PR #49, opened 2022-05-02
groups: add backend

liam-fitzgerald:
Adds a %groups agent with the following features and updates chat to use it for permissioning

Features:
- Roles
- Reader roles on each channel
- Writers roles for chats, inc. permission revocation
- A primitive joining flow (needs to expand for previews, doesn't respect permissions)


liam-fitzgerald:
Uncontrolled by default does sound nice, I'll look into it

PR #176, opened 2022-05-20
Po/initiate one on one dm

patosullivan:
Should be good to go, but there's something up with the dm const on line 220 of handlers.ts not matching the Handler def. Same thing is happening in master. Unrelated to this PR, though, so we should be good to merge.
Closes #86


liam-fitzgerald:
yeah

PR #205, opened 2022-06-03
chat-input: mobile formatting toolbar fix

rcrdlbl:
this resolves #73.

This code feels a bit messy to me, mostly due to ChatInputMenu getting split into two components as the megacomponent had too many discrete JSX sections and was hard to navigate – these components should probably be in two files imo, but I ran into issues w/ a dependency loop when trying to do so.

ways the menu could be shown/hidden conditionally on mobile:

- Seperate button to show/hide formatting tools (requires a good place for the "show formatting tools" button to live on mobile
- Automatically show/hide the menu when the text input box is focused
- automatically show/hide the input on select (pushing the text input up after something is selected is likely to cause a similar issue to the one that made moving the formatting toolbar necessary in the first place, this time with the native copy/paste covering up the user's text instead)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,29 @@
+---
+name: Bug report
+about: Is something broken?
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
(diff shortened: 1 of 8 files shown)


liam-fitzgerald:
in OOB discussion with Dan, it's been decided that the plus icon
misalignment is considered non-blocking, as he cannot reproduce it with his
version of iOS. I can repro and will fix this myself.

PR #490, opened 2022-07-28
groups: polymorphise properly for multi agent channels

liam-fitzgerald:
Introduces `$nest` which is simply a `(pair dude:gall flag)`

Uses this to refer to channels so that we aren't hardcoding references to chat.

Will need someone to pick up on the UI side of things, as part of the API has changed


liam-fitzgerald:
FE changes are included! merry christmas

PR #949, opened 2022-09-29
vscroll: switching to virtuoso

arthyn:
This replaces our virtual scroller with react-virtuoso. Seems to have less issues with getting stuck. Addresses #942

### DIFF
--- a/ui/src/chat/ChatScroller/ChatScroller.tsx
+++ b/ui/src/chat/ChatScroller/ChatScroller.tsx
@@ -9,15 +9,15 @@ import { differenceInDays } from 'date-fns';
 import { BigIntOrderedMap, daToUnix } from '@urbit/api';
 import bigInt from 'big-integer';
 import { Virtuoso } from 'react-virtuoso';
-import ChatWritScroller from './ChatWritScroller';
+import LoadingSpinner from '@/components/LoadingSpinner/LoadingSpinner';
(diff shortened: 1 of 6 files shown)


ui/src/chat/ChatScroller/ChatScroller.tsx
```diff
@@ -129,37 +167,53 @@ export default function ChatScroller({
...
-        />
-      ) : null}
+    <div className="relative h-full flex-1 p-2">
+      <Virtuoso
+        {...indexData}
+        followOutput
+        alignToBottom
+        className="h-full"
+        atBottomThreshold={250}
+        atTopThreshold={250}
```
tomholford, line 206: nit: would be nice to have a constant so we can tune this somewhere instead of inline magic numbers. alternatively could be calculated dynamically, e.g.  *x* number of messages * average message height

also, did you happen to play with the `overscan` prop to see if it affected the responsiveness and performance?

arthyn: No I didn't but might be worth it. It seemed pretty snappy as is, but could maybe be smoother.

liam-fitzgerald (approved):
sad to see it go, but probably the right move

nerveharp:
> sad to see it go, but probably the right move

If you truly love something...

PR #957, opened 2022-09-30
galleries: added option to sort by alphabet

### LINKED ISSUE #919: gallery: alpha/time sort broken
The alpha/time posted sort function in Galleries sends an event to the back-end, but the front-end doesn't respect it.

To reproduce:

Observe that the order does not change, but the poke goes through. Refreshing also does not re-order items.

rcrdlbl:
fixes #919.

this may be a bit rough as we're using the first content element to sort if an "official" title is unavailable (so many links will start with h as in https) but it does work.


arthyn (approved):
LGTM, maybe we can de-rank untitled links later

PR #958, opened 2022-09-30
groups: if group title text is white, use drop shadow

### LINKED ISSUE #939: groups: always add text-shadow to group cover if text is white

rcrdlbl:
fixes #939.


arthyn (approved):
LGTM

PR #963, opened 2022-10-03
groups: use landscape settings

nerveharp:
Dependent on urbit/garden#10.

Implements controls from Landscape to disable:
- avatars
- nicknames
- remote image fetching
- spellcheck

Manifests in the following product locations:
- Chat images
- Ship avatars
- Ship names
- Profile headers
- Notebook headers
- Notebook images
- Gallery items
- Notebook, gallery, profile bio, and chat text inputs
- Group icons
- Group headers

Also introduces a constant for the Landscape desk, since we're using `%garden` now and will eventually use `%landscape`.

fixes tloncorp/landscape-apps#936

### DIFF
--- a/ui/src/profiles/ProfileCoverImage.tsx
+++ b/ui/src/profiles/ProfileCoverImage.tsx
@@ -27,7 +27,7 @@ export default function ProfileCoverImage({
 }: CoverProps) {
   const contact = useContact(ship);
   const { disableRemoteContent } = useCalm();
-  const cover = contact || emptyContact;
+  const { cover } = contact || emptyContact;
 
   return (
     <div
--- a/ui/src/profiles/ProfileCoverImage.tsx
+++ b/ui/src/profiles/ProfileCoverImage.tsx
@@ -26,8 +26,8 @@ export default function ProfileCoverImage({
   children,
 }: CoverProps) {
   const contact = useContact(ship);
-  const calm = useCalm();
-  const cover = !calm.disableRemoteContent || contact || emptyContact;
+  const { disableRemoteContent } = useCalm();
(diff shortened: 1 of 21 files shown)


ui/src/profiles/ProfileCoverImage.tsx
```diff
@@ -25,7 +26,8 @@ export default function ProfileCoverImage({
   children,
 }: CoverProps) {
   const contact = useContact(ship);
-  const { cover } = contact || emptyContact;
+  const calm = useCalm();
+  const cover = !calm.disableRemoteContent || contact || emptyContact;
```
arthyn, line 30: I think something went wrong here this should have given a type error. Missing destructuring, but also the logic needs to be different I think.

arthyn (changes requested):
Just the one small issue, after that should be good to go

ui/src/profiles/ProfileCoverImage.tsx
```diff
@@ -26,16 +26,20 @@ export default function ProfileCoverImage({
   children,
 }: CoverProps) {
   const contact = useContact(ship);
-  const calm = useCalm();
-  const cover = !calm.disableRemoteContent || contact || emptyContact;
+  const { disableRemoteContent } = useCalm();
+  const cover = contact || emptyContact;
```
arthyn, line 30:
```suggestion
  const { cover } = contact || emptyContact;
```

arthyn (approved):
lgtm with suggestion

nerveharp:
> lgtm with suggestion

Thank you, _that's_ what you meant by destructuring.

PR #966, opened 2022-10-03
versioning: derisk deployment w/ poor man's versioning

liam-fitzgerald:
- Versions currently used marks starting at version `-0`
- Adds a %dummy mark
- Symlinks future upgrades of mark types to the %dummy mark, preventing kick-resubscribe loops, however this has poor UX, because the subscriber isn't aware that the versions are out of syncs. This technique also requires more effort when the marks are upgraded, because all subscriptions must be restarted by hand

This PR significantly derisks the deployment timeline, because once this is merged the only hard dependency for release is the actual OTA logic itself, which is fairly constrained in scope.

A full implementation of versioning is to follow in subsequent PRs

Reviewers:
- Want hard approval from @arthyn 
- @ryjm, @tomholford you should try to wrap your brains around this, it's worthwhile to understand why we're doing it and the necessity of version negotiation


arthyn (approved):
LGTM, my question would be why do we not require versioning for things like DMs and clubs is that simply because it's all pokes/scries?

ryjm:
spoke about this a little to liam but I really want to understand what steps we'd need to take to avoid having to manually update each (notwithstanding making everyone else do it). my naive assumption was to just write a gate that uses a `version:state` top level face that becomes the only change one needs to make their upgrade their marks. obviously this is easy from just a single player app, but i'd really like to understand what scaling / p2p syncing problem make a more elegant solution untenable. i'm currently assuming it's a fundamental design choice to avoid a much more malevolent problem, but that fact that i can't seem to nail it down. 

case in point - having to walk people through graphs store updates for each app was an experience i don't want to recereate, especially since i had to go trough a bunch of old ones and do the same. 

again this is coming from the right side of the bell curve but it's been nagging me for a while. @liam-fitzgerald gave me a good starting point to that today but i cant shake the feeling that there's a better way to do this. hopefully liam and i come to some sort understanding as he trashes my house tomorrow.

liam-fitzgerald:
We still need to do this for for DMs, but it's a less pressing issue, as it
doesn't materially impact the network as much

tomholford:
>@ryjm, @tomholford you should try to wrap your brains around this, it's worthwhile to understand why we're doing it and the necessity of version negotiation

Thanks for flagging this. Of course I am still a hoon n00b, but tead through the changes and seems pretty straightforward.

ryjm:
a hoon… what? someone needs to take a gander at the bible (you know the one)

nerveharp:
@ryjm @arthyn @tomholford We good here?

PR #967, opened 2022-10-03
devex: add scripts and config for use with bouncer

tomholford:
These are optional config files for use with [bouncer](https://github.com/tloncorp/bouncer).

Note that the paths in the yml files and bin scripts will need to be updated for your system.

The change includes an update to the gitignore for these directories. This way, others can tweak the bouncer config without introducing a git diff in their local repo. Eventually we can use ENV vars or similar to customize config for each individual system

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 node_modules
 .DS_Store
+ops
+bin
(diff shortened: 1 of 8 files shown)


arthyn (approved):
LGTM, seems like we need some way to change paths for people who have different setups, but we can do that later

nerveharp:
For sure, we can grab stuff from .env.local or whatever. Looks good!

Issue #801, opened 2022-09-08
talk: no pinning for chat channels


urcades:
Refer to the video below. 

I can pin arbitrary 1:1 DMs and Groups DMs, but I cannot pin group channels:

https://user-images.githubusercontent.com/1195363/189013806-49283977-3113-40d9-8bb6-c205cd9616e8.mov

It appears as though no "options dropdown menu" is surfaced for "group channels" in the Talk app.

I'm about to write up a "dropdown unification" issue I'll link this issue to momentarily.

nerveharp:
Needs back-end support.

PR #968, opened 2022-10-03
chat: added DMOptions dropdown to group chat channels

rcrdlbl:
This currently partially solves for #801 

I can send the request to pin a group chat channel, but the backend rejects it (think it'll only take DMs and MultiDMs at this point).

@liam-fitzgerald @arthyn 
mind helping resolve the backend portion of this?


arthyn:
fixes #801 no actual backend issues, just using the wrong keys

jamesacklin:
This seems fine, @patosullivan review?

ui/src/dms/MessagesSidebarItem.tsx
```diff
@@ -36,14 +40,7 @@ function ChannelSidebarItem({ whom, brief }: MessagesSidebarItemProps) {
     <SidebarItem
       to={`/groups/${groupFlag}/channels/${whom}`}
```
patosullivan, line 41:
```suggestion
      to={`/groups/${groupFlag}/channels/${nest}`}
```

If this is set to whom it won't route to the actual channel.

patosullivan (changes requested):
Looks good, just one issue.

patosullivan (approved):

Issue #785, opened 2022-09-07
talk: inviting someone into a 1:1/group dm causes the form to slightly jump in size


urcades:
This is a minor visual bug:

https://user-images.githubusercontent.com/1195363/188947467-532bd860-54a3-4560-8202-840605b2054e.mov

Note that when I select a name from the dropdown, the form field itself actually jumps a few pixels. The field should remain a static size as items are added to it.

PR #970, opened 2022-10-03
talk: set ShipSelector padding to avoid size jumping

rcrdlbl:
this one's about #785, it just needed a little bump in padding value to keep it from "filling up", and a corresponding mod to the button.

@urcades this is the simplest fix but it does make the input a tiny bit taller (as tall as it was after it "jumps" in the current version)


tomholford (approved):
LGTM

nerveharp:
Nice, thank you.

Issue #600, opened 2022-08-12
group/chat: mark channel as read on scrollback/load/etc


nerveharp:
Use the intersection of the last unread + vscroller height to update the brief and mark the channel as read.

rcrdlbl:
I'm having trouble reproducing this on the current master branch – @nerveharp could you try this again?

I think what might be missing is actually optimistic updating for the "mark read", and the lag on livenet is making it feel like it's not getting marked read (maybe?)

PR #975, opened 2022-10-05
chat: mark chat channel read when unread divider is visible on screen

rcrdlbl:
see: #600.

### DIFF
--- a/ui/src/chat/ChatMessage/DateDivider.tsx
+++ b/ui/src/chat/ChatMessage/DateDivider.tsx
@@ -1,17 +1,23 @@
 import React from 'react';
 import cn from 'classnames';
-import { makePrettyDay, pluralize } from '../../logic/utils';
+import { makePrettyDay, pluralize } from '@/logic/utils';
+import { useChatState } from '@/state/chat';
 
 interface DateDividerProps {
   date: Date;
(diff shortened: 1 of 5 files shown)


nerveharp (approved):
Gonna steal this one for private/future use, seems nice. Thanks!

ui/src/chat/ChatMessage/DateDivider.tsx
```diff
@@ -1,17 +1,23 @@
...
   date: Date;
   unreadCount?: number;
+  viewRef?: (node?: Element | null | undefined) => void;
 }
 
-export default function DateDivider({ date, unreadCount }: DateDividerProps) {
+export default function DateDivider({
+  date,
+  unreadCount,
+  viewRef,
```
arthyn, line 15: This really should use `React.forwardRef` instead

rcrdlbl: the library *might* not play well with it – but i'll poke around and see

rcrdlbl: Okay, just pushed.

arthyn (approved):
lgtm!

Issue #924, opened 2022-09-21
nav: small visual updates to sidebar row states


urcades:
Currently, when any of the rows in groups/talk are selected, and neighbor rows are hovered, we get a mildly unpleasant/unfinished-feeling effect in which the rows appear to visually blend together:

https://user-images.githubusercontent.com/1195363/191563740-268d157a-4863-409b-8970-bc7eb41d5b40.mov

In our latest sidebar designs, we've updated the hover behavior to account for this by 1) setting a one-shade-darker "selected" state for sidebar rows and 2) universally spacing out all row elements by ~2px:

https://user-images.githubusercontent.com/1195363/191564397-6ec6df00-6440-4b89-8c0d-c2b94f8b2787.mov

This is a small update, but it adds a significantly "more polished" feel to the sidebar's rows.

PR #977, opened 2022-10-05
sidebar: set active items one shade darker

rcrdlbl:
see: #924.


urcades (changes requested):
This is very close, with exception to the behavior of hovering over an already-selected item/row in the sidebar — we should probably omit selected rows from the onHover styling treatment, as there's no additional utility gained from hovering over a selected row (in the sense that clicking a selected row won't actually do anything)

To get a rough sense of the feeling we're trying to evoke here, you can check out this figma, and hover/click over any of the **groups** in the sidebar (the other rows don't do anything).

rcrdlbl:
@urcades updated!

urcades (changes requested):
The edits I see made to the top nav items in both Groups and Talk are perfect, I only see one last change that needs to be made in the list of DMs/Groups themselves:

https://user-images.githubusercontent.com/1195363/194147824-511e1a4e-bf07-444c-8e7c-d7774cd57d74.mov

You can see in the first part of this  video, the on-hover and highlighting behavior is as-specified, but as I hover the individual DMs, they remain butted up against one another — their onHover styling is perfect though.

All that remains is ensuring the small spacing is inherited by these row items as well!

urcades (approved):
Ignore the previous comment, @rcrdlbl alerted me to the fact that this PR is simply about the active item shading, not the row spacing.

rcrdlbl:
The latest commit should address your concern, @urcades

ui/tsconfig.rvpu8p5fi8d.json
```diff
@@ -0,0 +1,42 @@
+{
```
arthyn, line 1: Remove this file

arthyn (changes requested):
Just the one file to remove, then good to merge

ui/tsconfig.rvpu8p5fi8d.json
rcrdlbl: Done!

arthyn (approved):
lgtm

Issue #974, opened 2022-10-04
groups: pending invites appear twice, also in "Members"


nerveharp:
**To Reproduce**
1. Start a private group
2. Invite your buddies as part of the group setup wizard
3. Navigate to "Members & Group Info," click on the "Pending" tab
4. Observe duplicate invites (screenshot below).
5. Tab back over to "Members."
6. Observe all the invited ships.

**Expected behavior**
The invited ships should appear once in the "Pending" tab and only those who have accepted should appear in the "Members" tab.

**Screenshots**

Pending:
![image](https://user-images.githubusercontent.com/748181/193938398-e3a5331f-9242-4d1a-957b-060769836be6.png)

Members:
![image](https://user-images.githubusercontent.com/748181/193938658-42f3eff9-c6c3-469c-b423-643c026832c4.png)

**Additional context:**
Self-hosted, MacOS/Chrome

PR #980, opened 2022-10-05
groups: pending members show up where they should

rcrdlbl:
ref: #974 

the "concat" logic was off in the pending list leading to the list getting duplicated, and the members list needed a filter added to it, plus some restructuring so typescript wouldn't go off about the `Cordon` type.


nerveharp (approved):
Thank you, filter lord.

PR #981, opened 2022-10-05
diary: headers, lists, and rules

arthyn:
This implements #952 and additionally horizontal rules. It does not add any of the formatting UI to the menu only the extension support itself. so you can type them in, but we still need the extended formatting menu for notebooks @nerveharp .

Also adds state migration for diary so we don't blow up any state.


nerveharp:
Thank you, chef

arthyn:
we gotta cook

desk/sur/diary.hoon
```diff
@@ -120,14 +120,24 @@
...
 ::
 ::    %image: a visual, we record dimensions for better rendering
 ::    %cite: an Urbit reference
+::    %header: a traditional HTML heading, h1-h6
+::    %listing: a traditional HTML list, ul and ol
 ::
 +$  block
   $%  [%image src=cord height=@ud width=@ud alt=cord]
       [%cite =cite:c]
+      [%header p=?(%h1 %h2 %h3 %h4 %h5 %h6) q=inline]
```
liam-fitzgerald, line 138: should be `q=(list inline)`

liam-fitzgerald (changes requested):
Needs moar recursion

desk/sur/diary.hoon
arthyn: bah ok fine

liam-fitzgerald: come on..., it's like a 10 line change tops ;)

arthyn:
got verbal approval

PR #989, opened 2022-10-06
desk: fix kick resubscribe loop due to public/rolling mismatch

liam-fitzgerald:
The internal rolling release build is using numbered marks, whilst the public alpha is using unnumbered marks. As the public alpha lacks the mark files for the numbered marks and the rolling release lacks mark files for the unnumbered marks, this causes a kick-resubscribe loop whenever two ships on different builds attempt to communicate over a subscription. Rectifies this issue by returning to sending unnumbered marks over a subscription and adding the ability to receive numbered or unnumbered marks as pokes and %facts. Once this commit has been deployed to public alpha, we can return to giving numbered marks over subscriptions.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -141,7 +141,7 @@
     =+  !<(req=create:c vase)
     (create req)
   ::
-      %chat-action-0
+      ?(%chat-action-0 %chat-action)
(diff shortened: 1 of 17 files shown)


arthyn (approved):
lgtm

PR #992, opened 2022-10-06
quips: refactor

liam-fitzgerald:
- Moves quips inside the $seal data structure of a note, instead of existing separately to the noteId -> note mapping
- Changes the datatype of publish quips to be at parity with chat messages
- Reworks the fetching of notes so that we don't have to front load so much data

### DIFF
--- a/desk/app/diary.hoon
+++ b/desk/app/diary.hoon
@@ -258,24 +258,6 @@
   ^-  yarn:ha
   =/  id  (end [7 1] (shax eny.bowl))
   [id rope now.bowl con wer but]
-++  flatten
-  |=  content=(list inline:d)
-  ^-  cord
-  %-  crip
-  %-  zing
-  %+  turn
-    content
-  |=  c=inline:d
-  ^-  tape
-  ?@  c  (trip c)
-  ?-  -.c
-      %break  ""
-      %tag    (trip p.c)
-      %link   (trip q.c)
-      %block   (trip q.c)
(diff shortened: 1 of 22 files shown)


nerveharp:
Note to @liam-fitzgerald - add migrations for state

arthyn (approved):
LGTM, this is great will be ready once it has migrations

liam-fitzgerald:
Notes on the new OTA style

Previous OTA logic has been plagued by pollution of the namespace and the resultant copy paste to make the types align.

Skeleton of the new style
```
++  on-load
    |=  =vase
    =|  cards=(list card)
    |^  ^-  (quip card _this)
    =+  !<(old=versioned-state vase)
    |-
    ?-  -.old
      %1  [cards this(state old)]
    ::
      %0  $(old (state-0-to-1 old)
    ==
    ::
    +$  versioned-state
      $%  state-0
          state-1
      ==
    +$  state-0  [%0 =shelf:zero]
    ++  zero     zero:old:d
    +$  state-1  current-state
    ++  one      d
    ++  state-0-to-1
       ::  write your actual state adapter here
```
You should write all of the types in the state adapters with the numbered arms as the subject i.e. `diff:one` && `diff:zero`. This prevents future OTAs from having to touch code that is not actually being upgraded but needs editing to make it type check.
Hence, we could add another state adapter like by simply making `+one` evaluate to `one:old:d` and `+two` evaluate to `d`.

Issue #711, opened 2022-08-26
chat: loading state for pending messages


nerveharp:
When sending a message, we have three states:
1. Message has been sent from front-end to ship, but has not landed on ship yet
2. Ship has received message but has not been sent to group host/DM recipient
3. Group host/DM recipient has received the message

Currently, we only show the message once the group host/DM recipient has received it (state 3). This contributes to a sluggish or slow feeling in our new Chat channels.

This issue is to implement loading and confirmation states for in-flight messages.

Design: https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/Product-Handoffs-(Locked)?node-id=1292%3A69101

PR #993, opened 2022-10-07
chat: sent/delivered status for messages

rcrdlbl:
partially covers #711 

**Functionality that is covered:**
- Optimistic posting of sent messages from input field to ChatScroller
- "Sent from Browser, not delivered to host ship" and "Delivered to host shop" now have distinct visual states.

The effect on UX is that messages now feel like they post instantly now instead of having a percieved lag as the message makes its roundtrip from the host ship back to your urbit. (this is still so much faster than g1, it sometimes feels instant! but often it still doesn't!)

**Functionality that needs to be covered:**
- The distinct intermediary state of "sent from browser to user's ship, not yet reached host ship" detailed in issue #711
- design: https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/Product-Handoffs-(Locked)?node-id=1292%3A69107

 @arthyn and i have determined that some sort of backend update is necessary for this to be achievable in any reasonable + reliable fashion, so we'd prefer to break open a second issue to get this feature finished up.

I feel pretty strongly that this visually minor feature should be put in before general release – it's actually incredibly informative on *why* a message didn't deliver instantly (no more going to the terminal to `|hi` a group's host when messages don't seem to deliver! And people who wouldn't think to do this can get the same information at a glance!)

### DIFF
--- a/ui/src/chat/ChatMessage/ChatMessage.tsx
+++ b/ui/src/chat/ChatMessage/ChatMessage.tsx
@@ -15,7 +15,12 @@ import ChatContent from '@/chat/ChatContent/ChatContent';
 import ChatReactions from '@/chat/ChatReactions/ChatReactions';
 import DateDivider from '@/chat/ChatMessage/DateDivider';
 import ChatMessageOptions from '@/chat/ChatMessage/ChatMessageOptions';
-import { usePact, useChatState, useIsMessageDelivered } from '@/state/chat';
+import {
+  usePact,
+  useChatState,
+  useIsMessageDelivered,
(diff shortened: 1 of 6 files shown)


nerveharp:
👑, will test it out today

nerveharp (approved):
Feels really nice in situ, I encourage others to set their dev proxy to a live ship and test it out (~fabled-faster/interface-testing-facility is a real-enough case).

arthyn (approved):
lgtm with suggestions

liam-fitzgerald (approved):
Looks good! Ship it!

PR #994, opened 2022-10-07
sort DMs by most recent

rcrdlbl:
see: 988

apparently the "RECENT" flag in sortOptions was changed to sort by date, but in the wrong direction at some point. I'm reversing the sorted array, which is brute-forcey but works great.

### DIFF
--- a/ui/src/dms/MessagesList.tsx
+++ b/ui/src/dms/MessagesList.tsx
@@ -46,7 +46,8 @@ export default function MessagesList({ filter }: MessagesListProps) {
 
       return true; // is all
     })
-    .sort(sortOptions[RECENT]);
+    .sort(sortOptions[RECENT])
+    .reverse();
 
   return (
     <ul


ui/src/dms/MessagesList.tsx
```diff
@@ -46,7 +46,8 @@ export default function MessagesList({ filter }: MessagesListProps) {
 
       return true; // is all
     })
-    .sort(sortOptions[RECENT]);
+    .sort(sortOptions[RECENT])
+    .reverse();
```
liam-fitzgerald, line 50: this mutates in place, that's fine right, because there's only one reference to it?

liam-fitzgerald (approved):
minor nit about .reverse() being mutable LGTM!

ui/src/dms/MessagesList.tsx
rcrdlbl: this is the only place we're sorting DMs at the moment, and this is the only place we're doing this filtering logic w/ the list I'm 99% sure

rcrdlbl: DMs/Chat Channels/ETC

tomholford:
@rcrdlbl does this affect other instances of message or recent sorting?  could be worth refactoring `useMessageSort`'s usage of the `reverse` boolean

rcrdlbl:
so basically: the codebase doesn't use the sortMessages function useMessageSort exports anywhere, which is the only place that Talk uses the sortRecordsBy – sortMessages seems to straight up be broken in its current iteration.

we can keep the `reverse()` at the end of this as a quick fix (as we're not sorting stuff the "right" way here anyway) or hold off on merging and perform a more intensive fix on `useMessageSort`

tomholford:
> so basically: the codebase doesn't use the sortMessages function useMessageSort exports anywhere, which is the only place that Talk uses the sortRecordsBy – sortMessages seems to straight up be broken in its current iteration.
> 
> 
> 
> we can keep the `reverse()` at the end of this as a quick fix (as we're not sorting stuff the "right" way here anyway) or hold off on merging and perform a more intensive fix on `useMessageSort`

Given the timeline, let's keep the quick fix

PR #995, opened 2022-10-07
ui: make virtuoso set keys on components properly

liam-fitzgerald:
this & an actual production build would do wonders for the virtual scroller


arthyn (approved):
lgtm, our builds should be production now?

liam-fitzgerald:
Screen Shot 2022-10-07 at 6 08 43 pm
react devtools only work with development mode react

liam-fitzgerald:
oh wtf, since when is devtools compatible with production

PR #999, opened 2022-10-11
ops: editing actions to work with ref passing

arthyn:
OTT, this uses the latest edits in janeway so that when we manually deploy we can enter a tag to deploy. I'm not confident in the bash part if anyone knows if that will work to conditionally pass the argument.


arthyn:
Test fixed in other PR

Issue #782, opened 2022-09-07
groups: "canceling a join" of a group that doesn't exist is broken


urcades:
**Describe the bug**
After attempting to join a nonexistent (or slow/offline) group, I can't use the provided interface to cancel the join

**To Reproduce**
1. Ask someone to invite you to a nonexistent or offline group, or attempt to join one with a random endpoint
2. Click on the pending group row in the sidebar
3. Try clicking 'cancel join' — If it doesn't work, you've repro'd the bug

**Expected behavior**
No matter the state of the join, or why it's persisting, I should be able to cancel out of a join by clicking this button once. This is one of those interfaces that can't feel fiddly — it needs to perform what it says on the label, or else it undermines the trust of the interface (this applies to all "cancel"-type actions)

**Screenshots**

https://user-images.githubusercontent.com/1195363/188934121-38843a65-949d-47dc-b946-fe606f18e83c.mov

nerveharp:
I also experience this daily.

PR #1002, opened 2022-10-13
groups: hide group joins with "error" states

rcrdlbl:
see: #782.

group joins that have an "error" state (due to nonexistent target group etc) were still appearing, this PR prevents them from appearing. tradeoff: the errored joins still technically persist in the backend, and could potentially build up and affect performance after long periods of use

the ideal fix for this requires backend changes (a separate "possible error/nonexistent group" visual state/tooltip text arrangement, with a cancel button).

unfortunately the only function to cancel a group join currently doesn't seem to work if the join process itself has errored – if we desire this functionality soon, we should create a separate issue for it.


arthyn (approved):
lgtm, I'll check out the backend to see what we can do

PR #1004, opened 2022-10-13
mobile: route based nav

arthyn:
This converts our nav to being route based instead of detached, addressing #940. Also includes a change to `useMedia` so that we aren't constantly creating listeners. As soon as the first is created any subsequent calls skip and just the value from the first.

The only thing I'm unsure about is if all the links now point correctly if someone can help test.


patosullivan (changes requested):
The "back" link in the channel header doesn't work in Gallery or Notebook (both of which work in master), otherwise looks good.

patosullivan (approved):

Issue #984, opened 2022-10-06
groups: none of the "copy" actions work in safari


urcades:
**Describe the bug**
At the moment, none of the copy actions/buttons in the product appear to function in safari. I've tried out the buttons in a chromium-based browser (Arc) and they appear to function as intended

**To Reproduce**
1. Get on an Apple device and access your urbit with safari
2. Click any of the copy actions in the interface
3. Try and paste the copied name/endpoint into a chat channel
4. If you don't see anything, or see your clipboard's prior contents, you've replicated the bug

**Expected behavior**
I expect copy actions to function on all browsers

**Screenshots**

https://user-images.githubusercontent.com/1195363/194221083-ff988e45-6b70-406b-a438-a75244e75bf1.mov

**Additional context:**
- Ship: planet (~fabled-faster)
- Instance type: Tlon-hosted
- MacOS Monterey — 12.6 (21G115)
- Safari — 15.6.1 (17613.3.9.1.16)

nerveharp:
@ryjm identified that turning on HTTPS fixes this, but normal HTTP exhibits this behavior.

rcrdlbl:
@urcades 
It works fine for me on normal http – i'm unable to replicate this on safari version 16.0 on the current master branch.

note the notebooks/galleries "share" button isn't hooked up to a copy action due to a decision we made in the lead-up to the demo path

there is however some weirdness when attempting to paste a link that would result in a reference into a gallery comment i've noticed (i think the parser has the same rules applied as it would to chat channels), but that's a seperate issue.

PR #1007, opened 2022-10-14
groups: support copying to clipboard without HTTPS

rcrdlbl:
see #984.

The issue was actually with HTTP and not safari, specifically with HTTP's interaction with the recommended API for clipboard interaction, `navigator.clipboard.writeText`, which only supports HTTPS.

I'm bringing over the useCopy hook from old Groups, which requires a few changes to the rest of the code, and also depends on a function that while it does work, is technically deprecated

(so we gotta change it in the near future, which involves forcing HTTPS on landscape, which is mad out of scope of this PR or our release timeline)


patosullivan (approved):
Looks/works great!

PR #1008, opened 2022-10-14
groups: version negotiation

ryjm:
Follows the flow defined in `%chat`.


desk/app/groups.hoon
```diff
@@ -2,21 +2,39 @@
...
+        cool=epic:e
+        groups=(map flag:g [net:g group:g])
+        xeno=(map flag:g gang:g)
+    ==
+  +$  versioned-state  $%(state-0 state-1)
+  ::
+  ++  jack
+    |=  v=versioned-state
+    ?-  -.v
+      %0  [%1 cool=*epic:e +.v]
```
arthyn, line 31: This doesn't fully migrate because you need to migrate `%net`

desk/app/groups.hoon
```diff
@@ -2,21 +2,39 @@
...
+      %1  v
+    ==
+  ::
+  +$  adapted-state  $&(versioned-state jack)
```
arthyn, line 35: Can you explain what this does?

desk/app/groups.hoon
```diff
@@ -149,6 +163,26 @@
     =+  !<(=flag:g vase)
     ga-abet:ga-invite-reject:(ga-abed:gang-core flag)
   ==
+::  +load: load next state
+++  load
+  |=  =vase
+  ^+  cor
+  =+  !<(old=versioned-state vase)
+  =?  old  ?=(%0 -.old)
+    ;;(adapted-state !<(versioned-state vase))
```
arthyn, line 172: Can you explain how this works with the previous comment?

arthyn (changes requested):
Have a few questions. Also, this needs to fully migrate the `%net` state change.

desk/app/groups.hoon
ryjm: we `!<` the `vase` as a `versioned-state` so that we can dispatch off the old version (i.e if we need to pass cards when upgrading), and since `adapted-state` is still a `versioned-state` you can `;;` it to update it to the latest state definition as described in the previous comment.

this lets you cleanly separate the state update from any effects that need to be triggered during the update by defining the former in the `+jack` gate and the latter in `+on-load`. if you don't need to trigger any effects you just can just `;;` and be done.

desk/app/groups.hoon
ryjm: ah right, will fix.

desk/app/groups.hoon
ryjm: line 172 should probably just be `;;(adapted-state old)` though.

arthyn: ah ok, this is cool

ryjm:
closing in favor of #1031

PR #1012, opened 2022-10-14
s3: generic upload component

jamesacklin:
Adds S3 upload, cribbing a lot of work from https://github.com/arthyn/silo.

Introduces an otherwise unremarkable HTML file input component which accepts a forwarded ref. The thinking here is that we could use this elsewhere in the interface as a hidden element, pass the input an event the ref, and use the provided FileStore however we see fit.

### DIFF
--- a/ui/src/components/FileUpload.tsx
+++ b/ui/src/components/FileUpload.tsx
@@ -1,94 +1,97 @@
-import React, { useCallback, ChangeEvent, useEffect } from 'react';
+import React, { useCallback, ChangeEvent, useRef, useEffect } from 'react';
+import _ from 'lodash';
 import { PutObjectCommand } from '@aws-sdk/client-s3';
 import { dateToDa, deSig } from '@urbit/api';
-import { useFileStore, useS3Creds, useS3Config, Upload } from '@/state/storage';
+import { useFileStore, useStorage } from '@/state/storage';
+import { Upload, UploadInputProps } from '@/types/storage';
+import api from '../api';
 
-interface UploadInput {
-  multiple?: boolean;
-}
+const UploadInput = React.forwardRef<HTMLInputElement, UploadInputProps>(
+  (props, ref) => {
+    const { multiple, ...rest } = props;
+    const { s3, ...storage } = useStorage();
+    const { credentials } = s3;
+    const { setFiles, createClient, setStatus, ...fs } = useFileStore();
 
-export default function UploadInput(props: UploadInput) {
-  const { multiple } = props;
-  const { files, createClient, ...fs } = useFileStore();
-  const s3creds = useS3Creds();
-  const s3config = useS3Config();
+    useEffect(() => {
+      useStorage.getState().initialize(api);
+    }, []);
 
-  useEffect(() => {
-    const connect = async () => {
-      const client = createClient(s3creds);
-      return client;
-    };
-    connect()
-      .then(() => fs.setStatus('success'))
-      .catch((error: unknown) => {
-        console.error(error);
-      });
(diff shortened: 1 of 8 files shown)


liam-fitzgerald:
pls watch the `package-lock` thrashing!

ui/src/components/FileUpload.tsx
```diff
@@ -0,0 +1,107 @@
...
+      },
+      [fs, s3]
+    );
+
+    useEffect(() => {
+      files.forEach((upload: Upload, idx: number) => {
+        uploadFile(upload, idx);
+      });
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [files.length]);
```
liam-fitzgerald, line 78: accessors don't work properly in dependencies afaik?

ui/src/components/FileUpload.tsx
```diff
@@ -0,0 +1,107 @@
...
+              new PutObjectCommand({
+                Bucket: s3.configuration.currentBucket,
+                Key: `${window.ship}/${timestamp}-${file.name}`,
+                Body: file,
+                ContentType: file.type,
+                ContentLength: file.size,
+                ACL: 'public-read',
+              })
+            )
+            .then(() => {
```
liam-fitzgerald, line 58: please don't mix `await` and `.then()` unless absolutely necessary

ui/src/components/FileUpload.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    useEffect(() => {
+      const hasCredentials =
+        credentials?.accessKeyId &&
+        credentials?.endpoint &&
+        credentials?.secretAccessKey;
+      if (hasCredentials) {
+        const connect = async () => {
+          const client = createClient(credentials);
+          return client;
```
liam-fitzgerald, line 30: if you return a promise directly, the calling function doesn't need to be `async`, the result is still awaitable though

ui/src/components/FileUpload.tsx
```diff
@@ -0,0 +1,107 @@
...
+      },
+      [fs]
+    );
+
+    if (storage.loaded)
+      return (
+        <input
+          type="file"
+          multiple={multiple}
+          onChange={(e) => onFiles(e)}
```
liam-fitzgerald, line 97: drop unnecessary arrow function

liam-fitzgerald (changes requested):
looks mostly good, but some of the async work needs fixing

ui/src/components/FileUpload.tsx
```diff
@@ -0,0 +1,107 @@
...
+const UploadInput = React.forwardRef<HTMLInputElement, UploadInputProps>(
+  (props, ref) => {
+    const { multiple, ...rest } = props;
+    const { files, createClient, ...fs } = useFileStore();
+    const { s3, ...storage } = useStorage();
+    const { credentials } = s3;
+
+    useEffect(() => {
+      async function init() {
+        useStorage.getState().initialize(api);
```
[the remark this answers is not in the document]
jamesacklin: Calling directly now.

ui/src/components/FileUpload.tsx
jamesacklin: Cleaned this up a bit.

ui/src/components/FileUpload.tsx
jamesacklin: Cleaned this up, removed the `await`

ui/src/components/FileUpload.tsx
jamesacklin: Yes chef

patosullivan (approved):
LGTM!

PR #1015, opened 2022-10-17
chat: fix several issues with drafts

### LINKED ISSUE #978: Talk: when returning to a chat with unsent text, it inserts an extra space
**Describe the bug**
when returning to a chat with unsent text, it inserts an extra space
savetextissue

**To Reproduce**
1. Got a chat
2. Start typing a message
3. Go to a different chat message (don't send message)
4. Go back to the previous chat message (with unsent text)

**Expected behavior**
It shouldn't add any space there

**Screenshots**

**Additional context:**
- Ship type (comet/moon/planet/star/galaxy)
- Instance type (Tlon-hosted/self-hosted with Port/CLI)
- OS/browser (Safari on MacOS, Chrome on Windows, Firefox on Linux, etc.)

patosullivan:
Fixes the following issues with drafts:
- Adding a new line when returning to a draft (fixes #978)
- Not showing most recent draft if more than one draft has been created
- Unnecessarily updating drafts (when new draft is the same as the old draft)

There may be an issue with the backend code causing these empty paragraphs/breaks to get added to messages. If that's the case, it'll obviate most of the code here.

### DIFF
--- a/ui/src/chat/ChatInput/ChatInput.tsx
+++ b/ui/src/chat/ChatInput/ChatInput.tsx
@@ -1,9 +1,9 @@
 import { Editor } from '@tiptap/react';
-import { debounce } from 'lodash';
+import { debounce, isEqual } from 'lodash';
 import cn from 'classnames';
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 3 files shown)


arthyn (approved):
lgtm!

PR #1020, opened 2022-10-20
groups: set notifications sidebar item as active in default route

### LINKED ISSUE #928: nav: on first load, brought to "app notifications" page, but sidebar item isn't in a "selected" state
When I first load "Groups" via Landscape, I'm brought to the following screen:

Screen Shot 2022-09-21 at 5 07 05 PM

Notice that the sidebar row for "Notifications" is not set to a "selected" state, it should appear as so:

Screen Shot 2022-09-21 at 5 09 44 PM

patosullivan:
Fixes #928


jamesacklin (approved):
Looks good to me, pending @arthyn's stamp, should be g2g.

arthyn (approved):
lgtm

PR #1022, opened 2022-10-21
groups: set heap display mode in settings store

### LINKED ISSUE #926: heap: list/grid view a writer-setting
This is evident when arranging by list or grid in a channel you don't have permission to edit.

To reproduce:
1. Start a Gallery channel in a group on Ship A, an administrator. Mark it as "Members can view."
2. Visit this Gallery using Ship B, a group member. Interact with the grid|list view setting.
3. Observe poke failure, no UI change.

Expected result:
This should be a _per-user setting_ that persists between refreshes/sessions.

patosullivan:
Fixes #926 

We should allow the user to set which mode they'd prefer to view a gallery with on a per-user basis. Previously it was set on a per-gallery basis.

### DIFF
--- a/ui/src/heap/HeapChannel.tsx
+++ b/ui/src/heap/HeapChannel.tsx
@@ -14,7 +14,6 @@ import {
 } from '@/state/groups/groups';
 import {
   useCuriosForHeap,
-  useHeapDisplayMode,
   useHeapState,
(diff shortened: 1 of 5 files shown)


jamesacklin (approved):
Makes perfect sense, thanks!

PR #1023, opened 2022-10-21
groups: add ability to remove potential invite from invite list on new group creation

### LINKED ISSUE #599: groups/create: impossible to remove ships from invite list
When creating a new group you're presented with the opportunity to invite other users to the group. If you add a ship to that list there is no way to remove them from the list if you change your mind, you just have to skip the invite process all together.

patosullivan:
fixes #599 

@urcades @jyng 
thoughts on how this looks/works?

https://user-images.githubusercontent.com/1221094/197270463-6231382d-4dca-425d-8513-949afe6c05de.mov


jamesacklin (approved):
Perfect

PR #1026, opened 2022-10-24
s3: key FileStore on input ID

jamesacklin:
Embellishes the FileStore a bit by:
- Adding a pre-signed URL for all uploads
- Keying the FileStore, rather than referring to uploads by index
- Adds a 'for' field and a required ID on all image inputs
- Adds an example implementation to show how to produce a (or a list of) URL(s) from a file input

As a result, the component supports multiple inputs on the same page.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -10,6 +10,7 @@
       "hasInstallScript": true,
       "dependencies": {
         "@aws-sdk/client-s3": "^3.190.0",
+        "@aws-sdk/s3-request-presigner": "^3.190.0",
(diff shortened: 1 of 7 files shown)


arthyn (approved):
lgtm, I'm not entirely sure we need the presignedUrl since these are uploaded as public-read, but I can't really find what all capabilities you get from them other than access so I guess it's fine. I like the per file status tracking!

jamesacklin:
@arthyn Thanks, gonna merge so I can hand off to @patosullivan

PR #1027, opened 2022-10-24
vscroll: useMemo on objects generated on render that are deps

arthyn:
OTT, bah can't believe I didn't catch this before. Fixes issues @patosullivan and I were seeing trying to access the forge.


patosullivan (approved):
Looks good, would just remove the commented out console.log.

PR #1028, opened 2022-10-25
chat: add image upload (and pasting of image URLs) to chat input

### LINKED ISSUE #160: groups + chat: handle images in input component
This is by far the most complicated manifestation of #930, so it deserves its own case.

The expected behavior:

0. Assume that the user has already set up an S3 bucket in Landscape.
1. They click the "plus" button in the message input field.
2. A popup prompts them to upload a file or start a sig command. (Leave the sig command option off until we have more info on that feature.)
3. The desktop file browser lets them choose a file (let's allow our standard file extensions that we support in Gallery - png, jpeg, webm, mp3, etc).
4. The image is uploaded to S3 in the background and a placeholder appears in the message input field above the text input. (We're missing a loading state in the designs, but let's assume a square with a spinner in the middle and the "X" as a cancel button.)
5. Once complete, the image is added as block content with a preview. The "X" button should remove the block element from the message.
6. Press Send or Enter to send the image to the chat.

Design: https://www.figma.com/file/mdozWm05YsYcUdmYcsWcLA/2022-Q2-Talk?node-id=1340%3A51556

If step 0 above is not true and the user doesn't have S3 set up, show a popup explaining that they need to do so: https://www.figma.com/file/mdozWm05YsYcUdmYcsWcLA/2022-Q2-Talk?node-id=1340%3A51574

patosullivan:
[mostly] closes #160 

Added a new hook called useFileUpload that can be used anywhere we want to dynamically insert a hidden upload input.
Added the ability to upload images to chat.
Added the ability to paste an image URL into the chat input and have it get sent as an Image block type.
Added some loading/error states around chat input.

Note:
Does not include pasting of images directly into chat from the clipboard.
Does not show the image as a preview, just adds the image URL to the chat input.

https://user-images.githubusercontent.com/1221094/198136649-fba0f2f4-4878-461b-a71e-50c1cb008ed5.mov

https://user-images.githubusercontent.com/1221094/198136787-29971834-778b-420e-b35e-284d8949f8cb.mov

### DIFF
--- a/ui/src/logic/useFileUpload.ts
+++ b/ui/src/logic/useFileUpload.ts
@@ -73,8 +73,11 @@ function useFileUpload() {
         })
         .catch((error: any) => {
           setFileStatus([key, 'error']);
-          setErrorMessage([key, `${error.toString()}`]);
-          console.error(error);
+          setErrorMessage([
+            key,
+            `S3 upload error: ${error.message}, check your S3 configuration.`,
+          ]);
+          console.log({ error });
         });
 
       return uploadData;
--- a/ui/src/logic/useFileUpload.ts
+++ b/ui/src/logic/useFileUpload.ts
@@ -82,9 +82,7 @@ function useFileUpload() {
       const newFiles = _.keyBy(
         [...e.target.files].map((file) => ({
           file,
-          key: `${window.ship}/${deSig(
-            dateToDa(new Date())
-          )}-${encodeURIComponent(file.name)}`,
+          key: `${window.ship}/${deSig(dateToDa(new Date()))}-${file.name}`,
           for: e.target.id,
           status: 'initial',
           url: '',
--- a/ui/src/logic/useFileUpload.ts
+++ b/ui/src/logic/useFileUpload.ts
@@ -46,32 +46,38 @@ function useFileUpload() {
   const uploadFile = useCallback(
     async (upload) => {
       const { file, key } = upload;
-      if (fs.client) {
-        setFileStatus([key, 'loading']);
-        const command = new PutObjectCommand({
-          Bucket: s3.configuration.currentBucket,
-          Key: key,
-          Body: file,
-          ContentType: file.type,
-          ContentLength: file.size,
-          ACL: 'public-read',
-        });
-        const url = await getSignedUrl(fs.client, command);
-        const uploadData = fs.client
-          .send(command)
-          .then(() => {
-            setFileStatus([key, 'success']);
-            setFileURL([key, url.split('?')[0]]);
-          })
-          .catch((error: any) => {
-            setFileStatus([key, 'error']);
-            setErrorMessage([key, `${error.toString()}`]);
-            console.error(error);
-          });
-        return uploadData;
+
+      if (!fs.client) {
+        setStatus('error');
+        return false;
       }
-      setStatus('error');
-      return false;
+
+      setFileStatus([key, 'loading']);
+
+      const command = new PutObjectCommand({
+        Bucket: s3.configuration.currentBucket,
+        Key: key,
+        Body: file,
+        ContentType: file.type,
+        ContentLength: file.size,
+        ACL: 'public-read',
+      });
+
+      const url = await getSignedUrl(fs.client, command);
+
+      const uploadData = fs.client
+        .send(command)
+        .then(() => {
+          setFileStatus([key, 'success']);
+          setFileURL([key, url.split('?')[0]]);
+        })
+        .catch((error: any) => {
+          setFileStatus([key, 'error']);
+          setErrorMessage([key, `${error.toString()}`]);
+          console.error(error);
+        });
+
+      return uploadData;
     },
     [fs, s3, setStatus, setFileStatus, setFileURL, setErrorMessage]
   );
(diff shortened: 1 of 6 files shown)


ui/src/logic/useFileUpload.ts
```diff
@@ -0,0 +1,121 @@
...
+        .catch((error: unknown) => {
+          console.error(error);
+        });
+    }
+  }, [createClient, setStatus, credentials]);
+
+  const uploadFile = useCallback(
+    async (upload) => {
+      const { file, key } = upload;
+      if (fs.client) {
```
arthyn, line 49: I think flipping this check and early returning would help make this a little more readable

ui/src/state/storage/upload.tsx
```diff
@@ -20,7 +20,8 @@ export const useFileStore = create<FileStore>((set) => ({
         hostname: endpoint.host,
         path: endpoint.pathname || '/',
       },
-      region: 'global',
+      // this region is necessary for compatibility with other S3 providers (i.e., filebase)
+      region: 'us-east-1',
```
arthyn, line 24: we're going to have contention here until we request an actual region in our storage creds and use that.

ui/src/state/storage/upload.tsx
```diff
@@ -11,7 +11,7 @@ export function prefixEndpoint(endpoint: string) {
 export const useFileStore = create<FileStore>((set) => ({
   client: null,
   status: 'initial',
-  files: {},
+  files: {} as FileStoreFile[],
```
arthyn, line 14: Types unnecessary here if we type the other file correctly

ui/src/types/storage.ts
```diff
@@ -36,12 +36,22 @@ export interface Upload {
...
+  url: string;
+  for: string;
+  key: string;
+}
+
 export interface FileStore {
   client: S3Client | null;
   status: 'initial' | 'loading' | 'success' | 'error';
-  files: object;
+  files: FileStoreFile[] | null;
```
arthyn, line 51: This should be `Record<string, FileStoreFile>`

arthyn (changes requested):
A few small things otherwise lgtm. we can tidy this a bit once we do the previews and the logic has a place to live.

urcades (changes requested):
Hi @patosullivan , this is looking amazing — the only suggestion I have is a subtle tweak to the existing error-surfacing behavior you've built: 

Screen Shot 2022-10-26 at 6 38 25 PM

https://www.figma.com/file/7hegPepDf9wsRfLvciOI04/2022-Q3-Talk?node-id=1705%3A38431

~

What's being illustrated here is that once an upload has failed, we can immediately surface a non-blocking error **that dismisses if anything besides the single (optional) button is clicked**. This error should additionally only appear for ~5 seconds if no actions are taken.

The reasoning for this suggestion is that we'd like to be informative as to what's going on under the hood, to answer why a particular image upload has failed, but we don't want to block further interactivity if someone just wants to post an image.

Regarding the error element itself, we have some prior art for the styling here, displayed when a group is being joined and you click on the pending join in the sidebar.

The element should have a modular design, with affordances for extra body copy and a button leading to support/docs, which should exist eventually:

Screen Shot 2022-10-26 at 6 43 24 PM

ui/src/state/storage/upload.tsx
patosullivan: Agreed. Think we should go ahead and just add it?

patosullivan:
> Hi @patosullivan , this is looking amazing — the only suggestion I have is a subtle tweak to the existing error-surfacing behavior you've built:
> 
> Screen Shot 2022-10-26 at 6 38 25 PM
> 
> https://www.figma.com/file/7hegPepDf9wsRfLvciOI04/2022-Q3-Talk?node-id=1705%3A38431
> 
> ~
> 
> What's being illustrated here is that once an upload has failed, we can immediately surface a non-blocking error **that dismisses if anything besides the single (optional) button is clicked**. This error should additionally only appear for ~5 seconds if no actions are taken.
> 
> The reasoning for this suggestion is that we'd like to be informative as to what's going on under the hood, to answer why a particular image upload has failed, but we don't want to block further interactivity if someone just wants to post an image.
> 
> Regarding the error element itself, we have some prior art for the styling here, displayed when a group is being joined and you click on the pending join in the sidebar.
> 
> The element should have a modular design, with affordances for extra body copy and a button leading to support/docs, which should exist eventually:
> 
> Screen Shot 2022-10-26 at 6 43 24 PM

Thanks! So we should just show this popover and not use the exclamation point in place of the plus symbol at all, correct?

patosullivan:
> @patosullivan yes you are 100% correct

Ok, updated. How does this look? Note that I've commented out the button in the code I've pushed up because we don't have anywhere to link to at the moment afaik.

https://user-images.githubusercontent.com/1221094/198346822-ddebedce-0a34-435e-b014-edddbfbc3549.mov

urcades (approved):
This interaction is looking excellent per the last video — I think it could use one more standard increment of nudging upwards, to keep it from butting against the button and field, but that's a finer detail.

patosullivan:
@arthyn are we good to merge this before the s3 changes?

arthyn:
@patosullivan yeah that's fine

arthyn (approved):

jamesacklin:
Huge, thank you

PR #1032, opened 2022-10-27
groups: suppress superfluous notifications

### LINKED ISSUE #754: groups/notifications: superfluous notifications of ships joining + leaving on member ships
As a member of a group - not an administrator - I receive notifications for every other member joining and leaving the group.

Steps to reproduce:
1. Join a mainnet group you do not administer as a member
2. Using a second mainnet ship, join the same group
3. Observe the join notification on your first mainnet member ship

Expected behavior:
I should only see these notifications if I am an administrator of the group in question.

tomholford:
# Context

This closes #754 by filtering certain notifications to only be sent to Admin users.

# Changes

Suppress the following notifications for non-Admin users (ie, those not in the bloc):

- joined channel
- requested to join
- left channel
- member role change (user is now an ...)

# Preview

*On Admin ship*
image

*On Member ship*
image

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -77,6 +77,11 @@ This project uses husky to run git pre-commit hooks. You may disable Husky by ad
 
 ## Glossary
 
+_bloc_
+: superuser sects
+
+sects in the bloc set are allowed to make modifications to the group, and its various metadata and permissions
+
 _brief_
 : a representation of the last thing we've seen, and how many new items since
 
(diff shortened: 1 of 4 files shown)


tomholford:
@jamesacklin ended up filtering out 4 notifications from non-Admins:

- joined channel
- requested to join
- left channel
- member role change (user is now an ...)

Is this expected behavior for non-Admin users? And are there any notifications we should add to or remove from this list?

jamesacklin:
This is correct; I don't think we need to yank any others. It's good that we have the `bloc` construct if we need to, thank you.

arthyn (approved):
lgtm!

Issue #1005, opened 2022-10-13
nav: make mobile match desktop 


arthyn:
A few sections are unnavigable on mobile:

- main
  - profile
  - find groups
  - create group
- group 
  - all channels
  - group actions

@urcades these are the ones I found so far

PR #1033, opened 2022-10-27
nav: improving mobile

arthyn:
This addresses #1005 by adding more routes to the bottom nav. I decided it would be difficult to represent the difference between all channels and joined channels so I added that one to the top right header only on the channels page.

This also throws in some mobile styling tweaks to headers that @urcades was working on, and some improvements to how the virtual scroller manages keys/data and loading.

### DIFF
--- a/ui/src/chat/ChatMessage/ChatMessage.tsx
+++ b/ui/src/chat/ChatMessage/ChatMessage.tsx
@@ -64,7 +64,7 @@ const ChatMessage = React.memo<
       const isMessagePosted = useIsMessagePosted(seal.id);
 
       useEffect(() => {
-        if (inView === true && unread) {
+        if (inView && unread) {
           markRead(whom);
         }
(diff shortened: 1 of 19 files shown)


ui/src/chat/ChatMessage/ChatMessage.tsx
```diff
@@ -64,10 +64,10 @@ const ChatMessage = React.memo<
       const isMessagePosted = useIsMessagePosted(seal.id);
 
       useEffect(() => {
-        if (inView === true) {
+        if (inView === true && unread) {
```
tomholford, line 67:
```suggestion
        if (inView && unread) {
```

tomholford (approved):
Pulled and played with it locally, LGTM. Nice work @arthyn 

image

image

jamesacklin:
Feel free to yank the bug report star from the bottom nav.

Issue #788, opened 2022-09-07
groups: significant UX issues with group creation


urcades:
**Describe the bug**
At the moment, groups creation is littered with a lot of small issues that compound into a larger issue of groups creation feeling needlessly baroque/heavy. 

Edit: While this isn't a demo path concern, it's still a rather core set of interfaces that we can stand to improve for a v1 release.

In the following video, I attempt to create a group:

https://user-images.githubusercontent.com/1195363/188939772-60ef5350-a206-4aec-8512-1cce18b5083f.mov

- At 4 seconds, we already begin to see some issues with the form field design: The insertion point isn't automatically inserted into the "Group Name" field — I should be able to immediately begin typing the moment this field appears
- At 0:29 seconds I attempt to create my group after naming it and granting it a description, only to run into a wall — while the interface (barely) makes it clear that setting an avatar is a required action, the fact that an avatar needs to be explicitly provided in order to proceed feels close to a dark pattern, or at least a needlessly halting one. The "avatar" field should default to the "Letter" + Color presentation, so that all I need to do to proceed is give my group a name
- At 0:35 seconds, when I click into the "Group Icon" field, and click out of it to reset the state  (an expectation), it doesn't work. I need to explicitly press the "X" button to clear the field
- At 0:39 seconds, when I finally select the "Choose fill color" option, the default "black" icon color apparently isn't enough to proceed. Even _after_ setting the color of the icon at 0:51, I can't proceed to create my group.
- It isn't until 0:59 where I finally realize that I need to edit my group's name in order to proceed — this is almost certainly a bug

It nearly takes a minute for me to simply proceed from setting my group's default information — this is pretty insane, and certainly a blocker in getting people to the result they actually expect. Moving onward:

- Selecting a Group's privacy level is pretty straightforward, nothing particularly wrong here
- At 1:11 we run into an interface that visually blocks you from proceeding unless you invite people to the group. I think this tertiary "Skip" pattern is an antipattern that detracts from an otherwise straightforward interface.
- At 1:20, I'm dumped into a blank "groups" screen, rather than the group I just created

Combined, the above issues make for a creation process that feels more difficult than it needs to be.

There are a few updates to the groups creation flow that are on the cusp of being formalized:

**Group Details:**

<img width="1440" alt="Frame 2429" src="https://user-images.githubusercontent.com/1195363/188954778-e9b16169-0b32-4248-923a-029accdf6f98.png">

**Group Privacy**

[...]

Issue #651, opened 2022-08-19
groups/join: search autocomplete lists match at top and still includes match in the list below


lukestiles:
**Describe the bug**
When searching for a user or group, the match is inserted at the top, and remains in the list below. 

**To Reproduce**
1. Search for ~nocsyx-lassul
2. Watch the list down select while typing the name
3. ~nocsyx-lassul appears as the top choice 
4. ~nocsyx-lassul also appears at the bottom of the list

**Expected behavior**
I would expect to only see the match once, either at the top of the list for easy selection or preselected in some other sort order in the full list. 

**Screenshots**
<img width="1515" alt="Screen Shot 2022-08-19 at 8 52 29 AM" src="https://user-images.githubusercontent.com/1903610/185658876-31c64792-1502-4a55-8c52-f3a5402ebaa4.png">

**Additional context:**
- Searching from ~famhut-wolpyx
- Digital Ocean self-hosted
- Chrome on Mac

arthyn:
This isn't really a bug. This was only ever intended to be a ship search. Do we really need to do this?

PR #1046, opened 2022-10-29
groups: rework of group creation flow

rcrdlbl:
see: #788.

there are still a few things I may have missed with this w/r/t styling, but everything seems to work well.

I still wasn't able to figure out what was causing #651, however.


ui/src/groups/GroupInfoFields.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from 'react';
 import { useFormContext } from 'react-hook-form';
 import ColorBoxIcon from '@/components/icons/ColorBoxIcon';
```
patosullivan, line 3: We can probably ditch all of these unused imports.

ui/src/groups/NewGroup/GroupInfoPreview.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import cn from 'classnames';
+import { useFormContext } from 'react-hook-form';
```
patosullivan, line 3: Same here, we can get rid of these unused imports.

patosullivan (approved):
Looks good, one nit: we should probably get rid of the unused imports. I noticed them in two of these files, but there may be others.

urcades (approved):
@rcrdlbl this is looking good, the initial state of the creation flow is beautiful and implemented exactly as intended.

A minor bit of feedback that's polish-oriented (and would be nice to have, but shouldn't block merge) is to potentially "fix"/"pin" the top edge of the modal so that it's not jumping around vertically as one clicks through the options, as shown in the video below:

https://user-images.githubusercontent.com/1195363/199289084-fe028426-e6c2-49a6-9505-bae73f4f7354.mov

rcrdlbl:
@urcades I've updated this – note that this approach inevitably makes it much more difficult to keep the modal exactly centered on the screen

jamesacklin (approved):

PR #1051, opened 2022-10-31
chat: fix state migration

ryjm:
this *should* work, testing on a ~marnec install now.


arthyn (approved):
Took me a bit but I think I get it, lgtm!

Issue #1034, opened 2022-10-28
groups/edit: setting a channel to "default" on creation doesn't work on first try


jamesacklin:
To reproduce:

1. Set up a group
2. Add a channel
3. Click the "Default" checkbox and create the channel
4. Note its non-Default badge status in the list
5. Edit the channel again, note the Default checkbox is unchecked
6. Check the checkbox and save the channel
7. Note the appearance of the "Default" badge in the list of channels

Expectation:

This should work at the first go-around.

PR #1054, opened 2022-11-01
groups: set channel as default upon creation

rcrdlbl:
see: #1034 

had to add in a call to set the channel as default as apparently "CreateChannel" doesn't take a "join" argument


arthyn (approved):
lgtm, yeah we just need to add the join flag to all the channel create types, because it's already in the groups type

Issue #1053, opened 2022-10-31
groups/chat: image block - URL content regressed to bare text


jamesacklin:
**Describe the bug**
Uploading an image through S3 in the chat used to send it as an attachment as message block content, now it doesn't.

**To Reproduce**
1. Run the latest Groups from ~wannec (`jqln9`).
2. Set up S3.
3. Go to a chat.
4. Click the + button and attach an image.
5. Observe that your new URL is pasted into the chat input.
6. Click "Send."
7. Observe the URL echoed back to you as plaintext, not even as a link.

**Expected behavior**
Expected the image to appear as an image block.

**Screenshots**
![image](https://user-images.githubusercontent.com/748181/199124976-b5061ead-15bc-4a4d-9c79-cec412573791.png)

**Additional context:**
Self-hosted moon running latest Groups from `master` (~wannec), Chrome/Fedora.

PR #1057, opened 2022-11-01
chat: fix missing special characters in image url regex

patosullivan:
OTT


jamesacklin:
Does this fix #1053 ?

jamesacklin (approved):

PR #1058, opened 2022-11-01
handle frontend disconnects

### LINKED ISSUE #1039: groups+talk: silent disconnects
Most of our front-end performance problems are rooted in the front-end disconnecting from the back-end -- we have no crashes _per se_ in our JS or Hoon, but the user's session itself periodically disconnects and either reconnects, or doesn't.

We have this problem in Groups 1 and have solved it with a yellow spinner (or red reconnect button) in the titlebar; we have no such affordance in Groups 2.

Two parts to this issue:

1. Detect disconnects and generalize them as a hook so we can disable the UI where appropriate (looking at Groups 1 for inspiration)

2. Design a bit of UI to help users get themselves out of this state. This should be as natural as possible - we've been conditioned to wait for the spinner and to jab it if it turns red, but I'll leave it to @urcades what the solve looks like.

patosullivan:
fixes #1039 

https://user-images.githubusercontent.com/1221094/199313876-c5deb66f-df4e-4a11-b6af-c13e69550c1c.mov

Also, yes, resizing the viewport triggers a call to `http://localhost:3000/~/scry/hark/group/{flag}/latest.json` which results in a 404, which, while it provided a nice way to test this feature, is pretty weird.

### DIFF
--- a/ui/src/api.ts
+++ b/ui/src/api.ts
@@ -44,20 +44,11 @@ async function setupAPI() {
 
 const api = {
   async scry<T>(params: Scry) {
-    try {
-      if (!client) {
-        await setupAPI();
-      }
-
-      const clientScry = await client.scry<T>(params);
-      useLocalState.setState({ subscription: 'connected' });
-      useLocalState.setState({ errorCount: 0 });
-      return clientScry;
-    } catch (e) {
-      useLocalState.setState({ subscription: 'disconnected' });
-      useLocalState.setState({ errorCount: errorCount + 1 });
-      throw e;
+    if (!client) {
+      await setupAPI();
     }
+
+    return client.scry<T>(params);
   },
   async poke<T>(params: PokeInterface<T>) {
     try {
--- a/ui/src/api.ts
+++ b/ui/src/api.ts
@@ -5,6 +5,7 @@ import Urbit, {
   SubscriptionRequestInterface,
   Thread,
 } from '@urbit/http-api';
+import { useLocalState } from './state/local';
 
 export const IS_MOCK =
   import.meta.env.MODE === 'mock' || import.meta.env.MODE === 'staging';
@@ -13,7 +14,9 @@ const URL = (import.meta.env.VITE_MOCK_URL ||
 
 let client = undefined as unknown as Urbit | UrbitMock;
 
-async function setupAPI() {
+const { errorCount } = useLocalState.getState();
+
+export async function setupAPI() {
   if (IS_MOCK) {
     window.ship = 'finned-palmer';
     window.our = `~${window.ship}`;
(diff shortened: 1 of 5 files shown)


patosullivan:
We apparently often see calls to `/~/scry/hark/group//latest.json` that 404 and that'll cause the disconnect banner to show. @liam-fitzgerald @ryjm any idea why that is?

patosullivan:
Nevermind, this was a frontend issue. Pushing a fix.

patosullivan:
We still see 404s on calls to `/~/scry/hark/group/${flag}/latest.json` where we actually have a flag when viewing a group. I'm guessing it's because the group in question doesn't have any notifications. Should this return a 404? Seems like an empty response with a 200 would be better, but I don't know if there's some limitation preventing that. @arthyn @liam-fitzgerald @ryjm

urcades (approved):
Until we have a better sense of how to align the "universal" chrome interface across each of our apps, this banner is a perfect and hyper-obvious solve to surface the disconnect/reconnect interface.

jamesacklin:
> We still see 404s on calls to `/~/scry/hark/group/${flag}/latest.json` where we actually have a flag when viewing a group. I'm guessing it's because the group in question doesn't have any notifications. Should this return a 404? Seems like an empty response with a 200 would be better, but I don't know if there's some limitation preventing that. @arthyn @liam-fitzgerald @ryjm
> 
> Bringing this up because it causes our new subscription state to switch to "disconnected" briefly when loading a group, which causes the "disconnected" status bar to flash for ~300ms.

Well, we don't want that -- any advice for returning a 200 instead? @arthyn @liam-fitzgerald

patosullivan:
Hacky workaround: don't show a disconnect if it's a 404 on that type of path.

arthyn:
well I don't think scries should actually affect the subscription status at all so you can just remove it

ui/src/app.tsx
```diff
@@ -373,6 +378,8 @@ function App() {
   const location = useLocation();
   const isChat = useIsChat();
   const isMobile = useIsMobile();
+  const subscription = useSubscriptionStatus();
+  const errorCount = useErrorCount();
```
arthyn, line 382: doesn't seem like we're using this

ui/src/api.ts
```diff
@@ -41,39 +44,84 @@ async function setupAPI() {
...
-    return client.scry<T>(params);
+      const clientScry = await client.scry<T>(params);
+      useLocalState.setState({ subscription: 'connected' });
+      useLocalState.setState({ errorCount: 0 });
+      return clientScry;
+    } catch (e) {
+      useLocalState.setState({ subscription: 'disconnected' });
+      useLocalState.setState({ errorCount: errorCount + 1 });
+      throw e;
+    }
```
arthyn, line 60: Revert, scries aren't passed through the same channel system so their errors don't have bearing on the connection status

arthyn (changes requested):
This was a good way to do it, just a few changes, should help with the flashing.

ui/src/app.tsx
patosullivan: We're not using errorCount, yeah.

arthyn (approved):
lgtm

PR #1060, opened 2022-11-01
chat: atone

ryjm:
be sober, be vigilant, because your adversary the devil, as a roaring lion, walketh about, seeking whom he may devour


arthyn (approved):
🙏🏼

PR #1063, opened 2022-11-01
groups: click to copy ID

rcrdlbl:
see: #1037.

There's an `Author` component that consists of an avatar and a shipname that we're using in contexts where authorship is connoted that I'm pretty sure covers all cases where a ShipName is used and there isn't additional interactivity, let me know if i'm missing any instances with this.


jamesacklin (approved):
Based, good, thank you

Issue #962, opened 2022-10-02
gallery: text items get title [object, Object]


jamesacklin:
Steps to reproduce:

1. Add a new text curio to a heap.
2. Mouse-over the block.
3. Observe the errorenous title.

Expected result:
The title should be the first ~20 characters of the text content, with an ellipsis indicating truncation.

![image.png](https://images.zenhubusercontent.com/6245cc0024d90711bcf4ea06/ae76e44c-c26b-46c7-923b-94b58c10840c)

PR #1066, opened 2022-11-02
heap: title previews no longer show [Object, object] when formatted text is included in a text heap

rcrdlbl:
see: #962


patosullivan (approved):
Nice!

PR #1078, opened 2022-11-03
Chat: fallback to 'find groups' if join fails on group ref

### LINKED ISSUE #1043: chat: group references that haven't completed finding preview have broken join links
**Describe the bug**
If a subscription takes too long or never comes back, group refs will have a broken join link (possibly go too?)

**To Reproduce**
1. Go to Tlon
2. Click on Watercooler
3. Scroll up to ed's group ref

**Expected behavior**
If the subscription is taking too long we should just link directly to the find groups page with the group flag preloaded or possibly show some kind of status indicator instead of a button.

patosullivan:
OTT

fixes #1043 (I think)

tested by posting a ref to non-existent group.

@arthyn would this cover the case you mentioned in your issue? The ref you mentioned itself is no longer available.

### DIFF
--- a/ui/src/groups/FindGroups.tsx
+++ b/ui/src/groups/FindGroups.tsx
@@ -137,6 +137,11 @@ export default function FindGroups({ title }: ViewProps) {
       return;
     }
 
+    if (ship && name) {
+      navigate(`/find/${ship}/${name}`);
+      return;
+    }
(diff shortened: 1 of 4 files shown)


arthyn (approved):
yeah I think this will solve what I was talking about

Issue #1076, opened 2022-11-03
groups/mobile: disable CSS hovers


jamesacklin:
:hover styling (and interactions) lead to necessary yet annoying double-taps.

This may be rectified with a single-line fix in the Tailwind config: https://github.com/tailwindlabs/tailwindcss/pull/8394

PR #1082, opened 2022-11-04
everywhere: disabled CSS hover on touchscreens

rcrdlbl:
see #1076.

simple as.

tradeoff: some tooltips may no longer be visible on mobile without long-pressing. the only one I can think of that isn't purely instructional is the emoji react hover action that shows who reacted – we likely want to implement an alternative method to see this anyway (to account for cases where more than 5 or so people have reacted to a message with the same emoji, which happens pretty often in group-based chat apps like this)


jamesacklin (approved):
LGTM

PR #1085, opened 2022-11-04
groups: fix for double listing of matched value in join ship selector

### LINKED ISSUE #651: groups/join: search autocomplete lists match at top and still includes match in the list below

patosullivan:
fixes #651

### DIFF
--- a/ui/src/components/ShipSelector.tsx
+++ b/ui/src/components/ShipSelector.tsx
@@ -18,6 +18,7 @@ import {
   ClearIndicatorProps,
   InputActionMeta,
 } from 'react-select';
+import { includes } from 'lodash';
 import CreatableSelect from 'react-select/creatable';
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

Issue #1067, opened 2022-11-02
groups: after clicking into a group invite via notifications, I see no "invite row" in the "find groups" interface


urcades:
**Describe the bug**
I requested (out of band/product) an invite to Tlon's private group in new Groups and successfully received the invite from James in my app notifications. After clicking into the notification, and being brought to the "Find Groups" interface (which shows an invite count of (1) in the sidebar), I see no group invite interface/row:

<img width="1567" alt="Screen Shot 2022-11-02 at 10 06 51 AM" src="https://user-images.githubusercontent.com/1195363/199510538-fbe601e2-df3c-4177-bd4c-8b2b910d5442.png">

**To Reproduce**
1. Ask someone to invite you to a (private?) group in new groups
2. You should receive a new invite in your notifications interface
3. Clicking into the notification should bring you to find groups, where the invite should reside
4. If you see nothing here, you've replicated the bug

**Expected behavior**
After receiving the invite, I expected to be led to an interface where I could accept the group invite

**Additional context:**
- Ship: planet (~fabled-faster)
- Instance type: Tlon-hosted
- MacOS Monterey — 12.6 (21G115)
- Safari — 15.6.1 (17613.3.9.1.16)
- Chromium — Engine Version 106.0.5249.103

urcades:
An additional detail: searching for the group I've been supposedly invited to in "Find Groups" yields no results, i.e. searching for the host of the chat does not display Tlon Corporation

tomholford:
> An additional detail: searching for the group I've been supposedly invited to in "Find Groups" yields no results, i.e. searching for the host of the chat does not display Tlon Corporation

Just ran into this issue too - I think it's when the sig is not included in the search query

patosullivan:
In trying to repro this (haven't been able to) I've encountered a different bug: I can't leave a private group I was invited to. Not sure if it's because the group is hosted on a ~wannec ship and the invite was sent to my planet (on ~binnec).

patosullivan:
> > An additional detail: searching for the group I've been supposedly invited to in "Find Groups" yields no results, i.e. searching for the host of the chat does not display Tlon Corporation
> 
> Just ran into this issue too - I think it's when the sig is not included in the search query

shouldn't be able to search for it without the sig. How'd you get to the search in this instance?

arthyn:
[...]

PR #1087, opened 2022-11-04
groups: adding better handling for errored group joins

arthyn:
This addresses the immediate issues with #1067 by adding an ability remove "errored" joins and also lets new invites show up even if a join was errored. Also adds a new cancel flow for "errored" joins so they can be removed from the sidebar.

This also removes the "Cancel Join" button from the join in progress popover because it's currently not possible to remove a join in progress (only maybe hide it and it would still have the potential to go through). We may be able to come up with something later that gives us the ability to completely cancel, but we need quite a bit to do it.


patosullivan (approved):
LGTM

PR #1088, opened 2022-11-04
groups: fix for collision caused by channels using the same channel flag

### LINKED ISSUE #1068: groups: creating a channel with the same flag in different groups causes collision
**To Reproduce**
1. Create a group and a notebook within it, then add a test post
2. Create another group, then a new notebook with the same name as the previous notebook
3. Observe that the post from the first notebook is duplicated in the new notebook
4. Delete the post in the second notebook
5. Observe that the post is gone from the first notebook

**Expected behavior**
Channels should be able to have the same name across multiple different groups

**Additional context:**
This is because we are only considering the channel flag for some state operations and component rendering. Will need to refactor to use a composite key of group flag + channel flag

patosullivan:
fixes #1068


ui/src/channels/NewChannel/NewChannelForm.tsx
```diff
@@ -39,7 +39,22 @@ export default function NewChannelForm() {
...
+        need to append a timestamp to the end of the name of the new channel.
+
+        Timestamps are used because they are virtually guaranteed to be unique.
+
+        In the future, we will index channels by their full path (including group name), and this will no
+        longer be necessary. That change will require a migration of existing channels.
+       */
+      const tempChannelName = strToSym(values.meta.title);
+      const tempNewChannelFlag = `${window.our}/${tempChannelName}`;
+      const existingChannel = useChatState.getState().pacts[tempNewChannelFlag];
```
arthyn, line 54: We actually need to check the channel type that matches the type of channel trying to be created, not just chat. Also a more reliable place I think would be to check `chats` not `pacts` even though they should both have the same keys.

arthyn (changes requested):

ui/src/channels/NewChannel/NewChannelForm.tsx
patosullivan: Good point, just pushed a commit addressing this (and using chats instead of pacts).

liam-fitzgerald (changes requested):
just appending a simpler number i.e `~zod/test-300` make things a little bit more user friendly when we expose the flag to the user

patosullivan:
> just appending a simpler number i.e `~zod/test-300` make things a little bit more user friendly when we expose the flag to the user

k, switched to a random small number.

liam-fitzgerald (approved):

PR #1089, opened 2022-11-04
chat: fixing thread layout regression

arthyn:
OTT, I don't think this had a ticket


patosullivan (approved):
LGTM in both talk and groups.

Issue #1092, opened 2022-11-06
chat: chatinput aggressively grabs input on mobile


jamesacklin:
**Describe the bug, optionally with screenshots**
We appear to trap the cursor in the Chat input, especially on mobile.

**To Reproduce**
1. Visit a Chat channel or DM on mobile (responsive)
2. Enter some text and send it
3. Attempt to long-press + drag to select some text from the chat log
4. Observe that the chatinput traps focus, disabling virtually any other interaction

**Expected behavior**
The chat input should relinquish focus like any other form field.

**Additional context:**
- Self-hosted running 0.5.0, iOS/Safari
- Original reporter: @lukestiles

Issue #1083, opened 2022-11-04
talk: drafts causing infinite loop


jamesacklin:
I'm encountering a condition where I have a blank draft (single space) in a multi-DM in Talk. I unfocus the window, refocus the window, and get stuck in a loop (illustrated in the screenshot).

Steps to reproduce:
1. Open a multi-DM in Talk
2. Type a message, then select all -> delete
3. Unfocus the window without unfocusing the text field
4. Refocus the window
5. Observe loop:

![2022-11-04T09:19:04,412982151-04:00.png](https://images.zenhubusercontent.com/6245cc0024d90711bcf4ea06/8a01ae56-28f0-4ca6-921b-a556287a1429)

arthyn:
It's either drafts or competing focus triggering drafts, but yeah I saw the same

PR #1097, opened 2022-11-07
talk: fixing looping chatinput focus

arthyn:
This addresses #1083. Bug was introduced when I changed the useMedia hook to prevent multiple listeners, initialized was getting overwritten from true to false. I think this also addresses #1092?


jamesacklin:
Nice

PR #1100, opened 2022-11-07
notebooks: prevented insertion of extraneous newlines into notebook posts

rcrdlbl:
see: #981.

tests currently won't pass due to an ESlint rule, if someone can help me rewrite that check i'd appreciate it.

### DIFF
--- a/ui/src/logic/tiptap.ts
+++ b/ui/src/logic/tiptap.ts
@@ -279,7 +279,7 @@ export function JSONToInlines(
           return [{ break: null }];
         }
         const isContentFinal = idx === json.content.length - 1;
-        const isBreakDirectlyBefore = memo[idx - 1] && typeof memo[idx - 1] !== 'string' && memo[idx-1].hasOwnProperty('break');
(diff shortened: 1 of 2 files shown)


arthyn (approved):
lgtm

ui/src/logic/tiptap.ts
```diff
@@ -279,7 +279,8 @@ export function JSONToInlines(
           return [{ break: null }];
         }
         const isContentFinal = idx === json.content.length - 1;
-        if (isContentFinal) {
+        const isBreakDirectlyBefore = memo[idx - 1] && typeof memo[idx - 1] !== 'string' && memo[idx-1].hasOwnProperty('break');
```
arthyn, line 282:
```suggestion
        const isBreakDirectlyBefore = memo[idx - 1] && typeof memo[idx - 1] !== 'string' && 'break' in memo[idx-1];
```

arthyn: this should fix CI I think

PR #1101, opened 2022-11-08
Groups: show private group join requested in group reference

### LINKED ISSUE #1081: Groups/References: Private group with join requested has no visual state in references
Clicking "request to join" on a chat reference to a private group results in the request sending, and appearing in the group list sidebar, but the chat reference seems to not have a "Requested" state, so it feels like nothing happened until you leave the group and return to the list of your groups (in talk, you can't tell at all!)

**To Reproduce**
1. Receive a reference to a private group your ship is not in
2. Click on "Request To Join"
3. Receive no visual feedback
4. Return to the "All Groups" sidebar or open the group info modal from the reference, see the join was requested
5. Return to the chat, observe no indication you've requested to join the group

**Expected behavior**
The "request to join" button becomes disabled, and the text is replaced with "Requested" (perhaps a "cancel" option appears next to it)

**Additional context:**
- planet
- self-hosted, digitalocean
- MacOS Chrome

patosullivan:
Fixes #1081

### DIFF
--- a/ui/src/components/References/GroupReference.tsx
+++ b/ui/src/components/References/GroupReference.tsx
@@ -12,10 +12,8 @@ interface GroupReferenceProps {
 
 export default function GroupReference({ flag }: GroupReferenceProps) {
   const gang = useGang(flag);
-  const { group, privacy, open, join, reject, requested } = useGroupJoin(
-    flag,
-    gang
-  );
+  const { group, privacy, open, join, reject, requested, button } =
+    useGroupJoin(flag, gang);
   const { ship } = getFlagParts(flag);
 
   const meta = group?.meta || gang?.preview?.meta;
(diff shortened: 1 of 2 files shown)


ui/src/components/References/GroupReference.tsx
```diff
@@ -23,6 +26,19 @@ export default function GroupReference({ flag }: GroupReferenceProps) {
...
+    }
+    if (requested) {
+      return 'Join Requested';
+    }
+    if (privacy === 'private') {
+      return 'Request to Join';
+    }
+    return 'Join';
+  };
+
```
arthyn, line 41: we already have this logic if you just use `button.text` from `useGroupJoin`, maybe we should it's language out for these though

patosullivan: Thanks, missed that in useGroupJoin.

arthyn (approved):
lgtm

Issue #1102, opened 2022-11-08
talk: small updates to message element spacing


urcades:
In the current implementation of our "message row component", found in Talk, talk channels, and comments, there are **two** areas where additional vertical spacing could be added:

The first of these being the space between 1) the last message sent by Author A, and 2) the first message/author row demarcation row by Author B, demonstrated in the video below:

https://user-images.githubusercontent.com/1195363/200629293-d7b94e4b-1106-495a-9aa1-ff9cecda7e42.mov

The second of these being after the final message in a Talk channel or DM:

https://user-images.githubusercontent.com/1195363/200629994-4f55ff9b-eb44-4fe4-ada2-b47452d2d571.mov

~

### **Fix: Spacing between distinct authors' messages**

As specified in our previous designs for Talk message rows, we've intended for there to be a noticeable gap/margin/space between messages sent from distinct authors:

![Screenshot 2022-11-08 at 12 30 16 PM](https://user-images.githubusercontent.com/1195363/200634673-f1f38a03-ceff-48d6-983c-994286afddf7.png)

Note here, between Romina's last message and Ed's first message there is a space of approximately 12px (or whatever the tailwind-measure equivalent is).

~

### **Fix: Final message bottom-padding:**

Between the final message row in a Talk channel, and the bottom edge of the scroll interface, we should provide an equivalent amount of bottom margin/spacing that is afforded the left/right margins of the row elements:

![Screenshot 2022-11-08 at 12 22 17 PM](https://user-images.githubusercontent.com/1195363/200633981-bf215cdf-a6b2-48ac-8d24-9ee345b1e842.png)

Note in the following image how the left edge of a given message row does not encroach into the padding that rings the chat scroll, while the last message dips into the established **padding.**

![Screenshot 2022-11-08 at 12 23 32 PM](https://user-images.githubusercontent.com/1195363/200633434-332cf3d4-d9e7-4d0c-887c-9fb391ffc964.png)

urcades:
For a precise look into the intended spacing, check out this frame in the Talk handoff files:

https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/Product-Handoffs-(Locked)?node-id=1292%3A69151

urcades:
As always, I'm deferring pixel-precise measures to whatever the Tailwind implementation's spacing is set to

PR #1104, opened 2022-11-08
Always open chat at bottom

### LINKED ISSUE #1096: chat: navigating between channels jumps pin chat log around
A breakout of #965:

Clicking away from a chat channel and back into it also positions my scrollback way above where I was (presumably when I opened the app, before others chatted more and increased the log length).

May be related to #1095.

patosullivan:
Fixes #1096 and #1102 

Note that this doesn't account for scrolling directly to a message, as it appears that for now we don't actually use the `scrollTo` prop. Should it?

### DIFF
--- a/ui/src/chat/ChatScroller/ChatScroller.tsx
+++ b/ui/src/chat/ChatScroller/ChatScroller.tsx
@@ -94,6 +94,7 @@ function createRenderer({
           newDay={newDay}
           ref={ref}
           unread={unreadBrief}
+          isLast={keyIdx === keys.length - 1}
         />
       );
     }
--- a/ui/src/chat/ChatScroller/ChatScroller.tsx
+++ b/ui/src/chat/ChatScroller/ChatScroller.tsx
@@ -5,12 +5,13 @@ import React, {
   useCallback,
   useEffect,
   useMemo,
+  useRef,
   useState,
 } from 'react';
 import { isSameDay } from 'date-fns';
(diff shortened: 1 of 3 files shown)


ui/src/chat/ChatScroller/ChatScroller.tsx
```diff
@@ -184,6 +187,19 @@ export default function ChatScroller({
...
+    // This is a hack to force it to scroll to the bottom.
+
+    if (indexData.firstItemIndex === FIRST_INDEX && diff === 0) {
+      // We need to wait to make sure the virtuoso component has been updated.
+      setTimeout(() => {
+        virtuoso?.current?.scrollToIndex({
+          index: indexData.data.length - 1,
+        });
+      }, 50);
+    }
```
arthyn, line 202: did moving this into the `oldWhom !== whom` statement not work? I guess the index data wouldn't be accurate...

patosullivan: Moving what into the oldWhom !== whom statement? Like moving this inside that if statement:

```
    const diff = mess.size - indexData.data.length;
    if (diff !== 0) {
      setIndexData({
        firstItemIndex: indexData.firstItemIndex - diff,
        data: keys,
      });
    }
```

No, doesn't work.

patosullivan:
Going to spike on a different way to do this that doesn't require the ref/setTimeout.

tomholford (approved):
LGTM

Issue #1090, opened 2022-11-05
groups: admin actions might not be propagating to host


arthyn:
When trying to invite my moon to the Tlon group as an admin, it seemingly worked from my ship ~nocsyx-lassul, and the invite went through to my moon. However, this causes the group join to fail because that ship never ends up in the pending cordon on the host. So we're either not propagating all actions to host from admins or specifically missing cordon edits, needs investigation.

arthyn:
This was probably the cause of lots of invite issues seen in #1067

PR #1106, opened 2022-11-09
groups: proxy pokes to host

arthyn:
This addresses #1090, I'd like two approvals because if the logic is wrong here you get looping.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -452,6 +452,13 @@
   ::
   ++  go-pass
     |%
+    ++  update
+      |=  =action:g
+      ^-  card
+      =/  =wire  (snoc go-area %proxy)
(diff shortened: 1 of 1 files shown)


desk/app/groups.hoon
```diff
@@ -678,6 +685,9 @@
   ++  go-update
     |=  [=time =diff:g]
     ^+  go-core
+    ?:  &(!=(our.bowl p.flag) from-self) 
+      =.  cor  (emit (update:go-pass [flag [time diff]]))
+      go-core
```
liam-fitzgerald, line 690: this logic need to go in the `%group-action` case of the +poke arm, see the pattern established in `%chat`

liam-fitzgerald (dismissed):

ryjm (approved):
lgtm

Issue #566, opened 2022-08-08
groups: implement optimistic loads and indeterminate spinners


jamesacklin:

jamesacklin:
Split into other issues.

PR #1108, opened 2022-11-10
groups: adding new harness for tracking poke results

arthyn:
This adds a pathway for tracking poke responses that come down subscriptions so that we can start culling the missing loading states in #566 (and not just track the poke's request itself which is only half of the loading). 

This adds a loading state to group creation as an example. Essentially, when doing a poke that will eventually produce a fact down a subscription, you call and await `useSubscriptionState.getState.track(...)`. The function you pass must be able to read the facts on a subscription and determine the correct fact has come down. When this promise is finished you can assume that the function you passed returned true and it's safe to show success/proceed.

### DIFF
--- a/ui/src/groups/NewGroup/NewGroupInvite.tsx
+++ b/ui/src/groups/NewGroup/NewGroupInvite.tsx
@@ -9,6 +9,9 @@ import CaretDownIcon from '@/components/icons/CaretDownIcon';
 import IconButton from '@/components/IconButton';
 import XIcon from '@/components/icons/XIcon';
 import { PrivacyType } from '@/types/groups';
+import { Status } from '@/logic/status';
+import LoadingSpinner from '@/components/LoadingSpinner/LoadingSpinner';
+import X16Icon from '@/components/icons/X16Icon';
 
 interface NewGroupInviteProps {
   groupName: string;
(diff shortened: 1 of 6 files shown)


ui/src/groups/NewGroup/NewGroupInvite.tsx
```diff
@@ -187,10 +195,20 @@ export default function NewGroupInvite({
...
+            submitText
+          ) : status === 'loading' ? (
+            <>
+              <LoadingSpinner className="mr-2 h-4 w-4" />
+              <span>Creating</span>
+            </>
+          ) : status === 'error' ? (
+            <>
+              <X16Icon className="mr-2 h-4 w-4" />
+              <span>Errored</span>
```
tomholford, line 209: Is it possible to show a more descriptive error to the user?

tomholford (approved):
This is great, nice work. One question inline, but not a merge-blocker. LGTM

ui/src/groups/NewGroup/NewGroupInvite.tsx
arthyn: The only thing we get back here is a stack trace so not really. Erroring here means the poke itself failed. I do wish we had more descriptive errors. I think this is an architectural/philosophical discussion that we need to have. Currently we just do lots of assertions, but it'd be nice to change those to not crash and actually send down useful errors.

PR #1111, opened 2022-11-10
groups: mobile nav redesign

jamesacklin:
A wholesale redesign of the mobile navigation. Unless otherwise noted, these changes only affect the mobile view and should have no effect on the desktop view.

**Groups app home**
- Add labels to all icons
- Replaces Groups icon
- Switches notification counter to activity bell
- Places Find Groups in the nav
- Adds grayscale/opacity effects to the ship avatar in the inactive Profile tab
- Adds an overflow menu containing the Submit Feedback link
- Adds a Add Group button in the header of the screen
- Relocates the group sorting control to the header

**Find Groups, Notifications**
- Unifies page headers with Groups app home
- Removes gray background and card section treatments

**Group home**
- Adds labels to all icons
- Relocates the Group Info item
- Adds grayscale/opacity effects to the group avatar in the inactive Group Info tab
- Drops the Activity tab
- Adds an Invite tab, which pops the Invite modal
- Adds a back button (goes back to Groups app home) to the header with the group avatar + name
- Relocates the channel sorting dropdown to the header
- For administrators, adds a New Channel button in the header

**New Channel modal**
- Redirects the user back to the group home on success, rather than the channel editor

**Invite modal**
- Switches to shorter placeholder text on mobile to prevent ugly wrapping

**Profile edit and Profile modal (including desktop)**
- Restores details to sigils

https://user-images.githubusercontent.com/748181/200985314-0ee220ca-f789-4e5b-94af-5347a910fba4.mov

**Does not address**
Horizontal overflow issues with group creation or editing; our new group metadata editor is a bit too wide. Logging in a separate issue.

### DIFF
--- a/ui/src/components/Sidebar/MobileSidebar.tsx
+++ b/ui/src/components/Sidebar/MobileSidebar.tsx
@@ -21,15 +21,15 @@ export default function MobileSidebar() {
       <footer className="flex-none border-t-2 border-gray-50">
         <nav>
           <ul className="flex justify-items-stretch">
-            <NavTab to="/" aria-label="Groups">
+            <NavTab to="/">
               <AppGroupsIcon className="mb-0.5 h-6 w-6" />
               Groups
             </NavTab>
-            <NavTab to="/notifications" aria-label="Activity">
+            <NavTab to="/notifications">
               <BellIcon className="mb-0.5 h-6 w-6" />
               Activity
             </NavTab>
-            <NavTab to="/find" aria-label="Find">
+            <NavTab to="/find">
               <MagnifyingGlassIcon className="mb-0.5 h-6 w-6" />
               Find Groups
             </NavTab>
@@ -44,7 +44,7 @@ export default function MobileSidebar() {
               />
               Profile
             </NavTab>
-            <NavTab to="/actions" aria-label="Options">
+            <NavTab to="/actions">
               <ElipsisIcon className="mb-0.5 h-6 w-6" />
               Options
             </NavTab>
--- a/ui/src/components/Sidebar/MobileSidebar.tsx
+++ b/ui/src/components/Sidebar/MobileSidebar.tsx
@@ -1,19 +1,19 @@
-import cn from 'classnames';
 import React from 'react';
-import { Outlet, useLocation, useMatch } from 'react-router';
+import cn from 'classnames';
+import { Outlet, useMatch } from 'react-router';
 import { useNotifications } from '@/notifications/useNotifications';
 import NavTab from '../NavTab';
-import GroupIcon from '../icons/GroupIcon';
-import ActivityIndicator from './ActivityIndicator';
-import AsteriskIcon from '../icons/Asterisk16Icon';
+import AppGroupsIcon from '../icons/AppGroupsIcon';
+import ElipsisIcon from '../icons/EllipsisIcon';
+import BellIcon from '../icons/BellIcon';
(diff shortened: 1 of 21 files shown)


jamesacklin:
Alerting @urcades and @lukestiles for visibility.

ui/src/components/Sidebar/MobileSidebar.tsx
```diff
@@ -22,34 +22,31 @@ export default function MobileSidebar() {
...
+                className={cn(
+                  'mb-1 h-6 w-6',
+                  !profileMatch && 'opacity-50 grayscale'
+                )}
               />
+              Profile
+            </NavTab>
+            <NavTab to="/actions" aria-label="Options">
+              <ElipsisIcon className="mb-0.5 h-6 w-6" />
+              Options
```
arthyn, line 49: aria labels can be removed now that these have actual text

ui/src/profiles/EditProfile/EditProfile.tsx
```diff
@@ -253,6 +257,8 @@ function EditProfileContent() {
 }
 
 export default function EditProfile({ title }: ViewProps) {
+  const isMobile = useIsMobile();
```
arthyn, line 260: seems unused

ui/src/groups/GroupSidebar/MobileGroupSidebar.tsx
```diff
@@ -1,20 +1,18 @@
 import React from 'react';
```
arthyn, line 1: A little confused about the changes for this nav. we removed the ability to see group filtered notifications and group actions, and still don't have a way to leave the group. 

Seems like notifications should be top level and invite should be in the ellipsis menu with the other things I mentioned?

ui/src/groups/MobileGroupsActions.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import SidebarItem from '@/components/Sidebar/SidebarItem';
+import AsteriskIcon from '@/components/icons/Asterisk16Icon';
+
+export default function MobileGroupsActions() {
+  return (
+    <nav className="h-full flex-1 overflow-y-auto p-2">
+      <ul className="space-y-3">
```
arthyn, line 8: Feels weird to have this just for one item

ui/src/nav/MobileGroupRoot.tsx
```diff
@@ -1,27 +1,46 @@
...
-          <span>All</span>
-          <CaretRight16Icon className="ml-1 h-4 w-4 text-gray-400" />
-        </Link>
+      <header className="flex items-center justify-between p-4 pt-3 pr-5">
+        <div>
+          <Link
+            to="/"
+            className="default-focus inline-flex items-center text-base font-semibold text-gray-800 hover:bg-gray-50"
+          >
+            <CaretLeft16Icon className="mr-2 h-6 w-6 text-gray-400" />
```
arthyn, line 23: use `CaretLeftIcon` if you're going to size it to 24px, convention so far has been only ever size down otherwise it'll look too thick as it does like this.

arthyn (changes requested):
A few questions and a few changes. Overall I think this is getting better, but I think we're still missing things and dropped some functionality unnecessarily. I really like the convention of creation living in top right.

arthyn (approved):
lgtm, we should add those missing options in another pr

PR #1114, opened 2022-11-10
channels: fix recent sort in sidebar

### LINKED ISSUE #1073: groups: channels not sorting by recency
At the moment, channels are sorted statically:

...where posts in the Watercooler (among others) occurred more recently than any new posts in the Tlon Notebook.

This ordering appears to correspond with edits in the Channel Editor.

tomholford:
# Context

This resolves #1073 by refactoring the channel list sort behavior and adding an effect hook to prefetch messages. When selecting a group, the messages for each channel will be fetched, then the last message timestamp will be compared to determine the sort order.

For now this is done somewhat ineffciently. In discussion with @arthyn, we came up with a more scalable approach that can be implemented after the initial release. In short, implement a new backend endpoint that publishes the latest message for each channel.

# Preview

https://user-images.githubusercontent.com/16504501/201065746-db26cb48-d881-495b-9d9e-e955f96bc767.mp4

### DIFF
--- a/ui/src/chat/ChatChannel.tsx
+++ b/ui/src/chat/ChatChannel.tsx
@@ -7,7 +7,12 @@ import ChatInput from '@/chat/ChatInput/ChatInput';
 import ChatWindow from '@/chat/ChatWindow';
 import Layout from '@/components/Layout/Layout';
 import { ViewProps } from '@/types/groups';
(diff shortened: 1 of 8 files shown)


arthyn (approved):
lgtm, we should follow up PR the other apps to be supported

PR #1119, opened 2022-11-10
groups: fix for jumping/mis-ordered admin channel list

### LINKED ISSUE #1035: groups/edit: various channel editing mishaps
During testing, @zalberico ran into several issues with the front-end of the channel editor getting out of sync with ship state. Our dream of direct manipulation is nice, but only works if the execution is flawless.

To combat this without a wholesale rewrite of editing the zone/channel model, I propose (with @urcades' OK) we add a "Save" button to the channel-editing form thusly:

- This Save button should appear after any changes to the group channel structure such as rearranging zones, adding/removing/renaming channels, etc.
- The front-end should update optimistically and immediately.
- The save button should effectively submit these changes to the back-end, and disappear when the action is complete.
- Navigating away from this page in an unsaved state should present a window.confirm warning ("You have made unsaved changes to your group. Navigate away?" or something of the sort).

patosullivan:
Fixes #1035 by reloading group state (await useGroupState.getState().start()) when moving a channel. Incorrect list will flash briefly (a few milliseconds) before correct one is shown.

### DIFF
--- a/ui/src/state/groups/groups.ts
+++ b/ui/src/state/groups/groups.ts
@@ -420,7 +420,7 @@ export const useGroupState = create<GroupState>(
           },
         };
         await api.poke(groupAction(flag, diff));
-        await useGroupState.getState().start();
+        await useGroupState.getState().updateGroups();
       },
       setChannelPerm: async (flag, nest, sects) => {
         const currentReaders = get().groups[flag].channels[nest]?.readers || [];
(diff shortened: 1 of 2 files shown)


patosullivan:
> 

good idea, just added the method.

arthyn (approved):
lgtm, I wonder how we could test this with a "slow" ship

PR #1134, opened 2022-11-10
chat: use scrollTo to scroll to messages, fix thread exits

### LINKED ISSUE #1105: chat: we need the abillity to scroll to a particular message in the chat scroller

patosullivan:
Fixes #1105 

We can now scroll to a message using the URL param `msg` (such as when clicking a chat reference).

Also, when closing a thread we'll return to the message attached to that thread in the scroller (important for mobile).


arthyn (approved):
very nice, lgtm!

PR #1136, opened 2022-11-11
groups: avoid alien ships for now

ryjm:
until runtime fix is in, to avoid ships getting into a bad state.


desk/app/groups.hoon
```diff
@@ -78,6 +78,9 @@
 ++  emit  |=(=card cor(cards [card cards]))
 ++  emil  |=(caz=(list card) cor(cards (welp (flop caz) cards)))
 ++  give  |=(=gift:agent:gall (emit %give gift))
+++  scry
+  |*  [=mold care=term =desk =path]
```
liam-fitzgerald, line 82: This is a type-hole, you shouldn't polymorphism like this

liam-fitzgerald (changes requested):
Pls fix scry helper, else LGTM!

liam-fitzgerald (approved):

arthyn (approved):

PR #1138, opened 2022-11-11
channels: recent sort applies to all channel types

tomholford:
# Context

The previous PR (#1114) only supported chat channels; this adds support diary and heap channels too.

# Preview

https://user-images.githubusercontent.com/16504501/201310743-70d962c5-6be4-4727-abdf-218f25d55a35.mp4


jamesacklin (approved):
Makes sense to me

arthyn (approved):
lgtm ty!

PR #1145, opened 2022-11-14
chat: fix backscroll loading

### LINKED ISSUE #1143: chat: limited backscroll, disappearing old messages
This is observable in channels with long backscrolls. First noticed in the urbit app dev channel.

1. Scroll all the way back and observe the oldest message.
2. Post a new message in the channel.
3. Scroll back and observe that the oldest message is gone.

arthyn:
This fixes #1143 and throws in some extra word breaking for inline code and chat input.

Side note, we don't handle block code right now. We should add handling for that with the potential for overflow scrolling.

Btw when doing this I couldn't find our old message row/type spec in figma @urcades so that I could match what inline code was supposed to be.

### DIFF
--- a/ui/src/chat/ChatScroller/ChatScroller.tsx
+++ b/ui/src/chat/ChatScroller/ChatScroller.tsx
@@ -237,10 +237,10 @@ export default function ChatScroller({
 
   const fetchMessages = useCallback(
     async (newer: boolean) => {
-      const newest = mess.peekLargest();
-      const seenNewest = newer && newest && loaded.newest.leq(newest[0]);
-      const oldest = mess.peekSmallest();
-      const seenOldest = !newer && oldest && loaded.oldest.geq(oldest[0]);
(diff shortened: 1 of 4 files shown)


ui/src/chat/ChatScroller/ChatScroller.tsx
```diff
@@ -276,10 +276,18 @@ export default function ChatScroller({
         style={{
           overflowY: 'scroll',
         }}
-        atBottomThreshold={250}
-        atTopThreshold={250}
-        atTopStateChange={(top) => top && fetchMessages(false)}
-        atBottomStateChange={(bot) => bot && fetchMessages(true)}
+        atBottomThreshold={2500}
```
arthyn, line 279:
```suggestion
        atBottomThreshold={250}
```

tomholford (approved):
LGTM

PR #1146, opened 2022-11-14
chat: new unread tracking system

arthyn:
Addresses #1042 and #1122. This fixes the brief marking loop by making sure we use the correct server time, and changes a lot of the way we do unreads on the frontend, mostly to help manage UI interactions. **Please look through with a fine-tooth comb and give a spin locally if you can.**

First off, this further utilizes the chat store by adding an unread field, basically separating unread state from being directly tied to the backend. This lets us manage the state of showing unread markers and things much easier. We also track whether or not we're currently at the bottom of scroll in this particular chat.

1. When we get a new brief with unreads, we either mark it read immediately because we're at the bottom of the window or if we aren't at the bottom, we update a new separately tracked field in chat store (not state, trying to keep these separate).
2. Once we have an "unread" in chat store, we track whether it's been "seen", if it has a timer going to be marked read, and a brief that "lags" behind the current state.
3. If we have an "unread" in chat store then we show the banner at the top of the window and the marker in the chat where the last message was read. Once someone sees the marker in chat, the banner will hide itself, but the marker will remain for 15 seconds or until the user scrolls away. 
4. When either of those happen, the "unread" will be removed and we send the poke to mark read.


tomholford:
>this further utilizes the chat store by adding an unread field, basically separating unread state from being directly tied to the backend. This lets us manage the state of showing unread markers and things much easier. 

Huge!

liam-fitzgerald (changes requested):
Discussed OOB, proposed change won't work on kick

jamesacklin:
Blows away my channel subscriptions when proxying against my livenet ship; I assume this is expected?

arthyn:
Btw this does not solve the issue with having to look through threads for unread indicators, we should do that in a separate PR so this can get through.

liam-fitzgerald (approved):
hoon side LGTM!

ui/src/chat/ChatMessage/ChatMessage.tsx
```diff
@@ -57,19 +58,45 @@ const ChatMessage = React.memo<
...
     ) => {
       const { seal, memo } = writ;
-      const { markRead } = useChatState.getState();
-      const [viewRef, inView, entry] = useInView({
+      const chatInfo = useChatInfo(whom);
+      const unread = chatInfo?.unread;
+      const { ref: viewRef } = useInView({
         threshold: 1,
-        triggerOnce: true,
+        onChange: useCallback(
```
tomholford, line 65: Optional: Since this is a long function with stacked conditionals, would improve readability to have a comment to explain the decision tree.  Not a merge-blocker.

tomholford (approved):
Pull and played with it locally, buttery smooth :butter:  Nice work @arthyn !

One optional nit inline, otherwise LGTM

jamesacklin:
Experimented a bit locally, meets with my approval

PR #1149, opened 2022-11-14
Fix for ship selector with exact patps

### LINKED ISSUE #1117: ship-selector: doesn't prioritize exact match over suggestions
If you're trying to type in an exact name, but you have lots of other names in your contact store, the exact name will get pushed to the bottom.

**To Reproduce**
1. Go to new message in Talk
2. Type an exact patp that isn't in your contact store already
3. See that it's not the first result
4. Scroll down to see it at the very bottom.

**Expected behavior**
An exact match patp should be the first result

patosullivan:
I think this gets us where we want to be. See video, but please test for yourselves.

Fixes #1117 
Closes #1141 

https://user-images.githubusercontent.com/1221094/201757947-a5202f28-9373-4e2e-9b60-47cbe53b69d5.mov

### DIFF
--- a/ui/src/components/ShipSelector.tsx
+++ b/ui/src/components/ShipSelector.tsx
@@ -476,7 +476,7 @@ export default function ShipSelector({
         isValidNewOption={(val) =>
           includes(
             slicedOptions.map((o) => o.value),
-            val
+            preSig(val)
           )
             ? false
             : isValidNewOption(val)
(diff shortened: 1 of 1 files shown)


patosullivan:
Sorry, uploaded the wrong screen recording first.

rcrdlbl:
Oh, *THAT'S* where you were supposed to exclude stuff!

rcrdlbl (approved):
lgtm!

PR #1151, opened 2022-11-15
chat: scroller smorgasbord

### LINKED ISSUE #1116: Clicking on another thread while one is open is broken
See title. Example:

https://user-images.githubusercontent.com/2065299/201125685-d159e9be-b283-4990-8d59-b1ba59387d4b.mov

**To Reproduce**
1. Click a thread.
2. Click another while it's still open.
3. ...
4. Profit.

**Expected behavior**
All threads rendering correctly

tomholford:
# Context

This fixes #1116, and extends the existing scrollTo functionality.

# Changes

- refactor vscroller pre-pending and scrollTo
- clicking Unread banner scrolls to first unread message
- if brief unread count is higher than default page fetch size (100), fetch older messages so the unread deep link banner is functional

# TODO

- [x] @urcades: How should the linked message be styled? Currently using the same background color as hovering the message, but we should probably have a distinct color like G1 does for this flow. Here's what it looks like now:

image

# Preview

## Opening Multiple Threads

https://user-images.githubusercontent.com/16504501/201763958-6458dd17-ede5-4445-8749-b427d8ba6cc6.mp4

## Deep Link to Chat

https://user-images.githubusercontent.com/16504501/201911500-7c8447b9-11e1-4a4f-a977-2fad13ce7e67.mp4

## Click Unread Banner to go to First Unread

https://user-images.githubusercontent.com/16504501/201938816-c2f50337-4399-4053-a69b-6efa76022c1f.mp4


jamesacklin:
Everything appears to work smoothly (tested on livenet). Nice job!

patosullivan (approved):
This is awesome. MUCH simpler.

ui/src/chat/ChatScroller/ChatScroller.tsx
```diff
@@ -151,68 +151,6 @@ export default function ChatScroller({
...
-    [keys, messages]
-  );
-
-  const [indexData, setIndexData] = useState<{
-    firstItemIndex: number;
-    data: bigInt.BigInteger[];
-  }>({
-    firstItemIndex: FIRST_INDEX,
-    data: keys,
-  });
```
arthyn, line 169: unfortunately we can't get rid of this. if you scroll back on any channel where it has to load the old posts, you'll see it start to load and then completely lose your position because of the way virtuoso does indexes.

This was their official suggestion: https://virtuoso.dev/prepend-items/

arthyn (changes requested):
this is awesome stuff. we have to keep the index stuff though, because we need to prepend data to the list and this changes the indexes.

arthyn:
This also doesn't seem to scroll to the bottom when switching between chats.

urcades:
@tomholford I think a quick hack for the visual styling Q you had re: visiting a deep link or viewing a first unread messages or getting linked to a thread should be a classic blue background (essentially what we do in Groups1) — we can use the existing specification + element coverage of the onHover state, but just ensure that it's blue to emphasize its connection to "unreads" or being "linked".

tomholford:
@urcades Thanks for the feedback, will use Groups 1 blue for now, easy to revisit this in the future.

Preview:

image

image

tomholford:
> This also doesn't seem to scroll to the bottom when switching between chats if you've scrolled back far enough. Try between watercooler and bulletin board, you have to at least scroll enough that it loads some older messages.

Unfortunately I am not able to connect to a mainnet ship using SHIP_URL, having SSL issue. Would you be able to test the latest changes and report back?

arthyn:
still seeing an issue where if you are scrolled back enough the scroller gets stuck at the top when navigating between chats.

ui/src/chat/ChatScroller/ChatScroller.tsx
```diff
@@ -239,7 +276,7 @@ export default function ChatScroller({
           overflowY: 'scroll',
         }}
         atBottomThreshold={250}
-        atTopThreshold={2500}
```
arthyn, line 280: I had set this really high to try and have the top get loaded really early so your scrolling is never interrupted by loading, but maybe you found an issue?

tomholford: Ah yeah, this was leftover from trying to get the darn thing to work. It has been restored in the latest commit

tomholford:
@arthyn:
> still seeing an issue where if you are scrolled back enough the scroller gets stuck at the top when navigating between chats.

Just pushed a commit to address this. Here is an excerpt from the commit message:

> This change forces a scroller remount for each channel switch.
> This resets the scroll position when switching channels;
> previously, when switching between channels, the virtuoso
> internal scroll index would remain the same. So, if one scrolled
> far back in a long channel, then switched to a less active one,
> the channel would be scrolled to the top.

It feels a little dirty, and I wonder about performance on lower end devices. Works well on the dev machine, but testing it now on mobile and there's definitely some jank. So, looking into that now

tomholford:
@arthyn Posted about this in the App Dev channel, but I think the mobile jank is present on master. So to keep this PR's scope from creeping any more, going to re-request review as is, and open a separate issue to track.

ui/src/chat/ChatScroller/ChatScroller.tsx
tomholford: @arthyn FYI added some mobile specific values to account for smaller page size; also tweaked the overscan. LMK what you think!

ui/src/chat/ChatUnreadAlerts.tsx
```diff
@@ -15,6 +18,27 @@ export default function ChatUnreadAlerts({ whom }: ChatUnreadAlertsProps) {
...
+  const chatKeys = useChatKeys({ replying: false });
+  const goToFirstUnread = useCallback(
+    (b: ChatBrief) => {
+      const { 'read-id': lastRead } = b;
+      if (!lastRead) {
+        return;
+      }
+      // lastRead is formatted like: ~zod/123.456.789...
+      const lastReadBN = bigInt(lastRead.split('/')[1].replaceAll('.', ''));
+      const firstUnread = chatKeys.find((key) => key.gt(lastReadBN));
```
arthyn, line 33: The reason you can't do eq here is that the IDs don't actually correlate to the time keys, but this is fine for now we have some other issues to fix before you can associate them correctly I think.

arthyn (approved):
I think this is much cleaner, remount ftw lol. We can address other jankiness in other issues, this solves what you originally wanted to do.

Issue #1024, opened 2022-10-21
`(sane %tas)` check on `q.flag` needed to prevent bad behavior


rabsef-bicrym:
Poking `groups` etc. with in`sane` `@tas` values for a `q.flag` results in an extant `flag` but one which cannot be used

With a group named ~pedia and a diary named ~pedia I broke front end actions entirely, despite being able to see the group and the notebook in it

trying to add a new chat to it:
```
index.5ecbd790.js:90 gall: poke-as cast fail :groups [a=%json b=%group-action]
syntax error
{1 6}
[%key 'flag']
/sys/vane/clay/hoon:<[832 51].[832 63]>
/sys/vane/gall/hoon:<[890 27].[890 45]>
```

trying to post to ~pedia diary:
```
index.5ecbd790.js:90 gall: poke-as cast fail :diary [a=%json b=%diary-action]
syntax error
{1 6}
[%key 'flag']
/sys/vane/clay/hoon:<[832 51].[832 63]>
/sys/vane/gall/hoon:<[890 27].[890 45]>
```

rabsef-bicrym:
They were also in a section called ~pedia - this seems to work, however.

rabsef-bicrym:
<img width="342" alt="image" src="https://user-images.githubusercontent.com/52549148/197289015-4e26e0b4-724d-4f08-be83-5be4e6c7bb5b.png">
![image](https://user-images.githubusercontent.com/52549148/197289077-43b4a255-c302-4a11-87f3-218b923f129d.png)
^ when wet tries to join, it fails, but wet can see the group.

PR #1167, opened 2022-11-16
apps: handle flag creation safer

arthyn:
This adds some better normalization to the frontend as well as adds `(sane %tas)` checking in places where we're creating flags addressing #1024


liam-fitzgerald (approved):
LGTM!

PR #1170, opened 2022-11-16
groups: channel creation flow rework

rcrdlbl:
partially resolves: #692.

### what's missing?

**header image uploader**
there doesn't seem to be an option to add a header to a channel at the moment, this will require backend changes.

**Granular per-user read/write/view permissions**
AFAIK we haven't written out how exactly these are going to work on the code side of things, and we'll need to discuss this.

**Ship Type Access Permissions**
If these are already built in, i'll just need a quick explanation of which scries to make, if not this will need backend support.


arthyn:
@jamesacklin @urcades an issue here that came up in galen's run through and others have brought up is that channel name should come first since it's the only required field. do we want to address that here or are we standing by channel type first?

ui/src/channels/NewChannel/NewChannelDetails.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <header className="mb-6 flex items-center">
+          <h2 className="text-lg font-bold">Customize: {watchTitle}</h2>
+        </header>
+      </div>
+      <div className="flex flex-col space-y-6">
+        <label className="font-semibold">
+          {channelTypeName} Description
+          <textarea
+            // TODO: set sane maxLength
+            {...register('meta.description', { maxLength: 300 })}
```
arthyn, line 49: although this isn't marked as required, the form seems to require me to put something here. I don't think this is supposed to be required is it?

ui/src/groups/GroupAdmin/AdminChannels/ChannelPermsSelector.tsx
```diff
@@ -1,29 +1,41 @@
...
     description: 'Members can only view',
     readerSects: ['all'],
     writerSects: ['admin'],
   },
   secret: {
     title: 'Admin Only',
+    visibility: 'closed',
     description: 'Only Admin can view and write',
     readerSects: ['admin'],
     writerSects: ['admin'],
```
arthyn, line 41: @urcades @jamesacklin @jyng some ux feedback

It's weird to me that we're splitting out visibility making the choices match the backend structure more, but then in this second step we still use the synthesized combination of permissions. why not make this only be about write permissions? and if not, why split them in the first place? 

It's also unclear why these take place on two separate screens. Seems like permissions could be its own screen (and skippable so the default of read/write all is applied) and the optional description field could be moved to the first screen. Overall this feels like we're further complicating the process of making/editing channels. I do realize these may get more complicated down the line, but that feels pretty far off.

urcades:
> @jamesacklin @urcades an issue here that came up in galen's run through and others have brought up is that channel name should come first since it's the only required field. do we want to address that here or are we standing by channel type first?

@arthyn I actually do completely agree that the name field should be focused initially, regardless of whether it's the first form element that appears in the interface. This has always been the intent behind these "creation" interfaces!

The thinking behind situating the "channel type selector" before the name field was to grant people the opportunity to consider the apps/channel types available to them before naming the channel. This isn't a strongly-principled idea, but it felt like it might be beneficial to grant people a sense of the "social shapes" available for them to inhabit right off the bat, that there's more to a group's capability than chat, for example.

Regardless, I still feel like the channel title should be focused so people can start typing the name of the channel right off the bat. If y'all feel like it's a good idea to move the channel name field to the top of the form, I'm not opposed.

ui/src/groups/GroupAdmin/AdminChannels/ChannelPermsSelector.tsx
urcades: To be clear, there are basically two branching off points for how channels can be dialed in, permissions/access-wise:

Screenshot 2022-11-17 at 2 47 54 AM

In this first branch, someone's decided to create an "open" channel, which means it's open for everyone in the group to read and write to. The second/last screen in this process then outlines a series of choices that allow you to **chip away** (subtractively define) at the "everyone has access" access model: You can limit the "all access" to a specific subset of urbit types, you can set write to a smaller subset of people, etc.

Screenshot 2022-11-17 at 2 48 35 AM

In this second branch, someone has decided to create a "closed" channel, which means it's completely closed off to people in the group. The second/last screen in this process then promotes an additively-defined access model for the new channel: One has to select individual contributors or readers to invite to the channel.

~

In a nutshell, these two branches of defining a channel's access model stem back to whether you want the channel to be maximally open (available to all to read/write) or maximally closed (no one can read it unless you invite them). The second screen in each of these flows should in fact be optional, and as indicated, one should be able to create the channel right away after finishing the first screen. 

The access/permissions nuance these second screens surface are provided so people can create completely open channels that only a few can write to, completely open channels for everyone but comets, completely closed channels that only 3 other can contribute to, etc.

> Overall this feels like we're further complicating the process of making/editing channels.

As ideated, I actually found the designed paths to be pretty simple: The first screen is comprised of mandatory selections/definition (channel type, channel name, channel visibility), and the second screen is completely optional, and can be skipped. Intermixing optional fields with mandatory fields is where form-filling complexity + confusion begins to emerge in my experience.

I would actually be very OK with removing both the description and header (for supported channel type) fields so as to make the second screen truly streamlined and only related to permissions tuning, but I'd want to run it by @jamesacklin first

arthyn: > As ideated, I actually found the designed paths to be pretty simple: The first screen is comprised of mandatory selections/definition (channel type, channel name, channel visibility), and the second screen is completely optional, and can be skipped. Intermixing optional fields with mandatory fields is where form-filling complexity + confusion begins to emerge in my experience.

But since we're using a radio with a default value the visibility isn't actually mandatory to interact with.

Anyway as designed this doesn't follow our current permissions scheme so we'll have to push all this to post launch. The way @rcrdlbl has implemented this works as a hold over, so we'll have to go with that until we make these breaking changes.

@jamesacklin I'm inclined to say this PR should just wait until the permissions model matches post launch. It does allow people to edit channel descriptions which we currently have no way of doing, but also we don't surface those anywhere currently either. I'll let you make the call, but to me the design change needs the permissions update for it to actually work and not feel like a step down.

urcades: Yeah I hear you re: all points — we should definitely take some time to refactor with what we actually have, I actually like the pared-down form I think you're suggesting here, I've never particularly been a fan of frontloading creation flows with non-mandatory items, they just end up too beefy + a little intimidating.

rcrdlbl: so it's best to leave this open and add in @urcades fixes? wondering how long and how extensive the streamlining will get

jamesacklin: > this PR should just wait until the permissions model matches post launch.

?

arthyn: @rcrdlbl let's go ahead and close it, but leave the branch in the repo don't delete.

rcrdlbl: close without merging, correct?
it may auto-delete on us but i'll push from local again if that occurs

PR #1174, opened 2022-11-16
Catch disconnects caused by subscription errors, use errorCount

### LINKED ISSUE #1086: groups: disconnect not thrown on ship shutdown
To reproduce:
1. Open Groups in a browser window
2. Shut your ship down and do not close Groups

Expected behavior:
The disconnect notice in #1058 should appear and attempt to reconnect me.

---

Further steps:
3. Restart your ship with Groups still open

Expected behavior:
I should be reconnected to my now-running ship.

patosullivan:
Closes #1086 

It seems we were already showing the disconnect bar on ship shutdown, but we weren't catching errors with subscriptions themselves. This *should* solve that.

### DIFF
--- a/ui/src/api.ts
+++ b/ui/src/api.ts
@@ -5,8 +5,8 @@ import Urbit, {
   SubscriptionRequestInterface,
   Thread,
 } from '@urbit/http-api';
-import { useLocalState } from './state/local';
-import useSubscriptionState from './state/subscription';
+import { useLocalState } from '@/state/local';
+import useSubscriptionState from '@/state/subscription';
 
 export const IS_MOCK =
   import.meta.env.MODE === 'mock' || import.meta.env.MODE === 'staging';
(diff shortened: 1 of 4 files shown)


ui/src/api.ts
```diff
@@ -57,12 +57,18 @@ const api = {
         await setupAPI();
       }
 
+      client.onError = () => {
+        (async () => {
+          useLocalState.setState({ errorCount: errorCount + 1 });
+        })();
+      };
```
arthyn, line 64: This and the one in thread can safely be moved up into the `setupApi` call where we set the client. We don't need to set a new onError each time we poke/thread.

arthyn (changes requested):
Two things to think about, old groups had a third state 'reconnecting' do we think that would be useful? Also I think that we should probably have two error counts, one for the actual airlock onError and another for pokes/subs/threads, because 2 pokes that error is a lot less likely to be a disconnect than any errors coming from the airlock itself. perhaps the threshold for pokes/threads/subs could be ~4-5 and general airlock connection errors the threshold could be 1-2?

patosullivan:
@arthyn changes you requested have been added. we actually are using a "reconnecting" status, but it was only getting set if the user clicked the "Reconnect" button in the Disconnect banner. It'll now show when we attempt to reconnect after an airlock error.

ui/src/components/DisconnectNotice.tsx
```diff
@@ -42,7 +43,11 @@ export default function DisconnectNotice() {
   return (
     <div className="z-50 flex items-center justify-between bg-yellow py-1 px-2 text-sm font-medium text-black dark:text-white">
       <div className="flex items-center">
-        <AsteriskIcon className="mr-3 h-4 w-4" />
+        {subscription === 'reconnecting' ? (
+          <LoadingSpinner className="mr-3 h-4 w-4" />
```
arthyn, line 47: let's see how often we see this, if it's often let's consider something more solid state

arthyn (approved):
lgtm!

Issue #1178, opened 2022-11-16
talk: editing multi-DM form has wrong initial values


jamesacklin:
**Describe the bug, optionally with screenshots**
Editing a multi-DM overrides the currently-set image/color simply by initializing the form.

![Screenshot 2022-11-16 at 5 12 14 PM](https://user-images.githubusercontent.com/748181/202306093-e953e617-f4ba-4828-b505-cc2cbbbafb70.png)

**To Reproduce**
1. Go to a multi-DM in Talk
2. Edit the chat info via the context menu
3. Note that the default color overrides any changes you have made to the icon

**Expected behavior**
The form should be pre-populated with the values for club.meta already set.

**Additional context:**
- Self-hosted, Chrome/MacOS

PR #1179, opened 2022-11-16
talk: refurb multi-DM chat info

jamesacklin:
Doesn't touch #1178, but brings the multi-DM "Chat Info" dialog up to spec:

- Adds the contents of MultiDmHero without the member list
- Breaks the members and pending lists out into their own components
- Adds the ability to DM each member directly from this modal
- Switches the modal to "edit mode" on button click
- Gets you out of "edit mode" by clicking Cancel

With pending members:
image

With no pending members and image + name set:
image

Spec: https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/Product-Handoffs-(Locked)?node-id=1206%3A42842


ui/src/dms/MultiDMEditModal.tsx
```diff
@@ -1,21 +1,101 @@
...
+                  <div
+                    className="flex items-center space-x-2 rounded-lg p-2 hover:bg-gray-50"
+                    key={ship}
+                  >
+                    <Avatar size="small" ship={ship} />
+                    <div className="flex grow flex-col">
+                      <ShipName
+                        name={ship}
+                        showAlias={true}
+                        className="font-bold text-gray-800"
```
arthyn, line 81: this looks semibold in the mocks to me, can't tell though because it's unclickable in figma

ui/src/dms/MultiDMEditModal.tsx
```diff
@@ -1,21 +1,101 @@
...
+                      />
+                      <ShipName
+                        name={ship}
+                        showAlias={false}
+                        className="text-sm font-semibold text-gray-400"
+                      />
+                    </div>
+                    <Link to={`/dm/${ship}`}>
```
arthyn, line 89: needs an aria-label

arthyn (changes requested):
just two small things

ui/src/dms/MultiDMInfoForm.tsx
```diff
@@ -87,11 +91,14 @@ export default function MultiDMInfoForm({ setOpen }: MultiDMInfoFormProps) {
...
-          </DialogPrimitive.Close>
+          <button
+            onClick={() => setEditing(false)}
+            className="secondary-button ml-auto"
+          >
+            Cancel
+          </button>
           <button type="submit" className="button">
-            Done
+            Save
```
arthyn, line 101: this should probably use the new loading state tracker thing and I don't think it's included in the other issues.

jamesacklin: Want to handle it in the loader mega-issue, or here?

arthyn: lets do here since it's not tracked in the mega

rcrdlbl:
Other than hunter's comment, this looks great + easy to understand!

arthyn (approved):

PR #1180, opened 2022-11-16
groups/find: Introducing Globe16Icon

rcrdlbl:
Today, we're introducing Globe16Icon. 

This is the most semiotically accurate way to represent the notion of something being "Public" at 1rem invented so far. Basically, it's what we should have been using this whole time.


jamesacklin (approved):
Zamenhof would be proud

PR #1183, opened 2022-11-17
migration: sketch out working migration

liam-fitzgerald:
Depends on the `lf/gladio-2` branch living in landscape

### To run the migration
`-landscape!migrate`

### Notes
Shaping up pretty nicely, addressed most issues uncovered in the most recent testate run


desk/lib/chat-graph.hoon
```diff
@@ -28,21 +28,63 @@
...
-  ^-  inline:c
+  ^-  (list inline:c)
   ?-  -.con
-    %text       text.con
-    %mention    `@t`(scot %p ship.con)
-    %url        [%link [. .]:url.con]
-    %code       [%inline-code expression.con]
-    %reference  'elided reference' :: TODO fix?
+    %text       (fall (rush text.con apex:parse) ~[text.con])
+    %mention    ~[`@t`(scot %p ship.con)]
```
arthyn, line 48: I think we can use the actual mention type now

desk/app/chat.hoon
```diff
@@ -156,7 +156,7 @@
     |=  =chat:zero
     %*  .  *chat:one
       remark  remark.chat
-      log     log.chat
+      ::  log     *log.chat
       perm    perm.chat
       pact    pact.chat
```
arthyn, line 161: I just ripped out all the state migrations in jake's branch since we're nuking anyway

arthyn (commented):
I think the flow of this makes sense. Left a few questions. One general question is are we going to do all the desk/agent renames with a slew of pokes from %groups or how do we encode that in the OTA?

desk/app/groups.hoon
```diff
@@ -374,7 +407,7 @@
     ?+  module.config.metadatum  ~
       %chat  `%chat
       %link  `%heap
-      %publish   ~  :: TODO
+      %publish   `%diary  :: TODO
```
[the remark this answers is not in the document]
liam-fitzgerald: nope

arthyn (approved):
we've pored over this enough, lgtm

Issue #1123, opened 2022-11-10
groups: admin only channels various issues


arthyn:
there seems to be an issue with admin only channels where members can see them and also admins don't seem to be able to actually post in the channel, suggesting some type of permissions issue. The visibility might just be a frontend issue, but the posting is likely backend.

PR #1189, opened 2022-11-17
permissions: make sure we're checking correctly

arthyn:
This addresses #1123 mostly on the frontend, the backend will be handled by #1195. Writing issues still need to be dug into, but I suspect they were more issues with changing permissions itself.


ui/src/app.tsx
```diff
@@ -495,6 +495,7 @@ function RoutedApp() {
   };
 
   const theme = useTheme();
+  console.log(theme);
```
patosullivan, line 498: Can we ditch this extra console log?

patosullivan (approved):
This looks good, but it's going to conflict with some of Tom's changes in his open PR. I'm guessing this is what you guys talked about in the crew catchup this morning though, so you probably have some plan for that.

ui/src/app.tsx
arthyn:
```suggestion
```

PR #1196, opened 2022-11-18
app: fix media query deduplication

arthyn:
This showed up as an issue because I noticed our media query not picking up theme changes. Essentially on first render before the setQuery call can update the initialized flag, each invocation was creating a query listener. I moved the tracking outside of react state so that it's updated immediately.


rcrdlbl (commented):
LGTM + Makes the codebase cleaner!

tomholford (approved):
LGTM

For future reference these are available too:

https://usehooks-ts.com/react-hook/use-media-query
https://usehooks-ts.com/react-hook/use-dark-mode

Issue #1187, opened 2022-11-17
vscroll: load messages before the user gets to the top


arthyn:
We should reduce the amount of loads when scrolling and load more if the user is scrolling fast. Also we should try to anticipate what messages the user will need by scroll acceleration so that we can preload messages.

PR #1197, opened 2022-11-18
omnibus: scroller perf, devex enhancements 

### LINKED ISSUE #1200: chat: sending a message doesn't scroll to bottom sometimes
Sending a message with a `?msg=...` always scrolls to the focused msg and not the bottom.

**To Reproduce**
1. Click a reference link
2. Then try to send a message in the same channel
3. Notice window stays scrolled on previous focused message from ref link.

**Expected behavior**
Window should scroll to bottom when sending a message

tomholford:
# Context

This addresses #1187. It also includes additional performance enhancements and  devex tweaks.

# Changes

- [x] consistent localStorage key for prerelease banner
- [x] reuse firstUnread hook
- [x] virtual scroller performance enhancements
  - [x] dynamic page fetching - short initial page for quick render, then larger as user scrolls back
  - [ ] detect high acceleration and pause / prefetch higher amounts
     - initially attempted to implement using virtuoso scroll seek, ran into a known issue (https://github.com/petyosi/react-virtuoso/issues/410)
     - now [trying an approach](https://github.com/petyosi/react-virtuoso/issues/410#issuecomment-1320405297) recommended by the maintainer of `react-virtuoso`
- [x] remove channel prefetching; use `last.brief` to sort by recent instead 
- [x] upgrade `react-virtuoso` from 2.19.0 --> 3.1.4 to fix a [scroller bug](https://github.com/petyosi/react-virtuoso/issues/788) on iOS Chrome
- [ ] refactor scrollTo to fix #1200


ui/src/logic/usePrefetchChannels.ts
```diff
@@ -39,8 +43,20 @@ export default function usePrefetchChannels(flag: string) {
...
 
   const fetchAll = useCallback(async () => {
-    await asyncForEach(Object.keys(group?.channels ?? {}), fetchChannel);
-  }, [fetchChannel, group]);
+    // first, prioritize the recent channel
+    if (recentChannel && recentChannel !== '') {
+      await fetchChannel(recentChannel);
+    }
+
+    // defer the rest, in batches
```
patosullivan, line 51: Thank you!

patosullivan (approved):
Looks good and runs well locally.

arthyn (changes requested):
Testing this is giving me some really bad behavior just putting this here now before I can write it all out

arthyn:
So I haven't been able to capture this on camera, but in either watercooler or urbit app dev, on a fresh reload, if you scroll up as soon as you see messages and then scroll up again once the second set of messages load, sometimes you get stuck in like a purgatory of sorts where messages are almost all missing and going either up or down scrolls in the wrong direction.

Other than that, it seems like scrolling backwards now gets stuck more often than before.

tomholford:
@arthyn Okay thanks, will look into this

arthyn (approved):

PR #1204, opened 2022-11-18
groups/talk: updated color/image input field to account for empty default values

rcrdlbl:
it's about #1178. basically just had to force this thing to actually use the right "mode" when given a certain initial value.


arthyn (approved):
makes sense, lgtm

jamesacklin (approved):
This makes way more sense, thank you.

PR #1205, opened 2022-11-18
sidebar: recent sorting

arthyn:
Unfortunately I think I sent @tomholford down the wrong path for #1114 my apologies! The `brief.last` is the correct thing to use for the last message we just had a few issues. We weren't using nests, looking at all channel types, updating briefs for posts in all channel types, and our chat brief was going off out of order so it was always one behind. This corrects all of that and also changes the default sort for channels to arranged and default heap view to grid.

This does rip out the channel prefetching which I think we need to revisit after launch. The prefetching was incurring too many requests + subscriptions holding up loading anything else for like 10s+ which I think is not a decent enough trade off yet. Also we don't really want to be subscribed to all channel updates because that was why we split them out in the first place.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -1112,8 +1112,6 @@
       (put:log-on:c log.chat time d)
     =.  ca-core
       (ca-give-updates time d)
-    =.  cor
-      (give-brief flag/flag ca-brief)
     ?-    -.d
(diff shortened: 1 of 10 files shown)


tomholford (approved):
Pulled and played with it locally, nice work! Much better than the janky solution we had previously

PR #1031, opened 2022-10-26
diary, groups, heap: version negotiation

ryjm:
Includes state update for `%net` migration.

Resolves #950, #578, and #579 (WIP)

PR #1210, opened 2022-11-21
versioning: fix outstanding issues

liam-fitzgerald:
Fixes versioning so it can be merged
Notes:
- Unifies saga editing callsites
- Does not depend on kernel kicking the subscription so works correctly with fake marks


midsum-salrux:
```
gall: poke-as cast fail :groups [a=%json b=%group-create]
[%key 'secret']
/sys/vane/clay/hoon:<[832 51].[832 63]>
/sys/vane/gall/hoon:<[1.058 27].[1.058 45]>
```

getting this error in the JS console when I try to create a group to test this with

arthyn:
@midsum-salrux you'll need to run the UI preview, if you check out DEVELOPMENT.md it has instructions under getting started

desk/lib/wood.hoon
```diff
@@ -0,0 +1,31 @@
+::  $wood: logging
```
ryjm, line 1: nice, thoughts on adding levels / `%stir`-like poke too?

ryjm (commented):
looks good (logging lib yes plz) but why remove keep?

jamesacklin:
@midsum-salrux Can you reproduce this kick-loop in this branch? https://github.com/tloncorp/landscape-apps/pull/1031#issuecomment-1322263026

midsum-salrux:
> @midsum-salrux Can you reproduce this kick-loop in this branch? [#1031 (comment)](https://github.com/tloncorp/landscape-apps/pull/1031#issuecomment-1322263026)

Was testing with a pair of livenet moons that I got into a bad state. Going to try to get a more reproducible fakezod setup.

liam-fitzgerald:
yes @ryjm definitely should flesh out the logging lib a bit more, just
wanted to get something going

arthyn (approved):
As far as I can tell this looks good, chi/dex/lev seems consistent across the agents.

Issue #1191, opened 2022-11-17
scroller: double render part deux


tomholford:
Not sure how I got in this state, but now it happens on every message

![image](https://user-images.githubusercontent.com/16504501/202533663-e1c48135-f1cb-4ad1-ab4a-c3ab4addbf5c.png)

Issue #1172, opened 2022-11-16
chat: investigate id/sent time mismatch 


arthyn:
it seems these are not the same, additionally we seem to be indexing the pact on the frontend with one, `sent`, but then later when we go to check if a message is in the index we use the full `id`. likely cause of double renders and alec needs to be able to match between facts and scries of same messages.

PR #1214, opened 2022-11-21
chat: ensure time gets corrected and id + sent match

arthyn:
This hopefully fixes both the double renders #1172 #1191, and allows @alecananian to match a message from an ID between the ID and sent time or if you want you can match what we're doing here and also subscribe to chat `/ui` and listen for the chat action to correct the time.


alecananian:
Thanks for the code snippet, @arthyn - I should be able to mimic the functionality here with the `/ui` subscription 👍

patosullivan (approved):
Looks good!

Issue #1169, opened 2022-11-16
chat: references for old messages not loading until scrolled


tomholford:
1. Post a reference to a message deep in the backscroll in another channel
1. Refresh the page
1. Observe that it is not laoding the chat reference (spinner per the screenshot below)
1. Scroll back in the first channel until the referenced message is loaded
1. Switch back and observe that the chat ref is loaded

![image](https://user-images.githubusercontent.com/16504501/202100167-bde5146a-fdc8-46c0-8d84-1bde50b3186e.png)

Expected behavior:

Poke to query for and load a single reference message regardless of whether or not its in the client side message cache.

PR #1215, opened 2022-11-21
chat: fix for unloaded writ refs

patosullivan:
For #1169 

We can just use useWritByFlagAndId for all writ references, no need for a separate component for unsubbed/subbed writ references, and fixes the issue described.


arthyn (approved):
LGTM

PR #1219, opened 2022-11-22
talk: fix sort and persist filter

arthyn:
With recent change to `useAllBriefs` using nests, message sort need to be adjusted as well. Also threw in persisted talk filter because it's really annoying to watch it change lol.


tomholford (approved):
LGTM

Issue #1154, opened 2022-11-15
mobile: emoji picker occluded


tomholford:
![image](https://user-images.githubusercontent.com/16504501/201958838-f14ec822-41d7-4be0-8ecc-cb7ef754230a.png)

Is there a directive to use the native device picker when on mobile?

rcrdlbl:
IIRC we have custom emoji/stickers in the roadmap so we'll be unable to use the native one forever.

urcades:
To Dan's point, we should eventually accommodate for a world in which we're adding additional custom functionality to this picker. In the current desktop implementation, we position the picker popover well, and we should adapt the existing behavior to whatever custom popover picker we implement in the future.

On mobile (web and apps) this gets a little more tricky, but we should essentially do whatever we can to maximize the size of the picker and align our interfaces to device-native patterns. On both iOS and Android, emoji pickers end up taking over or extending the keyboard interface, and end up emanating from the bottom of the screen as an edge-to-edge view:

<img width="727" alt="Screenshot 2022-11-16 at 1 46 54 PM" src="https://user-images.githubusercontent.com/1195363/202266786-9437857d-1dab-46ff-a553-0dc9e79d4a24.png">

I was doing a bit of research and ended up discovering that forcing the default emoji keyboard on mobile OSes [might not be supported](https://stackoverflow.com/questions/37873103/how-to-force-emoji-keyboard-on-input-type-text), so I imagine the next iteration of our selection interface might be abstracted as:

1. Click on the "react" button on a message
2. Open up the device's native mobile keyboard for quickly searching emoji, stack a bottom-anchored, left-edge to right-edge emoji/sticker selection interface directly above the keyboard (see the discord example above)
3. Tapping outside of the selection interface should dismiss it

PR #1220, opened 2022-11-22
chat: emoji react selector mobile fixes

rcrdlbl:
see: #1154. also includes a blue background on emoji reacts you've selected

the latest version of emoji-mart seems notoriously difficult to self-style for width https://github.com/missive/emoji-mart/issues/634 so i'm using a hack with "show per line" to make it somewhat responsive. Also bumping up the font size on mobile to prevent mobile browsers from auto-zooming on the search input field.

### DIFF
--- a/ui/src/components/EmojiPicker.tsx
+++ b/ui/src/components/EmojiPicker.tsx
@@ -25,24 +25,6 @@ export default function EmojiPicker({
     load();
   }, [load]);
 
-  // if (isMobile) {
-  //   return (
-  //     <Dialog.Root open={open} onOpenChange={setOpen}>
-  //       <Dialog.Overlay />
-  //       <Dialog.Portal>
-  //         <Dialog.Content>
-  //             {data ? (
-  //               <Picker data={data} size={'2em'} previewPosition="none" {...props} />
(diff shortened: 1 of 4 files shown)


ui/src/components/EmojiPicker.tsx
```diff
@@ -15,17 +17,45 @@ export default function EmojiPicker({
...
+  //             ) : (
+  //               <div className="flex h-full w-full items-center justify-center">
+  //                 <LoadingSpinner className="h-6 w-6" />
+  //               </div>
+  //             )}
+  //        </Dialog.Content>
+  //      </Dialog.Portal>
+  //    </Dialog.Root>
+  //   );
+  // }
```
tomholford, line 45: Is this still needed?

tomholford (approved):
Tested locally, LGTM

localhost_3000_apps_groups_groups__zod_h_channels_chat__zod_c1(iPhone SE)

ui/src/components/EmojiPicker.tsx
arthyn:
```suggestion
```

PR #1226, opened 2022-11-22
scroller: refactor scrollTo behavior

### LINKED ISSUE #1200: chat: sending a message doesn't scroll to bottom sometimes

tomholford:
This refactors the scroller to use a provided ref, which subsequently allows other components (such as the Unread banner) to control the scroll index. This is then used to scrollback the index to the first unread message when clicking on the banner. Further, a useEffect hook is no longer used; instead, virtuoso's `initialTopMostIndex` is used on load if there is a `msg` param in the URL.

This resolves #1200

### DIFF
--- a/ui/src/chat/ChatScroller/ChatScroller.tsx
+++ b/ui/src/chat/ChatScroller/ChatScroller.tsx
@@ -5,13 +5,12 @@ import React, {
   useCallback,
   useEffect,
   useMemo,
-  useRef,
   useState,
 } from 'react';
 import { isSameDay } from 'date-fns';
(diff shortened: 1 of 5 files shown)


patosullivan (approved):
Code looks good, having an issue getting unreads to work on my dev moons today, can't test actually functionality.

arthyn:
seemed to work fine locally

PR #1235, opened 2022-11-22
groups: Add zindex to 'joining group' popover

patosullivan:
We had another z-index issue with the "You are currently joining this group" popover.


arthyn (approved):
lgtm

Issue #1195, opened 2022-11-17
groups: ensure reader permissions are enforced


arthyn:
We need to ensure that various group actions are actually protected by reader permissions. The readers sects of the channel must intersect with the sects of the user's vessel.

The two known areas to be enforced: 
- the scry for getting all groups
- joining a channel

There may be others, I'll add to this issue.

arthyn:
Also in the `watch-chan` arm, you shouldn't be able to get a preview without permissions.

PR #1237, opened 2022-11-22
Enforce reader permissions

midsum-salrux:
#1195 

This doesn't prevent seeing that secret channels exist yet, but you can't join them.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -318,6 +318,8 @@
   =/  [=flag:g =net:g =group:g]  i.gs
   ?.  (~(has by channels.group) nest)
     $(gs t.gs)
+  ?.  (go-can-read:(go-abed:group-core flag) src.bowl (~(got by channels.group) nest))
+    $(gs t.gs)
   =/  =preview:channel:g
     =,  group
     :*  nest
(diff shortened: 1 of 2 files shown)


desk/app/groups.hoon
```diff
@@ -635,13 +637,15 @@
       ?+    rest.pole  ~
           [%can-read src=@ ~]
         =/  src  (slav %p src.rest.pole)
-        ?:  =(%open -.cordon.group)  `loob+!>(&)
```
midsum-salrux, line 638: Note the removal of this check. Before, you could read everything in an open group, even secret channels.

arthyn: good catch

arthyn (approved):
this looks good to me as a first step. 

I'm wondering if it's possible to even prevent the channels themselves from ending up on the ship without permissions. right now we give all group updates unequivocally, but it seems like if a channel add comes in that's admin only it shouldn't even go to the receiving ship that's lacking. however, this makes sending updates much more complicated. I guess we could just guard for it on the receiving ship and if someone modifies their source to see them that's acceptable. @midsum-salrux @ryjm @liam-fitzgerald what do y'all think?

midsum-salrux:
> I'm wondering if it's possible to even prevent the channels themselves from ending up on the ship without permissions. right now we give all group updates unequivocally, but it seems like if a channel add comes in that's admin only it shouldn't even go to the receiving ship that's lacking. however, this makes sending updates much more complicated. I guess we could just guard for it on the receiving ship and if someone modifies their source to see them that's acceptable. @midsum-salrux @ryjm @liam-fitzgerald what do y'all think?

yeah, this is what I ran into. I'm OK with the solution you mentioned since even if they did modify their local source, they could only see the name of the secret channel, not actually join it.

ryjm (approved):
LGTG

 (sidenote: it's relatively easy to create channels associated with a group without initializing them. they'll never appear in the UI but show up in a `:chat +dbug` for example. might want to account for that at some point too)

PR #1244, opened 2022-11-22
groups: fix secret join and protect vs missing xeno

arthyn:
OTT, not sure if this is the issue but can't repro with the fix on. Also fixes erroneous logic for group joins now that we have secret groups.


tomholford (approved):
LGTM

midsum-salrux (approved):
LGTM

PR #1246, opened 2022-11-22
channel: prevent brief marking from triggering infinite loop

arthyn:
OTT, basically each time dismiss channel marked read it updated the brief, which updated the deps in its useEffect which in turn did it again.


rcrdlbl (approved):
lgtm

PR #1247, opened 2022-11-23
diary/Image: use UpdateValues

rcrdlbl:
added a new function that lets you manipulate diary node values directly. lets you actually use the s3 uploader now


arthyn (approved):
lgtm

PR #1255, opened 2022-11-23
heap, diary: use correct time for marking read

### LINKED ISSUE #1042: chat: unread marking sometimes loops

arthyn:
Similar to fix for chat, always use the last post's time + 1. Fixes #1042

### DIFF
--- a/desk/app/diary.hoon
+++ b/desk/app/diary.hoon
@@ -630,7 +630,9 @@
         %unwatch  remark.diary(watching |)
         %read-at  !!
       ::
-          %read   remark.diary(last-read now.bowl)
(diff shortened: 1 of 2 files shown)


tomholford (approved):
LGTM

Issue #1216, opened 2022-11-21
chat: we shouldn't load content references in chat while scrolling


patosullivan:
Virtuoso provides an isScrolling event property we can use to prevent render of certain items while scrolling to improve performance: https://virtuoso.dev/scroll-handling/

We could/should use this to prevent all of the network calls and renders that occur while scrolling through chats with lots of references.

first mentioned by @arthyn

PR #1256, opened 2022-11-23
groups: Prevent refs from loading while scrolling

patosullivan:
For #1216 

Notice that this does introduce some visual noise (switching from loading state to refs).

I tried to extract the scrolling state out into a hook, but the other components didn't seem to pick up on the changes in scrolling state from the hook so I ended up prop-drilling.

Might still be some jank in here. Please pull it down and play with it and feel free to make suggestions.

Also note that virtuoso has a built-in affordance for this type of behavior but it's meant to apply to all items in the list: https://virtuoso.dev/scroll-seek-placeholders/

### DIFF
--- a/ui/src/chat/ChatContent/ChatContent.tsx
+++ b/ui/src/chat/ChatContent/ChatContent.tsx
@@ -20,7 +20,7 @@ import { Link } from 'react-router-dom';
 
 interface ChatContentProps {
   story: ChatStory;
-  isScrolling: boolean;
+  isScrolling?: boolean;
 }
 
 interface InlineContentProps {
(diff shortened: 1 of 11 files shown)


arthyn (changes requested):
The problem I see here is that there's no remembering that we've already loaded a ref, so now if you scroll away and come back you'll get the loading spinner again. I think `useWritByFlagAndWritId` should probably take a param that blocks it from loading so we can pass `isScrolling`, and also we should make sure loaded refs are stored in the stores so we aren't reloading them and they can return their value immediately if they have it.

patosullivan:
@arthyn I'm not passing isScrolling into the various ref components, but we're still seeing re-renders on the writ and I'm not sure why. Any ideas?

arthyn (approved):
LGTM!

PR #1258, opened 2022-11-24
groups: hide empty sections from sidebar

rcrdlbl:
This also updates channelIndex to hide sections full of admin-only channels from users, and handles the "section full of unjoined channels" issue


patosullivan (approved):
Looks good, works well.

Issue #1253, opened 2022-11-23
chat: emoji picker runs off screen


jamesacklin:
Desktop-only.

![image.png](https://images.zenhubusercontent.com/6245cc0024d90711bcf4ea06/a7b92ce2-2734-4556-bac6-7feeb1a8cc7b)

rcrdlbl:
what platform/browser/size are you experiencing this at?

jamesacklin:
Chrome/MacOS.

This is in the initial response stage, in the message actions set of buttons off to the right.

rcrdlbl:
Yep, I haven't been able to replicate with that setup

patosullivan:
I noticed this on mobile view in dev tools on chrome as well.

jamesacklin:
This report comes from userland: 

![image.png](https://images.zenhubusercontent.com/6245cc0024d90711bcf4ea06/6ee32f4c-aa05-47eb-8cfb-718e951fe96f)

Firefox / MacOS

rcrdlbl:
Okay – I seem to be able to replicate this on the current beta branch (I have to set the window to fullscreen width) but am unable to do so going off this repo's master branch; can someone pull master and try to replicate it? I'll keep poking around in the meantime.

PR #1272, opened 2022-11-28
changed emoji picker popover to come out of left side of trigger

rcrdlbl:
see: #1253.

I had a lot of trouble recreating this locally, though I could get it to show up sometimes after checking out a pre-version bump commit – I think the version bump may have solved it?


arthyn (approved):
lgtm

PR #1273, opened 2022-11-28
clubs: prevent double 'join' and represent pending correctly

### LINKED ISSUE #1162: talk: joined dms/multi-dms show as unjoined on other devices
Take note of the double "Join" from jyng. I believe what happened here is that I had Talk open in Browser Window A and Browser Window B and clicked join for both. 

If I'm making changes and updates in one window they should be updated in the other window.

It looks like application state isn't shared across browser windows?

image

arthyn:
OTT, fixes #1162

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -735,10 +735,12 @@
         %team
       =*  ship  ship.delta
       =.  cu-core  (cu-give-delta delta)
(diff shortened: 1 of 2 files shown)


rcrdlbl (commented):
JS-Side this looks great!

midsum-salrux (approved):
LGTM

PR #1274, opened 2022-11-28
multi-dm: changed header truncation css class

rcrdlbl:
untracked issue – fixes header of multi-dm overflowing its boundaries if too many planets are added


arthyn (approved):
works!

PR #1275, opened 2022-11-28
chat: prevent users from copying refs in dm

### LINKED ISSUE #1259: talk: refs from the same chat not loading
Screenshot 2022-11-24 at 9 24 47 AM

**To Reproduce**
1. Create a new DM in Talk
2. Type text
3. Copy your text using the hover over button for copy ("reference")
4. Paste that reference in the DM
5. See spinner

**Expected behavior**
When pasted you should see the reference like you do in channels in groups.

**Additional context:**
- macOS, Google Chrome

patosullivan:
Fixes #1259 

Users should not be able to copy chat refs from DMs, this is not intended functionality. We happen to be using chat refs for "replies" in groups right now, so I can see why Zach attempted to use it in DMs, but refs were never intended to be used in DMs.

We *will* have replies in DMs, but that will come when we add proper replies for both groups and talk.

### DIFF
--- a/ui/src/chat/ChatMessage/ChatMessageOptions.tsx
+++ b/ui/src/chat/ChatMessage/ChatMessageOptions.tsx
@@ -23,6 +23,7 @@ export default function ChatMessageOptions(props: {
 }) {
   const { whom, writ, hideReply } = props;
   const groupFlag = useRouteGroup();
+  console.log({ groupFlag });
   const { didCopy, doCopy } = useCopy(
     `/1/chan/chat/${whom}/msg/${writ.seal.id}`
   );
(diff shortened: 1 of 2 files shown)


ui/src/chat/ChatMessage/ChatMessageOptions.tsx
```diff
@@ -23,6 +23,7 @@ export default function ChatMessageOptions(props: {
 }) {
   const { whom, writ, hideReply } = props;
   const groupFlag = useRouteGroup();
+  console.log({ groupFlag });
```
arthyn, line 26:
```suggestion
```

arthyn (approved):
lgtm

PR #1277, opened 2022-11-28
various: add version to %group-action mark

### LINKED ISSUE #1245: leaving a group and then rejoining messes up permissions with default channels
We don't clear any channel state when leaving a group which is fine I think, but upon rejoining a group that has default channels, when those channels join something about those channels actually existing previously is preventing state from getting updated correctly. We lose the group flag in `perm` at the very least which prevents you from writing to channels.

liam-fitzgerald:
Adds versions to marks, so that updates to permissions are handled correctly, instead of being dropped on the floor.

Fixes #1245

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -355,7 +355,7 @@
       (slog tank u.p.sign)
     ::
         %fact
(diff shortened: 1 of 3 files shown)


midsum-salrux (approved):
LGTM

PR #1278, opened 2022-11-29
chat: search backend

liam-fitzgerald:
Allows for searching either by ship, or by text


ryjm:
hell yeah, this means i can update `%graph-query` to work with new groups.

arthyn:
@ashelkovnykov is picking this up in another branch, tyfys @liam-fitzgerald

PR #1279, opened 2022-11-29
(wip) migration: ux and warnings

arthyn:
@urcades @jamesacklin here's the start of my attempt at signalling what's happening during migration. I ended up putting the same popover/tooltip over the pending migration button as in the sidebar. This is all inside the group, working on how we display that the group itself isn't migrated.

Screen Shot 2022-11-28 at 7 37 12 PM
Screen Shot 2022-11-28 at 8 38 08 PM


patosullivan (changes requested):
Code itself looks good, but I pointed this at my test ship and some groups are showing as migrated that aren't:

image

When I try to look at these groups, the channels themselves do show the orange dot/grayed out UI:

image

Is this right?

arthyn:
@patosullivan looks like that group is partially migrated maybe since one of the channels is joinable. who hosts the solutions channel?

liam-fitzgerald (approved):
LGTM!

jamesacklin (approved):
Good, great even.

Issue #1182, opened 2022-11-17
import/heap: links should be stored as the link content type


arthyn:
Currently our migration uses the link content type for heap links, but our current implementation stores them as plain text. We should probably store them as links in general and to increase compatibility with the migration.

jamesacklin:
Duplicate, covered in #1161

arthyn:
well I split this out because it's a frontend only change

jamesacklin:
![image.png](https://images.zenhubusercontent.com/6245cc0024d90711bcf4ea06/9cc0e0f9-600a-4ced-92a2-62ccf80367b9)

Use the `link` type here.

PR #1281, opened 2022-11-29
heap: store links as link content type

patosullivan:
For #1182 

Also, rather than returning `undefined` for titles we should just show the URL.


tomholford (approved):
Nice work LGTM

PR #1282, opened 2022-11-29
groups: usability fixes

rcrdlbl:
solving some overflow issues w/ long channel/user names (lots of setting the correct grow/shrink values) and makes the formatting toolbar's "link" feature actually work on phones


arthyn (approved):
this lgtm, all the shrinks make sense and playing around with it feels solid. curious about truncate vs line-clamp-1 use differences

rcrdlbl:
> this lgtm, all the shrinks make sense and playing around with it feels solid. curious about truncate vs line-clamp-1 use differences

I believe it's a "position: absolute/relative/fixed" and width: 100% interaction that goes wrong, as 'truncate' only works on desktop and `line-clamp-1` only works on mobile in these specific instances

Issue #1283, opened 2022-11-29
chat: respect code-fencing in messages


jamesacklin:
Chat messages need to parse or understand triple-backtick code-fencing in messages.

![image.png](https://images.zenhubusercontent.com/6245cc0024d90711bcf4ea06/b0785745-7abd-42d0-a1fb-1eae634ce298)

urcades:
Making a note here that we have some existing designs for code-fence message rows here:

https://www.figma.com/file/YLjErBShApNlWpE44nqnbC/Indigo-FW2022?node-id=2157%3A23839&t=wdtai91b9AJSQa7a-1

![Screenshot 2022-11-30 at 1 39 22 PM](https://user-images.githubusercontent.com/1195363/204881229-35926f90-fa40-4e92-bc06-1439f64b5d16.png)

Some markdown formats allow for an additional syntax label after the initial three ticks (ignore backslash):

```css
```css
...some code...
\```
```

And the thinking behind the "Hoon" label in the example above was to provide a visual affordance for indicating what type of syntax is being highlighted, if this is a feature we can easily support. If we're opting for a more simple initial implementation, we can ignore this label. 

The copy button feels like a nice affordance to keep, but again, if we're opting for a simple implementation, we can ignore this element.

PR #1290, opened 2022-11-30
chat: Handle code block input/display

patosullivan:
For #1283 

We need a way to differentiate inline code formatting vs code block formatting in the toolbar, @urcades . For now I used a right caret to indicate inline.

Video demo:

https://user-images.githubusercontent.com/1221094/204890576-d2c25220-e3ce-47bd-8eb7-0210f9e7170f.mov

### DIFF
--- a/ui/src/chat/ChatContent/ChatContent.tsx
+++ b/ui/src/chat/ChatContent/ChatContent.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import cn from 'classnames';
 import { ChatBlock, isChatImage, ChatStory } from '@/types/chat';
 import {
   isBlockquote,
@@ -125,7 +126,7 @@ export function InlineContent({ story }: InlineContentProps) {
 
   if (isBlockCode(story)) {
     return (
-      <pre className="rounded bg-gray-50 px-1.5">
+      <pre className="bg-gray-50 px-4">
         <code>{story.code}</code>
       </pre>
     );
@@ -166,6 +167,10 @@ export default function ChatContent({
 }: ChatContentProps) {
   const inlineLength = story.inline.length;
   const blockLength = story.block.length;
+  const firstBlockCode = story.inline.findIndex(isBlockCode);
+  // @ts-expect-error - this is a bug in typescript
+  // https://github.com/microsoft/TypeScript/issues/48829
+  const lastBlockCode = story.inline.findLastIndex(isBlockCode);
 
   return (
     <div className="leading-6">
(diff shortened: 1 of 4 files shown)


urcades:
Hi @patosullivan — just updated our icon set with an additional item for "Code Block":

https://www.figma.com/file/YLjErBShApNlWpE44nqnbC/Indigo-FW2022?node-id=89%3A1848&t=wdtai91b9AJSQa7a-1

Let me know if this works when dropped in, can tweak as necessary!

urcades (dismissed):
Apart from integrating a new/more distinct icon for "code block" input, I'd recommend we additionally increase the code block's internal margin/padding between the code text and the boundaries of the gray background it's contained within:

Screenshot 2022-11-30 at 3 59 27 PM

In the sketch above, the indigo-spec padding is set to 16p horizontal margin and 8p vertical margin in order to visually "center" the text within the container, ignoring the right-side margin as I assume we're allowing for horizontal overflow (no wrapping) of the text. We can use whichever tailwind stand-ins for the measurements indicated here.

patosullivan:
This actually builds fine, but vitest is not ignoring the ts error so the tests fail. Looking into it.

urcades:
I would not mind at all — here's a link to the 24px version:

https://www.figma.com/file/YLjErBShApNlWpE44nqnbC/Indigo-FW2022?node-id=89%3A1849&t=JE8dAqZkeZBqKrzi-1

arthyn:
still seem to be missing bottom padding:
Screen Shot 2022-12-01 at 1 04 24 PM

patosullivan:
> still seem to be missing bottom padding: Screen Shot 2022-12-01 at 1 04 24 PM

can't repro this, are you sure you pulled the latest?

arthyn:
@patosullivan yep this is on your branch latest

ui/src/chat/ChatContent/ChatContent.tsx
```diff
@@ -122,6 +124,14 @@ export function InlineContent({ story }: InlineContentProps) {
     );
   }
 
+  if (isBlockCode(story)) {
+    return (
+      <pre className="bg-gray-50 px-4">
```
arthyn, line 129: Seems like this should be `rounded bg-gray-50 py-2 px-4` and no need to do the checking in ChatContent

patosullivan: It *should* work like that, but there's a bug in tiptap that creates multiple lines for code blocks:
https://github.com/ueberdosis/tiptap/issues/1585

patosullivan:
fyi @urcades it looks like this: 
image

arthyn (approved):
lgtm!

urcades:
@patosullivan I was playing around with this a bit and I think it might help to re-order the options in this formatting bar such that all "inline" functions are grouped, and "block"-creating functions are grouped:

Screenshot 2022-12-05 at 3 49 50 PM

I definitely felt a little bit of visual imbalance here, and it helps somewhat to bundle the larger/more square-looking functions with one another. 

In the example above, I think it also helps to add a little bit of space between each group of icons — this lessens the amount of shape/size contrast we're seeing in the screenshot you took.

Issue #1286, opened 2022-11-29
refs: improve fetching


arthyn:
We need to add fetching of single curios and notes. We also should cache all refs so that they don't have to be fetched more than once.

PR #1297, opened 2022-12-01
diary, heap: ref fetching

arthyn:
This addresses #1286, adding remote note and curio fetching as well as keeping all remote refs cached so they are only fetched once. Also made curio times more similar to how notes and messages work and changed it so that any public group with open reader permissions can be read even if you're not joined to the group. So now, you can ref anything public and have it render.


desk/lib/curios.hoon
```diff
@@ -34,18 +34,21 @@
   (need (get time))
 ::
 ++  reduce
-  |=  [=time del=delta:curios:h]
+  |=  [now=time =time del=delta:curios:h]
   ^+  cur
   ?-  -.del
+    :: we use now on the add to enforce host ordering
+    :: any other actions should use time so that it's
+    :: the note that we're looking for.
```
liam-fitzgerald, line 42: This will use now on the add for both publisher and subscriber, so this comment is wrong.

liam-fitzgerald (changes requested):
generally looks good, want clarification on the now change though

desk/lib/notes.hoon
```diff
@@ -36,17 +36,14 @@
   (need (get time))
 ::
 ++  reduce
-  |=  [now=time =time del=delta:notes:d]
+  |=  [=time qt=time del=delta:notes:d]
```
liam-fitzgerald, line 39: why can't you just drop the time altogether and have one?

liam-fitzgerald (changes requested):
sry! want a clarification on something else

desk/lib/notes.hoon
arthyn: there's a difference between something being added vs referencing an existing note/curio/etc. we could handle this higher up to unify I think. basically we need to modify the diff's time to be `now` if it's an add otherwise it's a reference to something existing.

liam-fitzgerald: ok good

liam-fitzgerald (approved):
LGTM!

desk/lib/notes.hoon
arthyn: I'm doing all this because leaving as sender time felt wrong to me, could end up in bad UX if posts look out of order, etc.

PR #1298, opened 2022-12-01
migration: ux redux

arthyn:
This duplicates the changes I put in Liam's branch so that it actually gets globbed. It should work correctly with normal usage now and if not we should correct.


arthyn:
@liam-fitzgerald I'm wondering if we shouldn't just merge your migration branch instead

liam-fitzgerald (approved):
Bang! LGTM

PR #1302, opened 2022-12-02
migration: handle empty groups

arthyn:
OTT, this was in last redux but it was showing weird merge conflicts


arthyn:
going to merge since liam approved last one

Issue #1285, opened 2022-11-29
groups: "group avatar" corner radii styling is being overridden by `rounded-md`


urcades:
This image illustrates what's going on pretty precisely — basically we're overriding `rounded` radii styling (seen in the user's avatar/sigil) on group avatars with `rounded-md`, which is making them appear a little visually disjointed from how we've intended for left-aligned icons to appear in sidebars universally:

<img width="1470" alt="Screenshot 2022-11-29 at 2 05 33 PM" src="https://user-images.githubusercontent.com/1195363/204623032-dad429c0-6671-4b91-bf25-b6e0c926baf0.png">

We should eliminate the styling overriding here, and ensure that all left-aligned icon boxes / avatars are hewing to the `rounded` styling.

(Now that I have an up to date dev environment locally, I tried to find the source of the override myself but couldn't find it, would appreciate someone also pointing out where I can assess components individually in the source)

rcrdlbl:
components are available in the UI folder of the repo

jamesacklin:
This is likely `SidebarItem`

PR #1304, opened 2022-12-02
groups: channel join pending states, group list icon rounding

rcrdlbl:
see: #1264 and #1285.

#1264 is only partially resolved – we don't actually have the "pending state" of the channel accessible from the backend, so we'll need to find a good place to store it locally


arthyn (changes requested):
You need to follow the example of create group here which will basically give you a true backend "pending" state:

https://github.com/tloncorp/landscape-apps/blob/master/ui/src/groups/NewGroup/NewGroup.tsx
https://github.com/tloncorp/landscape-apps/blob/69815ac861927d77c5300022339a20075b2cfdea/ui/src/state/groups/groups.ts#L186

note the passing of onSuccess/onError and .track

PR #1307, opened 2022-12-05
migration: %bait placeholder

ryjm:
Adds a `%bait` field to `%cite` representing an imported reference.

a `%bait` will only appear through the migrate flow and only when the
group being imported is not the same as the group holding the referenced
source. This disctinction must be made because `%cite` may not contain
enough information to reconstruct the imported reference, since the host
of the given reference may not have migrated yet.

Thus, `%bait` is primarily a placeholder for a potentially unresolvable
reference that will be handled at a later date by the `%nark` app. This
will allow us to begin the migration without losing the necessary data
needed to reconstruct refs.

(This will be available to anyone who migrates, so an enterprising
developer could theoretically create custom reference types for
imported instances.)

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -1,4 +1,4 @@
-/-  c=chat, g=groups
+/-  c=chat, g=groups, n=nark
 /-  meta
 /-  ha=hark
 /-  e=epic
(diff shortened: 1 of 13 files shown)


liam-fitzgerald (approved):
Reviewed OOB, approved!
@arthyn anything needed for the FE end of this?

PR #1308, opened 2022-12-05
desk: add desk.release file

liam-fitzgerald:
Adds a desk.release file, so that integrating desks can check if the main net version has been installed


midsum-salrux (approved):
LGTM

PR #1311, opened 2022-12-05
migration: changing import flag tracking for better logic

arthyn:
Deleting the flags was removing a key piece of information so instead I changed it to work as a map and we just mark a boolean flag once import is complete. This way we know when to ignore migration guards for groups/channels. 

As it stands before this pr, if you were to make a completely new group/channel they would show as pending migration.


midsum-salrux (approved):
LGTM. A comment saying that true = already imported and false = not yet imported might be helpful though.

PR #1324, opened 2022-12-06
groups/diary: added not-prose to writ ref

rcrdlbl:
see: #1319


liam-fitzgerald (approved):

arthyn:
@rcrdlbl did you not need to add this to the other ref types?

rcrdlbl:
@arthyn i think the was only with the `<a>` tag that the Chat refs rendered interfering with the `prose-lg` tag in notebooks but i've put them in more places for safety

arthyn (approved):
lgtm, ty!

PR #1325, opened 2022-12-06
groups: finish initialising group before giving fact

liam-fitzgerald:
Fixes "unable to create channel after creating group"


ryjm (approved):
LGTG

PR #1343, opened 2022-12-07
app-ref: fix id and add subscribeOnce to api

arthyn:
OTT, this should fix remaining issues with app references


patosullivan (approved):
Looks good!

PR #1349, opened 2022-12-08
migration: only retry unjoined checking after 10 mins

arthyn:
This significantly reduces the subscription load since these will retry constantly if you're getting briefs.


patosullivan (approved):
looks good

PR #1350, opened 2022-12-08
Diary: fix issues with editing notes

patosullivan:
Fixes issues with editing migrated notes, also handles more cases for content within notes to display correctly in the editor.


arthyn (approved):
lgtm!

Issue #1398, opened 2022-12-09
groups: when disconnected, disable or hide chat input


zalberico:
**Describe the bug, optionally with screenshots**
- Have groups 2 open, turn off your ship
- See the yellow reconnecting bar at the top
- Try to type and send messages
- Observe messages disappear on enter without any indication to the user what's happening

There should be some indication to the user of what's going on when they do this (the yellow reconnecting is easy to miss and doesn't fully solve this anyway). The text they tried to send should also not vanish.

PR #1403, opened 2022-12-09
chat: prevent user from sending messages while disconnected

patosullivan:
For #1398 

Disables post the send button and the editor while a the frontend is disconnected.


arthyn:
yeah I agree with @rcrdlbl here

jamesacklin (approved):

rcrdlbl (changes requested):
looks good, except for the fact that we also need to modify `messageEditor's` `onEnter` to be aware of disconnects.

rcrdlbl (approved):
nice, LGTM

Issue #1396, opened 2022-12-09
chat: threaded chats truncating when navigating between


ImmanuelKant:
**Describe the bug, optionally with screenshots**

If I click on one thread and then click on a 2nd different thread, I'm only seeing the last message(s) in the 2nd thread.
In this case the 2nd thread had 3 replies

**To Reproduce**

Click the first thread (read circle):
![image](https://user-images.githubusercontent.com/28705048/206742733-2390058e-94d0-40eb-a730-2817d2346cff.png)

Then click the 2nd thread (red circle)
![image](https://user-images.githubusercontent.com/28705048/206742888-323df393-3b11-407a-8e71-1fd8de522b18.png)

The 2nd thread is not displaying correctly, only the last 2 comments are visible. The initial post + the 1st comment are missing

**Expected behavior**

The full thread should be visible.

PR #1407, opened 2022-12-09
chat/threads: added key to chatscroller in thread

rcrdlbl:
see: #1396.

the object being passed down to ChatScroller when the thread was changed was always correct, even when the issue was triggered (often when switching between threads quickly), so this was likely an issue with react being unable to tell that the thread being referred to by the thread component was actually different.

i've been clicking around for a while trying to trigger this and don't seem to be able to, but i highly recommend pulling this branch and testing yourself as part of the review.


patosullivan (approved):
LGTM

Issue #1354, opened 2022-12-08
groups: potential frontend resubscribe loop


Fang-:
Saw a huge, fast, output stream of the following on my ship:

<details><summary>printfs</summary>

```
%chat: on-watch on path /said/~bolbex-fogdys/watercooler-4926/msg/~rilfun-lidlen/170.141.184.505.967.92
9.684.771.413.931.777.536.819
%groups: on-peek on path /x/groups/~dabben-larbet/tlon/channel/chat/~bolbex-fogdys/watercooler-4926/can
-read/~palfun-foslup
%chat: on-watch on path /said/~bolbex-fogdys/watercooler-4926/msg/~rilfun-lidlen/170.141.184.505.967.92
9.684.771.413.931.777.536.819
%groups: on-peek on path /x/groups/~dabben-larbet/tlon/channel/chat/~bolbex-fogdys/watercooler-4926/can
-read/~palfun-foslup
%chat: on-watch on path /said/~bolbex-fogdys/watercooler-4926/msg/~rilfun-lidlen/170.141.184.505.967.92
9.684.771.413.931.777.536.819
%groups: on-peek on path /x/groups/~dabben-larbet/tlon/channel/chat/~bolbex-fogdys/watercooler-4926/can
-read/~palfun-foslup
%chat: on-watch on path /said/~bolbex-fogdys/watercooler-4926/msg/~rilfun-lidlen/170.141.184.505.967.92
9.684.771.413.931.777.536.819
%groups: on-peek on path /x/groups/~dabben-larbet/tlon/channel/chat/~bolbex-fogdys/watercooler-4926/can
-read/~palfun-foslup
%chat: on-watch on path /said/~bolbex-fogdys/watercooler-4926/msg/~rilfun-lidlen/170.141.184.505.967.92
9.684.771.413.931.777.536.819
%groups: on-peek on path /x/groups/~dabben-larbet/tlon/channel/chat/~bolbex-fogdys/watercooler-4926/can
-read/~palfun-foslup
%chat: on-watch on path /said/~bolbex-fogdys/watercooler-4926/msg/~rilfun-lidlen/170.141.184.505.967.92
9.684.771.413.931.777.536.819

...repeated many times
```

</details>

After reloading the groups tab I had open, things seem slightly more calm. Yet unclear what triggered this, but in case it's a clue, here's what was printed into scrollback right before the above started:

<details><summary>intro</summary>

```
[...]

PR #1414, opened 2022-12-09
subs: reduce ref fetching (again)

### LINKED ISSUE #1406: chat: endless on peek when looking @ a threaded response
When attempting to view a threaded response in UC, I am getting an endless stream of watch and peak alternation between groups and chat. Unclear if this is intended behavior.

image

image

arthyn:
This makes sure that we aren't attempting to initiate multiple requests for the same content at once. That state lives outside of the react state lifecycle so that it is immediately updated. This currently puts all retries at 10 mins. Fixes #1406 and #1354

### DIFF
--- a/ui/src/logic/subscriptionTracking.ts
+++ b/ui/src/logic/subscriptionTracking.ts
@@ -1,0 +1,37 @@
+export interface PreviewCheck {
+  inProgress: boolean;
+  attempted: number;
+}
+
+const DEFAULT_WAIT = 10 * 60 * 1000;
+
+export function getPreviewTracker(wait = DEFAULT_WAIT) {
+  const tracked: Record<string, PreviewCheck> = {};
+
+  const getPreviewTracking = (k: string) =>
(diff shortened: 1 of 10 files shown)


ui/src/logic/subscriptionTracking.ts
```diff
@@ -0,0 +1,37 @@
...
+      attempted: 0,
+    };
+
+  const isPastWaiting = (attempted: number) => Date.now() - attempted >= wait;
+
+  return {
+    tracked,
+    shouldLoad: (k: string) => {
+      const { attempted, inProgress } = getPreviewTracking(k);
+      console.log(k);
```
patosullivan, line 23:
```suggestion
```

patosullivan (approved):
Looks good and works well locally. One errant console.log

tomholford (approved):
Tested this with local fake zod and connected to my mainnet ship, feels good. LGTM

PR #1417, opened 2022-12-10
Docs broken: Fix for formatting in some udon files

vcavallo:
I was getting errors in docs UI:

docs

This PR addresses the errors in these docs files.


jamesacklin:
@vcavallo Nice to see you and thanks!

Is there a way to wrap to 80 characters that doesn't break docs?

jamesacklin:
OK, all good! We don't want broken docs, tyfys.

jamesacklin (approved):

Issue #1397, opened 2022-12-09
Something is filling up local storage:


patosullivan:
Reported in UC help. I also saw it on my own ship, hard refresh seemed to fix it. 

```
Encountered error:The quota has been exceeded.
setItem@[native code]
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:115:5794
d@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:115:6966
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:115:7070
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:1137:94952
N6e@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:97:6703
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:82:3901
UG@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:95:10840
UG@[native code]
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:93:39073
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:82:3901
lie@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:93:39020
vh@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:93:38954
Aie@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:95:11385
batchSet@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:116:20080
event@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:116:19565
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:116:9160
onmessage@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:116:4677
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:115:23154
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:115:22977
@https://socsel-pacted.arvo.network/apps/groups/assets/index.d96bf496.js:115:22694
asyncFunctionResume@[native code]
@[native code]
promiseReactionJobWithoutPromise@[native code]
promiseReactionJob@[native code]
```

PR #1434, opened 2022-12-12
talk: perf part 1 and local storage cleanup

arthyn:
Changes talk sidebar to be virtualized. Also introduces the ability for us to control when avatars load their images, like when scrolling. This should make going from channel back to all messages much faster.

Also threw in a fix for #1397 by removing fleet from persisted groups and fixed a few things with localStorage, like bad keys being saved or the fact that storage was named setting which was overriding persisted settings.


patosullivan (approved):
Code looks good, works well both in desktop and mobile viewports locally.

PR #1438, opened 2022-12-12
chat: keep unreads across tabs in sync

### LINKED ISSUE #1415: talk: unread status may be getting cached
If you go read chats in groups and then switch to talk you may still see them as unread.

arthyn:
OTT this fixes #1415


patosullivan (approved):
LGTM

PR #1445, opened 2022-12-12
groups: updated GroupAvatar to handle emoji + other unicode

rcrdlbl:
see: #1423.

This may still break if people use the multi-emoji sequences that display as one emoji (like the families where they all have different user-selectable skin tones) but it'll fall back relatively gracefully


patosullivan (approved):
Looks good and works well locally.

jamesacklin:
@rcrdlbl Solve that merge conflict and we're G2G.

PR #1462, opened 2022-12-13
chat: fix issue with chat scroller not loading older messages

### LINKED ISSUE #1437: chat: occasionally virtual scroller gets stuck unable to load older messages
Leaving a channel and coming back seems to fix.

patosullivan:
We're sometimes rendering Virtuoso without setting a firstItemIndex because the keys haven't loaded yet. We need to make sure we have keys before rendering Virtuoso.

Fixes #1437


tomholford (approved):
LGTM

PR #1472, opened 2022-12-13
On group join, ensure group is not already joined

midsum-salrux:
#1471


arthyn:
Going to close this until we revisit

Issue #1386, opened 2022-12-09
groups: admin edit actions succeed for group members, not reflected in local state


zalberico:
**Describe the bug, optionally with screenshots**
As an admin of the Tlon group I changed the group color, to me it didn't look like it worked but others on the call saw the change.

I also tried an incognito browser to see if it was a local cache issue, but I didn't see the change there either.

jamesacklin:
NB: this is for a _non-hosted_ but admin-roled user.

ngzax:
see https://github.com/tloncorp/landscape-apps/issues/1409 for what appears to be same issue.

jamesacklin:
@zalberico Are you still seeing this? Do you have a console log from the browser?

zalberico:
Okay - working on this with @arthyn, if I left the group and rejoined, then got promoted to admin I could see the changes update.

For groups I'm an admin of that I host on my ship it works fine.

Corrina also sees this issue and she didn't run double migrate, so we suspect this affects anyone who was an admin of a group that they don't personally host.

zalberico:
It'd be best if it was possible to ship a fix so people wouldn't have to leave and join the group.

More details:
 - We ran `:groups +dbug [%outgoing %ship ~bitpyx-dildus]`
 - In the broken state it returned no outgoing subscription
 - I left and rejoined the group and things were working
 - I ran the command again `groups +dbug [%outgoing %ship ~bitpyx-dildus]`
 - It returned the following:
```
%groups: on-poke with mark %dbug
>   [ wire=/groups/~bitpyx-dildus/tlon-support/updates
  agnt=[~bitpyx-dildus %groups]
  path=/groups/~bitpyx-dildus/tlon-support/updates/init
  ackd=%.y
]
> :groups +dbug [%outgoing %ship ~bitpyx-dildus]
>=
```

zalberico:
@arthyn and I confirmed that the release on wannec fixed this for me

tomholford:
> @arthyn and I confirmed that the release on wannec fixed this for me

Okay to close this out?

arthyn:
yep!

ngzax:
is this on the network yet? because if it is, i still can't update the setting for the Bookends group. (same host)

arthyn:
@ngzax it is, what's your %groups hash? and can you share what errors if any you see?

ngzax:
not sure what you mean by "hash?" do you mean ~darlur/bookends?

[...]

PR #1476, opened 2022-12-13
groups: restore missing subs from migration

arthyn:
OTT, this is a fix for #1386 it just loops through all groups on load and makes sure that we have a subscription if one isn't there. Need to double check that this is safe.


midsum-salrux (approved):
LGTM

arthyn:
@ryjm @liam-fitzgerald can one of you also take a look here?

PR #1488, opened 2022-12-14
channels: restore lost subscriptions

### LINKED ISSUE #1478: chat: channel is visible and has history, but cannot be posted to by group members
image

The same is also true for the Tlon > Bridge channel, so we can use that as a starting point for troubleshooting

arthyn:
Somewhere in the process of migration we lost subscriptions to the updates for certain channels. This should safely restore them if missing. Fixes #1478


ryjm (approved):
LGTM

PR #1490, opened 2022-12-14
chat: better handling of thread unreads

### LINKED ISSUE #1453: chat: mark thread as being unread if contained inside
currently if an unread is in a thread then you have to poke around to find it, we just need to add marking that it's the source of the unread

patosullivan:
Fixes #1453 

Shows a blue "unread" dot next to a thread if it contains the last unread message.

Also allows for a user to click "Click to view" button in the unread banner and get taken to the parent (replied-to message) in the chat scroller if the last unread message is in a thread under that message.

tl;dr: makes unread threads less annoying.

### DIFF
--- a/ui/src/chat/ChatUnreadAlerts.tsx
+++ b/ui/src/chat/ChatUnreadAlerts.tsx
@@ -1,11 +1,16 @@
 import React, { useCallback } from 'react';
 import { format, isToday } from 'date-fns';
 import { daToUnix, udToDec } from '@urbit/api';
-import bigInt from 'big-integer';
+import bigInt, { BigInteger } from 'big-integer';
 import { VirtuosoHandle } from 'react-virtuoso';
 import XIcon from '@/components/icons/XIcon';
 import { pluralize } from '@/logic/utils';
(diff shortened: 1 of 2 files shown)


ui/src/chat/ChatUnreadAlerts.tsx
```diff
@@ -18,6 +23,16 @@ export default function ChatUnreadAlerts({
...
+  if (maybeWrit) {
+    const [_, writ] = maybeWrit;
+    const replyTo = writ.memo.replying || '';
+    replyToId =
+      replyTo !== ''
+        ? bigInt(replyTo.split('/')[1].replaceAll('.', ''))
+        : bigInt(0);
+  }
   const markRead = useCallback(() => {
     useChatState.getState().markRead(whom);
```
arthyn, line 37: I know this isn't what you were trying to solve but can you change this too. It will make the X work instantly. You'll probably have to import useChatStore

```suggestion
    useChatState.getState().markRead(whom);
    useChatStore.getState().read(whom);
```

arthyn (approved):
lgtm!

PR #1491, opened 2022-12-14
perf: part 2

arthyn:
Optimize bootstrap sequence based on app and reduce group initial scry significantly until we need the full fleet.


ryjm (approved):
LGTM

Issue #1457, opened 2022-12-13
groups: can't leave a group that is awaiting migration from its host


lycurgus:
I'm in a few groups that are old/stale, where the host hasn't been on the network for a while. As the group hasn't "yet" migrated from my ship's perspective, I'm unable to interact with it, even to leave (which, btw, I hope would tell Groups not to bother trying to migrate it in future).

**To Reproduce**
1. Be a member of a group whose host is offline and hasn't/won't migrate the group
2. Click on the un-migrated group
3. Note that there is no option even to leave the group

**Expected behavior**
It would be nice if it was possible to leave a group that hadn't migrated yet, for cases where the user didn't expect it to migrate in future.

**Additional context:**
- Planet
- Self-hosted CLI
- Firefox 107 on Windows 11

PR #1506, opened 2022-12-15
groups: added leave button to unmigrated group popover

rcrdlbl:
see: #1457.

changed the "tooltip" to a "popover", as having an item that now has an interactable element that disappears when not hovering over the trigger is no fun.


patosullivan (approved):
Looks good and runs well. Switching over to a popover is a good idea.

PR #1508, opened 2022-12-15
ux: ESC to close an open thread

tomholford:
# Context

I have been reflexively hitting ESC to close threads, and so has ~simfur-ritwed

image

# Changes

This adds a keydown handler to open threads. On hitting ESC, the thread is closed.

https://user-images.githubusercontent.com/16504501/207866644-20dfb92d-1797-49b2-81f4-f69716071521.mp4

### DIFF
--- a/ui/src/chat/ChatThread/ChatThread.tsx
+++ b/ui/src/chat/ChatThread/ChatThread.tsx
@@ -1,6 +1,6 @@
 import cn from 'classnames';
 import React, { useRef } from 'react';
-import { useLocation, useParams } from 'react-router';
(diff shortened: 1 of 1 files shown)


patosullivan:
I like this, but imo it would be nice if we extracted keybindings out into a hook like this:

```
const action = useKeyBinding('close-thread')
```

along with an object that maps out all of the bindings. We could then use that object in the %garden settings store and allow the user to set their own keybinding preferences, as designed.

We could also easily add a vim mode 👀

arthyn (approved):
lgtm, we can do a comprehensive sweep later

Issue #1512, opened 2022-12-15
groups: paste image from clipboard to upload


eamsden:
**Problem**
Pasting an image in Talk or Groups has no effect.

**Proposed Solution**
Pasting image binary data from the system clipboard should initiate an S3 upload of the image data, as was the case in Groups 1.

tomholford:
add drop target to chat input

vcavallo:
yes please very much love this feature.

Issue #1509, opened 2022-12-15
chat: finish implementing images as designed


arthyn:
https://twitter.com/sor_wet/status/1603416813257121792 currently you can only send images with no other contents. The figma design had images as attachments similar to the way we do refs, we should do that soon.

jamesacklin:
See #1512 for a secondary requirement.

jamesacklin:
See #1425 as well (could close in favor of this).

Issue #1425, opened 2022-12-11
groups and talk: can't post image links after a newline


bonbud-macryg:
**Describe the bug, optionally with screenshots**
I can't post messages in this format in any group or channel in Groups or Talk:
```
test

https://0x0.st/onrp.png
```

Not only can I not post the message inside that code block, but I can't post the codeblock itself including the three backticks above and below. If you copy this codeblock into the chat input you'll see it removes the paragraph break; try to post without the paragraph break, just a newline, and the post will also vanish. The post isn't visible to me or anyone else in the group.

I've tried this on three channels across two groups, across Safari and Firefox, and across a planet and moon. Behaviour consistent in all settings.

**To Reproduce**
1. Copy the text in the code block above, or type it out manually
2. Try to post in any group channel, in Groups or Talk

**Expected behavior**
A post formatted as in the code block above should be posted to the group, and I'd expect that the image link would render in the channel as an image. If I posted "test" and the image link separately, I'd get the desired layout across two posts.

**Additional context:**
- Bug occurs on planet and moon
- Bug occurs on Safari 16.0 and Firefox 107.0.1

PR #1544, opened 2022-12-16
chat: image UX enhancements

tomholford:
# Context

Resolves several open issues regarding image UX in chat

# Changes

- [x] #1425 
- [x] #1509 
- [x] #1512 

# Preview

## Paste Single or Multiple to Upload with Caption

https://user-images.githubusercontent.com/16504501/208105471-c3d2ead4-1f94-4dbf-a9cb-4250fcc28e4f.mp4

## Drag and Drop

https://user-images.githubusercontent.com/16504501/208116734-660fbc18-40db-45d9-ac93-b38e0fa90ef5.mp4

## Click X to Remove

https://user-images.githubusercontent.com/16504501/208122484-19a783c7-63eb-4cb5-83b4-e00e2ab19d48.mp4


urcades (changes requested):
Oh man this is incredible @tomholford – the only minor change I’d suggest here is editing the label text in the input field to “images” from “references”, so that the label matches the content correctly, otherwise this is looking beautiful and I am so excited to use this.

jamesacklin:
Agreed, incredible. Two quick things:
- Can you add `.object-cover` to the image thumbnails to avoid squishing?
- We'll need some kind of indicative state (maybe a strong blue border) for the drop target while `isDragging`. People can learn this behavior but first-timers will need help discovering it.

tomholford:
> Oh man this is incredible @tomholford – the only minor change I’d suggest here is editing the label text in the input field to “images” from “references”, (as in `Attached: 2 images`) so that the label matches the content correctly, otherwise this is looking beautiful and I am so excited to use this.

Thanks Ed! Okay will do. It's possible to post multiple types of references at once, e.g. an image and a chat message reference. What should the copy say in this case?

patosullivan:
This looks awesome, but I did notice that if I paste an image link rather than pasting the image directly the image still shows up as full size in the chat scroller.

patosullivan:
Example:
image

The first image was a link, the second was just a paste. Same image.

tomholford:
> This looks awesome, but I did notice that if I paste an image link rather than pasting the image directly the image still shows up as full size in the chat scroller.

Thanks for the feedback, what is the expected behavior?

jamesacklin:
I think we want the smaller size for image links. We can follow up with another bug, I think?

vcavallo:
beautiful! very excited for this.

jamesacklin:
@tomholford If you update the screencaps, we can get a re-review from @urcades and we'll be good to go

patosullivan (approved):

urcades:
> Agreed, incredible. Two quick things:
> 
> * Can you add `.object-cover` to the image thumbnails to avoid squishing?
> * We'll need some kind of indicative state (maybe a strong blue border) for the drop target while `isDragging`. People can learn this behavior but first-timers will need help discovering it.

Very good catch – I didn't see the images were being squished in the mobile preview of the video.

> Thanks Ed! Okay will do. It's possible to post multiple types of references at once, e.g. an image and a chat message reference. What should the copy say in this case?
> 
> edit: we discussed out of band in the standup, going to show 'Images' if all attachments are images, otherwise fallback to 'References'

I think this is a fine solve, but after thinking about it for a bit, I think we could very well generalize the label copy to simply 'attachments', no need to distinguish between media types:

attachments

The proposed repositioning of the "Attachment" status bar to the top of the input is a nod to some future ideas we have behind placing a replies-signifying interface in that status bar as well.

PR #1550, opened 2022-12-16
chat: fix issue with scroll position on channel switch

patosullivan:
Should fix the issue we were seeing with scroll between chats. Please test it out and let me know if you still see it in some particular situation.

Only issue I see is if you add an emoji, remove it, and add it back the emoji will be hidden below the chat input.


jamesacklin (approved):
Looks good, works!

PR #1574, opened 2022-12-22
ux: ChatThread enhancements

tomholford:
# Changes

## ChatThread
- Currently when the writ data is still loading, an obnoxiously large 'No thread found' is rendered; this removes it in favor of rendering nothing until it's ready. Eventually we should consider a thread loading state
- Limit the scope of the Escape keydown listener to the containing div element. This addresses the issue where the user opens a thread, types a message, then clicks a user's profile within the thread to view their info modal, then hits escape to close the modal but also closes the thread, losing the drafted post. Now it will require re-focusing the thread context after dismissing the modal. Eventually we should add an `onunload` to prevent losing the drafted message, and / or restoring the save draft feature.


vcavallo:
> Currently when the writ data is still loading, an obnoxiously large 'No thread found' is rendered; this removes it in favor of rendering nothing until it's ready 

Interesting. I had been dismissing that "No thread found" window the moment I saw it, rather than waiting (I didn't know I should wait).

arthyn (approved):
lgtm

PR #1583, opened 2022-12-23
profile: reliable spacing for favorite group tags

### LINKED ISSUE #1241: profile (mobile web): Favorite Groups here need vertical margins
We need responsive margins here, I'm seeing that there's right margins being applied, so we should probably have that be 0 on mobile and then add vertical margin 2 on mobile.
IMG_4432

**Additional context:**
OS/browser: iPhone Mini 11 / Safari Browser

jamesacklin:
fixes #1241

Screen Shot 2022-12-23 at 12 48 15 PM


tomholford (approved):
Keep cranking! LGTM

PR #1594, opened 2022-12-28
fixed omission of channel group on json

polwex:
When scrying through the airlock, `diary` and `heap` channels were not showing the groups they belong to. That makes interop very hard. Please consider merging.


desk/lib/diary-json.hoon
```diff
@@ -96,6 +96,7 @@
     |=  p=perm:d
     %-  pairs
     :~  writers/a/(turn ~(tap in writers.p) (lead %s))
+        group/(flag group.p)
```
tomholford, line 99:
```suggestion
        flag/(flag group.p)
```

desk/lib/heap-json.hoon
```diff
@@ -60,6 +60,7 @@
     |=  p=perm:h
     %-  pairs
     :~  writers/a/(turn ~(tap in writers.p) (lead %s))
+        group/(flag group.p)
```
tomholford, line 63:
```suggestion
        flag/(flag group.p)
```

tomholford (changes requested):
Thanks for the PR, this will be useful for other integrations. Requested a minor rename inline (`group` > `flag`). Also, please update the corresponding `DiaryPerm` and `HeapPerm` Typescript types in `ui/src/types/diary.ts` and `ui/src/types/heap.ts`. Otherwise LGTM

arthyn:
@tomholford we probably should stay with `group` as the key to stay consistent with chat. Also makes it more clear what it's for.

tomholford:
@arthyn I addressed the PR feedback from both of us so we can close this one out. PTAL at your convenience

tomholford:
@jamesacklin Per discussion in standup, this PR does not need a state migration since it is not modifying any types. Good for merge upon approval

tomholford (approved):
LGTM

edit: @arthyn since I made some changes, PTAL at your convenience

PR #1612, opened 2023-01-03
chat: fix input clearing on new messages in DMs

### LINKED ISSUE #1404: talk: Text entry gets wiped when a new message appears (maybe just group DMs)
When drafting a new message in a group DM, if a new message comes in while you have text in the composition field, your text will disappear.  
~sicdev-pilnup and I have both experienced this in the same group DM. I have _not_ noticed it in one-on-one DMs, but perhaps they just aren't as active.

**To Reproduce**
1. Have a group DM with 3 or more participants
2. Begin typing a message...
3. While typing, have someone else send a message
4. Your composed message will be wiped.

**Expected behavior**
The text you're entering in the composer is never deleted except by you :)

**Additional context:**
- Ship type: planet
- Instance type: hosted on digital ocean VPS
- OS/browser: Brave 1.46.140 Chromium: 108.0.5359.99 (Official Build) (64-bit) on Ubuntu 20.04 focal

patosullivan:
fixes #1404

### DIFF
--- a/ui/src/dms/MultiDm.tsx
+++ b/ui/src/dms/MultiDm.tsx
@@ -27,10 +27,10 @@ export default function MultiDm() {
   const club = useMultiDm(clubId);
 
   useEffect(() => {
-    if (clubId) {
+    if (clubId && club) {
       useChatState.getState().initializeMultiDm(clubId);
     }
-  }, [clubId]);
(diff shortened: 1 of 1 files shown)


tomholford (approved):
LGTM

Issue #1426, opened 2022-12-12
groups: delay in clearing unreads


tinnus-napbus:
**Describe the bug, optionally with screenshots**

It can take up to a few seconds for the unread indicator to disappear after I've viewed the unreads

**To Reproduce**
1. See unread on group
2. open group
3. view channel with unread
4. return to group index
5. still unread for a while

**Expected behavior**
unread indicator should be cleared instantaneously upon viewing channel

**Additional context:**
- Ship type: planet
- Instance type: self-hosted
- OS/browser: qutebrowser & chromium on arch linux64

arthyn:
we need to clear the timer and mark final read on navigation away

patosullivan:
is this related to #1461 @arthyn ? if so, should be fixed in #1490

arthyn:
not related to the X, we set a delay upon seeing the banner after we hide it and the brief doesn't get marked read until that delay finishes. we probably should just call "read" on navigation away.

PR #1615, opened 2023-01-03
clearing notifications upon leaving chat even if timer hasn't finished

rcrdlbl:
see: #1426 
when the chat channel unmounts, if there's an "unread clear timer", we just mark the whole thing read.


arthyn (approved):
thank you been wanting this haha

Issue #1427, opened 2022-12-12
Inconsistent notification display


lukestiles:
**Describe the bug, optionally with screenshots**
On mobile, Tlon Internal shows as having unread messages in the group list. 

![image](https://user-images.githubusercontent.com/1903610/206944846-1011f8c4-e7ed-49c1-9115-70cbae2c6c46.jpeg)

When I click into the group, no channels have a blue dot indicating an unread message. 

![image](https://user-images.githubusercontent.com/1903610/206945037-5b0d091a-a707-4596-96e8-74dcb996a9a1.png)

**Expected behavior**
I'd expect to see a blue dot on at least one of the channels. 

**Additional context:**
- Ship type - planet
- Instance type - self-hosted on digital ocean
- OS/browser - Chrome on iOS

zalberico:
I don't think this is #1075 - I think the bug here is it requires a hard refresh to see the blue dots appear for the channels.

jamesacklin:
#1426 as well?

zalberico:
I think #1426 is also a different issue

PR #1626, opened 2023-01-04
chat: mark group notifications read on viewing group channels

### LINKED ISSUE #1475: groups: viewing a mention does not mark its notification read

If you're mentioned in a channel and view that mention it doesn't seem to mark the notification read - you seemingly have to actually click on the notification to view it or manually mark the notification read.

**To Reproduce**
1. get mentioned
2. view the mention in the channel directly rather than clicking on the notification
3. observe notification still inread

**Expected behavior**

viewing a mention should mark its notification read regardless of whether you got there via the notification

**Additional context:**
- Ship type: planet
- Instance type: self-hosted
- OS/browser: qutebrowser on arch linux

patosullivan:
Currently we show a blue "unread indicator" dot for groups for which a user has an unread non-new-message/meta notification (~sampel-palnet mentioned you, ~sampel-palnet joined, new channel added, etc). This PR allows us to mark those notifications as read upon viewing the associated channel in that group (or the group itself) so that the user doesn't have to individually mark-as-read or click on each notification on the notifications view.

Note that we don't show blue notification dots next to the channels themselves for this sort of thing. Should we?

Alternatively, we could *only* show the blue notification dot on groups for unread messages (and not these other types of notifications).

Fixes #1475 #1427 and maybe a few others.

### DIFF
--- a/ui/src/channels/ChannelHeader.tsx
+++ b/ui/src/channels/ChannelHeader.tsx
@@ -1,4 +1,9 @@
-import React, { PropsWithChildren, useCallback, useState } from 'react';
+import React, {
+  PropsWithChildren,
+  useCallback,
+  useState,
+  useEffect,
+} from 'react';
 import { Link, useNavigate } from 'react-router-dom';
 import cn from 'classnames';
 import * as Dropdown from '@radix-ui/react-dropdown-menu';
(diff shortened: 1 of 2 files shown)


urcades (commented):
> Alternatively, we could only show the blue notification dot on groups for unread messages (and not these other types of notifications).

I think this is the behavior we should support — somewhere along the lines we managed to correlate the "unread group dot" with its meta-activity, when the dot has always been designed to account for "unread social comms within channels".

patosullivan:
> 



> > Alternatively, we could only show the blue notification dot on groups for unread messages (and not these other types of notifications).
> 
> I think this is the behavior we should support — somewhere along the lines we managed to correlate the "unread group dot" with its meta-activity, when the dot has always been designed to account for "unread social comms within channels".

would you mind making a separate issue for this?

urcades:
Sure thing!

PR #1630, opened 2023-01-05
chat: fix notification links for mentions

### LINKED ISSUE #1528: landscape: @mention notifications in threads do not link to the thread
- If you're mentioned in an old thread you'll see the notification in landscape.
- If you click that notification it takes you to the chat channel
- If the thread is old (you have to scroll up to see the original comment) then nothing else will happen.
- If you happen to guess it's a thread and scroll up, when it loads then it may pop open on the side.

What should happen is clicking the notification brings you right to the thread (it'd even be nice if it highlighted the specific mention briefly or something).

patosullivan:
Fixes #1528, which was also an issue for *any* link to a mention.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -1360,7 +1360,7 @@
           ?:  (mentioned q.p.content.memo our.bowl)
             =/  yarn
               %^  ca-spin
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm, I added this and am struggling to remember why

Issue #1637, opened 2023-01-05
groups: sawSeam({all: null}) doesn't work


rcrdlbl:
`useHarkState.getState().sawSeam({ all: null });` seems to not work properly when called within groups. note that the poke goes through successfully on the browser side, however nothing shows in the dojo.

rcrdlbl:
note: it seems to work fine within garden, for some reason.

Issue #1265, opened 2022-11-28
groups: add "mark as read" to all notifications


jamesacklin:
Would assuage lots of notification anxiety.

Places to put this:
- in the All Groups view (all notifications for all groups)
- in the Notifications view in Landscape (all notifications for all apps)

Will need design input from @urcades on where to place. Feel free to continue with implementing the function; we can figure out where to place the control.

urcades:
I'm sure these recommendations won't come as a surprise, but here are the recommendations for placing a button in Landscape: ([figma](https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/branch/m1A533w3gFeDvJymTtUTim/Product-Handoffs-(Locked)?node-id=2025%3A91577&t=AXntuyraIcQKUFN9-1))

<img width="1249" alt="Screenshot 2022-11-29 at 11 53 15 PM" src="https://user-images.githubusercontent.com/1195363/204710639-d8dd6d98-d892-4977-89ec-b553268bf12d.png">

And here are the recommendations for a parallel placement in Groups: ([figma](https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/branch/m1A533w3gFeDvJymTtUTim/Product-Handoffs-(Locked)?node-id=1593%3A69221&t=AXntuyraIcQKUFN9-1))

<img width="1463" alt="Screenshot 2022-11-29 at 11 52 53 PM" src="https://user-images.githubusercontent.com/1195363/204710591-70c3d38d-26be-49a3-8d22-aa88be0d2f9f.png">

While the notifications stream is in a state of being scrolled, we can expect the header to stay fixed "above" it:

<img width="1350" alt="Screenshot 2022-11-29 at 11 55 36 PM" src="https://user-images.githubusercontent.com/1195363/204711073-0e0eb00d-aeb5-47cd-b39c-5b4cd27ae7d5.png">

urcades:
I'm making it a point to not suggest any clever interfaces in the above suggestions — the bifurcated "All Messages" and "Mentions Only" filter should be ignored for the time being, it's solely included in the above designs as I'm dialing in some additional alignment + geometry with the sidebar.

PR #1638, opened 2023-01-05
groups: mark all notifications as read

rcrdlbl:
see: #1265. (only partially resolved, the garden portion will be in a separate PR in the garden repo)

targeting the groups desk here, as writing this feature resulted in the discovery of #1637.


arthyn (approved):
lgtm

PR #1645, opened 2023-01-06
channels: edit flow

### LINKED ISSUE #1604: Groups: Renaming a channel kicks off a notification that a channel has been added
I renamed a channel from chat to EM/PM Chat (to get around the UX in talk of everything having the same name) and got a notification which implies a new additional channel was added instead of an existing channel being renamed. This notification should be: "chat was renamed to EM/PM Chat" or something like that.

Screenshot 2022-12-31 at 4 20 23 PM

tomholford:
# Context

Previously, both Add and Edit Channel frontend actions were using the same poke action: `%add`. This had the downstream effect of editing a channel triggering a 'created' notification. This PR resolves #1604 by adding an `%edit` poke and new 'edited' and 'renamed' notifications.

# Preview

## Editing Channel Metadata (but not the name)
image

## Editing the Channel Name
image

## Adding a New Channel
image


desk/app/groups.hoon
```diff
@@ -1424,6 +1424,32 @@
...
+        |=(=realm:zone:g realm(ord (~(push of ord.realm) ch)))
+      =.  channels.group  (put:by-ch ch channel.diff)
+      ?:  from-self  go-core
+      =/  link  (go-link /channels)
+      =/  yarn
+        %-  spin
+        :*  (go-rope /channel/edit)
+            link
+            `['Subscribe to channel' link]
+            ?:  &(!=(~ title.meta.prev) !=(title.meta.channel.diff title.meta.prev))
```
tomholford, line 1442: Question for the Hoon Crew: what's a more idiomatic approach to express this conditional?

arthyn:
This needs to include a state migration since we're changing the types here

desk/sur/groups.hoon
```diff
@@ -99,15 +99,17 @@
   ::
   ::  $diff: represents the set of actions you can take on a channel
   ::  
-  ::    add: create a channel, should be called from agent
+  ::    add: create a channel
+  ::    add: edit a channel
```
patosullivan, line 103:
```suggestion
  ::    edit: edit a channel
```

patosullivan (approved):
Looks great, one nit on a comment.

PR #1653, opened 2023-01-06
groups: remove extraneous migration checks

arthyn:
This removes the manual checking we were doing for unjoined channels now that the migration has stabilized. It also changes the verb flag to be off so that we aren't filling people's dojos with spam. Should help with urbit/urbit#6176


patosullivan (approved):
lgtm!

PR #1654, opened 2023-01-06
groups: only show 'mark as read' if unreads exist

jamesacklin:
On the tin. Only shows the "Mark All as Read" button in groups/notifications if there are any actual unread notifications.

image

After clicking: 

image

### DIFF
--- a/ui/src/notifications/Notifications.tsx
+++ b/ui/src/notifications/Notifications.tsx
@@ -6,7 +6,6 @@ import React, {
   useEffect,
   useState,
 } from 'react';
-import _ from 'lodash';
 import { Helmet } from 'react-helmet';
 import { Link } from 'react-router-dom';
 import { ViewProps } from '@/types/groups';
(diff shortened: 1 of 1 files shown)


ui/src/notifications/Notifications.tsx
```diff
@@ -65,6 +72,17 @@ export default function Notifications({
...
+
+  useEffect(() => {
+    const allNotifs: object[] = [];
+    notifications.map((day) => day.bins.map((bin) => allNotifs.push(bin)));
+    if (_.some(allNotifs, { unread: true })) {
+      setHasUnreads(true);
+    } else {
+      setHasUnreads(false);
+    }
+  }, [notifications]);
```
arthyn, line 85:
```suggestion
  const { notifications, count } = useNotifications(flag);
  const hasUnreads = count > 0;
```

arthyn (changes requested):
I think you can simplify this a little

jamesacklin:
Haha, I think I probably could :)

arthyn (approved):
lgtm

Issue #1357, opened 2022-12-09
notebooks: excessive newline insertion


zalberico:
**Describe the bug, optionally with screenshots**
If you write a note that contains a list, save the note, and then re-edit the note. It adds extra new lines.

Sometimes the formatting gets removed, sometimes it won't convert things to a list.

**To Reproduce**
1. Create a note
2. Write things in the note, including lists separated by a new line.
3. Publish and edit the note
4. See formatting issues

ixv:
This doesn't just happen with lists, if you click edit then publish without changing anything all paragraph breaks get an extra newline added to them.

PR #1657, opened 2023-01-09
notes: filter out extraneous paragraphs in note editor

patosullivan:
For some reason `JSONToListing` still seems to be generating extra newlines even though `limitNewlines` is set/defaulted to true in `tiptap.ts`.

To get around this, I'm filtering out extraneous "paragraph" blocks on the JSON content before passing it to the editor.

For #1357


arthyn (approved):
lgtm. I'm ok with this to stop the bleeding, but we should edit the tests and fix the parser in a follow up pr

PR #1658, opened 2023-01-09
groups/channels: rank level bans

### LINKED ISSUE #1593: group-store: Existing and new rank-level bans are not enforced.
I have recently noticed that comets are able to join and interact with groups I administer, despite these groups having a previously effective rank-level ban of comets (a.k.a. _'Antikomet Aktion'_) for reasons of accountability. I know that a number of other groups do the same for similar reasons, and I presume that these groups are similarly affected, though I have not seen any other reports of this. This regression is probably associated with Landscape 1. **This regression may represent a privacy and/or security risk, as well as a risk of serious nuisance in light of #1429.**

**To Reproduce**
1. Spawn a comet.
2. Using the comet, join a group with an extant rank-level ban of comets.
3. Observe that you can enumerate channels, read chats, and post chats.
4. Leave the group.
5. As group host, run the previously effective command `:group-store &group-update-0 [%change-policy [~YOUR_GROUP_HOST %YOUR_GROUP_NAME] [%open [%ban-ranks (sy ~[%pawn])]]]` to rank-ban comets.
6. Rejoin the group as the comet.
7. Observe that the command given in step 5 had no effect.

**Expected behaviour**
Rank-level bans should be effective. Banned ranks should not be able to interact with affected groups, and should probably not be enumerated in the search interface. Alternatively, the search interface should indicate that enumerated groups are inaccessible to the searching user based on their rank and rank-level bans (as well as any per-user bans that might be in effect, but that's a different matter).

**Additional context:**
- Ship type: planet
- Instance type: self-hosted, CLI
- OS/browser: Vivaldi on Linux

arthyn:
This fixes #1593 on the backend and also displays messages on the frontend explaining that you can't join a group if they have banned you or your rank.

This should cover the joining from the frontend, but also anyone trying to be sneaky from the terminal whether they are trying to join a group or a channel since this also updates the can-read/can-write checks to include a check for banned status.

Screen Shot 2023-01-10 at 4 11 03 PM

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -721,6 +721,18 @@
       out
     (~(put in out) who)
   ::
+  ++  go-is-banned
+    |=  =ship
+    =*  cordon  cordon.group
+    ?&  =(-.cordon %open)
+        ?-  -.cordon
+            ?(%shut %afar)  |
+            %open
+          ?|  (~(has in ranks.ban.cordon) (clan:title ship))
+              (~(has in ships.ban.cordon) ship)
+          ==
+        ==
+    ==
   ++  go-pass
     |%
     ++  leave
@@ -870,6 +882,10 @@
         [%fleet ship=@ %is-bloc ~]
       =/  src  (slav %p ship.pole)
       `loob+!>((~(has in go-bloc-who) src))
+      ::
+        [%fleet ship=@ %is-ban ~]
+      =/  src  (slav %p ship.pole)
+      `loob+!>((go-is-banned src))
       ::
         [%channel app=@ ship=@ name=@ rest=*]
       =/  nes=nest:g  [app.pole (slav %p ship.pole) name.pole]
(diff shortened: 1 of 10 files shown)


desk/app/groups.hoon
```diff
@@ -1264,17 +1289,19 @@
     ^+  go-core
     =/  user-join  &((~(has in ships) src.bowl) =(1 ~(wyt in ships)))
     =/  am-host  =(p.flag our.bowl)
+    =*  cordon  cordon.group
```
tomholford, line 1292: Curious to know as a new hooner: what is the purpose of using a deferred expression here vs `=/`?

arthyn: @Fang- any insight here?

tomholford:
Pulled and played with this locally, so far everything LGTM. Asked @arthyn for a poke to set a rank ban so I can test the full flow.

tomholford:
Here's the poke for future reference:

```
:groups &group-action [[~zod %test] [now [%cordon [%open [%add-ranks (~(gas in *(set rank:title)) ~[%earl])]]]]]
```

tomholford (approved):
WOMM, LGTM :+1: 

image

desk/app/groups.hoon
Fang-: It's just "alias" vs "new variable" basically. `cordon.group` might be too long for how frequently you want to use it, so you alias it to something shorter (`cordon`). In most cases this is more appropriate than `=/`, because of that semantic difference, and also changing `cordon` here would update the original in `cordon.group` I believe.

tomholford: Wonderful, thanks!

PR #1662, opened 2023-01-10
diary: fix wrap paragraphs bug

patosullivan:
Better fix for #1357, the extraneous newlines bug.


arthyn:
This seems to have the opposite problem when testing locally. If you make a post with varying amounts of empty space between paragraphs, after subsequent edits those empty paragraphs start to get culled when they shouldn't be.

arthyn (approved):
lgtm

PR #1670, opened 2023-01-10
kelvin: declare support for 416

philipcmonk:
This release is coming soon: urbit/urbit#6194.  It has no breaking changes that affect these apps, except that `+team:title` will no longer gives access to our moons (which is not something we depended on).  This diff declares support for both the current kelvin and that new one.


arthyn (approved):
lgtm

PR #1700, opened 2023-01-12
tiptap: fix small codeblock issues

arthyn:
The previous fix didn't carry the `codeWithLang` flag through the entire parsing so would break on notebooks. Also for some reason the code block editor does not give a default language and our JSON parser on the backend expects a non-empty string so I added a default.


patosullivan (approved):
Lgtm

Issue #1679, opened 2023-01-12
chat: pasting a reference link into chat input on mobile doesn't create an attachment/chat block


patosullivan:
Seen on brave on Android:

![Screenshot_20230111-204733_Brave.png](https://user-images.githubusercontent.com/1221094/211963983-29628e4c-d737-4aca-a404-3cf24b992ef4.png)

PR #1701, opened 2023-01-13
chat: fix paste issues on mobile chrome

patosullivan:
For #1679 

I *think* this is caused by both a peculiarity in the way focus works on Android when pasting and the newish onPaste handler we used for chatblock images.

Using the tiptap commands was the much more straightforward way of handling the href issue (not clear how to correctly update or replace a node without a specific selection in ProseMirror).

### DIFF
--- a/ui/src/chat/ChatInput/ChatInput.tsx
+++ b/ui/src/chat/ChatInput/ChatInput.tsx
@@ -21,7 +21,7 @@ import AddIcon from '@/components/icons/AddIcon';
 import ArrowNWIcon16 from '@/components/icons/ArrowNIcon16';
 import useFileUpload from '@/logic/useFileUpload';
 import { useFileStore } from '@/state/storage';
-import { IMAGE_REGEX, isImageUrl } from '@/logic/utils';
+import { IMAGE_REGEX, isImageUrl, isValidUrl } from '@/logic/utils';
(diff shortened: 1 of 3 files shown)


ui/src/chat/ChatInput/ChatInput.tsx
```diff
@@ -224,6 +226,19 @@ export default function ChatInput({
...
+      // this is necessary because the editor doesn't
+      // keep focus on mobile in some cases (chrome/android).
+      if (text && isValidUrl(text)) {
+        messageEditor.commands.selectAll();
+        messageEditor.commands.setLink({ href: text });
+        messageEditor.commands.selectTextblockEnd();
+      }
+    }
+  }, [messageEditor, text]);
+
```
arthyn, line 241: should this just be in the paste since this won't work if you paste a link in the middle of some text?

patosullivan: If you interact with the input it should just convert it to a clickable link anyway. We don't have access to this editor instance in the onPaste within the MessageEditor component itself afaik.

arthyn: ah ok

arthyn (approved):
lgtm

Issue #1703, opened 2023-01-13
chat: continuous requests slowing down load


arthyn:
There seems to be some kind of dependency loop which causes continuous marking of read on a group.

PR #1705, opened 2023-01-13
channels: constant rerender from hook

arthyn:
Possible partial fix for #1703


tomholford (approved):
worked during our pair sesh, LGTM

PR #1708, opened 2023-01-13
talk: minimal viable chat-cli for groups 2

Fang-:
What with the incoming release of session support for the (web)terminal, it'd be a shame if we didn't have an app for which that feature made some sort of sense. So here it is, chat-cli updated to work with groups 2's chat.

The good:
- Supports most of the features chat-cli supported.
- Includes session support.
- Pays respects to times long past: `/app/talk`.
- Tries its best to gracefully render gui-oriented contents such as `%img` and `%cite`.

The bad:
- No DMs or group DMs support.
- Does not even attempt to resolve message references, except for user-prompted history lookup with `;123` commands. We can scry for messages, but cannot scry for an _existence check_ of messages, so the content scry may crash if we don't know the message the reference points to.
- Took a somewhat shoehorn approach to message rendering. Most messages should display fine, and we do chat-cli usual best-effort at inline url rendering. But expanded messages no longer auto-open the url they contain, we do not make use of output styling capabilities where possible, and are overall just a lower-fidelity experience.
- No reacts support. 😮

The ugly:
- Passes smoke tests. Renders a handful of strange messages I threw at it. But haven't done Deep Testing™.
- What to do about dependencies? Do I just hard-copy /lib/shoe into this repo?
- Should we auto-start this? Users will want to `|link %talk` manually anyway, but should we require them to `|start %groups %talk` once?
- Does not import or clean up old chat-cli state at all.

Probably best to just review 32cd9ea standalone for now. 68f41c6 just imports the old chat-cli that had been updated for nu-term compatibility. Obviously requires zuse 416 (terminal release).

### DIFF
--- a/talk/app/talk-cli.hoon
+++ b/talk/app/talk-cli.hoon
@@ -1,13 +1,7 @@
-::  chat-cli: cli chat client using chat-store and friends
+::  talk-cli: cli chat client using groups' chat
 ::
 ::    pulls all known messages into a single stream.
 ::    type ;help for usage instructions.
-::
-::    note that while the chat-store only cares about paths,
-::    we mostly deal with [ship path] (aka target) here.
-::    when sending messages (through the chat hook),
-::    we concat the ship onto the head of the path,
-::    and trust it to take care of the rest.
 ::
 /-  chat
 /+  shoe, default-agent, verb, dbug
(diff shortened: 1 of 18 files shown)


vvisigoth:
What is the blocker on DMs/group DM support?

Fang-:
Me taking the time to figure out what is up with the distinction between those and "normal" chats in the chat api, and figuring out how to account for that in talk(-cli)'s implementation. Decided to skip on it because I wanted something usable out asap. I might return to this, but giving this as a task to an apprentice could also work.

arthyn:
My only issue is the agent name being talk will likely be confusing with the fact that we have a desk named talk and they aren't necessarily connected. Other than that, this is really cool. LMK if you have questions about the rest of chats.

jamesacklin:
@Fang- Do you mind renaming so we don't have even more desk/app/agent name confusion? I'd love to get this merged and out.

Fang-:
I already renamed to `talk-cli` in 8a8d736 above. As mentioned to @arthyn out of band, I'm intentionally not using `chat-cli` here to avoid shenanigans around migrating old state, and it still being running in the groups 1 desk, etc.

If `talk-cli` isn't distinct enough, let me know. If it makes more sense, I'd also be happy to ship it as part of the Talk desk instead, since the functionality offered here is more in line with that than the full Groups suite.

jamesacklin:
> in the Talk desk

would actually be quite good, @arthyn what think you?

arthyn:
yeah I think moving to %talk makes a lot of sense

Fang-:
Alright, it's in the talk desk now. Since neither this nor the main `/desk` seem to contain all external dependencies, I haven't included any here yet either. This only really needs `/sur/chat` and `/lib/shoe` and their dependencies.

For usage, users will need to `|start %talk %talk-cli` and then `|link %talk-cli`.

arthyn:
So right now we dynamically add base-dev and garden-dev at deploy time, for the %groups desk Jake and Liam just added any other deps directly. Ideally we'd keep the repo clean from deps but might not be feasible. Should we just copy them over right now?

arthyn:
I say lets copy lib/shoe and can we do a symbolic link for sur/chat? then it'd be ready to go

Fang-:
I symlinked all of the dependencies from this repo, but still hard-copied the ones we indirectly pull from urbit/urbit's `/pkg/landscape` (old groups).

`/lib/shoe` is part of base-dev, but I hard-imported it anyway, since changes to it might break the code herein, and we may not always want/need to use its exact latest version.

arthyn (approved):
lgtm!

PR #1711, opened 2023-01-16
perf: omnibus

### LINKED ISSUE #1617: talk: opening DMs in Talk makes Brave peg a core
Whenever I open up Talk in my browser to view "All Messages", I notice that the CPU utilisation of one of my cores (on the machine my browser is on, not the machine I'm running my ship on) skyrockets. Caught it in the little CPU monitor panel plugin I have, but screenshot from `btop`:

image

**To Reproduce**
1. Open Talk.
2. Watch as the CPU % number goes up.
3. Use heat generated by computer to warm house.

**Expected behavior**
Ideally, viewing my messages doesn't use a ton of CPU.

**Additional context:**
- Ship type (comet/moon/planet/star/galaxy): star (~sorwet) and planet (~nartes-fasrum)
- Instance type (Tlon-hosted/self-hosted with Port/CLI): self-hosted with CLI
- OS/browser (Safari on MacOS, Chrome on Windows, Firefox on Linux, etc.): Brave	1.46.133 Chromium: 108.0.5359.71 (Official Build) (64-bit) NixOS

### LINKED ISSUE #1631: perf: switching message list filter in Talk
When using Talk and switching view contexts (e.g. All Messages to Direct Messages), a user with many messages can experience dropped frames and an unresponsive browser.  Received a Devtools profile from a user with an example, here's a screenshot:

image

tomholford:
# Context

This PR has a smorgasbord of performance improvements. It resolves #1617, #1631 
as well as a few other untracked issues (aggressive scroll handler, Sidebar jank, 
unnecessary re-renders). Per the Results section below, **it improves the script 
execution performance between 30-40%**. Worth noting that this addresses a subset 
of the outstanding issues. There is likely significant room to further improve perf in a 
future dev cycle.

# Changes

- [x] virtual scroll for Groups list in Sidebar
- [x] memoization of frequently rendered / expensive components, and prop values
  that don't change between updates
- [x] replaced inline definition of prop values to reduce re-renders
- [x] refactored scrollhandler in Sidebar
- [x] added devex tool why-did-you-render
- [x] upgraded react-router 5 --> 6 (fixes a perf issue)
- [x] Vite 2 --> 3 migration (prerequisite for why-did-you-render)

# Results

## Talk: Toggling Sidebar Filter

Scenario: toggling the Talk sidebar filter from Direct Messages --> All Messages and back

**~40%** reduction in script execution time 

### Before
image

### After
image

## Groups: Navigating Between Large Groups

Scenario: navigating between a couple large groups with many members, messages, and channels

**~30%** reduction in script execution time 

### Before
image

### After
image

# TODO

- [x] ~figure out why vite is bundling `wdyr` even when the condition in `main` 
  is false~ it seems like this is not currently possible; filed an issue in Vite's repo:
https://github.com/vitejs/vite/issues/11714


ui/package.json
```diff
@@ -114,8 +114,8 @@
     "react-image-size": "^2.0.0",
     "react-intersection-observer": "^9.4.0",
     "react-oembed-container": "^1.0.1",
-    "react-router": "^6.3.0",
-    "react-router-dom": "^6.3.0",
+    "react-router": "^6.6.2",
```
arthyn, line 117: This changed the way relative links work, currently getting a bad redirect when entering a group where the previous recent channel sends me to '../../activity'

tomholford: Good catch; I spiked into updating our usage of relative paths, but since there are many it would be non-trivial. Something we should consider in the future. For now, downgraded back to 6.3.0 (the release before the breaking change).

arthyn (approved):
lgtm

PR #1721, opened 2023-01-17
utils: correct image url regex

### LINKED ISSUE #1677: groups/talk: message consisting of text and then image link does not get posted

When posting a message that consists of text, followed by an image link, the message disappears from the input box when Enter is hit, but the message is not sent. Sending a message purely consisting of text or an image link works fine.

**To Reproduce**

Into groups or talk type:

1. "hi foo bar"  ::  works
2. "https://upload.wikimedia.org/wikipedia/en/thumb/5/5f/Original_Doge_meme.jpg/300px-Original_Doge_meme.jpg"  ::  works
3. "hi check this out https://upload.wikimedia.org/wikipedia/en/thumb/5/5f/Original_Doge_meme.jpg/300px-Original_Doge_meme.jpg"  ::  does not work

**Expected behavior**

Message gets posted.

**Additional context:**
- Planet
- Self hosted from CLI
- On linux desktop, within brave browser. Both Talk and Groups.

arthyn:
Fixes #1677. This should only ever match a string that is entirely an image URL not a partial of a string.

### DIFF
--- a/ui/src/logic/utils.ts
+++ b/ui/src/logic/utils.ts
@@ -335,7 +335,7 @@ export const VIDEO_REGEX = /(\.mov|\.mp4|\.ogv)$/i;
 export const URL_REGEX = /(https?:\/\/[^\s]+)/i;
 export const PATP_REGEX = /(~[a-z0-9-]+)/i;
 export const IMAGE_URL_REGEX =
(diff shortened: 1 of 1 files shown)


jamesacklin (approved):
Phew!

patosullivan (approved):

PR #1725, opened 2023-01-18
chat: use normal sendMessage in MultiDm

### LINKED ISSUE #1696: Talk: Text entry field wiped when new messages arrive
Identical to https://github.com/tloncorp/landscape-apps/issues/1404

I haven't noticed this in a while, but started to notice it again today.

patosullivan:
Fixes #1696 

The useSendMultiDm hook was causing a re-render. We don't need to use a special sendMessages function for multiDMs after we've already created the multiDM, so I pulled it out.


tomholford (approved):
Nice, LGTM. Should the unused hook be removed?

patosullivan:
> Nice, LGTM. Should the unused hook be removed?

no, we still need it for creating new multiDMs

PR #1726, opened 2023-01-18
talk: store ships in ship selector locally in NewDm

### LINKED ISSUE #1484: talk: switching pane/window while inputting message target clears input field
Go to 'New Message', enter some text in the 'Search for Urbit ID..' field, and then switch panes or windows and switch back.  The text will have been cleared, but the more intuitive behaviour is that it be preserved.

**Additional context:**
- Brave on macOS.

patosullivan:
Fixes #1484


arthyn (approved):
lgtm

tomholford:
Does this need to be unset on send / create of new DM?

tomholford (approved):
One question in the comments, otherwise LGTM

patosullivan:
> Does this need to be unset on send / create of new DM?

good catch, yes.

PR #1734, opened 2023-01-18
chat: fix chat input focus loss

### LINKED ISSUE #1729: chat: chat input loses focus after sending a message in group chat channels
OTT

patosullivan:
Fixes #1729 

It turns out this was caused by including `onReference` in the dependency array of the useCallback in handlePaste.

I also removed the useEffect we originally had in ChatInput for clearing text if no draft exists (didn't cause the issue we were seeing, but it was useless anyway and could cause other issues, I imagine).

### DIFF
--- a/ui/src/components/MessageEditor.tsx
+++ b/ui/src/components/MessageEditor.tsx
@@ -93,13 +93,6 @@ export function useMessageEditor({
       if (!whom) {
         return false;
       }
-      const text = event.clipboardData?.getData('text/plain');
-
-      if (text && isRef(text)) {
-        onReference(pathToCite(text));
-
-        return true;
-      }
 
       if (
         uploader &&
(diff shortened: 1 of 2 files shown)


patosullivan:
this breaks chat blocks for refs on android. investigating further.

patosullivan:
We're going to have to figure something else out for https://github.com/tloncorp/landscape-apps/issues/1679

For now I removed all reference logic from the handlePaste function.

ui/src/components/MessageEditor.tsx
```diff
@@ -93,13 +93,6 @@ export function useMessageEditor({
       if (!whom) {
         return false;
       }
-      const text = event.clipboardData?.getData('text/plain');
-
-      if (text && isRef(text)) {
```
tomholford, line 98: If the pasted item is a ref, is it still handled as expected without this conditional branch?

tomholford (approved):
One question inline, otherwise LGTM

ui/src/components/MessageEditor.tsx
patosullivan: Yes, this was just a way to force it since it wasn't happening using the existing method on Android.

Issue #1683, opened 2023-01-12
all groups erased


tinnus-napbus:
**Describe the bug, optionally with screenshots**

I had switched to `~binnec-dozzod-marzod` as my groups/talk source as it fixed the d.replied.at bug https://github.com/tloncorp/landscape-apps/issues/1643 & was recommended there. Upon seeing the new talk/groups announcement from the urbit dev mailing list I switched my sources back with `|install ~sogryp-dister-dozzod-dozzod %groups` (and same for talk). When I reopened the groups front-end there were no groups listed in the sidebar. Refresh, cache-busting refresh, different browser didn't help. I tried switching back to `~binnec-dozzod-marzod` but they were still missing.

when i do the following scry it returns nothing:

```
> .^(groups:g %gx /=groups=/groups/noun)
{}
```

it looks like all of my groups were actually erased?

**Expected behavior**
should not erase all groups

**Additional context:**

```
> +vat %groups                                                                                                                                                                                                       
%groups                                                                                                                                                                                                              
  /sys/kelvin:      [%zuse 417]                                                                                                                                                                                      
  base hash:        0vs.n20oi.r18b8.ruh8k.250eh.iummk.p8udg.8l23j.b9628.n3sao.j2gsd                                                                                                                                  
  %cz hash:         0vs.n20oi.r18b8.ruh8k.250eh.iummk.p8udg.8l23j.b9628.n3sao.j2gsd                                                                                                                                  
  app status:       running                                                                                                                                                                                          
  force on:         ~                                                                                                                                                                                                
  force off:        ~                                                                                                                                                                                                
  publishing ship:  ~                                                                                                                                                                                                
  updates:          remote                                                                                                                                                                                           
  source ship:      ~sogryp-dister-dozzod-dozzod                                                                                                                                                                     
  source desk:      %groups                                                                                                                                                                                          
  source aeon:      6                                                                                                                                                                                                
  kids desk:        ~                                                                                                                                                                                                
  pending updates:  ~[[%zuse 416]]                                                                                                                                                                                   
::                                                                                                                                                                                                                   
> +vat %talk                                                                                                                                                                                                         
%talk                                                                                                                                                                                                                
  /sys/kelvin:      [%zuse 417]                                                                                                                                                                                      
  base hash:        0v1a.6gfhh.oeeuq.4b5nn.nibtq.va6jd.cogfr.11i6o.2rg1f.sto1s.06sqf                                                                                                                                 
[...]

PR #1735, opened 2023-01-18
chat: reuse existing multi dms

### LINKED ISSUE #1704: New multiparty DM created for same recipients of prior multiparty DM
When using the flow to create a new multiparty DM a new multiparty DM is created every time even if there is an existing multiparty DM with the same participants.

**To Reproduce**
1. Select + New Message
2. Enter participants
3. FInish
4. New multiparty DM created

**Expected behavior**
Instead of creating a new DM each time, the existing one should be loaded.

arthyn:
This fixes #1704 ensuring that multi dms get recycled when available. Also included a fix for the cause of #1683, which is a rip out of the state nuking stuff since that really needs to go asap.


ui/src/dms/NewDm.tsx
```diff
@@ -21,7 +22,29 @@ export default function NewDM() {
...
+      }
+
+      return key;
+    }, '');
+  }, [multiDms, shipValues]);
+  const who =
+    ships.length > 0
+      ? isMultiDm
+        ? existingMultiDm
+          ? existingMultiDm
```
tomholford, line 44: Confirming this nested ternary checking for `existingMultiDm` in sequence is expected?

ui/src/state/chat/chat.ts
```diff
@@ -900,7 +911,13 @@ export function useDmIsPending(ship: string) {
 
 const selMultiDms = (s: ChatState) => s.multiDms;
 export function useMultiDms() {
-  return useChatState(selMultiDms);
+  const dms = useChatState(selMultiDms);
+
+  useEffect(() => {
+    useChatState.getState().fetchMultiDms();
+  }, []);
```
tomholford, line 918: Confirming that multi DMs should be scried every time `NewDm` component is visited? Or is the initial subscription on application load sufficient?

tomholford (approved):
Grateful for the ☢️ removal! A couple questions inline, otherwise LGTM

ui/src/dms/NewDm.tsx
arthyn: Hmm this probably could be simplified

ui/src/state/chat/chat.ts
arthyn: I didn't want to add another scry to our initial load and scrying every time someone visited new dm felt decent because it's pretty infrequent. We should probably revisit in another perf PR sometime soon.

PR #1738, opened 2023-01-19
talk: DM Navigation UX enhancements

### LINKED ISSUE #1540: talk: starting a new DM with existing DMee with should navigate there
If you search a patp to dm someone you already have a dm with, it finds the patp and once you enter text on the blank screen it brings you to your existing one.

If your existing one is already in your sidebar, it should bring you there as soon as you search the patp.

tomholford:
If user enters ship(s) from an existing DM or multi-DM, navigate to it; and, keep the ShipSelector focused so that additional ships can be entered to continue. Hitting enter again will open the message.

This resolves #1540.

*note the end of this clip shows #1763, which  means that it is creating the new DM successfully :)*

https://user-images.githubusercontent.com/16504501/213981019-7c660a5d-9c96-4fab-a8f5-05ba91a83143.mp4


tomholford:
@urcades Any feedback on this design? We could navigate automatically, but per the conversation in #1540, that would then prevent the user from entering additional patps to start a multi-DM.

patosullivan (approved):
lgtm

urcades (changes requested):
Adding the feedback from standup here, to describe a further iteration:

- We can reduce the amount of context switching (from keyboard input to mouse usage) here if we do the work to invoke a "preview"/existing view of an _existing  dm_ upon autocompleting a name in the input field
- For example, if I've previously DMed James and somehow forgotten I've DMed him, and I'm constructing a new DM with James' name, I'd ideally be shown a "preview" of our past DM upon autocompleting his name in the "New DM" field.
- Upon autocompleting the name of someone I've already DMed, and seeing the preview of our previous chat, if I press `ENTER` while my insertion point is still in the field, I should be brought to our prior chat as though I'd navigated to it manually, and the "New DM Input" should be dispelled as though I had clicked "Cancel"
- After autocompleting a single person's name in the input field whom I've already DMed, I should continue to be granted the opportunity to continue typing to add more individuals to the new DM creation field, to construct a Group DM. This behavior is why I call the above renders of previous DMs while autocompleting "previews". The same behavior should persist to Group DM construction — if I type in both James and Romina, and we've shared a prior Group DM with one another, we should try and render a preview of the existing Group DM

@tomholford let me know if the above is clear — I'm happy to throw together some sketches that visually describe the stated behavior, but as you'd imagine, there are no "new" interfaces being described here, simply new state for existing interfaces that reincorporate existing interfaces.

tomholford:
> Adding the feedback from standup here, to describe a further iteration:
> 
> * We can reduce the amount of context switching (from keyboard input to mouse usage) here if we do the work to invoke a "preview"/existing view of an _existing  dm_ upon autocompleting a name in the input field
> * For example, if I've previously DMed James and somehow forgotten I've DMed him, and I'm constructing a new DM with James' name, I'd ideally be shown a "preview" of our past DM upon autocompleting his name in the "New DM" field.
> * Upon autocompleting the name of someone I've already DMed, and seeing the preview of our previous chat, if I press `ENTER` while my insertion point is still in the field, I should be brought to our prior chat as though I'd navigated to it manually, and the "New DM Input" should be dispelled as though I had clicked "Cancel"
> * After autocompleting a single person's name in the input field whom I've already DMed, I should continue to be granted the opportunity to continue typing to add more individuals to the new DM creation field, to construct a Group DM. This behavior is why I call the above renders of previous DMs while autocompleting "previews". The same behavior should persist to Group DM construction — if I type in both James and Romina, and we've shared a prior Group DM with one another, we should try and render a preview of the existing Group DM
> 
> @tomholford let me know if the above is clear — I'm happy to throw together some sketches that visually describe the stated behavior, but as you'd imagine, there are no "new" interfaces being described here, simply new state for existing interfaces that reincorporate existing interfaces.

@urcades Made the changes as suggested and updated the screencap in the description, please take a look! Could use your help prettifying the presentation, but otherwise the UX should be as described in the above comment.

urcades:
@tomholford this is so close, the interactive state you've built in is so good — basically the only improvement I can think of here is that rather than stacking the default channel header with the DM construction field below it, we can probably keep the "dm construction field" intact and set in place throughout the process, and not render the original header until ENTER is pressed (and the existing DM is invoked).

This seems fine to me, given the construction field itself representing the list of individuals associated with the DM, and in the case of the desktop view, the existing DM being represented in the sidebar.

tomholford:
> @tomholford this is so close, the interactive state you've built in is so good — basically the only improvement I can think of here is that rather than stacking the default channel header with the DM construction field below it, we can probably keep the "dm construction field" intact and set in place throughout the process, and not render the original header until ENTER is pressed (and the existing DM is invoked).
> 
> This seems fine to me, given the construction field itself representing the list of individuals associated with the DM, and in the case of the desktop view, the existing DM being represented in the sidebar.

@urcades good call, just made the change and it's much cleaner!

image

If this looks good, feel free to :heavy_check_mark:  so we can merge and get some more eyes on it

tomholford:
@patosullivan @arthyn Made some significant changes since the initial review request, and have addressed @urcades' feedback; so, this is now ready for re-review

urcades (approved):
Yes, stamp

arthyn (approved):
this is really good!

PR #1743, opened 2023-01-19
chat: avoid giving useMessageEditor any deps that change on submit

arthyn:
OTT this should solve the most recent case of #1729

### DIFF
--- a/ui/src/chat/ChatInput/ChatInput.tsx
+++ b/ui/src/chat/ChatInput/ChatInput.tsx
@@ -110,14 +110,14 @@ export default function ChatInput({
   // update the Attached Items view when files finish uploading and have a size
   useEffect(() => {
     if (
-      whom &&
+      id &&
       files &&
       Object.values(files).length &&
       !_.some(Object.values(files), (f) => f.size === undefined)
     ) {
       // TODO: handle existing blocks (other refs)
       useChatStore.getState().setBlocks(
-        whom,
+        id,
         Object.values(files).map((f) => ({
           image: {
             src: f.url, // TODO: what to put when still loading?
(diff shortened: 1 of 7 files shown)


patosullivan (approved):
This looks good and runs well locally!

patosullivan:
Looks like we still have one CI issue though.

patosullivan:
I'm still losing focus on mobile :(

patosullivan (changes requested):
The chat input focus issue seems to persist on mobile (at least on my Android phone).

arthyn:
ok lemme check it out

arthyn:
@patosullivan should work now

tomholford (approved):
LGTM

patosullivan:
It's broken on desktop now for me:

https://user-images.githubusercontent.com/1221094/213582458-fbe1e8de-f3c8-4643-864d-9d1eee39ec0c.mov

patosullivan:
Disregard the last comment (I was looking at groups while running the talk frontend), it works fine on desktop.

There is an issue with mobile losing focus and then regaining it, though. The keyboard disappears and then pops back up, it's pretty jarring.

patosullivan (changes requested):
Need to figure out a different way to get focus back without triggering the keyboard to disappear and reappear on mobile.

patosullivan:
https://user-images.githubusercontent.com/1221094/213583641-cfb48ed1-956a-4148-8355-17104d0b972b.mp4

Ignore cringe discord notifications pls

patosullivan (approved):
We can't repro the issue I'm seeing, we should probably merge for now and see if we can get a repro on another android phone internally.

PR #1750, opened 2023-01-20
chat: fix sigil alignment

### LINKED ISSUE #1748: chat: fix alignment of sigil with chat input
looks like this at the moment:
image

jamesacklin:
fixes #1748

image


jamesacklin:
Unfortunately, this creates this situation. I think it might be fine, since I don't want to get into measuring the height of the input field and reactively positioning the avatar.

image

patosullivan (approved):
Works for me!

Issue #1749, opened 2023-01-20
chat: keyboard closes and re-opens on send on some Android devices


patosullivan:
I've been able to repro this on two different phones running Android 13, in both Brave and Firefox.

patosullivan:
I can repro the focus loss/regain in chrome on desktop by using the send button rather than pressing enter to send.

PR #1751, opened 2023-01-20
chat: fix keyboard and input focus issues on android

patosullivan:
For #1749 

Switching from onClick to onMouseDown and throwing in an e.preventDefault() prevents the button from stealing focus from the input.

This also makes the explicit messageEditor focus() command unnecessary.

to test: log out messageEditor.isFocused in the main body of the ChatInput component, note that it stays focused when you click the button.

### DIFF
--- a/ui/src/chat/ChatInput/ChatInput.tsx
+++ b/ui/src/chat/ChatInput/ChatInput.tsx
@@ -190,7 +190,7 @@ export default function ChatInput({
 
         sendMessage(whom, memo);
       }
-      editor?.chain().setContent('').focus().run();
+      editor?.commands.setContent('');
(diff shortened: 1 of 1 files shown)


tomholford (approved):
LGTM

arthyn:
Did this work without switching to mouseDown? Seems like the `event.preventDefault()` might do it by itself. This stops a keyboard enter from working if they tab over.

patosullivan:
> Did this work without switching to mouseDown? Seems like the `event.preventDefault()` might do it by itself. This stops a keyboard enter from working if they tab over.

It should have worked with onClick but didn't. Not sure why. I figured it's ok since you can just use the enter key to send anyway.

PR #1757, opened 2023-01-20
chat: handle gboard pastes on android

patosullivan:
for #1679 
Fixes both pasting of refs and regular links from gboard in Android.

Google's gboard supports "pastes" while composing within an input. These pastes don't register because the user is actually in composition mode, it's as if the user just typed the content in, so a paste event never fires and therefore never gets caught by our existing paste rules.


arthyn (approved):
lgtm

PR #1759, opened 2023-01-20
groups: properly cleanup subs after leave/kick/etc

arthyn:
Not sure if we have an issue about this but getting kicked or having the group deleted from the host doesn't properly cleanup subscriptions.


tomholford (approved):
Pulled the branch and tested locally, LGTM

Issue #1763, opened 2023-01-21
groups: empty chat channel shows placeholders indefinitely


tinnus-napbus:
**Describe the bug, optionally with screenshots**

If you create a chat channel in a group & don't poast anything, it'll show those pulsing placeholder message block thingies indefinitely

**To Reproduce**
1. go to empty channel or create one
2. observe placeholder blocks don't go away

**Expected behavior**
if the channel is legitimately empty it shouldn't look like it's still loading, it should just be empty

**Additional context:**
- Ship type planet
- Instance type: self-hosted
- OS/browser: qutebrowser

tomholford:
Just ran into the same issue, this also applies for new DMs

![image](https://user-images.githubusercontent.com/16504501/213983022-ebec9787-3b0f-457b-b365-db0a3caa96c0.png)

PR #1769, opened 2023-01-23
chat: don't show placeholder for empty chat channels

patosullivan:
For #1763 

We need to check to see if we actually initialized chat state for the channel, empty channels always have zero keys.


arthyn (approved):
lgtm

PR #1772, opened 2023-01-23
groups, talk: add desk.ship

### LINKED ISSUE #1771: groups & talk: apps do not have a source on first boot
- Boot a new identity
- Observe that talk and groups have no upstream source
- Old ships also have their source set to their sponsor instead of the proper moon `sogryp-dister-dozzod-dozzod`
- I noticed this because users in UC help had their sponsor set to source for both of these (and as a result were not getting updates and were behind on kelvins)

@philipcmonk is working on a PR to fix this and we'll release it via `tloncorp/landscape`

philipcmonk:
These are required for a desk that's in the pill to have a sync source.

Fixes #1771

### DIFF
--- a/desk/desk.ship
+++ b/desk/desk.ship
@@ -1,0 +1,1 @@
+~sogryp-dister-dozzod-dozzod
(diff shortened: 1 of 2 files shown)


arthyn (approved):
lgtm

Issue #1765, opened 2023-01-21
groups: host should not be able to remove self from admin


zalberico:
- Be a group host
- Go to members
- Remove yourself from admin (I did this accidentally)
- Observe you can no longer do anything in the group

The host should always be an admin.

lukestiles:
In addition to not allowing this action, we should consider having a confirmation message for any important action like this.

PR #1774, opened 2023-01-23
groups: do not allow group host to be demoted from admin

patosullivan:
For #1765 

Prevent group host from being removed as an admin of a group, on both the frontend and backend.

Actual hoon changes are on lines 1285 and 1415. If I missed something obvious on the hoon, let me know.


patosullivan:
Need to fix something on the hoon side.

arthyn (approved):
lgtm

Issue #1754, opened 2023-01-20
groups: when you tag a user in chat that's not present in a group, UI should tell you


zalberico:
- Be in a group
- Tag a user in chat that's not in the group
- Observe nothing happen

We should do two things:
- Tell the user the ID they tagged is not in the group
- Ask the user if they want to invite that person to the group

Issue #1609, opened 2023-01-03
Interactive Elements Shifting


minderfolden:
<img width="739" alt="Screenshot 2023-01-02 at 20 46 48" src="https://user-images.githubusercontent.com/51895363/210290677-bcf9f475-a62b-4924-a88b-7541507e1ec0.png">
I go to click the "add attachment" plus sign.

<img width="739" alt="Screenshot 2023-01-02 at 20 47 24" src="https://user-images.githubusercontent.com/51895363/210290698-15cc1497-1da0-4cdc-b7a6-7891c0eb24e1.png">
But then the text formatting line disappears and the plus shifts down. Then I have to mouse down to click the plus again before the upload attachment flow starts. 

Same thing happens when the text input area is expanded but I then try to add an emoji react. I'll click the react element, and then the formatting area disappears and I have to click the emoji react element again.

**Additional context:**
- Ship type: planet
- Instance type: Digital Ocean droplet
- OS/browser: Stack Browser https://stackbrowser.com/

jamesacklin:
@minderfolden This is an interesting one—are you on a desktop browser with the mobile layout?

minderfolden:
Here's a screen recording: https://share.descript.com/view/UpxIBwDDt76 

It's on my desktop.

minderfolden:
Nope; not Realm. It's the Stack Browser and I have Grid open in one card, and Groups open in another card.

jamesacklin:
OK, interesting -- will repro and see what's up

arthyn:
hmm shouldn't be getting the mobile UI at the res, interesting

Issue #1600, opened 2022-12-30
Groups: keyboard shortcuts for text only last for one letter


nartesfasrum:
**Describe the bug, optionally with screenshots**
Keyboard shortcuts (ctrl+i, ctrl+b) for text formatting only work for one letter, then the text goes back to 'normal'.

**To Reproduce**
1. Ctrl+i.
2. Start typing.
3. Observe that your text, which was supposed to be italic, is only italic for one letter.

**Expected behavior**
The text should remain in the formatting I specify until I specify otherwise. 

**Additional context:**
- Ship type (comet/moon/planet/star/galaxy): planet, ~nartes-fasrum
- Instance type (Tlon-hosted/self-hosted with Port/CLI): selfhosted with CLI
- OS/browser (Safari on MacOS, Chrome on Windows, Firefox on Linux, etc.):  Brave | 1.38.115 Chromium: 101.0.4951.64 (Official Build) (64-bit) on Linux

Issue #1563, opened 2022-12-16
chat: appending an exclamation to tilde-prefixed text prevents post


jtobin:
**Describe the bug, optionally with screenshots**
Typing a patp or ~any-text-like-this and appending an exclamation mark to it will prevent 'return' from submitting a post.

**To Reproduce**
1. In a channel, type `~message!` in the input field.
2. Press return.

**Expected behavior**
The message '~message!' should immediately post.

**Additional context:**
- Brave on macOS

jamesacklin:
@jtobin Did you see this in Chat channels and DMs, or other places too?

jtobin:
IIRC it was in a chat channel, but I just confirmed that it occurs in DM's in Talk as well.

PR #1775, opened 2023-01-23
chat: omnibus #1

arthyn:
A bunch of really small things, mostly involving mentions and chat:
- #1500 
- #1563 
- #1600 
- #1405 
- #1609
- #1754


patosullivan (approved):
Looks great and mentions work as expected in both groups and talk when I run it locally.

PR #1780, opened 2023-01-24
desk: fix ship mark

tomholford:
Before this change, compilation was failing with:

```
[%error-validating /desk/ship]
[%validate-page-fail /desk/ship %from %mime]
/mar/ship/hoon:<[15 38].[15 55]>
/mar/ship/hoon:<[15 42].[15 55]>
```

Now it compiles successfully


jamesacklin:
@arthyn We need this for release, correct?

arthyn:
@jamesacklin I would think so, but I'm curious what changed. maybe it was never validated

arthyn (approved):
lgtm not sure where this code came from previously, but new stuff matches urbit/urbit base-dev

Issue #1523, opened 2022-12-15
chat: read-only channels have non-functioning message actions


thelifeandtimes:
**Describe the bug, optionally with screenshots**
When a chat restricts posters (I discovered this in the tlon public office hours channel), there is still a mouseover option to reply to a previous comment. The GUI pops up and you can type a response, but when hitting send it renders for a moment and then disappears (presumably because the backend permissions don't actually line up)

**To Reproduce**
1. Join Groups chat channel which has restricted permission
2. Click on the `#` symbol on mouseover to create a reply thread
3. Type in text to the chat box in the sidebar
4. See message render and then disappear

**Expected behavior**
Probably just removing the mouseover option to create a thread is the way to solve for this. alternatively, you could still allow that but in the threaded replies sidebar have the chat box removed like it is for the main chat channel.

**Additional context:**
- Ship type: planet
- Instance type: self hosted (locally on macOS, currently)
- OS/browser: Brave

patosullivan:
Threads now respect channel perms, but we still make it look like you can emoji react in these channels and that could be annoying for some people. Should be a simple fix.

PR #1781, opened 2023-01-24
chat: prevent emoji reactions in channels you can't write to

patosullivan:
for #1523 

Threads already respect channel perms, just making emojis do the same.


tomholford (approved):
LGTM

Issue #1596, opened 2022-12-29
Talk: Joining a new group from a reference in a chat channel takes you to a blank page


thelifeandtimes:
**Describe the bug, optionally with screenshots**
I was in a chat where someone shared a link to a new public group for which I was not yet a member, clicking `join` presumably gave me access to a group, but took me to a page with a blank main window and just my standard talk sidebar.

**To Reproduce**
1. Go to a group chat from within Talk
2. Click on the `join` button in a reference to a public group for which you are not a member

**Expected behavior**
A variety of things could happen here:
- automatically open a new browser tab with that group's main channel page
- take me to a chat channel within that group, likely the one with the most recent new messages?
- take me to a dialog box prompting me to open groups?

**Additional context:**
- Ship type: Planet
- Instance type: Self-Hosted on ubuntu
- OS/browser: Brave on MacOS

jamesacklin:
Similar to #1664, address at the same time.

arthyn:
and #1655

PR #1782, opened 2023-01-24
talk: proper navigation for group join actions

patosullivan:
for #1596 

When we're in talk we want the useGroupJoin actions to pop open a new tab with groups.

### DIFF
--- a/ui/src/groups/useGroupJoin.ts
+++ b/ui/src/groups/useGroupJoin.ts
@@ -6,6 +6,7 @@ import useHarkState from '@/state/hark';
 import { useGroup, useGroupState } from '@/state/groups';
 import { Gang, PrivacyType } from '@/types/groups';
 import { Status } from '@/logic/status';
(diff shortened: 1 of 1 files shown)


ui/src/groups/useGroupJoin.ts
```diff
@@ -70,14 +79,29 @@ export default function useGroupJoin(
...
         setStatus('error');
-        navigate(`/find/${flag}`);
-        if (requested) {
-          await useGroupState.getState().rescind(flag);
+        if (isTalk) {
+          if (requested) {
+            await useGroupState.getState().rescind(flag);
+          } else {
+            await useGroupState.getState().reject(flag);
+          }
```
arthyn, line 95: Could probably pull this one level above since it applies to both

arthyn (approved):
lgtm, one nit. this is a good solve until we can share+auto-join channels from talk

Issue #1621, opened 2023-01-04
groups: searching for a group on mobile doesn't let you paste a patp


zalberico:
- on iOS go to find groups
- attempt to paste patp
- observe you can't and also observe the field default text is not removed when the input field is selected

PR #1783, opened 2023-01-24
groups: disable autofocus for find groups on mobile

patosullivan:
for #1621 

autofocus was too grabby for mobile.


tomholford (approved):
LGTM

Issue #1655, opened 2023-01-09
groups: reference to msg in channel you're not a member of goes nowhere


tinnus-napbus:
**Describe the bug, optionally with screenshots**
If someone posts a reference to a message in a group you're a member of but in a channel you've not joined, it takes you to the group but doesn't open any channel, it's just empty

**To Reproduce**
1. follow link to msg in channel you've not joined
2. observe there's nothing

**Expected behavior**
it should probably prompt you to join the channel

jamesacklin:
We might make this implicit -- if you follow the reference, you join the channel. How does that sound @tinnus-napbus?

tinnus-napbus:
that makes sense for groups you're already a member of

PR #1784, opened 2023-01-24
groups: auto join channels in same group after clicking a reference

patosullivan:
For #1655 

Also 
- Eliminated the prop-drilling for the chat loading state (which I did, idk why) and cleaned up the ChatChannel component.
- We were using zustand's create/persist for heap and diary state and for some reason it wasn't actually persisting, this fixes that.
- We weren't actually scrying for perms when initializing heap, also fixed that.

~~I can't seem to figure out a way to actually load just-joined the channel content into the store without doing a page refresh. Any idea what I'm missing here? Try it out by commenting out the `window.location.reload()` in one of the useEffects.~~

### DIFF
--- a/ui/src/state/diary/diary.ts
+++ b/ui/src/state/diary/diary.ts
@@ -1,8 +1,5 @@
-import bigInt, { BigInteger } from 'big-integer';
+import bigInt from 'big-integer';
 import { unstable_batchedUpdates as batchUpdates } from 'react-dom';
-import _ from 'lodash';
-import create from 'zustand';
-import { persist } from 'zustand/middleware';
 import produce, { setAutoFreeze } from 'immer';
 import { BigIntOrderedMap, decToUd, udToDec, unixToDa } from '@urbit/api';
 import { useCallback, useEffect, useMemo } from 'react';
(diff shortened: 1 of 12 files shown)


patosullivan:
Just pushed a fix for the content-not-loading-after-join issue, using a local "joining" state for each channel and subscription tracking. Also:

- We were using zustand's create/persist for heap and diary state and for some reason it wasn't actually persisting, this fixes that. 
- We weren't actually scrying for perms when initializing heap, also fixed that.

ui/src/state/diary/diary.ts
```diff
@@ -74,314 +65,319 @@ function diaryNoteDiff(flag: DiaryFlag, time: string, delta: NoteDelta) {
...
+
+      await makeNotesStore(
+        flag,
+        get,
+        `/diary/${flag}/notes`,
+        `/diary/${flag}/ui`
+      ).initialize();
+    },
+  }),
+  ['briefs', 'shelf', 'notes'],
```
arthyn, line 379: A little concerned that notes is going to balloon our local storage usage even more. One thing that would be nice is if we could use the partialize function from zustand here to only pull outlines instead of full notes.

arthyn (approved):
lgtm! a few nits and we should probably edit base state to allow for zustand's partialize whether here or later.

ui/src/state/diary/diary.ts
patosullivan: You're right, will make a separate issue for this.

PR #1785, opened 2023-01-25
perms: secret group fixes

### LINKED ISSUE #1469: groups: "secret group" reference element "flashes" or renders in an unexpected manner after being invited to it
I had received an invite to a secret group and checked the chat where the original link was posted, only to view a highly buggy reference element in the chat:

https://user-images.githubusercontent.com/1195363/207427273-89ec1adb-fc80-4dc5-b819-9807eb1b7ca2.mov

**To Reproduce**
1. Ask someone to post a reference to a secret group in a channel you are in
2. It shouldn't appear at first
3. Ask the same individual to invite you to the group
4. Stay in the original chat, and see if you witness the "nonexistent" reference turn into the group invite
5. If the invite appears as buggy as it does in my video, you've replicated the issue

**Expected behavior**
There are two issues with the behavior I witnessed:
1. The obvious flashing shouldn't be happening, period
2. The group link appearing as an invite displays the group's privacy status as "public", which is incorrect

***Additional context:**
- Ship: planet (~tonlur-hidlyn)
- Instance type: Self-hosted (Digital Ocean)

Accessing the above urbit through the following machine, remotely:
- MacOS Ventura — 13.0.1 (22A400)
- Safari — Version 16.1 (18614.2.9.1.12)

tomholford:
This PR addresses a couple secret groups issues and resolves #1469 

- [x] show secret group reference when a member of the group (before, it would be hidden even if already a member)
- [ ] receiving an invite to a secret group causes a render loop for the reference (#1469)

### DIFF
--- a/ui/src/components/References/GroupReference.tsx
+++ b/ui/src/components/References/GroupReference.tsx
@@ -38,7 +38,10 @@ export default function GroupReference({
     }
   }, [gang, group, flag, isScrolling]);
 
-  if (privacy === 'secret') {
+  const referenceUnavailable =
+    privacy === 'secret' && !(window.our in (group?.fleet ?? {}));
+
+  if (referenceUnavailable) {
     return (
       <div className="relative flex h-16 items-center space-x-3 rounded-lg border-2 border-gray-50 bg-gray-50 p-2 text-base font-semibold text-gray-600">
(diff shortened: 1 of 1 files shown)


patosullivan (approved):
Nice!

PR #1794, opened 2023-01-26
ux: clearer Find Groups copy when Not Found

### LINKED ISSUE #1666: groups: improve Find Groups feedback for offline hosts
When searching for a host on Find Groups, if the ship is offline, an error message is rendered: "This ship doesn't host any groups"

This is causing confusion for some users - they know that a certain ship does in fact host a group. So, we should consider rephrasing this message to something like:

"This ship is either offline or does not host any groups"

tomholford:
When the search times out, it could be due to multiple root causes.

This clarifies the message as it has caused confusion in UC Help a few times now:

image

Resolves #1666.

# Preview

## Before
image

## After
image

### DIFF
--- a/ui/src/groups/FindGroups.tsx
+++ b/ui/src/groups/FindGroups.tsx
@@ -192,14 +192,10 @@ export default function FindGroups({ title }: ViewProps) {
       }
 
       return (
-        <div>
-          The search timed out. This happens when:
-          <ul className="list-inside list-disc">
-            <li>The ship doesn&apos;t host any groups</li>
-            <li>The ship is under heavy load</li>
-            <li>The ship is offline</li>
-          </ul>
-        </div>
+        <span>
+          Your search timed out, which may happen when a ship hosts no groups,
+          is under heavy load, or is offline.
+        </span>
       );
     }
 
--- a/ui/src/groups/FindGroups.tsx
+++ b/ui/src/groups/FindGroups.tsx
@@ -191,7 +191,16 @@ export default function FindGroups({ title }: ViewProps) {
         );
       }
 
-      return <span>This ship doesn&apos;t host any groups</span>;
+      return (
+        <div>
+          The search timed out. This happens when:
+          <ul className="list-inside list-disc">
+            <li>The ship doesn&apos;t host any groups</li>
+            <li>The ship is under heavy load</li>
+            <li>The ship is offline</li>
+          </ul>
+        </div>
+      );
     }
 
     return null;


tomholford:
@e-shee Any thoughts on the updated copy? 

@urcades Any thoughts on the aesthetique?

urcades:
Any suggestion for improvement I'd have is dependent on copy recommendations from @e-shee , but my gut instinct is to render this as a single line of discretionary text. I sketched this out -really- quickly here:

Screenshot 2023-01-26 at 12 12 07 AM

https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/branch/yonw9FxqhTxghFHdjHkLBH/Product-Handoffs-(Locked)?node-id=2638%3A95369&t=Y7zMQLvWKMEbX23q-1

Ideally, if we had robust FAQ/Support capability, I'd also recommend a "Learn more" link, for people who are curious about the edges of p2p networking and its implications on product design.

Lastly, if you look to the frames below the linked one above, there are some additional search states that might be worth considering that we have yet to implement — in some cases (like a search that may have been mis-spelled, but comes close to a match in your network), I think the best outcome would be to find -something- we could recommend to people.

Screenshot 2023-01-26 at 12 12 25 AM

Screenshot 2023-01-26 at 12 12 22 AM

Hopefully this is helpful!

tomholford:
> Any suggestion for improvement I'd have is dependent on copy recommendations from @e-shee , but my gut instinct is to render this as a single line of discretionary text. I sketched this out -really- quickly here:
> 
> Screenshot 2023-01-26 at 12 12 07 AM
> 
> https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/branch/yonw9FxqhTxghFHdjHkLBH/Product-Handoffs-(Locked)?node-id=2638%3A95369&t=Y7zMQLvWKMEbX23q-1
> 
> Ideally, if we had robust FAQ/Support capability, I'd also recommend a "Learn more" link, for people who are curious about the edges of p2p networking and its implications on product design.
> 
> Lastly, if you look to the frames below the linked one above, there are some additional search states that might be worth considering that we have yet to implement — in some cases (like a search that may have been mis-spelled, but comes close to a match in your network), I think the best outcome would be to find -something- we could recommend to people.
> 
> Screenshot 2023-01-26 at 12 12 25 AM
> 
> Screenshot 2023-01-26 at 12 12 22 AM
> 
> Hopefully this is helpful!

@urcades Thanks for the prompt review and design. Just made the change and updated the screenshot above. I love both ideas for the FAQ link and the enhanced discoverability. Will create new follow up issues to track both.

patosullivan (approved):
Great idea.

PR #1795, opened 2023-01-26
groups: only show unread dot in Groups Sidebar for comms events

### LINKED ISSUE #1649: groups: update "unread group dot" in sidebar to map to "new communication events" only
Lifting a comment out of #1626 for @patosullivan —

> Alternatively, we could only show the blue notification dot on groups for unread messages (and not these other types of notifications).

I think this is the behavior we should support — somewhere along the lines we managed to correlate the "unread group dot" with its meta-activity, when the dot has always been designed to account for "unread social comms within channels".

_Originally posted by @urcades in https://github.com/tloncorp/landscape-apps/pull/1626#pullrequestreview-1239019972_

tomholford:
Before this change, other group events would also trigger the dot.

For example, if a user joined or left a channel, it would show the Group as unread.

This resolves #1649

### DIFF
--- a/ui/src/logic/useIsGroupUnread.ts
+++ b/ui/src/logic/useIsGroupUnread.ts
@@ -1,6 +1,6 @@
+import { useCallback } from 'react';
 import { useNotifications } from '@/notifications/useNotifications';
 import { useGroups } from '@/state/groups';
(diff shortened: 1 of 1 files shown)


patosullivan (approved):
Nice

PR #1803, opened 2023-01-26
chat: removing time correction

### LINKED ISSUE #1494: chat: sender messages out of order
It seems like the senders messages can get out of order with respect to incoming messages:

Screen Shot 2022-12-14 at 3 49 34 PM

arthyn:
A lot of waffling here, but ultimately I think the client ordering we have/had was fine, but trying to correct the sender's time with the server time was a one-sided solution that we ultimately didn't need and was causing ordering issues. It was originally done when taking a pass at double renders, but I think it was unnecessary. Fixes #1494 also may fix the last of the double renders?

Also threw in some corrections for recent channel navigation stuff to make sure it respects `canRead` when searching for which channel to switch to and made the navigations absolute paths so that we don't end up on a bad path.

There are still cases where messages can end up out of order, such as a ship being down (and then coming back up) as well as heavy load or possibly an OTA.


tomholford (approved):
Pulled and tested locally, LGTM

PR #1804, opened 2023-01-26
chat: correctly rectify threads when deleting messages

### LINKED ISSUE #1625: threads: deleted reply does not return back to original state
A clear and concise description of what the bug is. If applicable, add screenshots to help explain your problem.

**To Reproduce**
I started the thread and I wrote a reply and then deleted it. This is the result:
image

**Expected behavior**
The thread should be deleted and return the user with thread sidebar closed (back to normal chat)

arthyn:
OTT this fixes #1625


tomholford (approved):
Good catch. Tested locally, LGTM

Issue #1157, opened 2022-11-15
groups: Switching windows while entering text into invite input field empties field


jtobin:
**Describe the bug, optionally with screenshots**
When inviting a ship to join a group, if one cmd+tabs away to another window and back, any entered text is erased.

**Expected behavior**
Any text already entered should be preserved.

**Additional context:**
- Brave on macOS.

PR #1806, opened 2023-01-26
keep input text on blur in ship selector

patosullivan:
for #1157 

Had a lot of trouble getting onInputChange to do what we wanted here before I realized that we could pass inputValue as a prop to the CreatableSelect component itself (and thus have onInputChange fire at appropriate times). Passing this prop doesn't seem to change any other functionality, but please test and let me know if you find something.


tomholford (approved):
Nice work. Tested locally, LGTM

Issue #1812, opened 2023-01-27
groups: channel list sorting is broken


patosullivan:
Users cannot change the sortmode on channel lists.

PR #1813, opened 2023-01-27
groups: fix sidebar sort persistence issues

patosullivan:
For #1812 

Both channel list sorting and group sorting settings were not persisting, it looks like we just weren't passing the mode into the sort function.

I also discovered that we were wiping the groupSideBarSort key each time we saved individual group channel sort options, so you could only have one group's sidebar sort settings saved at a time. This also fixes that.


arthyn (approved):
lgtm

Issue #1366, opened 2022-12-09
groups: no loading indicator for inviting a user


zalberico:
**Describe the bug, optionally with screenshots**
- Invite a user to a group
- Observe the UI tell you nothing about success or failure

arthyn:
I don't think this is a bug as much as an enhancement. We currently don't track these unless the group is private/secret.

zalberico:
I'd take an "invite sent"

PR #1816, opened 2023-01-27
groups: add loading spinner for group invite dialog and pending screen

patosullivan:
for #1366 

adds subscription tracking and relevant UI changes for `addMembers`, `invite` and `revoke` (which is both "cancel" and "reject" in secret/private groups)  in both the "Members" and "Pinned" views.

demo:

https://user-images.githubusercontent.com/1221094/215186214-bd68a788-c404-457e-af6b-9b99ac68ff79.mov

### DIFF
--- a/ui/src/state/groups/groups.ts
+++ b/ui/src/state/groups/groups.ts
@@ -315,8 +315,24 @@ export const useGroupState = create<GroupState>(
               await useSubscriptionState
                 .getState()
                 .track('groups/groups/ui', (event) => {
-                  const { diff } = event;
-                  if ('cordon' in diff) {
+                  const { update, diff } = event;
+                  if (update && update.diff) {
+                    if ('cordon' in update.diff) {
+                      const { shut } = update.diff.cordon;
+                      if ('add-ships' in shut) {
+                        const { kind, ships: addedShips } = shut['add-ships'];
+                        return (
+                          kind === 'pending' &&
+                          addedShips.every((ship: string) =>
+                            ships.includes(ship)
+                          )
+                        );
+                      }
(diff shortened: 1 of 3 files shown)


ui/src/state/groups/groups.ts
```diff
@@ -318,18 +357,60 @@ export const useGroupState = create<GroupState>(
...
+                      return (
+                        returnedKind === 'pending' &&
+                        addedShips.every((ship: string) => ships.includes(ship))
+                      );
+                    }
+                    return false;
+                  }
+                  return false;
+                });
+
```
arthyn, line 409: Is this duplicated or am I going crazy lol

ui/src/state/groups/groups.ts
```diff
@@ -298,18 +298,57 @@ export const useGroupState = create<GroupState>(
...
+                      const { kind, ships: addedShips } = shut['add-ships'];
+                      return (
+                        kind === 'pending' &&
+                        addedShips.every((ship: string) => ships.includes(ship))
+                      );
+                    }
+                    return false;
+                  }
+                  return false;
+                });
```
arthyn, line 347: same here

arthyn (approved):
now that we have this pattern going we should abstract it into something I think. one question, but maybe it's because we get different types on the same subscription?

patosullivan:
Yeah, we get different types on the same subscription. One with just diff, the other with update.diff right afterward. Without checking for both it'll fail on the second one. Not sure why, very annoying.

patosullivan:
Totally agree, we need to extract the tracking stuff.

Issue #1811, opened 2023-01-27
heap: unable to refresh curio detail view


tomholford:
This is happening on master (unreleased); perhaps related to the heap state changes in #1784

![image](https://user-images.githubusercontent.com/16504501/215040141-69ceced7-a1ca-420f-a5f3-4f014d9fd499.png)

1. Navigate to a curio detail view
2. Refresh the page

patosullivan:
Same thing is happening with diaries. Looking into it.

PR #1819, opened 2023-01-27
groups: fix for note and curio detail not loading on refresh

patosullivan:
For #1811 

It looks like switching out createState for create probably caused notes/curios to sometimes not return as a proper BigIntOrderedMap. I'm not totally sure why yet, but checking for the existence of the `.get` method stops the bleeding for now.


tomholford (approved):
LGTM

Issue #1529, opened 2022-12-15
groups + talk: layout gets stuck in responsive-small mode


zalberico:
**Describe the bug, optionally with screenshots**
- Be in Talk on desktop
- Make the window narrow until it switches to the mobile view
- Observe making the window wide won't switch back without a hard refresh

jamesacklin:
A bigger issue -- our media query listeners aren't continuously listening to the window size. @arthyn suspected he fixed but this also happens with theming.

We should look at which of these layout bugs occur as a result of Tailwind class names and which come from useIsMobile hook (which aren't as performant).

PR #1820, opened 2023-01-27
clubs: only accept once

### LINKED ISSUE #1565: talk: need to repeatedly accept group dm `invite`
image

`%talk` hash `0vc.9kn8n.68vo6.v4e7g.4mou1.ekqeg.65ttj.avtlg.02t5l.dts7e.qjfk1` still asking me to accept same group DM roughly every time I open it. This was resolved for a day or so and then re-occurred.

arthyn:
This fixes #1565 by ensuring that we hear facts about club's membership and meta even if we haven't visited said club and also by loading all clubs initially so that we can reconcile with whatever is stored in local storage.

This does change our subscriptions so this needs to be communicated with external parties.

Also threw in a fix for the issue of channels on refresh not loading, some console errors about nested buttons with chat reactions, and the issue with media query listeners disappearing #1529.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -321,7 +321,7 @@
   ^+  cor
   ?+    pole  ~|(bad-watch-path/path !!)
       [%imp ~]        ?>(from-self cor)
-      [%club %new ~]  ?>(from-self cor)
+      [%clubs %ui ~]  ?>(from-self cor)
       [%briefs ~]  ?>(from-self cor)
       [%ui ~]  ?>(from-self cor)
       [%dm %invited ~]  ?>(from-self cor)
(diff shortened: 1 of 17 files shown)


patosullivan (approved):
Looks great, works well locally.

tomholford:
>  some console errors about nested buttons with chat reactions

thank you! this was rather obnoxious

tomholford (approved):
This is great! Addresses some annoyances, and unifies the implementation and is thus more maintainable. LGTM

jamesacklin:
Very nice. Do we need state migrations for this?

arthyn:
no state migrations needed, but at the very least @alecananian will likely need to change his subscription for club meta updates

Issue #1733, opened 2023-01-18
Failed to fetch dynamically imported module: https://[redacted]/apps/escape/assets/DiaryAddNote.eeefdfc8.js


belisarius222:
```
TypeError: Failed to fetch dynamically imported module: https://[redacted]/apps/escape/assets/DiaryAddNote.eeefdfc8.js
```

This happens each time I click the "Add Note" button in the `halbex-palheb/uf-public/channels/diary/~halbex-palheb/infrastructure-6618` channel of the UF Public group.  The error also shows up in a modal dialog.

patosullivan:
It looks like this is happening in Escape. Can you try the same thing in Groups?

tomholford:
Smoothbrain: a dedicated `catch` clause for this specific error that triggers `location.reload`

Ideal: a healthcheck / hash endpoint that is polled periodically; when a new version is deployed / frontend is stale, show a butter bar notification to reload the app

PR #1829, opened 2023-01-30
support improvement omnibus

patosullivan:
This is for #1733, but also includes some other nice features.

I added a kiln state (required updated @urbit/api) that we can use to get info about desks the user has installed (same as in %garden/landscape). For now, we can use this to surface the base hash of the app that the user is currently looking at, but we could use it in the future to check for the existence of other apps that we could interoperate with.

This PR adds:

1) A way to poll the backend to find out if we have an update:

https://user-images.githubusercontent.com/1221094/215599321-af9a025d-1d23-4e54-b3d1-b8709d872487.mov

2) An "About Groups" (or Talk) modal that shows basic information:

https://user-images.githubusercontent.com/1221094/215598995-472cd514-4f69-4776-b730-ce0c26ba60ec.mov

### DIFF
--- a/ui/src/components/AboutDialog.tsx
+++ b/ui/src/components/AboutDialog.tsx
@@ -43,8 +43,10 @@ export default function AboutDialog() {
           </span>
           <span className="text-sm text-gray-500">{appDescription}</span>
           <div className="flex flex-row items-center text-sm">
-            <span className="font-bold text-gray-500">Hash:</span>
-            <code className="ml-2 break-all">{pike?.hash}</code>
+            <div className="flex flex-row items-end">
+              <span className="font-bold text-gray-500">Hash:</span>
+              <code className="ml-2 break-all">{pike?.hash}</code>
+            </div>
             <IconButton
               label="Copy Hash"
               icon={
--- a/ui/src/components/AboutDialog.tsx
+++ b/ui/src/components/AboutDialog.tsx
@@ -1,5 +1,6 @@
 import { useDismissNavigate } from '@/logic/routing';
 import { isTalk, useCopy } from '@/logic/utils';
+import { useCharge } from '@/state/docket';
 import { usePike } from '@/state/kiln';
 import Dialog, { DialogContent } from './Dialog';
 import IconButton from './IconButton';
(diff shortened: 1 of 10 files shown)


jamesacklin (changes requested):
Looks good, can you set the About Groups text color to the customary gray?

patosullivan:
> Looks good, can you set the About Groups text color to the customary gray?

@jamesacklin sure, in the modal itself or in the dropdown?

jamesacklin:
In the dropdown.

One more nit: please align the baselines of the hash and the label text beside it.

Screenshot 2023-01-30 at 5 10 44 PM

ui/src/components/AboutDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+  );
+
+  const onOpenChange = (open: boolean) => {
+    if (!open) {
+      dismiss();
+    }
+  };
+
+  const appDescription = isTalk
+    ? `Send encrypted direct messages to one or many friends.
```
arthyn, line 26: I think if you wanted to we could pull this from the treaty instead of hardcoding

ui/src/components/Sidebar/Sidebar.tsx
```diff
@@ -64,6 +66,15 @@ export function GroupsAppMenu() {
                 Submit Feedback
               </DropdownMenu.Item>
             </a>
+            <Link
+              to="/about"
+              className="no-underline"
+              state={{ backgroundLocation: location }}
+            >
+              <DropdownMenu.Item className="dropdown-item pl-3 text-blue">
```
arthyn, line 74: in the video it looked like this text was center aligned, but it should be left aligned

ui/src/state/bootstrap.ts
```diff
@@ -6,13 +6,16 @@ import { useDiaryState } from './diary';
...
+import useKilnState from './kiln';
 import { useSettingsState } from './settings';
 import { useStorage } from './storage';
 
-export default function bootstrap() {
+export default async function bootstrap() {
   useGroupState.getState().start();
   useChatState.getState().start();
 
+  useKilnState.getState().initializeKiln();
```
arthyn, line 17: I would call this last it's pretty low priority

arthyn (changes requested):
this is awesome and gives us a good place to potentially put release notes if we wanted. just a few nits

ui/src/components/AboutDialog.tsx
patosullivan: Yeah, this is a great idea.

ui/src/components/Sidebar/Sidebar.tsx
patosullivan: The other dropdown item is center aligned, though.

arthyn: weird I didn't think we center aligned anything lol

arthyn (approved):
lgtm!

PR #1832, opened 2023-01-30
Fix navigation on lack of read perms on channels

patosullivan:
I don't know if we have an issue for this, but users were being redirected to /groupflag/actvity (regardless of app) if they didn't have read permissions on a channel.

I had thought we already fixed this, but apparently we didn't.

### DIFF
--- a/ui/src/chat/ChatChannel.tsx
+++ b/ui/src/chat/ChatChannel.tsx
@@ -19,6 +19,7 @@ import {
   canReadChannel,
   canWriteChannel,
   isChannelJoined,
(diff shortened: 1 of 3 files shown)


patosullivan:
> I had this in my branch probably why you thought that

@arthyn which branch?

arthyn:
#1820

patosullivan:
@arthyn I'm not seeing it there

arthyn (approved):
I had changed the urls so they weren't relative which was the issue, but you took it a step further here. lgtm

Issue #1430, opened 2022-12-12
chat: visual duplication of some chat messages


litmus-ritten:
Since Landscape 1, I have noticed that some chat messages are temporarily visually duplicated on post. This only happens to my own messages, survives navigation to different contexts within Groups, but the duplicate message disappears on refresh. This affects both textual and graphical chat messages.

**Graphical example**

![urbit - landscape 1 duplicate message bug](https://user-images.githubusercontent.com/62781197/206976657-87f73c62-a72f-4e4a-80fa-60d0d24b9366.png)

**To Reproduce**
1. Chat in groups.
2. Observe that sometimes a message is duplicated on post.

**Expected behaviour**
Chat messages should only show up once per message.

**Additional context:**
- Ship type: planet
- Instance type: self-hosted with CLI
- OS/browser: Vivaldi on Linux

arthyn:
Possibly fixed with #1803, but can't reliably repro

litmus-ritten:
I'm still seeing this. Could reproducibility be a function of pier speed and/or network speed?

arthyn:
that pr hasn't been deployed yet sorry

PR #1838, opened 2023-01-31
chat: dont add messages already indexed

arthyn:
OTT this should be the final fix for #1430 our white whale


patosullivan (approved):
Great!

Issue #1787, opened 2023-01-25
chat: date dividers not rendered until unreads dismissed


tomholford:
This happens for any chat channel with unreads. When first loaded, no date dividers are rendered. Then, when the notifications are dismissed (either by the timer or by clicking the X), they are then subsequently rendered.

https://user-images.githubusercontent.com/16504501/214528164-6e74a3dd-dc37-4fb7-bc52-823ab6c6e4be.mp4

PR #1842, opened 2023-01-31
chat: do not hide date dividers while chat is unread

patosullivan:
For #1787


tomholford (approved):
Love the ol' one-liner. This also removes an unnecessary re-render :heart: LGTM

jamesacklin (approved):

PR #1850, opened 2023-02-01
chat: do not allow send on submit if send is disabled on chatinput

patosullivan:
This was a cause for the "broken images" appearing in chat. Users could just hit enter on the input and it would send the non-uploaded image url.


jamesacklin (approved):
Yes

PR #1855, opened 2023-02-02
heap: prevent double comments on Curios

### LINKED ISSUE #1576: gallery: Double comment
If I hit enter quickly enough, I can 

**To Reproduce**
1. Load a post in Collections
2. Write comment
3. Hit enter really fast, twice
4. /1/chan/heap/~dabben-larbet/interface-1720/curio/170141184505978111400233779297014251520

**Expected behavior**
There should be one instance of any comment

**Additional context:**
- Ship type (comet/moon/planet/star/galaxy)
- Instance type (Tlon-hosted/self-hosted with Port/CLI)
- OS/browser (Safari on MacOS, Chrome on Windows, Firefox on Linux, etc.)

tomholford:
This resolves #1576 by clearing the editor input before poking the curio. Previously, it would be cleared after submission, which would allowed a brief window to double post by spamming Enter (before the component could process the state update). I tried other approaches (e.g., do not send if isPending), but they were causing weird side effects or extraneous renders.

Also, add keys to mapped comments in HeapDetail view to suppress console error.

https://user-images.githubusercontent.com/16504501/216317586-6e287363-b332-4b58-94cb-1838c818c58a.mov

### DIFF
--- a/ui/src/heap/HeapDetail/HeapDetailSidebar/HeapDetailComments.tsx
+++ b/ui/src/heap/HeapDetail/HeapDetailSidebar/HeapDetailComments.tsx
@@ -32,8 +32,8 @@ export default function HeapDetailComments({ time }: HeapDetailCommentsProps) {
       >
         {Array.from(comments)
           .sort(([a], [b]) => a.compare(b))
-          .map(([, curio]) => (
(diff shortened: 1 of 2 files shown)


patosullivan (approved):
lgtm

PR #1859, opened 2023-02-02
ops: replace glob with new flow

arthyn:
OTT, should fix issues with capital letter containing imports like fonts and notebook editor


patosullivan (approved):
LFG

PR #1860, opened 2023-02-02
groups: allow group joins from references

### LINKED ISSUE #1841: groups: users cannot join groups from any refs
If a user clicks "continue reading" on a notebook reference for a group the user hasn't joined yet the user does not auto-join the group, instead they navigate to a blank page that looks like this:

image

patosullivan:
Fixes #1841 

Using the group join modal and some URL parameters to give users the option to join a (public, otherwise they wouldn't be able to see the ref anyway) group from a ref and then automatically navigating the user to the content from the ref after the join completes.

Demo video (excuse the confusion after I click through the first ref, I used the browser back button instead of the app nav):

https://user-images.githubusercontent.com/1221094/216381861-8ca124cc-ae9d-4a18-865c-ba9ea03304f9.mov

### DIFF
--- a/ui/src/chat/ChatChannel.tsx
+++ b/ui/src/chat/ChatChannel.tsx
@@ -26,7 +26,6 @@ import ChatScrollerPlaceholder from '@/chat/ChatScoller/ChatScrollerPlaceholder'
 
 function ChatChannel({ title }: ViewProps) {
   const navigate = useNavigate();
-  const joinRef = new URLSearchParams(window.location.search).get('joinref');
   const { chShip, chName } = useParams();
   const chFlag = `${chShip}/${chName}`;
   const nest = `chat/${chFlag}`;
@@ -51,26 +50,18 @@ function ChatChannel({ title }: ViewProps) {
 
   const joinChannel = useCallback(async () => {
     setJoining(true);
-    if (joinRef) {
-      setTimeout(async () => {
-        await useChatState.getState().joinChat(chFlag);
-      }, 300);
-    } else {
+    try {
       await useChatState.getState().joinChat(chFlag);
+    } catch (e) {
+      console.log("Couldn't join chat (maybe already joined)", e);
     }
     setJoining(false);
-  }, [chFlag, joinRef]);
+  }, [chFlag]);
 
   const initializeChannel = useCallback(async () => {
-    if (joinRef) {
-      setTimeout(async () => {
-        await useChatState.getState().initialize(chFlag);
-      }, 300);
-    } else {
-      await useChatState.getState().initialize(chFlag);
-    }
+    await useChatState.getState().initialize(chFlag);
     setLoading(false);
-  }, [chFlag, joinRef]);
+  }, [chFlag]);
 
   useEffect(() => {
     if (!joined) {
@@ -80,7 +71,7 @@ function ChatChannel({ title }: ViewProps) {
 
   useEffect(() => {
     setLoading(true);
-    if (joined && channel && canRead && !joining) {
+    if (joined && canRead && !joining) {
       initializeChannel();
       setRecentChannel(nest);
     }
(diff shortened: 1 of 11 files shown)


ui/src/chat/ChatChannel.tsx
```diff
@@ -26,6 +26,7 @@ import ChatScrollerPlaceholder from '@/chat/ChatScoller/ChatScrollerPlaceholder'
 
 function ChatChannel({ title }: ViewProps) {
   const navigate = useNavigate();
+  const joinRef = new URLSearchParams(window.location.search).get('joinref');
```
tomholford, line 29: can this be accomplished with `useSearchParams`?

https://reactrouter.com/zh/main/hooks/use-search-params

ui/src/chat/ChatChannel.tsx
```diff
@@ -50,14 +51,26 @@ function ChatChannel({ title }: ViewProps) {
 
   const joinChannel = useCallback(async () => {
     setJoining(true);
-    await useChatState.getState().joinChat(chFlag);
+    if (joinRef) {
+      setTimeout(async () => {
+        await useChatState.getState().joinChat(chFlag);
+      }, 300);
```
tomholford, line 57: Is the 300 to allow for the modal to open? Will this be sufficient for lower end devices? 

Also, please extract to a constant value so we can easily tune it here and on line 68

patosullivan: Actually we're waiting for the join to complete (on the new group/channel you're joining). For some reason we get the fact back that the group join completes (and the promise on the frontend returns true), but if we don't wait a bit before initializing the chat it'll fail. I had to do the same thing with notes. We need to figure out why that's happening on the backend, this is just a workaround for now.

ui/src/chat/ChatChannel.tsx
patosullivan: What will this get us in this case?

ui/src/heap/HeapDetail.tsx
```diff
@@ -22,6 +22,7 @@ export default function HeapDetail() {
   const nest = useNest();
   const [, chFlag] = nestToFlag(nest);
   const { time, curio } = useCurioFromParams();
+  const [loading, setLoading] = useState(false);
```
arthyn, line 25: we should use this in a spinner

ui/src/groups/useGroupJoin.ts
```diff
@@ -79,7 +82,25 @@ export default function useGroupJoin(
...
+              );
+            }
+          } else if (isTalk) {
+            const href = `/apps/groups/groups/${flag}/channels/${redirectItem.nest}/${redirectItem.type}/${redirectItem.id}&joinref=true`;
+            window.open(`${window.location.origin}${href}`, '_blank');
+          } else {
+            navigate(
+              `/groups/${flag}/channels/${redirectItem.nest}/${redirectItem.type}/${redirectItem.id}?joinref=true`
+            );
+          }
```
arthyn, line 102: this could be cleaned up a bit since these conditionals and paths are pretty similar

ui/src/chat/ChatChannel.tsx
arthyn: I thought we had figured this out, but having trouble remembering.

arthyn (changes requested):
I think it's time for us to make a navigate helper/util that does the if/else with navigate and window.open. Also let's dig into the setTimeout thing. I thought it had something to do with us not being sure about when the channel was available or beginning to sync or maybe having the group in perms? idk but let's pair on it

ui/src/chat/ChatChannel.tsx
tomholford: Consider this a nit, it feels more idiomatic but is certainly not a merge-blocker. If it doesn't get us anything then skip it

ui/src/chat/ChatChannel.tsx
patosullivan: That was a different instance, but I'm down to figure this out.

patosullivan:
Added a useNavigateByApp hook. Still need to figure out why setTimeout seems to be necessary for initializing in this instance.

patosullivan:
Should be good to go now.

arthyn (approved):
lgtm!

PR #1876, opened 2023-02-03
groups: notification overhaul

### LINKED ISSUE #1691: groups: notifications updates
Per the [project specification](https://docs.google.com/document/d/1z_qpCWGKv5Muio9P1-uxGnEVOr9_jZ3vQPwsSV10DME/edit?usp=sharing) written out by @jamesacklin, I've arranged a few updates to the notifications interface(s) across landscape and groups.

In this issue, I'll go over **Groups**-specific updates, which amount to:

1. Updates to the interface that contains App-level (all group) notifications, to make notifications a little easier to filter/parse through
2. A proposed per-group "Home" interface which can surface per-group notifications in a sort of "group home feed". This interface can then serve as the basis for a proper customizable group "Home"

(For changes to Landscape's notifications experience, and a description of the changes to notification rows themselves, refer to #1689)

To begin, this issue is proposing that we update the Groups app notifications experience from this:

Screenshot 2023-01-12 at 11 23 00 AM

To this:

GROUP

Each of the changes proposed in #1689 apply to this interface as well, which is scoped only to per-Group notifications. This is reflected in the "Notification Provenance" data label, which does not depict the **app** from which a notification hails, but the group/channel:

Screenshot 2023-01-12 at 11 29 13 AM

~

Next, the newest addition to the Notifications interfaces is a proposed new navigation spot within an individual group — the Group Home:

Header

Each of the proposed experiential updates hold true, and note that each of the notifications, being scoped to a particular group, also specify their provenance one level further, and omit the mention of the notification coming from the "Group itself", since it's obvious one is in a particular group while viewing the group's home screen. Instead, the particular channels or meta-group functions are displayed as icons:

Screenshot 2023-01-12 at 12 06 10 PM

~

These are the high-level notes for initial implementation, more specific design/state details can be found in [Figma](https://www.figma.com/file/luXjuX8fdV2mQrfqCQ4Cr2/2023-Q1-Groups?node-id=2380%3A7835&t=lxBt8HrEvYmnDr8C-1)

### LINKED ISSUE #1827: groups: while using a mobile device, tapping notifications yields no redirect
While using the Groups webapp on a mobile device, attempting to tap into a notification does not redirect me to its trigger:

https://user-images.githubusercontent.com/1195363/215558767-c4676bef-1eab-414f-9777-c6c2a432ed82.MP4

It's hard to see in this video, but I have a series of blue notifications that I'm tapping one at a time to see where they lead me. None of them appear to function correctly. The experience I'm witnessing on mobile essentially renders the notifications interface completely unusable.

patosullivan:
Fixes #1691
Fixes #1827 

Implements ed's designs for notifications in groups. 

Still need to figure out:
- [x] why sawRope is causing old mentions to disappear from the notifications list all together.
- [x] what, if anything, needs to change about the mobile implementation
- [x] what, if anything, needs to change about the dropdown menu on notifications.

Screen recording:
https://user-images.githubusercontent.com/1221094/216711277-b27eab96-59ad-49b8-b62c-1689460e2913.mov

Also note that that recording doesn't include `mix-blend-multiply` on the blue-on-blue mentions, but here is how it looks when applied:
image

Looks like this on mobile:

All notifications:
localhost_3000_apps_groups_(Pixel 5)

Group home:
localhost_3000_apps_groups_groups__dotnet-dotnet-finned-palmer_fake-group(Pixel 5)


patosullivan:
One other thing to note here: the sigils aren't detailed, but this is how they appear in garden as well. Should they be?

patosullivan:
Accounting for channel adds and group membership changes:

image

patosullivan:
Now with detailed sigils: 
image

patosullivan:
Can't seem to repro the sawRope weirdness I was seeing earlier.

patosullivan:
Ok, pending Ed's review I think this is good to go.

urcades (changes requested):
Changes are looking great, can't wait to test these out.

I have a few suggestions re: the overall legibility of these on mobile:

Screenshot 2023-02-06 at 1 37 46 PM

Right now it appears that we constrain the horizontal size of the notification body such that it doesn't fully extend to the edge of the notification row — since we're strapped for space in a mobile viewport, we should extend this out to the rightward edge.

Screenshot 2023-02-06 at 1 38 30 PM

A small quirk I'm seeing in these screens is the size of the group icon changing — these should remain fixed to the size of the other icons we deploy here (16 I believe)

Screenshot 2023-02-06 at 1 39 33 PM

A last legibility thing I'd suggest we incorporate here is to simply render the labels for (GROUP • CHANNEL) as a visually "single" string instead of the two strings we currently seem to render as separate elements — the idea being that we can truncate a single string in a much more elegant manner, rather than vertically stack two side-by-side strings as we currently do.

urcades (approved):
🫡

arthyn (approved):
lgtm

jamesacklin (changes requested):
Looks good, a few things.

(1) See what I did in Landscape to change the verbiage here ("Latest of n new messages").

Screenshot 2023-02-08 at 12 57 05 PM

(2) Let's tighten the white space up here. Especially noticeable with the "Reply" button.

Screenshot 2023-02-08 at 12 59 06 PM

(3) We also want to break the mentions and reply notifications up into a line-clamped excerpt of the source, followed by the content. It should look like this:

image

https://github.com/tloncorp/landscape/blob/master/ui/src/nav/notifications/Notification.tsx#L172-L203

(4) Make sure you're using the right leading for the notification content (`leading-5`). The line-height looks a little squished.

All shipnames should use the user-set alias (if set in CalmEngine).

patosullivan:
@jamesacklin updated based on your feedback, lmk what you think (lifted some more of your stuff out of landscape):

image

jamesacklin (approved):
Looks good! See #1899, was easier to do than to explain.

PR #1886, opened 2023-02-06
chat: detecting disconnects

### LINKED ISSUE #1575: groups: unread markers in sidebar stop updating eventually
When I leave groups open for a while, I see no unread blue dots on the groups in the sidebar. When I refresh, there are new blue dots. It seems to stop updating after a while.

**To Reproduce**
1. Read all unreads
2. leave it a while
3. see no unreads
4. refresh
5. see a number of unreads

**Expected behavior**
unreads should update live

**Additional context:**
- Ship type: planet
- Instance type: self-hosted
- OS/browser: qutebrowser on arch linux

arthyn:
This should solve the silent disconnect issue seen in #1575 as well as the issue where the blue dot still appeared even though the chat itself was marked read. We weren't using the same method in this case which was causing the discrepancy.

We determine if we still have an active subscription channel by poking setting store every 30s. If we don't get a response back within 15s then we claim disconnected. At this point we're not doing any kind of automatic reconnection so the backoff does nothing for now. I'm curious what we think is the best course here.

Fixes #1575


patosullivan (approved):
Looks great.

PR #1892, opened 2023-02-08
 add confirm modals for destructive actions

### LINKED ISSUE #1851: groups&talk: deleting a message sometimes appears to hang
- Write and send a message
- From the hover menu select the red X to delete it
- Observe it sometimes appear to do nothing before it eventually deletes

We should have some way to communicate in-progress state to the user (for sends we do the `>>` arrows, for others we use a spinner, not sure the proper answer for this case - maybe grey out the message and then remove?).

### LINKED ISSUE #1852: groups: confirm dialogs for all delete actions
As an admin it is very easy to delete messages accidentally and without warning.

Currently delete is next to thread and react. This could be moved, there could be a confirmation modal, or something else.

There may be other actions like this to protect.

patosullivan:
Fixes #1852 

Adds a generic "confirm" dialog that can be used for any destructive action. Specifically used here for message deletions and user kicks/bans.

image

Adds a loading spinner for message deletion, fixes #1851.

Also cleans up our Dialog component (use portals to avoid z-index and focus issues) and the Delete Group Dialog component (there was a bug where a user couldn't actually close that dialog).

I didn't replace the Delete Group Dialog component since it has some special functionality (type out the group name to enable the delete button).


jamesacklin:
A few issues with this, unfortunately.

Just for consistency's sake, let's make sure everything has a Cancel button as well.

If the spinner is active in the button, we don't need the label. Spinner or label, one or the other.

Let's rephrase this a bit, we're missing an article and "containing members" sounds robotic. Screenshot 2023-02-08 at 12 43 00 PM

No confirmation for deleting notes:
image

No confirmation for deleting curios:
image

Introduces a regression where interacting with the Add Channel or Edit Channel modals close them immediately:

https://user-images.githubusercontent.com/748181/217611791-fff07638-868f-4a35-981f-ee246efbbfb1.mov

https://user-images.githubusercontent.com/748181/217611809-73311c9e-036f-4d99-a198-fe3265ba1503.mov

jamesacklin (changes requested):
See comment.

ui/src/components/Dialog.tsx
```diff
@@ -9,7 +9,9 @@ export default function Dialog({
 }: DialogPrimitive.DialogProps) {
   return (
     <DialogPrimitive.Root {...props}>
-      <DialogPrimitive.Overlay className="fixed inset-0 z-40 transform-gpu bg-black opacity-30" />
+      <DialogPrimitive.Portal>
```
arthyn, line 12: I think you might want this to surround children too like they do in the example here: https://www.radix-ui.com/docs/primitives/components/dialog

ui/src/groups/GroupAdmin/GroupInfoEditor.tsx
```diff
@@ -163,7 +163,11 @@ export default function GroupInfoEditor({ title }: ViewProps) {
           >
             Delete {group?.meta.title}
           </button>
-          <DialogContent containerClass="max-w-[420px]">
+          <DialogContent
+            onInteractOutside={() => setDeleteDialogOpen(false)}
```
arthyn, line 167: You shouldn't have to do this or the onclick on `DialogClose`. you can add an onOpenChange to the `Dialog` component that does this for all these cases

arthyn (changes requested):
a few things with dialogs, also it looks like your newchannel snap shot got blown up

ui/src/groups/GroupAdmin/GroupInfoEditor.tsx
patosullivan: This seems to be the only instance it was used in, pushing the fix now.

ui/src/components/Dialog.tsx
patosullivan: Pushing the fix for this now.

ui/src/groups/GroupAdmin/GroupInfoEditor.tsx
```diff
@@ -182,7 +181,7 @@ export default function GroupInfoEditor({ title }: ViewProps) {
             <div className="flex justify-end space-x-2">
               <DialogClose
                 className="secondary-button"
-                onClick={() => setDeleteDialogOpen(false)}
+                // onClick={() => setDeleteDialogOpen(false)}
```
arthyn, line 184:
```suggestion
```

arthyn (approved):
lgtm!

jamesacklin (approved):
Much better, thank you!

PR #1900, opened 2023-02-08
leap: layout tweaks from design

### LINKED ISSUE #1660: leap: ux enhancements
cross-posted from the comments of the initial Leap PR (https://github.com/tloncorp/landscape-apps/pull/1309):

This is excellent work — I'm excited to begin reincorporating Leap in an intentional manner as we get time.

While we've been cooking up updates to Groups/Landscape writ large, we've pushed some edits to the overall Leap structure:

Screenshot 2022-12-12 at 11 30 26 AM

In @tomholford 's video, the current interface covered are the in-app commands list, but ideally we'd like to generalize + provide affordances for recently-viewed resources across app types — in Groups, we'd want to surface recent groups and channels, while in Talk, we'd want to surface recent conversations or files/links shared in conversations. 

For an initial release, we can shed these sections/features of course, but it's worth noting that Leap was always designed to accommodate more general search-like capabilities for power user types.

Accessing Leap/Search should happen by keystroke, as Tom has already implemented, but in line with this more "General Search Affordance" direction we'd like to take leap, we should surface it clearly in the nav bar of Groups:

Screenshot 2022-12-12 at 11 34 03 AM

It's worth noting that the eventual intention for most search affordances we provide to just be the universal Leap invocation, filtered/scoped down to the appropriate context.

On a more immediately-implementable note, all I'd ask is that we segment out Leap as drawn in the screens above, rather than contain in inside of a single "modal"-style lightbox interface. To make Leap's function as a "floating search bar" a little more transparent, we should visually distinguish it from our current modal interface pattern.

The current (very loose) specification for Leap will be developed here:

https://www.figma.com/file/DF5veycqn4Za6dYrMTG85N/2022-Q4-Groups?node-id=2146%3A26266&t=B3DyeDglX9yVkWBB-1

jamesacklin:
Adding a few tweaks from #1660 for a MVP blessing from @urcades.

No input:
image

- Selecting "Create New Group" pops a modal
- Selecting Talk opens that app in a new window

Matches for ships, channels, and groups:
image

Some stuff fell below the line for my hour or so of work on this that I think we can ship without. Obviously, we'll want avatars for groups/channels and ships, which will require a little refactoring. I couldn't gracefully add a LeapSectionRow header for Commands (default). We'll also need to define more hotkeys for those actions.

Fixes #1660


urcades:
Blessings sent, this is wonderful — let's get this out there 🫡

patosullivan (approved):
Looks nice

tomholford (approved):
Looks great! LGTM

jamesacklin:
Thanks @tomholford, feel free to take your main Leap PR out of draft mode.

Issue #1891, opened 2023-02-08
perms: bunt issue for can-read / can-write for a fresh comet joining UC Help


tomholford:
Seeing the `~zod` value in the scries below points to a bunt issue with the perms check.

To repro:

1. Boot a comet
2. Join UC > Help (~bitbet-bolbel)
3. See the `can-read` err
4. Attempt a dotpost
5. See the `can-write` err

<img width="431" alt="image" src="https://user-images.githubusercontent.com/16504501/217411435-c1a30ba3-41db-4395-8083-64099701552e.png">

<img width="449" alt="image" src="https://user-images.githubusercontent.com/16504501/217411450-98a18d6d-5caf-4dff-8746-779c7b836fca.png">

PR #1902, opened 2023-02-08
channels: guarantee we always have group flag immediately upon join

arthyn:
OTT, this is an extensive change but should ensure that we always have the group populated in `perm`. This crops up in issues like #1891


tomholford (approved):
Pulled and tested locally, LGTM

jamesacklin:
Also tested locally. Appears to work as intended -- there's no visible UI change so I can't comment there.

Issue #1567, opened 2022-12-17
chat: scrolling back on mobile loses place 


arthyn:
on mobile when loading old messages, your scrollbar stays at the same place so you end up much farther back than you were

jamesacklin:
Did you fix this?

arthyn:
I think so?

PR #1908, opened 2023-02-10
chat: optimize virtual scroller

### LINKED ISSUE #1510: chat: virtual scroller woes part 3
We have a few remaining issues with the virtual scroller/chat window centered around sending messages. It seems like after I send a message the window jumps when hearing the fact back from the poke. Similarly it seems that we are rerendering way too high in the tree after sending a message because if messages are sent in quick succession there's a chance one will get lost.

### LINKED ISSUE #1758: text wrapping in threads may interact strangely with scrollbar
Bit of an oddball here. I opened a thread, and this is what I saw in the sidebar:

basedjump

The scrollbar is trying to appear, reduces the width of the sidebar's contents, then wraps the url, massaging things in such a way that the scrollbar no longer appears... and then we repeat. Or something like that. Not actually sure what's going on here.

I can reproduce consistently in this thread, and have never seen it in others. Note that the "toggling" happens as fast as draw calls can happen, but the limited framerate of the gif makes it appear slower.

### LINKED ISSUE #1905: groups: search for group host timeout
- Go to Find Groups on a fresh ship
- Search `~nibset-napwyn` (if not on a fresh ship, any host you don't already know)
- Observe it spin and fail (with better failure text now thanks)
- Re-search it a second time after the timeout and observe the group it hosts appear

@jamesacklin and @arthyn this one is pretty bad for onboarding, this action is among the first things a new user does and it looks like they can't join any groups. Is it possible to bump this up the stack?

Tagging @emceetee for context (since she's the one that's been hitting it in onboarding)

arthyn:
This should fix most of what I saw in #1510. A few other cleanup things thrown in like the connection checking needing a bit of guarding and cleanup, and tweaking find groups timeout #1905.

Main lines of work here:
- Stop chat message options from rendering until hovered (might need to tweak this for mobile will do before merge). All these tooltip/popover/dialog/dropdown combos take a lot of time to render because they're complex. we should avoid these at all costs in messages or at least actively guard against rendering them until it's necessary.
- Remove scrolling based on height change. This was causing much of the jumping. It was supposed to be limited to only when you're at the bottom of scroll, but that seems to be unreliable. I left some adjustment for the "first pass" aka when things stop changing height after first mount. After that it won't fire anymore and virtuoso should take care of any adjustments.
- Remove unnecessary extra fetching for unread messages. This would be nice, but our measures for this are too inaccurate because of threads and how we filter replies out of our fetched results.
- Stop chat loader from showing if we have some messages. This was flickering the window on every channel change
- Debouncing isScrolling events except for when we start scrolling. virtuoso does a lot of weird updates for this value that happen frequently bouncing back and forth between false and true even though you're still scrolling.
- Adding back in the style to make sure that overflow is set to scroll. This is really important because without this, the scrollbar being added in causes a render loop from height changes. 

Fixes #1758
Fixes #1510 
Fixes #1905

Mobile issues in #1567 still seem to persist and may be an issue with virtuoso


tomholford (approved):
Tried it out locally, feels good. Nice work ! Understand that there are more changes coming per the PR description, but for now LGTM

arthyn:
Still seeing some issues on mobile investigating

jamesacklin:
Testing this out, will have to spend some time with it. Initial impressions:

- Only showing the message actions on hover made things much snappier. You'd have to know we made this change to notice the tiny delay. I think it's worth it.
- I was able to zoom back a week in the team chat channel with relative ease. This is pretty arduous on master.
- No noticeable regressions. Chat still scrolls.

jamesacklin (approved):

PR #1909, opened 2023-02-10
leap: escape to close

patosullivan:


jamesacklin (approved):
Does this interfere with an open thread / another open modal?

tomholford (approved):
LGTM

patosullivan:
> Does this interfere with an open thread / another open modal?

just checked, does interfere with threads and other places. fix incoming.

PR #1915, opened 2023-02-10
chat: add replies

### LINKED ISSUE #1747: groups/talk: reply to message without thread
oftentimes I don't wanna start a thread when I reply to a message, I just wanna quote-reference it inline. Currently I have to copy the message link and add it as an attachment. This is kinda annoying. It would be nice if there was a direct quote-reply button

patosullivan:
fixes #1747

Adds replies to the main chat input in chat channels. DMs and threads excluded.

https://user-images.githubusercontent.com/1221094/218164649-745b6296-841b-43ee-a9e9-4956a54e9910.mov


urcades:
Before I get to reviewing — is this an intentionally-downscoped + MVP iteration of replies to get back to parity with the previous version of Groups? As I recall, "replies" in Groups1 were similarly structured as:

- reference to OP's replied-to post
- OP mention
- insertion point added after mention

Which is what this PR appears to be doing

patosullivan:
> Before I get to reviewing — is this an intentionally-downscoped + MVP iteration of replies to get back to parity with the previous version of Groups? As I recall, "replies" in Groups1 were similarly structured as:
> 
> * reference to OP's replied-to post
> * OP mention
> * insertion point added after mention
> 
> Which is what this PR appears to be doing

yep, that was my goal here.

patosullivan:
I think in a future sprint we could come back through and style this appropriately (and add appropriate notifications without requiring the mention, which may require storing "replies" in a new way on the backend since we currently use that field for thread replies and we're going to start filtering those out on initial load of a chat channel).

urcades (changes requested):
This is functionally excellent, and the only request I'd have is for us to ensure that the elements I'm pointing between with a red arrow are spaced out similarly to the area indicated by the green measure:

Screenshot 2023-02-10 at 3 16 50 PM

If we're willing to slightly update the reference element/component in order to make replies more legible, I'd recommend we also make the reference metadata more visually diminutive:

Section 1

^ This is a side-by-side comparison of the existing reference component, and margin + element sizing updates we could potentially make so that replies are a little easier to read. If updating the reference element is out of scope, I can place these notes into a new issue for later.

patosullivan:
> This is functionally excellent, and the only request I'd have is for us to ensure that the elements I'm pointing between with a red arrow are spaced out similarly to the area indicated by the green measure:
> 
> Screenshot 2023-02-10 at 3 16 50 PM
> 
> If we're willing to slightly update the reference element/component in order to make replies more legible, I'd recommend we also make the reference metadata more visually diminutive:
> 
> Section 1
> 
> ^ This is a side-by-side comparison of the existing reference component, and margin + element sizing updates we could potentially make so that replies are a little easier to read. If updating the reference element is out of scope, I can place these notes into a new issue for later.

I think these are all doable.

patosullivan:
Hey @urcades, I made the sigil much smaller and added the 16px padding everywhere, let me know what you think:

image

urcades:
16pt padding on the message itself is great,  the last edit i’d make here is to clamp down the vertical margin (red) on the reference author element, making it equivalent to the left margin (green):

urcades:
16pt padding on the message itself is great,  the last edit i’d make here is to clamp down the vertical margin (red) on the reference author element, making it equivalent to the left margin (green):

37AA53B0-6520-49DC-A0DF-C683DD3F7D3A

patosullivan:
> 16pt padding on the message itself is great, the last edit i’d make here is to clamp down the vertical margin (red) on the reference author element, making it equivalent to the left margin (green):
> 
> 37AA53B0-6520-49DC-A0DF-C683DD3F7D3A

Done:
image

urcades (approved):
🫡

ui/src/chat/ChatInput/ChatInput.tsx
```diff
@@ -80,14 +83,20 @@ export default function ChatInput({
...
+  const groupFlag = useGroupFlag();
   const subscription = useSubscriptionStatus();
   const pact = usePact(whom);
   const id = replying ? `${whom}-${replying}` : whom;
   const chatInfo = useChatInfo(id);
-  const reply = replying || chatInfo?.replying || null;
-  const replyingWrit = reply && pact.writs.get(pact.index[reply]);
+  // const reply = replying || chatInfo?.replying || null;
+  // const replyingWrit = reply && pact.writs.get(pact.index[reply]);
+  const replyingWrit = chatReplyId && pact.writs.get(pact.index[chatReplyId]);
```
arthyn, line 99: Unclear why we needed to replace the chat store behavior with routing. It seems the same?

arthyn (approved):
lgtm, one question and we just need to remember that we'll need to rip a lot of this out when we implement these through the backend

PR #1916, opened 2023-02-10
groups: simplify grouplist, fix scroll pinned groups issue, remove dnd

### LINKED ISSUE #1824: groups: pinned groups scroll on mobile is broken
- Pin a couple groups
- On mobile observe there are two sections (pinned groups and all groups)
- Attempt to scroll
- Observe you can only scroll in the all groups section

This feels broken when you’re using it - also if you have more than 6 or so pinned groups, the all groups section gets pushed off screen and you can’t get to it at all since you can’t scroll.

Scrolling should scroll the entire sidebar, not just the all groups section.

### LINKED ISSUE #1835: talk: ui shows pinned when no channel is pinned
- Pin a group in groups, but don’t pin anything in talk 
- On mobile go to talk
- Observe there’s an empty pinned section at the top

61780603-C5D5-4F9F-A545-6D7BD56FB6CB

There should only be a pinned section if channels in talk are pinned. Only pinning a group shouldn’t add it.

patosullivan:
Fixes #1824 

Didn't create the context for scrolling since we don't have nearly as many items in this list as we do in Talk.

We do have a separate scrollable section here for non-pinned groups since we're using Virtuoso to render them separately. Should we?

https://user-images.githubusercontent.com/1221094/218182443-156d07a8-f7fa-421c-9ba6-b6b0b74f818d.mov

Also fixes #1835

### DIFF
--- a/ui/src/components/Sidebar/GroupList.tsx
+++ b/ui/src/components/Sidebar/GroupList.tsx
@@ -1,9 +1,8 @@
-import cn from 'classnames';
-import React, { PropsWithChildren, useEffect, useMemo } from 'react';
-import { uniq, without } from 'lodash';
-import { DndProvider, useDrag, useDrop } from 'react-dnd';
-import { HTML5Backend } from 'react-dnd-html5-backend';
-import { TouchBackend } from 'react-dnd-touch-backend';
+import React, { useCallback, useMemo, useRef, useState } from 'react';
+import { debounce } from 'lodash';
+// import { DndProvider, useDrag, useDrop } from 'react-dnd';
+// import { HTML5Backend } from 'react-dnd-html5-backend';
+// import { TouchBackend } from 'react-dnd-touch-backend';
 import * as Popover from '@radix-ui/react-popover';
 import { Virtuoso } from 'react-virtuoso';
 import { useIsMobile } from '@/logic/useMedia';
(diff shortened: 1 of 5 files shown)


arthyn:
I think this should mimic talk where it's all in one scroll zone. right now it's giving two scrollbars which is weird

patosullivan:
Went ahead and mimicked talk entirely. Why not?

jamesacklin:
Nice, fixes the offending pinned-scrolling issue.

A few notes on alignment:

image

patosullivan:
> Nice, fixes the offending pinned-scrolling issue.
> 
> A few notes on alignment:
> 
> image

lmk if the latest commit looks good.

ui/src/groups/GroupAvatar.tsx
```diff
@@ -42,7 +42,8 @@ export default function GroupAvatar({
   loadImage = true,
 }: GroupAvatarProps) {
   const { hasLoaded, load } = useAvatar(image || '');
-  const showImage = hasLoaded || loadImage;
+  const imageIsColor = isColor(image || '');
+  const showImage = hasLoaded || loadImage || imageIsColor;
```
arthyn, line 46: I think a lot of the logic is getting conflated here.

```suggestion
  const imageIsColor = isColor(image || '');
  const showImage = image && !calm.disableRemoteContent && !imageIsColor && (hasLoaded || loadImage);
```

Then below we can do:

```
if (imageIsColor) {
    background = image;
  } else {
    background = dark ? '#333333' : '#E5E5E5';
  }
  return showImage ? (
    <img className={cn('rounded', size, className)} src={image} onLoad={load} />
  ) : (
    <ColorBoxIcon
      className={cn('rounded', size, textSize(size), className)}
      color={background}
      letter={title ? symbols[0] : ''}
    />
  );
```

arthyn (changes requested):
Two changes then gtg I think

jamesacklin:
Chipped in with my label alignment nit. Hope you don't mind :)

image

jamesacklin (approved):

Issue #1305, opened 2022-12-02
Ability to add ships to the contact store


patosullivan:
At the moment we only have contact-store data that was added via old groups. We are not adding ships that are newly interacted with in groups/talk to the contact store.

jamesacklin:
This is a significant issue, has architectural implications, and needs to be talked about. Tagging in @lukestiles for visibility--this is what I mentioned earlier this week.

Notes from @arthyn: 

- Even if we add the pokes to propagate contact details through the host, we're still using the old %landscape agent and corresponding groups themselves. 
- @liamfitzgerald had updated %contact-store to work with Groups 2 and we have not implemented those pokes.
- We probably want to move the %contact-store agent (and other Groups 1 agents) into the proper desk as a prerequisite to this integration work.

jamesacklin:
@Bradyab So that you may watch.

zalberico:
@jamesacklin does this mean new ships can't set profile info or custom nicks? A user in UC help (~rivmyr-ronpem) can't see nicks, but I thought that was because his old groups wasn't up to date - is that expected for new ships/newly breached ships?

jamesacklin:
That is correct.

tomholford:
@jamesacklin Would like to sink my teeth into this, but per this comment:

> We probably want to move the %contact-store agent (and other Groups 1 agents) into the proper desk as a prerequisite to this integration work.

Have we decided if / when to rip this bandaid off?

lukestiles:
Some initial work done here https://github.com/tloncorp/landscape-apps/blob/master/desk/app/contact-store.hoon. 

It may not be necessary to federate, there are alternate strategies to reduce network load. Please confirm with @philipcmonk.

jamesacklin:
Posting incremental notes here for transparency.

There are a few lines of work on urbit/urbit informing our approach (namely https://github.com/urbit/urbit/pull/5827). Our belief is that the data volume of a profile update isn't that high; past problems with direct contact distribution were related to the lifecycle of subscriptions rather than the data load itself.

We will naively subscribe ships to profile updates using progressively passive methods. There are some actions in the UI that necessitate an immediate refresh of a single ship's profile data; other actions can acceptably defer the refresh of a large-ish number of ships (joining UC/general, for example).

User actions that will call for an immediate (re)subscription and refresh of one or more ships' profile data: 

- Viewing another ship's profile directly
- Starting (or being invited to) a DM with another ship 
[...]

PR #1931, opened 2023-02-14
ux: Leap sorts by pinned, unreads, and pals

tomholford:
# Context

After using preview Leap, noticed I was using it to primarily jump between my unreads and pins. So, this PR bakes in some of this behavior.

# Changes

- [x] score Group results by membership in `pinnedGroups`, then by `isGroupUnread`, then A-->Z
- [x] score Channel results by membership in `pinned`, then if it's a member of a group in `pinnedGroups`, then by if the channel is within the currently viewed group, then by `isChannelUnread`, then by A-->Z
- [x] score Ships results by prioritizing mutuals in %pals graph, the mention heuristics (e.g., downrank comets), then by unread status
- [x] fuzzy string compare when filtering

### DIFF
--- a/ui/src/components/Leap/MenuOptions.tsx
+++ b/ui/src/components/Leap/MenuOptions.tsx
@@ -29,18 +29,11 @@ export interface IMenuOption {
 
 export const groupsMenuOptions: IMenuOption[] = [
   {
-    title: 'Profile',
-    subtitle: '',
-    to: '/profile/edit',
-    icon: CommandBadge,
-  },
-  {
     title: 'Notifications',
     subtitle: '',
     to: '/',
     icon: CommandBadge,
   },
-
   {
     title: 'Find Groups',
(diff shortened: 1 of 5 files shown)


patosullivan (approved):
This is awesome.

jamesacklin:
Leaving a note for testing: make sure this works without %pals

patosullivan:
Not covered in this PR, but some issues I've noticed with leap:
1) if your ship does not have a contact store entry for a ship, you can't navigate to a DM with that ship (even if you've started one with them). This effects all new ships (which are currently effected by #1305).
2) I can't search for and navigate to a group DM.

patosullivan:
#1936 and #1937

arthyn:
Should we merge?

jamesacklin:
Does it work without pals installed?

PR #1932, opened 2023-02-14
chat: add embed content

patosullivan:
A first stab at embed content. @urcades I just did this off the cuff, let me know what you think.

Only supports youtube, twitter, spotify, and direct audio links for now.

https://user-images.githubusercontent.com/1221094/218745164-4965430f-f237-4636-b219-b841eeb5679f.mov

https://user-images.githubusercontent.com/1221094/218745175-f34b96bd-2501-4b87-8c8e-fe50ec8bf3bc.mov


tomholford (approved):
Pulled and tested this locally. Working as expected:

image

arthyn:
@urcades were you ever able to give this a look?

urcades (changes requested):
@arthyn finally getting around to it!

This functionally is looking fantastic — if we wanted to perform a few more stylistic/surface-level tweaks, I'd only recommend we style the lightbox interface as such (with close button anchored to the edge of the viewport), so that we don't get too much interference between the close/X button and any interfaces/buttons located within the embed itself:

Lightbox

One other tiny detail we can stand to tweak is the bottom padding/spacing of the embed's metadata row:

Screenshot 2023-02-23 at 10 38 21 AM

Note the uneven blue boxes here indicating the element spacing — we should tuck in the vertical spacing a bit such that it's equivalent to the horizontal spacing, and equidistant all around the embed contents.

Other than these tweaks, I only have a few questions:

In the included update, is there a way to play embeds within the chat, or must a lightbox be opened every time to view/play the embed content?

How does one view the source material each embed points to? Is this possible via the metadata text underneath the embed content? Through the lightbox view of the embedded content?

Since we're implementing a lightbox view here, would it be possible to apply the same styling to plain images posted "natively" to a talk channel? It'd be great if we could simply make lightboxing a regular pattern for all media posted to a talk channel, or to notebooks, etc.

patosullivan:
After merging master into this branch something is up with the audio player component. It's constantly re-rendering after a scroll. Is this a VS issue? @arthyn

patosullivan:
Hey @urcades,

Yes, they need to be rendered outside of the virtual scroller to avoid having the scroller redraw and jump around.

I've updated the spacing, looks like this now:

image

I've also created a new light box component that we could use anywhere in the app, looks like this:
image

I can add some functionality to show images in the light box pretty easily too, commit incoming.

patosullivan:
Ok, added the light box in for images:


https://user-images.githubusercontent.com/1221094/221283899-3fa261e3-23b8-4ac4-ab84-8d3564354307.mov

patosullivan:
> How does one view the source material each embed points to? Is this possible via the metadata text underneath the embed content? Through the lightbox view of the embedded content?

I made the title for the embeds a clickable link, but I've also now added the "Source" button from your mocks which is a link to the source material.

urcades (approved):
Updates look fantastic, thanks for answering the Qs!

ui/src/chat/ChatEmbedContent/YouTubeEmbed.tsx
```diff
@@ -0,0 +1,77 @@
...
+        style={{
+          backgroundImage: `url(${thumbnail})`,
+          backgroundSize: 'cover',
+          backgroundPosition: 'center',
+          width: isMobile ? 300 : thumbnailWidth,
+          height: isMobile ? 300 : thumbnailHeight,
+        }}
+        className="rounded"
+      />
+      <div
```
arthyn, line 41: Should be button

ui/src/chat/ChatEmbedContent/SpotifyEmbed.tsx
```diff
@@ -0,0 +1,66 @@
...
+        style={{
+          backgroundImage: `url(${thumbnailUrl})`,
+          backgroundSize: 'cover',
+          backgroundPosition: 'center',
+          width: thumbnailWidth,
+          height: thumbnailHeight,
+        }}
+        className="rounded"
+      />
+      <div
```
arthyn, line 34: Should be button

ui/src/chat/ChatContent/ChatContentImage.tsx
```diff
@@ -15,25 +16,39 @@ export default function ChatContentImage({
...
-      </a>
+      {calm?.disableRemoteContent ? (
+        <span>{src}</span>
+      ) : (
+        <img
+          src={src}
+          className="max-w-full cursor-pointer rounded"
+          height={height}
+          width={width}
+          onClick={() => setShowLightBox(true)}
```
arthyn, line 34: Probably should wrap the image in a transparent button with this onclick

arthyn (changes requested):
buttons!

PR #1934, opened 2023-02-14
Only show pinned if we have talk pins, filter out left/deleted group channels from briefs

patosullivan:
Applying the same fix applied to the mobile messages list (only show the "Pinned" section if we have pinned channels/DMs, not pinned groups, which won't render anyway), and fix a separate issue where a blank row could appear in the list if a user has left a group/channel or that group/channel was deleted (not sure how we missed this before).

### DIFF
--- a/ui/src/dms/MessagesList.tsx
+++ b/ui/src/dms/MessagesList.tsx
@@ -89,6 +89,10 @@ export default function MessagesList({
           return false;
         }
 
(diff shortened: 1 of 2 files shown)


tomholford (commented):
Changes look good, but unable to test locally because of https://github.com/tloncorp/landscape-apps/issues/1927

arthyn (approved):
lgtm

PR #1935, opened 2023-02-14
chat: fix new dms

### LINKED ISSUE #1927: New DM breaks on first message
When I go to create a fresh DM with an unknown ship, my first message doesn't seem to go through. 

**To Reproduce**
1. Create a DM with some @p
2. Send a message
3. See an error in the console.

**Expected behavior**
A message going through

**Additional context:**
- OS/browser: Safari on MacOS

Screen Shot 2023-02-13 at 1 43 25 PM

arthyn:
There were two bugs here. Fixes #1927 and also ensures that ship selector always outputs valid @p's, because if you were typing a patp without one and selecting it when it was an exact match (not in contact store) this would pass a bad `whom` along.

### DIFF
--- a/ui/src/components/ShipSelector.tsx
+++ b/ui/src/components/ShipSelector.tsx
@@ -395,7 +395,7 @@ export default function ShipSelector({
 
         if (parts.every((s) => isValidNewOption(preSig(s)))) {
           const options: ShipOption[] = parts.map((p) => ({
-            value: p,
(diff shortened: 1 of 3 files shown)


patosullivan (changes requested):
This does allow the chat to get created, but the first message itself doesn't appear.

patosullivan (approved):
LGTM

PR #1938, opened 2023-02-14
gallery: virtualize list, fetch new items at bottom

### LINKED ISSUE #1928: gallery: index only shows 50 posts
At the moment, we only fetch the first 50 posts in a Gallery channel. We should add pagination the same way we did to fix #1458.

Steps to reproduce:

1. Open a Gallery with >50 items (the "archive" channel in ~natnex-ronret/door-link is a good example)
2. See ~50 items

Expected behavior: paginate and lazy-load the next 50 curios.

patosullivan:
Fixes #1928 

Add VirtuosoGrid, fetch 50 new items when we reach the bottom of the current list. Had to create a fake curio and prepend it to the sortedCurio array (and then replace it) in order to get the NewCurioForm to render in the list.

https://user-images.githubusercontent.com/1221094/218874834-cf4f1e41-cb4b-4a86-b5e9-d335f5946c4e.mov

### DIFF
--- a/ui/src/heap/HeapChannel.tsx
+++ b/ui/src/heap/HeapChannel.tsx
@@ -16,6 +16,7 @@ import {
   useHeapState,
   useHeapPerms,
 } from '@/state/heap/heap';
+import { VirtuosoGrid } from 'react-virtuoso';
 import ChannelHeader from '@/channels/ChannelHeader';
 import {
   HeapSetting,
@@ -36,6 +37,7 @@ import {
 import { GRID, HeapCurio, HeapDisplayMode, HeapSortMode } from '@/types/heap';
 import useRecentChannel from '@/logic/useRecentChannel';
 import useAllBriefs from '@/logic/useAllBriefs';
+import makeCuriosStore from '@/state/heap/curios';
 import NewCurioForm from './NewCurioForm';
 
 function HeapChannel({ title }: ViewProps) {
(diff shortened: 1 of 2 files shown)


ui/src/heap/HeapChannel.tsx
```diff
@@ -138,41 +140,90 @@ function HeapChannel({ title }: ViewProps) {
...
-          <div className="aspect-h-1 aspect-w-1">
-            <HeapBlock curio={curio} time={time.toString()} />
-          </div>
-        ) : (
-          <HeapRow key={time.toString()} curio={curio} time={time.toString()} />
-        )}
-      </div>
-    ),
-    [displayMode, navigateToDetail]
+    (i: number, curio: HeapCurio, time: bigInt.BigInteger) =>
```
arthyn, line 143: I wonder if the `Header` component that virtuoso has would work here instead of having to fake it out

ui/src/heap/HeapChannel.tsx
```diff
@@ -197,22 +248,18 @@ function HeapChannel({ title }: ViewProps) {
...
-      <div className="h-full overflow-y-scroll p-4">
-        <div
-          className={cn(
+      <div className="h-full p-4">
+        <VirtuosoGrid
+          data={canWrite ? sortedCurios : sortedCurios.slice(1)}
+          itemContent={(i, [time, curio]) => renderCurio(i, curio, time)}
+          computeItemKey={computeItemKey}
+          style={{ height: '100%', width: '100%', paddingTop: '1rem' }}
+          atBottomStateChange={loadOlderCurios}
```
arthyn, line 257: this fires on both entry and leaving bottom

arthyn (approved):
lgtm, some things to maybe consider

ui/src/heap/HeapChannel.tsx
patosullivan: Tried it, it places it outside of the grid (above it).

ui/src/heap/HeapChannel.tsx
patosullivan: Good catch, fixed in latest commit.

PR #1940, opened 2023-02-15
groups: hide reject button on group refs for joined groups

### LINKED ISSUE #1910: groups, talk: group refs show a reject button
DB872BC7-A7A5-4882-8D77-DAD7D67EF612

Reject in that image is weird, there’s no need for it.

patosullivan:
Fixes #1910 

Still not clear on why these are showing up as gang invites, but for now we can hide the button if we know that you're in the group.

### DIFF
--- a/ui/src/components/References/GroupReference.tsx
+++ b/ui/src/components/References/GroupReference.tsx
@@ -101,7 +101,7 @@ export default function GroupReference({
           </span>
         ) : (
           <>
-            {gang.invite && button.text !== 'Go' && status !== 'loading' ? (
+            {gang.invite && !group && status !== 'loading' ? (
               <button
                 className="small-button bg-red text-white dark:text-black"
                 onClick={reject}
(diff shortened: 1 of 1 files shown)


ui/src/components/References/GroupReference.tsx
```diff
@@ -101,7 +101,7 @@ export default function GroupReference({
           </span>
         ) : (
           <>
-            {gang.invite && status !== 'loading' ? (
+            {gang.invite && button.text !== 'Go' && status !== 'loading' ? (
```
arthyn, line 104: might be safer to check if the group doesn't exist

arthyn (changes requested):

ui/src/components/References/GroupReference.tsx
patosullivan: Good idea.

arthyn (approved):
lgtm

PR #1946, opened 2023-02-15
chat: do not insert replies into browser history

### LINKED ISSUE #1945: chat: single-message-replies URL params are getting added to history, makes "back" not work as expected.

patosullivan:
Fixes #1945 
Also did some cleanup.


arthyn (approved):
lgtm

PR #1948, opened 2023-02-15
chat: fix chat input loss of focus

### LINKED ISSUE #1942: chat: chat input on mobile is losing focus after send

patosullivan:
fixes #1942 

Focus was apparently lost because we were passing a prop directly into sendMessage, which caused a re-render.


arthyn (approved):
lgtm

PR #1949, opened 2023-02-15
chat: fix message options not appearing on mobile

### LINKED ISSUE #1943: chat: message options sometimes don't appear on mobile

patosullivan:
fixes #1943


arthyn (commented):
the more I think about this I wonder if it would be better to do focus/blur so that way it goes away if you click something else. the only issue we may encounter here is that we might need to check focus within for tabbing

arthyn (approved):
lgtm

PR #1954, opened 2023-02-15
chat: drafts reborn

### LINKED ISSUE #1925: talk: text in chat gets wiped
- Be writing a message in a dm
- Take longer than 10s to send it
- Observe it get wiped

This was an issue I think was solved but appears to have been reintroduced in a regression as of this morning (Feb 13th)

### LINKED ISSUE #1953: groups, talk: input state wiped on reload
- Be typing a message
- Observe what you type get wiped
- It appears to happen more often on mobile and when the yellow banner appears or things reload?

arthyn:
This changes how we handle the `ChatInput` around the app. Since we really want distinct breaks between one chat or another I added keys to force remount between chats, similar to the scroller. This makes the initialization of the input much cleaner since we know that we won't have states lingering between. To combat any kind of wiping this now stores typed messages in local storage. Fixes #1953, fixes #1925

### DIFF
--- a/ui/src/chat/ChatChannel.tsx
+++ b/ui/src/chat/ChatChannel.tsx
@@ -113,7 +113,13 @@ function ChatChannel({ title }: ViewProps) {
             )}
           >
             {canWrite ? (
-              <ChatInput whom={chFlag} sendMessage={sendMessage} showReply />
+              <ChatInput
+                key={chFlag}
+                whom={chFlag}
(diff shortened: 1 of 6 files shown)


patosullivan (approved):
Looks great, works well locally and on mobile.

jamesacklin:
Pulling and checking locally.

jamesacklin (approved):
Works on my machine(s).

PR #1955, opened 2023-02-15
chat: fix emoji picker disappearing on subsequent messages

### LINKED ISSUE #1950: chat: emoji modal won't stay open on subsequent messages from the same user
If a user sends more than one message in sequence and you try to react to one of the subsequent messages the emoji modal will close as soon as you move your mouse away from the chat message options.

Certainly related to changes with handling hover state on ChatMessageOptions.

patosullivan:
Fixes #1950 

Was caused by the new way we handle hover state in chat messages. As soon as we leave the div for the message we'll lose its children that depend on hover. There was enough space between the message options and the emoji picker for the onOut to fire. Had to sacrifice some space between the picker and the chat message options.


tomholford (approved):
LGTM

PR #1957, opened 2023-02-15
chat: prevent stuck at top and add scroll to message

arthyn:
**scroll to:**
I thought we had this, but the scroller has been through so many changes not sure what happened to it. This adds back in the ability to detect `scrollTo` changes and scroll appropriately. All the dependencies in the `useEffect`'s are very carefully set here. Also found that we weren't always using the correct time for scrolling back so I added that as well.

**stuck at top**
I think recent changes I made made this more likely to happen or maybe it was always this way. Either way, this alters the atTop threshold so that we can escape the "top" on small screens if there's not enough messages to trigger a state change and also ensures that if we initially load at the top because of a `scrollTo` that we manually start loading older messages since `atTopChange` won't fire.

**misc**
- removed errant console log
- connection checker fires too early on slow loads


patosullivan (approved):
LGTM, works as expected locally.

jamesacklin (approved):
(astonished voice) Am I the blocker?!

Works for me, I'm able to click a reference in Chat and scroll back in the log to see its original context.

PR #1962, opened 2023-02-16
groups: prevent channel host removal

### LINKED ISSUE #1807: groups: channel hosts should not be able to be kicked or banned
Related: https://github.com/tloncorp/landscape-apps/issues/1765

It's a little weird that a channel host needs to be an admin to create the channel, but can then be removed as admin.

I think the whole channel hosts separate from group hosts in the current UX doesn't make sense (for it to make sense, groups would have to operate as a different type of abstraction?) - but as it is, probably if you're a channel host you should also be a group admin? Something about this is weird.

arthyn:
Although there are no real technical issues with removing channel hosts there are some UX concerns. Given that this removes the UI for kicking/banning channel hosts and also guards against it on the backend. Fixes #1807

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -715,6 +715,14 @@
       out
     (~(put in out) who)
   ::
+  ++  go-channel-hosts
+    ^-  (set ship)
+    %-  ~(gas in *(set ship))
+    %+  turn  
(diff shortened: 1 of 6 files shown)


patosullivan (approved):
lgtm

PR #1965, opened 2023-02-16
notifications: make sure we load on mobile and tweak layout

### LINKED ISSUE #1961: groups: home is blank on mobile
- Open groups on mobile (iOS safari)
- Click a group
- Go to blank 'home' with no notifications

I *do* have notifications on home for the group on the desktop (macOS chrome) and I have notifications on mobile in garden/landscape.

arthyn:
OTT we just forgot to initialize the group notifications, fixes #1961 also includes some tweaks so that the container stays the right size

### DIFF
--- a/ui/src/groups/Groups.tsx
+++ b/ui/src/groups/Groups.tsx
@@ -93,7 +93,7 @@ function Groups() {
   }
 
   return (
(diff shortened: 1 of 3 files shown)


jamesacklin (approved):
Makes sense

PR #1968, opened 2023-02-17
groups: fix group sidebar icons in dark mode

patosullivan:
OTT. No isssue for this one afaik.


jamesacklin (approved):
Yes. We should have a test that looks for `mix-blend-multiply` and warns if it doesn't have a `dark:mix-blend-*` class to go along with it.

PR #1969, opened 2023-02-17
chat: prevent extra break/space on messages with uploaded images

### LINKED ISSUE #1741: talk: fix bottom padding of media/image content posted to talk
While hovering over an image posted to Talk, note the superfluous bottom padding:

Screenshot 2023-01-19 at 2 12 05 PM

We should normalize the spacing around media content within a message row, 8px all around.

See #1675 for related margin/padding issues, we could probably solve this issue and the reference padding issues in one fell swoop.

patosullivan:
fixes #1741 

normalizeInline will produce an object like this { break: null } when we upload an image because tiptap sends an empty paragraph object when no text has been added in the chat input.


arthyn (approved):
lgtm

Issue #1840, opened 2023-01-31
notebooks: images do not appear to post when inserted into a bullet list


urcades:
Currently, it appears as though we do not render images added as children to a bullet list in Notebooks.

Here's a video of me attempting to add an image to a notebook post, and it failing:

https://user-images.githubusercontent.com/1195363/215844810-b09e6320-9e68-4b1a-8b4b-15ab77d54875.mov

When I inspect the node in developer tools, the `<li>` element is empty – nothing is attached to the notebook.

The resulting limitation is related to #1837

patosullivan:
My original plan was just to disable this menu in a list item, but there does not appear to be a way to disable the action menu based on the node type it's called from with tiptap/prosemirror.

The issue here is that we can't store block content within a list item on the backend. I've looked into doing this (i.e., allowing an item to be a `verse` type, which can be either inline or block), but this appears to be incompatible with the diary-graph migration. Tried investigating that, but the migration code is pretty complicated and I'm worried I might break something there.

@arthyn or @midsum-salrux do you know if there's a way we could change the typedef for listing (on lines 156-159 in sur/diary.hoon) to accept `p= (list verse)` for `%item` and not break the migration code?

PR #1972, opened 2023-02-17
notebooks: disable blocks in lists

arthyn:
This should solve #1840 temporarily until we can actually support blocks in lists.


patosullivan (approved):
Nice!

PR #1975, opened 2023-02-17
groups: add real progress to mark read

arthyn:
OTT, this makes the loading indicator actually follow the progress of updating. This does change a bit how we update notifications upon a "saw" because now they update once success is tracked. Also cleaned up the notifications page so there's less duplication.


tomholford (approved):
Smooth! LGTM

PR #1986, opened 2023-02-21
chat: using tiptap editor in a 'controlled' way is bad

### LINKED ISSUE #1982: talk: message input in fresh DM loses focus
From @ashelkovnykov:

I'm seeing some sort of instability in Talk. Right now, when I try to DM someone with whom I've never had a conversation yet, after I select his @p and start trying to write a message, after typing a few characters the text box will lose focus. This makes it almost impossible to type, as if I type at least one character and then pause for even half a second, the text box will lose focus and I can't keep writing without manually bringing focus back onto the textbox.

I think it has something to do with repeated attempts by Talk to fetch our conversation history (which doesn't exist)

My reasoning is as follows: this same issue happened to me yesterday while I was having connectivity issues between my browser and ship. It presented exactly the same, but when I noticed that for some reason my browser had lost connection via Eyre, I restarted my ship and my browser session. When I then went into Talk and tried to message the same ship again, this time it turned out that I had talked to this ship via DM before: when I put his @p into the "new message" box in Talk, it fetched our previous conversation history, and the above problem didn't occur.

arthyn:
Fixes #1982, my previous change for drafts in #1954 basically made the tiptap editor act in a controlled manner which was mostly fine except that since it was constantly recreating the editor it was pulling focus off and reapplying it in places where we have `autoFocus` on. However, in the new DM situation we don't want the input to pull focus over the ship selector so we saw this bug. This switches things around so that we only set the draft on first mount. We need to put this through it's paces though I'm worried about some kind of async bug where the draft gets applied repeatedly or after submitting, etc.

### DIFF
--- a/ui/src/chat/ChatInput/ChatInput.tsx
+++ b/ui/src/chat/ChatInput/ChatInput.tsx
@@ -287,7 +287,7 @@ export default function ChatInput({
    */
   const messageEditor = useMessageEditor({
     whom: id,
-    content: draft,
+    content: '',
     uploadKey,
     placeholder: 'Message',
     allowMentions: true,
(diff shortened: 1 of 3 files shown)


jamesacklin:
The only slightly odd behavior is that the message input retains focus across DMs/channels if you have a draft set in one of them. Doesn't matter if you unfocus after drafting the message. Drafts do clear after sending, though.

patosullivan (approved):
Looks good

PR #1990, opened 2023-02-22
diary: breaking up paragraphs correctly

### LINKED ISSUE #1959: Notebook removes spacing between paragraphs on save
When I'm editing a Notebook post, the formatting doesn't render spaces between paragraphs. When I save the post, the spacing is removed. When I go to edit the post, the spaces appear to be saved. See screenshots of saved and editing state.

**Expected behavior**
I expect formatting to be saved on publication.

**Additional context:**
- Using a Tlon hosted planet.
- Safari on MacOS

editing
published state

arthyn:
This parses our breaks so that they form paragraphs fixing #1959. We could probably refine this further with code blocks and blockquotes (these should probably not be part of the inline type anyway).

Fixes #1959


tomholford (approved):
always happy to see a test! LGTM

PR #1992, opened 2023-02-22
channels: avoid ungodly unread backlog

arthyn:
OTT I don't think we have this issue recorded, but joining any channel with a large backlog gives an ungodly amount of unreads. This sanely sets the last read time to the join time.


tomholford (approved):
subtly brilliant, very nice. LGTM

patosullivan (approved):
lgtm!

PR #1995, opened 2023-02-23
channel admin overhaul

### LINKED ISSUE #1770: groups: updates + improvements to channel management ux
At the moment, group "channel management" is an interface located within a group's 1) "admin settings" interface.

Unfortunately, this management UX has quite a bit of overlap with the 2) "All Channels" index, available to all members of a group, and the 3) "Channels" screen in a group's "Info + Members" interface for non-admins. These three distinct interfaces result in way too much surface area for what should be a relatively compact area of the groups interface (a list of channels).

The following enhancement proposal streamlines all of the channel-viewing/management interfaces into a single location, the "All Channels" list available in the group's sidebar to admins and members alike. Within this singular interface, channel options should be scoped appropriately to role. In reducing the complexity of this interface, I hope to make it a lot easier to reason about (and diagnose issues with).

Here's an overview of some suggested changes, which have summaries/explanations on the cards to the left of each row:

Channel Management Updates

In a nutshell, the following experiences are listed in the illustrations above:

1. Joining/Leaving channels
2. Adding a new channel, with a new sidebar shortcut and from within the channel list itself
3. Adding a new section, from the channel list itself
4. Editing a channel's information/membership
5. Role-specific interfaces + distinctions
6. Changes to "group settings" and "group info + members" interfaces available to admins/members alike, as a result of the prior updates (tl;dr, removing the channel viewing tabs from these respective interfaces)

Further details are located in Figma:
https://www.figma.com/file/luXjuX8fdV2mQrfqCQ4Cr2/2023-Q1-Groups?node-id=2438%3A10911&t=VwfUADS5OwY599zx-1

Please feel free to pepper the Figma, or this issue, with comments, issues, concerns, and rebuttals.

See #1618 for related concerns involving the performance of some of the interfaces described here. As usual, we should ensure that our interfaces are as reactive as possible, and act in a deterministic manner (if I rearrange a section, it should just work immediately, without needing to wait for a response from the remote urbit, if I click a button it should simply change state and process the interaction async, etc.)

patosullivan:
Fixes #1770 by reusing the admin channel list as the universal channel list.

https://user-images.githubusercontent.com/1221094/220970186-0e5eb293-d962-4e37-8f14-3056d337f7ed.mov


patosullivan:
Forgot to record the "plus button" action opening the "Add channel" dialog, but that is in there too.

urcades (approved):
This is gorgeous, it's looking great 🫡

One question: Now that this interface exists, are we deprecating this channels list buried in the "Groups Settings" interface?:

Screenshot 2023-02-23 at 11 45 24 AM

One of the major reasons we decided that integrating channel admin/management into "All Channels" was a good idea was because we wanted to cease the duplication of the "list of channels" interface, and make "All Channels" the one place to go to view/manage channels regardless of whether one was an admin or member.

patosullivan:
> This is gorgeous, it's looking great 🫡
> 
> One question: Now that this interface exists, are we deprecating this "channel management list" buried in the "Groups Settings" interface?:
> 
> Screenshot 2023-02-23 at 11 45 24 AM
> 
> One of the major reasons we decided that integrating channel admin/management into "All Channels" was a good idea was because we wanted to cease the duplication of the "list of channels" interface, and make "All Channels" the one place to go to view/manage channels regardless of whether one was an admin or member.

Yes, I removed it from the group management screen.

ui/src/groups/ChannelsList/ChannelManagerHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+export default function ChannelManagerHeader({
+  addSection,
+}: ChannelManagerHeaderProps) {
+  const location = useLocation();
+  const flag = useRouteGroup();
+  const isAdmin = useAmAdmin(flag);
+  const isMobile = useIsMobile();
+
+  return (
+    <div
```
arthyn, line 21: we should prefer using tailwind for responsive stuff where we can

arthyn (approved):
lgtm, small nit about responsive classes

ui/src/groups/ChannelsList/ChannelManagerHeader.tsx
patosullivan: Agreed. Worth going through these changes to replace all of the isMobile references?

patosullivan: Fixed.

jamesacklin:
Super minor quirk with renaming a section:
1. Click rename
2. Set a name
3. Press Enter
4. See name set
5. Click rename again
6. Set a name
7. Click _outside the field_
8. Name gets set to "New Section"

Would expect that this second new name persists on field blur, not just Enter-press.

rename

jamesacklin:
Creating a new channel and pressing Enter in the field name sends me to a blank route. I would expect that this would close the modal and keep me on the same page.

blank

jamesacklin:
Please match the border radii between these two islands.

image

jamesacklin:
NB for anyone following along at home: the ever-present reordering glitch in v2.7.0 and earlier only occurs if the bandwidth is severely limited and you intentionally and aggressively reshuffle channels. (We were updating much more than we needed to with each ordering change.)

reorder-glitch

arthyn (approved):
lgtm!

patosullivan:
Hey @jamesacklin, I just pushed some commits that I think address all of your concerns.

jamesacklin:
Amazing, thank you @patosullivan. Two last quirks.

Renaming a section shuffles it to the beginning of the list. I'd expect the section to stay in place.

rename-reorder

Changing permissions on a channel appears to never update after the modal closes. A refresh fixes this. Changing the name or default-join status behaves as expected.

perms

PR #1998, opened 2023-02-23
Group refs: add fixed with, make text small, wrap it

patosullivan:
- Adds a fixed with to group refs of 250px
- Sets font-size to text-sm
- Puts "by ShipName" on a separate line.
- Hides overflowing text with ellipsis.
- Makes it so the button text doesn't wrap.

image

### DIFF
--- a/ui/src/components/References/GroupReference.tsx
+++ b/ui/src/components/References/GroupReference.tsx
@@ -60,7 +60,7 @@ export default function GroupReference({
   return (
     <div
       className={cn(
-        'not-prose relative flex items-center rounded-lg  text-base transition-colors hover:border-gray-100 hover:bg-white group-one-hover:border-gray-100 group-one-hover:bg-white',
+        'not-prose relative flex max-w-[250px] items-center rounded-lg bg-white text-base transition-colors hover:border-gray-100 hover:bg-white group-one-hover:border-gray-100 group-one-hover:bg-white',
         {
           'border-2 border-gray-50': !plain,
         }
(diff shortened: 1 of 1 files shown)


ui/src/components/References/GroupReference.tsx
```diff
@@ -60,7 +60,7 @@ export default function GroupReference({
   return (
     <div
       className={cn(
-        'not-prose relative flex items-center rounded-lg  text-base transition-colors hover:border-gray-100 hover:bg-white group-one-hover:border-gray-100 group-one-hover:bg-white',
+        'not-prose relative flex max-w-[250px] items-center rounded-lg bg-white text-base transition-colors hover:border-gray-100 hover:bg-white group-one-hover:border-gray-100 group-one-hover:bg-white',
```
arthyn, line 63: would 600 better match our other max widths for chat stuff? can't remember, but it does look a little narrow

jamesacklin: Why a max-width at all?

arthyn: looks bad at full-width imo

patosullivan: Yeah, agreed, looks bad at full width. I feel like 600 is also too huge.

patosullivan: I'll add that this width makes these look the same when rendered in both gallery items and chat.

arthyn (approved):
lgtm

PR #2001, opened 2023-02-23
New channel manager fixes

### LINKED ISSUE #2000: groups: new channel manager bugs
Renaming a section shuffles it to the beginning of the list. I'd expect the section to stay in place.

rename-reorder

Changing permissions on a channel appears to never update after the modal closes. A refresh fixes this. Changing the name or default-join status behaves as expected.

perms

_Originally posted by @jamesacklin in https://github.com/tloncorp/landscape-apps/issues/1995#issuecomment-1442331768_

patosullivan:
Fixes #2000 

We had some weird logic in the SectionNameEditInput component that was causing the first issue. 

We were also re-adding channels to a section each time we updated the title, which appears to be unnecessary (maybe a previous hack to deal with the state not updating?), which was causing the channels to reorder when renaming a section.

The channel permissions state not updating was due to the relevant calls not being awaited.

### DIFF
--- a/ui/src/channels/EditChannelForm.tsx
+++ b/ui/src/channels/EditChannelForm.tsx
@@ -93,9 +93,9 @@ export default function EditChannelForm({
           : useDiaryState.getState();
 
       if (privacy !== 'public') {
-        chState.addSects(channelFlag, ['admin']);
+        await chState.addSects(channelFlag, ['admin']);
       } else {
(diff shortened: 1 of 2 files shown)


jamesacklin:
Amazing, thank you. I'll test when I get back to my desk.

Great work!

arthyn (approved):
lgtm

jamesacklin:
Perfect, thanks again.

PR #2003, opened 2023-02-24
Switch to preact

patosullivan:
Took ten minutes. Tests are broken and there's an issue with backgroundLocation/modals.


arthyn:
let's revisit later

PR #2005, opened 2023-02-25
chat: thread status cosmetics

jamesacklin:
Fixes some alignment issues in the chat thread indicator that have been annoying me for some time.

1. Properly aligns the left edge of the first avatar in series with the emoji/message content above.
2. Lightly overlaps sigils to conserve horizontal real estate. The outline masking effect is sensitive to hovered/active/linked states of both the thread indicator and the parent message such that the border always matches the background color behind it.
3. Hides the "Last reply " prefix text on small screens and correctly capitalizes the relative date string in its absence.
4. Matches the unread link color to the presence of the unread dot; falls back to normal black. This should help with spotting unread threads in a long chat scroll.

Unreads:
image

Read:
image


patosullivan (approved):
Lgtm

ui/src/styles/components.css
```diff
@@ -145,3 +145,13 @@
...
+
+/* Conditionally style reply avatars */
+
+.is-active .reply-avatar {
+  @apply outline-gray-50;
+}
+
+.bg-blue-softer .reply-avatar {
+  @apply outline-blue-50;
+}
```
arthyn, line 157: Doesn't really follow the way we've been conditionally applying classes I would move into JS

urcades (changes requested):
One tiny change I'd suggest we make is to move the "unread dot" over to a position that's a little more representative of how we surface the dot elsewhere (right-aligned to its container + object of attention):

Screenshot 2023-02-27 at 12 15 51 PM

This also has the benefit of keeping the "new replies" row visually left-aligned with the text it's coupled to.

jamesacklin:
@urcades:

With unreads:

image

At rest:

image

PR #2014, opened 2023-02-27
Add missing awaits and/or loader states omnibus

### LINKED ISSUE #1999: groups: immediate feedback for all server interactions
From ~natnex-ronret: 

Image

From ~rosmut-hodpub:

> i've noted this too. little lags like this make a big impression

To wit, any button that triggers a server-side action (sending a poke, sawing a rope, whatever) should be accompanied by (A) at bare minimum, an `await` function and (B) a temporary flag set on the button that indicates the user clicked it _even if the server-side action hasn't started yet_. This sleight of hand will help us mask slow connections and make Groups _appear_ faster without further optimizations.

patosullivan:
Fixes #1999


patosullivan:
> changes look good, in the places where you just added async/await and not a status were those spots just already responding immediately?

They were typically places where we wouldn't have had a "loading" state, like the action should just happen in the background.

PR #2018, opened 2023-02-28
fix loading of curios/notes from unjoined channels

### LINKED ISSUE #1976: chat: gallery ref doesn't link through, has impermissible actions
Gallery references in a Chat don't link through to the detail view itself and presents the user with "Edit" and "Delete" options. 

**To Reproduce**
1. Go to ~natnex-ronret/door-link
2. Find ~fabled-faster's post where he said "feels door" and links to a beautiful green wash in one of his Storage Galleries.
3. Click the ref. Note that the navigation works, there's a loading spinner, and then the item fails to load.
4. Also observe that the ref has Edit and Delete options. Ed has expressly made his Storage group view-only for all members.

**Expected behavior**
The Gallery item should load and should not display additional actions if I cannot take them. 

**Additional context:**
Groups v2.7.0 pre-release, reproducible in all browsers

patosullivan:
Fixes #1976, which includes two separate bugs:

1) Users could see an edit menu in the ref, this removes that.
2) Clicking on a gallery item ref for a gallery you haven't joined would take you to a heap detail page where the curio never loads. This PR auto-joins the user to the channel before attempting to load the curio.

This also fixes two other bugs:
1) A similar bug to the second bug listed above, but for notes (can't find an issue for this one).
2) A text overflow issue with author name/time in the heap detail sidebar.

To test:
Using two test ships, create one group on each ship. Have one ship in both groups initially. The other ship will create a group that the first ship is not a member of. In the first group, add a chat channel. In the second group, add a notebook and gallery channel and some content. Post refs from those notebook/gallery channels in the chat channel on the first group from the second ship and click through on the first ship.

Edit:
Also fixed the package lock, it appears that a dependency was added to master using an older version of npm.


arthyn (approved):
lgtm!

PR #2025, opened 2023-02-28
groups: Auto-join all channels

### LINKED ISSUE #1997: groups: auto-join all channels on group join
Contrary to our previous "calm" thinking, new users are confused when they land in a group with no default channels set and are required to opt-in to channel membership.

1. Join all new members of a group to all permissible channels.
2. Join all members of a group to a new channel when it is created.
3. Remove the "Set as Default" checkbox in the New Channel / Edit Channel dialog. This option is now redundant.

There is no need to change the channel list sidebar at the moment.

patosullivan:
Fixes #1997 

- All new members of a group will join all permissible channels.
- All members of a group will join a new channel when it is created.
- "Set as Default" checkbox in the New Channel / Edit Channel dialog has been removed
- "Default" tags on channel list have been removed.

Test by using two ships. Create a new group on one ship with some channels, send an invite to the other ship. Join from the other ship, all channels should be joined. Go back to the first ship, make a new channel. Go to the second ship, it should now have that channel.


desk/app/groups.hoon
```diff
@@ -759,7 +759,7 @@
         %+  skim
           ~(tap by channels.group)
         |=  [nes=nest:g =channel:g]
-        join.channel
+        %.y
```
arthyn, line 762: I think this can just be `~(tap by channels.group)` now

ui/src/channels/NewChannel/NewChannelForm.tsx
```diff
@@ -125,11 +125,9 @@ export default function NewChannelForm() {
         }
       }
 
-      if (values.join === true) {
-        await useGroupState
-          .getState()
-          .setChannelJoin(groupFlag, newChannelNest, true);
-      }
+      await useGroupState
```
arthyn, line 128: I think we can just remove this now

arthyn (commented):
let's pair on this tomorrow and do the new channel joins from the backend instead

arthyn (approved):
lgtm!

PR #2030, opened 2023-03-01
clubs: only send to part of list so we actually are gossiping

arthyn:
This should help with the explosion of networking activity in large group dms, but will not fix all the issues we're currently seeing with them. Definitely up to any better ways of doing this until we can fix the overarching issues. This is just to stop the bleeding.

Need to test this with a good amount of ships, maybe we can get some test ones in hosting plus some personal moons? @jamesacklin


arthyn:
After talking with @philipcmonk I think it's best to just implement something to keep track of gossip we've heard which shouldn't take too long.

PR #2039, opened 2023-03-01
groups: updates to groups sidebar element spacing

urcades:
I've updated a few of the styles contained in the `GroupSidebar` component:

- Decreased the top margin around the group header image block for equidistant spacing within the sidebar
- Removed extraneous padding from group nav cluster items
- Decreased the spacing between channel elements in the sidebar to better match the 2px border stroke we generally use

Visual summary of changes:

Updates


urcades:
While inspecting the list of Groups at the app's root to try similar edits, I couldn't manage to find where we style the child elements (the Group Rows) contained in the list. I tried editing a few spots, to no avail.

ui/src/components/Sidebar/GroupList.tsx
```diff
@@ -63,7 +63,7 @@ export default function GroupList({
       computeItemKey={(_i, [flag]) => flag}
       itemContent={itemContent}
       components={components}
-      className="h-full w-full overflow-x-hidden"
+      className="h-full w-full space-y-0.5 overflow-x-hidden"
```
urcades, line 66: @patosullivan @arthyn I'm attempting to apply this inter-element styling I applied to the channel list in this PR, but it doesn't appear to function/work as expected — any idea what might be going on here?

urcades: I'm specifically'referring to adding `space-y-0.5` to the ClassName here

patosullivan: It's because that utility is intended to be used with flex, but we aren't using flex here. You could add it to the classes on the Wrapper component within the SidebarItem component. I'll push a pr to show what I mean.

patosullivan: I also included an update to the pre-commit hook that will update the snapshots before commit so you don't see these build errors in CI. Something we should have done a long time ago tbh.

patosullivan (approved):
As long as the change I made looks right to you, everything here lgtm.

jamesacklin (approved):
Love it.

@urcades Pull Pat's tweaks and hit that merge button when you're ready.

Issue #1937, opened 2023-02-14
leap: can't search for and navigate to a group DM.


patosullivan:
OTT

PR #2047, opened 2023-03-02
leap: add missing leap results

### LINKED ISSUE #1936: leap: can't search for DMs without contact store
If your ship does not have a contact store entry for a ship, you can't navigate to a DM with that ship (even if you've started one with them). This effects all new ships (which are currently effected by https://github.com/tloncorp/landscape-apps/issues/1305).

patosullivan:
Fixes #1936 and #1937

### DIFF
--- a/ui/src/state/chat/chat.ts
+++ b/ui/src/state/chat/chat.ts
@@ -986,6 +986,17 @@ export function useMultiDms() {
   return dms;
 }
 
+const selDms = (s: ChatState) => s.dms;
+export function useDms() {
+  const dms = useChatState(selDms);
+
+  useEffect(() => {
+    useChatState.getState().fetchDms();
+  }, []);
+
+  return dms;
+}
+
 export function useMultiDm(id: string): Club | undefined {
   const multiDm = useChatState(useCallback((s) => s.multiDms[id], [id]));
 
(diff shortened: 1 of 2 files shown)


ui/src/state/chat/chat.ts
```diff
@@ -986,6 +986,17 @@ export function useMultiDms() {
   return dms;
 }
 
+const selDms = (s: ChatState) => s.dms;
+export function useDms() {
+  const dms = useChatState(selDms);
+
+  useEffect(() => {
+    useChatState.getState().fetchDms();
```
arthyn, line 994: we should have already fetched DMs on load or seen any come in on subscriptions. I think the `useMultiDms` above was from a time when we didn't do that on load for multi dms, hence the fetch, but we do now so it just needs to be corrected.

arthyn (changes requested):
lgtm, just one thing to correct

ui/src/state/chat/chat.ts
patosullivan: It won't load if we're in groups though, correct? So we'd still need a fetch in both cases if we're in groups.

patosullivan: pushing up a PR where we check if DMs are already in state before returning.

arthyn (approved):
lgtm

PR #2051, opened 2023-03-02
clubs: changing stop mechanism to be 'heard'

arthyn:
This changes our "stop" mechanism from echoes to already heard. So each piece of gossip travels with a unique identifier and if we've already heard that uid then we just ignore it. If we haven't heard it, we continue to gossip it to everyone and process like before. So at most we should hear `n` facts, where `n` is the number of group members, because the initial sender sends one, which everyone gossips to everyone else, so we should hear one per person. One of these will get recorded in our `heard` set and the rest ignored.

Clarifying because this was unclear to me: the reason we initially send to everyone and also gossip to everyone is that this ensures that we get close to proper ordering in the case of sender/receiver ships going offline in tandem and then coming online out of order. If the receiver comes back online and hasn't heard the senders poke yet (because they're offline) the gossipers should conceivably poke the gossip first before poking any other messages that they would have sent as well, conserving order.

@jamesacklin flagging that this is breaking change we should communicate to third party devs.


arthyn:
Notably this breaks connection between people on the old version of clubs vs the new. Started to go down a path to making it backwards compatible, but I wasn't sure how to mesh the two methods.

arthyn:
Pinging for review, one note here is that we talked about trying to make these backwards compatible by potentially hashing the contents but I think that only works forward aka old trying to talk to new. If we left echoes I suppose it could work for old as well?

midsum-salrux (approved):
LGTM

arthyn:
working on backwards compat, just making sure we don't merge yet

PR #2053, opened 2023-03-02
chat: move ChatEmbedContent container to inside component

### LINKED ISSUE #2052: chat: URLs appear on their own line
Image

jamesacklin:
fixes #2052

Keeps non-oEmbed URLs as inline:
image

Respects embeds:
image


patosullivan (approved):
Makes sense, thanks!

PR #2054, opened 2023-03-02
notes: remove kebab menu + share button

### LINKED ISSUE #2038: "Edit Note" for notebook post produces blank screen
Attempts to edit a new notebook entry (Landscape Pilot Program) in the group, ETHDenver, result in a blank screen. I tried four times with the same result. Refresh has no effect, can only exit the page/go back. 

The issue doesn't appear to be affected by browser (used Arc + Chrome).

Screenshot 2023-03-01 at 9 36 06 AM

patosullivan:
Fixes #2038


arthyn (approved):
lgtm

PR #2056, opened 2023-03-02
groups: set default group sort to 'Arranged'

### LINKED ISSUE #2044: groups: default channel sidebar ordering to "Arranged"
From a group onboarding:

> we made some sections in the group but realized these section might not show up with different sort options (such as A-Z) and only show for "arranged". Is there a way to set the group settings as admin so that it shows the "arranged" view to new people by default?

patosullivan:
fixes #2044


jamesacklin:
Checked this out. I joined ~fabled-faster/structure and:

1. I joined every channel (thank you) and
2. my sidebar was set to Arranged.

Works for me.

jamesacklin:
@arthyn look good to you? Fast-tracking a ~binnec/~wannec build

arthyn (approved):
lgtm, logic here is complex, but let's run with it

PR #2058, opened 2023-03-03
heap: don't check for embed if we know it's not

### LINKED ISSUE #2057: heap: don't check known things for embeds

patosullivan:
fixes #2057 

makes galleries load faster.

### DIFF
--- a/ui/src/heap/HeapBlock.tsx
+++ b/ui/src/heap/HeapBlock.tsx
@@ -239,8 +239,13 @@ export default function HeapBlock({
   useEffect(() => {
     const getOembed = async () => {
       if (isValidUrl(url) && !notEmbed) {
-        const oembed = await useEmbedState.getState().getEmbed(url);
-        setEmbed(oembed);
+        try {
+          const oembed = await useEmbedState.getState().getEmbed(url);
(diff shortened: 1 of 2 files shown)


ui/src/heap/HeapBlock.tsx
```diff
@@ -234,18 +234,19 @@ export default function HeapBlock({
   const flag = useRouteGroup();
   const isAdmin = useAmAdmin(flag);
   const canEdit = asRef ? false : isAdmin || window.our === curio.heart.author;
+  const notEmbed = isImage && isAudio && isText && isComment;
 
   useEffect(() => {
     const getOembed = async () => {
-      if (isValidUrl(url)) {
+      if (isValidUrl(url) && !notEmbed) {
         const oembed = await useEmbedState.getState().getEmbed(url);
```
arthyn, line 242: Can we surround in a try/catch and set embed to null if it errors?

arthyn (changes requested):
need to catch any errors here so we don't show spinner forever and also apply same logic to HeapRow

arthyn (approved):
lgtm

PR #2064, opened 2023-03-03
refs: improve reference perf

### LINKED ISSUE #1970: chat: reduce channel checking for refs
Currently these scry on every render, across all instances of a ref for the same channel. We should lock this up so that we only scry  less frequently across all instances of the same channel.

patosullivan:
Fixes #1970 

We don't need to initialize the relevant channel, we were still doing that for curios/notes.

I also memoized all of the reference components.


arthyn (approved):
lgtm

PR #2067, opened 2023-03-03
chat: add global state for pickers, lightboxes, modals

### LINKED ISSUE #2065: groups + chat: popups/menus/dropdowns closing on background events
Something is causing top-level re-renders, which have the annoying effect of closing something you're looking at / inspecting / attempting to do.

To reproduce:
- Join a channel in a moderately-active group
- Open an image in a lightbox, or open the emoji picker
- Observe it closing as a new event comes in (a message in another channel, a message in the same channel, etc).

patosullivan:
Fixes #2065 

Unfortunately we need to prop-drill the writIds down into the image content/embeds as well in order to have a persistent ID.


arthyn (approved):
sick lgtm

PR #2068, opened 2023-03-03
More mop functions: top, bot, and bat

### LINKED ISSUE #2042: chat: change data structure to allow better querying
We currently use a `mop` to store chats, notes, and curios. If you look at the way we've implemented the `older`/`newer`/`newest` scries in `dm.hoon`, `notes.hoon`, `curios.hoon` it's clear that we're not using these efficiently because we're missing some more ergonomically useful functions.

For chats this is especially bad because we have huge logs. This means when we reverse the log it takes forever. Also to do the "click to view last unread message" correctly we need to be able to arbitrarily jump back to a point in time and provide the surrounding context we can get one direction of messages using `tab` but not the preceding direction.

For this to work we need a few extra functions on mops:
- top/bottom X items
- a function similar to tab except going in the opposite direction

Once we have these we should replace the implementations of the scries mentioned above using these and should see better performance for all of them.

midsum-salrux:
Fixes #2042 

Example usage:

```
/+  mop-extensions
=/  longer  |=([a=tape b=tape] (gth (lent a) (lent b)))
=/  mymop  ((mop tape @) longer)
=/  myon   ((on tape @) longer)
=/  mope  ((mop-extensions tape @) longer)
=/  letters  (put:myon (put:myon (put:myon (put:myon (put:myon *foo "a" 1) "bb" 2) "ccc" 3) "dddd" 4) "eeeeee" 5)

:*  (top:mope letters 4)
    (bot:mope letters 4)
    (bat:mope letters `"bb" 2)
==
```

### DIFF
--- a/desk/lib/mop-extensions.hoon
+++ b/desk/lib/mop-extensions.hoon
@@ -1,0 +1,77 @@
+|*  [key=mold val=mold]
+=>  |%
+    +$  item  [key=key val=val]
+    --
+~%  %comp  +>+  ~
+|=  compare=$-([key key] ?)
+~%  %core    +  ~
+|%
+::  +bat: tabulate a subset excluding start element with a max count (backwards)
+::
+++  bat
+  |=  [a=(tree item) b=(unit key) c=@]
+  ^-  (list item)
+  |^
+  e:(tabulate (del-span a b) b c)
+  ::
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm!!!

Issue #1375, opened 2022-12-09
groups: improve performance of reloading groups and talk browser apps


zalberico:
**Describe the bug, optionally with screenshots**
Refreshing talk takes 22 seconds, groups.json is 2.7MB and takes 10 seconds to render, which is at least 10x larger than it should be

@philipcmonk mentioned there's low hanging fruit here (can you add details @philipcmonk)

zalberico:
Now that groups/talk work pretty well - this is the main issue that makes it hard to use.

I often have to manually refresh to trust that what I'm seeing is the current state. Tapping into channels often takes seconds to load. I can't trust that the state will update without another manual refresh (on both mobile and desktop).

arthyn:
Can we maybe clarify more in this issue? There's a bit more loading work we can do on the front-end, but it's pretty fast right now on reload especially if you have localstorage filled. Is this more to do with loading particular channels?

patosullivan:
Maybe this is the `bap` issue? If you don't regularly pack)meld then initializing the writ store for a channel can take several seconds (at least).

arthyn:
Likely yeah

jamesacklin:
Fixed by #2100

PR #2069, opened 2023-03-03
channels: improve querying across the board

arthyn:
Makes much better use of the data structures we had in place, the beginning of fixes for #1375. This should significantly speed up all channel loads.

Thanks to @midsum-salrux for his mop work in #2042


midsum-salrux (approved):
LGTM

PR #2070, opened 2023-03-04
ops: restore mop work post release

arthyn:
This restores what we reverted


jamesacklin:
Let's avoid merging in the case we investigate #2071 and deploy a solution before this is ready to go.

midsum-salrux (approved):
LGTM

PR #2074, opened 2023-03-06
Prevent unnecessary calls to noembed on chat embed

patosullivan:
Also, don't make calls to noembed on heap items if remote content has been disabled.

No issue created for this one, just something that occurred to me this morning.


jamesacklin (approved):
Makes sense, thanks

PR #2080, opened 2023-03-06
revert ref changes/remove container queries - fixes safari scroll bug

patosullivan:
I don't see an issue for this, but this fixes the safari scroll bug (scroller would jump up by one message repeatedly).

Appears to have been caused by the use of container queries in references. Removing the container queries made the scroller behave as expected. 

To test: run this branch locally and open up safari to any channel with a reference loaded into the scroller, notice that it doesn't jump.


arthyn (approved):
lgtm, let's deploy this for now. I think we can redo these changes with regular media queries and maybe some props that the greater calling context can pass in to tell it to be smaller? or we can find some other way maybe.

PR #2083, opened 2023-03-06
Prevent excess scries for DMs on ships without DMs/MultiDMs

patosullivan:
Avoid object checking in useEffect dep array


arthyn (approved):
lgtm

PR #2085, opened 2023-03-07
Add cancel button for stuck group joins

patosullivan:
There isn't an issue for this afaik, but romi saw this and I've actually seen this on one of my dev moons. If something is wrong with the host ship for the group you're trying to join you could get into a state where you're just stuck with the pending group item in your sidebar.

This adds a cancel button so you can get rid of the pending group item.

### DIFF
--- a/ui/src/components/Sidebar/GangItem.tsx
+++ b/ui/src/components/Sidebar/GangItem.tsx
@@ -79,6 +79,21 @@ export default function GangItem(props: { flag: string }) {
                 It may take a few minutes depending on the host&apos;s and your
                 connection.
               </span>
+              <button
+                className="small-button bg-gray-50 text-gray-800"
+                onClick={handleCancel}
+              >
+                {isPending ? (
+                  <>
+                    Canceling...
+                    <LoadingSpinner className="h-5 w-4" />
+                  </>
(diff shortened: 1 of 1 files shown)


ui/src/components/Sidebar/GangItem.tsx
```diff
@@ -79,6 +79,21 @@ export default function GangItem(props: { flag: string }) {
                 It may take a few minutes depending on the host&apos;s and your
                 connection.
               </span>
+              <button
+                className="small-button bg-gray-50 text-gray-800"
+                onClick={handleCancel}
+              >
+                {isPending ? (
+                  <>
+                    Canceling...
```
jamesacklin, line 88: Could probably yank this, just use the spinner by itself

ui/src/components/Sidebar/GangItem.tsx
```diff
@@ -79,6 +79,21 @@ export default function GangItem(props: { flag: string }) {
...
+                    <LoadingSpinner className="h-5 w-4" />
+                  </>
+                ) : requested ? (
+                  'Cancel Request'
```
jamesacklin, line 92: ...and could probably also shorten to "Cancel" in both cases

Issue #2077, opened 2023-03-06
groups: less aggressive disconnect behavior


jamesacklin:
The problem with knowing something is wrong is that now you know it's wrong.

1. Explore a less intrusive treatment for the reconnecting state. The butter bar occupies screen real estate by pushing the rest of the view downward and may be the cause of undue alarm. We could probably keep this for the "truly disconnected, please manually refresh" state.
2. We disable the Send button in the chat message input when disconnected. Let users send the message anyway and make use of the "sent from the front-end, but not received on-ship yet" pending state (light gray arrows).
3. The banner also steals focus from the chat input. Let users continue typing away even in the reconnecting state.

urcades:
I have a rough proposal for point 1.) above:

On desktop/larger-sized viewports, we can consider situating a loader/refreshing/async-load indicator within the sidebar, in a relatively unobtrusive (yet visible) location:

<img width="779" alt="Screenshot 2023-03-06 at 2 58 23 PM" src="https://user-images.githubusercontent.com/1195363/223218030-77ea0938-17cb-4f37-a680-acabd418d6ad.png">

This placement works across most contexts, with a tiny exception to potentially obscuring the dropdown arrow placement of the "Group Actions" dropdown, which is present under "normal" circumstances:

https://user-images.githubusercontent.com/1195363/223218634-dc3ad89e-00cd-45f2-8924-1f36d4f61f85.mov

On mobile viewports, or within any mobile applications we develop, we can situate the loading indicator in any given header context in an unobtrusive location. Here's one example:

<img width="615" alt="Screenshot 2023-03-06 at 3 00 45 PM" src="https://user-images.githubusercontent.com/1195363/223218811-d2a77edc-1f93-4f6c-81aa-6ab33512cddc.png">

A valid alternative to consider here is simply keeping the placement of the loading indicator in the given channel's header, so that there's only ever one spot to pay attention to the app's connection state:

<img width="883" alt="Screenshot 2023-03-06 at 3 08 27 PM" src="https://user-images.githubusercontent.com/1195363/223219292-124e2492-90f4-42cb-b180-3fd408e75251.png">

I'm not entirely opposed to this direction if the development lift is significantly lessened.

arthyn:
I think the spinner is good for a reconnecting state, but for disconnected I think we need something more static.

PR #2091, opened 2023-03-07
app: less frequent disconnect checking

arthyn:
Should address 2+3 in #2077, also adjusts the criteria for throwing disconnected by making it require 2 timeouts. The first timeout will add to the error count in case we've seen other errors so has the possibility of throwing the disconnected state because this should be a pretty good indicator that we're actually disconnected.

notably doesn't include the fix for double renders caused by reconnecting, will follow up

### DIFF
--- a/ui/src/chat/ChatInput/ChatInput.tsx
+++ b/ui/src/chat/ChatInput/ChatInput.tsx
@@ -19,7 +19,6 @@ import ShipName from '@/components/ShipName';
 import X16Icon from '@/components/icons/X16Icon';
 import {
   fetchChatBlocks,
-  useChatBlocks,
   useChatInfo,
   useChatStore,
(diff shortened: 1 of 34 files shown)


ui/src/chat/ChatInput/ChatInput.tsx
```diff
@@ -293,13 +290,10 @@ export default function ChatInput({
     allowMentions: true,
     onEnter: useCallback(
       ({ editor }) => {
-        if (subscription === 'connected') {
-          onSubmit(editor);
-          return true;
-        }
-        return false;
+        onSubmit(editor);
```
jamesacklin, line 293: There we have it. Thank you

jamesacklin (approved):

Issue #1913, opened 2023-02-10
Group load is slow. Caused spontaneous refresh.


marisarow:
**Describe the bug, optionally with screenshots**
First load of a group is very slow. When he opened Tlon Public, gallery and notebooks weren't loading content, but were showing the ability to add content (which is wrong based on our permissions). Eventually, the window actually hard refreshed itself and then galleries and notebooks were showing appropriately (with no ability to add content).

jamesacklin:
Should be fixed with #1902, but keeping around for a test case.

jamesacklin:
Fixed by #2100

PR #2100, opened 2023-03-09
perf: improve initial load

arthyn:
This is an effort to reduce our initial load times as per #1375 and #1913. 

This collapses our slew of scries into a single `init` scry for both groups and talk. In investigating this with @joemfb we found that pre-building these scry marks and also minimizing the amount of inter-agent scries was important so that's why you see lots of `init` scries in all the agents.

In addition, this PR adds a request scheduler for our initial run of requests. This is an attempt to make sure that we prioritize the main content of the page such as chat messages, notes, galleries, etc. We defer loading subscriptions and migration information to later in the queue for example. This could probably be improved/simplified.

There's still much we could do here, but this got us some pretty decent improvements.
I've pulled some numbers just from dev builds so that we can get a sense of how much of an improvement we're getting:

## new
**full load of home page**
- groups: 4.15s
- talk: 3.2s

**time to chat loading**
- groups: 2.1s
- talk: 2.4s

## old
**full load of home page**
- groups: 6.08s
- talk: 4.84s

**time to chat loading**
- groups: 5.14s
- talk: 4.6s


patosullivan (approved):
Frontend changes look great. Backend changes compile and the app loads *much* faster locally.

jamesacklin:
~150% faster to load the app, and ~200% faster to load a chat. I'll take it!

As soon as we merge I'll throw this on canary so we can live in the new world for a bit.

joemfb (approved):
this will do

(further gains could be had by adding new scry handlers so as to combine and reduce the number of inter-agent scries. also, ford cache hits for already-built marks seem to be more expensive then they should be.)

jamesacklin:
Fixes #1913 
Fixes #1375

PR #2101, opened 2023-03-09
updating padding values across chat message types

### LINKED ISSUE #1675: talk: visual refactor + improvements to "reference" message types
### One-line Summary: 

The "reference" component needs to be polished – while we've implemented its intended _functionality_, we haven't implemented it _precisely_ "to spec", which should help with its functional legibility.

### Issues

As of writing, the "reference" messaging component looks like this:

References

While references are more or less functioning in an MVP manner, there are a few key optimizations and simplifications we can implement to make them feel a lot more "solid" and meaningful as block-quote-adjacent structures.

The current issues with the component:

Screenshot 2023-01-10 at 1 50 25 PM

1. zero margin between an inaccessible/nonexistent reference and the following message (edit: this "zero margin" issue actually seems to affect any reference that's coupled with nominal message text, when attaching a reference to the message input)
2. an empty (and useless) metadata row that looks buggy/unintentional

Screenshot 2023-01-10 at 1 50 28 PM

3. The metadata row, when filled with data, is the same size as normal author rows, which causes them to "fight" for visual dominance, which makes the message feed harder to parse/read
4. there's a huge margin between a fully-rendered reference and a "reply" row appended to it

Screenshot 2023-01-10 at 1 50 32 PM

5. There are two "hover" states for a reference row, which decrease the feeling of 'stability' in the interface

With these issues laid out, I'll proceed to write some implementation polish suggestions in a following post.

urcades:
Fixes #1675 

I'm taking some time to comb through a number of components that have been overdue for some tweaking of minor values.

Some edits include:

- Increase the space between chat message authors in order to match design spec
- Updating padding values for DateDivider to match design spec
- Minor padding adjustments to various reference components: NoteReference, ReferenceBar, WritBaseReference


ui/src/chat/ChatMessage/Author.tsx
```diff
@@ -89,6 +89,9 @@ export default function Author({
       <div onClick={handleProfileClick} className="shrink-0">
         <Avatar
           ship={ship}
+          // The conditional here appears to have scoping
+          // which affects the author avatar's size in both "chats themselves"
+          // as well as within ReferenceRow components.
           size={isReply || isRef ? 'xxs' : 'xs'}
```
urcades, line 95: So I've been trying to figure out how to ensure author/op author components in references are completely uniform. Right now Talk message references render their author avatars as specified, while both Notebook Post + Gallery Item references render their author avatars larger:

Screenshot 2023-03-09 at 1 28 27 AM

While digging around across the various Reference-related files and the Author component definition, I came across the conditional on line `95` that momentarily seemed to be the correct value to tweak (it made the Note and Gallery references' avatars smaller), it seems as though this value also affected the size of talk message author components:

https://user-images.githubusercontent.com/1195363/223939587-0f3dbbb2-0465-44e7-847a-11756bb4c1eb.mov

My question regarding this line: Which component is responsible for rendering reference author avatars at the intended size of `xxs`?

jamesacklin: `isRef` and `isReply` are set one component up in @/components/References/ReferenceBar.tsx. Could probably just set size statically there to xxs, make the size prop optional in Author, and set xs as the default.

jamesacklin:
Assigning myself to unblock @urcades

jamesacklin (approved):
This is good to go by my eye; I pulled the responsive ReferenceBar changes back in and did away with the container queries. All other spacing edits look good.

ui/src/components/References/ReferenceBar.tsx
```diff
@@ -4,12 +4,15 @@ import Author from '@/chat/ChatMessage/Author';
 import { daToUnix } from '@urbit/api';
 import { BigInteger } from 'big-integer';
 import ChannelIcon from '@/channels/ChannelIcon';
+import GroupAvatar from '@/groups/GroupAvatar';
 import useNavigateByApp from '@/logic/useNavigateByApp';
+import { useGroup } from '@/state/groups';
```
patosullivan, line 9:
```suggestion
```

ui/src/chat/ChatMessage/Author.tsx
```diff
@@ -89,6 +89,9 @@ export default function Author({
       <div onClick={handleProfileClick} className="shrink-0">
         <Avatar
           ship={ship}
+          // The conditional here appears to have scoping
+          // which affects the author avatar's size in both "chats themselves"
+          // as well as within ReferenceRow components.
```
patosullivan, line 94:
```suggestion
```

patosullivan (approved):

PR #2106, opened 2023-03-10
sidebar sort issues

### LINKED ISSUE #2103: groups: sort groups by "recent" is broken
Groups with older activity are appearing before groups without newer activity in some cases.

patosullivan:
Fixes #2103 

Also fixes an issue where the sidebar would rerender on each update to the settings store (such as the connection checker getting called).


arthyn (approved):
lgtm

PR #2107, opened 2023-03-10
perf: initial load remainder

### LINKED ISSUE #2105: groups: pinned DMs and groups wiped as a result of #2100
When #2100 hits your ship, pinned groups and DMs are gone from the sidebar.

The pinned items get added to local storage, but are not present in the sidebar on refresh.

arthyn:
Forgot to include this commit in the original PR, also includes pins fix #2105

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -305,22 +305,12 @@
   |=  =(pole knot)
   ^-  (unit (unit cage))
   ?+    pole  [~ ~]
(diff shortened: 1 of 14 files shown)


desk/app/groups.hoon
```diff
@@ -333,6 +323,15 @@
       ``noun+!>((~(has by groups) [src name.pole]))
   ==
 ::
+++  groups-light
```
jamesacklin, line 326: Very chill indeed

jamesacklin (approved):

PR #2109, opened 2023-03-10
perf: more load tweaks

### LINKED ISSUE #1375: groups: improve performance of reloading groups and talk browser apps

### LINKED ISSUE #1913: Group load is slow. Caused spontaneous refresh.

arthyn:
Some more loading stuff and fix to the erroring call on empty quilts.

Fixes #1913 
Fixes #1375

### DIFF
--- a/ui/src/state/bootstrap.ts
+++ b/ui/src/state/bootstrap.ts
@@ -33,6 +33,42 @@ const emptyTalkInit: TalkInit = {
   pins: [],
 };
 
+async function startGroups(talkStarted: boolean) {
+  // make sure if this errors we don't kill the entire app
+  const { chat, heap, diary, ...groups } = await asyncWithDefault(
+    () =>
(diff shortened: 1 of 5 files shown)


patosullivan (approved):

jamesacklin:
Nice. From 10k feet it looks like we're aggressively choreographing the loading sequence and being smart about initializing things in the natural order of the UX. 

Seems good to me!

PR #2110, opened 2023-03-10
WIP: add leap button and app grid, add app results to leap

patosullivan:
https://user-images.githubusercontent.com/1221094/224428900-1b647f56-110b-48aa-acc9-cad48c78798f.mov


arthyn:
This is great! I'd love to start moving leap into its own thing soon.

Two things I saw so far, hitting enter while apps is selected both opens landscape in a new tab and also opens the app modal, and clicking an app doesn't close leap when opening a tab.

urcades:
Yeah, these are solid bones to build on — I can provide some small bits of feedback to refine this — one immediate line of thinking I'll try to express soon is how to potentially integrate the "grid of apps display" into leap itself so there's less of a dramatic state switch.

It'll also be good to think about how to ensure that the grid of apps is keyboard navigable from leap itself, so that you after access the apps section, you can use arrow keys to navigate the tiles or continue to type to refine the results.

jamesacklin:
WSAD for app tile navigation

But seriously, looks good. My only worry is code deduplication across apps; theoretically this could/should be a shared library. Good that the data is the same, OK that it's not DRY, bad that we could drift in appearance between apps.

patosullivan:
image

patosullivan:
localhost_3000_apps_groups_(iPhone 12 Pro)
localhost_3000_apps_talk_(iPhone 12 Pro)

arthyn:
let's resolve the merge conflicts and get this in for 2.9

ui/src/components/Dialog.tsx
```diff
@@ -53,6 +55,11 @@ export const DialogContent = React.forwardRef<
             <X16Icon className="h-4 w-4" />
           </DialogPrimitive.Close>
         )}
+        {showClose && appModal && (
+          <DialogPrimitive.Close className="icon-button absolute top-8 right-6 bg-white">
+            <X16Icon className="h-4 w-4" />
+          </DialogPrimitive.Close>
+        )}
```
arthyn, line 62: Wondering if maybe we shouldn't just get rid of the booleans and just make this a slot so whoever calling can just put the close where they want

ui/src/app.tsx
```diff
@@ -191,6 +195,9 @@ function ChatRoutes({ state, location, isMobile, isSmall }: RoutesProps) {
       {state?.backgroundLocation ? (
         <Routes>
           <Route path="/about" element={<AboutDialog />} />
+          <Route path="/grid" element={<Grid />} />
+          <Route path="/app/:desk" element={<AppModal />} />
+          <Route path="/app/:desk/info" element={<TileInfo />} />
```
arthyn, line 200: Should probably use suspense for each of these and load dynamically since they're kinda secondary

ui/src/components/Leap/useLeap.tsx
```diff
@@ -525,6 +593,103 @@ export default function useLeap() {
...
+      // pinned charges are strong signals
+      // const isPinned = charge in pinnedCharges;
+      // if (isPinned) {
+      //  newScore += 10;
+      //  }
+      //  prefer unreads as well
+      //  const isUnread = isChannelUnread(charge);
+      //  if (isUnread) {
+      //  newScore += 5;
+      //  }
```
arthyn, line 623:
```suggestion
```

arthyn (changes requested):
a few nits, but lgtm. whoever merges first between this and james sidebar stuff will give the other some good conflicts lol

ui/src/app.tsx
patosullivan: Pushing this now.

ui/src/components/Dialog.tsx
patosullivan: Agreed, but outside of scope here, right?

arthyn (approved):
lfg!

urcades (dismissed):
Apologies for the late thoughts here — as promised, the majority of my 2¢ here revolve around frontend/visual updates to the existing work here:

Leap Options

Option A is essentially @patosullivan 's existing flow, but I'm recommending we enclose the app section "within" the Leap interface, and the resulting app preview "within" a standard-looking view (with one of our standard headers) in order to continually reinforce existing UI patterns we've established. All else is the same, including the ability to actually hover over an app and view its info as @patosullivan  had demonstrated.

There's a small suggestion that the app view is in fact a "filter" of some sort on Leap's results, with the ability to continue typing to narrow down the app results, but this is a nice-to-have, not necessary for this update.

The second option presented here simply suggests inserting apps into Leap as a default row, which removes the necessity to dig down one level to view "all apps". Because navigating a ton of apps might not be the most efficient option, we should surface the ability to just view "all apps", which takes us to a view seen in the first option.

Out of these two options, we should probably roll with the first since it's closest to the existing implementation, and I'm not trying to slow anything down here — it's worth noting that as we think more about system chrome, we should start to reify standards for navigation (such as always keeping Leap results contained within Leap itself), rather than continue to invent new patterns.

patosullivan:
Hey @urcades, rendering Grid within the Leap interface itself will actually require a pretty significant refactor to the Leap interface. Right now, the Grid interface is independent from Leap (it's actually a modal that displays at a particular route), but only accessible through leap. I basically just added a menu option to Leap that links to that route. 

What do you think about shipping this as-is and then I can follow up with a better match to your design treatment?

urcades:
@patosullivan thats totally fine, let’s just get it in there

patosullivan:
I did add the app title bar from your design, though:

image

PR #2111, opened 2023-03-11
meta: tracking app activity with vita

arthyn:
As a complement to installing %vita on sogryp to track downloads, we should also track daily usage. This PR implements the vita-client wrapper onto our UI agents so that it's out of the way of our main agents and adds the ability to poke for tracking app opens and also for disabling tracking. It will initially poke settings store to set show message to true, so that the default state of the setting can be false (no flashing).

The popover will stay up until it is interacted with. I chose defaulting to "on", but we can flip it if we feel that's a better look.

image

Any help with copy would be greatly appreciated here. I tried adding stuff about not transmitting data or personal information but it felt too wordy.


jamesacklin:
No bones about this being useful, just have to carefully consider the default and the language we use. Will run this down with the higher-ups and possibly test the language with some actual users.

patosullivan:
We should probably explicitly state the kind of data we mean when we say we're "tracking daily usage".

jamesacklin:
Here are my thoughts on language and approach. I encourage us to exhibit _opt-in_ behavior to tracking user activity, even as something as innocuous as this.

New Wireframe 1

Welcoming thoughts from @urcades and @lukestiles on approach, behavior, and language.

urcades:
The latter option is a pattern I'd also suggest we implement — it's the respectful choice.

arthyn:
yeah I like the second option I'll start heading that direction

arthyn:
image

lukestiles:
This looks like it landed in a good place. Please monitor network response in case we get negative feedback and need to yank or change the default. 


Is %vita linked to an app explainer?

patosullivan (approved):
Looks good to me. Should we add it to Landscape too?

arthyn:
Yeah we should do for landscape too

PR #2115, opened 2023-03-13
chat: fix image flickering

### LINKED ISSUE #2112: chat: broken image causes flickering
This image comes from a bucket with broken CORS. The height seems to be reflected correctly in the DOM, but the lack of a image data is causing a recalculating loop in the virtual scroller.

Image

Image

patosullivan:
This fixes #2112 

It turns out that chrome was forcing this content to load via HTTPS and palfun's minio instance doesn't have an HTTPS cert. When the image fails to load the height of the message changes, causing the scroller to re-render, which causes the image to load again, resulting in a loop.

To fix it we need to handle the error when the image doesn't load, store that state globally in the chat store, and display some static content in the scroller.

This also fixes an issue where we were missing a key for the list of thread reply authors in the ChatMessage component and adds a missing cleanup function to the useEffect in the ChatEmbedContent component.

image

### DIFF
--- a/ui/src/chat/ChatContent/ChatContentImage.tsx
+++ b/ui/src/chat/ChatContent/ChatContentImage.tsx
@@ -43,8 +43,8 @@ export default function ChatContentImage({
     return (
       <div className="embed-inline-block group">
         <div className="flex h-full flex-col items-center justify-center">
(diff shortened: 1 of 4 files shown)


jamesacklin (approved):
Elegant solution, nice that we also include the URL.

arthyn (approved):
lgtm, don't forget to change to gray

PR #2119, opened 2023-03-14
groups: appropriate null-groups routing

jamesacklin:
I got sick of having to hit Groups in the nav every time I opened the app on my phone.

For `/apps/groups`:
- If you aren't in any groups, you get routed to Find Groups, no matter what kind of device you're on. This should cover the first-run/newbie case.
- If you're in at least one group and you're on a mobile device, you get routed to the list of groups you're in.
- If you're in at least one group and you're **not** on a mobile device, you get routed to Activity ("All Notifications"), since there's a list of groups within view in the sidebar. @ashelkovnykov said something about this last week/the week before.

For `/apps/groups/notifications`:
- If you aren't in any groups, you get routed to Find Groups, which theoretically prevents us from showing the empty list at all for first-runners/newbies.
- If you're in at least one group, you get routed to Activity ("All Notifications").

This has the side effect of disallowing most navigation away from Find Groups on mobile on a fresh ship. However, we lack a "Don't see a group you want to join? Create your own" CTA bubble on Find Groups. Maybe @urcades has an opinion here?

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -1,6 +1,7 @@
 import cookies from 'browser-cookies';
 import React, { Suspense, useEffect, useState } from 'react';
 import { Helmet } from 'react-helmet';
+import _ from 'lodash';
 import {
   BrowserRouter as Router,
   Routes,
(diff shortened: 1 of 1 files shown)


patosullivan (approved):
Looks good!

Thanks, this was annoying me too.

PR #2120, opened 2023-03-14
groups: sane intra-group nav

jamesacklin:
This makes some similar changes as #2119.

- If you're on a mobile device, as you navigate into the group, you get dropped into the Channels list. Tapping into a channel and "back out" navigates you to the group root, rather than the named `/channellist` route (in case we change this in the future).
- If you're *not* on a mobile device, as you navigate into the group, you're either navigated to the most recent channel or Activity (current behavior).

mobile-nav


arthyn (approved):
lgtm

PR #2124, opened 2023-03-14
Add long press to view options on heap blocks

### LINKED ISSUE #2061: groups, gallery: can't delete on mobile, delete on desktop shows no spinner and hangs
- Be an admin of a group with a gallery channel that has stuff in it
- Attempt to delete gallery entry on mobile (I couldn't find out how)
- Attempt to delete same entry on desktop
- Observe it hang when you click delete with no spinner
- Hard refresh the page
- Observe that it is in fact deleted

patosullivan:
Also add a useLongPress hook that we can start using elsewhere.
Fixes #2061

### DIFF
--- a/ui/src/heap/HeapBlock.tsx
+++ b/ui/src/heap/HeapBlock.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { HeapCurio, isLink } from '@/types/heap';
 import cn from 'classnames';
 import { isValidUrl, validOembedCheck } from '@/logic/utils';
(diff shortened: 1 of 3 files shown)


ui/src/heap/HeapBlock.tsx
```diff
@@ -100,7 +105,7 @@ function TopBar({
           )}
         </div>
         {canEdit && (
-          <div className="relative hidden group-hover:block">
+          <div className={longPress ? 'relative' : ' hidden group-hover:block'}>
```
arthyn, line 108: seems like you may still want relative when not longpressed?

arthyn (approved):
lgtm, just the one question

ui/src/heap/HeapBlock.tsx
patosullivan: Yeah, you're right.

PR #2132, opened 2023-03-15
emoji picker: use only controlled hover, no css

### LINKED ISSUE #2108: chat: emoji picker impossible to hit with mouse
The emoji picker for reactions is nigh impossible to use in Groups and frequently does not appear reliably in Talk (I think this is something to do with how we're handling the mouse-out/close event).

No reproduction other than "play with it and see."

patosullivan:
It looks like we forgot to take out the CSS hover stuff when we switched to a controlled hover state for the chat message options, so we had both the JS hover state and the CSS hover state clashing with each other. Now we just use the JS state.
Fixes #2108
Also fixes the issue where the message options wouldn't appear on first hover.

### DIFF
--- a/ui/src/chat/ChatMessage/ChatMessageOptions.tsx
+++ b/ui/src/chat/ChatMessage/ChatMessageOptions.tsx
@@ -84,7 +84,7 @@ export default function ChatMessageOptions(props: {
 
   const navigate = useNavigate();
   return (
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

jamesacklin:
tysm

PR #2134, opened 2023-03-15
Po/fix group membership issues

### LINKED ISSUE #2129: groups: issues with membership + invite capabilities
Somehow, the ability for any group member to invite someone to an open group has been dropped from the menus I'd expect it in:

The group actions dropdown list available within the group itself:

Screenshot 2023-03-14 at 8 03 26 PM

The same/mirrored list of actions that's found in the "..." menu in the "All Groups" sidebar:

Screenshot 2023-03-14 at 8 03 32 PM

Within the "All Group Members" screen in any particular open group:

Screenshot 2023-03-14 at 8 02 49 PM

As seen in the image above, I'm also unable to currently view the membership list of an open group.

patosullivan:
Fixes #2129

Allow group members in public groups to invite new members.

Also, load members into group store on navigating to a group. It looks like this was caused by changes in #2100. To improve initial load, we don't load all members for all groups, but we have the option to load them for a particular group when necessary, we just didn't update Groups.tsx to reflect this. This also fixes the issue that Zach first noticed where the mention pop-up shows "not in group" for group members.

### DIFF
--- a/ui/src/groups/GroupActions.tsx
+++ b/ui/src/groups/GroupActions.tsx
@@ -12,8 +12,8 @@ import { useChatState, usePinnedGroups } from '@/state/chat';
 import LeaveIcon from '@/components/icons/LeaveIcon';
 import useIsGroupUnread from '@/logic/useIsGroupUnread';
 import UnreadIndicator from '@/components/Sidebar/UnreadIndicator';
-import { citeToPath, useCopy } from '@/logic/utils';
-import { useAmAdmin } from '@/state/groups';
+import { citeToPath, getPrivacyFromGroup, useCopy } from '@/logic/utils';
+import { useAmAdmin, useGroup } from '@/state/groups';
 
 const { ship } = window;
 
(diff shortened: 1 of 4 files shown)


arthyn (approved):
lgtm, can we wrap the members scry inside `initialize` with `useSchedulerStore.getState().wait(() => {...}, 3)` or w/e priority works to keep this from blocking chat messages loading first

patosullivan:
It seems to load simultaneously with chat messages regardless of the priority unless I set it to 6 (in which case it doesn't load). Not sure what I'm doing wrong here:

```
        const { wait } = useSchedulerStore.getState();
        if (getMembers) {
          wait(async () => {
            console.log('getting members');
            const group = await api.scry<Group>({
              app: 'groups',
              path: `/groups/${flag}`,
            });
            console.log('got members', group);
            get().batchSet((draft) => {
              draft.groups[flag] = group;
            });
          }, 6);
        }
```

patosullivan:
We talked out of band and this is fine as-is.

PR #2136, opened 2023-03-15
groups: add lure integration

midsum-salrux:
Merge lure frontend changes from https://github.com/tloncorp/lure

image

image


arthyn:
@midsum-salrux if you could take a look at the moves I made here to make sure it follows the same logic

PR #2143, opened 2023-03-16
vita: account for ships with previous ~ state

arthyn:
This should fix the issue @patosullivan saw earlier about the missing mark. The state transition didn't work for ships who previously had the groups/talk ui agents. Too much fakezod juggling meant this didn't happen locally when I was developing it.


midsum-salrux (approved):
LGTM

PR #2144, opened 2023-03-16
perf: add service worker

patosullivan:
This branch uses vite-plugin-pwa to add a service worker at build time. Adding this service worker will improve load times and allow us to extend our apps with more app-like (PWA) functionality later (push notifications, better offline support, background sync, performance optimizations available via network request interception, etc).

At the moment, the service worker will get us some performance improvement by serving precached files.

Note that in order to test a SW locally you *cannot* use https (without a lot of trouble), so I added a new "sw" mode and corresponding scripts to be used when we need to work on SW stuff locally.

I also added a patch that will allow us to inject our base app routes into the `generateRegisterDevSW` function in vite-plugin-pwa (the patch looks huge because of prettier, unfortunately).


arthyn:
can you copy the patch change in a code block w/o the prettier changes?

patosullivan:
Made the apps installable as PWAs (i.e., the browser will recognize that they are PWAs and are installable) and added a prompt:

image

jamesacklin (approved):
Overall, seems like a net good to me. It will be good to develop a caching strategy that allows us to hold app data on the device and stream updates to it.

My only concern is with the safe area of the PWA icons (attachment forthcoming).

patosullivan:
Good catch. @urcades can you provide some icons with an appropriate safe area?

arthyn (approved):
lgtm

urcades:
@jamesacklin + @arthyn I'm going to be aiming for a crop ratio of hereabouts:

Screenshot 2023-03-28 at 12 37 24 PM

I'm unsure what dimensions are actually required here, but I've provided 1x/2x/3x + svg formats in the attached .zip:

[pwa-app-icons.zip](https://github.com/tloncorp/landscape-apps/files/11092180/pwa-app-icons.zip)

patosullivan:
> @jamesacklin + @arthyn I'm going to be aiming for a crop ratio of hereabouts:
> 
> Screenshot 2023-03-28 at 12 37 24 PM
> 
> I'm unsure what dimensions are actually required here, but I've provided 1x/2x/3x + svg formats in the attached .zip:
> 
> [pwa-app-icons.zip](https://github.com/tloncorp/landscape-apps/files/11092180/pwa-app-icons.zip)

hey @urcades, could I get these in 192x192 and 512x512 sizes?

urcades:
@patosullivan gotcha:

[pwa-2.zip](https://github.com/tloncorp/landscape-apps/files/11093570/pwa-2.zip)

PR #2147, opened 2023-03-16
clubs: rearchitecture for better gossip control

### LINKED ISSUE #2021: clubs: gossip is too extreme, needs better mechanism
The exponential effects of our current gossip scheme for clubs is crushing ships that are in group chats of more than 10 people. Duplicate facts are not ignored as well so this leads to extreme amounts of hark notifications.

arthyn:
> This changes our "stop" mechanism from echoes to already heard. So each piece of gossip travels with a unique identifier and if we've already heard that uid then we just ignore it. If we haven't heard it, we continue to gossip it to everyone and process like before. So at most we should hear n facts, where n is the number of group members, because the initial sender sends one, which everyone gossips to everyone else, so we should hear one per person. One of these will get recorded in our heard set and the rest ignored.
>
> Clarifying because this was unclear to me: the reason we initially send to everyone and also gossip to everyone is that this ensures that we get close to proper ordering in the case of sender/receiver ships going offline in tandem and then coming online out of order. If the receiver comes back online and hasn't heard the senders poke yet (because they're offline) the gossipers should conceivably poke the gossip first before poking any other messages that they would have sent as well, conserving order.
>
> @jamesacklin flagging that this is breaking change we should communicate to third party devs.

Remade PR because of some base branch changes. You can see me floundering around with versioning stuff which I left in the commits in case we need it later, but the types actually nest so I was able to do a little sleight of hand. Old versions will hear a piece of gossip with a really large echo and just not gossip it forward because it's above the threshold. New versions will check their `heard` set and see if it's in there and ignore if so.

This now does not break communication between versions so it should be super smooth to deploy.

Fixes #2021


jamesacklin:
@alecananian FYI

philipcmonk (approved):
Looks good!

PR #2152, opened 2023-03-17
chat: insert images after other block content

### LINKED ISSUE #2137: talk: image uploads inserted above quote-reply

If I hit the reply button to a chat message & then I hit the button to upload image to s3, the image ends up above the reference

**To Reproduce**
1. reply & upload image
2. observe image above reference

**Expected behavior**
image should go below the mention

patosullivan:
Fixes #2137
Inserting the image before the quoted reply looked weird.

### DIFF
--- a/ui/src/chat/ChatContent/ChatContent.tsx
+++ b/ui/src/chat/ChatContent/ChatContent.tsx
@@ -185,12 +185,29 @@ function ChatContent({
   const blockLength = story.block.length;
   const firstBlockCode = story.inline.findIndex(isBlockCode);
   const lastBlockCode = findLastIndex(story.inline, isBlockCode);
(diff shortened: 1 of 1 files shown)


ui/src/chat/ChatContent/ChatContent.tsx
```diff
@@ -185,12 +185,29 @@ function ChatContent({
   const blockLength = story.block.length;
   const firstBlockCode = story.inline.findIndex(isBlockCode);
   const lastBlockCode = findLastIndex(story.inline, isBlockCode);
+  const blockContentImages = story.block.filter(isChatImage);
```
arthyn, line 188: could we just use a sort here so there's less moving parts?

arthyn (approved):
lgtm

Issue #2161, opened 2023-03-20
chat: text formatted as link renders as embed


alecananian:
**Describe the bug, optionally with screenshots**
Related to #2146 

When you format a specific portion of your message as a link to any of the supported embed sites (Twitter, YouTube, etc.), the text is not rendered and you instead get an inline embed:

<img width="302" alt="Screenshot 2023-03-20 at 10 39 43 AM" src="https://user-images.githubusercontent.com/1013230/226423000-257b44db-d656-4a70-b34c-fd98546aa196.png">
<img width="304" alt="Screenshot 2023-03-20 at 10 39 48 AM" src="https://user-images.githubusercontent.com/1013230/226423002-bf11ed31-c521-40f3-a7ab-fdb4ad62109c.png">
<img width="433" alt="Screenshot 2023-03-20 at 10 39 54 AM" src="https://user-images.githubusercontent.com/1013230/226423003-e8121923-e1dd-4720-8cd3-b6ed5abe8f58.png">

**To Reproduce**
1. Type a text message
2. Highlight part of the message and format it to be a link from Twitter
3. Send the message
4. See that the highlighted text is now an embed and the original text is not visible anywhere

**Expected behavior**
The full text message is available along with thembed

**Additional context:**
- Tlon-hosted planet
- macOS 13.0, Chromium

PR #2162, opened 2023-03-20
chat: show link custom content when rendering as plain link

### LINKED ISSUE #2146: chat: text formatted as link renders as the link without the text
from a user: 
"highlighted text in chat box, clicked the formatting button for a link, entered a URL, clicked 'done'. it shows as the text i highlighted with the link inserted as one would expect before i send the message, but after i send the message the text disappears and is replaced with the link"

alecananian:
Fixes #2146 

Passes custom content from `Link` inline story type to the embed renderer so the `<a>` tag can render it. Does not address the issue when the URL is a supported embed site, logged that separately in #2161

### DIFF
--- a/ui/src/chat/ChatContent/ChatContent.tsx
+++ b/ui/src/chat/ChatContent/ChatContent.tsx
@@ -98,7 +98,13 @@ export function InlineContent({
   }
 
   if (isLink(story)) {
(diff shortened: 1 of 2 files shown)


patosullivan (approved):
Awesome, thanks!

PR #2163, opened 2023-03-20
app modal: fix dark mode issue

patosullivan:
Some apps (like %pals) don't explicitly set a background color and just expect the default white background color in the browser. I originally set the background color for the iframe to white using tailwind's `bg-white`, this caused the background to change to black automatically in dark mode. Apps like %pals then became unusable because they were using black text on our black background.

This change explicitly sets the background to `#fff` so that apps like %pals will render properly. Apps that set their own background (like groups or talk) will be unaffected.


arthyn (approved):
lgtm

PR #2171, opened 2023-03-21
pwa: fix some issues for both platforms

patosullivan:
Fixes an issue where extraneous bottom padding would be displayed on Android. 
Fixes an issue where a bad start_url in the manifest would cause the PWA to show a blank screen on iOS.


jamesacklin (approved):
Makes sense to me, thanks!

PR #2172, opened 2023-03-21
app modal: always close to background location

patosullivan:
This makes some changes that ensure that the user is always redirected to the background location upon closing an AppModal (which seems more natural, less annoying than having to go back through Grid and Leap).

It does the same thing for closing Grid.

Also fixes the app links in Leap for opening a new tab for groups/talk.


jamesacklin (approved):
Thanks. This trailing slash issues are pervasive but I'm glad we're catching them as we see them.

Issue #2178, opened 2023-03-21
talk appears to be borked after the recent update


jalehman:
Navigating to the talk app opens a blank screen that never loads.

<img width="1337" alt="Screenshot 2023-03-21 at 1 59 09 PM" src="https://user-images.githubusercontent.com/550978/226739603-7f6d71cb-ef2c-4fbc-8409-3c295f0b5262.png">

The console shows a few errors:

1. A 404 on attempting to GET https://my-domain/~/~scry/talk-ui/init.json
2. Uncaught (in promise) SyntaxError: Unexpected token '<', "<html><hea"... is not valid JSON

<img width="1339" alt="Screenshot 2023-03-21 at 1 59 40 PM" src="https://user-images.githubusercontent.com/550978/226739909-159f067b-0b6f-4425-a9d0-f5535e4329e3.png">

Relevant system stats:

```
> +vat %talk
%talk
  /sys/kelvin:      [%zuse 417] [%zuse 416] [%zuse 415] [%zuse 414]
  base hash:        0vk.vqbgc.ff98d.e9vp3.ld9l5.2uj8b.o5lg5.6pnhi.1l9q9.furqm.mi53e
  %cz hash:         0vk.vqbgc.ff98d.e9vp3.ld9l5.2uj8b.o5lg5.6pnhi.1l9q9.furqm.mi53e
  app status:       running
  force on:         ~
  force off:        ~
  publishing ship:  [~ ~sogryp-dister-dozzod-dozzod]
  updates:          remote
  source ship:      ~marnec-dozzod-marzod
  source desk:      %talk
  source aeon:      52
  kids desk:        ~
  pending updates:  ~[[%zuse 416] [%zuse 414]]
::
> +vat %base
%base
  /sys/kelvin:      [%zuse 415]
  base hash:        0va.mn6n0.2ni3n.vpr1r.9b98s.q2jve.4osvm.g4hvn.magll.uqoll.3jbrb
  %cz hash:         0va.mn6n0.2ni3n.vpr1r.9b98s.q2jve.4osvm.g4hvn.magll.uqoll.3jbrb
  app status:       running
  force on:         ~
  force off:        ~
  publishing ship:  ~
  updates:          remote
  source ship:      ~binnec-dozzod-marzod
  source desk:      %kids
  source aeon:      62
  kids desk:        %kids
  pending updates:  ~[[%zuse 414]]
```

Am available for more info if you need help reproducing or inspecting, you know how to reach me.

[...]

PR #2180, opened 2023-03-21
talk: offer fallback if talk out of sync

arthyn:
OTT prevents issues like #2178


jamesacklin (approved):
Seems like a fine pattern, should support the case where someone uninstalls Groups and leaves Talk.

patosullivan (approved):

PR #2183, opened 2023-03-22
groups: Disable invites for non-admins

patosullivan:
This hides invite actions for non-admins while we figure out why invites to public groups from non-admins are failing.


jamesacklin (approved):
All in due time, thanks.

PR #2187, opened 2023-03-23
curio: detail view fixes

jamesacklin:
Cleans a bunch of things up w/r/t layout in the curio detail view.

- Stacks content and comments vertically for smaller viewports.
- Uses the window scroll in Virtuoso for smaller viewports (eliminates any scroll-within-scroll problems).
- Realigns the comment content.
- Brings the sidebar metadata up to spec (https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/Product-Handoffs?node-id=2799-114533&t=v23dQDf7A7iGe9BC-4).
- Manually truncates the "back" button in the header to avoid the line-clamp bug which destroys the layout on smaller screenss.

https://user-images.githubusercontent.com/748181/227311717-533ee2b5-aa31-4459-b3d9-ab251ae0ec5e.mov

https://user-images.githubusercontent.com/748181/227311731-8117dbb1-c01d-466d-b540-a9d829e7ab9e.mov

### DIFF
--- a/ui/src/heap/HeapDetail.tsx
+++ b/ui/src/heap/HeapDetail.tsx
@@ -124,7 +124,7 @@ export default function HeapDetail() {
     </div>
   ) : (
     <Layout
-      className="flex-1 bg-gray-50"
+      className="flex-1 bg-white"
       header={
         <HeapDetailHeader
           flag={groupFlag}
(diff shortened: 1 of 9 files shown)


arthyn:
This is good! when I load on mobile though I only see the first 4 comments, not until I click the input box that I get the rest of the comments. maybe something to do with window scroll and how that plays with top/bottom loading?

jamesacklin:
@arthyn Possibly. Let me look into it. Currently the flexbox layout prohibits any kind of scrolling/loading whatsoever.

Is there a really amazing/compelling reason we're using Virtuoso here at all? Rare is the curio with even 3+ comments.

arthyn:
Probably safe to switch back for now, not sure if it was added for posterity or for an issue @patosullivan. we still need to convert this to quips model for better querying anyway so we can just add it back when we do that or something.

jamesacklin:
@arthyn Switched out, much simpler.

arthyn (approved):
lgtm

Issue #2140, opened 2023-03-16
Refresh on mobile loses context at groups root


lukestiles:
**Describe the bug, optionally with screenshots**
When using Groups on mobile web and viewing the groups list, if the view refreshes for any reason, the notifications screen is loaded

**To Reproduce**
https://user-images.githubusercontent.com/1903610/225678907-44476f84-460b-417d-83ed-f024e185adc8.mp4

**Expected behavior**
I'd expect to stay on the screen I was viewing after a refresh. 

**Additional context:**
- self-hosted DO planet, Chrome on iOS

patosullivan:
I can't repro this by just hitting the refresh button (i.e., the correct page is showing on refresh). Did you do anything else with the URL bar? Is this issue still occurring?

jamesacklin:
I've noticed that the `/apps/groups/` path resolves flawlessly; refreshing strips the trailing `/`, which fails.

PR #2190, opened 2023-03-24
Fix bottom spacing issues on PWAs

patosullivan:
Fixes the issue where non-IOS PWAs were getting extra bottom padding, also ensure that we *always* have bottom padding on iOS PWA to avoid colliding with the home bar.

Also fixes an issue where navigating back to the "home" route would appear as navigating to an external route in the PWA (by updating manifest settings and making sure our basename in app.tsx includes a trailing slash).

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -12,7 +12,7 @@
       data-react-helmet="true"
     />
     <link rel="apple-touch-icon" href="/src/assets/apple-touch-icon.png" />
-    <meta name="apple-mobile-web-app-capable" content="yes" />
+    <meta name="mobile-web-app-capable" content="yes" />
   </head>
   <body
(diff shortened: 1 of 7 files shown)


jamesacklin (approved):
Nice, good improvement.

Does this also fix #2140 w/r/t https://github.com/tloncorp/landscape-apps/pull/2190/files#diff-093127603bee40a36f6c601028f9af200504cb9dba3a73b808ae2b2643e5980cR603 ?

patosullivan:
>

> Nice, good improvement.
> 
> Does this also fix #2140 w/r/t
> 
> https://github.com/tloncorp/landscape-apps/blob/092fc4e910c777485dc121afdc647ccfd57d11bf/ui/src/app.tsx#L601-L607
> 
> ?

yes, does seem to fix it.

PR #2191, opened 2023-03-24
groups: lures for lure

arthyn:
This adds an entry into the invite people dialog screen so that people now see that they can invite others whether on urbit or off. This combines the editor in the previous PR with display so that the host can enable from either location. This also pulls the "invite people" link up into the group sidebar so it's more present. There are more locations in @urcades onboarding figma, but I thought we should just get this going first.

Also refactors the way we do dialogs so that there's less componentry to use out the gate and cleans up our scrollbars so they're transparent except for the thumb.

Screen Shot 2023-03-24 at 5 26 22 PM

**host**
Screen Shot 2023-03-24 at 5 26 13 PM

Screen Shot 2023-03-24 at 5 25 58 PM

**member**

Screen Shot 2023-03-24 at 5 26 29 PM
Screen Shot 2023-03-24 at 5 26 55 PM


zalberico (changes requested):
I think the invite headers could be clearer, but I like the separate modal and where you have the link in the sidebar.

> Send Landscape Invite to Someone

I'd probably change this to: Send Invite Link to Friends Not on Urbit

> Invite people to group

I'd change this to: Send Invite to Friends Already on Urbit

Thoughts?

urcades:
Here's a set of suggestions, and an interface to grant them some context:

---

NEW — **Invite someone from outside the Urbit network**
Courtesy of Tlon Corporation

Have friends, family or collaborators who aren’t on Urbit? You can now gift them an Urbit ID and onboard them to your group all in one free and easy sweep. Just copy and share the link below. 

---

**Invite by Urbit ID** 
(e.g. ~sampel-palnet) or display name. 
[Search]

---

Interface Example:

Invite to Group Modal

arthyn:
updated to look like this: 
Screen Shot 2023-03-27 at 2 17 07 PM

ui/src/dms/MessagesSidebar.tsx
```diff
@@ -140,7 +140,7 @@ export default function MessagesSidebar() {
   );
 
   return (
-    <nav className="flex h-full w-64 flex-none flex-col border-r-2 border-gray-50 bg-white">
```
patosullivan, line 143: Why make this full?

patosullivan (approved):
Looks good! Just one style related comment, but I'm probably missing something.

jamesacklin (approved):
Degrades nicely if you don't have lure support installed (or the group host doesn't have it turned on).

zalberico (approved):
Thanks - Let's ship it!

ui/src/dms/MessagesSidebar.tsx
arthyn: bah I saw this comment and forgot lol

PR #2192, opened 2023-03-25
groups: fix group header color in dark mode

jamesacklin:
I don't work at night enough to notice these things, I guess.

Before:

image

After:

image


arthyn (approved):
lgtm

PR #2209, opened 2023-03-28
leap,dialog: fix default styling and adjust leap to account

arthyn:
OTT more cleanup from #2191 also adjusts global default styling for dialogs (adjusting from pulling from grid)


patosullivan (approved):
Looks good

PR #2211, opened 2023-03-28
lure: grouper agent

arthyn:
For sake of types and colocation of concerns we decided to move %grouper and %greeting to this desk so that landscape can stay clean of groups stuff.


midsum-salrux (approved):
LGTM

jamesacklin:
Will retag and redeploy.

PR #2216, opened 2023-03-29
http-api: incorporating new reconnect logic

arthyn:
What this does is incorporate upcoming changes to @urbit/http-api which make the reconnect logic behave much better. Because this now behaves better we have to actually handle subscription kick/quit now and ensure that we rescry everything that could have been missed in between.

To do this, I refactored our usage of `api` and `useSubscriptionState` so that it's now combined into a cohesive class with everything wrapped together. 

**subscriptions**
- now are deduped by default, so we no longer need to guard against that and can call the same sub over and over with no issues. This enables us to call bootstrap anytime we want now.
- all scheduled by default to pass 4 but can be changed w/ param

**poke tracking**
- this now is a first class function `api.trackedPoke...` which accepts a poke with a subscription app/path and a hook to track response
- default hook/validator for when poke === response

Also this new @urbit/http-api enables us to use @tloncorp/eyrie so we can debug subscriptions better


patosullivan:
I just used eyrie/this branch to debug another issue. This is great.

ui/vite.config.ts
```diff
@@ -107,6 +107,45 @@ export default ({ mode }: { mode: string }) => {
   console.log(process.env.APP);
   console.log(mode, app, base(mode, app));
 
+  const rollupOptions = {
+    external:
+      mode === 'mock' || mode === 'staging'
+        ? ['virtual:pwa-register/react']
```
patosullivan, line 113: Why do we need this in mock/staging?

patosullivan (approved):
Looks great, works well locally, just left one question and one suggestion.

ui/vite.config.ts
arthyn: was running into type errors from the SW since this is in the update banner

PR #2222, opened 2023-03-30
Check bait URL for invite links more often

midsum-salrux:
Fixes https://github.com/tloncorp/lure/issues/49

Currently this checks every 30 seconds. If that feels slow we can make it more often.


arthyn (approved):
lgtm

PR #2227, opened 2023-03-30
hark: more sane usage

### LINKED ISSUE #2022: latest.json responses for talk sometimes really big
Occasionally I get into a state (not clear to me yet how) where my ship spends large amounts of time spinning on `<<http>>`. I suspect the Talk frontend is the one making requests that cause that, because:
1) Closing the tab with Talk in it usually alleviates the issue.
2) Looking in the networking tab for Talk when the problem occurs, requests for `latest.json` receive a response that's a couple of _megabytes_ in size. Latest I saw was 3.8mb.

If the backend is trying to serialize that much chat data, no wonder it's taking some time. Whatever "latest" we're scrying could probably get a more tightly-scoped response from the backend.

arthyn:
A refactor of how we update hark state in the app. The previous was too aggressive and possibly causing issues. We now only update at the desk level when hearing a fact, otherwise we only update when visiting the notifications page. Not sure this was @ashelkovnykov 's issue, but seems like it could be part of it.

The other part of this introduces a new single scry `skeins` which gets us only the data we need to display the notifications screen and count which should fix #2022 and possibly the other part of alex's issue although I think there might be something else at play.


ashelkovnykov:
@arthyn Thanks a lot - eager to see if this helps!

patosullivan (approved):
Looks good to me. Compiles and runs well locally.

jamesacklin (approved):
Looks good to me, doesn't appear to break anything for our other clients.

PR #2247, opened 2023-04-05
lure: reduce timeout when querying for url

### LINKED ISSUE #2242: [Lure] Fail faster when checking for lure link
Right now when you're in a group and click the Invite People button in the sidebar the modal for external invites will spin and timeout if no link is set (after a long time). When a link is set it typically succeeds in under 5s. I think we can fail faster when one isn't set (maybe 5s? maybe less?).

tomholford:
This addresses feedback from initial testing of the Lure + Groups integration.

Resolves #2242


arthyn (approved):
seems fine, but the test broke or something?

tomholford:
> seems fine, but the test broke or something?

Something is up, the test is passing locally but failing in CI. Going to skip it for now

PR #2252, opened 2023-04-05
chat: fix double render on reconnect

### LINKED ISSUE #2026: chat: return of the double render
Now seeing double renders in some group chat channels: 

image

patosullivan:
Fixes #2026

@arthyn found this while we were on a call


arthyn (approved):
lgtm

PR #2256, opened 2023-04-06
sw: Update service worker via UpdateNotice bar.

patosullivan:
This changes our service worker update strategy to `prompt` and then allows for the service worker to be updated by the user when they click the "Reload" button in the UpdateNotice bar.

Before we were using `autoUpdate`, but a) we didn't include the appropriate function to autoUpdate in main.tsx and b) this would force a refresh on users without their prompting, which could be pretty annoying in a chat context.

I also updated the eslint rules to ignore the virtual module import (and the `@` import, since this would still be flagged by eslint in some situations).

I also updated the tsconfig to use the types for the virtual module.


arthyn (approved):
lgtm

PR #2271, opened 2023-04-07
notifications: fix mention links

### LINKED ISSUE #2263: groups: notifications for thread msgs open the msg as its own thread
Whenever I get mentioned in a message that is posted _in a thread_, this shows up correctly in my notifications.

However, clicking it opens the mention-message _as its own thread_, as opposed to opening in the context of the thread it was originally in.

I can respond in that "nested thread", but it won't be accessible in the groups chat interface through normal means. There is no "1 reply" indicator on the message in its original thread context.

What I really want is for clicking the notification to take me to the message _in its original context_, like it does correctly with top-level messages.

patosullivan:
fixes #2263 by opening up the thread and scrolling to the mention in question.

Also: for normal mentions (i.e., mentions in the main chat window), this routes the user to the mention within the chat window rather than prompting the user to start a new thread for the mention.

### DIFF
--- a/ui/src/logic/usePathFromHark.ts
+++ b/ui/src/logic/usePathFromHark.ts
@@ -3,10 +3,6 @@ import { useWritByFlagAndWritId } from '@/state/chat';
 import { Skein } from '@/types/hark';
 
 export default function usePathFromHark(bin: Skein, isMention: boolean) {
-  if (!isMention) {
-    return bin.top?.wer || '';
-  }
-
   const writId = bin.top?.wer.split('/')[10];
   const ship =
     typeof bin.top.con[0] !== 'string' && 'ship' in bin.top.con[0]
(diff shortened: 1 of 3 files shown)


arthyn:
does this work for all notifications? seems like it's only for chat-related ones

patosullivan:
> does this work for all notifications? seems like it's only for chat-related ones

do we generate mention notifications for anything other than chats?

patosullivan:
Looks like the only place we do it is line 1445 in app/chat.hoon

ui/src/logic/usePathFromHark.ts
```diff
@@ -0,0 +1,45 @@
+import { udToDec } from '@urbit/api';
+import { useWritByFlagAndWritId } from '@/state/chat';
+import { Skein } from '@/types/hark';
+
+export default function usePathFromHark(bin: Skein, isMention: boolean) {
+  const writId = bin.top?.wer.split('/')[10];
```
arthyn, line 6: Oh sorry I see now, I think it would be more clear if this early returned `wer` if it's not a mention above all the rest.

arthyn (changes requested):

ui/src/logic/usePathFromHark.ts
```diff
@@ -0,0 +1,46 @@
...
+    bin.top?.wer.split('/')[7]
+  }`;
+  const idWrit = `${ship}/${writId}`;
+  const replyToWrit = useWritByFlagAndWritId(channelFlag, idWrit || '', false)
+    ?.memo.replying;
+  const wer = bin.top?.wer;
+
+  if (!isMention) {
+    return wer;
+  }
```
arthyn, line 21: I think this needs to be at the top otherwise this will error out when trying to reference specific segments like `[10]`

arthyn (changes requested):

arthyn (approved):
I was wrong in my last comment, pushed a revert and gtg

PR #2273, opened 2023-04-10
emoji picker: fix closing immediately on mobile

### LINKED ISSUE #2253: emoji picker closes immediately on mobile (sometimes)

patosullivan:
This stops the emoji picker from closing immediately on mobile by removing the search input.

Disabling autoFocus for the input isn't enough, if the user selects the input the picker will still close.

This stops the bleeding while we figure out what is causing the picker to close when the search within the picker is focused (or we switch to another picker or another method for displaying the picker).

Also pops the picker up closer to the message itself since we don't have to worry about the keyboard anymore.

Fixes #2253

### DIFF
--- a/ui/src/components/EmojiPicker.tsx
+++ b/ui/src/components/EmojiPicker.tsx
@@ -36,18 +36,16 @@ export default function EmojiPicker({
       ) : (
         children
       )}
-      {(isMobile || !withTrigger) && (
-        <Popover.Anchor className={isMobile ? 'fixed inset-x-0 top-12' : ''} />
-      )}
+      {(isMobile || !withTrigger) && <Popover.Anchor />}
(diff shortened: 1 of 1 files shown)


patosullivan:
> I have a feeling people will be pretty upset if we take the search input away.
> 
> Is this turning into something where we evaluate another approach altogether? This has given us nothing but trouble.

kinda feels like it, yeah. I could go back to looking into other emoji picker options?

patosullivan:
Note that this only removes the search input on mobile, though. Also note that this issue only seems to happen with latest Android. Not sure how we could detect if the user is on the *latest* version of Android, but we could detect if they're on Android. Would be a regression for everyone on Android that isn't on the latest version, though.

PR #2275, opened 2023-04-10
chat: lift chat hover state to chat store, use router for emoji picker open state on mobile

### LINKED ISSUE #2253: emoji picker closes immediately on mobile (sometimes)

patosullivan:
- Fixes the issue where the emoji picker would disappear if you hovered slightly out of/away from the chat message that spawned it by lifting hover state up into the chat store
- Adds 100px of padding on the left and the top of the emoji picker to prevent it from accidentally closing if the user hovers over another message outside of the picker.
- Fixes #2253, the issue where users on latest android couldn't use the emoji picker on mobile

### DIFF
--- a/ui/src/chat/useChatStore.ts
+++ b/ui/src/chat/useChatStore.ts
@@ -12,7 +12,7 @@ export interface ChatInfo {
     brief: ChatBrief; // lags behind actual brief, only gets update if unread
   };
   dialogs: Record<string, Record<string, boolean>>;
-  hovering: Record<string, boolean>;
+  hovering: string;
   failedToLoadContent: Record<string, Record<number, boolean>>;
 }
 
@@ -49,7 +49,7 @@ const emptyInfo: ChatInfo = {
   blocks: [],
   unread: undefined,
   dialogs: {},
-  hovering: {},
+  hovering: '',
   failedToLoadContent: {},
 };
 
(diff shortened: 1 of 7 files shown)


ui/src/chat/useChatStore.ts
```diff
@@ -12,6 +12,7 @@ export interface ChatInfo {
     brief: ChatBrief; // lags behind actual brief, only gets update if unread
   };
   dialogs: Record<string, Record<string, boolean>>;
+  hovering: Record<string, boolean>;
```
arthyn, line 15: I think you just want this to be a string, because we shouldn't ever have more than one message hovered. As is right now you can get multiple hovers.

ui/src/chat/useChatStore.ts
```diff
@@ -225,6 +239,26 @@ export function useChatDialog(
   };
 }
 
+export function useChatHovering(
+  whom: string,
+  writId: string
+): { hovering: boolean; setHovering: (hovering: boolean) => void } {
+  const { setHovering } = useChatStore.getState();
+  const hovering = useChatStore(
```
arthyn, line 247: Given what I said above this would just change to be `hovering === writId`

arthyn (changes requested):
Let's make sure we can't get duplicates, also getting no routes matched when I try this on mobile in talk

PR #2278, opened 2023-04-11
perf: update indexedb persistor to store queries separately

patosullivan:
re: our discussion yesterday, took a stab at writing our own persistor client.


patosullivan:
> seems much slower actually, I guess because we're separately awaiting each query to be read from the cache on restore.

might have been a temporary issue with the dev moon I was using, seems fine now.

jamesacklin:
Won't delete the branch, just closing the PR for now.

PR #2280, opened 2023-04-11
Fix twitter embeds in heap detail in safari

### LINKED ISSUE #2279: heap: twitter embeds in heapdetail don't work in safari
Probably just need to match what we're doing with the twitter embeds in the lightbox in chat.

patosullivan:
Fixes #2279 by providing a definite height/width for twitter embeds like we do in the lightbox (and which safari seems to require).

Also account for scrollable content in the heapdetail (if a user were to link to a tweet in a thread, for instance)


arthyn (approved):
lgtm

PR #2284, opened 2023-04-11
emojis: indicate which were sent by the user

### LINKED ISSUE #2098: chat: indicate which emoji reactions were sent by me
UX suggestion: there's currently no indication which emoji reaction was left by me and which was left by someone else, until I hover over and see my name and the X icon. Having a separate color background for reactions I have added to, and removing the (+) icon to the right if I've already reacted should provide a clearer experience here.

Example: in the screenshot below, the top reaction is one someone did on my post while the bottom is one I did - they look the same at a glance.

Screenshot 2023-03-08 at 10 46 46 AM

patosullivan:
Fixes #2098


arthyn (approved):
lgtm

PR #2293, opened 2023-04-12
devex: Create multiship dev mode

patosullivan:
Add a VITE_SHIP_URL2 and run `npm run multi` or `npm run multichat` to run the dev server for two ships concurrently.


arthyn (approved):
ty this will help a lot with backend changes

PR #2295, opened 2023-04-13
Lay the groundwork for dynamic user themes.

patosullivan:
This PR lays the groundwork to allow dynamic user themes by replacing our hardcoded colors in the tailwind config with CSS custom properties. At the moment, we're just using our existing "default" theme. A theme would consist of both a "light" and "dark" set of colors.

This PR does not add the actual functionality to allow the user to change the colors, but that will be straightforward. We could surface the functionality in settings and stick their theme colors in the settings store for the groups or talk desks. We could also just store their theme settings in localStorage on the browser (via zustand and persist).

Only outstanding issue I've found is the GroupHeader SidebarItem not highlighting properly. @jamesacklin, you worked on that, is there something I'm missing there?


jamesacklin:
Won't delete the branch, just closing the PR for now.

PR #2298, opened 2023-04-13
pwa: Switch to injectManifest strategy for sw.

patosullivan:
This allows us to add custom functionality to the SW later (i.e., web push notifications).

Overall functionality should be exactly the same for now.


arthyn (approved):
seems good as long as `skipWaiting` is ok in our case. I think it is?

patosullivan:
> seems good as long as `skipWaiting` is ok in our case. I think it is?

As long as you see the new service worker running and not waiting it should be good

PR #2299, opened 2023-04-13
Add settings modal, switch to react query for settings state.

patosullivan:
This adds a simple settings modal that currently just displays toggles for various calm settings.

Also switches us over to react-query for settings state.

@urcades I just threw this together to kick this thing off, feel free to critique or point me at some figma files.

Looks like this: 

https://user-images.githubusercontent.com/1221094/231873705-07cc64b6-a944-48be-a767-2e9542b452fd.mov


ui/src/components/icons/PencilSettingsIcon.tsx
```diff
@@ -0,0 +1,19 @@
+import { IconProps } from './icon';
+
+export default function PencilSettingsIcon({ className }: IconProps) {
+  return (
+    <svg
+      className={className}
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 18 18"
```
arthyn, line 8: seems like a weird size you may want to re-export making sure you get the surrounding frame should either be 16 or 24

ui/src/components/SettingsDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { SettingsState, useCalm, useSettingsState } from '@/state/settings';
+import { isTalk } from '@/logic/utils';
+import Dialog from './Dialog';
+import Setting from './Setting';
+
+async function toggle(property: keyof SettingsState['calmEngine']) {
+  const selProp = (s: SettingsState) => s.calmEngine[property];
+  const state = useSettingsState.getState();
+  const curr = selProp(state);
+  await state.putEntry('calmEngine', property, !curr);
```
arthyn, line 11: I think putEntry uses poke under the hood, if we switched it to `trackedPoke` progress would be a little more accurate.

arthyn (approved):
lgtm, two suggestions

patosullivan:
Made @arthyn's suggested changes, now using trackedPoke and an appropriately sized icon:

image

Also worth noting that the calm settings here are *per desk*, not universal. I.e., if you set something in this modal it will just be for groups or talk.

ui/src/state/settings.ts
```diff
@@ -161,8 +176,39 @@ export const useSettingsState = createState<BaseSettingsState>(
...
+              'entry-key': key,
+              value: val,
+            },
+          },
+        },
+        {
+          app: 'settings-store',
+          path: `/desk/${window.desk}`,
+        },
+        (event) => {
```
arthyn, line 196: I think you may be able to omit this checking function and it will still work since the subscription already has the reduce and the default checker on tracked poke does a deep equal

urcades:
> Also worth noting that the calm settings here are per desk, not universal. I.e., if you set something in this modal it will just be for groups or talk.

This is perfect, exactly how I'd imagine this to work as someone encountering it!

I only have a few minor suggestions, mostly around spacing:

Screenshot 2023-04-14 at 10 07 40 AM

Screenshot 2023-04-14 at 10 08 45 AM

Essentially, we should try and ensure that "settings forms/pages" adhere to a wider 32px spacing scheme around major section elements, and set items within sections to 16px — this makes it a little easier to visually parse which sections field clusters belong to.

urcades:
In addition to minor spacing updates, I have a few additional thoughts around changing the terminology/copy of these interfaces to make it more clear this is Groups-scoped:

Overview:

App Settings Modal

Details:

We can stand to update the label in the dropdown to state "App Settings", which make it a little clearer that the resulting interface is for this particular app, and nothing else. We probably shouldn't use "Groups Settings" since the objects contained in the app are called Groups, and people might figure that this interface is targeting actual groups, not the Groups app.

Screenshot 2023-04-14 at 10 30 49 AM

Next, are a few instances of clarifying the language of the fields' labels in order to make it clear that the settings are scoped to the app itself:

Screenshot 2023-04-14 at 10 33 29 AM

You can copy out the text for these fields from [Figma](https://www.figma.com/file/QMM0PcJyuKWrXY2n4VpQp1/2023-Q2-Groups?node-id=122%3A25788&t=vjAMUj1GAAJYwYfU-1), using the inspect interface.

urcades (changes requested):
I forgot to leave a review status

patosullivan:
image

ui/src/state/settings.ts
```diff
@@ -300,27 +324,58 @@ export function useDiaryCommentSortMode(flag: string): 'asc' | 'dsc' {
...
 }
 
-const selLoaded = (s: SettingsState) => s.loaded;
-export function useSettingsLoaded() {
-  const loaded = useSettingsState(selLoaded);
-  return loaded;
+export function useMessagesFilter() {
+  const { data, isLoading } = useSettings();
+
+  if (isLoading || data === undefined || data.groups === undefined) {
```
arthyn, line 365:
```suggestion
  if (isLoading || data === undefined || data.talk === undefined) {
```

arthyn (approved):
just need to commit that suggestion then gtg

PR #2301, opened 2023-04-13
reconnect: fan out refetching a little so we don't choke reconnection

arthyn:
OTT, this should make reconnects a little smoother. We have a little more room here we can play with if this isn't enough.


patosullivan (approved):
Makes sense to me

PR #2308, opened 2023-04-14
Add loading state for initial group load

### LINKED ISSUE #2306: groups: joining a group has no loaders
A feeling of helplessness and confusion accompanies joining a new group, even if the group "join" succeeds. Most of this boils down to the following conditions:
(A) you've joined the group
(B) you've navigated to the new group
(C) you're waiting for more content to fill itself in

image

Sometimes, the group name populates here. Sometimes, it doesn't. We should fix this.

**To Reproduce**
1. Join a new group
2. Allow yourself to be navigated to the group
3. Observe empty/questionable group state while content gets downloaded from the host

**Expected behavior**
I would expect some kind of loading indicator. A centered spinner in the main stage at the very least would be appropriate here. Lots of apps generate some silly technobabble "Reticulating splines..." to keep you around while waiting -- we could do something more straightforward "Downloading group data from host..." _or_ we could fake out the actual dojo response one would expect to see when transferring this data across the network. Up to us, really -- anything but a blank screen.

patosullivan:
Fixes #2306

Also tweaked the queryKeys (use 'groups', not 'group' for useGroup) and tweaked the behavior of useGroupLeaveMutation due to some weirdness I saw around group leaves.

### DIFF
--- a/ui/src/groups/GroupSidebar/ChannelList.tsx
+++ b/ui/src/groups/GroupSidebar/ChannelList.tsx
@@ -12,7 +12,6 @@ import {
 } from '@/logic/utils';
 import { useIsMobile } from '@/logic/useMedia';
 import { useGroup, useGroupFlag, useVessel } from '@/state/groups';
-import CaretDown16Icon from '@/components/icons/CaretDownIcon';
 import SortIcon from '@/components/icons/SortIcon';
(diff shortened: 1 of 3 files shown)


arthyn (approved):
makes sense to me

PR #2309, opened 2023-04-14
diary/heap: fix initialization and stabilize refs

arthyn:
There were multiple issues here. First of all outlines and notes were missing the type property which is what we were using to differentiate between the two. Second of all, we were persisting our rich map types, but those weren't rehydrating when the app opens so we weren't actually getting any benefit from those. 

The main issue though was that the initialize call was not gated anymore because we were implicitly using the "is subbed" flag previously to also prevent subsequent initializations. However to work with the new reconnect behavior we needed allow it to be called multiple times. So to stabilize it, I've made it check to see if it already has all the new values and if it does it simply doesn't update the map and therefore won't trigger new renders.

Also I found a huge source of slowness, groups were too aggressively subscribing we should probably be cautious about granular subscriptions and default them to off.

I'm still working on a few things here, but this should fix the issues we were seeing.

### DIFF
--- a/ui/src/diary/DiaryChannel.tsx
+++ b/ui/src/diary/DiaryChannel.tsx
@@ -117,7 +117,6 @@ function DiaryChannel() {
     ? canReadChannel(channel, vessel, group?.bloc)
     : false;
 
-  console.log(letters.size, joined, joining, channel, canRead);
 
   useEffect(() => {
     if (!joined) {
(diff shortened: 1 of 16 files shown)


ui/src/diary/DiaryChannel.tsx
```diff
@@ -116,6 +117,8 @@ function DiaryChannel() {
     ? canReadChannel(channel, vessel, group?.bloc)
     : false;
 
+  console.log(letters.size, joined, joining, channel, canRead);
```
arthyn, line 120:
```suggestion
```

patosullivan (approved):
Looks good to me and runs well locally.

PR #2321, opened 2023-04-17
skeins: Use appropriate queryKeys

### LINKED ISSUE #2320: Notifications aren't marked as read without refresh
When I click 'Mark as read' I still have unread notifs.

**To Reproduce**
1. Go to the root Groups screen
2. Click on Mark as Read
3. See error

**Expected behavior**
The notifs should be marked as read

Here's a video: 
https://drive.google.com/file/d/1GwqpbG_WNV0OQT5wwmuPMd2pbA2kRcdq/view?usp=share_link

**Additional context:**
- OS/browser Safari on MacOS

patosullivan:
We were using 'undefined' in cases where we weren't fetching skeins for a whole desk, just a group. We should just use the desk name.

We also weren't invalidating/cancelling existing skeins queries for groups.

The first issue may have been causing some weird behavior, the second issue definitely prevented a user from marking all groups activity updates as read.

Should fix #2320

### DIFF
--- a/ui/src/state/hark.ts
+++ b/ui/src/state/hark.ts
@@ -62,7 +62,7 @@ export function useBlanket(flag?: Flag) {
 
 export function useSkeins(flag?: Flag) {
   const { data, ...rest } = useReactQuerySubscription({
-    queryKey: ['skeins', flag ? flag : undefined],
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

Issue #2235, opened 2023-04-03
talk: dm disappeared from sidebar (until refresh)


Fang-:
Had gone afk for just a minute, returned to my talk tab, and wanted to view the dm I had had just a little while ago with ~habsul-rignyr. But.. it wasn't in my sidebar anymore?

After a page refresh, it showed up just fine again.

![chat-gone](https://user-images.githubusercontent.com/3829764/229566725-a38b5d34-c383-4646-98a1-bad79dd4206f.gif)

Not sure what caused this. Maybe it just got sorted poorly, and was somewhere way down? A quick manual search didn't reveal it though.

arthyn:
this should be fixed as of a few releases ago

PR #2327, opened 2023-04-18
chat: restore missing DM invites

arthyn:
While building the marks for the new UI init scries I forgot to change the ship gate which oddly converts patps without the sig. Because this didn't error since it's in `enjs:format` I didn't realize, and previous DMs still showed up so everything looked fine. Possibly the cause of #2235 and maybe also Jack's issue in the support email?


tomholford (approved):
LGTM

patosullivan (approved):

PR #2333, opened 2023-04-18
chat: switch to club-action-0 for backwards compat with old clients

arthyn:
OTT, this helps old clients stay functional by keeping `club-action` using `echo` and we switch new clients to `club-action-0` only for sending. This does introduce somewhat of an asymmetry with the marks since `club-action-0` is only used for clients, and when `club-action` gets encoded to json it uses `uid` not `echo`. However everything still works with just this change so I'd prefer to not alter things further.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -318,7 +318,7 @@
       %club-create
     cu-abet:(cu-create:cu-core !<(=create:club:c vase))
   ::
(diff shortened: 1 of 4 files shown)


alecananian:
Related Small Talk PR: https://github.com/tloncorp/pocket/pull/312

patosullivan (approved):

PR #2347, opened 2023-04-19
groups: reconnect and perf

arthyn:
This PR addresses some issues around reconnecting seen recently where things get locked up or take too long. Turns out we were doing too much here. React query handles a lot of this for us. However, because we do some of the work in the `init` scry we should turn off reconnect handling for queries updated by that.

Also when reconnecting we have a possibility of getting a large number of facts in a row. With the subscription handling as written this would cause an enormous amount of scries to happen. To combat this the subscription handling will now use a debounce that fires both on the leading and trailing edge so we get immediate response, but if a lot more is happening they all get delayed until the final call.

### DIFF
--- a/ui/src/state/bootstrap.ts
+++ b/ui/src/state/bootstrap.ts
@@ -163,6 +163,12 @@ useLocalState.setState({
     bootstrap('reset');
 
     useLocalState.setState({ lastReconnect: Date.now() });
-    wait(() => queryClient.invalidateQueries(), 5);
+    wait(
+      () =>
+        queryClient.invalidateQueries({
+          refetchType: 'none',
+        }),
+      5
+    );
   },
 });
--- a/ui/src/state/bootstrap.ts
+++ b/ui/src/state/bootstrap.ts
@@ -158,13 +158,10 @@ export default async function bootstrap(reset = 'initial' as Bootstrap) {
 
 useLocalState.setState({
   onReconnect: () => {
-    // let event stream finish resolving before starting requests
-    setTimeout(() => {
-      const { reset } = useSchedulerStore.getState();
-      reset();
-      bootstrap('reset');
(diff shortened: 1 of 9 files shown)


ui/src/state/groups/groups.ts
```diff
@@ -76,21 +80,35 @@ export function useGroups() {
...
+  const { data, ...rest } = useReactQueryScry({
+    queryKey,
     app: 'groups',
-    path: `/groups/${flag}/ui`,
-    initialScryPath: `/groups/${flag}`,
-    enabled: !!flag && flag !== '' && withMembers && subscribe,
-    initialData: group,
+    path: `/groups/${flag}`,
     options: {
+      enabled: !!flag && flag !== '' && withMembers && subscribe,
```
patosullivan, line 93: I missed this before, but we can't just set withMembers to be true, we must always set both withMembers and subscribe?

ui/src/state/bootstrap.ts
```diff
@@ -158,11 +158,13 @@ export default async function bootstrap(reset = 'initial' as Bootstrap) {
...
 useLocalState.setState({
   onReconnect: () => {
-    const { reset, wait } = useSchedulerStore.getState();
-    reset();
-    bootstrap('reset');
-
-    useLocalState.setState({ lastReconnect: Date.now() });
-    wait(() => queryClient.invalidateQueries(), 5);
+    // let event stream finish resolving before starting requests
+    setTimeout(() => {
```
patosullivan, line 162: Why does setTimeout do that for us?

patosullivan (approved):
Looks good, just a couple questions but I imagine you did both of these things for good reasons, so it should be good to merge.

patosullivan (approved):
changes look good

Issue #555, opened 2022-08-05
groups: per-channel cordon


jamesacklin:
For each channel, users need to be added to one of several roles:

- **Viewer**: can only read the channel's contents
- **Commenter**: can comment on items in the channel (add to thread but only on existing items)
- **Contributor**: can add items (write) to the channel, as well as edit/delete said items
- **Editor**: all of the above, plus edit and delete others' items

Group admin abilities supersede all of these permissions.

jamesacklin:
Additional request in #1766

jamesacklin:
Also implicated: #1572

lukestiles:
Editor sounds like moderator in this context. </bikeshed>

PR #2350, opened 2023-04-20
groups: add role management

patosullivan:
This adds the ability to add custom roles and assign read/write permissions for channels based on membership in those roles.

https://user-images.githubusercontent.com/1221094/233364279-26f2355e-ced7-4959-9892-57bed087653e.mov


urcades (changes requested):
This is looking good — I'm going to scope my feedback accordingly so that I'm only covering immediate-term changes we can make to the interface we're surfacing, not any deeper functionality brought up in yesterday's call.

What I'm addressing below is:

1. Improving the display + setting of roles in the member's list
2. Preparing the Role Add/Edit experience to more readily accept the updates we'll want to push later

## Setting Roles in Member's List

One immediate point of UX improvement I think we can ship soon is to slightly reconfigure the presentation of "Members lists" role display. At the moment, selecting roles in the members list is done via a dropdown, which additionally updates its label to show what role is applied to the member. If more than two roles are applied, this interface becomes illegible, so I think we can move the display of applied roles to the left side of the member row:

**Current interface:**

Screenshot 2023-04-25 at 12 05 36 PM

**Recommended update:**

Screenshot 2023-04-25 at 11 54 13 AM

Coupled with this change in the placement of role labels, we can update the dropdown to take on an iconographic appearance:

Screenshot 2023-04-25 at 11 55 24 AM

I'm not sure if we want to tackle the action list in this PR, but it's worth noting that we could mirror the above role-setting functionality in the "action list" for each individual member:

Screenshot 2023-04-25 at 11 56 25 AM

I think these combined improvements will promote a more legible interface that should work on smaller screens as well.

## Role Add/Edit Page

I don't have any immediate improvements for the role-adding/editing experience currently implemented, but I think if we want to smooth over the current implementation and a more feature-rich permissions-oriented implementation, we might want to structure the role editing experience as so:

**Current interface:**

Screenshot 2023-04-25 at 12 09 11 PM

**Recommended update:**

Screenshot 2023-04-25 at 11 59 28 AM

Instead of surfacing "Edit" and "Delete" inline with the role, we can provide an arrow here, which indicates that editing the role happens in another screen/page.

This role editing page could be quite involved (as we've already covered), but I think we can still keep it pretty simple now, and only allow for the editing of the role name, its description, and the ability to delete it. Regardless of what we place in this page, it should roughly be structured as so:

Screenshot 2023-04-25 at 12 03 21 PM

- Role Details
- Role Permissions (for later)
- Delete Role

jamesacklin:
@urcades Ideally, we would ship these two together to the public. It may be ugly for the time being until we refactor the group settings interface/modal.

urcades:
@patosullivan I added some notes on the placement of role-limiting read/write access here:

https://github.com/tloncorp/landscape-apps/issues/2348#issuecomment-1522417426

Lmk if you have any questions: The first iteration (single scroll) is close to what you've implemented, but if we're going with a paged design for Groups Settings, it might be worth our while to segment out Channel Settings similarly.

jamesacklin:
Fixes #555?

patosullivan:
Seems to.

patosullivan:
@urcades how are we looking? 

We can't set colors or custom icons the way things currently work with roles, but I think this is about what you wanted. Lmk if there's anything we need to change here.

https://user-images.githubusercontent.com/1221094/236928059-86d6a3cb-8906-4d91-a8f1-e6d2379361fc.mov

patosullivan:
> Fixes #555?

@jamesacklin just note that we won't have the ability to assign moderator permissions to anyone other than admins at this point. just read/write.

urcades (approved):
This is excellent, changes looking good

arthyn (approved):
lgtm

PR #2351, opened 2023-04-20
bugfix: Fix copy to clipboard in safari.

### LINKED ISSUE #1767: groups: copying references on mobile (iOS) works only about half the time
- Be in a group chat (urbit help), copy a chat reference
- Go to a different group chat (tlon support)
- Attempt to paste the reference
- Observe that instead of the reference you see old content from your clipboard pasted

I think the reference is not getting copied despite it turning into a checkmark, when this occurs it doesn't work even if attempting to copy multiple times.

### LINKED ISSUE #2346: Copy Group Link broken in Groups UI
The copy group link action in the upper left Groups nav is not populating the copy-paste buffer. 

**To Reproduce**
1. Copy text into copy-paste buffer
2. Click upper left Groups nav to open menu
3. Click "Copy Group Link"
4. Paste in a DM
5. Previously copied text from step 1 is still in buffer

**Expected behavior**
The Landscape-friendly Group link should be pasted.

**Additional context:**
Currently synced to marnec

patosullivan:
Switches to usehooks-ts's useCopyToClipboard copy function rather than our own writeText function (which didn't account for safari weirdness) within our useCopy hook.

Fixes #2346
Fixes #1767

### DIFF
--- a/ui/src/logic/utils.ts
+++ b/ui/src/logic/utils.ts
@@ -29,6 +29,7 @@ import {
 import { CurioContent, Heap, HeapBrief } from '@/types/heap';
 import { DiaryBrief, DiaryQuip, DiaryQuipMap } from '@/types/diary';
 import bigInt from 'big-integer';
+import { useCopyToClipboard } from 'usehooks-ts';
 
 export const isTalk = import.meta.env.VITE_APP === 'chat';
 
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

PR #2352, opened 2023-04-20
bugfix: Fix for clicking on names in ship ShipSelector

### LINKED ISSUE #2341: ship selector: can't click on names in the menu
May be related to portal-ing menu

patosullivan:
react-select's built in 'menuPortal' feature somehow prevented the user from being able to click on items in the menu.

I added the menuPortal changes because the menu's opening would cause the dialogs containing the ShipSelector to move, causing the ShipSelector itself to move out of view.

Rather than use the menuPortal feature, we can just add bottom margin to the dialog container. This fixes both issues.

Fixes #2341


arthyn (approved):
lgtm

PR #2353, opened 2023-04-20
bugfix: make send button in heap detail clickable

### LINKED ISSUE #2342: gallery: send button doesn't work on comments in some galleries

patosullivan:
A previous change to heap detail styles made it so that the div containing the CaretRightIcon/previous button was occluding the send button in heap detail. The div containing the left/right carets (next/prev buttons) should have had group and relative classes in order for the child divs to render/behave properly, i.e., set opacity to 100 on hover and appear on the sides of the HeapDetailBody (and not occlude the items on the far left of the screen, including the send button).

Fixes #2342


arthyn (approved):
lgtm

PR #2361, opened 2023-04-22
chat: add full-text and mentions search

ashelkovnykov:
Extended version of the code originally published by @liam-fitzgerald in #1278.

Tested on a set of moons for DMs + chat, then tested on my own ship. Tests documented below. All times are the average of 10 runs.

| dm / chat        | # messages |
|------------------|------------|
| `~master-morzod` | `1.010`    |
| `~datder-sonnet` | `6.997`    |
| Tlon Watercooler | `41.119`   |

| test | dm / chat | Avg. Time (ms) |
|------|-----------|----------------|
| find all instances of 3-character string that doesn't exist | `~master-morzod` | 142 |
| find all instances of 15-character string that doesn't exist | `~master-morzod` | 123 |
| find first 1000 messages that contain the letter 'e' | `~master-morzod` | 29 * |
| find all instances of 3-character string that doesn't exist | `~datder-sonnet` | 754 |
| find all instances of 15-character string that doesn't exist | `~datder-sonnet` | 640 |
| find first 1000 messages that contain the letter 'e' | `~datder-sonnet` | 34 * |
| find all instances of 3-character string that doesn't exist | Tlon Watercooler | 4.303 |
| find all instances of 15-character string that doesn't exist | Tlon Watercooler | 3.307 |
| find first 1000 messages that contain the letter 'e' | Tlon Watercooler | 32 * |
| find first 10 mentions of `~ravmel-ropdyl` | Tlon Watercooler | 86 |

*: That's how long it took to find the results, but mark conversion and rendering to dojo took significantly longer

Example scries:
```
=chat-types -build-file /=groups=/sur/chat/hoon
.^((list writ:chat-types) %gx /=chat=/dm/(scot %p ~datder-sonnet)/search/text/0/5/['012345678901234']/chat-scan)
.^((list writ:chat-types) %gx /=chat=/dm/(scot %p ~datder-sonnet)/search/text/0/(scot %ud 1.000)/['e']/chat-scan)
.^((list writ:chat-types) %gx /=chat=/chat/(scot %p ~bolbex-fogdys)/['watercooler-4926']/search/text/0/100/['yzy']/chat-scan)
.^((list writ:chat-types) %gx /=chat=/chat/(scot %p ~bolbex-fogdys)/['watercooler-4926']/search/mention/0/10/(scot %p ~ravmel-ropdyl)/chat-scan)
```
Prepend with `~>  %bout` to measure wall-clock run time of scry.

Notes:
- Currently, the scry path searches from oldest to newest
- Search was useful for its own testing; I forgot how to measure wall-clock run time of arbitrary Hoon in dojo, but I remembered that `~master-morzod` had told me, so I was able to search my chat history for the answer
- Empty string input (`''`) for the text search matches all messages

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -1243,9 +1243,32 @@
   ++  ca-peek
     |=  =(pole knot)
     ^-  (unit (unit cage))
-    ?+  pole  [~ ~]
-      [%writs rest=*]  (peek:ca-pact rest.pole)
-      [%perm ~]        ``chat-perm+!>(perm.chat)
+    ?+    pole  [~ ~]
+        [%writs rest=*]
+      (peek:ca-pact rest.pole)
+    ::
+        [%perm ~]
+      ``chat-perm+!>(perm.chat)
+    ::
+        [%search %text skip=@ count=@ nedl=@ ~]
+      %-  some
+      %-  some
+      :-  %chat-scan
+      !>
+      %^    text:search:ca-pact
+          (slav %ud skip.pole)
+        (slav %ud count.pole)
+      nedl.pole
+    ::
+        [%search %mention skip=@ count=@ nedl=@ ~]
+      %-  some
+      %-  some
+      :-  %chat-scan
+      !>
+      %^    mention:search:ca-pact
+          (slav %ud skip.pole)
+        (slav %ud count.pole)
+      (slav %p nedl.pole)
     ==
   ::
   ++  ca-revoke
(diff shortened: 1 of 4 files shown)


mcevoypeter:
Logging for posterity from a conversation Alex and I had offline: is it possible to incrementally surface search results? For example, if I'm searching through a channel with 41 million messages, can the search implementation surface a match the instance it finds one so I don't have to wait for the entire search to run before seeing some of my results?

arthyn:
@mcevoypeter we'd have to change this to more of a "job" style flow where you initiate a search which gets assigned some id that we can then subscribe to. then we could have it start searching chunks at a time progressively, and sending facts as it finds things. 

it doesn't seem crazy, but def should be a v2. the other thing to consider is that if we started indexing on write, the one shot scry may be fast enough for all use cases even really large ones since it would only be traversing a small set of already processed data.

ashelkovnykov:
@arthyn The direction I was thinking while speaking with @mcevoypeter is that after this version but before indexing, it's also possible to cache the partial search and resume searching from where we left off. The code to do so would be very simple, and we're already returning the partial-search `mop` - we just throw it away instead of caching it.

PR #2372, opened 2023-04-24
bugfix: use correct lastReply for chat reply time

### LINKED ISSUE #1868: chat: "last reply" times for threads often off by hours
OTT

patosullivan:
Fixes #1868 by sorting replies by time client-side.

### DIFF
--- a/ui/src/chat/ChatMessage/ChatMessage.tsx
+++ b/ui/src/chat/ChatMessage/ChatMessage.tsx
@@ -4,7 +4,7 @@ import cn from 'classnames';
 import _, { debounce } from 'lodash';
 import f from 'lodash/fp';
 import { BigInteger } from 'big-integer';
-import { daToUnix } from '@urbit/api';
+import { daToUnix, udToDec } from '@urbit/api';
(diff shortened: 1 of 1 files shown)


ui/src/chat/ChatMessage/ChatMessage.tsx
```diff
@@ -153,8 +153,14 @@ const ChatMessage = React.memo<
         f.uniq,
         f.take(3)
       )(seal.replied);
-      const lastReply = seal.replied.length ? _.last(seal.replied)! : '';
-      const lastReplyWrit = useWrit(whom, lastReply)!;
+      const repliesSortedByTime = seal.replied.sort((a, b) => {
+        const aTime = udToDec(a.split('/')[1]);
```
arthyn, line 157: probably should use `useChatState.getState().getTime(a)` and then use big int compare so we at least try to use server time if we have it.

arthyn (changes requested):

ui/src/chat/ChatMessage/ChatMessage.tsx
patosullivan: ok, just switched to times from getTIme().

PR #2387, opened 2023-04-25
grouper: use add-ships for shut cordon

arthyn:
This fixes the lure invites for private/secret groups. We should likely change the flow of invites so that they all flow through the group-invite mark instead of forcing people to know about the cordon add-ships stuff. Also re-enables lure on the frontend.

### DIFF
--- a/desk/app/grouper.hoon
+++ b/desk/app/grouper.hoon
@@ -79,8 +79,23 @@
     ?>  (~(has in enabled-groups) token.bite)
     ?>  ?=([%bite-1 *] bite)
     =/  =invite:groups  [[our.bowl token.bite] joiner.bite]
-    :_  this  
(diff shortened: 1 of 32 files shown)


desk/app/grouper.hoon
```diff
@@ -79,8 +79,23 @@
...
+    ?+  -.cordon.group  ~
+        %open
+      ~[[%pass /invite %agent [our.bowl %groups] %poke %group-invite !>(invite)]]
+    ::
+        %shut
+      ~&  ['inviting' joiner.bite]
+      =/  =action:groups  
+        :-  [our.bowl token.bite]
+        :-  now.bowl
+        :-  %cordon
```
midsum-salrux, line 95: NB: `:*` is more idiomatic

midsum-salrux (approved):
LGTM

arthyn:
closing in favor of #2391

PR #2390, opened 2023-04-25
bugfix: no duplicate message options on thread OPs

### LINKED ISSUE #2383: talk: reaction button unreliable for threaded messages
Depending on the UI state, the "React" button for messages that are threaded may or may not work.

**To Reproduce**
1. Go to a Chat channel or a DM.
2. Click "#" to open a thread.
3. Attempt to react to the root post in either the thread pane or the main message log.
4. Observe that the button does not work.
5. Type a few replies. Allow someone else to reply in the thread as well.
6. Click "React" on their post and observe the misplaced emoji reaction picker.

These buttons are inoperable when the UI is in this state:

Screenshot 2023-04-25 at 12 59 50 PM

However, this one works (a follow-up message):

Screenshot 2023-04-25 at 1 00 28 PM

Incorrect positioning of the emoji picker in a follow-up message:

Screenshot 2023-04-25 at 1 02 01 PM

patosullivan:
We need to check whether the message we're rendering is a thread OP and not allow for it to display message options unless it's being rendered in the main chat scroller, otherwise ChatMessageOptions renders twice and the user won't be able to react to the message.

Fixes #2383

### DIFF
--- a/ui/src/chat/ChatMessage/ChatMessage.tsx
+++ b/ui/src/chat/ChatMessage/ChatMessage.tsx
@@ -6,7 +6,7 @@ import f from 'lodash/fp';
 import { BigInteger } from 'big-integer';
 import { daToUnix } from '@urbit/api';
 import { format, formatDistanceToNow, formatRelative, isToday } from 'date-fns';
-import { NavLink } from 'react-router-dom';
+import { NavLink, useParams } from 'react-router-dom';
 import { useInView } from 'react-intersection-observer';
 import { ChatBrief, ChatWrit } from '@/types/chat';
 import Author from '@/chat/ChatMessage/Author';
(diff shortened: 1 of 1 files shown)


ui/src/chat/ChatMessage/ChatMessage.tsx
```diff
@@ -79,10 +79,18 @@ const ChatMessage = React.memo<
...
+        idTime: string;
+      }>();
+      const isThread = idShip && idTime;
+      const threadOpId = isThread ? `${idShip}/${idTime}` : '';
+      const isThreadOp = threadOpId === seal.id && hideReplies;
       const chatInfo = useChatInfo(whom);
       const unread = chatInfo?.unread;
       const unreadId = unread?.brief['read-id'];
       const { hovering, setHovering } = useChatHovering(whom, writ.seal.id);
+      console.log({ hovering, whom, id: writ.seal.id });
```
arthyn, line 93:
```suggestion
```

arthyn (approved):
makes sense, lgtm

PR #2391, opened 2023-04-25
grouper: use add-ships for shut cordon

arthyn:
This fixes the lure invites for private/secret groups. We should likely change the flow of invites so that they all flow through the group-invite mark instead of forcing people to know about the cordon add-ships stuff. Also re-enables lure on the frontend.

Also adds resubscribing to reel in case it gets dropped.


midsum-salrux (approved):
LGTM

PR #2398, opened 2023-04-26
ops: upload only necessary files to artifacts

patosullivan:
Avoids uploading JS source and node modules to github artifacts.


arthyn (approved):
lgtm

PR #2426, opened 2023-04-27
chat: reappearing dm invites

arthyn:
This should fix the issue people are seeing where DMs have to be accepted multiple times. We had added functionality for people to be able to load the apps offline which would load locally stored state, but that was erroneously getting merged with latest state even if you weren't offline. This rips that functionality out since we aren't remotely offline compatible yet.

Also throws in a few things, like setting a line-height on the Authors where we're doing line-clamping so it doesn't cut off things like tall emoji, and also stops channels from showing a loader when unnecessary like moving from a dm to a channel that was previously already loaded.


patosullivan (approved):
Looks good.

PR #2427, opened 2023-04-27
fix for group icon svg that wasn't compliant with CSP on tlon.network

patosullivan:
tlon.network implemented a content security policy that did not allow for setting a base64 string as a src on an image, this caused the groups icon to look broken on notifications on tlon hosted ships.


arthyn (changes requested):
we should just re-export the default group icon to make sure it's right (not using an embedded base64 image) and use that

patosullivan:
Switched to using the existing AppGroupsIcon, matched the styling, looks like this:

image

arthyn (approved):
lgtm

PR #2428, opened 2023-04-27
meta: 413 compat

arthyn:
This gives us compatibility with kelvin 413 and also allows us to control which branch/tag the deploy uses when installing urbit repo deps.


joemfb (changes requested):
`:contacts` uses `%gu` to check if `:contact-store` is running: https://github.com/tloncorp/landscape-apps/blob/80cebfa1be2e6cb8ef896bc16e6898b23d4fde60/desk/app/contacts.hoon#L402

That path needs to change for 413k

desk/app/contacts.hoon
```diff
@@ -398,7 +398,7 @@
         --
     ::
     ^+  cor
-    =/  bas  /(scot %p our.bowl)/contact-store/(scot %da now.bowl)
+    =/  bas  /(scot %p our.bowl)/contact-store/(scot %da now.bowl)/$
     ?.  .^(? gu+bas)  cor
     =/  ful  .^(rolodex:legacy gx+(weld bas /all/noun))
```
joemfb, line 403: The %gu scry needs to change, but the %gx does not. You should revert the change to `bas` and just ` .^(? gu+(weld bas /$))`. This one's a pain ...

arthyn: ah yeah you're right ok

joemfb (approved):
LGTM wrt `:contacts`. I also grepped for other %gall scrys, didn't see any affected.

arthyn:
@joemfb tysm

Issue #2431, opened 2023-04-28
Cannot read property 'title' of undefined


ilyakooo0:
All I did was open groups.

Desk hash: `0v11.kq88i.9hcnd.c1mei.b3mfo.4vgqd.r3qss.04st6.pimdi.p39ol.ajmpq`

```
TypeError: Cannot read property 'title' of undefined
    at https://urbit.iko.soy/apps/groups/assets/index.c13885b0.js:48:16464
    at Kl (https://urbit.iko.soy/apps/groups/assets/react-beautiful-dnd.fde4d763.js:19:52884)
    at Ka (https://urbit.iko.soy/apps/groups/assets/react-beautiful-dnd.fde4d763.js:19:62361)
    at Zd (https://urbit.iko.soy/apps/groups/assets/react-beautiful-dnd.fde4d763.js:19:61688)
    at Ls (https://urbit.iko.soy/apps/groups/assets/react-beautiful-dnd.fde4d763.js:19:61273)
    at mf (https://urbit.iko.soy/apps/groups/assets/react-beautiful-dnd.fde4d763.js:23:11906)
    at pf (https://urbit.iko.soy/apps/groups/assets/react-beautiful-dnd.fde4d763.js:23:947)
    at yg (https://urbit.iko.soy/apps/groups/assets/react-beautiful-dnd.fde4d763.js:23:875)
    at rn (https://urbit.iko.soy/apps/groups/assets/react-beautiful-dnd.fde4d763.js:23:730)
    at sl (https://urbit.iko.soy/apps/groups/assets/react-beautiful-dnd.fde4d763.js:21:10884)
```

patosullivan:
This was just at the index of groups, right? like `https://urbit.iko.soy/apps/groups/`, not some particular route?

ilyakooo0:
Just https://urbit.iko.soy/apps/groups/ yields an error

But now it is different:

Encountered error:Cannot use 'in' operator to search for 'shut' in undefined
```
TypeError: Cannot use 'in' operator to search for 'shut' in undefined
    at Gd (https://urbit.iko.soy/apps/groups/assets/index.c13885b0.js:1:45332)
    at dn (https://urbit.iko.soy/apps/groups/assets/index.c13885b0.js:1:45449)
    at https://urbit.iko.soy/apps/groups/assets/index.c13885b0.js:48:12098
[...]

PR #2436, opened 2023-04-28
bugfix: check for group.cordon before checking for shut

patosullivan:
Partial fix for #2431 

This should fix the second issue that the user saw. I think the issue was actually caused by getPrivacyFromCordon which is ultimately being called by useGroupJoin, which is called by FindGroups. I went ahead and added checks for the existence of group.cordon elsewhere. 

Still not sure where the 'Cannot read property 'title' of  undefined' error was coming from. 

I'll let him know to update the ticket if he sees the first issue again and can produce repro steps.

### DIFF
--- a/ui/src/groups/GroupAdmin/GroupMemberManager.tsx
+++ b/ui/src/groups/GroupAdmin/GroupMemberManager.tsx
@@ -32,7 +32,7 @@ export default function GroupMemberManager() {
       return [];
     }
     return Object.keys(group.fleet).filter((k) => {
-      if ('shut' in group.cordon) {
+      if (group.cordon && 'shut' in group.cordon) {
         return (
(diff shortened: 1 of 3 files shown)


arthyn (changes requested):
This means we have a type that's faulty because you should never have a group without a cordon. I think the cause is the way we do:

```
  if (rest.isLoading || rest.isError) {
    return undefined;
  }

  return {
    ...(data as Group),
  };
  ```

in `useGroup`. this if statement should also check if data is undefined, because if it is then you'll end up doing `{...(undefined)}` which just returns `{}` so our normal checks won't catch it and neither will our types. So instead of putting checks that go against our types everywhere, we should just fix this.

patosullivan:
Closed in favor of #2437

PR #2438, opened 2023-04-28
chat: adding search capabilities

### LINKED ISSUE #2249: groups/talk: search interface recommendations
In speaking with James and Luke on the subject of Q2's Search implementation, I've come to understand that it's probably ideal to let engineering lead with this particular feature set, rather than circumscribe or prematurely limit capability here. 

We all understand channel search is incredibly important, and I believe we're all aware of what a ubiquitously "good" search experience feels like. Search speed and overall performance are probably the biggest elements of the UX to finesse here.

That said, I think it's still important to properly situate/contextualize where we inject search, so I've taken some time to illustrate some simple base cases/placement for a search-triggering interface:

Search Placement

Read this from left to right, top to bottom.

I've illustrated five potential placements for our search interface. I think out of all of them, the Leap-integrated approach is my first choice, but either of the three preceding options (a search field integrated into the header) are acceptable for a v1.

I'd recommend strongly against situating search interface + results in a sidebar. After thinking about it, I realized we might get some collision or undesired "stackiness" with threads. Regardless of the placement we decide upon, I recommend strongly that we situate channel search triggering as a magnifying glass in the upper right hand side of a channel's header, as a general affordance (like the "..." menu). 

We can "sneak" group-wide search into the channel search interface and just heavily prioritize "in-channel" results, but if we want to create a distinct location for group-scope search, the group sidebar is probably the place to do it. This grants us the ability to have a very solid mobile navigation which repeats itself across contexts:

Equivalent placements for "Find/Discover Groups" and "Group Search", with "Channel Search" remaining the same:

Screenshot 2023-04-05 at 2 51 07 AM

arthyn:
This PR adds a UI to the work that @ashelkovnykov did to get search going. I used one of @urcades recommendations in #2249 to take up the entire header which maps nicely to mobile. You can see it used here:

https://user-images.githubusercontent.com/5466421/235269888-60ac7ad2-d90b-465c-92f0-34ce72a0c886.mov

Notably this does not include recent searches, mention support, keyboard navigation, or even loading results past the first 1000 instances, but it at least gets us going. This does however highlight our poor ability to navigate in backscroll because of the current thread architecture so we should probably do those updates immediately following this.

@ashelkovnykov one thing I did notice is that the search is currently case sensitive we should probably change that.

Fixes #2249
Fixes LAND-418


urcades (approved):
This is huge. I’m in full support of the implementation and placement, and feel like this is a great base to get started with and merge for stress testing purposes.

I’m already noting that the visual-interface similarity of the search results with chat itself makes for a tiny amount of potential confusion, but we can distinguish the search results properly (as in, place search results in a distinctly “search”-looking interface) when we begin to address the various updates you mentioned.

urcades:
@jamesacklin i can come up with something simple in a few hours, well before EOD

urcades:
I actually had these designs completed earlier today, but Dan and I were tracing through some incredibly distracting (but great) history of our work. My apologies.

Search Results

The ideal state I'd want us to push on is one in which we design the search results to be highly distinct as an interface that lives "above" the primary groups/landscape/channel surface, like Leap or Landscape search.

Of particular interest in this micro flow is this interface:

Screenshot 2023-05-02 at 11 49 27 PM

There are a couple of important details going on here:

1. The Search results themselves are hovered above the chat/channel context
2. We're reusing one of the patterns implemented in DM construction: the cancel button for the search interface (which should be abled to be escaped by also typing ESC
3. Channel context is retained by keeping the channel + group name element in the upper left of the channel header
4. In a manner similar to james' prior work for autocomplete, we're graying out the string result representing the search input, to make it clear the search prompt is "completing" the fetched result
5. Message rows, rather than take on the actual/literal styling of the messages in a Talk channel, are surrounded by a border while not hovered/focused, and are given a light gray background and accompanying "Go to Message" button when hovered

Each of these details, as a whole, build up a very solid/distinct interface for search results emanating from the search result. If we can't implement them -all- in this PR, I totally understand that — the most important element in this list is the first item, which will go a long way in making search results more obvious.

urcades:
If, for whatever reason, we can't manage to implement the results as a "floating element", this is the next best thing:

Screenshot 2023-05-02 at 11 58 39 PM

jamesacklin:
Update ~2023.5.30 - @arthyn is attempting to make this work without #2468.

arthyn:
pinging for re-review @patosullivan @urcades @dnbrwstr @jamesacklin @ashelkovnykov 

Unfortunately this got to be an even bigger update ha. The newest commits here add @urcades UI updates to make the search look more like a normal search. To achieve that we had to build a combobox similar to what we had in Landscape for app/dev search. In addition, we needed the capability to jump around in chat and view disparate windows of writs. This mostly meant managing stuff on the frontend but also required a new writs scry "around" which gives 25 messages around the requested time. Here's what it looks like:


https://github.com/tloncorp/landscape-apps/assets/5466421/401ce5bb-95e4-47e3-80fd-abd8beb89f07


This also adds search to both dms and group dms, and includes some changes around channel headers to make them more specific to each channel and remove all the polymorphic stuff we were juggling in the channelheader component and some other small a11y changes.

jamesacklin:
Huge. I'll check it out locally tonight.

ui/src/chat/ChatMessage/Author.tsx
```diff
@@ -33,19 +28,15 @@ export default function Author({
...
+  const timeDisplay = date ? makePrettyDayAndDateAndTime(date) : undefined;
 
   const handleProfileClick = () => {
     modalNavigate(`/profile/${ship}`, {
       state: { backgroundLocation: location },
     });
   };
 
-  if (!date) {
+  if (!timeDisplay) {
```
patosullivan, line 39: Why do we need to update this?

ui/src/chat/ChatSearch/ChatSearch.tsx
```diff
@@ -0,0 +1,208 @@
...
+}>;
+
+export default function ChatSearch({
+  whom,
+  root,
+  placeholder,
+  children,
+}: ChatSearchProps) {
+  const navigate = useNavigate();
+  const { query } = useParams<RouteParams>();
```
patosullivan, line 42: It might be a good idea to store the most recent query somewhere outside of the route in case the user wants to look through a few different results without having to search again.

patosullivan (approved):
Code looks good for the most part, added two questions in-line. The biggest issue I found while using the feature locally was that if I select a message from a thread the thread pane and the message itself show up as expected, but it's difficult to find the parent message in the main chat scroller. Not sure if we want to punt on that or not.

ui/src/chat/ChatMessage/Author.tsx
arthyn: mostly just didn't want to do ?. everywhere were referencing time display. but if you mean the time display stuff itself, it's because the screen reader was reading the bullet lol.

ui/src/chat/ChatSearch/ChatSearch.tsx
arthyn: like a recent searches or something? I think that's a good idea, we def need a follow-up with something like that and actually using the mentions because the mocks had those and I'm sure people would appreciate

Issue #2425, opened 2023-04-27
lure: group host should check connectivity to lure service when creating lure link


zalberico:
If a group host cannot communicate with the lure service, the following can happen:
 - Group host creates lure link for their group
 - They share that link to users
 - Users click that link and end up at a "group not found page"

This happens because the lure service (currently hosted by us on ~loshut-lonreg) does not have the metadata for the lure link because the user's group host ship cannot communicate with it (recently this was because a bunch of galaxies were offline).

Proposal:
 - When a user enables the lure link, the system should check for connectivity to the lure service (which is set in the landscape settings).
 - If the group host cannot contact the lure service, it should throw a red warning in the UI and tell the user the lure link could not be created because the lure service was not accessible.

@midsum-salrux I though you'd be a good candidate for this one, but feel free to pull @tomholford in for any help with the UI warning bit.

Also if you have a better proposal for this that's great too (this was just my first thought).

This was one of the things that came out of the discussion about how to avoid edge cases with lure failures (we'll share the doc broadly).

midsum-salrux:
This all makes sense. We track whether we've gotten a poke-ack from the bait service already, so we can just expose this in the UI.

jamesacklin:
Happy to pull this in for 3.1.0 if it's not a lot of investigative effort

zalberico:
Reopening this since we've seen it not work.

zalberico:
Latest is in https://github.com/tloncorp/landscape-apps/pull/2444 - it sounded like it was on binnec/marnec but not yet deployed? I asked some questions on that PR

zalberico:
I think we're good - closing out based on the questions answered in the PR.

PR #2444, opened 2023-05-01
Show lure connection issue

midsum-salrux:
https://github.com/tloncorp/landscape/pull/161

### DIFF
--- a/ui/src/groups/LureInviteBlock.tsx
+++ b/ui/src/groups/LureInviteBlock.tsx
@@ -25,7 +25,7 @@ export default function LureInviteBlock({
   group,
   className,
 }: LureInviteBlock) {
-  const { supported, fetched, enabled, url, toggle } = useLure(flag);
+  const { supported, fetched, enabled, enableAcked, url, toggle } = useLure(flag);
   const { didCopy, doCopy } = useCopy(url);
 
   if (!supported) {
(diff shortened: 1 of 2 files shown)


jamesacklin:
Fixes #2425 ?

tomholford (approved):
Paired with @midsum-salrux on this, the flow was working successfully on my machine. LGTM

zalberico:
What happens when the connection is fixed? Does the link automatically appear there, does the error text go away?

zalberico:
Tagging @tomholford and @midsum-salrux for visibility

zalberico:
Is it possible for that check to pass but the link to still not get set? Does the check happen before sending the link (not in one transaction). We saw galen have a link set that didn't work do to communication issues with dem, but it may have been intermittent.

zalberico:
@midsum-salrux can you take a look at the questions above when you see this? I can create new tracking issues depending on the answers.

midsum-salrux:
> What happens when the connection is fixed? Does the link automatically appear there, does the error text go away?

You'd need to navigate away and back or refresh for the error to go away I think


> Is it possible for that check to pass but the link to still not get set? Does the check happen before sending the link (not in one transaction). We saw galen have a link set that didn't work do to communication issues with dem, but it may have been intermittent.

I think that may have been due to this not being deployed. I can't think of a state where the link wouldn't get set but the poke still gets acked.

PR #2459, opened 2023-05-02
react native android app

alecananian:
- Adds code for React Native-based Android app that currently acts as a webview wrapper around a ship's Talk mobile UI
- Adds two GitHub Actions workflows for creating Android builds (manually or when tagged with `vX.X.X-android`) and publishing OTA updates (when pushing to `master`)


native-android/src/lib/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+      },
+    },
+  ]);
+
+  if (Platform.OS === 'android') {
+    await Notifications.setNotificationChannelAsync('default', {
+      name: 'default',
+      importance: Notifications.AndroidImportance.MAX,
+      vibrationPattern: [0, 250, 250, 250],
+      lightColor: '#FF231F7C',
```
arthyn, line 60: do we need a dark color here?

native-android/src/lib/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+export const handleNotification = async (
+  notification: Notifications.Notification
+) => {
+  const { identifier, content } = notification.request;
+  console.log('Handling notification', identifier, content.data);
+  await Notifications.scheduleNotificationAsync({
+    identifier,
+    content: {
+      categoryIdentifier: 'message',
+      title: 'You received a new message',
```
arthyn, line 129: do we not want to scry for the contents here?

native-android/src/state/hark.ts
```diff
@@ -0,0 +1,227 @@
+import _ from 'lodash';
```
arthyn, line 1: if we can we should trim this up to only what we need, also this is a bit outdated now since we use skeins.

arthyn (approved):
lgtm, mostly just question/comments no blockers or anything

native-android/google-services.json
```diff
@@ -0,0 +1,47 @@
+{
```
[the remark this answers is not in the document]
alecananian: It is okay to check into the repo as it only contains public keys

native-android/src/lib/notifications.ts
alecananian: We will scry for contents in a future build

native-android/google-services.json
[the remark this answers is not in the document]
arthyn: cool

native-android/src/state/hark.ts
alecananian: @patosullivan I wasn't sure if you had made React Native-specific changes to this file. Is it okay to copy `hark.ts` from the UI project?

patosullivan: Yes, just make sure to copy this pattern for accessing the api:

```
  const { api } = useStore.getState();
    if (api === null) {
      return;
    }
 ```

Comment (patosullivan) [reply native-android/src/state/hark.ts:1]: Actually, the new hark.ts uses react-query, I'm not sure if that's what we want in react native. I'll look into it.

Comment (patosullivan) [reply native-android/src/state/hark.ts:1]: I think we should probably wait on the react-query stuff and just update this file to use skeins.

Comment (arthyn) [reply native-android/src/state/hark.ts:1]: The only thing we seem to be using it for currently is `sawRope` let's just rip everything out but that since we don't really need it.

Comment (alecananian) [reply native-android/src/state/hark.ts:1]: Will take care of this now!

Comment (alecananian) [reply native-android/src/state/hark.ts:1]: @arthyn `sawRope` calls `await get().update(rope.group)`, which pretty much requires the rest of the functions except for `sawSeam`. Is this update call safe to rip out as well?

Comment (alecananian) [reply native-android/src/state/hark.ts]: Sounds good! I've pared down the hark state file a bit and also fixed a `lodash` import in another file

File: native-android/src/lib/notifications.ts (hunk 1 above)
Comment (alecananian) [reply native-android/src/lib/notifications.ts:60]: `darkColor` does not appear to be an option:

```
export interface NotificationChannel {
  id: string;
  name: string | null;
  importance: AndroidImportance;
  bypassDnd: boolean;
  description: string | null;
  groupId?: string | null;
  lightColor: string;
  lockscreenVisibility: AndroidNotificationVisibility;
  showBadge: boolean;
  sound: 'default' | 'custom' | null;
  audioAttributes: AudioAttributes;
  vibrationPattern: number[] | null;
  enableLights: boolean;
  enableVibrate: boolean;
}
```

arthyn: ah gotcha didn't know just usually there's a complement haha

arthyn:
@alecananian @patosullivan if y'all are good, I'm good with deploying this.

alecananian:
Let's do it!

patosullivan (approved):

patosullivan:
I'm good with it. Occurred to me earlier today that we might want to provide a refresh button so users can get out of bad states without killing the app. Probably not necessary to add that before we merge, though.

PR #2465, opened 2023-05-03
emojis: fix route issue on mobile

### LINKED ISSUE #2445: Emojis in threads on iOS mobile web don't work
- Open thread on mobile web iOS in groups
- Try to +1 some message
- Observe that the emoji picker never appears

patosullivan:
Fixes #2445 by accounting for routing to the emoji picker from within a thread on mobile.

Also adds some RTL tests around ChatMessageOptions. Attempted to test for showing/hiding the emoji picker itself, but it ended up being a rabbit hole. Going to attempt to write those tests using playwright in another PR.


arthyn (approved):
lgtm, hopefully we can make this easier somehow going forward

PR #2471, opened 2023-05-05
android: fix login state mgmt and api init

alecananian:
- Fixes React version peer dependency mismatch
- Fixes broken `lint:format` command because of missing package
- Fixes `api` helper not being initialized upon login
- Refactors state management and flow for login and returning to app on relaunch

I discovered that the `api` helper was only being initialized when you relaunched the app from storage, it was not initialized after manual login so the push notification registration request wasn't being sent through. I ended up doing a major refactor of the state management to make it a bit simpler, particularly around only allowing one ship at a time to be configured.

### DIFF
--- a/native-android/src/Login.tsx
+++ b/native-android/src/Login.tsx
@@ -106,12 +106,14 @@ export default function Login() {
           });
         } else {
           setLoginProblem(
-            'An error occurred while loggin in. Please try again.'
+            'An error occurred while logging in. Please try again.'
           );
         }
       } catch (err) {
         console.error('Error validating access key', err);
-        setLoginProblem('An error occurred while loggin in. Please try again.');
+        setLoginProblem(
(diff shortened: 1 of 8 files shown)


native-android/src/Login.tsx
```diff
@@ -9,164 +9,116 @@ import {
...
+            authCookie,
+          });
+        } else {
+          setLoginProblem(
+            'An error occurred while loggin in. Please try again.'
+          );
+        }
+      } catch (err) {
+        console.error('Error validating access key', err);
+        setLoginProblem('An error occurred while loggin in. Please try again.');
```
arthyn, line 114:
```suggestion
        setLoginProblem('An error occurred while logging in. Please try again.');
```

arthyn (approved):
Much simpler, lgtm

PR #2473, opened 2023-05-05
Add playwright tests

patosullivan:
This adds playwright tests that run against two fake ships (~zod and ~bus) that can be stood up and taken down on each run of the e2e tests. These tests should also run on pull requests to github via the update to ci.yml.

At the moment this really just tests for whether login to landscape works (which is necessary for subsequent tests) and whether a user can accept an invite to a group and actually make it into the group.

Doesn't yet copy/commit the desk files into the fake ships, just tests frontend changes against the groups/talks desks already installed on the ships.


ui/.husky/pre-commit
```diff
@@ -3,6 +3,9 @@ cd ui
 
 npm run test -- run -u
 
+npm run e2e
+npm run e2e:chat
```
arthyn, line 7: this might be brutal lol

arthyn (approved):
lgtm!

PR #2475, opened 2023-05-07
%u scry endpoint for chat

polrel-witter:
This came about while updating `%talk-cli`; a few parts needed to check for messages before scrying them out to avoid crashing.


arthyn (approved):
lgtm

PR #2480, opened 2023-05-08
groups: fix issue with private groups in find groups

### LINKED ISSUE #2478: groups: private groups marked as public, impossible to "ask"
Private groups are set appropriately on the host side, but they appear as Public groups in the Find view. Clicking "Join" does not issue an ask to the group host.

**To Reproduce**
1. Create a private group on ship A, or create a public group and switch it to private.
2. Find ship A on ship B. Note the "Public" status of the private group.
3. Click "Join." 
4. Watch group join hang.
5. On ship A, note that ship B does not appear in the pending queue.
6. Eventually cancel join.

**Expected behavior**
A. Private groups should appear labeled as such in the Find view.
B. The action button for a private group should read "Request to Join" (see [Figma](https://www.figma.com/file/sCFYLlAGjehlYXUkT2rarg/Product-Handoffs?type=design&node-id=1593-69981&t=A1IXD8rvK9VQWC6t-4))
C. The ship asking to be admitted to the group should appear in the Pending list in the group admin view.

patosullivan:
Fixes #2478

We weren't properly checking the privacy of groups returned from the group index. useGroupPrivacy defaulted to `public` if we didn't have group or gang data for a particular flag, which means that we didn't show the accurate privacy setting for groups we only knew of via group index (i.e., groups we had found by searching for a ship in Find Groups).

This PR updates useGangs so that it will insert gang indexes into the cached gangs data. Also fixes an issue we had with usePendingGangs and usePendingGangsWithoutClaim returning gangs without an invite key at all.

### DIFF
--- a/ui/src/state/groups/groups.ts
+++ b/ui/src/state/groups/groups.ts
@@ -28,6 +28,7 @@ import { BaitCite } from '@/types/chat';
 import useReactQuerySubscription from '@/logic/useReactQuerySubscription';
 import useReactQuerySubscribeOnce from '@/logic/useReactQuerySubscribeOnce';
 import useReactQueryScry from '@/logic/useReactQueryScry';
+import { getFlagParts } from '@/logic/utils';
 
 export const GROUP_ADMIN = 'admin';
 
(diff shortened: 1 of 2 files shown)


ui/src/state/groups/groups.ts
```diff
@@ -212,7 +214,26 @@ export function useGangs() {
...
+      });
+      return newAcc;
+    }
+    return acc;
+  }, {});
+
   return {
+    ...groupIndexDataAsGangs,
     ...(data as Gangs),
   };
```
arthyn, line 238: Probably should wrap everything in useMemo

arthyn (changes requested):

arthyn (approved):
lgtm

PR #2481, opened 2023-05-08
reconnect: tweaking pt3

arthyn:
Stop fetching on focus and reconnect by default, better briefs syncing, new channel reaping handling, stopping channel notifications from being automatically marked read (clogging request queue).


patosullivan (approved):
Looks good, runs well locally.

PR #2485, opened 2023-05-09
bugfix: properly render note references

### LINKED ISSUE #2230: Groups: Notebook embed previews in Chat elide rich text elements entirely and concatenate paragraphs
Notebook embeds in Chat provide a two-paragraph preview of the contents of the notebook post. However, there are two problems:

1. The two paragraphs are directly concatenated, with no gap between them (not even a space).
2. Rich text elements (for instance, hyperlinks, bold text, italics, _etc._) are totally dropped. This is not to say that the rich text elements are rendered as plain text, but rather that they are not rendered at all.

Please consult the following image:

urbit - landscape notebook embed error

Note:

**a)** A hyperlink fails to render.
**b)** Italicised words fail to render.
**c)** There is no space between the two paragraphs.

**To Reproduce**

1. Create a Notebook post containing rich text elements and at least two paragraphs.
2. Copy the link to the Notebook post.
3. Paste it into a Chat.

**Expected behaviour**

Rich text elements render in the Notebook embed, within reason. If multiple paragraphs are to be rendered they should be separated by at least a space.

**Additional context**

- Ship type: Planet
- Instance type: Self-hosted CLI
- OS/browser: Vivaldi on Linux

patosullivan:
Fixes #2230

This PR updates the NoteReference component to use DiaryContent to render a truncated version of the content of a note, including formatting and block content.

Looks like this:

image

### DIFF
--- a/ui/src/components/References/BaitReference.tsx
+++ b/ui/src/components/References/BaitReference.tsx
@@ -3,11 +3,11 @@ import { useHasMigratedChannels } from '@/logic/useMigrationInfo';
 import { useGroup, useShoal } from '@/state/groups';
 import { BaitCite } from '@/types/chat';
 import { udToDec } from '@urbit/api';
-import cn from 'classnames';
 import React from 'react';
 import ExclamationPoint from '../icons/ExclamationPoint';
 // eslint-disable-next-line import/no-cycle
 import CurioReference from './CurioReference';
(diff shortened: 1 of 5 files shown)


arthyn (changes requested):

arthyn (approved):
lgtm

Issue #2466, opened 2023-05-03
groups: updates to group creation, incorporate lure activation


urcades:
The latest updates to the group creation flow involve three critical changes:

1. Shortening the overall process from three to two steps, removing the ability to invite people to the group, the reasoning being that it makes little sense to invite people to a group that has no channels added to it
2. Incorporating the ability to enable lure links for the group, during the privacy-configuration step
3. Optimizing + updating the first step (customizing the group) with a few changes:
    a. Simplifying the customization flow: removing the ability to specify hex colors or image links separately for the group's avatar and header, opting for a single "accent color" feature which sets the overall initial theme for the group. Per element customization can happen from within the group's settings interfaces
    b.  Showing a "live" preview of the sidebar state, to give people a sense of how their visual changes will manifest
    c. Most notably, granting the selected accent color the ability to set the color for the entire sidebar, primary buttons, and the gray "background" for each channel (seen most prominently in notebooks and galleries, but also in all channels)

<img width="7880" alt="Group Creation with Lure" src="https://user-images.githubusercontent.com/1195363/236021254-ee4a5385-f4c4-4bd7-842b-de9a4465bfc3.png">

As illustrated in c.) this update suggests some major changes to group owner's ability to theme their group in the form of a single variable for an "Accent color". This may have some overlap with #2295 , and we should probably reason through how admins can make their groups unusable for their members via bad color accessibility. On the other hand, this reinforces that people are the true owners of their urbit's resources.

urcades:
Another small detail: The privacy-setting step is visually updated to provide more space to list bullet points indicating what each privacy setting "means" in terms of group access.

Issue #2348, opened 2023-04-20
groups: omnibus improvements to group settings ux


urcades:
At the moment, our group settings interfaces present themselves as a page-like structure, which unlike most full-screen "pages" of a group (channels, home, activity, all channels, etc.) has no sidebar nav presence:

<img width="1665" alt="Frame 2514" src="https://user-images.githubusercontent.com/1195363/233264265-50ccbcd7-d81c-4c77-98eb-9c10d0870cf3.png">

In order to further ground people's expectations around how we resolve "sidebar nav" to "pages" **and** how we resolve non-sidebar nav to more "modal" interfaces (dropdowns, popovers, dialogs, etc.), I'm proposing that we transition our paged "Group Admin/Settings/Config) interfaces into a modal experience, with clearer segmentation of Group Admin/Settings concerns:

<img width="1440" alt="Section_ Group Info-Details" src="https://user-images.githubusercontent.com/1195363/233264506-3cdfc98a-5881-4935-8e31-824fc95184fd.png">

All-in-all, this redesigned settings experience should resolve as such:

![Group Settings](https://user-images.githubusercontent.com/1195363/233264752-d6706e90-ccfd-4270-9656-f4a087ad8805.png)

This update lives in Figma [here](https://www.figma.com/file/QMM0PcJyuKWrXY2n4VpQp1/2023-Q2-Groups?node-id=216%3A30099&t=rdICRm2en7kiYGoH-1).

urcades:
The redesigned "Group Info" section now incorporates a proposed color picker for setting a group's "accent color":

<img width="881" alt="Screenshot 2023-04-20 at 1 25 41 AM" src="https://user-images.githubusercontent.com/1195363/233266269-c70b4186-eb06-42bb-9c7a-f42ce51e61bc.png">

Accent colors have yet to be technically defined and scoped, but the idea is for a single color, chosen by an admin, to override Tlon's default theming for a single group. This might include color overrides for "primary"-style button backgrounds, desaturated-accent channel backgrounds (to replace the default gray we use everywhere), and the sidebar itself.

We'll have to weigh this proposed functionality against accessibility/readability concerns as well, of course.

See #2295 , which seems related.

urcades:
I've extended the controls surfaced in the "Invites & Privacy" section:

[...]

PR #2492, opened 2023-05-10
admin: reformat group settings

### LINKED ISSUE #1553: groups: host shows existing members as pending
Screenshot 2022-12-16 at 11 11 27 AM

Marisa is in the tlon group, but the host (dabben-larbet) shows her as pending.

jamesacklin:
Wholesale reboot of the Group Settings interface.

- Converts the interface to a tabbed-sidebar modal for large screens.
- Stacks the navigation and settings interfaces for small screens.
- Adjusts routes to consistently provide a background location for the modal.
- Makes the delete group name-check insensitive of special characters.
- Breaks the on-network invitation form out into its own component.
- Refactors the privacy selector.
- Virtualizes the pending list.
- Differentiates between Badge and AddBadge icons (and fixes the latter).
- Removes the header element of the Roles editor while editing a role.

A tour of Tlon Local and Tlon:

https://github.com/tloncorp/landscape-apps/assets/748181/9a17df48-9314-40cc-8ae4-8f4f445f59f8

Fixes tloncorp/landscape-apps#2348
Fixes tloncorp/landscape-apps#2486

Please note:

- Impacts #2466 in adjusting the privacy selector but nothing else
- Does not fix #1553 
- Does not reformat the basic "Group Info" form in #2348
- Does not implement group accent colors


jamesacklin:
@patosullivan This messes with a lot of the group admin interfaces that you presumably changed in #2350. How would you prefer to coordinate?

patosullivan:
> @patosullivan This messes with a lot of the group admin interfaces that you presumably changed in #2350. How would you prefer to coordinate?

After @arthyn's review mine should be good to go, how about we merge that first? I can push a commit to this branch to work out the merge conflicts if you like.

jamesacklin:
I agree, let's merge your stuff to develop first. As long as your role components are relatively self-contained, it shouldn't be too hard to add a route/SidebarItem and drop them in.

patosullivan (changes requested):
Overall it looks good, but there are two things:

- Normal group members can no longer access group info (biggest downside here imo is you can't see who is in the group with you).

- The close button in the modal seems like its should be in a different spot. The width of the cards within the main modal view apparently changes, which makes the X look kind of odd sitting where it does now IMO. I'll upload a video to show what I'm talking about.

patosullivan:
https://github.com/tloncorp/landscape-apps/assets/1221094/bacc60e2-6c07-4eca-b360-06fa27a85408

jamesacklin:
Agreed, thanks for the catch on non-admins.

> The close button in the modal seems like its should be in a different spot.

This is due to the scroll bar occupying some width (vs not when it's not needed) - @urcades do you have an interesting solution for where we should locate the close button? Or should it move over to acommodate the scroll bar?

urcades:
> @urcades do you have an interesting solution for where we should locate the close button?

@jamesacklin providing an idea/solution for this problem by EOD

One other minor note I'll make is that when first assessing @patosullivan 's work on the roles paging animation, it felt good in the context of the video presented in the PR. When actually used (in my experience) it's a little disorienting and not yet fine-tuned enough to warrant its inclusion in the new settings ux. Secondly, no other sections appear to use this pattern, so it's a little jarring  to only surface this interaction in one place.

If it's easy to do in the context of this PR, I'd recommend we extract the animation from this experience until we can fine-tune its behavior or more generally apply it to other similar interfaces.

jamesacklin:
@urcades Animation gone, will update the preview.

urcades:
If we feel like we need to surface an explicit "x" to close out our modals, I think we should just situate them within an explicit space, such as this header (which is using the "Generic Channel Header" indigo component):

Screenshot 2023-05-30 at 5 34 03 PM

I personally feel like "tap out of modal" interaction is ubiquitous enough to warrant an implicit "ESC" key or "click outside of modal to close" to be acceptable, but I know there's no replacing explicit affordances like this. Might as well make it very obvious.

jamesacklin:
@urcades and @patosullivan, updated with the latest

patosullivan (approved):
Looks good, I only found one thing that is annoying but probably not a blocker. The menu for the ship selector for group invites appears within the modal instead of on top of it:

image

We can probably fiddle with that separately since it involves the can of worms that is the ship selector.

PR #2493, opened 2023-05-10
apps: hosting support

arthyn:
This enables the Tlon hosting platform to run analytics. Also fixes an issue where newly created heaps/diaries were erroring on actions like adding posts, etc.


patosullivan (approved):
Looks good.

PR #2495, opened 2023-05-11
note reference cleanup

patosullivan:
This fixes several issues with the new note references:

1) node types can be nested, so we need to ensure that we are truncating an actual string before calling slice. 
2) list items need to be truncated.
3) `not-prose` class should not be used on the NoteReference container as this will cause weird issues with the way we render the note (also...it is prose).

### DIFF
--- a/ui/src/components/References/NoteReference.tsx
+++ b/ui/src/components/References/NoteReference.tsx
@@ -64,7 +64,7 @@ function NoteReference({
   const prettyDate = makePrettyDate(new Date(outline.sent));
 
   return (
-    <div className="note-inline-block not-prose group max-w-[600px]">
+    <div className="note-inline-block group max-w-[600px]">
       <div
         onClick={handleOpenReferenceClick}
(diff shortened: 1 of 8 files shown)


ui/src/components/References/NoteReference.tsx
```diff
@@ -64,7 +64,7 @@ function NoteReference({
   const prettyDate = makePrettyDate(new Date(outline.sent));
 
   return (
-    <div className="note-inline-block not-prose group max-w-[600px]">
```
arthyn, line 67: there was some reason I did this that I'm having trouble remembering, but I think it had to do with nested refs in notes. will look real quick

arthyn (approved):
lgtm with suggestion

ui/src/components/References/NoteReference.tsx
patosullivan: Cool, just added it, looks good.

PR #2497, opened 2023-05-11
tests: check for tiles after login in auth setup

patosullivan:
We need to make sure the tiles actually load on landscape, this seems like a good place to do that. 

Note that this will only tell us if landscape/garden is working *with the version installed on the fake ships*, which is not updated before running this test. Also note that landscape/garden lives in a different repo, so I'm not sure how we want to handle copying/building landscape on these fake ships.

cc @jamesacklin


arthyn (approved):
lgtm

PR #2508, opened 2023-05-12
roles: Fake out the "members" role on the frontend for channel perms

### LINKED ISSUE #2499: channel: update role field to include "member" case
Scenario:

- I join the group, I get dropped into the General chat.
- I can read a notebook channel but can't post new notes.
- Eventually, I gain the trust of the group and they invite me to write some notebook posts.
- A group admin adds me to a specific role that is a "Writer" of the notebook channel.
- I can write to the notebook channel.

This is based on my working understanding of our group/channel permissions model:

Image

In the above scenario, here's how I, a group admin, would construct these permissions:

1) The general chat, which everyone can read and write to.

Image

2) An "Experts" role, who are not administrators.

Image

3) A "Zinc Facts" notebook, which I want everyone to be able to read, but only admins and "Experts" to write to.

Image

This ticket concerns the fact that I can't express this in the current UI. My suggestion is to add a "Members" token to the dropdown/multi-select "Readers" and "Writers" fields such that I can explicitly allow regular members of the group to read but not write.

patosullivan:
Fixes #2499

Also removes the "Members can read only" privacy type and sets the equivalent of that as the default when selecting "Custom" (user can then add/remove roles from that as necessary).

### DIFF
--- a/ui/src/channels/EditChannelForm.tsx
+++ b/ui/src/channels/EditChannelForm.tsx
@@ -70,13 +70,20 @@ export default function EditChannelForm({
 
   const onSubmit = useCallback(
     async (values: ChannelFormSchema) => {
-      const { privacy, ...nextChannel } = values;
+      const { privacy, readers, ...nextChannel } = values;
 
       if (presetSection) {
         nextChannel.zone = presetSection;
       }
       try {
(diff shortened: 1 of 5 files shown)


arthyn (approved):
lgtm

PR #2509, opened 2023-05-12
android: refactor notification permission request flow

alecananian:
- Refactors the way Android is checking for and request notification permissions because in the pre-Android 13 world there is no explicit permission request. We will no longer skip registration just because permission had been granted before launch.
- Fixes issue with receiving notifications on logged out state
- Replaces the Expo network static check with a listener that will auto-update while app is launched


native-android/src/lib/notifications.ts
```diff
@@ -60,15 +73,6 @@ export const initNotifications = async () => {
   });
   Notifications.registerTaskAsync(BACKGROUND_NOTIFICATION_TASK);
 
-  // Handle receiving notifications while app is in foreground
-  Notifications.setNotificationHandler({
```
patosullivan, line 64: I think it's a good idea to take out foreground notifications, we should check with @jamesacklin too, though.

patosullivan (approved):

PR #2510, opened 2023-05-12
chat input: fix focus loss on mobile.

### LINKED ISSUE #2506: chat input losing focus on mobile
Happens after every send. Android/chrome and the Android native app.

Have seen it a few times on desktop recently but couldn't reliably reproduce.

patosullivan:
Not clear why this isn't always triggered on desktop, but adding `chatInfo.blocks` to the dependency array for the onSubmit useCallback was causing the ChatInput to lose focus after send on mobile after every message. I changed the dependency to `chatInfo.blocks.length` and that seems to have solved it. I also cleaned up the if statements at the top of onSubmit.

Fixes #2506


arthyn (approved):
makes sense because any changes to onsubmit re-initializes tiptap which loses focus

PR #2518, opened 2023-05-15
heap: fix for authors not receiving notifications for comments

### LINKED ISSUE #1872: groups: user does not receive notifications for comments on gallery items
If a user makes a gallery item and later receives a comment, the user that posted it will not be notified.

**To Reproduce**
1. Create a gallery item.
2. Have another user create a comment on it.
3. Notice that you don't receive a notification.

**Expected behavior**
User should received a "commented on:" notification, similar to notes.

patosullivan:
Fixes #1872 by making sure that we emit a yarn in this particular instance. They would receive notifications if they were to follow up to the comment, but they weren't getting notified on the initial comment because their patp wasn't in the replies. Now they'll always be notified on any comment on their heap posts.

One question for @arthyn about why it's necessary to return he-core twice for this to compile.

### DIFF
--- a/desk/app/heap.hoon
+++ b/desk/app/heap.hoon
@@ -952,9 +952,8 @@
           ~
         =/  am-op-author  =(author.curio.u.op our.bowl)
         =/  am-author  =(author.heart our.bowl)
-        ?:  |(&(!am-author in-replies) &(am-op-author !am-author))
-          =.  cor  (emit (pass-hark & & yarn))
-        he-core
+        =?  cor  |(&(!am-author in-replies) &(am-op-author !am-author))
+          (emit (pass-hark & & yarn))
         he-core
       ==
     ::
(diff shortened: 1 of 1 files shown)


desk/app/heap.hoon
```diff
@@ -951,7 +950,11 @@
                 (flatten q.content.heart)
             ==
           ~
-        =.  cor  (emit (pass-hark & & yarn))
+        =/  am-op-author  =(author.curio.u.op our.bowl)
+        =/  am-author  =(author.heart our.bowl)
+        ?:  |(&(!am-author in-replies) &(am-op-author !am-author))
+          =.  cor  (emit (pass-hark & & yarn))
+        he-core
```
arthyn, line 957: the reason is that doing `=.` makes a state change and then returns the value following using that updated subject so `=.  cor  (emit (pass-hark & & yarn))` does the state change, but you have to return something.

```suggestion
        =?  cor  |(&(!am-author in-replies) &(am-op-author !am-author))
          (emit (pass-hark & & yarn))
        he-core
```

this will do the state change conditionally and then always return the core

arthyn (approved):
lgtm

PR #2519, opened 2023-05-15
groups: add "host offline" state

### LINKED ISSUE #2414: groups: "host offline" state for join flow
> Requires peer availability implementation

If we're unable to contact a group's host, or to finish loading its content because the host is offline, we should reflect that in the load state.

Host availability unknown:

Image

Host offline:

Image

For discussion: how's this copy? Do we want to cover other states here?

patosullivan:
fixes #2414

looks like this:

image

I added a new subscription path just for `hi`ing a ship, returns the ship name if it's successful, also added some UI state around disconnected ships (using zustand on the frontend).

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -292,6 +292,10 @@
     ?.  ?=(%known -)  (hi-and-req-gang-index ship)
     (req-gang-index ship)
   ::
+      [%hi ship=@ ~]
+    =/  =ship  (slav %p ship.pole)
+    (hi-ship ship)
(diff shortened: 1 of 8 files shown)


arthyn (approved):
lgtm, we may want to tweak the timeout

patosullivan:
@urcades thoughts on how this looks?

urcades:
@patosullivan This is great (for a liminal/loading state) — basically anything we can do to surface up whatever data/info is available to us is a positive development

PR #2522, opened 2023-05-16
update workflows to not reference janeway

dosullivan:
We're not using the janeway service account anymore, so this changes out the references to be more generic.


arthyn (approved):
lgtm

jamesacklin:
Thank you 🫡

PR #2524, opened 2023-05-16
android: skip local notification when yarn contents scry fails

### LINKED ISSUE #2521: android: "You have a new notification" -> routes you to a route ending in undefined.
I get this every now and then, even if my ship is offline.

alecananian:
Resolves #2521 

Removes the "You have received a new message" notification and simply skips the local notification send if the notification contents could not be retrieved (i.e., yarn details scry fails)


patosullivan (approved):
Looks good

PR #2525, opened 2023-05-17
groups: allow users to navigate to "joining" state groups

### LINKED ISSUE #2413: groups: initial load improvements
### Collapse differences between "waiting for host" and "downloading content"

Right now, groups whose hosts cannot be contacted are grayed-out in the sidebar and cannot be opened, while groups that can be contacted but are still loading can be opened, but to a blank group.

Instead, in either state, a group that is not fully initialized should be grayed out in the sidebar, but openable.

### Add loading indicator to initial group state

Right now, it's not clear what's happening while a group is loading. We should add a loader and status text to indicate what's going on.

Image

https://www.figma.com/file/lPb6AIjQbg6ECbhxc6vXer/2023-Q2-Onboarding?node-id=184-19781&t=My1PDbEoTnS6S6yn-11

patosullivan:
Fixes #2413 by allowing for users to navigate into a group that is currently in a "joining" state. Also:

- adds "Cancel Join" button to group actions (available from either the sidebar or the dropdown in the group header).
- adds "Retry" button to the "Group host is offline" screen.
- Fixes some CSS issues with "joining" groups listed in the sidebar.
- Shows the group's title from the gang preview in the group header if the group is in the "joining" state.

### DIFF
--- a/ui/src/groups/Groups.tsx
+++ b/ui/src/groups/Groups.tsx
@@ -7,7 +7,6 @@ import {
   useGroupConnection,
   useGroupConnectionState,
   useGroupHostHi,
-  useGroupIndex,
   useRouteGroup,
   useVessel,
 } from '@/state/groups/groups';
@@ -25,7 +24,7 @@ function Groups() {
   const group = useGroup(flag, true);
   const gang = useGang(flag);
   const { ship } = getFlagParts(flag);
-  const { isError, isSuccess, isLoading, ...rest } = useGroupHostHi(ship);
+  const { isError, isSuccess, isLoading } = useGroupHostHi(ship);
   const connection = useGroupConnection(flag);
   const vessel = useVessel(flag, window.our);
   const isMobile = useIsMobile();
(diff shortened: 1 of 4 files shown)


ui/src/groups/Groups.tsx
```diff
@@ -95,10 +94,18 @@ function Groups() {
...
       <div className="flex min-w-0 grow items-center justify-center bg-gray-50">
-        <div className="flex items-center justify-center">
+        <div className="flex flex-col items-center justify-center space-y-4">
           <span className="ml-2 text-gray-600">
             Group host ({ship}) is offline.
           </span>
+          <button
+            className="small-button"
+            onClick={() =>
+              useGroupConnectionState.getState().setGroupConnected(flag, true)
```
arthyn, line 104: what will this do exactly? so like we're in an "offline" state so we see that messaging, but if the group join goes through behind the scenes we'll have to hit this to see that?

patosullivan: If we're in the "group is down" cached state, yeah.

patosullivan: Yeah, good idea.

arthyn (approved):
lgtm, let's see how it works out

PR #2530, opened 2023-05-17
notebook refs: more fixes

### LINKED ISSUE #2526: diary refs: refs within diary refs cause browser tab to crash, also looks weird
OTT

patosullivan:
This fixes #2526 by preventing refs from being rendered within a notebook reference, also: 

- fixes some issues around the way we were calculating remainingChars (it was getting reset early)
-  limits the number of images that can be rendered in a ref (to one) 
- handles header text (which users can apparently add by pasting in content from somewhere else, this is not available directly in our UI)
- adds an ellipsis to the last inline string in the truncated text
- fixes the "missing keys in list" issue in DiaryContent.

You can test this by making some fake notes with lots of text, images, references, etc., and then embedding it as a reference somewhere.

### DIFF
--- a/ui/src/diary/DiaryContent/DiaryContent.tsx
+++ b/ui/src/diary/DiaryContent/DiaryContent.tsx
@@ -66,8 +66,8 @@ export function InlineContent({ story }: InlineContentProps) {
   if (_.isArray(story)) {
     return (
       <>
-        {story.map((s) => (
-          <InlineContent story={s} />
+        {story.map((s, k) => (
+          <InlineContent story={s} key={k} />
         ))}
       </>
     );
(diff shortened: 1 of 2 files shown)


arthyn:
> handles header text (which users can apparently add by pasting in content from somewhere else, this is not available directly in our UI)

is available in the UI, we just don't have a button for it, but you can do it by using #'s like in markdown

patosullivan:
> > handles header text (which users can apparently add by pasting in content from somewhere else, this is not available directly in our UI)
> 
> is available in the UI, we just don't have a button for it, but you can do it by using #'s like in markdown

Cool, didn't realize.

arthyn (approved):
lgtm

PR #2533, opened 2023-05-18
hark: don't crash on empty %desks scry

philipcmonk:
Was looking at some first-boot stuff, and was noticing this crash fairly frequently, eg when I refresh /apps/groups/find.  It looks to me like we're just scrying notifications from a desk (%groups) which hark has no entry for.  It seems like defaulting to an empty rug should be fine, but I don't actually know what the call-site is.  This does silence that crash, at least.


arthyn (approved):
lgtm

PR #2536, opened 2023-05-19
Add a 'not' to fix backwards error logic

### LINKED ISSUE #2535: lure: in-group link block is broken
The Lure link for public groups I administrate are displaying as "Not enabled" or offline. The ship itself running the service is fine, according to @lukebuehler. 

My Landscape settings look like this:

image

A group I administrate but am not the host of:

image

A group I both administrate and host:

image
image

**To Reproduce**
1. Run the latest %groups, %talk, and %garden from the Tlon internal prerelease moon.
2. Navigate to a group you administrate that has Lure links enabled.
3. Attempt to copy the link.
4. Navigate to another group you administrate that does not have Lure links enabled.
5. Attempt to enable the link.

midsum-salrux:
Fixes #2535

### DIFF
--- a/ui/src/state/lure/lure.ts
+++ b/ui/src/state/lure/lure.ts
@@ -136,14 +136,14 @@ export const useLureState = create<LureState>(
           undefined
         );
 
-        const enableAcked = await asyncWithDefault(
+        const enableAcked = !(await asyncWithDefault(
           () =>
             api.scry<boolean>({
               app: 'reel',
               path: `/outstanding-poke/${flag}`,
             }),
           true
-        );
+        ));
 
         set(
           produce((draft: LureState) => {


ui/src/state/lure/lure.ts
```diff
@@ -136,14 +136,14 @@ export const useLureState = create<LureState>(
...
         );
 
-        const enableAcked = await asyncWithDefault(
+        const enableAcked = !(await asyncWithDefault(
           () =>
             api.scry<boolean>({
               app: 'reel',
               path: `/outstanding-poke/${flag}`,
             }),
           true
```
arthyn, line 145: does the default need to change too or is it good?

midsum-salrux: The default is good

arthyn (approved):
lgtm

PR #2537, opened 2023-05-19
set emoji picker theme based on current theme

alecananian:
Resolves https://github.com/tloncorp/talk-android/issues/7

Uses value from `useCurrentTheme` hook to set the emoji picker's theme


arthyn (approved):
lgtm

Issue #2476, opened 2023-05-08
groups: add channel description


dnbrwstr:
Channel descriptions will enable us to better contextualize channels, and in particular will make it easier to suggest to new users how they should engage. A field for this should already exist; we just need to surface it in the UI.

[Figma link with comments](https://www.figma.com/file/lPb6AIjQbg6ECbhxc6vXer/2023-Q2-Onboarding?type=design&node-id=411-25570&t=IwWbOfPUSNipBYkl-11)

![Image](https://user-images.githubusercontent.com/357216/236841209-da64fe79-f1e7-4ca3-933c-d2e67cf7ed16.png)

PR #2543, opened 2023-05-22
groups: unified channel headers

jamesacklin:
Made another pass on the heels of #2476.
- Fixes an issue where URLs in descriptions would not wrap and therefore truncate artificially early (thanks to relying on line-break)
- Fixes an issue where long descriptions (particularly those with URLs) would push the content width past the viewport on mobile
- Removes the gray box around channel icons
- Ensures all headers match


patosullivan (approved):
Looks good

PR #2546, opened 2023-05-23
groups: update find groups, group references

### LINKED ISSUE #2477: groups: improve find groups
Right now "Find groups" is confusing, as it's not a real search. We should rework it to emphasize suggestions + directory, and tweak display of the name input for clarity.

[Figma link with comments
](https://www.figma.com/file/lPb6AIjQbg6ECbhxc6vXer/2023-Q2-Onboarding?type=design&node-id=411-27316&t=IwWbOfPUSNipBYkl-11)

Image

patosullivan:
Closes #2477 by matching Find Groups (now "Discover") to Figma, and making associated changes to the buttons in group references. 

Also
- makes group references a bit wider to accommodate the larger button (now says "Join Group" rather than "Join").
- adds some margin below group refs so that they don't bump up directly with other refs.

Looks like this:

image

Group refs now look like this:

image

Note that I still only have three groups listed on the "Discover"/"Find Groups" page, though the design calls for six in two columns. We probably need to come up with a few more groups to list.


arthyn (approved):
lgtm

Issue #2544, opened 2023-05-22
Notifications for Talk messages received, messages not received


lukestiles:

jamesacklin:
Fixed by #2549

PR #2549, opened 2023-05-23
channels: make sure we use unique timestamps

arthyn:
This addresses #2544 by ensuring we have unique timestamps and we're not overriding previous messages.


midsum-salrux (approved):
LGTM

PR #2550, opened 2023-05-23
talk: unify DM and club headers

jamesacklin:
Unifies the Talk headers to match what I did in #2543.

Also (finally) lists the ships that are pending in a club behind the blue "Pending" label. 

image

### DIFF
--- a/ui/src/dms/Dm.tsx
+++ b/ui/src/dms/Dm.tsx
@@ -17,22 +17,6 @@ import CaretLeft16Icon from '@/components/icons/CaretLeft16Icon';
 import ReconnectingSpinner from '@/components/ReconnectingSpinner';
 import MessageSelector from './MessageSelector';
 
-function BackLink({
-  children,
-  mobile,
-}: {
-  children: React.ReactNode;
-  mobile: boolean;
-}) {
-  return mobile ? (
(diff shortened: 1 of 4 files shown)


ui/src/dms/Dm.tsx
```diff
@@ -69,43 +53,51 @@ export default function Dm() {
...
+                      </span>
+                      <span className="w-full break-all text-sm text-gray-400 line-clamp-1">
+                        {ship}
+                      </span>
+                    </>
+                  ) : (
+                    <span className="text-sm font-bold text-gray-800 sm:font-semibold">
+                      {ship}
+                    </span>
+                  )}
```
arthyn, line 90: you'll have to import shipname, but it should probably be:

```suggestion
                      <ShipName full name={ship} className="w-full break-all text-sm text-gray-400 line-clamp-1" />
                    </>
                  ) : (
                    <ShipName full name={ship} className="text-sm font-bold text-gray-800 sm:font-semibold" />
                  )}
```

arthyn (changes requested):
small change, but seems good

ui/src/dms/Dm.tsx
jamesacklin: Thanks!

PR #2551, opened 2023-05-23
groups: add message to connectivity check hi, block scry until hi completes

patosullivan:
This breaks playwright for now, I'll make a follow-up PR to fix playwright.


arthyn (commented):
might need to review this together

arthyn (approved):
lgtm

PR #2560, opened 2023-05-26
groups: unify dropdowns

### LINKED ISSUE #2176: system: dropdown component improvements
For a while, our "universal" product dropdown styling has been out of sync with the intended spec.

A selection of existing dropdowns:

Section 2

Some general issues with these dropdown interfaces:

- the onHover state for dropdown rows isn't standard across 
- some of these dropdowns include iconography/icon helpers while others don't
- some of these dropdowns include labels *for* the dropdown, while most don't
- some dropdowns omit nested corner radii
- the current drop shadow treatment for our dropdowns results in a top edge that bleeds into the background

These points signal that our dropdowns may be ripe for stricter componentization.

Within indigo, we've stated a relatively stable spec for a while that if implemented, would render most of the issues above null:

Section 1

The current spec lives here (use figma's inspect feature to view the details of shadows, radii, etc.):

https://www.figma.com/file/YLjErBShApNlWpE44nqnbC/Indigo-SS2023?node-id=2723%3A31969&t=9jMQqefjAx03CoXq-1

jamesacklin:
Unifies all system dropdowns. Fixes #2176 

image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image

### DIFF
--- a/ui/src/channels/ChannelHeader.tsx
+++ b/ui/src/channels/ChannelHeader.tsx
@@ -169,7 +169,7 @@ function ChannelActions({
                 Edit {prettyAppName()}
               </Dropdown.Item>
               <Dropdown.Item
-                className="dropdown-item text-red"
+                className="dropdown-item-red"
                 onClick={() => setDeleteChannelIsOpen(!deleteChannelIsOpen)}
               >
                 Delete {prettyAppName()}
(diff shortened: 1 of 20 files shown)


urcades (approved):
So beautiful, thanks so much for taking the time to implement these.

While I think we can ship as-is, one optional (minor) point of improvement we could tack on is to increase the upper edge's drop shadow spread by a few pixels/units, in order to ensure this part of the dropdown isn't lost against other white backgrounds.

urcades (approved):
So beautiful, thanks so much for taking the time to implement these.

While I think we can ship as-is, one optional (minor) point of improvement we could tack on is to increase the upper edge's drop shadow spread by a few pixels/units, in order to ensure this part of the dropdown isn't lost against other white backgrounds.

jamesacklin:
@urcades Heard, chef:

image

patosullivan (approved):
Looks great.

PR #2573, opened 2023-06-01
chat: threads refactor (wip)

arthyn:
Fixes LAND-460


arthyn:
after our meeting today:

- hang threads off knots
	- return thread summaries instead of threads when returning knots in main queries
	- when querying threads just return the messages hanging off the knot
- briefs
	- key: becomes whom or whom + thread
- version number bump
- marks updates based on types

arthyn:
closing until we revisit

PR #2577, opened 2023-06-02
groups: role admin dialog grabbag

### LINKED ISSUE #2556: groups: role deletion appears impossible
In the new Roles editor, deleting a role appears to be impossible just by reading the copy and looking at the UI. The "New York" role has a single member in it.

image

Although I've satisfied the condition for input (the inputted name matches the name of the role), the UI is still blocking me from deleting the role. Through some investigation, I learned that roles must be empty (not have any members assigned to them) before deleting them. Note that the copy above says this is possible to do, which makes the UI state all the more confusing.

**To Reproduce**
1. Go to a group you administrate.
2. Create a role.
3. Add people to the role.
4. Attempt to delete the role.

**Expected behavior**
- **Threshold**: the UI text should explain how to remove members from the role before deletion is possible.
- **Objective**: the UI should ignore the "does this role have members assigned to it" check, remove the members from the role, and delete the role (functionally unblocking the action).

### LINKED ISSUE #2575: chat: message metadata cramped in threads
The message metadata row has become extra cramped in threads now when the author also has a role badge applied to them. See the screenshot for reference. The role badge also gets a scrollbar added to it when it shrinks:

Screenshot 2023-06-01 at 7 47 02 PM

**To Reproduce**
1. Go to a threaded reply posted by an admin (e.g. /1/chan/chat/~nibset-napwyn/support/msg/~nocsyx-lassul/170.141.184.506.246.922.914.222.855.715.130.896.809)
2. Hover on the message metadata
3. Observe row doesn't have enough room for all the info in it

**Expected behavior**
Info is rendered in a line, or has some different design applied to it when in a thread

**Additional context:**
- Tlon-hosted planet
- macOS 13.3.1, Chromium

### LINKED ISSUE #2576: groups: member role sticks around in admin settings/badge overflow
When adjusting roles for a member in a group, the "Member" role sticks around even if it's not selected. This might be correct (everyone is in that sect) but the UI is slightly confusing.

image

Here's what I'd expect:
1. A group member starts out as a Member. They don't have a badge in channels, nor does their membership count towards the `sects.length` to determine the overflow rule.
2. An admin assigns a role to a member. This is reflected in the badges in channels and in the listing (the "Member" text disappears in the listing).
3. An admin removes a role from a member. The role badge disappears in channels and the "Member" text reappears in the listing.

jamesacklin:
fixes #2576
fixes #2556
fixes #2575


patosullivan (approved):
Looks good!

PR #2578, opened 2023-06-03
groups: show members outside of routed group

jamesacklin:
Fixes an untracked bug caused by #2492 where accessing the group info or edit modals from outside the group context (i.e. via a group sidebar item in the the All Groups view) would only show you as a member.


arthyn (approved):
lgtm

PR #2585, opened 2023-06-08
groups: Update groups protocol number

philipcmonk:
This addresses a live issue on the network, where group hosts use 100% CPU because of a resubscribe loop when their subscribers haven't received recent groups OTAs.

The solution is to update the protocol version number for groups subscriptions.  This is expected to happen every time the groups subscriptions result type changes.  It changed a little while ago with the addition of group roles, but we forgot to update the version.

This includes a small refactor of how those protocol numbers are used to make it easier to keep them in sync across the desk.  This refactor applies to groups, chat, diary, heap, and contacts.  We move the version handling into the sur file, so that anything using these apps will automatically use the most recent version (as long as the values still match the types).

Expected impact:
- As soon as either a host or joiner receives this OTA, the 100% CPU usage bug should subside.
- If you're in a group, then during the time when either you or the host have received this OTA but the other has not, the group metadata will not update, and there will be no notification about this in the UI.  This means you will not see updates to the group description, roles, or channel list.  You will, however, be able to use chat channels, notebooks, and galleries like normal.  Once both you and and the host have received this OTA, you will receive the backlog of group metadata changes, and everything will continue as normal.
- If you attempt to join a group while either you or the host have received this OTA but the other has not, the UI will hang on a spinner with the text "Wait a sec".  If you go back to the top of groups, you will see an grayed-out entry in the sidebar for this group (it may be blank or have the ship and group name in it, eg "~ship/group").  This will not automatically resume when the other party receives this OTA.  Instead, you should either "cancel join" or "leave group", and then when you're both up-to-date, you will be able to join the group successfully.

Fixes LAND-531

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -22,7 +22,6 @@
         odd=&
         veb=|
     ==
-  ++  okay  `epic:e`0
   ++  club-eq  2 :: reverb control: max number of forwards for clubs
   +$  current-state
     $:  %2
@@ -54,7 +53,7 @@
       abet:init:cor
     [cards this]
   ::
-  ++  on-save  !>([state okay])
+  ++  on-save  !>([state okay:c])
   ++  on-load
     |=  =vase
     ^-  (quip card _this)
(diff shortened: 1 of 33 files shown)


arthyn (approved):
lgtm

PR #2589, opened 2023-06-09
chat: post message to native app when copying to clipboard

alecananian:
- Fixes `useCopy` hook always returning `true` even though the copy failed
- Updates `useCopy` to post a message to the React Native webview when attempting to copy text


alecananian:
Looks like the change to `ChatMessage.test.tsx.snap` came during my pre-commit hook run. Anything I should do about that?

patosullivan (approved):
Looks good, just revert the change to the chat message snapshot.

alecananian:
Reverted! Though CI is still failing because it has its own timezone that doesn't match the snapshot

PR #2600, opened 2023-06-12
Fix link to members view in join notification

patosullivan:
We were still sending the old URL in the notification.

Fixes LAND-477


arthyn (approved):
lgtm

PR #2601, opened 2023-06-12
channels: reorg hook usage, change all brief usage

arthyn:
This should fix some performance issues we're seeing lately LAND-562 unsure why they've cropped up now though.

Fixes LAND-562


patosullivan (approved):
lgtm!

PR #2608, opened 2023-06-13
fix missing quips issue

patosullivan:
fixes LAND-544

We were inserting a letter rather than an outline into the cache, so if we attempted to load from the cache before the update came back from the backend we wouldn't have quips. Fixes other type issues around letter vs outline.


arthyn (approved):
lgtm, there's a reason we did diaryletter here but I can't remember exactly why and that might have changed with the react query refactor

PR #2617, opened 2023-06-16
chat: fix reaction/deletion updates

arthyn:
This fixes issues seen in recent updates that prevent reacts from showing up or being deleted until refresh. This seems to also improve the performance of sending/receiving chats based on the `ChatScroller` changes.

Fixes LAND-585


patosullivan (approved):
Looks good, runs well locally.

PR #2621, opened 2023-06-16
perf: mark warming

arthyn:
This change makes a "mark warmer" file which ensures that particular conversions are pre-compiled and cached so that we aren't doing the compilation on every request. This gives us a tremendous boost to perf. Thanks to @pkova and @philipcmonk


talk/app/talk-ui.hoon
```diff
@@ -1,5 +1,6 @@
 /-  u=ui, g=groups, c=chat
 /+  default-agent, dbug, verb, vita-client
+/%  init  %ui-init
```
joemfb, line 3: should this be `/$  init  %ui-init  %json`?

philipcmonk (approved):
Looks good

desk/app/chat.hoon
```diff
@@ -4,13 +4,14 @@
...
 /+  default-agent, verb-lib=verb, dbug
-/+  chat-json
 /+  pac=dm
 /+  ch=chat-hark
 /+  gra=graph-store
 /+  epos-lib=saga
 /+  wood-lib=wood
 /+  mig=chat-graph
+::  performance, keep warm
+/+  chat-json
```
joemfb, line 14: Is this (and all the other `*-json` lib imports in the %groups desk) still necessary with `lib/mark-warmer` being imported by `:groups-ui`?

joemfb (commented):
A couple questions, otherwise looks good.

desk/app/chat.hoon
arthyn: good question I'm not sure

talk/app/talk-ui.hoon
arthyn: yeah you're right

```suggestion
/$  init  %ui-init  %json
```

PR #2626, opened 2023-06-18
groups: make sidebar resizable

alarivan:
As a user I want to be able to easily see full names for channels/groups when they are longer than the sidebar width. This is especially annoying when there is important information cut off from names.

#### Changes:
- Make groups sidebar resizable using [tailwind resize class](https://tailwindcss.com/docs/resize)

#### Before:
Some channel names are not fully visible. The full name of the channel can only be viewed by hovering over channel name or by navigating to a channel.
image

#### After:
Sidebar can be resized to view full names for all the channels. 

https://github.com/tloncorp/landscape-apps/assets/48297104/b0d053b9-03bf-48d1-a9e6-2ce7e34c9c0b

### DIFF
--- a/ui/src/groups/GroupSidebar/GroupSidebar.tsx
+++ b/ui/src/groups/GroupSidebar/GroupSidebar.tsx
@@ -112,7 +112,7 @@ export default function GroupSidebar() {
   const privacy = group ? getPrivacyFromGroup(group) : 'public';
 
   return (
-    <nav className="flex h-full w-64 flex-none flex-col bg-white">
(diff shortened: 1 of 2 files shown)


arthyn:
@alarivan this is awesome. do you mind also doing it for the `MessagesSidebar` so this works in Talk as well? after that should be gtg

alarivan:
> @alarivan this is awesome. do you mind also doing it for the `MessagesSidebar` so this works in Talk as well? after that should be gtg

done

arthyn (approved):
lgtm, ty!

PR #2630, opened 2023-06-20
diary: fix more cache issues

patosullivan:
Fixes LAND-597 and another missing quips issue (i.e., the note in the cache was the wrong shape, did not include quips).

Also updates our optimistic update functionality to be in-line with the tanstack query docs on setQueryData: https://tanstack.com/query/v4/docs/react/reference/QueryClient#queryclientsetquerydata

Fixes LAND-589 as well.


patosullivan:
Should fix LAND-589 as well.

arthyn (approved):
lgtm

PR #2632, opened 2023-06-21
edit-channel: restore writer editing

arthyn:
During the notebooks refactor we deleted the role add for other channel types. This normalizes the way we do add/delete so that they always flow the same way so all channels have the same functionality.


patosullivan (approved):
lgtm

PR #2635, opened 2023-06-21
groups: fix cannot read properties of undefined (title) issue

patosullivan:
Apparently the `meta` property returned on a Group from useGroups could be undefined in some cases.

Fixes LAND-601


jamesacklin (approved):
THANK YOU!

PR #2636, opened 2023-06-21
notebooks: fix bad notebook cache issue

patosullivan:
Some users were still seeing the missing quipCount/quippers issue because we had accidentally stored full notes instead of outlines in some situations previously. Now, when we call useNotes, we'll check to see if they have a non-outline note and then clear out the cache for that query key if they do.


patosullivan:
Closed in favor of https://github.com/tloncorp/landscape-apps/pull/2637

PR #2638, opened 2023-06-22
add posthog tracking to talk and groups

alecananian:
Resolves LAND-552

- Adds PostHog JS/React library as a dependency and connects tracking init call
- Identifies logged in ship with PostHog


alecananian:
Closing this because we are taking a different strategy with event logging

PR #2651, opened 2023-06-26
notebooks: more block content fixes

patosullivan:
- fixes LAND-389 with consistent "Remove" buttons for all blocks
- fixes LAND-599 by making sure we always have a cursor available after inserting a block (i.e., after the code block in particular), and making sure the image block replaces the current empty block.
- Some more style fixes (language selector dropdown in code blocks matches our other dropdowns).


arthyn (approved):
lgtm

PR #2658, opened 2023-06-28
groups: add PostHog analytics; capture Groups events

alecananian:
Fixes LAND-552

- Wraps application in the PostHog provider with all auto-capturing settings disabled
- Captures Groups-related events specific in the Linear ticket, taking into account group privacy settings


patosullivan (approved):
Code looks good! I think we should maybe rethink using the ship name as an identity, I'm not sure that really gives us any useful information and it may prove to be a liability later on.

I'm also not seeing any data streaming into the landscape mainnet project on posthog (even after setting `  opt_out_capturing_by_default: false,`). Am I looking at the right project? Feel free to ping me about it out of band.

jamesacklin:
Thanks!

PR #2665, opened 2023-06-29
chat: display status indicator for pals

Fang-:
If a ship is known in the pals state, we display a status indicator similar to the one in the original pals interface. Tooltip displays the tags/lists for that pal, if outgoing.

Since we already have the pals data (for use in leap), this is simply the addition of an indicator component.

image

Be careful, baby's first new-groups PR.


ui/src/components/PalIcon.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const outgoingSvg = (
+  <svg viewBox="0 0 100 100" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path
+      d="M50 95
+C44.0905 95 38.2389 93.836 32.7792 91.5746
+C27.3196 89.3131 22.3588 85.9984 18.1802 81.8198
+C14.0016 77.6412 10.6869 72.6804 8.42542 67.2208
+C6.16396 61.7611 5 55.9095 5 50"
+      stroke="black"
```
arthyn, line 14:
```suggestion
      stroke="currentColor"
```

ui/src/components/PalIcon.tsx
```diff
@@ -0,0 +1,146 @@
...
+L60.2062 60.2062
+L67.6777 32.3223
+Z
+M34.0901 69.4454
+L53.5355 50
+L50 46.4645
+L30.5546 65.9099
+L34.0901 69.4454
+Z"
+      fill="black"
```
arthyn, line 35:
```suggestion
      fill="currentColor"
```

ui/src/components/PalIcon.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const incomingSvg = (
+  <svg viewBox="0 0 100 100" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path
+      d="M95 50
+C95 55.9095 93.836 61.7611 91.5746 67.2208
+C89.3131  72.6804 85.9984 77.6412 81.8198 81.8198
+C77.6412 85.9984 72.6804 89.3131 67.2208 91.5746
+C61.7611 93.836 55.9095 95 50 95"
+      stroke="black"
```
arthyn, line 48:
```suggestion
      stroke="currentColor"
```

ui/src/components/PalIcon.tsx
```diff
@@ -0,0 +1,146 @@
...
+L27.9939 33.0295
+L31.5294 29.4939
+Z
+M53.0962 51.0607
+L31.5294 29.4939
+L27.9939 33.0295
+L49.5607 54.5962
+L53.0962 51.0607
+Z"
+      fill="black"
```
arthyn, line 69:
```suggestion
      fill="currentColor"
```

ui/src/components/PalIcon.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const mutualSvg = (
+  <svg viewBox="13 0 113 100" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path
+      d="M108 50
+C108 61.9348 103.259 73.3807 94.8198 81.8198
+C86.3807 90.2589 74.9347 95 63 95
+C51.0653 95 39.6193 90.2589 31.1802 81.8198
+C22.7411 73.3807 18 61.9347 18 50"
+      stroke="black"
```
arthyn, line 82:
```suggestion
      stroke="currentColor"
```

ui/src/components/PalIcon.tsx
```diff
@@ -0,0 +1,146 @@
...
+C76.0057 34.7635 78.5945 37.3523 81.7585 37.3523
+C84.9226 37.3523 87.5114 34.7635 87.5114 31.5994
+Z
+M51.0767 31.5994
+C51.0767 28.4354 48.4879 25.8466 45.3239 25.8466
+C42.1598 25.8466 39.571 28.4354 39.571 31.5994
+C39.571 34.7635 42.1598 37.3523 45.3239 37.3523
+C48.4879 37.3523 51.0767 34.7635 51.0767 31.5994
+Z"
+      fill="black"
```
arthyn, line 98:
```suggestion
      fill="currentColor"
```

ui/src/components/PalIcon.tsx
```diff
@@ -0,0 +1,146 @@
+import { useMemo } from 'react';
+import _ from 'lodash';
+import { usePals } from '@/state/pals';
+import * as Tooltip from '@radix-ui/react-tooltip';
```
arthyn, line 4: seems unused

arthyn (changes requested):

ui/src/components/PalIcon.tsx
Fang-: Good catch. I was thinking of using this for on-hover details, but `title` works fine for now.

urcades:
This placement is so good, exactly what I'd personally expect 🙏

@Fang- are you amenable to a custom icon being developed for pals here? I won't take offense if you'd rather keep them, just thinking about overall legibility and modularity of this little area.

Right now (from the looks of the image, I haven't run this locally) it appears as though these badges are being treated similarly to the role badges, as text or an image within a div with padding. I'd gently encourage us to consider what a visual element here that makes full use of the "element space" looks like, with no padding, using all ~16pt of the vertical height.

Fang-:
I'm open to whatever Tlon would do here, because I'm implementing this as a Tlon employee, not a paldev spy (for the record (; ). I just used the svgs from the pals ui because I didn't have anything better on hand, would be open to whatever we have that looks better.

(But, if we don't have anything ready to go, _maybe_ better to just merge and update the visuals later?)

urcades:
@Fang- Agree — let's merge this (it's amazing), and I'll create a task to track a more dialed-in tlon-specific implementation of the pals icons

urcades (approved):
Let's go. (Adding Linear task to track design update)

PR #2668, opened 2023-06-30
dms: fix empty scroller on new DMs

patosullivan:
At the point at which we create a new dm, a writWindow does not yet exist in state. In that case, we can just show the writs we have.

Fixes LAND-630.

I could not repro the DM not appearing in the sidebar, but it would be a separate issue afaict.


arthyn:
@patosullivan pushed up something that fixes what I was seeing, this should be gtg

PR #2669, opened 2023-06-30
profiles: include pals icon & actions in modal

Fang-:
Includes the same `PalIcon` that's displayed in chat, and includes add/remove pal buttons if pals is installed.

image

Putting it by the rest of the actions felt most natural. The "remove" action gets a light red color because it's destructive. But I'm open to other options. Maybe we want a gray "Remove Pal..." instead that asks for confirmation?

image

I suspect we may find the color in these elements clashing with the customization in the profile modal. Maybe it's fine? Maybe it's not?

(Targeting #2665 because this builds on top of that, but once that merges we can just target develop as normal.)

### DIFF
--- a/ui/src/profiles/ProfileModal.tsx
+++ b/ui/src/profiles/ProfileModal.tsx
@@ -6,6 +6,8 @@ import useContactState, { useContact } from '@/state/contact';
 import Avatar from '@/components/Avatar';
 import Dialog from '@/components/Dialog';
 import ShipName from '@/components/ShipName';
+import PalIcon from '@/components/PalIcon';
+import usePalsState from '@/state/pals';
 import useNavigateByApp from '@/logic/useNavigateByApp';
 import ProfileCoverImage from './ProfileCoverImage';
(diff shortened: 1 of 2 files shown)


ui/src/profiles/ProfileModal.tsx
```diff
@@ -114,7 +118,22 @@ export default function ProfileModal() {
...
+            className="secondary-button ml-auto bg-red-100"
+            onClick={() => pals.removePal(ship.slice(1))}
+          >
+            Remove Pal
+          </button>
+        ) : (
+          <button
+            className="secondary-button ml-auto"
+            onClick={() => pals.addPal(ship.slice(1))}
+          >
```
arthyn, line 132: using `deSig` might be more legible here

arthyn (approved):
lgtm

urcades (approved):
@Fang- I think your hunches here for placement + color are correct

A part of me is thinking a little ahead and wanting to slightly overhaul the arrangement of profiles (for example, how do other apps stack additional actions here? that bottom bar can only hold so many actions), but that can wait another day.

lgtm!

PR #2671, opened 2023-07-02
groups: add privacy notice

jamesacklin:
Adds a deep explainer for our PostHog analytics. This is a generally triggerable modal so we can use it in whatever context we need; at the moment, I threw a simple trigger in the "?" wayfinding menu.

This isn't the option itself, which we'll need to add to App Settings - this is just to explain the what and why.

fixes LAND-552

### DIFF
--- a/ui/src/groups/PrivacyNotice.tsx
+++ b/ui/src/groups/PrivacyNotice.tsx
@@ -110,7 +110,7 @@ export default function PrivacyNotice() {
               <NavLink
                 to="/settings"
                 className=""
-                state={{ backgroundLocation: state.backgroundLlocation }}
+                state={{ backgroundLocation: state.backgroundLocation }}
               >
                 opt out in App Settings
               </NavLink>{' '}
--- a/ui/src/groups/PrivacyNotice.tsx
+++ b/ui/src/groups/PrivacyNotice.tsx
@@ -1,0 +1,189 @@
+import React from 'react';
+import { useDismissNavigate } from '@/logic/routing';
+import Dialog from '@/components/Dialog';
+import { NavLink, useLocation } from 'react-router-dom';
+
+export default function PrivacyNotice() {
+  const dismiss = useDismissNavigate();
+  const { state } = useLocation();
+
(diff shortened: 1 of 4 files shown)


ui/src/groups/PrivacyNotice.tsx
```diff
@@ -0,0 +1,189 @@
...
+              Tlon doesn&rsquo;t and will never collect your data without
+              permission. We will never sell your data. We will never use your
+              data without explaining why or what we&rsquo;re doing with
+              anything we&rsquo;ve collected.
+            </p>
+            <p>
+              The most important point you should take away from this disclosure
+              is our continued commitment to transparency. We&rsquo;re not
+              trying to pull one over on you. You can{' '}
+              <NavLink
```
arthyn, line 110: same here

ui/src/groups/PrivacyNotice.tsx
```diff
@@ -0,0 +1,189 @@
...
+            <p>
+              By default,{' '}
+              <em>
+                we never track anything you do on Landscape as a self-hosted
+                user.
+              </em>{' '}
+              However, you can opt in to help us. If you&rsquo;re piloting a
+              Tlon-hosted ship, we collect surface level data about your ship
+              activity unless you opt out,{' '}
+              <NavLink
```
arthyn, line 73: Not sure why but this link doesn't work, brings you to a blank page

arthyn (changes requested):

ui/src/groups/PrivacyNotice.tsx
jamesacklin: Had a typo in state.backgroundLocation.

ui/src/groups/PrivacyNotice.tsx
jamesacklin: This one works; the other didn't -- they look very similar; can you re-verify?

arthyn: hmm both seem to do this now 
image

arthyn (approved):
lgtm

PR #2673, opened 2023-07-03
groups: better roles handling, ensure valid states

arthyn:
This ensures that we recheck permissions upon any relevant changes. It also tries to clean up any role state that is currently invalid. Also makes sure the frontend doesn't actually send "members" as a role and cleans up a few things there.

Fixes LAND-616
Fixes LAND-621

### DIFF
--- a/desk/app/diary.hoon
+++ b/desk/app/diary.hoon
@@ -487,11 +487,11 @@
       [%cabal * %del *]
     =/  =sect:g  (slav %tas p.diff)
     %+  recheck-perms  affected
-    `(~(gas in *(set sect:g)) ~[p.diff])
+    (~(gas in *(set sect:g)) ~[p.diff])
   ==
 ::
 ++  recheck-perms
-  |=  [affected=(list flag:d) sects=(unit (set sect:g))]
+  |=  [affected=(list flag:d) sects=(set sect:g)]
   ~&  "%diary recheck permissions for {<affected>}"
   %+  roll  affected
   |=  [=flag:d co=_cor]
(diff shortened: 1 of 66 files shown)


desk/app/diary.hoon
```diff
@@ -465,33 +465,38 @@
...
-    =/  di  (di-abed:di-core:co flag)
-    di-abet:di-recheck:di
+      [%cabal * %del *]
+    =/  =sect:g  (slav %tas p.diff)
+    %+  recheck-perms  affected
+    `(~(gas in *(set sect:g)) ~[p.diff])
   ==
 ::
+++  recheck-perms
+  |=  [affected=(list flag:d) sects=(unit (set sect:g))]
```
midsum-salrux, line 494: NB: would calling this with an empty set in the null case be clearer than using a unit or is there a semantic difference between the two?

midsum-salrux (approved):
LGTM with one non-blocking question

ui/src/groups/GroupAdmin/GroupRoles.tsx
```diff
@@ -370,7 +370,8 @@ export default function GroupRoles({ title }: { title: string }) {
               >
                 remove all members
               </Link>{' '}
-              from this role before it can be deleted.
+              from this role and remove it from all channels before it can be
+              deleted.
```
jamesacklin, line 374: Ah-ha! Thanks.

desk/app/diary.hoon
arthyn: hmm yeah we probably could use an empty set here

PR #2678, opened 2023-07-06
profiles: hide pals action if not installed

Fang-:
The conditional for this was implemented incorrectly. Additionally, the styling would not account for it.

Alongside the fix, we refactor the `!contacts` branch into a single main rendering branch, within which we check that condition to decide whether to include a component. Doing so prevents us from duplicating the now slightly more complex button rendering logic, and will make additions and modifications easier going forward.

This seems to behave correctly now, but please scrutinize closely. This is the best factoring I could come up with, but maybe it was deliberately not done like this initially.

Fixes LAND-643.

### DIFF
--- a/ui/src/profiles/ProfileModal.tsx
+++ b/ui/src/profiles/ProfileModal.tsx
@@ -84,33 +84,28 @@ export default function ProfileModal() {
           </div>
         )}
       </div>
-      <footer className="flex items-center py-4 px-6">
+      <footer className="flex items-center justify-end space-x-2 py-4 px-6">
         {pals.installed &&
           ship !== window.our &&
           (pals.pals.outgoing[ship.slice(1)] ? (
             <button
-              className="secondary-button ml-auto bg-red-100 dark:bg-red dark:text-white"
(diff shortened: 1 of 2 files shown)


jamesacklin (approved):
Verified, does not appear on a pals-less ship.

ui/src/profiles/ProfileModal.tsx
```diff
@@ -107,36 +71,43 @@ export default function ProfileModal() {
...
+              className="secondary-button ml-auto"
+              onClick={() => pals.addPal(ship.slice(1))}
+            >
+              Add Pal
+            </button>
+          ))}
+        <button
+          className={`secondary-button ${
+            pals.installed && ship !== window.our ? 'ml-2' : 'ml-auto'
+          }`}
```
arthyn, line 108: let's remove this and all the margins on the buttons and use `justify-end space-x-2` on the footer, should be simpler

arthyn (changes requested):
small change, but other than that looks good to me

jamesacklin (approved):

arthyn (approved):
lgtm

PR #2682, opened 2023-07-07
gallery: generalized cleanup, better refs

jamesacklin:
- Adds a gradient mask to text HeapBlock items
- Reformats the attribution row to HeapBlock items to show poster (sigil), time, and comment count
- Refactors the HeapRow component to behave more like the HeapBlock and brings into alignment with design
- Adds a `contextApp` prop and `children` to all content references in order to:
  - Reformat group references to conform with LAND-366 in both Block and List views
  - Reformat all other references to match in both Block and List views
  - Add isomorphic curio reference rendering to Gallery (the source and destination presentation appear the same in the Block and List views; click through to see reference in situ)

Fixes LAND-366
Fixes LAND-387

### DIFF
--- a/ui/src/components/References/BaitReference.tsx
+++ b/ui/src/components/References/BaitReference.tsx
@@ -16,9 +16,13 @@ import WritBaitReference from './WritBaitReference';
 function BaitReference({
   bait,
   isScrolling,
+  contextApp,
+  children,
 }: {
   isScrolling: boolean;
   bait: BaitCite['bait'];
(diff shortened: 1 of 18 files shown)


arthyn:
we probably should component-ize the block/row display in refs because there's lots of similar things going on

jamesacklin:
@arthyn Pushed a few changes, PTAL

arthyn (approved):
KILLING BLOW

PR #2686, opened 2023-07-09
groups: mobile layout issues

dantish:
Fixes LAND-607


patosullivan (approved):
Looks great

jamesacklin (approved):
Agreed, looks great. Thank you for the showSheet prop; we can use that pattern elsewhere. 

I think we just need an X button in the GroupInviteDialog; other than that, we should be good to go.

PR #2688, opened 2023-07-10
analytics: limit capturing even further

arthyn:
I though that opt-ins were more session based, but they are persistent and also span the origin. So that means that the talk check that I moved was incorrect and needed to stay since once you're cookied the captures will fire no matter what. 

In addition I've made some attempts at limiting the URL properties that get sent, it seems to be working, but I'm unsure if it was the `unset` or the `setOnce` that did it. Either way we seem to properly hide that information now.


alecananian (approved):
Looks good, thanks for finding these additional properties

patosullivan (approved):

PR #2689, opened 2023-07-10
notifications: various route issues

patosullivan:
fixes LAND-641 (routes to the correct place for join/leave notifs), also updates paths sent from the backend for leaves and role changes.


arthyn (approved):
lgtm

PR #2693, opened 2023-07-11
notebooks: add markdown editing

patosullivan:
Fixes LAND-368

This works independently of tiptap using bare prosemirror but users can still switch back to tiptap if they want.

@urcades thoughts on how this is presented?

Looks like this:

https://github.com/tloncorp/landscape-apps/assets/1221094/1c8c0a2f-ee57-497f-b30a-525e783cc743


patosullivan:
Spent most of the day working on parsing our special blocks from MD into the prosemirror AST and working on edge cases/weirdness. Next step is to work on using diaryMixedToJSON to hydrate the MD editor directly without going through the extra tiptap steps.

patosullivan:
This should be good for review now.

ui/src/diary/diary-add-note.tsx
```diff
@@ -16,18 +16,24 @@ import {
...
 import DiaryInlineEditor, { useDiaryInlineEditor } from './DiaryInlineEditor';
+import DiaryMarkdownEditor from './DiaryMarkdownEditor';
 
 export default function DiaryAddNote() {
   const { chShip, chName, id } = useParams();
   const [loaded, setLoaded] = useState(false);
+  const [editWithMarkdown, setEditWithMarkdown] = useState(true);
+  const [updateMarkdown, setUpdateMarkdown] = useState(false);
+  const [updateTipTap, setUpdateTipTap] = useState(true);
+  const [editorContent, setEditorContent] = useState<JSONContent | null>(null);
```
arthyn, line 36: the logic around updating with these flags seems like it would be easy to mess up. can we streamline this further or make these unnecessary somehow? like maybe we always have both the editor content and the markdown contents ready so flipping the switch is just really visual?

arthyn (changes requested):
also saw an error when trying to publish a pasted ref:

Screenshot 2023-07-18 at 4 43 50 PM

ui/src/diary/DiaryMarkdownEditor.tsx
```diff
@@ -0,0 +1,257 @@
...
+    if (markdownInput === null) {
+      return;
+    }
+    const newContent = deserialize(schema, markdownInput);
+    setEditorContent(newContent);
+    setUpdateTipTap(true);
+  }, [markdownInput, setEditorContent, setUpdateTipTap]);
+
+  return (
+    <div className="h-[600px] w-full">
```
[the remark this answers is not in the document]
patosullivan: I just set a fixed height while waiting on input from ed, basically

urcades (changes requested):
As hunter may have mentioned elsewhere, I think we can stand to drop the gray background and border surrounding the markdown text, and simply treat the markdown input as though it's another state of the writing surface.

If we wanted to further emphasize the difference between markdown editing and WYSIWYG editing, we could maybe switch the markdown editing text to our monospace typeface. This gets us feeling a little bit like how IA writer distinguishes between markdown editing and the text preview mode.

patosullivan:
Just made some style updates, now looks like this, @urcades:

 

https://github.com/tloncorp/landscape-apps/assets/1221094/ad23a414-448d-48da-87c9-d84362ab985f

urcades (approved):
beautiful

ui/src/components/Setting.tsx
```diff
@@ -34,7 +37,12 @@ export default function Setting({
           loading={status === 'loading'}
         />
         <div className="flex flex-1 flex-col justify-center">
-          <h3 id={id} className="flex items-center font-semibold leading-6">
+          <h3
+            id={id}
+            className={cn('flex items-center leading-6', {
+              'font-semibold': bold,
```
arthyn, line 43: maybe we should just open className above this and remove the font-semibold altogether and add it to anywhere setting was used?

arthyn (approved):
lgtm, LFG!

ui/src/components/Setting.tsx
patosullivan: good idea, just made that change, will merge now

PR #2694, opened 2023-07-11
GRO-598 Automatic Group Join for Lure

midsum-salrux:
https://linear.app/tlon/issue/GRO-598/lure-automatic-group-join

Read a cookie (to be set by the Lure landing page in the future) and auto-join the corresponding group when it's present.


ui/src/groups/autoJoinLureInvites.tsx
```diff
@@ -0,0 +1,13 @@
+import { Gangs } from '@/types/groups'
+import useGroupJoin from '@/groups/useGroupJoin';
+import cookies from 'browser-cookies';
+
+export default function AutoJoinLureInvites(pendingGangs: Gangs) {
```
patosullivan, line 5: This should probably be converted into a hook that returns a "autoJoinLureInvites" function. 

`useGroupJoin` returns some functions that I think you probably want to use (maybe just the `join` function) rather than calling the hook directly.

patosullivan (changes requested):
Hey, thanks for working on this, left a comment about doing this a slightly different way.

arthyn (approved):
lgtm!

PR #2695, opened 2023-07-12
groups/chat: show message options when content is larger than screen height

latter-bolden:
Fixes LAND-644

Ensures the react/thread/etc. menu is always visible if a message extends off the screen.

https://github.com/tloncorp/landscape-apps/assets/90741358/5068ffc1-2dc1-45d9-952b-a1de728ba8ff


arthyn (approved):
lgtm!

PR #2700, opened 2023-07-13
diary: update increment

arthyn:
This updates @patosullivan 's branch so that we correctly increment the epic and negotiate versions correctly. Notably this means we need some signalling around version mismatch, will handle in LAND-532.

### DIFF
--- a/desk/app/diary.hoon
+++ b/desk/app/diary.hoon
@@ -117,7 +117,6 @@
     ==
   +$  state-1  current-state
   ++  zero  zero:old:d
-  ++  one  d
   ++  state-0-to-1
     |=  s=state-0
     ^-  state-1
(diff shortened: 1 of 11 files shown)


desk/app/diary.hoon
```diff
@@ -110,12 +110,14 @@
...
-        voc=(map [flag:d plan:d] (unit said:d))
+        shelf=shelf:zero
+        voc=(map [flag:zero plan:zero] (unit said:zero))
         ::  true represents imported, false pending import
-        imp=(map flag:d ?)
+        imp=(map flag:zero ?)
     ==
   +$  state-1  current-state
+  ++  zero  zero:old:d
+  ++  one  d
```
midsum-salrux, line 120: NB: Doesn't look like this is used.

midsum-salrux (approved):
LGTM with one non-blocker

desk/app/diary.hoon
arthyn:
```suggestion
```

PR #2703, opened 2023-07-14
upgrade Vite to latest

latter-bolden:
On the tin.

I confirmed `dev`, `chat`, and `build` scripts still run fine after the update, but I'm still unfamiliar with what our workflow looks like more broadly — let me know if anything seems off.


arthyn (approved):
lgtm

PR #2704, opened 2023-07-14
groups: only allow self or host to join a channel

patosullivan:
fixes LAND-671

Prevents anyone other than ourselves or a group host from joining our ship to a channel.


arthyn (changes requested):
lgtm, can you add the same thing for group joins?

patosullivan:
> lgtm, can you add the same thing for group joins?

only our ship though, right (that's what I did in my latest commit)?

arthyn (approved):
yep, lgtm!

PR #2706, opened 2023-07-15
groups: use existence checks to ensure we can scry

arthyn:
We weren't checking whether or not a channel existed before scrying so this caused any updates that deleted roles to crash, causing a kick-resubscribe loop. This should fix the resubscribe loop for those who are receiving updates.  

Fixes LAND-675

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -771,14 +771,18 @@
   ::
       [%u %dm @ *]
     =/  =ship  (slav %p i.t.t.path)
-    ?.  (~(has by dms) ship)
+    =/  has  (~(has by dms) ship)
+    ?.  has
       ``loob+!>(|)
+    ?~  t.t.t.path  ``loob+!>(has)
     (di-peek:(di-abed:di-core ship) %u t.t.t.path)
   ::
       [%u %club @ *]
     =/  =id:club:c  (slav %uv i.t.t.path)
-    ?.  (~(has by clubs) id)
+    =/  has  (~(has by clubs) id)
+    ?.  has
       ``loob+!>(|)
+    ?~  t.t.t.path  ``loob+!>(has)
     (cu-peek:(cu-abed:cu-core id) %u t.t.t.path)
   ::
       [%u %chat @ @ *]
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -772,21 +772,23 @@
       [%u %dm @ *]
     =/  =ship  (slav %p i.t.t.path)
     ?.  (~(has by dms) ship)
-      ``flag+!>(|)
+      ``loob+!>(|)
     (di-peek:(di-abed:di-core ship) %u t.t.t.path)
   ::
       [%u %club @ *]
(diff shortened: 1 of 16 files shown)


desk/app/groups.hoon
```diff
@@ -254,6 +254,8 @@
     ?.  (gth ~(wyt in readers) 0)  cr
     =/  action  [flag now.bowl %channel nest %del-sects readers]
     cr(cards [[%pass /groups/role %agent [our.bowl dap.bowl] %poke [act:mar:g !>(action)]] cards.cr])
+  =+  .^(has=? %gu (welp (channel-scry nest) ~))
```
Fang-, line 257:
```suggestion
  =+  .^(has=? %gu (channel-scry nest))
```

Also consider just inlining this `.^` into the `?.`.

desk/sur/groups.hoon
```diff
@@ -3,7 +3,7 @@
 /-  grp=group-store
 /-  metadata-store
 |%
-++  okay  `epic:e`1
```
Fang-, line 6: Not clear to me what's up with this and the v3 marks. We didn't change the types or subscription API, right?

desk/app/chat.hoon
```diff
@@ -772,21 +772,23 @@
...
-      ``flag+!>(|)
+      ``loob+!>(|)
     (di-peek:(di-abed:di-core ship) %u t.t.t.path)
   ::
       [%u %club @ *]
     =/  =id:club:c  (slav %uv i.t.t.path)
     ?.  (~(has by clubs) id)
-      ``flag+!>(|)
+      ``loob+!>(|)
     (cu-peek:(cu-abed:cu-core id) %u t.t.t.path)
```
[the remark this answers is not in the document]
arthyn: might need to get on gather for this

philipcmonk (changes requested):
This looks good, except I think we need to add `%group-action-2` in +poke and change the tags in +go-take-update.

How has this been tested?

arthyn:
bah thought I pushed that change ty @philipcmonk

arthyn:
@philipcmonk before I added the new marks, this was tested by having one moon with channels that are protected and not joinable along with a some extra roles existing. a second moon joins the group. the original moon then deletes a role which triggers `verify-group-cabals`.

philipcmonk:
Great.  Have you tested it after adding the new marks?  At least to ensure that two moons with the changes can get groups updates from each other, and that they don't get updates from old ships (similar to our tests last time).

arthyn:
not yet attempting now

desk/sur/groups.hoon
arthyn: yeah this was the reason, no state changes just a semi-convenient way to break the subscription

arthyn:
@philipcmonk testing looks good

philipcmonk (approved):
Great, then this looks good

arthyn:
I have a separate branch not on develop with the same changes that will be deployed: https://github.com/tloncorp/landscape-apps/tree/release-cabal-hotfix, but I'll merge this so this can be in the next release too

PR #2707, opened 2023-07-15
groups: app info and group modals no longer scrollable; incorrect spacings

dantish:
Fixes LAND-666

### DIFF
--- a/ui/src/components/SettingsDialog.tsx
+++ b/ui/src/components/SettingsDialog.tsx
@@ -48,7 +48,7 @@ export default function SettingsDialog() {
 
   function renderContent() {
     return (
-      <div className="flex flex-col space-y-8 overflow-y-auto">
+      <div className="flex flex-col space-y-8">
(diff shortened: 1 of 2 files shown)


arthyn:
sorry I should have put more context on LAND-666 it was meant to be for https://github.com/tloncorp/landscape/blob/master/ui/src/components/AppInfo.tsx in Landscape

dantish:
> sorry I should have put more context on LAND-666 it was meant to be for https://github.com/tloncorp/landscape/blob/master/ui/src/components/AppInfo.tsx in Landscape

Ooops....😅
Okay, let's fix it too👍

dantish:
@arthyn seems like the app info modal is scrollable, can you reproduce it now?
https://github.com/tloncorp/landscape-apps/assets/22919953/62b38d25-8514-421c-a387-185df551a95e

arthyn:
@dantish ah sorry again haha, it's only through search apparently! if you click get urbit apps and then click get on one of the apps you don't have you should see.

PR #2708, opened 2023-07-16
talk: flickering channel actions dropdown

dantish:
Fixes LAND-574


arthyn (changes requested):
so we put the simple rendering of the bullet in when not hovering to stop rendering so many dialogs when we have a giant list, so they only are there when you hover. 

I think you can solve this by going back to the original and adding `&& !isOpen` to line 92 and then you can remove `!isOpen &&` from 93.

dantish:
> so we put the simple rendering of the bullet in when not hovering to stop rendering so many dialogs when we have a giant list, so they only are there when you hover.
> 
> I think you can solve this by going back to the original and adding `&& !isOpen` to line 92 and then you can remove `!isOpen &&` from 93.

Totally makes sense, thank you for the suggestion! 👍

arthyn (approved):
lgtm

PR #2710, opened 2023-07-17
handle case where notebook host is offline

patosullivan:
This allows a user to create a post while the notebook host is offline or unreachable.

The post gets stored in the cache and we add the temporary noteId to an array of pendingNotes via a zustand store.

We check if a noteId is in that array, if it is then the note headline in the notebook list his grayed out and a warning is shown on the note itself when viewed.

We use vitals to check if the host is online. If they are, and if we have pendingNotes, then we fetch the notes from the live server and see if the live notes include notes where the `sent` time is equal to our cached note IDs. If they are, we have cached notes that need to be replaced by real notes.

Also, if the user is looking at one of the cached notes and we know we have a real note, navigate the user to the real note.

*Note*: this does *not* implement drafts, it just fixes some cacheing issues around notes and surfaces up the fact that a note is cached to the user (similar to the way we handle cached/not-yet-posted messages in chat). We need a separate conversation on a drafts feature. We most likely need two types of drafts, one that is stored only on your machine (a "private draft", which would be stored on only the user's ship and could be better than relying solely on this optimistic update strategy, and a "public draft" which other writers could see and edit).

Looks like this:

image
image

fixes LAND-673


ui/src/state/diary/diary.ts
```diff
@@ -534,16 +574,14 @@ export function useAddNoteMutation() {
 
       if (notes !== undefined) {
         // for the unlikely case that the user navigates away from the editor
-        // before the mutation is complete, we update the cache optimistically
+        // before the mutation is complete, or if the host ship is offlie,
```
arthyn, line 577:
```suggestion
        // before the mutation is complete, or if the host ship is offline,
```

ui/src/state/diary/diary.ts
```diff
@@ -160,6 +170,25 @@ export function useNotes(flag: DiaryFlag) {
...
+      });
+    getNotes().then((n) => {
+      if (n !== null && typeof n === 'object') {
+        notes.current = n as DiaryOutlines;
+      }
+    });
+  }, [flag]);
+
+  return notes.current;
+}
```
arthyn, line 190: This should probably use react query with a special key for host only. as is if this takes too long notes will just be undefined by the time it gets used, and since this uses a ref it won't trigger re-renders to recheck w/e is using this.

arthyn (changes requested):
this looks good besides the comment on host notes

latter-bolden (commented):
Is the case where our ship is offline something we care about for this PR?

Testing locally, it worked like a charm when the group host is down, but killing our ship and reconnecting seemed to leave the draft in an unpublished state.

Not sure if that situation is in scope for this work or something we should handle once we get to the larger drafts refactor we'd spoken about.

jamesacklin:
Yeah, let's handle that in a separate issue. I think there will be much more drastic effects in the hosted case (might 404/redirect before they even get to this point).

latter-bolden (approved):

PR #2715, opened 2023-07-21
notebooks: fix two issues from markdown work

patosullivan:
This updates the conditions for hydrating the markdown editor which fixes an issue where if the user started writing a new note using tiptap and then switched to markdown they wouldn't see anything in the markdown editor.

Also fixes another issue in a settings.ts hook where we attempted to parse diary.settings when it didn't exist because we didn't need to check that before (if we didn't have a diary key in settings before, we would just early return, if we did then we assumed we had the settings key within it, now there can be cases where wee don't have the settings key because we can have {diary: { markdown: true }}, which exists outside of the per-notebook {diary: {settings: ['flag', 'settingName', 'value']}} structure).


latter-bolden (approved):
tested locally, lgtm

PR #2717, opened 2023-07-21
groups: support non-@tas characters in role names

latter-bolden:
OTT

Fixes LAND-667


jamesacklin (approved):
image

Works on my machine.

patosullivan (approved):

PR #2721, opened 2023-07-23
find: show all groups if the group is not found

dantish:
Fixes LAND-381


jamesacklin (changes requested):
Remove the group ID from the flag here; should be "Groups hosted by ~nibset-napwyn:".
Screenshot 2023-07-24 at 11 46 47 AM

jamesacklin (approved):
Looks good, thank you!

PR #2724, opened 2023-07-24
notebooks: various fixes (and fix tests)

patosullivan:
Fixes all issues in LAND-707, also fixes an issue with navigating back from a new note to the notebook and issues with jest tests failing around posthog imports.

The react suspense/lazy routing thing I mentioned in standup was a red herring, the issue was actually caused by a bad schema spec (and somehow we only hit this error in safari, which is odd).


jamesacklin (commented):
Works in all cases for me:
- Untitled notes are now forbidden (thank you)
- Flipping between Markdown and rich-text works as expected
- Everything appears to work in Safari

latter-bolden (approved):

PR #2725, opened 2023-07-24
chat: fix issue with loading old threads on mobile

patosullivan:
Fixes LAND-682

We were previously initializing chat threads by calling `useChatState().initialize()`. This wasn't an issue if

a) you clicked into the thread on desktop, so we already loaded the relevant messages b) you navigated to it directly on mobile (via notification or similar) and the thread happened to be included in the first 50 messages in the channel

It *was* an issue if you navigated to the thread directly on mobile and the thread wasn't included in the initial 50 messages.

We actually need to fetch the messages around the thread parent ID to initialize the channel in that instance.


arthyn (approved):
lgtm

PR #2698, opened 2023-07-13
groups: connectivity checks

jamesacklin:
Adds stoplight-style connection status (with title attributes) to:

- 1:1 DM headers in Talk
- ChatInfo modals for multi-DMs in Talk
- Profile modals in Groups
- Group host search in Discover / Groups
- Channel headers in Groups

Hides the connection status in the Group Admin view and the Profile for your own ship; displays a green-light status in channel headers and the ChatInfo modal.

Fixes LAND-664
Fixes LAND-391

PR #2726, opened 2023-07-24
groups/channels: connection and saga visibility

arthyn:
This includes @jamesacklin 's previous work to add connection indicators everywhere in #2698. This PR adds on top of that to also include protocol compatibility checking and adds a few more places to check including the invite dialog and all channels list as well as anywhere that we perform actions on the group/channel.

Fixes LAND-532


jamesacklin:
Checked this out locally (and committed the desk changes), connectivity indicators look good and function.

Is there a way I can bump a saga here or there to check out the disabled states?

patosullivan (changes requested):
I don't see an issue with the code, but when trying to access a group hosted by another dev moon it never loads, just gets stuck in the loading state. It looks like it has something to do with an issue with the indexdb cache, I see this error before the group loading issue starts:

image

arthyn:
You might have to commit the backend changes to the moon? unless you did that already. I know we see those messages when we try to cache an error response (which we should probably handle).

latter-bolden (changes requested):
The JS changes look good to me. When stubbing the compatibility hooks, it works as expected. The only spot I found where I could still trigger updates was by modifying roles for members in the Admin Panel.

However, when I try to simulate an incompatibility by bumping the `okay` value on the group host, I'm unable to trigger the blocked inputs. RQ dev tools show the `saga` for the group as ahead, but when I log from the compatibility hooks, it shows up as synced. 

@arthyn were you able to successfully test this with a simulated mismatch during development?

jamesacklin:
@latter-bolden Feel free to add interactivity blockers to:
- drag action in All Channels
- edit capabilities for each channel and creating/deleting channels in All Channels

The "Leave" should stay; you should be able to leave regardless of protocol version.

latter-bolden (approved):

PR #2727, opened 2023-07-25
ops: fix ci

arthyn:
This adds a two minute delay so that our changes getting committed have enough time to compile. Also fixes various queries that the tests were having trouble running. Fixes LAND-628


latter-bolden (approved):
lgtm

PR #2729, opened 2023-07-26
groups: no-op on invalid sect adds 

arthyn:
This makes sure we ignore invalid sects attempting to be added to someone. Also makes sure we accept fleet adds from the host even if they aren't in our pending state and makes sure we save activity modal state.


jamesacklin:
We released this, so g2g?

arthyn:
yep

PR #2730, opened 2023-07-27
mobile: add DMs to groups mobile

patosullivan:
this does not include adding DMs to desktop, but it could easily be added. this does not include the new icon treatment, but it could if we want it to.

fixes LAND-711

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -333,6 +333,28 @@ function GroupsRoutes({ state, location, isMobile, isSmall }: RoutesProps) {
                 />
               }
             />
+            <Route path="/messages" element={<MobileMessagesSidebar />} />
+            <Route path="/dm/" element={<Dms />}>
+              <Route index element={<DMHome />} />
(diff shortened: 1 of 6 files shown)


jamesacklin:
@alecananian Can you point @patosullivan to the "isWebview" (or whatever) hook you mentioned yesterday? Might help feature-flag DMs-in-Groups in this case.

alecananian:
Sure, if something should show in the app only, you can use the `isNativeApp` util flag: https://github.com/tloncorp/landscape-apps/blob/develop/ui/src/logic/native.ts

patosullivan:
So we don't want this available if a user navigates to it in a mobile browser?

patosullivan:
> Correct - if it's an easy-enough hook to use (and remove), we can enable when we feel the time is right.

fixed, chef

jamesacklin:
Approved pending the merge conflict from the connectivity + saga stuff we merged today.

PR #2734, opened 2023-07-27
posthog: use persistent anonymous UUID

patosullivan:
This PR stores a persistent UUID in the ship's setting store for use with posthog (we generate the UUID client-side and then store it, we can't grab the user's existing posthog ID AFAICT).

It also gives the user the ability to manually reset the UUID via a button in the Settings modal.

The latter looks like this:

https://github.com/tloncorp/landscape-apps/assets/1221094/bb472c97-14b0-4c2c-9c1e-50529beb348b

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -118,6 +118,7 @@
         "urbit-ob": "^5.0.1",
         "use-pwa-install": "^1.0.1",
         "usehooks-ts": "^2.6.0",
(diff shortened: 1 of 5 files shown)


jamesacklin:
Excellent, thank you. I'll include in this immediate release.

cc @lukestiles

jamesacklin (approved):

PR #2735, opened 2023-07-27
diary: better saga tracking

arthyn:
This makes saga tracking more explicit and constant so we can know at all times if we're mismatched or in sync.

### DIFF
--- a/desk/app/diary.hoon
+++ b/desk/app/diary.hoon
@@ -99,11 +99,15 @@
     =.  cor
       (emil (drop load:epos))
     =/  diaries  ~(tap in ~(key by shelf))
+    =.  cor  
+      %+  roll
+        ~(tap in (~(gas in *(set ship)) (turn diaries head)))
+      |=  [=ship cr=_cor]
+      (watch-epic:cr ship &)
     |-
     ?~  diaries
       cor
-    =.  cor
-      di-abet:di-upgrade:(di-abed:di-core i.diaries)
+    =.  cor  di-abet:di-upgrade:(di-abed:di-core i.diaries)
     $(diaries t.diaries)
   ==
   ::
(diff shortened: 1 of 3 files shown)


midsum-salrux:
https://wicdev.pcmonk.me:9000/wicdev/wicdev-wisryt/2023.7.28..01.36.29-versions.md extra context for posterity

midsum-salrux (approved):
LGTM

desk/app/diary.hoon
```diff
@@ -99,11 +99,16 @@
     =.  cor
       (emil (drop load:epos))
     =/  diaries  ~(tap in ~(key by shelf))
+    =.  cor  
+      %+  roll
+        ~(tap in (~(gas in *(set ship)) (turn diaries head)))
+      |=  [=ship cr=_cor]
```
Fang-, line 105: No need to `gas:in` if you're gonna `tap:in` right after. Also, you can do away with the extra `turn` iteration for `head` by implicitly discarding the tail in the `roll`:

```suggestion
      %+  roll  diaries
      |=  [[=ship *] cr=_cor]
```

desk/app/diary.hoon
```diff
@@ -144,29 +149,26 @@
...
   ::
       %fact
     ?.  =(%epic p.cage.sign)
       ~&  '!!! weird fact on /epic'
       cor
     =+  !<(=epic:e q.cage.sign)
-    ?.  =(epic okay:d)
-      cor
-    ~&  >>  "good news everyone!"
+    ~&  >>  "resolving epic change, current: {<okay:d>} received: {<epic>}"
```
Fang-, line 171: Are we ok with printing this also in the case where current & received are identical? If we don't care for that, `~?` might be better.

Also, do we _really_ need the priority on this printf?

desk/app/diary.hoon
```diff
@@ -628,21 +630,31 @@
   ::
   ++  di-upgrade
     ^+  di-core
+    ::  we're the host so ignore
```
Fang-, line 633: +1 for adding comments to code/logic we want to make sure is exactly right. Rather risk being overly verbose, than not being clear enough.  
That said, could use a little paragraph for `+di-upgrade` as a whole, explaining what "ignoring" and "syncing" imply here.

Gentle recommendations: Official comment style is "flag" comments. And this might be made more legible by being more explicit about the branching.

```suggestion
    ::  if we're the host, no-op
    ::
```

desk/app/diary.hoon
```diff
@@ -144,29 +149,26 @@
   --
 ::
 ++  watch-epic
-  |=  her=ship
+  |=  [her=ship leave=?]
```
Fang-, line 152: If I understand this PR correctly: we may previously not have always subscribed to a publisher's epic/saga, but that is important, so now we make sure it's set up for everyone we subscribe to? Then, what is the `leave=&` case here for? Is it just so we can indiscriminately re-set-up these subscriptions (without getting "duplicate wire on sub" printouts) or do we actually care about getting the initial/current subscription result again?

Fang- (commented):
Some questions, some suggestions.

This is modifying `+on-load`'s logic for going from version `%0` to `%1`. Does this mean version `%1` isn't out in the wild yet? If it is, and we want the logic you added to always run, you may want to add a `%2` instead.

desk/app/diary.hoon
philipcmonk: I think `tap:in` after `gas:in` is to get rid of duplicate ships if you're subscribed to multiple things from a single ship.

desk/app/diary.hoon
arthyn: no we don't probably lol

desk/app/diary.hoon
arthyn: It's both. We need the initial data because we likely didn't have it and we want to tell people right away if they're mismatched, and I'm also trying to start cleaning up the dojo output.

desk/app/diary.hoon
arthyn: Yeah that's correct trying to avoid sending extraneous cards here that will thrash unsub/resub to the same person over and over. If there's a more concise way to do this I would love that lol.

Fang-: Oh duh, that makes sense. Good catch!

PR #2736, opened 2023-07-28
groups: iOS touch tweaks

jamesacklin:
- Removes the gray box on touch for iOS
- Removes the double-tap-to-zoom gesture to make taps to fire faster
- Adds focus color to the mobile tab bar items for immediate feedback
- Smooths the font rendering


patosullivan (approved):
lgtm!

PR #2737, opened 2023-07-28
notes: fix useArrangedNotes map of undefined crash

patosullivan:
We weren't handling the case where arranged notes were undefined.


arthyn (approved):
lgtm

PR #2738, opened 2023-07-28
GRO-600: Lure link checking

midsum-salrux:
https://linear.app/tlon/issue/GRO-600/lure-handle-pending-metadata-states-for-lure-link-creation

### DIFF
--- a/desk/ted/lure-check-link.hoon
+++ b/desk/ted/lure-check-link.hoon
@@ -1,0 +1,19 @@
+/-  spider
+/+  *strandio
+=,  strand=strand:spider
+=,  strand-fail=strand-fail:libstrand:spider
+^-  thread:spider
+|=  arg=vase
+=/  m  (strand ,vase)
+^-  form:m
+;<  our=@p   bind:m  get-our
+=/  arguments  !<((unit [cord ship cord path]) arg)
+=/  [baseurl=cord target=ship group=cord =path]  (need arguments)
+=/  url  (crip "{(trip baseurl)}{(trip (scot %p target))}/{(trip group)}")
+;<  ~  bind:m  (send-request %'GET' url ~ ~)
+;<  rep=client-response:iris  bind:m
+  take-client-response
+?>  ?=(%finished -.rep)
+=/  result  =(200 status-code.response-header.rep)
+;<  ~  bind:m  (poke [our %grouper] grouper-link-checked+!>([result path]))
+(pure:m !>(~))
--- a/desk/ted/lure-check-link.hoon
+++ b/desk/ted/lure-check-link.hoon
@@ -3,17 +3,31 @@
 =,  strand=strand:spider
 =,  strand-fail=strand-fail:libstrand:spider
 ^-  thread:spider
-|=  arg=vase
 =/  m  (strand ,vase)
-^-  form:m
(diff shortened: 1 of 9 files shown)


midsum-salrux:
Link check successful:
image


Link check not successful:
image

jamesacklin:
@dnbrwstr and @urcades, do you have opinions on this micro-interaction?

urcades:
A simple idea here, if I understand this problem correctly (ignore the greater context, focus on the form field itself)

Lure Link Generation

This recommendation basically splits up the lure link generation process into three discrete steps:

1. Disabled form, indicating that it must be enabled to show a link
2. If enabled, Lure Link is "generated" and indicates it needs time before it can display
3. Only when we've able to render a link that actually works, do we complete the "generation" process and show a working link that can be copied
4. If the generation process fails, or we can't reliably surface a working link, we should error out the form and provide people the opportunity to retry the generation process

desk/ted/lure-check-link.hoon
```diff
@@ -0,0 +1,33 @@
...
+=/  m  (strand ,vase)
+|^  ted
+++  ted
+  |=  arg=vase
+  ^-  form:m
+  ;<  our=@p   bind:m  get-our
+  =/  arguments  !<((unit [cord ship cord path]) arg)
+  =/  [baseurl=cord target=ship group=cord =path]  (need arguments)
+  =/  target-tape  (trip (scot %p target))
+  ?~  target-tape  !!
```
arthyn, line 15: I'm pretty unfamiliar with threads. Is it ok to just crash? Should we maybe give output everywhere that we crash?

desk/ted/lure-check-link.hoon
```diff
@@ -0,0 +1,33 @@
...
+    take-client-response
+  ?>  ?=(%finished -.rep)
+  =/  result  =(200 status-code.response-header.rep)
+  ;<  ~  bind:m  (poke [our %grouper] grouper-link-checked+!>([result path]))
+  (pure:m !>(~))
+++  url
+  |=  [baseurl=cord target=ship group=cord]
+  ^-  cord
+  ?.  =(baseurl 'https://tlon.network/lure/')
+    (crip "{(trip baseurl)}{(trip (scot %p target))}/{(trip group)}")
```
arthyn, line 27: should `baseurl` be checked for an ending slash? Also should we maybe move the target tape check to be first since both cases use the target?

arthyn (approved):
lgtm, a few questions though

desk/ted/lure-check-link.hoon
midsum-salrux: The input field for baseurl doesn't let you save it unless it has a trailing slash. You're right about moving the check up.

desk/ted/lure-check-link.hoon
midsum-salrux: Console output would be helpful here, I'll add that.

midsum-salrux: It'd be very surprising for this to actually get hit though. The null check is just to mollify the type system so we can remove the initial sig.

jamesacklin:
@arthyn Are we good to go? I want to cut a branch for these guys so we can get everything out next week.

PR #2739, opened 2023-07-28
groups: new mobile navbar

patosullivan:
fixes LAND-710

I also updated the action sheet to include a link to the profile editor and updated the profile editor to include a back button. I know we'll change this soon with the other design updates, this is just a stop-gap for now.

Looks like this:

https://github.com/tloncorp/landscape-apps/assets/1221094/939607e9-8a40-411c-93d0-10a055458b87

### DIFF
--- a/ui/src/profiles/EditProfile/EditProfile.tsx
+++ b/ui/src/profiles/EditProfile/EditProfile.tsx
@@ -15,6 +15,10 @@ import Avatar from '@/components/Avatar';
 import ShipName from '@/components/ShipName';
 import GroupSelector, { GroupOption } from '@/components/GroupSelector';
 import { useAnalyticsEvent } from '@/logic/useAnalyticsEvent';
+import { useIsMobile } from '@/logic/useMedia';
+import Layout from '@/components/Layout/Layout';
+import { Link } from 'react-router-dom';
+import CaretLeft16Icon from '@/components/icons/CaretLeft16Icon';
 import ProfileFields from './ProfileFields';
 import ProfileCoverImage from '../ProfileCoverImage';
 import ProfileGroup from './ProfileGroup';
@@ -134,7 +138,7 @@ function EditProfileContent() {
   const watchCover = form.watch('cover');
 
   return (
-    <div className="w-full">
+    <div className="w-full p-6">
       <FormProvider {...form}>
         <div>
           <ProfileCoverImage
@@ -216,15 +220,37 @@ function EditProfileContent() {
 
 export default function EditProfile({ title }: ViewProps) {
   useAnalyticsEvent('profile_edit');
+  const isMobile = useIsMobile();
 
   return (
     <div className="flex grow overflow-y-scroll bg-gray-50">
       <Helmet>
         <title>{title}</title>
       </Helmet>
-      <div className="w-full p-6">
+      <Layout
+        header={
+          isMobile ? (
+            <div className="flex items-center justify-between border-b-2 border-gray-50 bg-white py-2 pl-2 pr-4">
+              <Link
+                to=".."
+                className="default-focus ellipsis w-max-sm inline-flex h-10 appearance-none items-center justify-center space-x-2 rounded p-2"
+              >
+                <div className="flex h-6 w-6 items-center justify-center">
+                  <CaretLeft16Icon className="h-5 w-5 shrink-0 text-gray-600" />
+                </div>
+                <div className="flex w-full flex-col justify-center">
+                  <span className="ellipsis text-sm font-bold line-clamp-1 sm:font-semibold">
+                    Edit Profile
+                  </span>
+                </div>
+              </Link>
+            </div>
+          ) : null
+        }
+        className="flex-1"
+      >
         <EditProfileContent />
-      </div>
+      </Layout>
     </div>
   );
 }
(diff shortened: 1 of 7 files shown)


urcades (approved):
The bones are here, and they're looking really good.

I support merging this + performing the deeper visual fixes to the profile page + the resulting card itself, which may end up encompass changes to the way we depict all profile cards, app-wide.

jamesacklin:
Merge at will.

ui/src/components/icons/PlaceholderIcon.tsx
```diff
@@ -0,0 +1,23 @@
+import { IconProps } from './icon';
+
+export default function PlaceholderIcon({ className }: IconProps) {
```
latter-bolden, line 3: is this just for development use?

ui/src/profiles/EditProfile/EditProfile.tsx
```diff
@@ -214,17 +218,58 @@ function EditProfileContent() {
...
       <Helmet>
         <title>{title}</title>
       </Helmet>
-      <div className="w-full p-6">
-        <EditProfileContent />
-      </div>
+      <Layout
+        header={
+          isMobile ? (
+            <MobileHeader title="Edit Profile" pathBack="/profile" />
```
latter-bolden, line 254: this leaves me stranded if clicked from Talk

ui/src/components/Settings/Settings.tsx
```diff
@@ -0,0 +1,188 @@
...
+          <p className="leading-5 text-gray-600">
+            Turn off automatically-displaying media embeds across{' '}
+            {isTalk ? 'Talk' : 'Groups'}. This may result in some software
+            appearing to have content missing.
+          </p>
+        </Setting>
+      </div>
+      <div className="space-y-2">
+        <div className="flex flex-col">
+          <h2 className="mb-2 text-lg font-bold">Theme</h2>
```
latter-bolden, line 163: on mobile I can see the Theme section header, but the control is unreachable

ui/src/components/ButterBar.tsx
latter-bolden, line 1: looks like this component is never referenced, should we remove instead of updating?

latter-bolden (changes requested):
Nice work, there's a lot packed in here. Code is looking good, just a few small issues/questions.

On a meta note, it feels like we're using a lot of ad hoc pixel values to match the Figma mocks. Wondering if this will start feeling unwieldy as we get deeper into the mobile updates.

patosullivan:
> Nice work, there's a lot packed in here. Code is looking good, just a few small issues/questions.
> 
> On a meta note, it feels like we're using a lot of ad hoc pixel values to match the Figma mocks. Wondering if this will start feeling unwieldy as we get deeper into the mobile updates.

Yeah, I'd prefer that we create new classes or update our existing size system, but for now these need to be fluid as Ed and Dan figure out which sizes actually work well on their new mobile UX.

ui/src/profiles/EditProfile/EditProfile.tsx
patosullivan: How did you get to this view on mobile talk?

latter-bolden: Grid icon in the header → Profile

patosullivan: Ah, via leap, got it.

latter-bolden (approved):

ui/src/components/MobileHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <div className="grid w-full grid-cols-3 items-center justify-between bg-white py-3 pl-2 pr-4 font-system-sans">
+      {pathBack ? (
+        <Link className="flex items-center" to={pathBack}>
+          <CaretLeftIcon className="h-6 w-6 text-gray-400" />
+          {pathBackText && (
+            <span className="text-[17px] leading-6 text-gray-800 line-clamp-1">
+              Cancel
+            </span>
+          )}
```
arthyn, line 26: Maybe we could just conditionally apply `sr-only` here if they don't want the text that way it always has an accessible label?

arthyn (approved):
lgtm

jamesacklin:
A few comments.

In the group view, it appears we're lighting up every navbar item as active:
image

Moreover, we have labels on those navbar items -- is that intentional @urcades ?

Navigating to the profile view pushes the navbar off-screen (was unable to scroll down on mobile, had to confirm on desktop):
Screenshot 2023-08-08 at 5 01 55 PM

Same thing with Find Groups: 
image

Missing some bottom padding on the new full-screen Settings view:
image

---

I also noticed a few things that I can help to tune up in a separate issue.

Noticeable jumps in icon position between the All Groups view and the group view itself:

https://github.com/tloncorp/landscape-apps/assets/748181/3b8cedd6-d5d1-4daa-af94-36a6878a63d1

Noticeable jumps in title and connectivity icon position between a group view and a channel:

https://github.com/tloncorp/landscape-apps/assets/748181/18d27c4e-aa0a-4452-9752-d1ab58307bdb

Icon-buttons are misaligned along the horizontal (window for ruler):
image

jamesacklin:
@patosullivan I'll do a cleanup pass, feel free to merge.

PR #2741, opened 2023-07-31
gallery: fix refs in gallery comments

patosullivan:
Styles the refs so that they can be displayed sanely in heap comments.

Fixes LAND-718.

Looks like this on desktop:

image

Looks like this on mobile:

https://github.com/tloncorp/landscape-apps/assets/1221094/399f7ec8-6357-4d1a-b565-3aa322a4e306


latter-bolden (approved):
Code looks good and worked for all reference types locally.

PR #2743, opened 2023-07-31
notebook: prevent create/editing on saga mismatch

jamesacklin:
Hides Add, Edit, Pin, Delete, and Comment capabilities for notebooks where you are not in sync with the host's version.

Also fixes the ShipConnection component to show the tooltip correctly.

### DIFF
--- a/ui/src/components/ShipConnection.tsx
+++ b/ui/src/components/ShipConnection.tsx
@@ -16,6 +16,18 @@ interface ShipConnectionProps {
   className?: string;
 }
 
+export function getText(ship: string, status?: ConnectionStatus) {
+  if (ship === window.our) {
+    return 'This is you';
+  }
+
+  return !status
+    ? 'No connection data'
+    : 'pending' in status
+    ? getPendingText(status, ship)
+    : getCompletedText(status, ship);
+}
+
 export default function ShipConnection({
   status,
   ship,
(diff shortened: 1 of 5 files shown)


ui/src/components/ShipConnection.tsx
```diff
@@ -24,28 +36,28 @@ export default function ShipConnection({
...
-        <Tooltip content={text}>
-          <Bullet16Icon
-            className={cn('h-4 w-4 flex-none', color)}
-            tabIndex={0}
-          />
+        <Tooltip content={getText(ship, status)}>
+          <span tabIndex={0}>
+            <Bullet16Icon
+              className={cn('h-4 w-4 flex-none', color)}
+              tabIndex={0}
```
arthyn, line 52:
```suggestion
```

ui/src/components/ShipConnection.tsx
```diff
@@ -16,6 +16,18 @@ interface ShipConnectionProps {
...
+    return 'This is you';
+  }
+
+  return !status
+    ? 'No connection data'
+    : 'pending' in status
+    ? getPendingText(status, ship)
+    : getCompletedText(status, ship);
+}
+
```
arthyn, line 30: you shouldn't need to do this, the previous should work

arthyn (changes requested):
two small changes, rest lgtm

arthyn (approved):
lgtm

PR #2750, opened 2023-08-02
lure: autojoin from any view

tomholford:
# Context

Before this change, the auto join logic was only executed on the Home Route. Unfortunately, we immediately drop new users into the Find Groups view, so this effect was not being executed.

This along with https://github.com/tloncorp/ylem/pull/1597 resolves GRO-598

# Changes 
This PR extracts the logic to a dummy component that's nested under the root router (since it depends on the router context). Thus this will now work for wherever a newly onboarded user first opens Groups.

# Testing

I tested this flow successfully with the following steps:

1. Onboard a new account using https://test.tlon.systems/lure/~nibset-napwyn/tlon?token=1-800-ZOD
2. Once the flow is finished, the ship is booted, and Landscape has been launched, use a local instance of Groups to connect to the hosted ship:
```
SHIP_URL=https://wanlud-simler.test.tlon.systems/ npm run dev
```
3. Observe that after some time elapses while loading the group invites, the hook fires and auto-joins the Tlon Local group


patosullivan (approved):
lgtm!

PR #2757, opened 2023-08-04
bugfix: land-736

patosullivan:
Apparently using destructuring to pull values out of this ref could be interpreted as attempting to assign a value to the ref. Eliminated the destructuring and added a check to make sure we have a ref.


arthyn (approved):
lgtm

PR #2760, opened 2023-08-05
groups: revert breaking api change, introduce new endpoint for new type

arthyn:
For the connectivity and compatibility checks we added a new type which includes the `saga` so that we could know when our agents are out of sync. Unfortunately this broke compat for anyone using that scry, including our %bark agent. This reverts that change so that the new types are at a new scry endpoint `/groups/ui` and `/groups/light/ui` respectively, and transitions the UI to use that instead.  

Fixes GRO-612

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -338,14 +338,15 @@
   ^-  (unit (unit cage))
   ?+    pole  [~ ~]
       [%x %gangs ~]  ``gangs+!>(xeno)
-      [%x %init ~]  ``noun+!>([groups-light-ui xeno])
+      [%x %init ~]  ``noun+!>([groups-light xeno])
+      [%x %init %v0 ~]  ``noun+!>([groups-light-ui xeno])
       [%x %groups %light ~]  ``groups+!>(groups-light)
-      [%x %groups %light %ui ~]  ``groups-ui+!>(groups-light-ui)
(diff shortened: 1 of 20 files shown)


jamesacklin:
Cool, thanks. Is this going out to mainnet tonight/tomorrow?

desk/app/groups.hoon
```diff
@@ -338,11 +338,15 @@
   ^-  (unit (unit cage))
   ?+    pole  [~ ~]
       [%x %gangs ~]  ``gangs+!>(xeno)
-      [%x %init ~]  ``noun+!>([groups-light xeno])
+      [%x %init ~]  ``noun+!>([groups-light-ui xeno])
```
Fang-, line 341: Looks like this has produced a `[groups-light xeno]` for some time (at least since before the breaking change to `/groups`). Why are we changing this here/now? Won't that potentially run into the same api breakage?

arthyn: while this was always groups-light, previously that type was `(map flag group)` which changed to `(map flag group-ui)` with the api change. this needs to stay changed so it's called something new to reflect that change. this scry is what we use to initialize groups through the %groups-ui agent, and that uses the `group-ui` type.

Fang- (approved):
This seems fine for now.

PR #2761, opened 2023-08-05
chat: use vercel's react-tweet to improve twitter embeds

patosullivan:
No more opening a modal to view the full tweet (since our old embed would tend to garble them if they included mentions, pictures, or videos).

Requires react-18, so this is currently based on hunter's branch.

Looks like this:

image
image
image
image
image

Fixes LAND-741


arthyn (approved):
lgtm

PR #2764, opened 2023-08-07
4.2.1 -> august-growth

jamesacklin:
Brings the bugfixes for the right-hand error in Notebooks, a guard against DMing yourself, and the v0 backcompat changes for 4.2.1 into the `august-growth` branch.


arthyn (approved):
lgtm

PR #2765, opened 2023-08-07
groups: don't ask for tracking on first run

patosullivan:
Wait until user has opened groups a second time.

Separately: there is something wrong with the %settings initialization. showActivityMessage is not getting initialized to `true`, so I've hacked around that for now.

### DIFF
--- a/ui/src/state/settings.ts
+++ b/ui/src/state/settings.ts
@@ -504,14 +504,24 @@ export function useSideBarSortMode() {
 
 export function useShowActivityMessage() {
   const { data, isLoading } = useMergedSettings();
+  const { mutate } = usePutEntryMutation({
+    bucket: 'groups',
+    key: 'hasBeenUsed',
+  });
 
   return useMemo(() => {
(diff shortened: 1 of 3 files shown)


ui/src/state/settings.ts
```diff
@@ -504,14 +504,24 @@ export function useSideBarSortMode() {
...
+    key: 'hasBeenUsed',
+  });
 
   return useMemo(() => {
     if (isLoading || data === undefined || window.desk !== 'groups') {
       return false;
     }
 
+    if (!data.groups?.hasBeenUsed && data.groups?.showActivityMessage) {
+      mutate({ val: true });
```
arthyn, line 518: I don't think we should mutate inside this hook.

arthyn (changes requested):
in addition to the comment, I think with how this is written we'd need something different because this would trigger true as soon as any second settings change would happen.

arthyn (approved):
lgtm

jamesacklin:
Looks good.

PR #2768, opened 2023-08-08
bugfix: drag and drop could get stuck in "drag" mode

patosullivan:
I originally had an unnecessary early return before we get to the setIsDragging(false) and setIsOver(false) calls.

This removes that.


jamesacklin (approved):
We will need this in the `august-growth` branch for release tomorrow.

PR #2771, opened 2023-08-08
Fix position of Talk/Groups containers

latter-bolden:
Fixes LAND-702

Small update that fixes the position of the container components that wrap Talk and Groups. Apart from hopefully resolving the scroll issue referenced, this should help the mobile app feel more native by ensuring navigational elements can't be dragged out of position. 

Give it a spin and let me know what y'all think. If we want to preserve the existing feel on desktop, I can add a mobile check.


jamesacklin:
I can't tell the difference at all on desktop, maybe it's too subtle for my eye. Seems good!

PR #2774, opened 2023-08-08
ops: fix CI by replacing docker/bash scripts for playwright with node scripts

patosullivan:
Fixes LAND-628


patosullivan:
fixes LAND-628

PR #2775, opened 2023-08-08
GRO-600: Lure link check UX followup changes

midsum-salrux:
image

UX changes for lure link checking.

- The link is checked every 5 seconds
- The retry button is disabled for the first 6 seconds
- The retry button toggles lure off and then back on again
- I changed the copy while the link is unverified

@zalberico @urcades


arthyn (approved):
lgtm

PR #2777, opened 2023-08-09
lure: more link checking status

arthyn:
This refactors the lure checker so that it uses react query which gives us a little more info into what's going on to present more status to the user, and brings the UI more inline with the rest.

Sidenote we have a lot flags here that combine into one state that should probably be combined somehow.

### DIFF
--- a/ui/src/groups/LureInviteBlock.tsx
+++ b/ui/src/groups/LureInviteBlock.tsx
@@ -5,6 +5,7 @@ import CheckIcon from '@/components/icons/CheckIcon';
 import { Group } from '@/types/groups';
 import LoadingSpinner from '@/components/LoadingSpinner/LoadingSpinner';
 import ShipName from '@/components/ShipName';
+import { useCallback } from 'react';
 
 interface LureInviteBlock {
   flag: string;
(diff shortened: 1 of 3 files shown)


ui/src/groups/LureInviteBlock.tsx
```diff
@@ -96,35 +103,27 @@ export default function LureInviteBlock({
...
-                disabled={retryDisabled}
-                className="button"
-                onClick={() => {
-                  toggle(group?.meta || emptyMeta)();
-                  setTimeout(toggle(group?.meta || emptyMeta), 500);
-                }}
-              >
+              <div className="flex w-full flex-1 items-center space-x-2 rounded-lg border-2 border-blue-soft bg-blue-soft py-1 px-2 text-lg font-semibold leading-5 text-blue mix-blend-multiply dark:mix-blend-screen sm:text-base sm:leading-5">
+                <span className="font-semibold text-blue">
+                  Failed to successfully create link with lure service
```
jamesacklin, line 118: I think we can go with "Error creating link" here. "Lure" makes sense to us for internal terminology but we don't explain it anywhere (nor should we IMO)

jamesacklin (approved):
Code looks good, left a comment inline on copy

ui/src/groups/LureInviteBlock.tsx
arthyn:
```suggestion
                  Error creating link
```

PR #2779, opened 2023-08-09
Add Cosmos

dnbrwstr:
This PR adds Cosmos, an example component, and some utilities and data for mocking out chunks of functionality. I added some basic docs + description to the readme, which I'm including below. 

A question: Cosmos has two conventions for fixture files: subfolders called `__fixtures__`, or inidividual files named `<whatever>.fixture.ts`. This PR uses the folder, but I'm curious if anyone has a different preference.

----

## Cosmos

Run cosmos: 
```
npm run cosmos
```

### Fixtures

[Fixtures – Cosmos Docs](https://github.com/react-cosmos/react-cosmos/blob/main/docs/usage/fixtures.md)  

Cosmos uses fixture files to generate component previews. You can create fixture files in two ways (from the docs).
1. End fixture file names with `.fixture.{js,jsx,ts,tsx,md,mdx}.`
2. Put fixture files inside   `__fixtures__`.

Fixture files and folders can be nested at any level under `/src`.

### Decorators

[Decorators – Cosmos Docs](https://github.com/react-cosmos/react-cosmos/blob/main/docs/usage/decorators.md)

Many components need to be wrapped with providers or other context to work properly. Cosmos uses a `cosmos.decorator.tsx` files to share wrappers between components. From the Cosmos docs:

> A decorator only applies to fixture files contained in the decorator's directory. Decorators can be composed, in the order of their position in the file system hierarchy (from outer to inner).

### How it works

Cosmos runs two servers: a UI wrapper on port `5000`, and a renderer on port `5050`. The wrapper renders the Cosmos UI, and the renderer renders the actual component previews in an environment that's very close to our standard application environment.

Our setup uses `react-cosmos-plugin-vite` to configure the renderer. When you start cosmos, it evaluates our vite configuration file, does some module rewiring to replace our `main.ts` file with its own, and spins up a wrapped vite dev server on a new port.

[Cosmos Docs](https://github.com/react-cosmos/react-cosmos/blob/main/docs/README.md#getting-started)

Fixes LAND-732


ui/package.json
```diff
@@ -44,7 +44,9 @@
     "postinstall": "patch-package",
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
-    "e2e:codegen": "concurrently \"npm run dev-no-ssl\" \"npx playwright codegen http://localhost:3000\""
+    "e2e:codegen": "concurrently \"npm run dev-no-ssl\" \"npx playwright codegen http://localhost:3000\"",
+    "cosmos": "cosmos",
+    "cosmos:export": "cosmos-export"
```
latter-bolden, line 49: Not sure how crucial this functionality is, but running the exported bundle didn't work for me. We should also `.gitignore` that generated directory if we're gonna keep this in

latter-bolden (commented):
Maybe worth adding a variant with a profile picture for Avatar. I'd also favor ripping out the legacy storybook stuffs before merging.

But this is looking really good. If there are other components you were considering moving in right away, I think it's ready for that.

arthyn:
would second ripping out storybook

arthyn (approved):
lgtm this is so much nicer! I think we can worry about more avatar options later. we should probably just meet about how we can best split components to make them dumber and we can address it then.

ui/package.json
dnbrwstr: Yeah, makes sense. I've removed for now, we can add back in and debug when we actually want to deploy.

dnbrwstr:
Thanks y'all -- ripped out Storybook and removed the export command. Should be good to merge in now.

PR #2780, opened 2023-08-09
groups: add notification volume management

Fang-:
Groups now keep track of notification settings. A global setting, which may be overridden by per-group settings, which may be overridden by per-channel settings.

Users may choose between three notification levels: %loud, %soft and %hush, meaning "everything", "some things", and "nothing" respectively.

It remains up to individual agents to determine what falls into the "soft" bucket. We do not propose a unified theory of notifications.

/lib/volume provides utilities for reading out notification levels for any given "scope".

Includes example integration with existing /app/chat notifications.

For GRO-614. Remaining work, which I personally won't get to before reassembly, but happy to help someone else take over:
- [ ] Test properly.
- [ ] Integrate with other agents. Add more notifications to chat (and mb others).
- [ ] Integrate with frontend.


desk/mar/volume/set.hoon
```diff
@@ -0,0 +1,30 @@
...
+  |%
+  ++  noun  ,[=^scope =^value]
+  ++  json
+    =,  dejs:format
+    |^  (ot 'scope'^scope 'value'^value ~)
+    ++  scope
+      |=  jon=^^json
+      ^-  ^^scope
+      ?~  jon  ~
+      ((of %group^flag:dejs:gjs %channel^nest:dejs:gjs ~) jon)
```
arthyn, line 20: this is kinda hard to parse

desk/app/groups.hoon
```diff
@@ -179,6 +187,26 @@
       %invite-decline
     =+  !<(=flag:g vase)
     ga-abet:ga-invite-reject:(ga-abed:gang-core flag)
+  ::
+      %volume-set
```
arthyn, line 191: should only be from self

desk/app/groups.hoon
```diff
@@ -361,6 +403,34 @@
...
+      [%x %volume dude=@ ship=@ name=@ ~]
+    =/  =ship    (slav %p ship.pole)
+    =/  =nest:g  [dude.pole ship name.pole]
+    :^  ~  ~  %volume-value
+    !>  ^-  value:v
+    ?^  vol=(~(get by chan.volume) nest)
+      u.vol
+    ::NOTE  searching through all groups like this is... inefficient,
+    ::      but the alternative is depending on the dude knowing what
+    ::      group the nest belongs to and scrying that out of it...
```
arthyn, line 426: yeah, we should keep a backwards map index to reference these easier

arthyn (changes requested):
some small changes and comments, but lgtm otherwise

PR #2788, opened 2023-08-11
Gallery: Update block input & other fun fixes

latter-bolden:
Very WIP, but basic structure is here. Still need to:

- [x] handle mobile input
- [x] handle drag drop/paste
- [x] rich text?

Will eventually...
Fixes LAND-614
Fixes LAND-401


latter-bolden:
This work was moved to another branch.

PR #2791, opened 2023-08-14
Groups: fix channel manager width & loading placeholder overflow

latter-bolden:
Fixes LAND-769

also resolves the issue where the gray chat messages loading animation overflows into the input.


jamesacklin (approved):
Thanks for catching my debug `bg-red`.

PR #2793, opened 2023-08-14
Cache embeds

latter-bolden:
Updates our embed lookups to run through `react-query` which causes them to get cached using IndexedDB.

Fixes LAND-438


jamesacklin (approved):
I see the XHR requests going out when scrolling through an embed-heavy channel, but they're being caught by the cache. Nice work!

PR #2800, opened 2023-08-16
groups: all group joins must be via join modal

patosullivan:
fixes LAND-647 to force all group joins to go through the group join modal flow. Also does not redirect the user to the joined group until the data for the group has already been loaded.


arthyn (approved):
lgtm

PR #2803, opened 2023-08-16
tools: cleanup eyrie usage

arthyn:
This pulls eyrie directly into our app so we can control it better and not incur the cost of a react + deps bundle.


jamesacklin:
Makes sense to bundle. It's ours, we can count it as a dev feature.

patosullivan (approved):
lgtm!

PR #2804, opened 2023-08-17
chat: disable user-select on messages for mobile

jamesacklin:
Kills text selection in Chat messages on mobile, which interferes with the long-tap action menu.

Note that this doesn't disable long-tap gestures on things like images, links, or other elements (so you can still long-tap to preview a link on iOS, long-tap to save an image, etc.).


dantish (approved):
Man, it’s an awesome and elegant solution, I’ll definitely need to learn css/classnames more😅

PR #2805, opened 2023-08-18
groups: improve "Share with Friends" action on mobile

jamesacklin:
Adds a QR code and a system share action (with clipboard copy fallback) to the "Share with Friends" action in the mobile profile menu.

This view has become slightly overwrought and we likely want to do the same for all groups, so I'm leaving this as a draft.

https://github.com/tloncorp/landscape-apps/assets/748181/7b6fcb26-6c9b-4a96-816e-58b1da8f99c6


jamesacklin:
Closing the PR but keeping the branch.

PR #2807, opened 2023-08-20
Chat: Add video player embed

patosullivan:
Fixes LAND-620. Shows a thumbnail, renders/plays-inline. No lightbox.

Looks like this:

https://github.com/tloncorp/landscape-apps/assets/1221094/56432d05-defa-48f2-847c-c63b7eaacfea

We might want to limit the height/width? It can actually be customized quite a bit, too, if we want to do that:
https://video-react.js.org/customize/customize-component/


arthyn (approved):
lgtm

latter-bolden:
Confirmed it only buffers a few MB at a time. Tested with a feature length video and scrubbing around worked as expected without triggering a download of the entire file.

Will say if we're going to avoid investing the design/eng resources in a custom player theme right now, I prefer the look and feel of the native video tag (which we're using for Gallery embeds right now). If we go that route the controls for full-screen, picture in picture, and playback speed all work out of the box.

PR #2809, opened 2023-08-21
groups: slight nav fixes

jamesacklin:
- Makes the center area of the MobileHeader more tappable (and simpler)
- Shrinks the channel header down to one line
- Uses GroupActions in the mobile header of a group and removes the "..." menu actions (fixes LAND-788)
- Moves the connection check explanations for groups and channels into GroupActions and ChannelActions (fixes LAND-783)
- Foregoes the MobileGroupSidebar for the time being and uses a single MobileSidebar (fixes LAND-791)
- Fixes an issue where the Profile card would inconsistently apply x-margins
- Adds Members and Invite actions to the root Group menu, eliminating any need for more menus (fixes LAND-419)
- Adds some other tweaks from @arthyn


dantish (approved):
Nice one 👍

PR #2819, opened 2023-08-23
chat: fix emoji reactions not sticking

patosullivan:
When adding or deleting a feel `her` is the writ author rather than the emoji feeler. Since that's the case, we were returning %.n early and the writ is never updated. This removes the early return, and instead adds the check to the appropriate branches in the `?-`.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -887,7 +887,7 @@
   =*  should  =(her src.bowl)
   ?-  -.delta
       %add  ?.(should | =(src.bowl author.p.delta))
-      %del  ?.(should | &)
+      %del  should
       %add-feel  =(src.bowl p.delta)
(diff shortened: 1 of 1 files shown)


desk/app/chat.hoon
```diff
@@ -884,13 +884,13 @@
   |=  diff=diff:writs:c
   =*  her    p.p.diff
   =*  delta  q.diff
-  ?.  =(her src.bowl)  |
+  =*  should  =(her src.bowl)
   ?-  -.delta
-      %add  =(src.bowl author.p.delta)
-      %del  &
+      %add  ?.(should | =(src.bowl author.p.delta))
+      %del  ?.(should | &)
```
arthyn, line 890:
```suggestion
      %del  should
```

arthyn (approved):
lgtm, suggestion for clarity

PR #2821, opened 2023-08-23
chat: fix various clicking issues

patosullivan:
Fixes both LAND-801 and LAND-802, in addition to switching to a normal long press for long presses (rather than mouseUp).

Also cleans up a few errant console.logs().

Drag and drop still works normally on desktop.

Fixes LAND-801
Fixes LAND-802


jamesacklin:
Can you see if this fixes LAND-809 as well? (in triage)

patosullivan:
It won't, I think that's an issue with the dialog component.

latter-bolden (approved):
Looks and feels good to me

PR #2822, opened 2023-08-24
chat: fix video player size bug

patosullivan:
Needed to pass an explicit `fluid={false}` prop and set a width on the video player itself.

Fixes LAND-810

Test it by posting a tall video followed by another message.


jamesacklin:
Does this work on window sizes <600px?

PR #2823, opened 2023-08-24
mobile: dms header

arthyn:
We were missing the titles for these so I added them in. Was also considering updating the alignment like @urcades latest designs, but figured that might be too much to chew. Did throw in hiding chat formatting bar.


patosullivan (approved):
lgtm!

PR #2830, opened 2023-08-27
groups: react-query on-mount refresh not scrying for settings

dantish:
Fixes LAND-750

The fix is to set `staleTime` _(20 mins)_ to prevent infinite loop of refetching/re-rendering

Please, feel free to say your arguments against it or the potential drawbacks of this approach.

### DIFF
--- a/ui/src/logic/useReactQuerySubscription.tsx
+++ b/ui/src/logic/useReactQuerySubscription.tsx
@@ -55,7 +55,7 @@ export default function useReactQuerySubscription({
   }, [app, path, queryClient, queryKey]);
 
   return useQuery(queryKey, fetchData, {
-    staleTime: 20 * 60 * 1000,
+    staleTime: 60 * 1000,
     ...options,
   });
 }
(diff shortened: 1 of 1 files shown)


ui/src/logic/useReactQuerySubscription.tsx
```diff
@@ -55,8 +55,7 @@ export default function useReactQuerySubscription({
   }, [app, path, queryClient, queryKey]);
 
   return useQuery(queryKey, fetchData, {
-    retryOnMount: false,
-    refetchOnMount: false,
+    staleTime: 20 * 60 * 1000,
```
arthyn, line 58: I think we probably want something much shorter. Not sure what the right default is though. Worst case scenario you're behind by a minute, but we may need to override in certain places. Let's try this for now!

```suggestion
    staleTime: 60 * 1000,
```

arthyn (approved):
lgtm, with suggestion. we still probably want to figure out the cyclic dependency causing this, but at least this probably gets us closer to where we should be.

PR #2835, opened 2023-08-28
Gallery: Speed up group refs

latter-bolden:
Updates our gang preview to check for existing data before issuing a new fetch. 

Believe this is what was slowing down refs in the "Local Directory" channel, but let me know if you guys are still seeing lag there.

Fixes LAND-794


arthyn (commented):
Might be good to just handle this through react query https://tanstack.com/query/v4/docs/react/guides/initial-query-data

arthyn (approved):
lgtm

PR #2836, opened 2023-08-28
chat: restore formatter bar in mobile on highlight

arthyn:
OTT never really meant to take away completely. This seems to be the best compromise imo, but open to hear what else we could do.


latter-bolden:
Code looks good to me, but we should talk during standup. I know you mentioned pop up style formatting display clashing with native selection, but for me this was also colliding pretty intrusively.

PR #2838, opened 2023-08-29
notebooks: add markdown checkboxes

patosullivan:
Fixes LAND-756.

Looks like this:

https://github.com/tloncorp/landscape-apps/assets/1221094/18acd3e2-663a-4ff2-be3e-9cc8ef542c41

### DIFF
--- a/ui/src/diary/DiaryContent/DiaryContent.tsx
+++ b/ui/src/diary/DiaryContent/DiaryContent.tsx
@@ -165,6 +165,22 @@ export function InlineContent({ story }: InlineContentProps) {
     return <span className="text-blue">{story.ship}</span>;
   }
 
+  if ('task' in story) {
+    return (
+      <div className="flex items-center">
+        <input
+          type="checkbox"
+          checked={story.task.checked}
+          className="pointer-events-none mr-2"
+          aria-disabled
+        />
+        {story.task.content.map((con, i) => (
+          <InlineContent key={`${con}-${i}`} story={con} />
+        ))}
+      </div>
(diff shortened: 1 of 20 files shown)


latter-bolden:
Ran into some funk while testing:

- Inserting a Task List seems to remove other content in the note
- Task Lists don't seem to persist when saved (not seeing them in the view or edit screens for the saved note)


https://github.com/tloncorp/landscape-apps/assets/90741358/0af7901d-47c3-4ba2-8486-554c0039a1a6

patosullivan:
Yeah, realized this would be an issue right as I was falling asleep last night, ha.

patosullivan:
Updated so that we can persist tasklists/taskitems to the backend and render them in notes. Not sure if we need a state transition here or not.

ui/src/diary/DiaryContent/DiaryContent.tsx
```diff
@@ -165,6 +165,22 @@ export function InlineContent({ story }: InlineContentProps) {
...
   }
 
+  if ('task' in story) {
+    return (
+      <div className="flex items-center">
+        <input
+          type="checkbox"
+          checked={story.task.checked}
+          className="pointer-events-none mr-2"
+          aria-disabled
```
arthyn, line 175: maybe also add an aria label of checked/unchecked?

arthyn (changes requested):
lgtm, won't need a state transition I think, but will need a version update.

arthyn (approved):
lgtm

PR #2839, opened 2023-08-29
Chat: Fix DM reacts on mobile

latter-bolden:
OTT

We just needed to move the picker paths for DMs over to the Groups router now that chat is embedded there for mobile.

Fixes LAND-835


patosullivan (approved):
lgtm

PR #2840, opened 2023-08-30
gallery: Use existing curios as initial data in HeapDetail

latter-bolden:
Updates the `HeapDetail` view to use existing data from the main grid query as an initial value while fetching the full curio with comments in the background. This should make sure it's always snappy when clicking into a block from the channel view.

Fixes LAND-841


patosullivan (approved):
lgtm

PR #2841, opened 2023-08-30
groups: fix infinite loop on mark read in diaries/heaps

patosullivan:
markRead was getting fired in an infinite loop via the useEffect in useDismissChannelNotifications. We need to prevent it from firing while the initial call is still loading and we need to invalidate the cache on success to keep it from firing again after the first call completes.

Also, mutateAsync was unnecessary here, mutate is fine.

Fixes LAND-842


arthyn (approved):
lgtm

PR #2843, opened 2023-08-31
refs: don't show roles in bar

arthyn:
They get crowded too easily imo and you can always click the person's profile.


patosullivan (approved):
Yeah, needs to happen.

PR #2846, opened 2023-09-01
groups: eliminate scrollbars in invite dialog on desktop

mrozanski:
Part of GRO-615

Review notes:

* Scrollbar still appears at small resolutions near the mobile breakpoint but gone at most window sizes in between.
* I don't know this app enough to tell if removing that margin-bottom class has other side effects.
* The main issue of the flickering is still in progress. More in comment later.

no-scrollbar

### DIFF
--- a/ui/src/groups/GroupInviteDialog.tsx
+++ b/ui/src/groups/GroupInviteDialog.tsx
@@ -168,7 +168,7 @@ export default function GroupInviteDialog() {
       open={true}
       onOpenChange={(isOpen) => !isOpen && dismiss()}
       containerClass="w-full max-w-xl"
-      className="mb-64 bg-transparent p-0"
+      className="bg-transparent p-0"
       close="none"
     >
       {renderContent()}


mrozanski:
https://github.com/tloncorp/landscape-apps/blob/3a5883e7ac0a6baf112566a60dc3571a734f3032/ui/src/groups/LureInviteBlock.tsx#L104
GRO-615

regarding the flickering issue:

If I'm following correctly,

1. After checking the box, `enabled` is coming back `true` from `useLure(flag)`
2. Then, it comes back false momentarily
3. This causes a few re-renders, changing the returned component visibly based on the condition linked above

`{enabled && enableAcked ? (...`

4. Then the hook returns `enabled=true` again with the opposite effect.

@arthyn we may need to talk to figure out the best solution, so I'll pause here.
It could be addressed from the component with a debouncer function that waits for the next change if the value changes rapidly.

Or it could be addressed from the `useLure` hook itself.

Maybe something else is not working as expected and these `true-false` flips should not happen in the first place.

Screenshot 2023-09-01 at 5 04 32 PM

patosullivan (approved):

mrozanski:
I've tested this and the flickering is avoided if we delay setting the value of the `enabled` flag that is used for conditional rendering.
I can't commit my POC because it violates a couple of linter rules but this should illustrate the point:

```
  // Use enalbedLocal in the return block for conditions instead of enabled

  const [enabledLocal, setEnabledLocal] = useState(enabled);
  const enabledTimeout = useRef(setTimeout(()=>{}, 0));

  useEffect(() => {
    clearTimeout(enabledTimeout.current);
    enabledTimeout.current = setTimeout(() => {
      setEnabledLocal(enabled);
    }, 700);
  }, [enabled]);
```

This is not a good solution because it delays the response from the UI and feels laggy.

There should be a way to avoid `useLure()` from returning the intermediate `false` values.

mrozanski:
The remaining scrollbars that still appear at viewport heights < 575px and the extra margin at the top that makes the cards look slightly off to the bottom go away bu removing the `mt-10` from this line in GroupInviteDialog

```
<div className={cn('flex flex-col space-y-6', !isMobile && 'mt-10')}>
```

https://github.com/tloncorp/landscape-apps/blob/b5ac597c324d92de6b8f7185de3453ace0a180e9/ui/src/groups/GroupInviteDialog.tsx#L150C76-L150C763

@jamesacklin 
Same as with the previous comment, I don't know if this has other consequences, but IMHO the dialog looks better with these changes.

If everyone agrees, I can commit the `mt-10` change and this can be merged.

Let me know and so close the PR w/o merge otherwise.

Screenshot 2023-09-08 at 4 59 57 PM

mrozanski:
Just to be clear, my suggestion is limited to addressing the scrollbars and alignment issues, **not the flickering**.
The scrollbars are mentioned as nice to have in the OT https://linear.app/tlon/issue/LAND-872#comment-358002ee
The flickering will require a separate PR (see my last comment from 9/7 in the ticket)

jamesacklin:
Looks good, thanks!

PR #2847, opened 2023-09-02
groups: add frontend for notification volume control

patosullivan:
Fixes LAND-761

Looks like this one mobile:

https://github.com/tloncorp/landscape-apps/assets/1221094/dc6d8bb5-e5ad-4c7a-9ddf-5492cda1cfc6

And like this on desktop:
 
https://github.com/tloncorp/landscape-apps/assets/1221094/aada0fc7-ecbd-4b2d-a66e-8bf35f96f52f


patosullivan:
Updated to account for %loud notifications (every chat message or note or gallery item from a channel):

https://github.com/tloncorp/landscape-apps/assets/1221094/990d23a8-8f25-4033-854e-c304fed73b41

localhost_3000_apps_groups_groups_~dotdev-dotdev-finned-palmer_new-dotdev-group_activity(iPhone 12 Pro)

urcades (changes requested):
The implementation of the interface itself is feeling solid, and looks functional — the one issue that I'm having is with the specification of the muting levels themselves:

Screenshot 2023-09-07 at 12 43 08 PM

~~The various levels here indicated don't conform to how notification muting/allowing interfaces often work on other platforms (I only checked out Slack and Discord here as reference points fwiw) — "All Messages" here, by definition, should enable push/activity notifications for -all- notifications, including mentions or replies.~~

~~Having a separate level for "All Messages and Replies" seems duplicitous here, and in most apps I've quickly surveyed, the other option often listed is "(Send a notification for..) Mentions Only".~~

~~If it isn't too much to ask for (or a massive lift on pushing this PR through), I'd like us to update the UX here to accommodate for what feels like a more normal approach to the various levels of notification allowance we offer here.~~

I misread the label! Ignore all of this.

I think we should just change the label of the row to "Mentions and Replies Only", that's all!

~~cc @jamesacklin ~~

urcades (approved):
W/ label change, this is looking good, let's get it in

ui/src/components/RadioGroup.tsx
```diff
@@ -0,0 +1,72 @@
...
+          key={option.value}
+          id={option.value}
+          value={option.value}
+          aria-label={option.ariaLabel ?? option.label}
+          className={cn(
+            'flex h-14 w-full cursor-pointer items-center justify-between gap-4 border border-gray-100 py-2 px-6',
+            index === 0 ? 'rounded-t-xl' : '',
+            index === options.length - 1 ? 'rounded-b-xl' : '',
+            index !== 0 && index !== options.length - 1 ? 'border-t-0' : '',
+            index === options.length - 1 ? 'border-t-0' : '',
```
latter-bolden, line 46: looks like we're checking on this condition twice, can we fold that into a single check?

latter-bolden (approved):
One micro nitpick, but frontend code LGTM. 

Nice work!

PR #2852, opened 2023-09-06
Chat: view all reactions

latter-bolden:
Adds support for viewing every reaction and it's author for a message (rather than the limited preview capabilities we have now). Accessible on desktop from the hover menu, on mobile from long press main menu or via long press on the reaction row.

Fixes LAND-369

 

https://github.com/tloncorp/landscape-apps/assets/90741358/91d45538-0abe-4419-87ce-101569cd85df

https://github.com/tloncorp/landscape-apps/assets/90741358/2c816e9e-e0ba-40e7-8155-262052e09d6b


jamesacklin (approved):
image

Beautiful, thank you!

ui/src/chat/ChatReactions/ReactionsWidget.tsx
```diff
@@ -0,0 +1,123 @@
+import _ from 'lodash';
```
alecananian, line 1: I'm not sure if this has been benchmarked for this repo, but you may want to use `import invertBy from 'lodash/invertBy';` to avoid import bloat: https://www.blazemeter.com/blog/import-lodash-libraries

ui/src/chat/ChatReactions/ReactionsWidget.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const feelCount = useMemo(() => Object.keys(feels).length, [feels]);
+  const feelsByShortcode = useMemo(() => _.invertBy(feels), [feels]);
+  const allFeels = useMemo(
+    () => Object.entries(feels).sort(byShortcode),
+    [feels]
+  );
+  const allShortcodes = useMemo(
+    () => Object.entries(feelsByShortcode).map(([code]) => code),
+    [feelsByShortcode]
+  );
```
alecananian, line 76: The memoization here could be unnecessary overhead (see https://blog.isquaredsoftware.com/2020/05/blogged-answers-a-mostly-complete-guide-to-react-rendering-behavior/#memoize-everything)

ui/src/chat/ChatReactions/ReactionsWidget.tsx
```diff
@@ -0,0 +1,123 @@
...
+      </Tabs.List>
+      {tabIds.map((tabId) => (
+        <Tabs.Content
+          key={tabId}
+          value={tabId}
+          className="h-full overflow-y-auto pl-3 pr-5"
+        >
+          {tabId === 'all' &&
+            allFeels.map(([ship, shortcode]) => (
+              <FeelRow ship={ship} shortcode={shortcode} />
```
alecananian, line 113: Missing `key` prop

ui/src/chat/ChatReactions/ReactionsWidget.tsx
```diff
@@ -0,0 +1,123 @@
...
+            ))}
+          {tabId !== 'all' &&
+            feelsByShortcode[tabId].map((ship) => (
+              <FeelRow ship={ship} shortcode={tabId} />
```
alecananian, line 117: Missing `key` prop

ui/src/chat/ChatReactions/ReactionsWidget.tsx
latter-bolden: All of these involve iterating over or sorting the list of reactions which is why they were memoized.

It is a bit of a toss up for now since that list is usually quite small. I think we're okay leaving this in though since the component is rendered infrequently and any overhead incurred won't affect app performance more broadly.

ui/src/chat/ChatReactions/ReactionsWidget.tsx
latter-bolden: We use this import style for lodash pretty regularly in the repo. From an ergonomic standpoint, it's nice to have them all accessible without needing to think about explicit imports.

We probably should do a more serious pass around import bloat at some point, but that's outside the scope of this update. Ideally, we could keep our current import style and ensure tree shaking is appropriately handled by the compiler.

PR #2853, opened 2023-09-07
groups: retain scroll position through nav changes

dantish:
Fixes LAND-804

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -110,7 +110,7 @@
         "react-router-dom": "^6.3.0",
         "react-select": "^5.3.2",
         "react-tweet": "^3.0.4",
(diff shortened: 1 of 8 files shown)


patosullivan:
Looks good to me!

arthyn:
it said `currentVirtuosoRef?.getState` is not a function

dantish:
Probably it means that the version of the lib hasn’t been updated on your machine for some reason, maybe try to remove node_modules and install again?

jamesacklin:
A note from standup: clearing .vite cache seems to work, as well as node_modules

arthyn (approved):
lgtm, let's play around with it and adjust if necessary

PR #2854, opened 2023-09-08
chat: transfer channels

arthyn:
This allows us to move full chat logs to another channel. The current channel will be truncated based on the time given. This is a temporary fix for us "truncate" chat channels in Tlon Local for new joins, so should speed up group joins significantly.


arthyn:
Note I don't want to merge develop into it yet so that we can deploy this independently to ~nibset-napwyn

midsum-salrux (approved):
LGTM

desk/app/chat.hoon
```diff
@@ -1668,6 +1685,45 @@
...
+    =/  writ-log
+      =-  +:-
+      %^  (dip:log-on:c @)  log.chat  ~
+      |=  [st=@ =time =diff:c]
+      :_  [%.n st]
+      ::  only keep writs
+      ?.  ?=(%writs -.diff)  ~
+      `diff
+    =/  filtered-log
+        =-  +:-
```
Fang-, line 1700:
```suggestion
        =<  +
```

Same for `=/  writ-log`.

desk/app/chat.hoon
```diff
@@ -1668,6 +1685,45 @@
...
+        %^  (dip:log-on:c @)  log.chat  ~
+        |=  [st=@ =time =diff:c]
+        :_  [%.n st]
+        ::  keep non-writ events
+        ?.  ?=(%writs -.diff)  `diff
+        ::  only keep writs after time
+        ?.  (gth time tim)  ~
+        `diff
+    =+  .^(=group:g %gx (weld ca-groups-scry /noun))
+    =/  =channel:g  (~(got by channels.group) [%chat flag])
```
Fang-, line 1710:
```suggestion
    =+  .^(=group:g %gx (weld ca-groups-scry /noun))
    =/  =channel:g  (~(got by channels.group) [%chat flag])
    ?>  !=(group.channel new-group)
```

Should probably assert we're actually moving it to a different group (because otherwise this archive, too, is just auto-joined).

Likewise for making sure we're moving it to a non-existent channel, not the same exact channel, etc.

Fang- (commented):
Comment about guaranteeing sanity, and a style nit, but otherwise this seems good. (:

Fang- (approved):
Changes look good. Let's get it! (:

jamesacklin:
Let's trim! Thanks.

PR #2857, opened 2023-09-11
diary: fix edit note crash and missing add note button

latter-bolden:
Fixes two bugs we'd been tracking:

- The `cannot read 0 of undefined` issue that was preventing notes from being edited
- The permissions issue causing  the "Add Note" button to not be shown

Also adds some stronger typing and defensive naming to prevent similar issues in the future.

Fixes LAND-870


patosullivan (approved):
Looks good.

PR #2859, opened 2023-09-12
heap: consistent curio delete behavior

latter-bolden:
A few fixes to ensure curio deletion works as expected:
- make sure you always have permission to delete if the option is shown in the UI
- properly track the delete and spin while waiting
- successful delete triggers consistent removal from cache and query invalidation

Also updates the delete curio modal to close when you click outside of it.

Fixes LAND-839


jamesacklin (approved):
Verified that this works on a livenet moon operating on a gallery hosted by a mainnet planet. Nice work!

PR #2864, opened 2023-09-13
nav: persist location state for groups and messages when navigating between tabs on mobile

latter-bolden:
OTT

Double clicking returns you to the root location.

https://github.com/tloncorp/landscape-apps/assets/90741358/eed83afb-5809-4a8c-a733-9f0c9e4ce41a

Fixes LAND-848


urcades:
This is looking good

ui/src/state/ephemeral.ts
```diff
@@ -0,0 +1,21 @@
+import create from 'zustand';
```
arthyn, line 1: we use "local.ts" for random stuff like this

arthyn (approved):
lgtm, might be good to combine that state with local

ui/src/state/ephemeral.ts
latter-bolden: was going to add there, but looked like that still gets persisted to `localStorage`?

this felt like something that should fully reset when you close the app

latter-bolden: jk, missed the partialize call there. Will update

PR #2865, opened 2023-09-13
groups: fix native safe area insets

dnbrwstr:
This PR is one half of a fix for safe area insets in the webapp. The other half involves changes to the native app repo here: https://github.com/tloncorp/talk-android/pull/31.

In order to properly render modals and various headers and footers, the webapp needs to have total control over the rendering surface. To make that possible, I've removed the native `SafeAreaView` wrapping the entire app and instead opted to inject safe area values into the webapp. The safe area values are accessed through a hook, and default to `0` values if nothing has been injected. This PR contains plumbing for that, as well as application to headers/footers where necessary.

Fixes LAND-900


patosullivan (approved):
Looks good.

PR #2866, opened 2023-09-14
fix lure link toggling when check request times out

alecananian:
Fixes issue with lure link toggling / popping in and out on the "invite people" dialog, which was occurring because the API call would time out and default to `false`. Now it will default to the previously returned value if a timeout occurs.

### DIFF
--- a/ui/src/state/lure/lure.ts
+++ b/ui/src/state/lure/lure.ts
@@ -10,7 +10,7 @@ import {
 import { GroupMeta } from '@/types/groups';
 import { useQuery } from '@tanstack/react-query';
 import produce from 'immer';
-import { useEffect, useCallback, useState } from 'react';
+import { useEffect, useCallback, useRef } from 'react';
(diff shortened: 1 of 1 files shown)


arthyn:
https://github.com/tloncorp/landscape-apps/assets/5466421/da7ddbf9-5a4e-47a5-99cf-139eb83ba94e

still seeing this on this branch

alecananian:
@arthyn what's the desired behavior here? The bug I was observing was that while you had "Invite Links" checked, the input box below was popping in and out periodically.

alecananian:
I've tried to cover the blank state shown in the video as well, but the amount of states being checked here and an API call that is constantly timing out has made this really difficult to manage. I'll merge this and revisit the rest of the flickering when implementing the QR code.

arthyn:
yeah we'd like to collapse a lot of that stuff on the backend so the frontend is easier to manage, so yeah feel free to defer for that

PR #2868, opened 2023-09-18
groups: remove mobile redirects

arthyn:
This fixes both LAND-902 and LAND-897 by removing the redirects. For 902 we also add in a prompt to go to the find tab since we no longer have a redirect. This is a stopgap until we have a better empty state.

image

Fixes LAND-902
Fixes LAND-897


patosullivan (approved):
Lgtm

jamesacklin (approved):
Redirect (or lack thereof) works naturally. Thank you.

PR #2870, opened 2023-09-18
groups+talk: add basic error logging

dnbrwstr:
Adds a PostHog `error` event, triggered by the app `ErrorBoundary` or the global window error event.


arthyn (approved):
lgtm

PR #2872, opened 2023-09-19
gallery: undo complicated placeholder for simpler solution

arthyn:
We tried to replace the placeholder with something that included an upload button. This had more consequences than we'd like including not showing cursor and breaking link pasting. This fixes that by restoring the old placeholder and just adding a simple button at the bottom of the modal. Fixes LAND-928

image
image

This is mostly untested just wanted to get it up before the end of the day.


jamesacklin:
Thanks. I think this is completely acceptable; we can iterate on that form in the future if we have more time.

jamesacklin (approved):

PR #2873, opened 2023-09-19
fix take photo flow in mobile safari

alecananian:
Fixes the "Take a Photo" flow in mobile Safari (and therefore the Groups app webview) by appending the file input to the DOM during file selection


arthyn (approved):
lgtm, interesting haha

PR #2877, opened 2023-09-19
messages: adds MobileHeader

jamesacklin:
Fixes LAND-933 again by adding a MobileHeader to the new DM screen (if you're in the mobile app). The Back button goes back to the Messages index.

image


arthyn (approved):
lgtm

PR #2879, opened 2023-09-22
groups: echo invites in groups list

jamesacklin:
- Reuses GroupJoinList in MobileRoot, which fixes LAND-947
- Shifts GroupJoinItem to use SidebarItem to allow them to slot in to the Groups list with uniform appearance
- Adds an optional highlightAll prop to GroupJoinList, which is passed down to GroupJoinItem to show a blue-soft background
- Adds an optional "preload" prop to GroupJoinItem for pre-filling the group item with static data until the gang data arrives; used in the "Find Groups" screen for the UF, door, and Tlon Local suggestions
- Unifies the separators in the Messages list with the Groups list
- Does NOT remove the "Find Groups" screen from the nav or invites from the Activity stream; meant to be an incremental change

### DIFF
--- a/ui/src/components/Sidebar/SidebarItem.tsx
+++ b/ui/src/components/Sidebar/SidebarItem.tsx
@@ -168,7 +168,7 @@ const SidebarItem = React.forwardRef<HTMLDivElement, SidebarProps>(
           )}
         </Action>
         {actions ? (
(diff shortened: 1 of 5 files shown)


latter-bolden (approved):
lgtm, nice work.

invites definitely feel more natural there

PR #2880, opened 2023-09-22
groups: activity indicators in mobile navbar

jamesacklin:
- Reformats the navbar slightly to accommodate tap targets with the intended margins, which fixes LAND-894
- Adds useIsAnyGroupUnread, a superset of what is returned by useIsGroupUnread
- Uses a variety of methods (useIsAnyGroupUnread, useChannelUnreadCounts, useNotifications) to show or hide a blue dot under the Groups, Messages, and Activity navigation items
- Converts the Activity navigation item to a DoubleClickableNavTab, since we will want to support linking to the notification context relative to the activity stream (both events are the same now, which might be premature)

Fixes LAND-945

### DIFF
--- a/ui/src/components/Sidebar/MobileSidebar.tsx
+++ b/ui/src/components/Sidebar/MobileSidebar.tsx
@@ -4,6 +4,7 @@ import { isNativeApp, useSafeAreaInsets } from '@/logic/native';
 import { useIsDark } from '@/logic/useMedia';
 import { useIsAnyGroupUnread } from '@/logic/useIsGroupUnread';
 import { useChannelUnreadCounts } from '@/logic/channel';
+import { useNotifications } from '@/notifications/useNotifications';
 import { useLocalState } from '@/state/local';
 import NavTab, { DoubleClickableNavTab } from '../NavTab';
 import BellIcon from '../icons/BellIcon';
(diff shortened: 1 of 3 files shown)


ui/src/components/Sidebar/MobileSidebar.tsx
```diff
@@ -29,12 +33,17 @@ function GroupsTab(props: { isInactive: boolean; isDarkMode: boolean }) {
...
     >
       <HomeIconMobileNav
         isInactive={props.isInactive}
         isDarkMode={props.isDarkMode}
-        className="mb-0.5 h-6 w-6"
+        className="h-6 w-6"
+      />
+      <div
+        className={cn(
+          'mt-[2px] h-1.5 w-1.5 rounded-full',
```
arthyn, line 44: image

so our icons in the FE are sized at 24px with no frame, but in the nav mocks here their frame is 32px with the dot butt up against the frame.

image

I think that distance is 4px between icon and bottom of frame so in the outlines here is top margin at 4px which I think is closer to the mocks.

jamesacklin:
@arthyn PTAL, we now have:
- a 48px navbar
- 32px icon containers with 24px centered icons
- 6px bullets stacked immediately underneath the icon containers
- 2px padding between the bullet and the bottom of the navbar

arthyn (approved):
primo

PR #2882, opened 2023-09-25
author: useNavigate() for profile modal

jamesacklin:
Changes the `Author` component to use useNavigate instead of useModalNavigate. This fixes LAND-953 (where the useModalNavigate hook would direct you to the profile modal without the matching backgroundState, resulting in a blank white screen). 

Note that this is global to `Author`. I checked it everywhere I could possibly find an `Author` with no regression, but would appreciate a second look.


latter-bolden (approved):
also kicked the tires for regressions, looking good.

PR #2883, opened 2023-09-25
HAK-26: Documentation for channels API after refactor

midsum-salrux:
https://linear.app/tlon/issue/HAK-26/docs-information-and-examples-for-how-to-work-with-the-new-backend

https://github.com/tloncorp/landscape-apps/blob/ns/docs-for-refactor/docs/channels.md


midsum-salrux:
TODO: find and fix ambiguous heading links

docs/channels.md
```diff
@@ -0,0 +1,760 @@
...
+`%blockquote`: blockquote surrounded content
+`%block`: link/reference to blocks
+`%code`: code formatting for large snippets
+`%tag`: tag gets special signifier
+`%link`: link to a URL with a face
+`%break`: line break
+### kind-data
+```hoon
++$  kind-data
+  $%  [%channel title=@t image=@t]
```
arthyn, line 511:
```suggestion
  $%  [%diary title=@t image=@t]
```

docs/channels.md
```diff
@@ -0,0 +1,760 @@
...
++$  nest  [=kind =ship name=term]
+```
+
+Uniquely identifies a channel.
+
+[kind](#kind)
+
+### kind
+```hoon
++$  han  ?(%notebook %gallery %chat)
```
arthyn, line 757:
```suggestion
+$  kind  ?(%notebook %gallery %chat)
```

docs/channels.md
```diff
@@ -0,0 +1,760 @@
+# Channels
```
arthyn, line 1: We probably should have some kind of introductory paragraph about what channels is, what kinds of posts we have and how those are displayed. @jamesacklin wrote some stuff up in the readme, maybe can help with this. This can wait until later though just leaving a note here.

docs/channels.md
```diff
@@ -0,0 +1,760 @@
...
+      [%read-at =time]
+      [%watch ~]
+      [%unwatch ~]
+  ==
+```
+
+Response (subscriber to client communication) for a channel.
+### posts
+```hoon
++$  notes  ((mop id-post (unit note)) lte)
```
arthyn, line 602:
```suggestion
+$  posts  ((mop id-post (unit post)) lte)
```

docs/channels.md
```diff
@@ -0,0 +1,760 @@
...
+  sent=now
+==
+=kind-data [%chat ~]
+=essay [memo kind-data]
+=post-action [%add essay]
+=action [%post post-action]
+
+:channels &channel-action [%channel nest action]
+```
+
```
arthyn, line 207: A marker here for where commands start and maybe an explainer about what that means semantically.

arthyn (changes requested):
This is looking good. I think there's still a few instances of "note" lingering, and I would double check the types around search and said since those changed recently. We'll have to do another pass before the refactor goes out too so I think we should just get this merged soon.

docs/channels.md
```diff
@@ -0,0 +1,768 @@
...
+Watch a channel for unreads.
+###### %unwatch
+```hoon
+=nest [%chat our %mychat]
+=action [%unwatch ~]
+
+:channels &channel-action [%channel nest action]
+```
+
+Stop watching a channel for unreads.
```
Fang-, line 176: What does it mean to "watch a channel for unreads"?

docs/channels.md
```diff
@@ -0,0 +1,768 @@
...
+
+###### %join
+```hoon
+=nest [%chat ~palnet-sampel %mychat]
+=action [%join group=[~sampel-palnet %mygroup]]
+
+:channels &channel-action [%channel nest action]
+```
+
+Join a channel. In this example, ~sampel-palnet is the group host and ~palnet-sampel is the channel host.
```
Fang-, line 131: Humbly requesting more obviously distinct sample `@p`s. Something like:

````suggestion
###### %join
```hoon
=nest [%chat ~sampel-banter %mychat]
=action [%join group=[~hostes %mygroup]]

:channels &channel-action [%channel nest action]
```

Join a channel. In this example, ~hostes is the group host and ~sampel-banter is the channel host.
````

File: docs/channels.md [hunk 8]
```diff
@@ -0,0 +1,768 @@
...
+=nest [%chat our %mychat]
+=action [%unwatch ~]
+
+:channels &channel-action [%channel nest action]
+```
+
+Stop watching a channel for unreads.
+
+##### Actions
+All of the previous pokes affected your local ship (joining and leaving, unreads, and so on). The following pokes are *actions*, meaning they will be synced across the network. When `%channels` on your ship hears an action, it issues a *command* to the host's `%channels-server` (see the above [diagram](#diagram)).
```
Fang-, line 179: Should probably explain this ahead of time. But also, the previous ones were actions, too, by our type naming convention. Marking some as "affects local state only" is fair, but shouldn't be made to seem like a big deal, and certainly shouldn't conflate terminology (especially if we're not talking about `%channels-server`/`$command`s here).

The heading for this could be nicer, but per the above, I'd argue this doesn't need a separate heading/nesting level at all.

Fang-: Looking at the `/sur` file, I can see how this emphasis came about: the post & reply action types are defined in terms of their command types. Considering that the action types are the outward-facing ones, that should probably be the other way around. I may touch that up.

docs/channels.md
```diff
@@ -0,0 +1,768 @@
...
+
+[arranged-posts](#arranged-posts)
+###### %add-writers
+```hoon
+=nest [%chat our %mychat]
+=action [%add-writers (sy ~[%role1 %role2])]
+:channels &channel-action [%channel nest action]
+```
+
+Enable writing to a channel for a certain set of roles
```
Fang-, line 345: Here and below, "allow" and "disallow" might be better wording than "enable" and "disable" respectively. More evocative of dealing with permissions.  
Maybe also just "certain roles".

docs/channels.md
```diff
@@ -0,0 +1,768 @@
...
+  secret=%.n
+  ==
+
+:groups &group-create create
+```
+
+This is done with `%groups` instead of `%channels`, but it's included here since it's necessary for testing.
+### Subscriptions
+#### /unreads
+Subscribe to unread & preview information. Each fact is a `[channels unreads]`
```
Fang-, line 385: These should all be telling the mark for the facts, in addition to the type contained within that mark.

docs/channels.md
```diff
@@ -0,0 +1,768 @@
...
+
+[channels](#channels) \| [unreads](#unreads)
+#### /
+"Firehose" subscription path. Each fact is an `r-channels` (response channels)
```
Fang-, line 389: I'm not sure if "firehose" is a broadly understood technical term. Consider elaborating on what exactly you mean by it.

docs/channels.md
```diff
@@ -0,0 +1,768 @@
...
+
+[r-channels](#r-channels)
+#### /[kind]/[ship]/[name]
+(This is effectively /[nest])
+
+Similar to `/`, but only include updates for a particular channels. Each fact is an `r-channels` (response channels)
+
+[nest](#nest) \| [r-channels](#r-channels)
```
Fang-, line 397: `/[nest]`, especially written next to all these other examples, suggests a path with a single element in it. If you want to point this out, I would just write something like:

```suggestion
#### /[kind]/[ship]/[name]

Similar to `/`, but only include updates for a particular channels. Note how the path fully describes a specific [`$nest`](#nest). Each fact is an [`$r-channels`](#r-channels) (channels response).
```

For other paths, I would either repeat the same short phrasing, or omit it entirely.

docs/channels.md
```diff
@@ -0,0 +1,768 @@
...
+#### /said/[kind]/[ship]/[name]/post/[time]/[(unit reply)]
+(This is effectively /said/[nest]/post/[time]/[(unit reply)])
+
+Read a reference. Facts are either of the `%channel-denied` mark (meaning you cannot view the channel) or of the `%channel-said` mark and the `said` type
+
+[nest](#nest) \| [said](#said)
+## Types
```
Fang-, line 404: Unclear to me if we should include these here, or just do permalinks to the type definitions in `/sur/channel`. Arguably any comments on the types we give here should be included in the sur file itself.

I guess having them in the document with anchor tags makes it faster for the reader to browser-back-button (#bbb) between the definitions and the part they were reading...

docs/channels.md
```diff
@@ -0,0 +1,768 @@
...
+  ==
+```
+
+Permissions for a channel
+
+[sect:groups](#sectgroups) \| [flag:groups](#flaggroups)
+### v-channel
+Internal representation of a channel. Contains syncing information and revision numbers.
+
+[channel](#channel)
```
Fang-, line 752: This one is never referenced, and described as "internal", so per my other comments we shouldn't be including this.

docs/channels.md
```diff
@@ -0,0 +1,768 @@
...
+Uniquely identifies a channel.
+
+[kind](#kind)
+
+### kind
+```hoon
++$  kind  ?(%notebook %gallery %chat)
+```
+
+Channel type.
```
Fang-, line 768: I would probably put these at the very top of the types section, and generally order them so that primitives/foundationals come first. Right now the first type is described as "A nest and a note", but we haven't learned what either of those are.

Fang- (commented):
Physically hurting because of the lack of whitespace around these `###` heading lines. (;

I don't really like the general layout of the pokes section, but I can't put my finger on what it is about it... Maybe the absence of a proper type specification?  
In that same vein, things like the marks for scries, and the fact that the poke headings are the marks, and the types for the poke marks, all that being _implicit_ is a little bit rough. I don't want to tell you to carefully do "mark | type | description" tables or w/e, maybe a simply "here's how to read this" line at the start of each section suffices...

...Although, maybe tables _would_ be nice for helping with scanability of the doc. Just chuck each section into a big table, instead of having all these small headings. (I've certainly been feeling "this looks not great" throughout, but that alone isn't a constructive thought.) Does the `%docs` app markdown renderer support tables?

docs/channels.md
midsum-salrux: @arthyn what do you think?

arthyn (approved):
lgtm!

docs/channels.md
```diff
@@ -0,0 +1,103 @@
...
+| `/u/[kind]/[ship]/[name]` | loob | Check the existence of a channel | <code>.^(? %gu /=channels=/chat/~sampel-palnet/mychat/loob)</code>
+
+## Pokes
+All pokes here have the [channel-action](https://github.com/tloncorp/landscape-apps/blob/hm/hackweek/desk/mar/channel/action.hoon) mark.
+
+### Channels
+| Description | Example
+|-------------|--------
+| Create a channel. This also updates the metadata in the `%groups` agent. See [$create-channel](https://github.com/tloncorp/landscape-apps/blob/hm/hackweek/desk/sur/channel.hoon#L245) | <pre>=create-channel :\*<br>  %chat<br>  %mychat<br>  [our %mygroup]<br>  title='My chat channel'<br>  description='Chat about things'<br>  readers=\~<br>  writers=\~<br>==<br><br>:channels &channel-action [%create create-channel]</pre>
```
Fang-, line 63: Wondering if these would benefit from having a first column that super briefly describes the shape or tag of the action.

Maybe like `[%create *]`, `[%channel * %unwatch *]`, `[%channel * %post %reply * %del-react *]`,  
or even terser like `%create`, `%unwatch`, `%del-react` respectively, and then just mentioning the `[%channel *]` case and such, as pointers to the following section.

Unsure if that would be better or not, but currently feels like there's some indexing/skimmability stuff missing here. The descriptions themselves are a bit too verbose for that.

Fang- (commented):
This feels a lot nicer already.

Also, we discussed this out of band, but ftr: there might be some merit to using explicit `<table>` tags instead of table syntax (cleaner source, ability to put ids/anchors in), but we can hold off on that until the need arises.

docs/channels.md
```diff
@@ -0,0 +1,103 @@
...
+| Disallow writing to a diary for a certain set of roles. | <pre>=nest [%chat our %mychat]<br>=action [%del-writers (sy ~[%role1 %role2])]<br><br>:channels &channel-action [channel nest action]</pre>
+| Pin channels | <pre>=pins ~[[%chat ~pinner %mychat]]<br><br>:channels &channel-action [%pin pins]</pre>
+
+### Unreads
+| Description | Example
+|-------------|--------
+| Mark a channel as read. | <pre>=nest [%chat our %mychat]<br>=action [%read ~]<br><br>:channels &channel-action [%channel nest action]</pre>
+| Mark a channel as read up to a certain time. | <pre>=nest [%chat our %mychat]<br>=action [%read-at now]<br><br>:channels &channel-action [%channel nest action]</pre>
+| Watch a channel for unreads. This affects notifications on the frontend. | <pre>=nest [%chat our %mychat]<br>=action [%watch ~]<br><br>:channels &channel-action [%channel nest action]</pre>
+| Stop watching a channel for unreads. This affects notifications on the frontend. | <pre>=nest [%chat our %mychat]<br>=action [%unwatch ~]<br><br>:channels &channel-action [%channel nest action]</pre>
```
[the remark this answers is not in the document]
midsum-salrux: Just found out that these are currently unused, so I added a note about that.

PR #2884, opened 2023-09-25
Chat scroller rework

dnbrwstr:
This PR fixes major issues with the scroller on mobile. It's not absolutely perfect, but should be an improvement.

Further work/known issues:
- Fix scrolling to newest items
- Fix main pane scroll when scrolling to an item in a thread
- Polish loader appearance
- Implement better `estimateSize`
- Improve key and spacebar navigation on desktop

Fixes LAND-898
Fixes LAND-406
Fixes LAND-691


arthyn (approved):
lfg!!!! would be cool to walk through this with the group at some point

latter-bolden:
code is looking great! This is feeling buttery on both desktop and mobile, was immediately sad when i had to go back to the production version 🙃

the only non-fine tuning issue I ran into a few times was a channel getting stuck and not loading anything beyond the current messages it had. I _think_ this is just an issue with the top/bottom detection in the fetch new messages effect because resizing the screen did seem to resolve it. I'll try to snag a video, but it was pretty sporadic.

"view new messages" banner was also working intermittently for me, but i belive that's a known issue.

PR #2887, opened 2023-09-27
navigation: allow user to navigate to last route from back button

patosullivan:
Fixes LAND-920 by having back buttons in particular components just navigate back to the last route with react-router's `-1` utility.

### DIFF
--- a/ui/src/channels/ChannelHeader.tsx
+++ b/ui/src/channels/ChannelHeader.tsx
@@ -71,7 +71,7 @@ export default function ChannelHeader({
             {children}
           </div>
         }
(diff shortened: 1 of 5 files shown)


latter-bolden (changes requested):
After following a thread reply notification, clicking back doesn't return me to the activity page. Instead, I get stuck in a navigation loop between the thread and it's containing channel.


https://github.com/tloncorp/landscape-apps/assets/90741358/74a7e835-0567-4f77-a58b-934c98b72843

arthyn:
if we do this from a native notification you will also get stuck as well because we have no back history. @jamesacklin @urcades @dnbrwstr maybe we should discuss next week

PR #2889, opened 2023-09-28
chat: prevent overfetching from scroller

arthyn:
We were missing a few stops so that we didn't just continuously fetch.


dnbrwstr (approved):
lgtm

PR #2890, opened 2023-09-28
ui: Edit Profile layout for large devices

tomholford:
This PR adds a style tweak to ensure that the Edit Profile view renders at full width on larger screens. Previously (on `0vp.gv2u9.9hg31.7jgob.84vt5.jobhe.he7k6.kfnia.95i4b.00vu4.eaocr`) it was rendering the content at the mobile width, with extraneous space to the right like so:

image


patosullivan (approved):
Thanks Tom!

PR #2891, opened 2023-09-28
landscape: update /grid paths --> /landscape

tomholford:
In light of the recent migration from `/grid` --> `/landscape` for 412k, this updates remaining path references throughout the project. 

- `AppReference` - when rendering a link to an app, it now deep links to Landscape instead of Grid
- `Sidebar` & `MessagesSidebar` - both have a 'Back to Landscape' button that now points to `/apps/landscape`  instead of `/apps/grid`

If we're lucky, this might also be a step in the direction of fixing e2e tests (currently broken on `develop`)


jamesacklin (approved):
Thanks @tomholford!

PR #2892, opened 2023-09-28
Fix scroller crash in empty channels

dnbrwstr:
Fixes LAND-963

Two issues here:
- We were attempting to scroll to an undefined index in empty channels
- `hasLoadedOldest` wasn't properly set for empty channels


patosullivan (approved):

arthyn (approved):
lgtm

PR #2894, opened 2023-09-29
groups: fix chat input jump

dnbrwstr:
Fixes [LAND-803](https://linear.app/tlon/issue/LAND-803/chat-apparent-flicker-between-chat-channels-on-mobile)

The message editor is initially null when rendering the input, which means that the input is never rendered on the first pass. Currently, this causes content to jump up immediately after the first render. This PR fixes that by rendering the input HTML before the editor is initialized.


arthyn (approved):
bless you

PR #2896, opened 2023-09-29
groups: mobile layout & select fixes

latter-bolden:
- Long press in channel/message list views no longer selects the whole screen
- Focusing on the channel filter input no longer zooms in
- Update banner is positioned visibly just below the top navigation
- Smuggled in a fix for misattributed Gallery author in references

Note, this adds `select-none` fairly high up in the component tree on mobile which affects much of the app. Since generally selecting arbitrary elements comes across as webby and non-native, I think this makes sense. But we'd need to be careful about marking places where select should be possible (Notebook content for example is explicitly made selectable in this PR).

Fixes LAND-916
Fixes LAND-891
Fixes LAND-917
Fixes LAND-951

### DIFF
--- a/ui/src/components/ActionMenu.tsx
+++ b/ui/src/components/ActionMenu.tsx
@@ -73,7 +73,7 @@ const ActionMenu = React.memo(
                 disabled={disabled}
                 asChild={asChild}
                 aria-label={ariaLabel}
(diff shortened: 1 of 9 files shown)


arthyn (approved):
lgtm

PR #2897, opened 2023-09-29
groups: fix message options placement

dnbrwstr:
This PR fixes the placement of message options on desktop post-new-scroller.

The issue was that `position: sticky` behaves differently inside the inverted scroller. We were only using `sticky` to ensure that options remained visible when a message was scrolled off the page, so for now I've replaced that with a simple hack to correct position after render.

A more comprehensive + performant solution would be to move options rendering outside of the message and deal with positioning there. I started down that route, but it looked like fully extracting options from message was a bigger change than it made sense to make just for this. I think it is worth doing, but holding off for now.


arthyn (approved):
lgtm

PR #2900, opened 2023-09-29
nav: use more accurate unread state, and unify native nav and non-native

arthyn:
This should make unread state in the nav match user action for the messages tab.

Also this makes the non-native nav and native nav the same so people can get the same experience.


patosullivan (approved):
Lgtm

PR #2901, opened 2023-09-30
bootstrap: revert reconnect changes for now

arthyn:
OTT, seeing bad behavior, will investigate


patosullivan (approved):
I'm seeing it too.

PR #2903, opened 2023-10-02
chat: fix unread banner

arthyn:
Fixes LAND-970 and LAND-977. This involved adding a few things to the backend: 
- Times on the subscription fact of all "added" writs. This ensures that times for new messages exist and are correct.
- A scry endpoint for fetch around id, since briefs only gives us an ID we needed a way to make sure we can get messages near the unread
- Changing when we give updates about writs so that it's always after we've resolved the pact
- Adding a missing mark for a single writ, there's a few places we ask for a single writ which were relying on the messages existing already, instead we just scry for them

So now if you have an unread:
- we fetch around the message
- if it's in a thread we also fetch around the thread OP

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -950,6 +950,18 @@
       %del-feel  =(src.bowl p.delta)
   ==
 ::
+++  diff-to-response
+  |=  [=diff:writs:c =pact:c]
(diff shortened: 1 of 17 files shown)


patosullivan (approved):
Looks good to me!

PR #2904, opened 2023-10-03
chat: mobile search overhaul

latter-bolden:
- Updates search on mobile to take up the whole page
- Displays previous few queries and the last set of results
- Changes the search function to fetch fewer results for faster initial load and populates more infinitely as you scroll
- Works across channels, DMs, and group chats

Fixes LAND-923

https://github.com/tloncorp/landscape-apps/assets/90741358/bb1fdacc-e3df-4742-ae1c-e1fdbeabf3f3

* note in the vid it hesitates on loading the thread when the search result is clicked. This is normally instantaneous, investigating


ui/src/app.tsx
```diff
@@ -797,7 +807,7 @@ function RoutedApp() {
           <Scheduler />
         </TooltipProvider>
         <LureAutojoiner />
-        {showDevTools && (
+        {false && (
```
arthyn, line 810: you can always just call `toggleDevTools()` from the console, maybe we should make the toggle persist so this is less annoying

arthyn (approved):
lgtm, this is a great improvement

PR #2905, opened 2023-10-03
Scroll improvements

dnbrwstr:
- Somewhat better optimized processing of incoming chat messages
- Improved anchoring of `scrollTo` message

Fixes LAND-975
Fixes LAND-973
Fixes LAND-976
Fixes LAND-978

Fix for 978 isn't 100%, but it should be much improved, and we can reassess after wider testing.


dnbrwstr:
Reasonable stopgap solution here, going to keep hacking on some edge cases. This should also fix LAND-973.

dnbrwstr:
@arthyn latest commit should fix the jump we were seeing last night. This should be good to go.

arthyn (approved):
lgtm, let's get it in to feel it

PR #2906, opened 2023-10-03
Add tlon hosting upload method

guaraqe:
I will sync the PR with the proxy one, and open both at the same time, with the instructions for testing.


guaraqe:
I added more details on how to test in https://github.com/tloncorp/ylem/pull/1715.

This PR has an issue I am not able to debug, which makes an "eyrie" button to appear in the frontend, blocking the upload button. It works if I remove the button, and the file is uploaded, but this is certainly a bug introduced by this PR.

I can try to solve it myself, but I need some guidance, as I don't undrestand the source of this problem.

patosullivan:
> I added more details on how to test in https://github.com/tloncorp/ylem/pull/1715.
> 
> This PR has an issue I am not able to debug, which makes an "eyrie" button to appear in the frontend, blocking the upload button. It works if I remove the button, and the file is uploaded, but this is certainly a bug introduced by this PR.
> 
> I can try to solve it myself, but I need some guidance, as I don't understand the source of this problem.

That button only appears when running the app via the local dev server. Are you saying you ran `npm run prod` and still saw it? Or were you running `dev`?

guaraqe:
Yes, but this button seems to appear because it detects some bug in the frontend. The question is: I don't understand what exactly is the bug, I was not able to identify it from the console logs or requests.

patosullivan:
> Yes, but this button seems to appear because it detects some bug in the frontend. The question is: I don't understand what exactly is the bug, I was not able to identify it from the console logs or requests.

It appears all the time, by default, when you run the app locally.

patosullivan (approved):
Looks good!

PR #2908, opened 2023-10-03
mobile: show copy text button on chat message options

patosullivan:
Adds a "Copy Text" button to the ActionMenu on mobile, fixes LAND-972. Also applies 'select-none' to the actions in the ActionMenu so that users don't accidentally select the button text rather than just pressing the button (not sure if we have an issue for that, couldn't find one).

I intentionally did not add it to desktop since a) it's superfluous b) we would be using a duplicate icon without an easily apparent label.


jamesacklin (approved):
Function works as expected, thank you. Double-copy icon gets a little confusing but we can refine later.

PR #2909, opened 2023-10-03
chat: fix issue with clicking on app references in chat

patosullivan:
Fixes LAND-974, which was caused by:

1) the parent div for chat message options had height set to 100%, so they created an invisible box blocking the content underneath.
2) the app reference was set to width 100%, unlike our other refs which have a max width of 600px.

Tagging @dnbrwstr since I noticed you had set the h-full on the chat message options. Did this solve some particular problem? Can we safely set it to min-h-fit?


dnbrwstr (approved):
Looks good to me -- I think `h-full` was left over from the previous method of adjusting options position to stay on page (was used with `position: sticky`). Afaik should be fine to remove now.

PR #2911, opened 2023-10-03
thread: fix input overlap

arthyn:
This fixes the overlap seen in threads here 

image

Also removes a sigpam that was spamming dojo

### DIFF
--- a/ui/src/chat/ChatThread/ChatThread.tsx
+++ b/ui/src/chat/ChatThread/ChatThread.tsx
@@ -103,7 +103,7 @@ export default function ChatThread() {
 
   return (
     <div
-      className="relative flex h-full w-full flex-col overflow-y-auto bg-white lg:w-96 lg:border-l-2 lg:border-gray-50"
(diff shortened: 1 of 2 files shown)


ui/src/chat/ChatThread/ChatThread.tsx
```diff
@@ -103,7 +103,7 @@ export default function ChatThread() {
 
   return (
     <div
-      className="relative flex h-full w-full flex-col overflow-y-auto bg-white lg:w-96 lg:border-l-2 lg:border-gray-50"
+      className="relative flex h-full w-full flex-col overflow-y-auto bg-white pb-[50px] sm:pb-0 lg:w-96 lg:border-l-2 lg:border-gray-50"
```
arthyn, line 106:
```suggestion
      className="relative flex h-full w-full flex-col overflow-y-auto bg-white lg:w-96 lg:border-l-2 lg:border-gray-50"
```

PR #2913, opened 2023-10-04
mobile: create group refinements

patosullivan:
Fixes LAND-987

https://github.com/tloncorp/landscape-apps/assets/1221094/ae1f07ef-f7de-4da1-a544-563e58312e20

Note that this modifies the mobile header to allow the arbitrary back function, which touches on the general "back button" discussion.


latter-bolden (approved):
lgtm!

PR #2915, opened 2023-10-04
analytics: disable posthog decide check

patosullivan:
When the app initially loads posthog fires off a request to their `/decide` endpoint.  This is just a check to see if the analytics client supports certain features (none of which we use). No user data is sent with the request, but this may make users worry. We can safely disable this.

More details here: https://posthog.com/docs/libraries/js#disable-decide-endpoint


arthyn (approved):
seems fine ty!

PR #2916, opened 2023-10-05
mobile: fix mention suggestion positioning

latter-bolden:
Fixes mentions to appear above the input bar on mobile and reverses result order so the best suggestion is easier to click.

Fixes LAND-929

https://github.com/tloncorp/landscape-apps/assets/90741358/8f6632de-8da3-4834-b4d4-832e57756957


jamesacklin (approved):
Very nice, works well on mobile-web as well:

image

PR #2923, opened 2023-10-05
chat: fix issues with leaving clubs

patosullivan:
Fixes LAND-1001 and possibly other issues.

The reason the blue dot wasn't going away on the Messages icon was because we were still getting briefs for a club we already left. This PR updates briefs so that when we scry for briefs we just send an empty brief for a club that we're neither a member of or invited to. It also makes sure we don't hear about writ updates for the club if we're not invited or a member.

There were two other issues I found while working on this:

1) If you rejected a club invite this would actually cause the messages list to not render at all, since the MessagesSideBarItem would return null and the virtualscroller wouldn't be able to calculate heights. A user might just assume that they lost all of their messages. (I'm currently effected by this bug on my ship, I rejected an invite from a test club). This PR updates the filter that creates `organizedBriefs` in the MessagesList component to filter out briefs for clubs we're not a member of, so they won't appear in the list (clubs we're invited to will still show up).

2) Leaving a club/rejecting an invite wouldn't work from the options dropdown for clubs because we were calling the dmRsvp function rather than multiDmRsvp (because we were not passing the `isMulti` prop into DmOptions). You could only do it via the DmInvite component. This PR passes `isMulti` for that case and now it's fine.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -958,9 +958,8 @@
   ^-  briefs:c
   %-  ~(gas by *briefs:c)
   %+  welp
-    %+  turn  ~(tap in ~(key by clubs))
-    |=  =id:club:c
-    =/  =club:c  (~(got by clubs) id)
+    %+  turn  ~(tap by clubs)
+    |=  [=id:club:c =club:c]
     =/  loyal  (~(has in team.crew.club) our.bowl)
     =/  invited  (~(has in hive.crew.club) our.bowl)
     ?:  &(!loyal !invited)
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -960,6 +960,11 @@
   %+  welp
     %+  turn  ~(tap in ~(key by clubs))
     |=  =id:club:c
+    =/  =club:c  (~(got by clubs) id)
+    =/  loyal  (~(has in team.crew.club) our.bowl)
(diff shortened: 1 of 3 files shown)


desk/app/chat.hoon
```diff
@@ -960,6 +960,11 @@
   %+  welp
     %+  turn  ~(tap in ~(key by clubs))
     |=  =id:club:c
+    =/  =club:c  (~(got by clubs) id)
```
arthyn, line 963:
```suggestion
    %+  turn  ~(tap by clubs)
    |=  [=id:club:c =club:c]
```

ui/src/dms/MessagesList.tsx
```diff
@@ -66,6 +67,20 @@ export default function MessagesList({
         const group = groups[groupFlag || ''];
         const vessel = group?.fleet[window.our];
         const channel = group?.channels[`chat/${b}`];
+        const club = useChatState.getState().multiDms[b];
+        const dm = useChatState.getState().dms.filter((d) => d === b)[0];
```
arthyn, line 71: probably should just use `useMultiDms` and `useDms` outside the loop and reference them here

arthyn (changes requested):
just some small changes

arthyn (approved):
lgtm

PR #2925, opened 2023-10-05
groups: disable wayfinding block on mobile

dnbrwstr:
Currently, the wayfinding block on mobile:

- Overlaps with the bottom nav
- Has stale content
- Duplicates content in settings + discover tabs

We could fix some of these things, but I don't think we lose much by just removing it for now.


patosullivan (approved):
It's time

PR #2928, opened 2023-10-06
profile: add link to manage account

latter-bolden:
OTT

Fixes LAND-1018


patosullivan (approved):

jamesacklin (changes requested):
This only applies to hosted customers, so let's wrap this sidebaritem/link in a isHosted() check.

jamesacklin (approved):

PR #2934, opened 2023-10-10
scroller: refactor prefix logic and improve perf

dnbrwstr:
This PR refactors the `prefixElement` feature in the `ChatScroller`. Before, the prefix element (the element at the very top of the list once all items are loaded, for example the DM header block) was inserted as part of the first item in the list. Now, it's treated as an independent item, and is inserted once all items are loaded. This makes rendering more consistent and easier to estimate/memoize.

In doing this, I uncovered a pretty significant performance regression: messages were being reprocessed on every single scroller render. I fixed that, which should significantly improve the performance of the scroller.

Fixes LAND-1037


patosullivan (approved):
Looks great, thank you!

PR #2935, opened 2023-10-10
scroller: fix loading newer messages after jump

dnbrwstr:
Fixes message loading after jumping to a new scroll position by removing `hasEverLoadedNewest` hack in `useScrollerMessages`. The need for `hasEverLoadedNewest` should have been removed by improvements to `hasLoadedNewest` statefulness in the writ window, but we should keep an eye on this to see whether we're getting unnecessary loads.

This pr also adds a nice lil debug helper: `useObjectChangeLogging`, in `utils.ts`. It'll log a message when any property of the object you pass changes. Useful for tracking down excessive re-rendering.

Fixes LAND-1035.

### DIFF
--- a/ui/src/chat/ChatScroller/ChatScroller.tsx
+++ b/ui/src/chat/ChatScroller/ChatScroller.tsx
@@ -28,7 +28,7 @@ import {
   useMessageData,
 } from '@/logic/useScrollerMessages';
 import { useChatState } from '@/state/chat/chat';
-import { createDevLogger } from '@/logic/utils';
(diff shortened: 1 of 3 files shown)


dnbrwstr:
The trick is that it requires a channel with enough unread messages that the first unread message won't be within the first page of results loaded. That means that you need to have 50+ (100+ to be safe) unread messages. 

The repro is basically this:
- Open a channel with a lot of unreads
- Click the banner when it pops up
- Scroll down towards newer messages

On the current live site, newer messages will fail to load in after scroll. In this updated branch, they'll load in correctly.

PR #2937, opened 2023-10-10
groups: improve group list rendering performance

dnbrwstr:
I noticed that when scrolling the main list of groups on both mobile + desktop, we were sending new requests for all group avatars on pretty much each scroll event.

After some investigation, it turned out that this was related to the `Head` component being passed to the `Virtuoso` scroller. Because that component was being dynamically generated, every time its children changed the whole tree under it was getting wiped and recreated, rather than just updated as we'd really want. 

In order to prevent recreating all the list items, I replumbed the list to display child content as an element in the list, rather than using the header component prop. It would be a little more elegant to just pass all the individual pinned groups as list items and render all items + section headers in `itemContent`, but I kept it simple for this pass.

In investigating this, I added a number of `useMemo` and `useCallback`s to reduce rerenders. They should be strictly better for performance + not have side effects, so I've left them in. 

Fixes LAND-1038.

### DIFF
--- a/ui/src/components/Sidebar/GroupList.tsx
+++ b/ui/src/components/Sidebar/GroupList.tsx
@@ -1,16 +1,32 @@
-import React, { useEffect, useMemo, useRef } from 'react';
+import React, { ReactNode, useEffect, useMemo, useRef } from 'react';
 import { StateSnapshot, Virtuoso, VirtuosoHandle } from 'react-virtuoso';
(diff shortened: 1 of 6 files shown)


jamesacklin (approved):
Scrolled through and watched the image requests in the Network tab of dev tools with cache enabled (normal use-case). None of the images loaded in duplicate. Good work!

arthyn:
ahhhh this makes sense with what I was seeing, ty!

PR #2941, opened 2023-10-10
mobile: fix stuck chat input focus state on navigate away

patosullivan:
We weren't properly clearing the chat input focus context on unmount of the chat input component, this makes sure the chat input context is properly reset if the component unmounts.

Fixes LAND-1032


arthyn (approved):
lgtm

PR #2942, opened 2023-10-10
scroller: initial fix for short scrollers

dnbrwstr:
Working pr for fixing LAND-1040 and LAND-1041. The initial commit should fix most issues, but I'm still testing, and want to dig into the `hasLoadedNewest` flag, since that should probably have prevented these issues.

### DIFF
--- a/ui/src/chat/ChatScroller/ChatScroller.tsx
+++ b/ui/src/chat/ChatScroller/ChatScroller.tsx
@@ -148,7 +148,7 @@ const thresholds = {
 
 const loaderPadding = {
   top: 40,
-  bottom: 50,
+  bottom: 0,
 };
 
 export interface ChatScrollerProps {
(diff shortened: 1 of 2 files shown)


jamesacklin:
Confirmed fixes 1040

jamesacklin:
Confirmed fixes 1041

dnbrwstr:
Also add a visible loader when loading at the top + fixed an issue with scroll tracking.

patosullivan:
1041 still seems to be present:

https://github.com/tloncorp/landscape-apps/assets/1221094/2d4c9155-982c-4170-b8a4-70754eb04a9e

dnbrwstr:
Ya, it's still present in the cases when you scroll up then down, but it should be fixed for the base case (just loading the page). I think the up and down issue has always been there? But I can work on a fix separately.

jamesacklin:
I think this eases enough pain we can merge as-is and make a secondary case for scroll-up-down-then-post.

PR #2944, opened 2023-10-11
lure: provide a fallback url if branch.io is blocked (or down)

patosullivan:
Fixes LAND-1047

Apparently branch.io appears on many adblock lists. (see https://github.com/BranchMetrics/web-branch-deep-linking-attribution/issues/538)

We were only returning a lure link if we could successfully create a deep link. We should return the normal/non-branch lure link if we fail to contact branch.io.

### DIFF
--- a/ui/src/logic/branch.ts
+++ b/ui/src/logic/branch.ts
@@ -20,7 +20,7 @@ export const getDeepLink = async (alias: string) => {
 
 export const createDeepLink = async (canonicalUrl: string, lure: string) => {
   const alias = lure.replace('~', '').replace('/', '-');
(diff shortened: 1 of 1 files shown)


jamesacklin (approved):
Makes perfect sense, thank you

alecananian:
The share URL is falling back to the original URL when the deep link one is `undefined`: https://github.com/tloncorp/landscape-apps/blob/develop/ui/src/state/lure/lure.ts#L273

Could be enough to just return `undefined` from the catch? Either way you'd get the same functionality in the end, so not a major issue. Just thought I'd mention it

PR #2945, opened 2023-10-11
lure: use svg for qr code rather than data url and img

patosullivan:
Fixes LAND-1045 by using an SVG rather than an image with a data url for the QR codes (data urls are blocked by tlon hosting's CSP).


jamesacklin (approved):
Screenshot 2023-10-11 at 2 59 03 PM

What a nice, crisp SVG.

PR #2947, opened 2023-10-11
scroller: fix bottom anchoring for new messages

dnbrwstr:
This fixes bottom anchoring of the chat scroller by ensuring that it doesn't flip into "newer" loading mode when a new message arrives.

### DIFF
--- a/ui/src/state/chat/writs.ts
+++ b/ui/src/state/chat/writs.ts
@@ -165,7 +165,7 @@ export function writsReducer(whom: string) {
         {
           oldest: time,
           newest: time,
-          loadedNewest: false,
+          loadedNewest: true,
           loadedOldest: false,
         },
         draft.writWindows[whom]


ui/src/state/chat/writs.ts
```diff
@@ -165,7 +165,7 @@ export function writsReducer(whom: string) {
         {
           oldest: time,
           newest: time,
-          loadedNewest: false,
```
dnbrwstr, line 168: @arthyn could you confirm that this won't cause issues elsewhere?

My thinking is that if this represents a single, newly created message, the window that corresponds to it should assume that it has the newest message, but not older ones.

dnbrwstr:
There's one more case that needs to be handled -- when you load the page from a message link, start scrolling down from back in the history, and scroll all the way to current messages. That's rarer, but in that case scrolling will still be unanchored atm.

ui/src/state/chat/writs.ts
arthyn: I think true here is right

arthyn (approved):
lgtm

PR #2954, opened 2023-10-13
chat/groups: update banner refactor

latter-bolden:
Our "needs update" logic didn't work on iOS and the update banner wasn't mobile friendly. This changes the banner to instead pop up as a sheet as well as an option in the bottom nav. Also updates the banner on desktop (groups and talk) which fixes the issue where it was pushing the chat input offscreen.

https://github.com/tloncorp/landscape-apps/assets/90741358/3bbb790a-ad9a-486f-bc82-1d5d54495fe2

https://github.com/tloncorp/landscape-apps/assets/90741358/e2ec7dc2-df29-49b4-974c-b83381584b48

The new logic for determining if an update is needed uses our existing docket subscription to listen for changes to the `%groups` version. We'll need to make sure we're consistent about bumping that for any main net releases cc @jamesacklin 

@arthyn want to explicitly mention this removes the old update component which housed our manual service worker code. I believe after removing it, the vite pwa library we're using will handle that automatically, but let me know if that sounds off base.

Fixes LAND-995


jamesacklin:
How do docket version bumps and updating the front-end pertain to our release candidates? Usually we tag `v4.20.0-rc1`, `-rc2`, etc in git, while the docket gets a singular `4 20 0`.

latter-bolden:
good point! I'll update it to check the pike hash instead docket version. That should work across candidate and mainnet releases.

ui/src/groups/GroupSidebar/GroupSidebar.tsx
```diff
@@ -95,9 +98,16 @@ function GroupHeader() {
...
         to=".."
-        className="h-6-w-6 absolute top-2.5 left-2 z-40 flex items-center justify-center rounded bg-white bg-transparent p-1 text-gray-400"
+        className={cn(
+          'h-6-w-6 absolute top-2.5 left-2 z-40 flex items-center justify-center rounded bg-white bg-transparent p-1 text-gray-400',
+          needsUpdate ? 'bg-yellow' : 'bg-transparent'
+        )}
       >
-        <CaretLeft16Icon className="h-4 w-4" />
+        {needsUpdate ? (
+          <AsteriskIcon className="h-4 w-4 text-black dark:text-white" />
```
arthyn, line 107: will this look right on dark with yellow + white?

arthyn: oh nvm ignore, I forget they're flipped

latter-bolden: `text-black`/`text-white` have inverted meaning when theme is dark. Counterintuitively, this forces it to always render as black, not sure if there's a cleaner way to do it

arthyn (commented):
This looks good, should be ready once you get the pike check in

latter-bolden:
@arthyn updated it to check the pike hash on an interval. 

Also noticed an old commit message from Pat mentioning some limitations of using the auto update for our service worker, so I added that functionality back in as well.

arthyn (approved):
lgtm with suggestion

PR #2955, opened 2023-10-13
diary: dont crash in update path

arthyn:
This should fix LAND-1061 by safely accessing notes when we try to make notifications. I'm not sure however of the cause of why one might not have the note being commented on, but seems highly possible with funny permissions.

Also put a little rider in here to fix the notification dot color.


patosullivan (approved):
lgtm

PR #2959, opened 2023-10-16
chat: message focusing

latter-bolden:
Changes our chat message longpress action to a "focus view" that closer resembles other native experiences. Works with all kinds of messages (text, image, tweet, reference, etc.). Usable from chat channels and DMs.

The animations need a design pass and there are still a handful of bugs I'm chasing down, but the bulk is there.

Fixes LAND-1054


latter-bolden:
This will need to be reworked a bit with all the changes that have landed since. Keeping the branch, but closing the PR for now.

PR #2960, opened 2023-10-17
New DM flow fixes

dnbrwstr:
- Fixes issue where creating a new dm would appear to get stuck loading indefinitely
- Fixes an issue that caused leaving a dm to redirect to the groups home tab
- Modifies new DM interface -- removes "Connect" action, making actually sending the message the primary action, and removes "Cancel" on mobile, as the back arrow in the header makes it redundant. 

Fixes LAND-1070
Fixes LAND-1071

### DIFF
--- a/ui/src/dms/Dm.tsx
+++ b/ui/src/dms/Dm.tsx
@@ -161,7 +161,7 @@ export default function Dm() {
           isSelecting ? (
             <>
               {isMobile && (
-                <MobileHeader title="New Message" pathBack="/messages" />
+                <MobileHeader title="New Message" pathBack={dmListPath} />
               )}
               <MessageSelector />
             </>
--- a/ui/src/dms/Dm.tsx
+++ b/ui/src/dms/Dm.tsx
@@ -159,7 +159,12 @@ export default function Dm() {
         className="padding-bottom-transition flex-1"
         header={
           isSelecting ? (
-            <MessageSelector />
+            <>
+              {isMobile && (
+                <MobileHeader title="New Message" pathBack="/messages" />
+              )}
+              <MessageSelector />
+            </>
           ) : (
             <Routes>
               <Route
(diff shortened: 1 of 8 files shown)


ui/src/dms/NewDm.tsx
```diff
@@ -4,25 +4,30 @@ import ChatInput from '@/chat/ChatInput/ChatInput';
...
     <Layout
       className="flex-1"
+      style={{
+        paddingBottom: shouldApplyPaddingBottom ? 50 : 0,
+      }}
       header={
-        isNativeApp() && (
-          <MobileHeader title="New Message" pathBack="/messages" />
-        )
+        isMobile && <MobileHeader title="New Message" pathBack="/messages" />
```
arthyn, line 30:
```suggestion
        isMobile && <MobileHeader title="New Message" pathBack={dmListPath} />
```

will have to import it

ui/src/dms/MultiDm.tsx
```diff
@@ -110,7 +122,12 @@ export default function MultiDm() {
         className="padding-bottom-transition flex-1"
         header={
           isSelecting ? (
-            <MessageSelector />
+            <>
+              {isMobile && (
+                <MobileHeader title="New Message" pathBack="/messages" />
```
arthyn, line 127: `dmListPath` instead of `/messages`

ui/src/dms/Dm.tsx
```diff
@@ -146,7 +159,12 @@ export default function Dm() {
         className="padding-bottom-transition flex-1"
         header={
           isSelecting ? (
-            <MessageSelector />
+            <>
+              {isMobile && (
+                <MobileHeader title="New Message" pathBack="/messages" />
```
arthyn, line 164: `dmListPath` instead of `/messages`

arthyn (changes requested):
just a few things, but looks good otherwise

arthyn (approved):
lgtm

PR #2961, opened 2023-10-17
skip branch deep linking if no lure web link provided

alecananian:
Skips the branch deep link look-up and creation flow if the specified `canonicalUrl` (the web lure flow URL) is empty. This prevents a situation where a deep link is created without a custom desktop redirect URL, resulting in the user being able to land on a Branch-hosted app download page.


arthyn (approved):
lgtm

PR #2966, opened 2023-10-21
notebook: tasks bug and misc

arthyn:
This fixes LAND-1075 LAND-1076 and LAND-1077. Tasks were missing from the incoming json parser so that was added in. We also missed a place where we let people try to edit a pending note. The biggest chunk here was letting notes actually error out so that people can at least save a backup somewhere else (or change the content), previously we were just navigating away.


patosullivan (approved):
lgtm

PR #2968, opened 2023-10-23
groups: fix icon position/size for mobile nav

latter-bolden:
OTT

Also: 
- updates the "Discover" icon to match the Figma mock 
- fixes touch targets for Discover & Profile (were unintentionally smaller than the others)

Screenshot 2023-10-23 at 3 44 00 AM

Fixes LAND-1074


latter-bolden:
Fixes LAND-1074

PR #2969, opened 2023-10-23
channels: don't allow multiple form submissions in new channel form

patosullivan:
fixes LAND-1051

We don't actually get a status change from the add channel mutation unless we're adding a channel to a zone. Rather than check statuses on each type of mutation here, we can just use react-hook-form's form state (and not allow the user to click the button while we're submitting).


arthyn (approved):
lgtm

PR #2970, opened 2023-10-23
storage: don't break existing functionality if landscape not up to date

arthyn:
This fixes LAND-1096 by making sure we default the new values even if Landscape doesn't send them to us. This ensures that uploads still work until Landscape is updated. This also clears credentials from being stored in local storage.


jamesacklin:
@arthyn Any blockers? Currently broken in testing.

PR #2971, opened 2023-10-23
images: compress images (and strip exif data) prior to upload

patosullivan:
fixes LAND-1092

I added a new library called `browser-image-compression` that uses web workers to compress images in a performant way. It also removes all exif data by default.

To test just upload a large image with exif data, then download the image after uploading/posting it and notice that it's much smaller and doesn't have exif data.

The library: https://github.com/Donaldcwl/browser-image-compression

### DIFF
--- a/ui/src/state/storage/upload.ts
+++ b/ui/src/state/storage/upload.ts
@@ -107,7 +107,6 @@ export const useFileStore = create<FileStore>((set, get) => ({
 
     const compressionOptions = {
       maxSizeMB: 1,
-      maxWidthOrHeight: 1920,
       useWebWorker: true,
     };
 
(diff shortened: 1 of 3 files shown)


ui/src/state/storage/upload.ts
```diff
@@ -92,6 +105,23 @@ export const useFileStore = create<FileStore>((set, get) => ({
     const { key, file } = upload;
     updateStatus(uploader, key, 'loading');
 
+    const compressionOptions = {
+      maxSizeMB: 1,
+      maxWidthOrHeight: 1920,
```
arthyn, line 110: will this mess with the sizes we record from the input?

patosullivan: Good point. I think it would if the image is larger than 1920x1920. Looking into it more now.

patosullivan: Looks like setting this value is unnecessary, I'm going to remove it for now.

arthyn (approved):
lgtm

PR #2973, opened 2023-10-23
storage: hardcode s3 url

arthyn:
We're going to add a hardcoded URL so that we don't have to poke all ships with the presigned URL endpoint. This makes it so that any hosted ship which has not set S3 credentials gets defaulted to the presigned flow. Will mimic these changes in Landscape.

This also improves hosting checks so that we can specifically target staging vs prod based on the vite ship url you pass and persists show/hide dev tools.


patosullivan (approved):
lgtm

PR #2977, opened 2023-10-23
group invite: fix @p input layout issues on mobile

latter-bolden:
Hides the rest of the modal when the input is focused and positions the menu bar below the input. Workable right now, but still flickers.

Fixes LAND-1098


latter-bolden:
This was rushed and incomplete. Should be addressed properly, closing for now.

PR #2979, opened 2023-10-24
hide invite people button until group has loaded

alecananian:
- Fixes "Invite People" button (and blank invite modal) showing on pending/loading groups sidebar and dropdown actions

Screenshot 2023-10-23 at 9 15 14 PM


latter-bolden (approved):

jamesacklin (approved):
Undefined until we get it makes sense.

PR #2981, opened 2023-10-24
deeplinks: defend against Branch link poisoning 

latter-bolden:
If branch links are initially malformed when created, they can never be automatically updated. We make no attempt to do so right now, and changing our logic to fix them would need to happen elsewhere since it requires the Branch private key.

This means it's really important not to create one with bad data. This PR adds checks to `createDeepLink` to ensure it has a valid `lure` or `wer` link before attempting creation. It also updates the lure logic to avoid calling `createDeepLink` altogether unless we have the requisite info to do so.

If your reviewing this, please confirm you can still create both types locally.

Fixes LAND-1106

### DIFF
--- a/ui/src/logic/branch.ts
+++ b/ui/src/logic/branch.ts
@@ -1,3 +1,6 @@
+import { isValidPatp } from 'urbit-ob';
+import { whomIsFlag } from './utils';
+
 const fetchBranchApi = async (path: string, init?: RequestInit) =>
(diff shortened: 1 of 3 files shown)


alecananian (approved):
Stronk 💪

patosullivan (approved):
Looks good, works locally (after disablying ublock origin in my browser).

PR #2984, opened 2023-10-24
QR codes: handle non success states

latter-bolden:
OTT

LAND-1104


latter-bolden:
Fixes LAND-1104

arthyn (approved):
lgtm!

PR #2991, opened 2023-10-26
images: flip usewebworker back on

patosullivan:
Since we determined the web worker wasn't the issue (it was the CSP changes on the hosting platform not playing nicely with Safari) we ought to go back to using the web worker for better performance on image uploads.


jamesacklin (approved):
Fine by me! Let me retag.

PR #2993, opened 2023-11-02
mobile: hide nav on gallery item and diary note comment input focus

patosullivan:
Fixes LAND-1118. Looks like the issue with the nav still appearing on a new DM was already fixed, but I removed it from the comment input on gallery items and diary notes.


jamesacklin (approved):
Looks good to me, stays on-pattern.

PR #2995, opened 2023-11-06
images: fix drag and drop in chat threads

patosullivan:
Fixes LAND-1110. We weren't using the currentTargetId in the handleDropWithTarget handler, we were just using whichever targetId we received in the parameters for the hook itself.


latter-bolden (approved):
Tested locally, looks good

PR #2997, opened 2023-11-08
LAND-1138 Fix e2e tests (update fakes tars and outdated locators)

mrozanski:
Tests were not running because the tar files with the fake ships images had the same groups and talk hashes as after committing desks, which we use to verify if the ships are ready to start testing.
Fixed by uploading new tars.

This commit also updates a couple of locators.


patosullivan (approved):
Thank you!

PR #3002, opened 2023-11-13
groups: visual fix omnibus

jamesacklin:
- Fixes LAND-1160 (group home screen):
  - Adds optional props to the MobileHeader to make the background transparent. Also removes all inline text-color declarations for UI elements in the MobileHeader such that we can pass a color down from its parent context.
  - Adds the group cover color or image behind the MobileHeader and channel list. If there's an image, we blur the image and scale it to cover the header area.
  - Adds a hook to determine a suitably-contrasting color for text and UI elements in the group home MobileHeader based on the group's cover color or image. This is similar to the `fast-average-color` method we tried a year ago, albeit more tunable to adjust for edge-cases.
    - This still doesn't avoid the problem of images with invalid CORS headers, which the hook needs to render the image in a `<canvas>` element to analyze. We default to white UI elements in these cases, which design has deemed an acceptable tradeoff.
  - Traps the channel list in a rounded-corner sheet.
  - Adjusts typography of channel section dividers.
  - Rearranges the group actions.
  - Traps the channel and group action icons in circles and adds right-carets for actions that lead to another screen.
  - Adds smile, plane icons and changes channel type icons. 
- Fixes LAND-775 by adding a platform-specific share action to the "Share with Friends" settings item. This falls back to https://tlon.io, which has app install buttons. This item is not present on non-native app contexts or non-https connections due to the requirements of the clipboard and share APIs.
- Fixes LAND-1159 by changing the icon in the "no Groups" placeholder to the compass.
- Fixes LAND-1080 by adding `-1` to the `tabIndex` prop of the group name. This is bad for Web accessibility, but couldn't figure out any other way to handle the auto-focus behavior.
- Fixes LAND-1143 by renaming the "About Groups" settings item to "About."


jamesacklin:
@arthyn @patosullivan @latter-bolden This is now pointing to the hackweek branch. Review at your leisure.

patosullivan (approved):
Looks good.

latter-bolden:
we should merge this anyhow, but one nitpick:

@jamesacklin `useTextColor` isn't working for me, were you getting CORS issues while developing? I see security errors loading those images and if that succeeds, issues reading a tainted canvas element (both on localhost and from a deployed ship).

This can result in hard to read text/buttons if there is in fact a background image that clashes with the default light mode/dark mode colors.

Screenshot 2023-11-21 at 5 50 12 PM

PR #3006, opened 2023-11-15
hackweek: only show 'go to latest' button when we're not already at the latest

patosullivan:
Fixes LAND-1171

We'll need the same thing on #2996 for DMs/Clubs

### DIFF
--- a/ui/src/chat/ChatWindow.tsx
+++ b/ui/src/chat/ChatWindow.tsx
@@ -153,7 +153,7 @@ export default function ChatWindow({
           isScrolling={isScrolling}
         />
       </div>
-      {scrollTo ? (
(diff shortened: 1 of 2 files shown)


arthyn (commented):
I think you'd still want to go to latest even if there aren't new pages, because you're just really far in the backlog. So maybe it's next page or "too far from bottom"?

patosullivan:
> I think you'd still want to go to latest even if there aren't new pages, because you're just really far in the backlog. So maybe it's next page or "too far from bottom"?

Updated for that case, but also realized that goToLatest wouldn't actually fetch the latest as it was (it would just move the scroller to the bottom, which wouldn't trigger onAtBottom on its own), so I made some more involved changes.

arthyn (approved):
lgtm!

PR #3007, opened 2023-11-15
groups: fix query invalidation loop

latter-bolden:
We have a hack in place to account for the situation where the frontend loses track of a group if it's been removed from the gang list, but is taking a long time to join. The way it's implemented right now means that we're invalidating the groups query too frequently which can trigger loops.

This update just adjusts that groups invalidation logic to only fire if we believe a gang was removed rather than every time the query is fetched.

Fixes LAND-1208


patosullivan (approved):
Lgtm

PR #3009, opened 2023-11-16
analytics: fix url info logged during opt in event

alecananian:
Fixes an issue where the current browser URL was logged to the analytics provider on the opt in event (updated on the identify event as well for good measure)


arthyn (approved):
lgtm

PR #3013, opened 2023-11-16
hackweek: make unreads work with threads/comments

arthyn:
This makes unreads work with threads/comments now that those are stored separately, fixing LAND-1180. We needed to add tracking of unread threads upon hearing new replies so that we could efficiently calculate unreads without going over the entire message log and then pass which threads are unread to the frontend so that we could mimic the current behavior on mainnet. The type changes here unlock the ability to make thread unread state be completely separate from the main window if we want. 

Works for both DMs and channels.

### DIFF
--- a/ui/src/chat/ChatUnreadAlerts.tsx
+++ b/ui/src/chat/ChatUnreadAlerts.tsx
@@ -34,7 +34,7 @@ export default function ChatUnreadAlerts({
   }
 
   const { threads } = unread;
-  const threadKeys = Object.keys(threads).sort();
+  const threadKeys = Object.keys(threads).sort((a, b) => a.localeCompare(b));
 
   const topId = threadKeys[0];
   const to =
--- a/ui/src/chat/ChatUnreadAlerts.tsx
+++ b/ui/src/chat/ChatUnreadAlerts.tsx
@@ -28,11 +28,11 @@ export default function ChatUnreadAlerts({
   }
 
   const { unread } = chatInfo.unread;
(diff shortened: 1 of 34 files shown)


Fang- (approved):
Hoon side looks good to me. (:

ui/src/chat/ChatUnreadAlerts.tsx
```diff
@@ -28,12 +28,19 @@ export default function ChatUnreadAlerts({
...
-  if (unread.count === 0 || !unread['read-id']) {
+  const id = unread['unread-id'];
+  if (unread.count === 0 || !id || typeof id === 'object') {
     return null;
   }
 
-  const id = unread['read-id'];
-  const to = `${root}?msg=${id}`;
+  const { threads } = unread;
+  const threadKeys = Object.keys(threads).sort();
```
patosullivan, line 37: We need to explicitly sort here

ui/src/chat/ChatWindow.tsx
```diff
@@ -70,24 +70,18 @@ export default function ChatWindow({
...
-  useEffect(
-    () => () => {
-      if (readTimeout !== undefined && readTimeout !== 0) {
-        markRead({ nest });
-      }
-    },
-    [readTimeout, markRead, nest]
-  );
+    useChatStore.getState().setCurrent(nest);
+  }, [nest]);
```
patosullivan, line 74: Do we need this other useEffect that was taken out?

patosullivan (approved):
We're good after a few changes

PR #3014, opened 2023-11-17
chat, heap, diary: tighter epic logic

Fang-:
When joining channels for the first time, these agents were failing to keep version negotiation state for their host around properly. Here, we make a couple of changes that help them keep the channel status straight. (`/app/diary` already had many of these changes. Now they're all consistent with each other.)

- During `+*-join`, explicitly initialize the $net as appropriate, instead of always bunting it to be %pub.
- During `+*-join`, do not subscribe to the channel right away, instead send an /epic version negotiation subscription first. Only once versions match do we open the subscription for the channel (in `+*-make-chi`).
- When receiving an `/epic` `%fact`, always call into `+*-take-epic` for all affected channels, regardless of what the state of the new epic is.
- In `+*-take-update`, don't change the version negotiation state, only touch up the load.net, and base that value off of watch ack vs nack.
- In `+load`, re-establish `/epic` subscriptions for all channel hosts. This lets ships in bad states because of the previously-broken logic recover.

For LAND-1195. This does unfortunately mean that we're going to need this big, bulky backend change to go out as an appetizer. On the bright side, it's not a strictly necessary appetizer, it only improves the UX for a subset of the "I'm behind the channel host" cases, and shouldn't affect behavior of the migration itself at all.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -126,6 +126,11 @@
     =.  cor  restore-missing-subs
     =.  cor  (emit %pass ca-area:ca-core:cor %agent [our.bowl dap.bowl] %poke %recheck-all-perms !>(0))
     =.  cor  (emit %pass ca-area:ca-core:cor %agent [our.bowl dap.bowl] %poke %leave-old-channels !>(0))
+    =.  cor
+      %+  roll  ~(tap in `(set @p)`(~(run in ~(key by chats)) head))
+      |=  [=ship cr=_cor]
+      ?:  =(ship our.bowl)  cr
+      (watch-epic:cr ship &)
     ?:  =(okay:c cool)  cor
     :: =?  cor  bad  (emit (keep !>(old)))
     %-  (note:wood %ver leaf/"New Epic" ~)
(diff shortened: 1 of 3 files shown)


Fang-:
The thing this doesn't do, unfortunately, is prompt the frontend to re-load the status for the affected channels. For reasons yet unknown, in the case of a fresh group join, it initially shows the channels as green (usually, depending on timing). Only after a page refresh does it receive the most-up-to-date statusses from the backend and renders the red dots.

I didn't find a way to push the new epic status onto the frontend. Did I just overlook it? If not, _maybe_ this needs to be paired with a frontend change that more aggressively re-checks channel status?

desk/app/chat.hoon
```diff
@@ -1754,12 +1762,17 @@
...
+    =|  =chat:c
+    =.  net.chat
+      ?:  =(our.bowl p.chan.j)  [%pub ~]
+      [%sub p.chan.j | %chi ~]
+    =.  chats  (~(put by chats) chan.j chat)
     =.  ca-core  (ca-abed chan.j)
     =.  last-read.remark.chat  now.bowl
     =.  group.perm.chat  group.j
     =.  cor  (give-brief flag/flag ca-brief)
-    ca-sub
```
arthyn, line 1762: we don't need sub here because eventually we do `ca-make-chi` and that subs?

Fang-: Correct.

PR #3015, opened 2023-11-17
Group cloning for testing refactor migration on large groups

midsum-salrux:
This expands on the existing transfer/import-channel code for archiving by allowing moves between ships

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -1,4 +1,4 @@
-/-  g=groups, ha=hark, h=heap, d=diary, c=chat
+/-  g=groups, ha=hark, h=heap, d=diary, c=chat, cl=clone
 /-  g-one=group
 /-  m-one=metadata-store
 /-  meta
@@ -183,6 +183,10 @@
       %group-cancel
     =+  !<(=flag:g vase)
     ga-abet:ga-cancel:(ga-abed:gang-core flag)
+  ::
+      %group-clone
+    =+  !<(=flag:g vase)
+    ga-abet:ga-clone:(ga-abed:gang-core flag)
   ::
       %invite-decline
     =+  !<(=flag:g vase)
@@ -977,6 +981,7 @@
         [%updates rest=*]  (go-pub rest.pole)
         [%ui ~]            go-core
         [%preview ~]       go-preview
+        [%clone ~]         go-clone
     ::
         [%bait host=@ name=@ ~]
       ?>  ?=(%open -.cordon.group)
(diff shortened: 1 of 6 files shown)


desk/app/groups.hoon
```diff
@@ -1137,9 +1168,36 @@
...
+      ==
+    =.  cor
+      ?-  +>-.clone
+          %chat
+        (emit [%pass /clone %agent [our.bowl %chat] %poke %import-channel !>([[our.bowl group.clone] create log.clone])])
+          %heap
+        ::  TODO implement this poke
+        (emit [%pass /clone %agent [our.bowl %heap] %poke %import-channel !>([[our.bowl group.clone] create log.clone])])
+          %diary
+        ::  TODO implement this poke
```
Fang-, line 1196: These `TODO`s are in fact `DONE`s, right?

desk/app/groups.hoon
```diff
@@ -183,6 +183,10 @@
       %group-cancel
     =+  !<(=flag:g vase)
     ga-abet:ga-cancel:(ga-abed:gang-core flag)
+  ::
+      %group-clone
+    =+  !<(=flag:g vase)
```
Fang-, line 188: This is a bit misleading because it uses a mark that actually exists, but with a different type. Assuming this is for local-only use (in which case we might also want a `?>  =(our src):bowl`), probably fine to rename to `%group-do-clone` or w/e.

desk/app/groups.hoon
```diff
@@ -1012,6 +1017,32 @@
...
+      ^-  (set card)
+      %-  %~  run  in
+      ^-  (set nest:g)
+      %~  key  by  channels.group
+      |=  =nest:g
+      ?+  p.nest  !!
+          %chat
+        =/  =log:c  .^(log:c %gx (welp (channel-scry nest) /logs/noun))
+        =/  =clone:cl  [q.flag q.q.nest [%chat log]]
+        [%give %fact ~[/clone/(scot %p p.q.nest)/[q.q.nest]] group-clone+!>(clone)]
```
Fang-, line 1034: We probably want to only give these facts to the ship that just opened the subscription, instead of all ships that still have this subscription open.

Along the same vein, we probably want to `%kick` the incoming subscription after we give it these facts.

Fang- (commented):
Some comments, but this seems fine overall.

Fang- (approved):
Thanks for the changes, this should suffice!

(tbc, ftr, we don't want to merge & release this broadly, only on a manual per-ship basis)

desk/app/groups.hoon
```diff
@@ -1810,6 +1877,24 @@
...
+          image=''
+          cover=''
+          *cordon:g
+          members=~
+          secret=%.n
+      ==
+    =.  cor
+      (emit %pass /groups/(scot %p our.bowl)/[q.flag]/create %agent [our.bowl %groups] %poke %group-create !>(create))
+    =.  cor
+      (emit %pass /groups/(scot %p p.flag)/[q.flag]/clone %agent [p.flag %groups] %watch /groups/(scot %p p.flag)/[q.flag]/clone)
```
arthyn, line 1895: If we sent this to `our.bowl` instead of `p.flag` would that let us clone groups w/o installing this on the host? seems like it might work

midsum-salrux: this does in fact work

PR #3016, opened 2023-11-17
Mariano / land-1210 landscape e2e testing group tests to run in parallel by

mrozanski:
Addresses LAND-1210

New test scenario

* Adds set of tests to cover non-admin of a group visiting channels (chat, notebook, gallery) and posting (text only).
* Adds some `data-testid` attributes to the code to use by locators
* Combine existing tests in a single file (use files to group tests by scenario)
* Sets workers to 2 for parallel runs

The goal is to leverage parallel runs and make test scenarios not depend on each other.

(Will follow up in PR discussion)


mrozanski:
Opening as draft to verify tests in CI with updated desk tarballs and start discussion about code and tests organization.

mrozanski:
Currently tests are organized by specs and execution is organized by ship.
There are two scenarios in two spec files.
Workers is set to 2.
The assumption is that Playwright will execute the two specs at the same time.

Specs cover different use cases and there are no dependencies between specs. This allows to selectively run one spec and not the others.

`001-spec` and `002-spec` depend on the `authenticate` task in `e2e/auth.setup.ts`

`authenticate`, `001-spec` and `002-spec` depend on the ships being already booted and running

The setup `authenticate` task runs twice, one for each ship

e2e  - Groups

Piers used for the tests (1) already contain a group created by ~zod with one channel of each type, and ~bus is already invited and accepted before archiving the piers. This enables spec 2 to **not depend on previous steps of the e2e process but on preconfigured conditions**.

Tests (2) (3) are configured to be skipped if they're not associated to the current ship running tests.

```
    await runTestForShip('~bus'); // --> (2)
    await runTestForShip('~zod'); // --> (3)
```

Proposed Improvements:

* Bump efficiency by not running the complete spec scrips twice and skipping depending on authenticated ship. Requires confirming if Playwright supports multiple contexts.
* Automate the creation of the piers. This would run if required by new versions of landscape, base desk or urbit,  archived and uploaded to bucket (note that file names would need to change in `rube/index.ts` or moved to separate config file)

I should be able to trigger one spec that involves two ships and completes in one run with either multiple ships authenticated or sequentially if that's not possible, but not starting tests just to skip them.

mrozanski:
Need to look into the last run in Github. Tests are failing and passing inconsistently.

mrozanski:
Adding to the proposed improvements:

* Running the `playwright` command multiple times results in an overwritten report and possibly tests result as reported by Github. I think ideally we want to call it only once and have all options-based behaviors defined using their config or cmd line options (which ships or specs to execute for example)

mrozanski:
@patosullivan @arthyn This is now ready for review.

mrozanski:
Also fixes LAND-1239

patosullivan:
Hey @mrozanski the changes look good, but it looks like we're still seeing a timeout on one of the tests (and it's still doing the thing where that happens and then the check is still reported as successful)

mrozanski:
@patosullivan where do you see that?

JIC if you're looking at the check linked in this PR, I believe it is a case of:

`"flaky" - tests that failed on the first run, but passed when retried;`

https://playwright.dev/docs/test-retries

patosullivan:
Oh yeah, it was just flaky. Ok, this looks good to me.

patosullivan (approved):

PR #3017, opened 2023-11-20
dms: smoother joins & leaves

latter-bolden:
Should fix the issue on the latest hackweek deploy where joining and leaving DMs and multi DMs succeed, but doesn't update the interface.

Fixes LAND-1225

### DIFF
--- a/ui/src/state/chat/chat.ts
+++ b/ui/src/state/chat/chat.ts
@@ -48,6 +48,8 @@ import emptyMultiDm, {
   appendWritToLastPage,
   buildCachedWrit,
   buildWritPage,
+  removePendingFromCache,
+  removeUnreadFromCache,
 } from './utils';
 
 export interface State {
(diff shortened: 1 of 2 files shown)


ui/src/state/chat/chat.ts
```diff
@@ -745,9 +747,8 @@ export function useMarkDmReadMutation() {
   });
 }
 
-const emptyUnreads: DMUnreads = {};
-const dmUnreadsKey = ['dm', 'unreads'];
 export function useDmUnreads() {
+  const dmUnreadsKey = ['dms', 'unreads'];
```
arthyn, line 751: we actually should move these keys up and reuse them everywhere we use the unreads key

arthyn (approved):
makes sense to me

ui/src/state/chat/chat.ts
latter-bolden: agreed. Rather than doing it one off here, I'll open a separate PR to extract keys more consistently.

PR #3019, opened 2023-11-21
version negotiation: make additional info accessible in the hooks

latter-bolden:
Makes `isLoading` and other react query goodies accessible from the `useNegotiation` hook. We need this for DMs where we're initially showing an erroneous mismatch warning. Now it will display normally if the negotiation is still loading.

Fixes LAND-1224


arthyn (approved):
lgtm

PR #3020, opened 2023-11-21
negotiation: insert facts into cache as we hear them

patosullivan:
Also make useNegotiation generic, used only in useNegotiate and useNegotiateMulti, fix issue with incompatibility shown on group join, fix css issue on incompatible message in DMs/MultiDMs.

Fixes LAND-1216.
Fixes LAND-1227.
Fixes LAND-1228


latter-bolden (approved):
looks good!

PR #3022, opened 2023-11-21
dms: fix delivery tracking when a dm is first created

latter-bolden:
OTT

Fixes LAND-1229


patosullivan (approved):
Looks good, worked as expected locally.

PR #3024, opened 2023-11-21
references: fix issue with reference paths causing notebook/gallery comment references to display improperly

patosullivan:
Fixes LAND-1226 by papering over the issue on the frontend in the notebook comments case and fixing a bug in the gallery comments case (we were looking at the wrong position in the segments array).


latter-bolden (approved):
Looking good, worked locally.

PR #3026, opened 2023-11-21
dms: fix loading placeholder

latter-bolden:
We'd moved this into ChatScroller, but the logic was broken. After fixing the logic, there are still display issues with it overlapping content that's set in the parent component. Clawing back the placeholder to it's original location in `DmWindow`.

Fixes LAND-1234


patosullivan (approved):
Works for me

PR #3027, opened 2023-11-21
negotiation: switch to using status from negotiation rather than boolean

patosullivan:
And assume we're waiting and everything is fine until we know otherwise


arthyn (approved):
lgtm

PR #3028, opened 2023-11-22
chat: backwards compatibility with pre-hw dms

Fang-:
Implements least-effort (20-80) backwards compatibility with the old chat backend for dms and clubs. To get there, we make the following changes:

- Rename the new dm- and club- marks to be prefixed with chat- (resulting in chat-dm- and chat-club-), so that the new codebase may distinguish between new and old-style nouns unambiguously.
- Re-includes json-less versions of the old marks.
- Make chat accept & process old-style dm and club pokes. These operate at 100% compatibility. iow, old chat talking to new chat will always work.
- Add conditional retry logic where, if sending a new-style poke fails, we may in some cases retry with an old-style poke instead. Due to dms and clubs not storing a log of events*, this comes with two caveats:
  - For dms, reacts added/removed by new chat will not propagate to old chat instances.
  - For clubs, _only_ messages are propagated from new chat to old chat. This means that club invites, member additions/removals, and metadata changes will be dropped on the floor in "new talking to old" scenarios.

\* In theory, we could jam the entire diff we were trying to send into the wire, so that we can always recover it on-nack, but this seems rather ill-advised...

Includes mark renamings in both the back- and frontend code included in this repo. UX depends on 9e6535c for dms, we may or may not want a similar change for clubs. cc @patosullivan 

This does nothing to handle the poke-nack case in general. Doing _something_ there seems desirable, but it outside the scope of minimum viable backwards compatibility.

For LAND-1231.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -1,4 +1,4 @@
-/-  c=chat, d=channels, g=groups, e=epic
+/-  c=chat, d=channels, g=groups, e=epic, old=chat-2
 /-  meta
 /-  ha=hark
 /-  contacts
@@ -430,7 +430,7 @@
   |^  ^+  cor
   ?+    mark  ~|(bad-poke/mark !!)
   ::
-      %dm-rsvp
+      %chat-dm-rsvp
     =+  !<(=rsvp:dm:c vase)
     di-abet:(di-rsvp:(di-abed:di-core ship.rsvp) ok.rsvp)
   ::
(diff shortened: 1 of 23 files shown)


desk/app/chat.hoon
```diff
@@ -1528,11 +1637,69 @@
...
+          `[%dm-rsvp !>(`rsvp:dm:old`[our.bowl ok])]
+        ::
+            [@ ~]
+          %-  some
+          :-  %dm-diff
+          !>  ^-  diff:dm:old
+          =/  id=time  (slav %ud i.t.wire)
+          :-  [our.bowl id]
+          =/  msg=(unit [=time =writ:c])
+            (get:di-pact our.bowl id)
```
Fang-, line 1676: Because it's a message _we_ are trying to send, the "full" id will always start with `our`, correct?

arthyn: yep

arthyn (approved):
beast mode

PR #3030, opened 2023-11-22
negotiation: host connection should always be matched, shouldn't assume

patosullivan:
We were assuming it was fine to allow DMs to the old world, but we shouldn't do it for group hosts.


latter-bolden (approved):
looks good!

PR #3031, opened 2023-11-22
dms: block new club creation if some of the ships have version mismatch

latter-bolden:
OTT

Fixes LAND-1238


patosullivan (approved):
Looks good.

PR #3035, opened 2023-11-27
chat, heap, diary: migrate chat refs properly

Fang-:
Chat was migrating refs largely correctly, but wasn't accounting for
cases where the referenced message lives outside of the channel where
the reference was posted.

Heap and diary were not migrating refs correctly at all. Since
converting old chat ids into new ones requires chat's lookup maps, we
expose those for old chat state through a new /old scry endpoint in
chat. We cache the scry, to avoid overhead for large/reference-heavy
backlogs.

Diary was not modifying references in top-level posts at all, which we
also correct for here.

All the above fixes LAND-1245. We also tack on two small, unrelated patches.

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -155,7 +155,6 @@
       %channel-migration
     ?>  =(our src):bowl
     =+  !<(new-channels=v-channels:c vase)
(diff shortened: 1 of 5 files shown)


arthyn (approved):
lgtm

PR #3038, opened 2023-11-28
always show author avatar if last post was notice

alecananian:
Updates the `newAuthor` logic to always consider a chat message to be a new author if the last post was a notice type

Screenshot 2023-11-27 at 5 41 17 PM


latter-bolden (approved):
Looks good, worked in testing.

PR #3039, opened 2023-11-28
Land-1253 chat search doesn't work

mrozanski:
- Ensure decToUd() receives a string type arg
- Default to '0' (decToUd returns empty string when a '0' is passed)
- Avoid running query if search term is empty

before:
Screenshot 2023-11-27 at 7 34 53 PM

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -106,6 +106,7 @@ import GroupVolumeDialog from './groups/GroupVolumeDialog';
 import ChannelVolumeDialog from './channels/ChannelVolumeDialog';
 import DMThread from './dms/DMThread';
 import MobileChatSearch from './chat/ChatSearch/MobileChatSearch';
+import MobileDmSearch from './dms/MobileDmSearch';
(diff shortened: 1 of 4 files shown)


ui/src/app.tsx
```diff
@@ -217,10 +218,7 @@ function ChatRoutes({ state, location, isMobile, isSmall }: RoutesProps) {
               />
             )}
             {isMobile && (
-              <Route
-                path=":ship/search/:query?"
-                element={<MobileChatSearch />}
-              />
+              <Route path=":ship/search/:query?" element={<MobileDmSearch />} />
```
arthyn, line 221: Also needs to be changed in `GroupRoutes`

arthyn (changes requested):
LGTM, but needs the same change for Groups app routes

arthyn (approved):
lgtm

latter-bolden (approved):
Working for me on chat and DMs for both mobile and desktop

PR #3040, opened 2023-11-28
chat: fixing duplicate message issues

arthyn:
This should fix LAND-1258. We ensure that we de-dupe messages at all times. We also guard against any fetching while other fetching is happening.


patosullivan (approved):
lgtm

latter-bolden (approved):
Looking good, that seems to be all the references. Fixes the issue testing locally.

PR #3041, opened 2023-11-28
perf: fix cpu pegging issue in chat threads

patosullivan:
fixes LAND-1254 by ~throttling the setChatInputFocus setter in the chat input focus context. Also applied some useMemos along the way and fixed an issue where we were never detecting if we were in a thread in ChatChannels.~ by no-op'ing on chat input state changes on desktop.


arthyn (approved):
lgtm

arthyn (approved):
lgtm

PR #3044, opened 2023-11-29
chat, heap, diary: send migration edits to hosts

Fang-:
During migration, chat identifiers change. As a consequence, references
to chat messages must change if they want to stay pointing to the same
message. However, the old ref alone is not sufficient for constructing
the new ref: you need to know the message being referenced.

Previously, we were doing this transformation during migration strictly
locally. This meant that chat hosts, not uncommonly only aware of the
messages in their own channels, were unable to migrate the chat
references correctly. Newcomers joining the channels would find "bad"
chat references in the backlog, and would be stuck with these forever.

The fix implemented herein depends on the assumption that the author of
a message containing a reference is the one most likely to be able to
resolve the reference successfully. After all, they had already seen the
message previously.

Using this assumption, we further depend on message authors to notify
channel hosts about the updated versions of their own messages. This way
the host can learn of the correctly-migrated reference, and share it
with its subscribers, without needing to chase down the reference itself
(which would require building entirely new pipelines).

The behavior here is as follows: on a per-host basis, when a subscriber
learns about a version match with the host (crucially guaranteeing that
the host has already migrated itself), it goes through its pre-migration
content again, looking for posts/replies authored by itself, containing
references to chat messages. We re-migrate those messages, but instead
of injecting them into the local %channels agent directly, we send an
action, signalling an edit of the original message (in chat's case, with
the new id), that must be sent to the channel host. This then goes on to
resolve as normal, as if the user had manually edited the content
through the ui.

The changes in 56569eb enable support for editing replies. This is a required part of this solution, because replies may contain references also. Previously, only top-level messages could be edited in-place.

Obvious caveat here is that this depends on message authors being online
and running the migration. An acceptable "risk" for p2p software.

Less obvious is the potential network load introduced by this. Each
affected message results in an individual edit, which the host must echo
out to all its subscribers. We don't expect gigantic numbers here, but
it will be a little stress test nonetheless. There are options for
spreading the load slightly, but unclear as of yet whether the
complexity required for that is a worthwhile tradeoff. Investigation
into the numbers on this to follow.

Logic herein untested at this time, further patches might follow.

For LAND-1261.

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -7,7 +7,7 @@
 /+  default-agent, verb, dbug, neg=negotiate
 ::
 %-  %-  agent:neg
-    [| [~.channels^%0 ~ ~] ~]
+    [| [~.channels^%1 ~ ~] ~]
 %-  agent:dbug
 %+  verb  |
 ::
@@ -16,7 +16,7 @@
   |%
   +$  card  card:agent:gall
   +$  current-state
-    $:  %1
+    $:  %2
         =v-channels:c
     ==
   --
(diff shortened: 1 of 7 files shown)


mrozanski:
fixes LAND-1261

arthyn (approved):
LGTMMMM

PR #3045, opened 2023-11-29
Fix some scroller issues

dnbrwstr:
Remove merge artifact which caused some issues with scroller.

Also adds optional debug overlay for scroller.


patosullivan:
Just confirmed that this fixes LAND-1268

PR #3047, opened 2023-11-29
emojis: update posts in cache within onMutate, show emoji reactions optimistically

patosullivan:
fixes LAND-1270 by updating the post within the pagemap in the cache. Previously we were just updating the cache if it existed (via the usePost query).

Follows the pattern of doing cache updates directly in the onMutate of the mutation.


latter-bolden (approved):
Updater code looks good. Worked when tested.

PR #3048, opened 2023-11-29
posts: prevent actions on undelivered posts

latter-bolden:
Hides actions (threading, copying, replying, etc.) on posts which aren't yet delivered. Since we're updating all post types optimistically right now, this PR takes a totalizing blunt approach rather than targeting specific problematic interactions.

 In the common case, the change shouldn't be noticeable for end users. If performance issues cause delayed delivery, this will defend against any unexpected hiccups interacting with manually cached posts.

Fixes LAND-1264


patosullivan (approved):
Looks good

PR #3052, opened 2023-11-30
notebook: fix issue with creating notes with refs

patosullivan:
Fixes LAND-1287 by adding 'cite' to the list of types we consider blocks in constructStory.


arthyn (approved):
lgtm

PR #3053, opened 2023-11-30
scroller: fix jumping in threads

dnbrwstr:
This PR fixes scroller jumps inside threads. The root issue was that `hasLoadedNewest` and `hasLoadedOldest` didn't match reality in thread contexts. While fixing that, I realized it'd be simpler to track load direction by just passing loading parameters in directly and updating state when they change. That way, we only invert the scroller if we're *actually* loading a new page. 

Fixes LAND-1295


arthyn (approved):
lgtm, didn't test yet

PR #3056, opened 2023-12-01
chat: only allow "safe" URL protocols in links

darighost:
Currently, users can post links with `javascript:` URLs. Although links open in a new tab by default, there are several tricky ways to turn this into XSS. Viz,

Screenshot 2023-12-01 at 0 58 39

This change borrows a [regex from dompurify](https://github.com/cure53/DOMPurify/blob/main/src/regexp.js#L9-L11) to only allow safe protocols in links. "Bad" protocols are simple replaced with a blank link. Even when Tlon implements a stronger CSP, it's worthwhile to prevent `javascript:` links, because some users run Landscape apps outside of Tlon (eg red horizon). They should be protected, even if their hosting environment lacks a strong CSP.

Really this PR is a chance for me to practice interacting with this repo, so I invite any criticism you may have. I'm excited to tackle bigger bugs, thanks!

### DIFF
--- a/ui/src/chat/ChatEmbedContent/ChatEmbedContent.tsx
+++ b/ui/src/chat/ChatEmbedContent/ChatEmbedContent.tsx
@@ -49,6 +49,14 @@ function ChatEmbedContent({
   }, [isError, error]);
 
   if (url !== content) {
+    // secure URL protocol regex borrowed from dompurify:
+    // https://github.com/cure53/DOMPurify/blob/main/src/regexp.js#L9-L11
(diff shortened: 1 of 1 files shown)


mrozanski:
Thank you. Labeling as not merge just because we're in the middle of a delicate update and only merging changes related to the upcoming release until it ships.

patosullivan (approved):
This looks good, thank you!

latter-bolden (approved):
Wasn't able to uncover any issues testing various kinds of links. 

@mrozanski I think this can be merged

PR #3058, opened 2023-12-01
scroller: fix issue that would cause chat input to lose focus after sending while in the scrollback

patosullivan:
Fixes LAND-1283 by ignoring negative adjustments (always seems to be -1) from the virtualizer in certain cases (such as a message with replies being pushed into view as new messages are received), which would cause a visually imperceptible scroll that would then trigger the chat input to lose focus.

Also: adds some changes for showing chat scroller debug overlay by default when in dev mode by using useShowDevTools().


patosullivan:
Tagged 1282 by mistake, should just be 1283.

arthyn (approved):
lgtm

PR #3059, opened 2023-12-01
refs: show user when reference content is unavailable

patosullivan:
fixes LAND-1262 by showing the UnavailableReference component when we get an error on fetching the referenced post.

Looks like this:

image


arthyn (approved):
lgtm

PR #3060, opened 2023-12-01
chat: fix duplicate messages in threads

patosullivan:
Fixes LAND-1299 by checking for and removing cached messages when we hear a fact about a new message that matches a cached message. We were also unnecessarily invalidating the usePost query for the parent, so I removed that.

I also found that we were still passing a channel flag into the whom prop in the ChatScroller in both ChatWindow and ChatThread, so I fixed that (and the things downstream from it).

### DIFF
--- a/ui/src/chat/ChatThread/ChatThread.tsx
+++ b/ui/src/chat/ChatThread/ChatThread.tsx
@@ -191,7 +191,7 @@ export default function ChatThread() {
           <ChatScroller
             key={idTime}
             messages={orderedReplies || []}
-            whom={flag}
+            whom={nest}
             isLoadingOlder={false}
             isLoadingNewer={false}
             scrollerRef={scrollerRef}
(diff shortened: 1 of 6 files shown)


ui/src/chat/ChatThread/ChatThread.tsx
```diff
@@ -191,7 +191,7 @@ export default function ChatThread() {
           <ChatScroller
             key={idTime}
             messages={orderedReplies || []}
-            whom={flag}
```
arthyn, line 194: I had intentionally changed these back to whom/flags because we were using them inconsistently. chatinfo and other places expect flags not nests, curious where we expect whoms to be nests

patosullivan: In the markRead call in ReplyMessage

arthyn:
This will fix the issue @zalberico just mentioned about always seeing the same point in a thread unread.

arthyn (approved):
lgtm

PR #3061, opened 2023-12-01
sidebars: better defaulting, moving constants around

arthyn:
Fixes LAND-1251 by improving defaulting behavior here and makes constants more clear.


patosullivan (approved):
Looks good

PR #3062, opened 2023-12-01
chat: use correct ship for notice

arthyn:
Fixes LAND-1282 by adjusting which ship is in the noticed based on who the join is coming from


patosullivan (approved):
Lgtm

PR #3063, opened 2023-12-01
dms: fix missing messages after getting an unread banner

latter-bolden:
Setting up the subscription was failing if you were in a chat that you'd just initiated. This updates the logic to wait until we're sure the agent is ready to accept the sub.

Fixes LAND-1306


arthyn (approved):
lgtm

PR #3064, opened 2023-12-02
fix notebook comment refs copying and replying

alecananian:
- Fixes issue with copying notebook comment refs (`chat/` was hard-coded in the `nest` value)
- Fixes `cite` path when making a notebook/gallery comment reply to another notebook/gallery comment

### DIFF
--- a/ui/src/replies/ReplyMessageOptions.tsx
+++ b/ui/src/replies/ReplyMessageOptions.tsx
@@ -47,11 +47,13 @@ export default function ReplyMessageOptions(props: {
 }) {
   const { open, onOpenChange, whom, reply, openReactionDetails, showReply } =
     props;
-  const nest = `chat/${whom}`;
   const { seal, memo } = reply ?? emptyReply;
+  const threadParentId = seal['parent-id'];
+  const isParent = threadParentId === seal.id;
+  const isDMorMultiDM = useIsDmOrMultiDm(whom);
+  const nest = isDMorMultiDM || isParent ? `chat/${whom}` : whom;
   const groupFlag = useRouteGroup();
   const isAdmin = useAmAdmin(groupFlag);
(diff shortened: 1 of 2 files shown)


ui/src/replies/ReplyMessageOptions.tsx
```diff
@@ -47,11 +47,13 @@ export default function ReplyMessageOptions(props: {
 }) {
   const { open, onOpenChange, whom, reply, openReactionDetails, showReply } =
     props;
-  const nest = `chat/${whom}`;
   const { seal, memo } = reply ?? emptyReply;
+  const threadParentId = seal['parent-id'];
+  const isParent = threadParentId === seal.id;
+  const isDMorMultiDM = useIsDmOrMultiDm(whom);
+  const nest = isDMorMultiDM || isParent ? `chat/${whom}` : whom;
```
patosullivan, line 54: I think we need the `nest` to have "chat" prepended on `whom` in every case where we know this is a chat message, because we just pass the channelFlag for chat messages (because of the way unreads work). @arthyn recently worked on this and could speak to it better than I can.

alecananian: That was my intention with `isDMorMultiDM || isParent`, but if there's a better way to determine this is a chat message, let me know! I can also add a new prop that gets set when the component is rendered in the chat routes.

mrozanski:
To do
Review again.
Ensure the Linear ticket is the correct one.

ui/src/replies/ReplyMessageOptions.tsx
patosullivan: I think if we revert your changes to this file then this PR is good to go.

alecananian:
I've reverted changes other than the `cite` path fix

patosullivan (approved):

PR #3065, opened 2023-12-02
dms: fix new multidm negotiation blocking

latter-bolden:
Previously, you could not set up a new multi DM if you hadn't already messaged with the ships in question because they wouldn't have negotiation status and we're blocking the input if we don't have a match.

This PR updates the backend to allow a poke that initiates negotiation and updates the frontend to send that poke when attempting to create a new multi dm. The UI will indicate that it's loading while that negotiation happens and inform the user if they can't create the DM because of a version mismatch.

Fixes LAND-1307

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -427,6 +427,9 @@
   |=  [=mark =vase]
   |^  ^+  cor
   ?+    mark  ~|(bad-poke/mark !!)
+      %chat-negotiate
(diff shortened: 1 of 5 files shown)


latter-bolden:
@Fang-  we'll still have an issue here if negotiation has a bug where status doesn't update after a ship goes through migration. In that case, the UI will block since from its point of view the ship is still behind.

patosullivan (approved):
looks good!

PR #3066, opened 2023-12-02
sort: need better defaulting since we switched from override

arthyn:
FIxes LAND-1311 by adding more correct defaulting since the old "default" was more like "override"


patosullivan (approved):
Looks good.

PR #3068, opened 2023-12-04
notifications: make sure user is routed to appropriate path from a reply notification

patosullivan:
This fixes LAND-1271, which was an issue where clicking on a reply notification would route the user to either a blank page (for notebooks and galleries) or the parent chat channel (without opening the thread pane).

The backend was sending a generic path that looks like this:
`/post/{postId}/{replyId}`

This was a problem for the frontend for two reasons:
1) The router does not expect anything at /post/ in any channel.
2) The router also doesn't expect anything after the {postId} token.

This PR updates the backend to send a path that includes the appropriate post type in the path (e.g., `note`, `curio`, `message`) for reply notifications.

It also updates the router to allow an optional param after the {postId}, which I called `idReplyTime`. We currently don't use it on the frontend, but we could use it in the future, so I decided to leave that in the path on the backend.

It also adds a backcompat function in Notification.tsx to fix the routes for any previously sent reply notifications for %channels.

It also fixes an issue with ReplyMessageOptions that can occur with replies to notes or gallery items. We were hardcoding chat (because we pass the channel flag as `whom` to the chat scroller in the ChatThread and ChatWindow components) and this was breaking for gallery item and note replies.

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -1093,7 +1093,11 @@
       =;  cs=(unit (list content:ha))
         ?~  cs  ca-core
         =/  =path
-          /post/(rsh 4 (scot %ui id-post))/(rsh 4 (scot %ui id.reply))
+          ?-    -.kind-data.post
+            %diary  /note/(rsh 4 (scot %ui id-post))/(rsh 4 (scot %ui id.reply))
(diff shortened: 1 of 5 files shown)


arthyn (approved):
lgtm

PR #3071, opened 2023-12-04
channels: allow admins to delete posts

patosullivan:
We were only allowing a post to be deleted if the ship attempting to delete it was the author, we need to allow admins to delete posts too.

Fixes LAND-1317


latter-bolden:
Fixes LAND-1317

Fang- (approved):
Simple enough change, let's get it!

PR #3072, opened 2023-12-04
channels: update cache when we hear a fact about showing/hiding a post

patosullivan:
We were ignoring updates about showing/hiding posts, and we were also failing on hearing an update that didn't match the ChannelsResponse type.

Fixes LAND-1321


mrozanski (approved):
Tested locally and it works.

latter-bolden (approved):

PR #3080, opened 2023-12-05
emojis: fix issues with adding emojis in chat

patosullivan:
This fixes LAND-1309.

There were several issues:

1) ReplyReaction's call to addReplyFeel was broken due to a bad nest. This caused adding +1 reactions to fail.
2) We were calling onSettled in useAddPostReact, which would cause an unnecessary re-scry which could temporarily make it look as if your reaction was not successful in the main chat window.
3) We weren't appropriately highlighting *your* reaction, which may make it look like the reaction didn't go through.


latter-bolden (approved):
Code looks good, was working well.

Added a small change to remove `onSettled` for del react as well (was still getting the flicker on remove) and assign the author correctly for cached reacts on single posts (could not remove if cached previously)

PR #3082, opened 2023-12-05
dms: fix failure to update invalidate function

dnbrwstr:
Previously, the dm query invalidate function wasn't updating when the query key changed, which could cause issues in the UI. This PR updates `invalidate` to use `useCallback` instead of `useRef`, fixing those issues.


arthyn (approved):
lgtm

latter-bolden (approved):

PR #3086, opened 2023-12-05
join modal: don't automatically navigate after join if in Talk

latter-bolden:
When you join a group through a reference in Talk, it navigates to a blank page since the path only works in Groups. 

This updates it to not navigate automatically. Instead the group modal will stop spinning and show a "Go" button which navigates correctly.

Fixes LAND-1320


patosullivan (approved):
looks good

PR #3088, opened 2023-12-05
Revert "Revert "talk-cli: import chat structures v2 always""

Fang-:
We have to go deeper.

This is #3075 re-applied. /sur/chat-2 already ships with the talk desk.

Talk-cli is still gonna be broken/useless for the new backend, but at least this way it won't block the update.


patosullivan (approved):
Have been using this change since Monday on my ship, allows for talk to actually compile.

PR #3091, opened 2023-12-06
notifications: group reply notifications appropriately

patosullivan:
fixes LAND-1304 by ensuring the thread doesn't include the reply id itself (which is unique for each reply).


arthyn (commented):
Maybe we should combine what this was trying to do https://github.com/tloncorp/landscape-apps/pull/2950 with what we need here. Checking length here seems kinda fraught

patosullivan:
@arthyn should I close this in favor of a new PR tackling the notification path issues generally?

arthyn:
sure yeah that works

PR #3092, opened 2023-12-06
chat: make sure we load in appropriate data when scrolling to a message outside of our current messages

patosullivan:
Fixes LAND-1331 (and any other case where our scrollTo message falls outside of currently loaded messages).

Previously, the scroller would just scroll to the top of what we currently have and useInfinitePosts wouldn't refetch to get messages around the message in question.

I also added a comment to another, related useEffect to make things a bit clearer.

### DIFF
--- a/ui/src/chat/ChatWindow.tsx
+++ b/ui/src/chat/ChatWindow.tsx
@@ -48,6 +48,7 @@ export default function ChatWindow({
     hasPreviousPage,
     fetchPreviousPage,
     refetch,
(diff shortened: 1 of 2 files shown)


latter-bolden:
Wasn't working for me in testing. Would also be curious to walk through it with you guys

patosullivan:
We looked at brian's setup out of band, turns out he was having an issue with git not pulling down branches (code would not change when he changed branches).

PR #3093, opened 2023-12-06
notifications: add support for receiving a notification for all messages in a channel

patosullivan:
fixes LAND-1313

If the user selects "All messages" in the notification settings for a channel, they will now receive notifications for all messages.


arthyn (approved):
lgtm

PR #3094, opened 2023-12-06
refs: allow posting of refs in notebook and gallery comments

patosullivan:
fixes LAND-1285 by allowing users to paste in refs in notebook and gallery comments (following the same pattern we use in ChatInput). Note that DiaryCommentField is used as the input in both diary comments and heap comments.


arthyn (approved):
lgtm

PR #3099, opened 2023-12-07
refs: show as unavailable when remote post data is null

patosullivan:
Fixes LAND-1346. The backend will return `null` for our `said` requests if we're requesting a post from a private group that we don't have access to. This change accounts for that.


latter-bolden (approved):
Lgtm, passed testing

PR #3104, opened 2023-12-07
threads: account for the fact that we sometimes only hear facts via the subscription in useChannels

patosullivan:
Sometimes (such as viewing a thread in mobile, after you've directly navigated to it via a notification or similar), we don't get updates via the subscription in useInfinitePosts, but we do via the subscription in useChannels. This PR adds a call to infinitePostsUpdater in the onEvent handler in useChannels so that we can be sure to update the cache appropriately when we hear these facts.

Calling it in both places shouldn't cause an issue, and doesn't seem to in my testing.

Fixes LAND-1338

### DIFF
--- a/ui/src/state/channel/channel.ts
+++ b/ui/src/state/channel/channel.ts
@@ -844,6 +844,16 @@ export function useChannels(): Channels {
       );
     }
 
+    if ('response' in event && 'post' in event.response) {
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):
Code change looks good.

Can now reliably repro the issue in mobile-view threads and confirmed this resolves it.

PR #3105, opened 2023-12-07
unreads: mark a chat as read when we hear that fact on unreads subscription

patosullivan:
Fixes LAND-1354 by marking a chat as read via useChatStore's read method if we hear a fact telling us that the chat is read. Previously we just called the `unread` function from useChatStore no matter what on hearing an unread fact about a chat channel.

### DIFF
--- a/ui/src/state/channel/channel.ts
+++ b/ui/src/state/channel/channel.ts
@@ -1030,16 +1030,24 @@ export function useUnreads(): Unreads {
 
     if (unread !== null) {
       const [app, flag] = nestToFlag(nest);
+
       if (app === 'chat') {
-        useChatStore
-          .getState()
-          .unread(flag, unread, () => markRead({ nest: `chat/${flag}` }));
+        if (unread['unread-id'] === null && unread.count === 0) {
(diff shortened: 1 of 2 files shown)


patosullivan:
Just pushed a similar fix for DMs

arthyn (approved):
lgtm, dms was going to be my comment but you got it

mrozanski:
I still see the same behavior I noticed before the DMs fix.

Steps:

Ship1 Talk client is open in two browsers.
`Ship1 develop`: pointing to the `develop` moon at hash `pqft6`.\
`Ship1 local`: used to test this fix, is pointing to localhost on branch `po/land-1354-mark-chats-as-read-when-we-hear-fact`

1. Ship2 sends a 1:1 message
2. Ship1 sees blue dot next to Ship2's name in Talk sidebar from both browsers
3. `Ship1 develop` opens the chat, blue dot is cleared
4. Expected: blue dot in `Ship1 local` is also cleared without the user opening the chat

Screenshot 2023-12-07 at 3 13 15 PM
Screenshot 2023-12-07 at 3 22 52 PM

PR #3106, opened 2023-12-07
chat spacing: resolve sent time width being too small

latter-bolden:
The update fixing the width of the sent time was off by a few pixels, should cover the worst case now without wrapping the line.

Also adjusts padding to line chat text start up with author text.

Fixes LAND-1357


patosullivan (approved):

patosullivan:
I think this is causing an issue, will add it to the ticket.

PR #3118, opened 2023-12-11
chat: when replying, only reset the message editor if we don't already have the mention

latter-bolden:
We have a bug right now where if you're replying in Chat and a new message comes in, the hook that sets the editor mention is always firing which clears away any other text you already had. 

This PR just checks to see if we actually need to set the mention before doing so.

Fixes LAND-1340


patosullivan (approved):
Looks good

PR #3119, opened 2023-12-11
channels: more accurate unread counts

Fang-:
The improvement here is two-fold:

First, we no longer recalculate the "most recent content" timestamp when marking a channel as read. Instead, we simply reuse the `.recency` which we were already tracking anyway. I don't remember why we thought we had to re-derive that on the spot. (2928b4b)

Second, we update the `$unread` type to provide, in addition to the total unread count, sub-counts for each context in which unreads may exist: top-level and within reply threads. (be58f62)

Draft because this does not yet include the frontend updates required for the latter change. cc @patosullivan 

For LAND-1393 and LAND-1394.
Fixes LAND-1412


desk/app/channels.hoon
```diff
@@ -1213,38 +1209,39 @@
...
         %+  skim
-          %~  tap  by
+          %-  bap:on-v-replies:c
           (lot:on-v-replies:c replies.u.u.parent `last-read.remark.channel ~)
         |=  [tim=time reply=(unit v-reply:c)]
         ?&  ?=(^ reply)
             !=(author.u.reply our.bowl)
         ==
-      :-  (add sum (lent unreads))
+      =/  count=@ud  (lent unreads)
```
midsum-salrux, line 1240: NB: I don't like reusing the name here but it was already doing that for `unreads`.

midsum-salrux (approved):
LGTM

Fang-:
Actually the top-level count (not the total) provided by this is inaccurate for cases where both a top-level message _with replies_ is below the top-level marker: in that case the replies should be included in the top-level count, but aren't.

Fixing...

latter-bolden (approved):

mrozanski:
Label `do not merge` added only because we're approaching a long holiday break, and we may prefer to merge once everyone is back to manage risks. (I'm open to reviewing this decision)

arthyn (approved):
lgtm lets test on binnec

PR #3120, opened 2023-12-11
perf: tweaks to speed up sidebar transition on desktop

patosullivan:
(partially) Fixes LAND-1360 by reducing re-rendering throughout the Sidebar and GroupSidebar components and their children. 

One major source of re-rendering was the data from useGroup() getting wiped on navigating away from a group, which caused the GroupSidebar itself to re-render and show loading state/skeleton on the way out. I added `keepPreviousData` to the useGroup() query to keep current data *until* we want to load in a new group.

I also memoized as much as seemed appropriate throughout.

I played around with animationConfig settings in GroupsNav, but nothing stood out as actually improving the perceived speed.

I also attempted to use framer-motion's `useIsPresent` hook to prevent certain components from rendering (including preventing group images from loading in the sidebar), but it usually either didn't make an appreciable difference or it added too much overhead for it to be worth it.

I also realized that the number of notifications in the Activity list in the home screen has an effect on how quickly we transition from viewing a group to returning to the home screen, and attempted to mitigate this by using Virtuoso to turn the notifications list into a virtualized list. This didn't make any appreciable difference. I think we ought to think of other ways to optimize here (maybe we ought to page the notifications?).

To get a good feel of how these changes will actually feel in production, I recommend running a build, running serve, and pointing the frontend at your personal ship. If you want to see what I'm talking about, re: the notifications themselves causing it to feel slow, make sure you have a lot of notifications. You ought to also test in Safari, where the issue was most pronounced.

Here is how it looks now in that scenario:

https://github.com/tloncorp/landscape-apps/assets/1221094/ebc88f62-3cd6-4775-a2ae-e9919871a705

### DIFF
--- a/ui/src/components/Avatar.tsx
+++ b/ui/src/components/Avatar.tsx
@@ -1,6 +1,6 @@
 import { isValidPatp } from 'urbit-ob';
 import classNames from 'classnames';
-import React, { CSSProperties } from 'react';
+import React, { CSSProperties, useMemo } from 'react';
 import '@urbit/sigil-js';
 import { Contact, cite } from '@urbit/api';
 import _ from 'lodash';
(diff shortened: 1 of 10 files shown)


latter-bolden (approved):
Code looks good, does appear to be sped up on Safari.

PR #3121, opened 2023-12-12
groups: fix broken "requested to join" notifications

latter-bolden:
OTT

Fixes LAND-1370

To reproduce, create a private group and from another ship, request to join it. You'll get a notification in your activity feed that doesn't work without this PR. You'll have to install this branch's `%groups` desk to test.


mrozanski:
Thank you for the instructions in the PR description, they were very helpful, please keep doing that.
Tested and approved.

PR #3126, opened 2023-12-15
groups: redesign group options sheet

jamesacklin:
Fixes LAND-1397. 

Brings GroupActions into alignment with the design linked in LAND-1397:
- Adds a "row" mode to HostConnection to fit the sheet design
- Groups common actions together and adds spacing between groups
- Adds subtitles to each action
- Conditionally hides/shows the "Leave group" option for members (vs admins, who should 'delete' the group or ask to have their role removed)
- Removes the "Channels" | "Members" | "Invite People" items from the top of the mobile ChannelList and adds a TODO to add the welcome bubble for admins who have created a new group.

Note that this PR converts all ActionMenus to use the "sheet" treatment on mobile (full-width, no gutter, attached to screen bottom). This PR also only touches the contents of the GroupActions dropdown/sheet.

This PR does *not* unify the GroupActions dropdown/sheet with any other dropdown/sheet (LAND-1221). We will need to update all action dropdown/sheets to match this new design (search in the codebase for `actions.push({` for an idea of the scope of change involved).

Revised group home with no options other than channels:
image

A group we don't administrate with no public invites:
image

A group we don't administrate with public invites:
image

A group we administrate (doesn't matter if invites are on/off, we're administrators and can enable links if we want to using the "Invite People" menu):
image

### DIFF
--- a/ui/src/channels/HostConnection.tsx
+++ b/ui/src/channels/HostConnection.tsx
@@ -11,6 +11,7 @@ import { Saga } from '@/types/groups';
 import Bullet16Icon from '@/components/icons/Bullet16Icon';
 import Tooltip from '@/components/Tooltip';
 import { useNegotiate } from '@/state/negotiation';
+import { get } from 'lodash';
 
 interface HostConnectionProps {
   ship: string;
@@ -85,7 +86,14 @@ export default function HostConnection({
             <Bullet16Icon
               className={cn(
                 'h-4 w-4 flex-none',
-                `text-${getHostConnectionColor(saga, status, matched)}-400`
+                getHostConnectionColor(saga, status, matched) === 'red' &&
+                  'text-red-400',
+                getHostConnectionColor(saga, status, matched) === 'yellow' &&
+                  'text-yellow-400',
+                getHostConnectionColor(saga, status, matched) === 'green' &&
+                  'text-green-400',
+                getHostConnectionColor(saga, status, matched) === 'gray' &&
+                  'text-gray-400'
               )}
             />
           </span>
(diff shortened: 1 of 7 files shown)


urcades (approved):
Beautiful, stamp!

latter-bolden (approved):
Looking good, worked in testing.

ui/src/channels/HostConnection.tsx
```diff
@@ -85,7 +85,7 @@ export default function HostConnection({
             <Bullet16Icon
               className={cn(
                 'h-4 w-4 flex-none',
-                getHostConnectionColor(saga, status, matched)
+                `text-${getHostConnectionColor(saga, status, matched)}-400`
```
mrozanski, line 88: @arthyn was this a concern for you? (concatenating Tailwind class names). 
This was already a value generated in runtime, the difference is it's being composed in this line now and between backticks, but still runtime, right?

jamesacklin: @arthyn Heard - take a look

ui/src/groups/GroupSidebar/ChannelList.tsx
```diff
@@ -201,48 +193,9 @@ const ChannelList = React.memo(({ paddingTop }: { paddingTop?: number }) => {
...
-          state={{ backgroundLocation: location }}
-          actions={<ElipsisIcon className="h-6 w-6 text-blue" />}
-        >
-          Invite People
-        </SidebarItem>
-      </div>
-    );
-  }, [isMobile, flag, location, paddingTop]);
+    // TODO: Add welcome message to group for admins and non-admins
+    return <div className={cn(paddingTop && `pt-${paddingTop}`)} />;
```
arthyn, line 197: this was previously there, but we should change at some point

ui/src/components/ShipConnection.tsx
```diff
@@ -33,7 +33,7 @@ export default function ShipConnection({
   const color = isSelf
     ? 'text-green-400'
     : matchedOrPending
-    ? getConnectionColor(status)
+    ? `text-${getConnectionColor(status)}-400`
```
arthyn, line 36: need to change here

arthyn (changes requested):
just needs one more change, then gtg

arthyn (approved):
lgtm

PR #3127, opened 2023-12-15
groups: add SoundCloud embeds

darighost:
SoundCloud already works via Noembed, so supporting it in groups is very simple. Here's what it looks like:

Screenshot 2023-12-15 at 3 08 52

Originally, I rendered whatever `iframe` SoundCloud gave me, like so...

```typescript
  const iframe = DOMPurify.sanitize(rawIframe, { ADD_TAGS: ['iframe'] });
  return <div dangerouslySetInnerHTML={{ __html: iframe }} />;
```

Bad! What if a sneaky hacker snuck some untrusted HTML into the `SoundcloudEmbed` component somehow? They could load any iframe they want! 

So we can either:

1. Hope there's no way to sneak in a malicious `iframe`, or
2. Make the iframe myself from the URL

Option 2 is the right option. However, maybe my implementation of option 2 is really bad! That's okay, criticism is welcome.

---

PS - Tumblr and DeviantArt also support oembed (though not via noembed) and I added them. But I removed them for this PR, because I perceived that maybe they wouldn't be as desirable as SoundCloud. If I miscalculated, let me know, and I can add them back in.

### DIFF
--- a/ui/src/chat/ChatEmbedContent/ChatEmbedContent.tsx
+++ b/ui/src/chat/ChatEmbedContent/ChatEmbedContent.tsx
@@ -8,6 +8,7 @@ import DOMPurify from 'dompurify';
 import YouTubeEmbed from './YouTubeEmbed';
 import TwitterEmbed from './TwitterEmbed';
 import SpotifyEmbed from './SpotifyEmbed';
+import SoundcloudEmbed from './SoundcloudEmbed';
 import AudioPlayer from './AudioPlayer';
 
 const trustedProviders = [
@@ -22,6 +23,10 @@ const trustedProviders = [
   {
     name: 'Spotify',
     regex: /^https:\/\/open\.spotify\.com\//,
+  },
+  {
+    name: 'SoundCloud',
+    regex: /^https:\/\/soundcloud\.com\//,
   },
 ];
 
@@ -133,6 +138,15 @@ function ChatEmbedContent({
           />
         </div>
       );
+    }
+
+    if (provider === 'SoundCloud') {
+      // SoundCloud offers an iframe we can embed, so we have to explicitly allow
+      // that via DOMPurify.
+      const embedIframe = DOMPurify.sanitize(rawEmbedHtml, {
+        ADD_TAGS: ['iframe'],
+      });
+      return <SoundcloudEmbed iframe={embedIframe} />;
     }
 
     return (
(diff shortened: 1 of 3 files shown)


latter-bolden:
Nice thanks for adding this! Embeds seem to be loading well and play music as expected.

Am noticing scroller inversion when hovering over them though. In the video, whats happening is my native scroll down on the mousepad causes it to nav up (opposite) while over the iFrame. It then reverts to normal and goes down again once I initiate another scroll from above it. In the video, I'm dragging down in the same direction throughout.

https://github.com/tloncorp/landscape-apps/assets/90741358/d77529c8-0767-460f-a4bc-140a68a4a32f

darighost:
>Am noticing scroller inversion when hovering over them though.

What a curious bug! Thanks a lot for testing so thoroughly.

Looking through the code in the embedded Spotify `iframe`, they're capturing scrolling with Javascript and inversing it.

I googled this problem and the conventional advice is that [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) go to the `iframe` first in this situation, so there's nothing you can do. 

Based on that, the two easy solutions are:

1. Accept that scrolling will be weird in this situation
2. Don't accept SoundCloud embeds

But a more creative solution would be to make sure _our_ page captures pointer events first, then passes along acceptable ones (ie, clicks but not scrolls) to the `iframe`. I think we could do thusly:

* create an invisible `div` with the same location and size as the `iframe`.
* set the z-index of the `div` so that it's "on top" of the `iframe`.
* capture pointer events to the `div`
    * if the pointer event is a click, 
        * momentarily disable the `div`
        * refire the same click (this time it will hit the `iframe`)
        * Re-enable this `div` 
    * else let it fire normally

Kinda like a [clickjacking attack](https://owasp.org/www-community/attacks/Clickjacking), but without the "attack" part.

Should require only a few lines of styling and an event listener in JS. Tonight I won't have internet, but I'll try this approach tomorrow and let you know :)

darighost:
Update: My idea, while fun to implement, turned out to be bad!

Overlaying a `div` to intercept scrolls (and pass clicks) does work. However, it also introduces new problems. 

For example: when hovering over clickable parts of the iframe, nothing happens (because the iframe doesn't receive the hovers). 

Closing this PR, but if any "real" web devs have a suggestion I'm all ears :)

PR #3130, opened 2023-12-15
groups: reformats Group Settings

jamesacklin:
This PR proposes a near-complete overhaul of Group Settings to fit it into a mobile form factor (since we are going to push group creators to customize the group after the fact).

- Breaks the settings editor out of the stacked modal/sidebar-modal form factor in favor of full-screen controls.
- Preserves nested routes for the desktop and duplicates them to flat routes for mobile. Mobile necessitates full-page navigation, while desktop makes use of an `<Outlet />` component. I couldn't figure out a way to handle this in the link handler itself, since Outlets need nested routes and our mobile views cannot feasibly use Outlets.
- Eliminates the gray backgrounds and rounded-corner islands in favor of flat rows. Unifies spacing and padding to accommodate. 
- Makes global changes to the LureInviteBlock and the GroupInviteDialog components to mark the difference between "shar(ing) your group with others by sending them a link" and "invit(ing) people to your group who are already using Urbit". These changes are visible both in the Group settings and in the "invite people" actions in groups themselves.

This PR does **not** address the woeful split-screen "invites" vs "members" view, which is used to assign roles. We should create another task to move the role-assignment concerns over to the Members list (`/ui/src/groups/Members.tsx`) and to fix the list virtualization there.

https://github.com/tloncorp/landscape-apps/assets/748181/d25a3fe4-8307-4f17-b39c-0bdbebf42230

The code and interface itself is ready for review; I will need to fix the E2E tests (@mrozanski may need your help here).

Fixes LAND-1081


urcades (approved):
Based on what was presented earlier, stamping.

It's worth noting that the desktop rendering for settings is a little strange due to the optimizations we're making for mobile, which is our focus at this time.

This is a cost we're accepting to making this feel at home in a smaller viewport.

jamesacklin:
@mrozanski E2E tests appear to be failing due to this change: https://github.com/tloncorp/landscape-apps/pull/3130/files#diff-a0210da039b1dcd3f5c2d9921028634bb07503369351fa8b6cd552fdf28aac0cL106

I removed the "Cancel" button from the GroupInviteDialog component because it seemed duplicitous but I can see why we needed it. Amending the PR for this to only appear in dialog contexts (and hopefully fix the tests).

mrozanski (approved):

arthyn:
I think we're missing the connection check somewhere here. How does an admin know if their changes aren't taking place because of an error or an offline host? If we don't want to address now, soon maybe we should make a button component or wrapper that is disabled when host mismatches?

jamesacklin:
> we should make a button component or wrapper that is disabled when host mismatches?

I'd love your help with this. There are lots of pieces of this interface that "save" differently (update on click, use formik or react-form-hook to save, etc.).

arthyn:
let's just writeup an issue, will merge

PR #3135, opened 2023-12-16
treat webm urls as videos

alecananian:
Updates regular expressions to treat `.webm` links as videos. This fixes an issue where when a message is just a single `.webm` URL, it wasn't posting at all. It also allows `.webm` videos to be rendered in embedded `<video>` tags on the front-end.


arthyn (approved):
lgtm

PR #3137, opened 2023-12-18
welcome card: update style, copy, and behavior

latter-bolden:
This brings the welcome card up to date with the copy and styling in Figma. Adapts it to be shown on desktop views as well, positioned above the activity feed on the home screen.

Closed status is no longer stored locally on the frontend, instead its tracked in the settings agent. This means closing the card once will hide it on all clients.

Screenshot 2023-12-18 at 10 13 12 AM

Screenshot 2023-12-18 at 10 13 55 AM

Screenshot 2023-12-18 at 10 15 24 AM

Screenshot 2023-12-18 at 10 14 50 AM

Fixes LAND-1399

### DIFF
--- a/ui/src/components/WelcomeCard.tsx
+++ b/ui/src/components/WelcomeCard.tsx
@@ -1,42 +1,68 @@
-import { Link } from 'react-router-dom';
-import { useLocalStorage } from 'usehooks-ts';
-
-import { createStorageKey } from '@/logic/utils';
+import { usePutEntryMutation, useSeenWelcomeCard } from '@/state/settings';
+import { useCallback, useState } from 'react';
+import { useIsMobile } from '@/logic/useMedia';
+import X16Icon from './icons/X16Icon';
 
 export default function WelcomeCard() {
-  const [isWelcomeSeen, setIsWelcomeSeen] = useLocalStorage<boolean>(
-    createStorageKey('welcome-seen'),
-    false
-  );
+  const isMobile = useIsMobile();
+  const alreadySeen = useSeenWelcomeCard();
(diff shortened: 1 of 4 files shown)


ui/src/components/WelcomeCard.tsx
```diff
@@ -1,42 +1,68 @@
...
+    return (
+      <div className="mx-6 mt-2 mb-4 space-y-4 rounded-[32px] bg-green-soft py-6 px-8">
+        <h3 className="text-lg font-medium">Welcome to Tlon</h3>
+        <p className="leading-5">
+          This is Tlon: An app for messaging friends and constructing
+          communities.
+        </p>
+        <p className="leading-5">
+          Visit the{' '}
+          <span className="font-semibold dark:font-bold">Tlon Studio</span>{' '}
```
alecananian, line 34: Might make sense to have `Tlon Studio` be a link?

urcades: @alecananian during first time boots, this card will exist above the two pre-invited groups "Tlon Locals" and "Tlon Studio", so they'll technically be right where they need to be! That being said, I think it'd a good idea to duplicate the ability to jump to Tlon Studio in case curious people tap the bolded word.

urcades (changes requested):
Linking to Tlon Studio for duplicity's sake is probably a good idea — after this it's g2g!

latter-bolden:
Added that affordance for any curious clickers :~)

alecananian (approved):

urcades (approved):
👍 nice

PR #3139, opened 2023-12-19
fix crash when navigating to deleted reply from activity

alecananian:
Fixes crash when you click into an activity notification for a thread reply that was deleted. It still navigates you to a weirdly empty thread, but that can be addressed with better UX at a different time


arthyn (approved):
lgtm

PR #3140, opened 2023-12-19
dm: in dm threads, make sure we don't display normal date divider if unread divider is shown

latter-bolden:
OTT

Fixes LAND-1434


arthyn (approved):
lgtm

PR #3141, opened 2023-12-20
groups: fix issue with deleting gallery item from item page

alecananian:
- Fixes issue with gallery item not able to be deleted from the item details page
- Fixes the `useDeletePostMutation` hook never firing `onSuccess` because of mismatched time formats


arthyn (approved):
lgtm

PR #3145, opened 2023-12-20
groups: poke %channels-server, not old agents

Fang-:
+verify-group-cabals was still poking agents using the names given in the nest, but those have been deprecated in favor of channels-server. Here, we make sure to always poke channels-server with a new-style channel-command instead.

Somewhat untested, somehow I couldn't get into this branch of the code (or even get the role deletion to stick). @arthyn this is probably better than "100% certainly broken", but we should probably pair on actually running the new code here.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -250,6 +250,8 @@
 ++  verify-group-cabals
   |=  [[=flag:g [* =group:g]] core=_cor]
   =.  core
+    ::  repair members as needed
+    ::
     %+  roll
       ~(tap by fleet.group)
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm, @midsum-salrux and I can try running this while you're out

mrozanski:
Does this address, partially address, or relate to an existing Linear ticket?

arthyn:
@mrozanski no this is just leftover bad behavior from the refactor

arthyn:
@Fang- found it while chasing LAND-1343

PR #3148, opened 2023-12-21
groups: fix gallery image view and delete button in list mode

alecananian:
- Fixes issue with gallery image thumbnails not loading in list view
- Fixes issue with gallery item delete button not working in list view

https://github.com/tloncorp/landscape-apps/assets/1013230/df7b75df-9261-4474-9577-0d3d6259900f

### DIFF
--- a/ui/src/heap/HeapRow.tsx
+++ b/ui/src/heap/HeapRow.tsx
@@ -17,7 +17,6 @@ import ElipsisSmallIcon from '@/components/icons/EllipsisSmallIcon';
 import MusicLargeIcon from '@/components/icons/MusicLargeIcon';
 import LinkIcon from '@/components/icons/LinkIcon';
 import CopyIcon from '@/components/icons/CopyIcon';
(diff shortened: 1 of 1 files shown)


ui/src/heap/HeapRow.tsx
```diff
@@ -64,7 +58,7 @@ function Actions({
   author,
 }: TopBarProps) {
   const [deleteOpen, setDeleteOpen] = useState(false);
-  const nest = useNest();
+  const chFlag = useChannelFlag();
```
arthyn, line 61: hmm curious why this would be wrong, is there an issue with the nest hook?

arthyn (approved):
lgtm

ui/src/heap/HeapRow.tsx
alecananian: Yes it's probably obsolete. It looks for an `app` router param, which I don't see defined in any routes.

PR #3153, opened 2023-12-22
groups: expose "can-read" as gate through scry

Fang-:
The groups agents (channels-server, diary, heap) do some dumb stuff. Specifically, whenever permissions change _at all_, they re-run permission checks for _all subscribers_ to _all channels_ in the group where the change happened.

This is dumb, and they should be smarter about it, but urbit is fast, it shouldn't be a _huge_ deal. Except, "running a permission check" involves scrying into the groups agent, which has non-trivial overhead.

The fix is equally dumb: don't improve the logic, just remove the overhead. This brings us down from "tens of milliseconds per subscriber" to "tens of _microseconds_ per subscriber".

Instead of only exposing a per-channel per-ship scry endpoint for checking read permissions, groups now also exposes a generic endpoint that produces a gate. This gate can be used to run a read permissions check with varying arguments repeatedly.

We update the `+can-read:perms` helper from `/lib/channel-utils` to use this gate, and make sure its scry of the gate is memoized. This removes the scry overhead for subsequent invocations within the same event.

There's light concerns here around the context of the gate being used to exfiltrate otherwise-private data from the groups agent. However, two reasons not to care:
- These endpoints are only accessible locally.
- The groups agent already exposes nearly all of its state on other scry endpoints, and userspace permissions don't exist yet, so this won't be "leaking" data that isn't already accessible.

That second point will not hold indefinitely, and so we will want to revisit this in the near future. Concretely, groups (and its clients) should be reconsidered to improve the ergonomics around reacting to permissions. This will likely be part of the groups agent refactor coming up next year.

Until then though, this should rid us of exceedingly painful performance when managing roles/permissions. 

Possibly fixes LAND-1343

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -343,7 +343,7 @@
     |=  [=nest:c co=_cor]
     ^+  cor
     %+  roll  ~(tap in p.diff)
-    |=  [=ship ci=_cor]
+    |=  [=ship ci=_co]
     ^+  cor
     =/  ca  (ca-abed:ca-core:ci nest)
     ca-abet:(ca-revoke:ca ship)
--- a/desk/app/diary.hoon
+++ b/desk/app/diary.hoon
@@ -4,7 +4,7 @@
 /+  default-agent, verb, dbug
 /+  not=notes
 /+  qup=quips
-/+  volume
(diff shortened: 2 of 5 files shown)


midsum-salrux (approved):
LGTM

PR #3155, opened 2023-12-23
fix gallery next and prev navigation

alecananian:
Simplifies gallery navigation logic and fixes issue with navigating to empty/deleted detail page

### DIFF
--- a/ui/src/heap/HeapDetail.tsx
+++ b/ui/src/heap/HeapDetail.tsx
@@ -139,7 +139,7 @@ export default function HeapDetail({ title }: ViewProps) {
           ) : null}
         </div>
         <div className="flex w-full flex-col lg:h-full lg:w-72 lg:border-l-2 lg:border-gray-50 xl:w-96">
-          {essay ? <HeapDetailSidebarInfo essay={essay} /> : null}
+          <HeapDetailSidebarInfo essay={essay} />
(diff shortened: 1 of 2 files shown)


mrozanski:
What are the actual steps to reproduce the issue being fixed?
Here's what I tested:

From Newest post

1. Open a gallery (from a mobile size screen) with two posts (photos)
2. Open the first photo in the list (latest uploaded)
3. Expected: no left arrow, and right arrow present (takes me to next photo in the list)
4. Observed: **Left arrow present (tapping closes the gallery) and no right arrow**

This branch fixes that behavior

From Oldest post

1. Open a gallery (from a mobile size screen) with two posts (photos)
2. Open the last photo in the list (oldest uploaded)
3. Expected: no right arrow, and left arrow present (takes me to previous photo in the list)
4. Observed: **both arrows present (tapping right one generates exception)**

This branch fixes that behavior too

What else should I test?

alecananian:
There seems to be some way that a post becomes `null`, so this PR fixes issues that arise from that by adding the filter `.filter(([, v]) => v !== null)` that is used elsewhere. I'm not sure the steps to reproduce (may have to do with a deleted post), but you can see the issue by following the path in the ticket and navigating through that gallery.

mrozanski (approved):

PR #3159, opened 2024-01-02
optimistically remove deleted roles from group cache

alecananian:
- Updates the delete group role flow to wait for success, optimistically update group cache, and invalidate group query when complete
- Updates create group role flow to go back to roles list on create success
- Fixes styling on various button loading states


arthyn (approved):
lgtm

PR #3161, opened 2024-01-02
add character counter to group description editor

alecananian:
Adds a character counter and error state to the group description editor

https://github.com/tloncorp/landscape-apps/assets/1013230/ce3333ee-95f4-472c-9bfc-09501dde6ad8


mrozanski (changes requested):
Works when. creating new groups. For some reason it doesn't register the error when editing an existing one.

Screenshot 2024-01-02 at 10 26 46 AM

alecananian:
Good catch @mrozanski! I've updated the edit group form to use the same `onChange` validation strategy as the create group form

mrozanski (approved):

PR #3162, opened 2024-01-02
fix issue with joining/leaving groups containing custom channels

sidnym-ladrut:
These changes fix a bug reported by `~darted-laster` in `~tommur-dostyn/tlon-studio/~tommur-dostyn/support` about new ships being unable to join groups containing a custom channel.

Please note that I've only tested the following workflows:

1. Joining a group containing a custom channel (i.e. `~hiddev-dannut/new-hooniverse`) on the livenet from a comet
2. Leaving the same group on the livenet from a comet

Please let me know if other workflows should be testing before merging.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -875,32 +875,32 @@
     ++  leave-channels
       |=  nests=(list nest:g)
       ^-  (list card)
-      %+  turn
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm, thank you!

PR #3163, opened 2024-01-05
Groups e2e: add/edit note using markdown Land-1240

mrozanski:
This PR:

- Adds configuration to boot two fake planets (from pre-saved piers in gcloud) for e2e tests
- Keeps the two galaxies we were using (~zod and ~bus) to enable "fake" routing between the two planets
- Updates existing tests to use the two new planets instead of ~zod and ~bus
- Adds new specs file 003 to cover markdown and editing notes
- Updates the readme

### DIFF
--- a/ui/E2E.md
+++ b/ui/E2E.md
@@ -8,11 +8,17 @@ The script:
 
 1. fetches fake ships piers from GCS
 2. downloads the urbit binaries for the local architecture
-3. boots the ships
+3. boots the test planets and sponsors
 4. executes the tests against the current ui code version
 5. kills spawned processes
 
 Test specs are in the `e2e/` directory.
+
+Each File can run in parallel, and they will while there are available workers.
+
+Number of workers is configured in ui/playwright.config.ts
+
+Specs in one file run in sequence.
 
 ## Running locally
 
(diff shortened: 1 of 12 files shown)


ui/e2e/003-post-note-with-markdown.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    .getByTestId('edit-with-markdown-toggle')
+    .getByRole('button')
+    .click();
+  await expect(
+    page.getByTestId('edit-with-markdown-toggle').getByRole('button')
+  ).toHaveAttribute('aria-pressed', 'true');
+
+  await page.getByTestId('note-markdown-editor').click();
+  await page.getByTestId('note-markdown-editor').fill('## Heading level 2');
+  await page.getByTestId('save-note-button').click();
```
arthyn, line 41: Do we actually need to use test ids or can we target these with text? Wondering if we have an accessibility issue?

arthyn (changes requested):
Looks good except the question

ui/e2e/003-post-note-with-markdown.spec.ts
mrozanski: It's just for practicality.
Using test IDs is a decision I made (maybe should have been a recommendation to discuss).

In my experience, they make e2e tests easier to maintain and prevent false negatives.
If we want to diagnose accessibility issues we can use any tool specifically designed for that, and will get better coverage and realistic insights.

My preference for using test IDs is based on making explicit when an element is targeted by a functional test.
Otherwise, when changes are made to behaviors or text based on business or design decisions, tests may break unless their locators are updated accordingly.
If it's not clear at a glance that a component I'm modifying is expected (by the tests) to be there, do something, or read something, I may make my changes and omit updating the tests. Meaning I'll have to come back once they were merged and the failing tests triggered.

If we prefer to use text, I can make the changes. LMK

ui/E2E.md
```diff
@@ -22,20 +28,33 @@ After installing  this project's dependencies with `npm install`,
 npm run e2e
 ```
 
+### Adding tests
+
+Create a new spec file in the `e2e/` directory.
+
+The file name should be descriptive of the test suite.
+
+Add data-testid attributes to the components you want to test and accessthem using `page.getByTestId()`.
```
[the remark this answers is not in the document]
mrozanski: It _may_ indicate but not necessarily.

ui/e2e/003-post-note-with-markdown.spec.ts
arthyn: yeah that makes sense to me

arthyn (approved):
lgtm

PR #3164, opened 2024-01-08
talk: fix dropdown on multi-dms invite dialog

alecananian:
- Fixes the dropdown getting cut off on the multi-party DMs invite dialog
- Adds `lint-staged` to `devDependencies` because the pre-commit hook using `npx` was grabbing v15 by default, which is not compatible with Node v16

Screenshot 2024-01-08 at 9 19 44 AM


arthyn (approved):
lgtm, probably time to update node

PR #3165, opened 2024-01-08
remove outdated useNest hook

alecananian:
Removes the outdated `useNest` hook as discussed in #3148, since the `app` param is no longer defined on any routes


arthyn (approved):
lgtm

PR #3168, opened 2024-01-10
groups: flag content for admin review

arthyn:
Fixes LAND-1442 by implementing a new action on all types of posts and replies to flag them to an administrator. This notably refactors a little bit how notifications are the frontend are routed, but it should be improved over what's currently on mainnet.


midsum-salrux (approved):
Backend looks good.

mrozanski:
Looking.

mrozanski:
When I click on the new option to report, the message is hidden immediately. If I cancel instead of sending (x icon) the message remains hidden.
I would expect that the message is hidden only if send the report and stays visible otherwise.

Also, in that state (message hidden) the option for reporting has an alt text of "show message". Feels a bit confusing.

Screenshot 2024-01-12 at 10 24 22 AM

mrozanski:
I tested a bunch of use cases and nothing new came up.
I only tested on desktop, I can do a mobile pass from this branch (not today though). If this is merged I can do that after too.

I still think (as discussed with @arthyn) that as long as the report action has a confirmation step, the message should't be hidden until the "send report" button that confirms the action is clicked in the dialog so that if I change my mind and don't report, everything stays as it was. Not blocking but not ideal.

mrozanski:
Pending items:

Let's consider this a viable initial version once these are added:

* Don't show me the report tooltip icon on posts already flagged by me (still can be un-hidden from the eye icon but not flagged more than once)
* Indicate who is the reporter in the activity screen notification to the host

@arthyn lmk if I'm missing anything we want to have for a first release, we can keep improving after.

mrozanski (approved):
* Testing from two self-hosted moons, running from vite locally
* Using a recently created group, public, hosted by a testing moon

- [ ] guest, non admin, can flag post from chat channel created by host
- [x] message becomes hidden for ship that flagged
- [x] reporter ship: option to flag a flagged message is disabled
- [x] reporter ship: message can be unhidden and hidden again
- [x] Can report a note
- [x] Can report a gallery post
- [x] Can report a note comment
- [x] admin sees event in activity including reporter ship name
- [x] Report in activity links to flagged post (recent, visible in newest scroll position)
- [x] Report in activity links to flagged post (not recent, off newest scroll position)
- [x] admin can delete message
- [x] deleted message dissapears from reporter history
- [x] Repeat all cases from a chat channel created by non-host admin
- [x] Repeat all cases from a note channel created by non-host admin
- [x] Repeat all cases from a gallery channel created by non-host admin

Passed :tada:

PR #3169, opened 2024-01-11
LAND-1057: Fetch contacts for authors on receiving checkpoint

midsum-salrux:
https://linear.app/tlon/issue/LAND-1057/contacts-on-group-join-grab-latest-set-of-authors


arthyn (approved):
lgtm

PR #3171, opened 2024-01-11
app: fix grid/landscape redirects

arthyn:
This was a regression from when we renamed %garden/grid to %landscape, reported by ~nallux-dozryl


arthyn:
Fixes LAND-1468

alecananian (approved):

PR #3172, opened 2024-01-11
connection: fix color regression

arthyn:
Fixes LAND-1462 by refactoring the color passing a little so it's less likely to miss cases. 

Also the linter got triggered on a bunch of files, if you hide whitespace changes it should be easier to review.

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -458,6 +458,11 @@
     ?~  p.sign  cor
     %-  (slog leaf+"Failed to hark" u.p.sign)
     cor
+      [%contacts %join-heed ~]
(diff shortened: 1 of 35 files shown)


alecananian:
There are a lot of formatting changes in this PR that get overwritten when running `npm run lint:format`. @arthyn you may have some extra setting on your IDE that formats with a different config than what's specific in the repo.

arthyn:
@alecananian hmm so running this manually seems to reverse, and the pre-commit hook didn't change things so it's possible this is something to do with not every branch having those recent node changes. I'll keep an eye on it.

alecananian (approved):

PR #3174, opened 2024-01-12
Handle poke-ack from adding contacts

midsum-salrux:
This addresses an issue from https://github.com/tloncorp/landscape-apps/pull/3169 where poke acks from adding contacts on channel join weren't handled.


arthyn (commented):
looks good, but e2e is failing

arthyn (approved):
lgtm

PR #3175, opened 2024-01-12
LAND-1455: rut:by -> urn:by and bump kelvin version

midsum-salrux:
https://github.com/urbit/urbit/releases/tag/411k-rc0


Fang-:
If we put both `412` and `411` in the `sys.kelvin` here we should just be free to merge/release this ahead of the actual 411 release, right? `urn:by` already exists in `412`.

arthyn:
@Fang- we should merge to this kelvin branch and then we can deploy it separately to the kelvin integration moon

arthyn (approved):
lgtm

arthyn:
@midsum-salrux @Fang- mentioned we could probably signal both kelvins, which would mean we could just merge this to develop instead. y'all want to test that and then update the base branch here based on what you find?

PR #3177, opened 2024-01-15
add route to support thread reacts

alecananian:
Resolves LAND-1391
Resolves LAND-1460

Adds a new route to support showing the emoji picker on mobile when deep-linking to a specific message in a thread via activity tab/push notifications


latter-bolden (approved):
Reacting to root thread messages was still failing for me, but pushed a small change to the  picker logic that seems to fix it.

With that change, is working as expected.

alecananian:
I am realizing now I very stupidly did not even tap on an emoji to test the sending of it 🙂 Thanks for the change!

PR #3178, opened 2024-01-15
fix crash when pasting bold/italics into new gallery block

alecananian:
Fixes recursion crash when pasting pre-formatted text with bold/italics into the New Block modal for galleries


latter-bolden (approved):
Makes sense, tested & fixes the issue.

Worth noting this function seems to only be referenced once to check if the curio input is empty. Refactored a bit, we could probably remove this function entirely and lean on the TipTap editor's built in [`isEmpty`](https://tiptap.dev/docs/editor/api/editor#is-empty) method which we use elsewhere.

PR #3179, opened 2024-01-16
perf: update mark warmer

arthyn:
We missed quite a few of these when refactoring. This was something random @latter-bolden and I found when discussing fetching.


midsum-salrux (approved):
looks warm to mark

PR #3180, opened 2024-01-16
LAND-1058: Fetch new post authors as contacts

midsum-salrux:
https://linear.app/tlon/issue/LAND-1058/contacts-on-fetch-of-posts-collect-new-authors-to-register-as-contact

This doesn't check whether we're adding someone we already have as a contact. I think that it's better for %contacts to make that idempotent than checking it here, which it looks it does but I'm not 100%.


arthyn:
@midsum-salrux yeah %heed is idempotent

arthyn (approved):
lgtm

PR #3183, opened 2024-01-17
Add publicUrlBase to S3 configuration

mrdomino:
Needed for Cloudflare R2.


mrdomino:
See also tloncorp/landscape#252.

cc @Fang-.

Currently untested.

mrdomino:
As mentioned in the other PR, fixed an issue with this, but would love help actually testing it on a fakezod.

arthyn (approved):
lgtm

PR #3186, opened 2024-01-18
groups: fetching tweaks

latter-bolden:
Handles a few improvements for fetching posts:

- avoids incomplete message history when navigating to a channel by only adding new messages to the cache if we've already fetched the query
- on boot, any cached post queries with unreads get refetched so new messages are immediately available upon navigating to the channel
- bumps page size to 100 in the app and 300 on desktop to prevent extra fetching on scroll and allow fewer requests on refetch

We should monitor performance after this lands on our ships. Particularly for any of us with lots of groups that are used frequently. We may need to restrict on boot cache updates to pinned groups for now if it proves too expensive. 

Fixes LAND-1473
Fixes LAND-1474


arthyn (approved):
lgtm

PR #3188, opened 2024-01-19
various: remove old/stale/unused dependencies

Fang-:
We remove all references to and dependencies on graph-store, metadata-store, group-store, and versions of channel types older than the previous generation.

This does mean that we remove migration/import logic for versions of groups preceding the previous one, but the migration path coming from that far back was already known to need some manual guidance.

Notably inclusion is a bugfix in /app/notify: it was still scrying into %group-store, when it should have been updated to scry into %groups instead. PTAL.

Marking as draft for now, because:
- [x] I should test a clean install off of this (plus base-dev and landscape-dev folders).
- [x] I should cull old marks, make sure all the remaining ones still build. (At the very least, there's still marks for old update types.)
- [ ] We may want to hold off on release until we've helped out known stragglers stuck in the graph-store era (there's at least one).

Fixes LAND-1243.


Fang-:
Both the groups and talk desks can now be constructed from the following copy operations:

```bash
cp -RL urbit/pkg/base-dev/* ship/desk/
cp -RL landscape/desk-dev/* ship/desk/
cp -RL landscape-apps/desk/* ship/desk
```

(Replacing `desk/*` at the end there with `talk/*` for the talk desk, obviously.)

Neither of them need _all_ the files that the dev desks provide, but given we don't have a way of specifying/copying the exact subset we need, we just stay with the status quo of "superfluous inclusions" for now.

We _should_, when we next deploy, probably do a full mounted desk wipe + the above copy operation, to make sure we're at least not leaving any of the now-removed files in place. Perhaps CI's deploy logic already takes care of this?

desk/app/notify.hoon
```diff
@@ -362,7 +362,15 @@
     =/  gs  ~(tap in groups.whitelist.entry)
     |-
     ?~  gs  %.n
-    ?:  (~(is-member group bowl) who i.gs)
+    ::REVIEW
```
arthyn, line 365: lgtm

desk/sur/notify.hoon
```diff
@@ -15,8 +15,15 @@
 ::
 +$  uid  @uvH
 ::
-+$  notification
-  [=bin:hark =body:hark]
+++  notification
```
arthyn, line 18: realistically unused but we'd have to migrate state to remove

arthyn (approved):
lgtm

PR #3189, opened 2024-01-22
groups: create & join group desktop update

latter-bolden:
- Implements a modal version of the recent add group sheet on desktop. 
- Extracts some of the common logic used in both views where possible. 
- Slightly adjusts the look of the desktop root sidebar to accommodate the change.
- Tested locally against livenet ships (self & Tlon hosted).

Fixes LAND-1477

PR Checklist


latter-bolden:
Looking into E2E failures + adding tests

ui/src/components/Sidebar/AddGroupSidebarItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+    setOpen(openChange);
+  };
+
+  const navigateToModal = (path: string) => {
+    navigate(path, { state: { backgroundLocation: location } });
+  };
+
+  return (
+    <Dropdown.Root open={open} onOpenChange={onOpenChange}>
+      <Dropdown.Trigger>
```
arthyn, line 23: Needs a `default-focus` class

ui/src/groups/AddGroup/InvitedGroupsDisplay.tsx
```diff
@@ -0,0 +1,20 @@
...
+}: {
+  selectFlag: (flag: string) => void;
+}) {
+  const pendingGangs = usePendingGangsWithoutClaim();
+  return (
+    <div className="h-full">
+      <h3 className="mb-4 text-lg font-bold">Pending Invites</h3>
+      <div className="h-5/6 overflow-auto">
+        {Object.keys(pendingGangs).map((flag, index) => (
+          <GroupReference key={index} flag={flag} customOnClick={selectFlag} />
```
arthyn, line 15: Something weird going on here:

Screenshot 2024-01-23 at 5 30 08 PM

clicking blue button does this:

Screenshot 2024-01-23 at 5 30 50 PM

ui/src/components/Sidebar/AddGroupSidebarItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <SidebarItem
+          icon={<HomeIconMobileNav className="m-1 h-4 w-4" />}
+          actions={<AddIcon16 className="h-4 w-4" />}
+        >
+          <span data-testid="add-group-sidebar-button">Groups</span>
```
arthyn, line 28: Need some kind of indicator that we've been invited to a group because it's easily lost in the notifications on the right if you have a lot (also maybe that screen is going away eventually).

old:
groups

new:
Screenshot 2024-01-23 at 5 29 58 PM

arthyn (changes requested):
New group creation flow is so much better, I think we just need to resolve a few small things.

ui/src/components/Sidebar/AddGroupSidebarItem.tsx
latter-bolden: Agreed, good thought. I think sticking em near the top of the sidebar like we do on mobile would make sense here?

arthyn: that would probably work yeah, we can address later if it gets too crazy

arthyn:
btw I think the tlon block was exported without padding so it looks really big

ui/src/components/Sidebar/AddGroupSidebarItem.tsx
latter-bolden: Having their own section felt too heavy. Popped them into pinned since they're highly visible there and it's expected you'll act on them immediately.

invite

latter-bolden:
@mrozanski this should be ready for testing

mrozanski:
> @mrozanski this should be ready for testing

Reviewing and testing now

arthyn (approved):
lgtm

PR #3191, opened 2024-01-24
groups: backwards compat for scries and use proper scry paths

arthyn:
PR Checklist

This fixes an issue where scries were failing after the recent flagging PR tloncorp/landscape-apps#3168 in going to fix I realized we broke our versioning scheme. So, this implements backwards compatibility for the v0 scries and adds new scries for latest version v1. Also noticed we're missing some marks in the mark warmer.

To test I opened the app from an old frontend that still used v0 scries and everything was working, then loaded the new frontend with v1 paths and everything was still working.


desk/app/groups.hoon
```diff
@@ -431,14 +431,19 @@
...
       [%x %groups %light ~]  ``groups+!>(groups-light)
-      [%x %groups %light %v0 ~]  ``groups-ui+!>(groups-light-ui)
+      [%x %groups %light %v0 ~]  ``groups-ui-0+!>(groups-light-ui-0)
+      [%x %groups %light %v1 ~]  ``groups-ui+!>(groups-light-ui)
   ::
       [%x %groups ~]
     ``groups+!>(`groups:g`(~(run by groups) tail))
   ::
       [%x %groups %v0 ~]
+    ``groups-ui-0+!>(`groups-ui:zer:old:g`(~(run by groups) to-group-ui-0))
```
midsum-salrux, line 444: NB: very very minor nitpick but I don't like that the mark is `foo-0` while the scry is `/foo/v0/`

midsum-salrux (approved):
LGTM

PR #3195, opened 2024-01-25
hosted url: update post-migration mark augmentation

latter-bolden:
OTT, updates how we augment mark data after the backend migration for use in hosting logs.

Fixes LAND-1508


arthyn (approved):
lgtm

PR #3196, opened 2024-01-25
drag & drop: prevent getting stuck in drag state

latter-bolden:
The main issue I was able to reproduce while testing was that events would fire more often than expected which prevented us from triggering the logic to end the drag when it detects a `dragleave` event. This updates it to always cancel the drag if it leaves the window. Also prevents a drag from being initiated if the document isn't visible. 

Together this seemed to fix any issues I found locally, but we'll need to see if the bug persists after merging since it presents sporadically.

Tested locally against self & hosted ships.

Fixes LAND-1463

PR Checklist


arthyn (approved):
lgtm

PR #3198, opened 2024-01-25
chat: prevent gap in message history when viewing scrollback

latter-bolden:
If a new message comes in while you're viewing a scroll-back (search or message ref clicked), it will be added to the cache directly after the loaded message history. This can lead to a weird gap between new messages and the point in the history you were viewing. An automated refetch is fired which resolves the situation, but it can take a bit depending on network conditions leaving you in a weird state until the refetch completes.

This prevents that scenario by checking to see if you have the newest messages before adding incoming ones to the cache.

Fixes LAND-1478

PR Checklist


arthyn (approved):
lgtm

PR #3199, opened 2024-01-26
members view: fix role selection on desktop

latter-bolden:
When the members view got overhauled for mobile, the desktop version had broken role selection. This adds a corresponding dialog to be used when not on mobile.

Screenshot 2024-01-26 at 10 45 41 AM

Tested locally against a hosted ship.

Fixes LAND-1482

PR Checklist

### DIFF
--- a/ui/src/groups/RoleInput/SetRolesDialog.tsx
+++ b/ui/src/groups/RoleInput/SetRolesDialog.tsx
@@ -21,7 +21,7 @@ export default function SetRolesDialog({
 
   return (
     <Dialog open={open} onOpenChange={onOpenChange} className={className}>
-      <h3 className="mt-1 font-bold">Set Roles</h3>
+      <h3 className="mt-1 font-bold">Edit Roles</h3>
       <div className="my-5 flex flex-col items-center justify-center">
         <Avatar ship={member} size="small" className="mb-2" />
         {contact?.nickname ? (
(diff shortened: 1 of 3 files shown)


ui/src/groups/RoleInput/SetRolesDialog.tsx
```diff
@@ -0,0 +1,47 @@
...
+          </div>
+        ) : (
+          <span className="font-medium">
+            <ShipName name={member} full />
+          </span>
+        )}
+      </div>
+      <div className="max-h-[300px] overflow-auto">
+        {roles.map((role) => (
+          <div className="my-2">
```
alecananian, line 40: Looks good, only note is this should have a `key` prop on it

latter-bolden: good catch!

ui/src/groups/RoleInput/SetRolesDialog.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <span className="font-medium">
+            <ShipName name={member} full />
+          </span>
+        )}
+      </div>
+      <div className="max-h-[300px] overflow-auto">
+        {roles.map((role, index) => (
+          <div className="my-2">
+            <RoleSelect key={index} role={role} member={member} />
+          </div>
```
alecananian, line 42: `key` should be up here on the `div` element, and best to use a unique value that relates to the role, rather than an index:
```suggestion
          <div key={role} className="my-2">
            <RoleSelect role={role} member={member} />
          </div>
```

alecananian (approved):

PR #3203, opened 2024-01-26
drag & drop: fix dragging Mac screenshots into galleries

latter-bolden:
This seems to just have been an issue with spaces in the filename as expected. Uploading screenshots on Mac should now work in Safari.

While testing this, I noticed a regression with the drag & drop fix from the other day as well. There's a long outstanding bug with handling a key `dragleave` event property in Safari which was causing the drag hover state to flicker or fail. This PR also smuggles in a way to combat that while leaving in place the fix for other browsers.

Worked locally against a self-hosted ship in Safari & Chrome. Whoever tests  this, please verify drag & drop works in both gallery and chat contexts on both browsers.

Fixes LAND-1446

PR Checklist

### DIFF
--- a/ui/src/logic/DragAndDropContext.tsx
+++ b/ui/src/logic/DragAndDropContext.tsx
@@ -9,6 +9,7 @@ import React, {
 import { uniq } from 'lodash';
 import { useIsMobile } from './useMedia';
 import { createDevLogger } from './utils';
+import { isSafari } from './native';
 
 interface DragTargetContext {
   targetIdStack: string[];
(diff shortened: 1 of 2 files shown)


latter-bolden:
Note: this will need to be included in the next release since this patches the other D&D fix which is already on `staging`.

arthyn (approved):
code lgtm

arthyn:
@latter-bolden you might have to rebase if we're going to point at staging

latter-bolden:
Yep, should be in sync with `staging` now.

Was also confirming multiple drop zones works as expected, behavior seems to be in line with livenet 👍

arthyn:
@mrozanski going to deploy this to binnec since this fixes an issue introduced by one of the earlier drag fixes.

PR #3204, opened 2024-01-30
profile: public profile page

Fang-:
Work in progress, further exploration pending.

Adds an `/app/profile` agent that binds to `/profile` and serves a publicly accessible page displaying your profile details (nickname, bio, avatar, etc).

Remaining must-haves:
- [x] Put the page into Eyre cache,
- [x] Add minimally viable OpenGraph metadata,
- [x] Figure out if we want to serve on `/` or if just `/profile` is fine,
- [x] Figure out if we're ok with setting a cache entry on `/profile` preventing it from serving dynamic content indefinitely (see also https://github.com/urbit/vere/pull/391#issuecomment-1546150438)
  - [x] ~~Alternatively,~~ just go and fix this (presumably incorrect) behavior. (Done: urbit/urbit#6909, urbit/vere#603.)
- [x] Update Tlon CTA copy
- [x] Include Tlon app meta tag
- [x] Backend:
  - [x] Display sigil with profile's color
  - [x] Expose endpoint for checking if public profile is 
enabled
  - [x] Implement the "main" profile component as a widget too
  - [x] Be careful to ensure we don't end up scrying during `+on-load`.
- [x] Editor ui:
  - [x] Display en/disable toggle conditionally based on feature-flag (and probably also show if enabled already, so it can always be easily disabled)
  - [x] Use endpoint to check if public profile is enabled
  - [ ] ~~Group widgets per app~~
  - [x] Don't display a widget's backend identifier
  - [x] Should refresh/re-poll the widget list whenever the modal opens
- [x] Page styling:
  - [x] Display correctly on mobile
  - [x] Become scrollable instead of squashing tall pages
  - [x] Make sure individual widgets have unique container IDs
  - [x] Reserve space at the bottom for the Tlon CTA
  - [x] Dark mode (adaptive) page styling
  - [x] Update styling based on @urcades' feedback
- [ ] Widgets:
  - [ ] Ship some simple widgets with Groups itself
    - [x] "DM me link/button"
  - [ ] Write documentation

Known nice-to-haves:
- [x] Experimental "widget"/app data display system,
- [ ] Research possibilities for dynamic/eauth-based page content in the presence of Eyre caching.
- [ ] Grid-based layout

Fixes LAND-1529
Fixes LAND-1532
Fixes LAND-1536


arthyn (approved):
simple durable yours

PR #3205, opened 2024-01-30
LAND-1514: Types for activity agent

midsum-salrux:
PR Checklist
`desk/sur/activity.hoon`
N/A
- [x] If a new feature, includes automated tests
N/A

### DIFF
--- a/desk/sur/activity.hoon
+++ b/desk/sur/activity.hoon
@@ -47,17 +47,19 @@
   +$  unreads
     (map whom unread)
   +$  unread
-    $:  recency=time
-        count=@ud
+    $:  count=@ud
         unread=(unit [message-key count=@ud])
         threads=(map message-key [message-key count=@ud])
     ==
   +$  update
     (pair whom unread)
   --
++$  message-key
+  $%  [%time =id =time]
+      [%recency =id =time recency=time]
+  ==
 +$  whom
   $%  [%ship p=ship]
       [%club p=id:club]
   ==
-+$  message-key  [=id =time]
 --
--- a/desk/sur/activity.hoon
+++ b/desk/sur/activity.hoon
@@ -1,18 +1,18 @@
 |%
-+$  stream=(mop time event)
-+$  indices=(map concern stream)
-+$  concern
-  $%  [%group group-concern]
-      [%channel channel-concern]
++$  stream  (mop time event)
++$  indices  (map index [=stream =unreads])
++$  reads  [floor=time (set event-id=time)]
++$  derived-unreads  [time count=@ud threads=(list [time count=@ud])]
++$  index
+  $%  [%channel channel-concern]
       [%dm dm-concern]
-      [%post post-concern]
-      [%writ writ-concern]
   ==
 +$  group-concern    group=flag
 +$  channel-concern  [channel=nest group=flag]
 +$  dm-concern       =whom
-+$  post-concern     [=post-key channel=nest group=flag]
-+$  writ-concern     [=writ-key =whom]
++$  post-concern     [=message-key channel=nest group=flag]
++$  reply-concern    [=message-key target=message-key channel=nest group=flag]
++$  dm-post-concern  [=writ-key =whom]
 ::  $event: an instance of activity
 ::
 ::    $flavor: what activity generated the event and where did it happen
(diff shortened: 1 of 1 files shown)


desk/sur/activity.hoon
```diff
@@ -0,0 +1,63 @@
...
+  ::  specific occasions TBD
+  $%  [%group group-concern occasion=?(%join %kick)]
+      [%channel channel-concern occasion=?(%message %mention %reply %notice)]
+      [%dm dm-concern occasion=?(%message %mention %reply %notice)]
+      [%post post-concern occasion=?(%message %mention %reply %notice)]
+      [%writ writ-concern occasion=?(%message %mention %reply %notice)]
+  ==
+::
+::  this is similar to the way unreads are handled for DMs. in the case of
+::  chats, recency will just happen to be the same as the key time.
```
arthyn, line 42: I think this was supposed to be about `message-key`, but I'm thinking now we'll still need a way to differentiate between whether one actually needs the id vs the time so we may want to change message key to be something different like a `$%`

midsum-salrux:
```
++  unreads
  =<  unreads
  |%
  +$  unreads
    (map whom unread)
  +$  unread
    $:  count=@ud
        unread=(unit [message-key count=@ud])
        threads=(map message-key [message-key count=@ud])
    ==
  +$  update
    (pair whom unread)
  --
+$  message-key
  $%  [%time =id =time]
      [%recency =id =time recency=time]
  ==
```

Something like this?

arthyn: No so `recency` is always time and relevant for both because that's just used to sort the sidebar. 

I thought we might need discrimination between DMs and posts given what we're doing here:
https://github.com/tloncorp/landscape-apps/blob/607e5e01c5ef4de53eab6b8ffed476d2c6862c67/ui/src/chat/unreadUtils.ts#L25

but if we make `parent-time` and `time` the same as the ids then I think we're ok
https://github.com/tloncorp/landscape-apps/blob/607e5e01c5ef4de53eab6b8ffed476d2c6862c67/desk/lib/chat-json.hoon#L142

So ultimately I think we can just remove the comment and proceed as is.

desk/sur/activity.hoon
```diff
@@ -0,0 +1,57 @@
...
+      [%channel channel-concern]
+      [%dm dm-concern]
+      [%post post-concern]
+      [%writ writ-concern]
+  ==
++$  group-concern    group=flag
++$  channel-concern  [channel=nest group=flag]
++$  dm-concern       =whom
++$  post-concern     [=post-key channel=nest group=flag]
++$  writ-concern     [=writ-key =whom]
```
Fang-, line 15: This feels like the following is being implied here:

```suggestion
+$  group-concern    group=flag
+$  channel-concern  [channel=nest group-concern]
+$  dm-concern       =whom
+$  post-concern     [=post-key channel-concern]
+$  writ-concern     [=writ-key dm-concern]
```

As in, these "scopes" nest a certain way. But channel events don't appear in the group event stream also, do they? You'd construct a "mixed" stream like that on-demand, instead of through the way it's structured/stored?

desk/sur/activity.hoon
```diff
@@ -0,0 +1,57 @@
+|%
++$  stream=(mop time event)
++$  indices=(map concern stream)
++$  concern
```
Fang-, line 4: Would "scope" be misleading? Is it more like "context"?

desk/sur/activity.hoon
```diff
@@ -0,0 +1,57 @@
...
+      [%emph p=cord]
+  ==
++$  flavor
+  ::  specific occasions TBD
+  $%  [%group group-concern occasion=?(%join %kick)]
+      [%channel channel-concern occasion=?(%message %mention %reply %notice)]
+      [%dm dm-concern occasion=?(%message %mention %reply %notice)]
+      [%post post-concern occasion=?(%message %mention %reply %notice)]
+      [%writ writ-concern occasion=?(%message %mention %reply %notice)]
+  ==
```
Fang-, line 37: I suspect in practice these two types might be kind of coupled. Do we perhaps want per-occasion metadata instead of general `$content`? (Doesn't preclude us from using `$content` for the metadata where appropriate, like `%message`.)

Fang- (commented):
Some questions/thoughts.

desk/sur/activity.hoon
arthyn: I think I miscopied from %hark, but content in event was supposed to be `content=(list content)` and used to generate the message like "x messaged you this: ....". Do you mean that different flavors might want other primitives in the content type?

desk/sur/activity.hoon
```diff
@@ -1,54 +1,39 @@
 |%
-+$  stream=(mop time event)
-+$  indices=(map concern stream)
-+$  concern
-  $%  [%group group-concern]
-      [%channel channel-concern]
++$  stream  (mop time event)
++$  indices  (map index [=stream =unreads])
```
arthyn, line 3:
```suggestion
+$  indices  (map index [=stream =reads])
```

arthyn (approved):
lgtm, let's proceed as is with these and see how they play out

PR #3207, opened 2024-01-31
Tlon desktop: move Talk functionality into main app + sidebar updates

latter-bolden:
Adds the ability to access the core Talk sidebar from within the main app on desktop. Overhauls the way the Sidebar looks, adds Search for group/message lists, adds flow for desktop DM creation, etc.
Tested locally against self & hosted ships.

Fixes LAND-1513
Fixes LAND-1522

### DIFF
--- a/ui/src/components/Sidebar/MessagesSidebar.tsx
+++ b/ui/src/components/Sidebar/MessagesSidebar.tsx
@@ -63,7 +63,7 @@ export default function MessagesSidebar({
     {
       key: 'groups',
       onClick: () => setFilterMode(filters.groups),
-      content: 'Group Talk Channels',
+      content: 'Chat Channels',
       containerClassName: cn(
         'flex items-center space-x-2 rounded-none',
         messagesFilter === filters.groups && 'bg-gray-50 text-gray-800'
--- a/ui/src/components/Sidebar/MessagesSidebar.tsx
+++ b/ui/src/components/Sidebar/MessagesSidebar.tsx
@@ -1,130 +1,19 @@
 import { useState, useRef, useCallback, useContext } from 'react';
 import cn from 'classnames';
 import { debounce } from 'lodash';
-import { Link, useLocation } from 'react-router-dom';
-import AddIcon from '@/components/icons/AddIcon';
 import Filter16Icon from '@/components/icons/Filter16Icon';
 import { usePinnedChats } from '@/state/pins';
-import SidebarItem from '@/components/Sidebar/SidebarItem';
-import Avatar from '@/components/Avatar';
-import ShipName from '@/components/ShipName';
-import TalkIcon from '@/components/icons/TalkIcon';
-import MenuIcon from '@/components/icons/MenuIcon';
-import ArrowNWIcon from '@/components/icons/ArrowNWIcon';
 import {
   filters,
   SidebarFilter,
   useMessagesFilter,
   usePutEntryMutation,
 } from '@/state/settings';
-import ReconnectingSpinner from '@/components/ReconnectingSpinner';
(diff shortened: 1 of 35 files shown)


ui/src/components/Sidebar/useSearchFilter.ts
```diff
@@ -0,0 +1,79 @@
...
+  //         ? b.score + 1
+  //         : b.score;
+
+  //       return right - left;
+  //     })
+  // ), [query, searchSpace]);
+
+  // const results = useMemo(() => fuzzScores.map(result => records[result.index]), [fuzzScores, records]);
+
+  // return results;
```
arthyn, line 78:
```suggestion
```

ui/src/components/Sidebar/useSearchFilter.ts
```diff
@@ -0,0 +1,79 @@
...
+      [...groupsToRecords(groups), ...gangsToRecords(gangs)],
+      comparator
+    );
+  }, [groups, gangs]);
+
+  if (!query) {
+    return [];
+  }
+
+  return searchSpace.filter(
```
arthyn, line 52: we could use fuzzy here maybe, but this is fine for now

ui/src/components/Sidebar/MessagesSidebar.tsx
```diff
@@ -0,0 +1,224 @@
...
+              return false;
+            }}
+          >
+            <ArrowNWIcon className="text-gray-400" />
+          </a>
+        </div>
+      </SidebarItem>
+    </ActionMenu>
+  );
+}
```
arthyn, line 127: unused

arthyn (approved):
lgtm, a few suggestions/comments

ui/src/components/Sidebar/useSearchFilter.ts
latter-bolden: ya fell back to filter cause fuzz was slowing it down a bit

arthyn: ahh gotcha cool

mrozanski:
Found while testing this branch:

Desktop, Groups, chat, Clicking on a sigil of a message's author
Expected:
stay in the same browser tab/window, open DM UI in the Messages app nav section
For example: https://localhost:3000/apps/groups/dm/~nocsyx-lassul
Observed:
New browser  tab pointing to https://localhost:3000/apps/talk/dm/~nocsyx-lassul

Note: Desktop browser in small (mobile) size does the same.
Haven't tested from the mobile app.

mrozanski:
I’ve noticed that when I click on an item in Activity that opens a group chat, if then I click the back arrow at the top of the side nav, I’m taken to the Groups home with no contents in the main pane. I would expect to go back to Activity instead.

On mobile and if the activity belongs to a thread this is more noticeable because clicking on `<` sends me to the main thread of that channel, then to the group home, and then to groups.
The behavior is probably designed like this and it would make sense that I can navigate back that way in other circumstances. But when coming from the activity list, I personally prefer to go back to that list in one step, as I’d most likely be catching up with activity in general.

mrozanski:
I assume it is expected (found the same in Figma) but noting that the Groups icon is now a house instead of the one currently being used.
group-icon

mrozanski:
Messages > Filter icon next to Direct Messages (drop down)
"Group Talk Channels" doesn't sound right.

grouptalk

Maybe "Group channels"?

mrozanski:
Member links in Chat info point to talk URL

Opening a multi-DM, clicking on the three dots at the top > Chat info

multi-dm-info

mrozanski (approved):
Comments verified.
Approving.

PR #3209, opened 2024-02-01
channels: fix kick/resubscribe loop caused by flagging

arthyn:
PR Checklist

This fixes LAND-1528 which was causing a kick/resubscribe loop because our scry caused a crash. This changes the scry so that it no-ops if it doesn't find the content and also changes the `ca-abed` path to no-op if a channel is not found.


desk/app/channels.hoon
```diff
@@ -575,6 +575,7 @@
     ==
   ++  ca-abed
     |=  n=nest:c
+    ?.  (~(has by v-channels) n)  ca-core
```
arthyn, line 578: I don't think this is actually good and could end up doing weird things. I tried moving up to line 541 and doing some no-oping there but I'm not sure that's right either.

Fang-: Yeah this changes the semantics of `+ca-abed` enough to make a lot of the code here "clearly wrong".

Was gonna say that, what we want instead, is a `%u` scry for channel existence, but we already have it! Whoever is doing a scry that fails on the `got:by` here should first be doing the `%u` version of the scry.

PR #3211, opened 2024-02-04
utils: better IMAGE_URL_REGEX

pkova:
This dude had a s3 bucket with urls that always contain a `:` and talk wasn't rendering them inline. This fixes the problem.

Also, `|` is not the or condition inside a character set.


arthyn (approved):
you're a saint

PR #3214, opened 2024-02-06
chat: allow admins to delete replies

latter-bolden:
Currently the delete reply arm asserted that you were the author in order to delete a reply. It should instead also allow the action if you're an admin. This PR just updates that line accordingly.

This also explains why it was working for top level posts, but not replies (the top level post arm already had both checks).

Tested with a local ship, deleting replies from a hosted ship.

Fixes LAND-1483

PR Checklist


arthyn (approved):
lgtm

PR #3216, opened 2024-02-08
ops: move dev to staging

arthyn:
PR Checklist


patosullivan (approved):
Most of these changes are great, have noticed two issues:

1) On one of my dev moons I'm seeing a count for activity without any new activity (I think it's from a DM).

2) Groups you're attempting to join don't immediately show in the sidebar, you need to refresh the page to see that it's there and the join is in progress (or has failed).

Probably not blockers for moving to staging imo.

I'm also wondering how old/out-of-date ships will catch up with the changes to state transitions in chat/diary/heap, but we can discuss that out of band (I'm probably misunderstanding something).

PR #3227, opened 2024-02-09
channels: make reply scry response match mark type

Fang-:
Scrying for a reply message would produce a noun marked with the %channel-reply mark, but the noun itself wouldn't nest in the type of that mark.

Here, we add a cast to the appropriate type, and "unversion" the reply noun that was being produced to coerce it into the correct shape.

Resolves LAND-1531.


arthyn (approved):
lgtm, should this be going out sooner than later? if so we should target staging

PR #3236, opened 2024-02-12
scroller: invert list based on reading direction

patosullivan:
Fixes LAND-1353.

This changes our logic about whether or not to invert the list. Rather than relying on load direction, we make the invert/uninvert decision base on reading direction (whether the user perceives to be scrolling up or down). If scrolling up, we invert. If scrolling down, we "uninvert". This avoids jumpiness associated with scrolling "up" in an uninverted state, or "down" in an inverted state (see https://github.com/TanStack/virtual/issues/659). We also assume that if a user has scrolled directly to a message they are going to begin reading "down" from that message.

This also moves a few variables around for easier readability, and moves a few variables into useMemo where it seems appropriate.

Tested on local dev moons on livenet.

PR Checklist


patosullivan:
Moving to draft while I diagnose a bug when exiting/re-opening thread pane.

mrozanski:
There is a jump I get consistently when a thread view is closed. The scroll position in the main conversation jumps to an older point, presumably the oldest that was fetched so far.
(We reviewed this on a call, just adding here to ensure we test that before merging)

mrozanski:
I detected something that may be related to the thread open/close issue, or at least has a similar result.

On desktop,
If I open a chat with unreads,
click on the "n new messages since xyz. - Click to view"
takes me to the first new message (which - separate issue -  doesn't display in a consistent position, but in one case mid screen and in another near the bottom)
Then as soon as I start scrolling to continue reading below the oldest new one, the conversation jumps way back.

patosullivan:
Ok, I think we're back in a state where someone else should try banging on this to find if there are any remaining issues.

latter-bolden:
@patosullivan here's a clip of the issue I was trying to describe in standup. Steps to reproduce:

1) Be scrolled back in the history and start scrolling down (towards newer messages)
2) Stop and initiate a scroll back up (towards older messages)

You can sometimes see oscillating behavior when the scroller is fighting your movement, other times it stays in place and doesn't move despite the trackpad input (see first & last pause in the video).

I think the real world case where this might occur is if you're scanning chat history quickly and you pass a message you actually want to read/focus on. Seemed to present even with a longer, more natural interval between switching directions than what's demoed.

https://github.com/tloncorp/landscape-apps/assets/90741358/15e5a8f6-2353-49e5-9355-96ab033651cc

patosullivan:
I'm having a hard time reproducing the frozen state that @latter-bolden found, might need to pair to see exactly how to trigger that one.

I've been looking at how to reduce the jitter on the flip between inverted/uninverted, not coming up with much as of yet. Will keep trying.

mrozanski:
Testing the latest locally, on desktop, I've noticed a new issue.

1. Go to a chat channel
2. Using the scroll wheel or equivalent gesture (with a magic mouse, that is vertical swipe with one finger)
3. Perform a short movement (I swipe vertically the equivalent to 200-300 px or the hight of 4-5 lines of text) and let go, to allow momentum to carry the motion until it stops.
4. Once it stops, perform the oposite scroll movement, again swiping vertically for a short distance, and let it stop
5. One more time, now in the oposite direction, perform a short scroll movement

After two momentum / full-stop cycles in oposite directions what I observe is that the third one stops abruptly after a fraction of a second and makes two or three short jumps back and forth.

I think letting the momentum stop by itself is the trigger

latter-bolden (approved):
Nothing glaring in the code. Let's get this merged so we can feel it out in practice

PR #3238, opened 2024-02-12
icons: update to tlon icon throughout

patosullivan:
We were still using the old Groups icon in several places, this PR switches us over to the Tlon icon throughout the app. I also uploaded a new icon to GCS for us to use in the desk.docket-0 file.

Fixes LAND-1549

Tested by running `npm run build && npm run serve` locally (and pinning in safari, or installing the PWA in chrome).

PR Checklist


arthyn (changes requested):
mask-icon and safari-pinned-tab seem off, also can we do the app tile in docket as well?

arthyn (approved):
lgtm

PR #3240, opened 2024-02-13
desktop sidebar: fix double scrollbar and persist nav location when tab switching

latter-bolden:
Moves search up into the top section of the sidebar to avoid double scroller after adjusting the viewport. Also adds logic to save the navigation state for the groups and messages tabs so you can pick up where you left off when returning to the tab. 

Right now this is only enabled for messages since it's unclear how this should work for groups given the current in-group sidebar transition and back behavior. It's a one line change to add it to groups as well if we figure out a way to go that route.

Fixes LAND-1553
Fixes LAND-1554

PR Checklist


mrozanski (approved):

patosullivan (approved):
Looks good, saving the nav state will be nice.

PR #3241, opened 2024-02-13
chat: fix copying root message from thread view

latter-bolden:
Updates the logic for calculating ref paths in chat reply options. If the message being copied is the root, link to it as a top level message rather than as a thread reply.

Also updates ref display logic to account for previously malformed paths. Old broken ones will show correctly now as well.

Tested against a local ship. For review, worth checking that normal messages, thread replies, and thread reply parents all work as expected.

Fixes LAND-1557

PR Checklist

### DIFF
--- a/ui/src/components/References/ContentReference.tsx
+++ b/ui/src/components/References/ContentReference.tsx
@@ -82,13 +82,14 @@ function ContentReference({
     if (app === 'chat') {
       const idWrit = segments[2];
       const idReply = segments[3];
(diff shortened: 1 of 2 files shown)


latter-bolden:
Not sure if we want this targeted against staging?

arthyn:
@latter-bolden no, deploy after next needs to go out quickly too so it can just go then imo

patosullivan (approved):

PR #3249, opened 2024-02-15
unreads: use chat store unreads for sidebar count

patosullivan:
Fixes LAND-1574.

We were using the unread state synced from the backend for the unread count on the messages tab button. This caused a delay in updating the count due to the delay in the updated state.

If we switch to using the chat store unreads, the count updates immediately (we use this same state to update the unread "blue dots").

Tested locally with livenet moon and planet messaging each other, and with pinning some active group chat channels.

PR Checklist


arthyn (approved):
lgtm, I had started to do the same thing last week but @latter-bolden and decided on a lighter fix, but I think this is more correct.

PR #3250, opened 2024-02-15
refs: fix issue with clicking links in refs

patosullivan:
Fixes LAND-1409.

Fixes the issue where clicking a link within a ref would both follow the link and navigate to the ref.

Fix consists of making sure the user is actually clicking on the ref itself, not the link.

Further explanation in in-line comments.

Tested locally on livenet dev moons by making some refs with links in them.

PR Checklist


latter-bolden (approved):
Looks good & makes sense. 

Worked when tested locally against a hosted ship

PR #3254, opened 2024-02-16
chore: fix some monorepo issues

patosullivan:
- adds the build step back in for the shared package. we're going to want `npm run dev` in there so we can update files in there while working in dev on either app, afaict. updates gh workflows appropriately.
- fixes the import sort order issue. prettier rc had been updated to use `@trivago/prettier-plugin-sort-imports` and wasn't configured with our `@/` root alias in mind. This made eslint mad every time you opened a file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,9 @@ jobs:
       - name: Install Dependencies
         run: npm ci
 
+      - name: Build Shared Packages
+        run: npm run build:shared
+
       - name: Run Lint
         run: npm run lint --workspaces --if-present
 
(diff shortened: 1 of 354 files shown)


alecananian (approved):

alecananian:
Thanks! `tsup.config.ts` got removed in a merge: https://github.com/tloncorp/landscape-apps/blob/b3700b9f00da58ee358131487721f68bbe27c06f/packages/shared/tsup.config.ts
But the inline options on the `build` command achieve the same thing - looks good 👍

latter-bolden (approved):

PR #3255, opened 2024-02-16
native: add tamagui

dnbrwstr:
This PR performs initial setup of tamagui in the native mobile app. Tamagui components are kept in a separate package (`ui`) to help keep it encapsulated, since the longer term goal is to use it in both the web and native apps.

How to test:
- Build the mobile app
- Try using `Button` from `@tloncorp/ui`


dnbrwstr:
Converting to draft while I resolve the type error, which weirdly didn't show up in VSCode.

dnbrwstr:
Ok, should be good.

PR #3257, opened 2024-02-17
mobile: add new preview profile for building apps with separate bundle ID

alecananian:
PR Checklist

Changes
- Configures Android to have `preview` and `production` product flavors with differing app icons and names
- Configures iOS to have `Landscape-preview` and `Landscape` app targets with differing app icons and names
- Updates Expo/EAS configs to support `preview` and `production` profiles
- Updates CI workflows to allow user to choose profile while triggering jobs
- Disables the push notification permission prompt for now in this repo so it doesn't conflict with the existing version of `tlon-mobile` on testers' devices

Unfortunately, I had to change the `engines` value in `package.json` because Expo's build servers don't support Node v20 yet. I set the Node engine as `>=18.18.0` so we can continue using v20 locally and in other build processes.

Tested by running the `eas build` commands locally. New bundles are available in Google Play and Apple App Store already.


mrozanski:
> Hey, could we include the latest changes from tlon-mobile (notifications linking fix) on here too?

We can, although since we still need to add push notifications to this version and we're not planning to use this one for production releases yet, I think it's OK to add the latest from tlon-mobile in a separate PR.

mrozanski:
Tracking @patosullivan 's suggestion (syncing changes from the other repo) in Linear. Thanks for surfacing.

PR #3262, opened 2024-02-20
chore: add npm run build:shared to postinstall on root package

patosullivan:
Without this you could run into eslint/ts errors on importing the package (until you manually build the package or run the dev script for it).


patosullivan:
> Makes sense! My only note is to now remove the build step from any CI workflows where we're already installing dependencies, otherwise it will be built twice.

good point! just removed them.

PR #3267, opened 2024-02-21
channels: make sure channel data is up to date when we have unreads

patosullivan:
Fixes LAND-1556.

getQueryData was always returning undefined in my testing locally, so rather than only refetching if we have cached data for a query, we just fetch the query if we have an unread for the channel (since we know we *will* use that data later).

I also noticed that channelKey in state/channel/keys.ts was using 'channel' rather than 'channels'. 'channel' isn't used as a queryKey anywhere afaict, I assumed this was supposed to be 'channels', so I changed that.

Tested locally with livenet moons.

PR Checklist

### DIFF
--- a/apps/tlon-web/src/state/bootstrap.ts
+++ b/apps/tlon-web/src/state/bootstrap.ts
@@ -3,13 +3,14 @@ import _ from 'lodash';
 
 import api from '@/api';
 import { useChatStore } from '@/chat/useChatStore';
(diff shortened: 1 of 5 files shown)


patosullivan:
Just realized this won't work for dms/clubs too. Will reopen when it does.

patosullivan:
Ok, now works for dms/clubs too.

PR #3268, opened 2024-02-21
channels: redirect on bad id in thread URL

patosullivan:
Fixes LAND-1450

Not clear where the bad link was generated, but this is where the error was occurring.

Tested on local livenet moons.

PR Checklist

### DIFF
--- a/apps/tlon-web/src/chat/ChatThread/ChatThread.tsx
+++ b/apps/tlon-web/src/chat/ChatThread/ChatThread.tsx
@@ -165,7 +165,7 @@ export default function ChatThread() {
   }, []);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):
Looks good, worked locally with chat threads.

Not worth hunting down every spot where we'd want to gracefully redirect on a bad URL, but probably worth updating `DMThread` as well.

patosullivan:
> Looks good, worked locally with chat threads.
> 
> Not worth hunting down every spot where we'd want to gracefully redirect on a bad URL, but probably worth updating `DMThread` as well.

I checked DMThread and we don't do a similar call to `bigInt()` there, so we shouldn't throw in that case afaict

PR #3274, opened 2024-02-21
channels: support non-knot text in searches

Fang-:
On the backend, we were taking the search query directly out of the path. Assuming all paths are well-formed, this meant we had no way of receiving text that didn't fit into the path's `@ta` (aka knot) character limitations.
Here, we first try to decode the query part of the search scry path as a knot-encoded string. Only if that fails do we take the path element as-is. (That way, we maintain backwards compatibility, and continue supporting trivial search cases trivially.)

On the frontend, we were injecting the search query into the scry url directly, even if it contained non-url-safe characters. Something further down the stack would url-encode the url string, but the backend wasn't prepared to handle this (and shouldn't have been). Here, we add a "stringToTa()" utility for knot-encoding strings, and make sure to do that to the query before putting it into the url.

With the combination of these, we now support arbitrary text (including unicode!) in the search.

Fixes LAND-1581.


Fang-:
There's some frontend bug, where searching for "Abc" returns results for both "Abc" and "abc" (lowercase), but searching for "abc" only returns the "abc".

I thought that was introduced by this PR, but it happens on both old and new backends, and using the frontend from `develop`. I wanted to try `git bisect`ing it, but ran into monorepo migration shenanigans, and decided to punt.

arthyn (approved):
lgtm, let's see if we can use [@urbit/aura-js](https://github.com/urbit/aura-js). This PR does something similar to what you've done but they are different: https://github.com/urbit/aura-js/pull/6

PR #3281, opened 2024-02-23
mobile: generate profile tab icon from user sigil

alecananian:
PR Checklist

Changes
- Updates profile tab icon to be an Urbit sigil for the logged in user's ship
- Fixes Tamagui dark mode

Screenshot 2024-02-22 at 10 07 04 PM

Note: very simple implementation for now that doesn't have all the different sizing and color options we'll need. We can address the sizing when we start to use the sigil component on other tabs, and the colors when we get contacts syncing hooked up.


dnbrwstr:
This looks great! What do you think about moving the sigil wrapper to the shared ui lib? Seems like a pretty core component.

apps/tlon-mobile/src/components/UrbitSigil.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  const sigilXml = useMemo(
+    () =>
+      sigil({
+        point: ship,
+        detail: 'none',
+        size: 12,
+        space: 'none',
+        foreground: '#ffffff',
+        background: isDarkMode ? '#333333' : '#000000',
```
dnbrwstr, line 25: Would be good to hook this up in a way that uses Tamagui themes/tokens system. `@tamagui/core` exports a  `useTheme` hook, which you can use to get current theme values, and there's also `@tamagui/get-token`, which lets you pull token values for sizing/spaces. You can also wrap this component in a Tamagui component that will resolve input tokens (see the `Icon` component).

apps/tlon-mobile/src/components/UrbitSigil.tsx
```diff
@@ -0,0 +1,39 @@
...
+      }),
+    [ship, isDarkMode]
+  );
+
+  return (
+    <View
+      style={tailwind(
+        'p-1.5 h-5 w-5 rounded flex items-center justify-center bg-black dark:bg-tlon-black-80'
```
dnbrwstr, line 33: Same as above, can we style using Tamagui?

dnbrwstr (changes requested):
Looks good, just a few comments on styling consistency.

dnbrwstr:
Although re-reading the above, maybe you're just wanting to hold off on styling w Tamagui until we build out a bit more -- fine with me if that's the case.

dnbrwstr:
> For this, do you think we should move the memoized XML to the shared lib? The full component probably doesn't make sense because it's React Native-specific.

Ultimately, the goal is for the `ui` package to export interfaces for core components, so that they'd be importable in other contexts (for example, another prototype/testbed app). It could export a sigil component that in the long run could be backed by `.native.ts` and `.web.ts` implementations, but for now could be native-only. 

That's a secondary goal though, so feel free to ignore if it's laborious. 

> That was my initial thinking, but lemme dig into the Tamagui side a little today and see what I can do here 🙂

Cool :)

dnbrwstr (approved):

PR #3282, opened 2024-02-23
upload: handle no publicUrlBase

jamesacklin:
Fixes a TypeError in upload.ts when uploading to a S3 bucket without a publicUrlBase set in the %storage config. 

Uploading an image was throwing a `Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL` error. I interpret this to mean that we were failing to instantiate a variable with a URL type because the `new URL()` function takes a single argument (which cannot be null). Open to other, cleaner fixes.

Tested locally with the Vite dev server proxied against my livenet planet.

PR Checklist


patosullivan (approved):
lgtm, thanks!

PR #3284, opened 2024-02-26
mobile: use single webview & handle native/webview navigation ipc

latter-bolden:
This moves from a webview per-tab architecture to a singleton webview overlay simplifying state & navigation required for moving between the two runtimes. Back buttons, internal tab navigations, etc. should work correctly now.

There are a few remaining items related to this work, but I'd like to get it merged so we can try it against the test build. I logged issues in Linear for tracking the following remaining items:
- manage account & branch wer functionality still need to be refactored
- theres an added `dev:ios:local` script that simplifies running against the local dev server. Needs to be fixed, hardcoded right now
- the webview context got a bit complicated and should probably have its state management simplified

Fixes LAND-1504


latter-bolden:
@alecananian addressed the points we spoke about this afternoon  👍

PR #3288, opened 2024-02-27
chat: add edit capability to group chat channels

patosullivan:
Adds the ability to edit chat channel posts.

Looks like this:

https://github.com/tloncorp/landscape-apps/assets/1221094/edbb39d2-d607-42a8-a1ea-7f5bcedc3be5

Tested locally with livenet dev moons.

Fixes LAND-378

PR Checklist


arthyn:
the menu keeps growing lmao

patosullivan:
Going to move the `edited` key over to seal rather than kind-data now.

patosullivan:
Ok, this should be good to go. Rather than surface up `rev`, I surfaced it as `edited`. If `rev` is any number other than `0`, then the message has been edited and we set `edited` to true in `channel-json` while encoding the json.

I also realized that I didn't account for replies when I originally posted the PR, so this now accounts for replies.

You should be able to either long press and press the edit button on mobile, or use the edit button in the message options on desktop, or use the up arrow on desktop to edit your last message.

arthyn (commented):
frontend changes look good, will discuss backend @Fang-

patosullivan:
After discussing with Mark, we decided that we need to version our scry and subscription paths so that we don't break API compatibility for changes like this. Now working on:

- creating new "v1" path for things as they are on livenet today
- creating new "v2" path for posts with the "rev" number
- making sure we fallback to the our current-world unversioned paths if no version is specified (i.e., default to version 1)

We also decided that surfacing up the revision number to the frontend is a better way to go, rather than `edited: "true" | "false"`, so I'll be making that change as well.

patosullivan:
Ok, with Mark's changes, this should be good to go @mrozanski, if you want to work on automated tests.

desk/lib/channel-utils.hoon
```diff
@@ -27,16 +31,45 @@
...
+++  s-posts
+  |=  =posts:c
+  ^-  simple-posts:c
+  %+  gas:on-simple-posts:c  *simple-posts:c
+  %+  turn  (tap:on-posts:c posts)
+  |=  [=id-post:c post=(unit post:c)]
+  ^-  [id-post:c (unit simple-post:c)]
+  [id-post ?~(post ~ `(s-post u.post))]
+::
+++  suv-posts
```
arthyn, line 43: 🚙

desk/app/channels.hoon
```diff
@@ -1446,10 +1537,38 @@
       =/  post  (get:on posts.channel `@da`time)
       ?~  post  ~
       ?~  u.post  `~
-      (ca-peek-replies id.u.u.post replies.u.u.post rest.pole)
+      (ca-peek-replies-1 id.u.u.post replies.u.u.post rest.pole)
+    ==
+  ::
+  ++  ca-peek-replies-0
+    |=  [parent-id=id-post:c replies=v-replies:c =(pole knot)]
```
arthyn, line 1544: I think these are giving the new type not the old

arthyn (changes requested):
I think `ca-peek-replies-0` needs updating. My only other question is are we still breaking compat by changing the mark names @Fang- ?

desk/app/channels.hoon
patosullivan: It looks like they always looked like this. Maybe that was wrong, though?

arthyn: right but we changed what those types are, the old types are now the `simple-` which we aren't giving here. just need to do something similar to what we did for `ca-peek-posts-0`

patosullivan: But v-reply has always had the rev on it, and these were using v-reply (except for the individual reply scry that I switched over). Maybe we just didn't notice because we don't actually use these anywhere.

If anyone was using these, they were getting the `rev`, in other words.

patosullivan: Yep

patosullivan: Yeah, it fails on v1 and v0 because of the mark. Will fix.

patosullivan: This should be fixed in the latest commit.

arthyn (approved):
lgtm, we should ask spandrell if this breaks anything with trill

PR #3290, opened 2024-02-27
scroller: revert reading direction inversion changes

patosullivan:
This reverts the changes made to invert the scroller direction based on reading direction, from #3236

Also adds more logging so we can get some more insight into what is changing and when.

Dan and I used the afternoon to try to find what was causing the issue in iOS, couldn't find anything within a reasonable amount of time.

So LAND-1353 is now unresolved.

We ought to pull in the @tanstack/virtual library to our shared packages so we can get some clearer debugging around what's happening in virtual and maybe modify to our needs. I looked at pulling it in, but the package itself uses workspaces and pnpm, so not super straight forward to pull in directly. We should create a ticket for it IMO.

@dnbrwstr let me know if you think we should add logging anywhere else (or feel free to make changes).

For the original LAND-1353 issue:
Dan and I discussed another way we can try to guess the correct inverted/not inverted state for the scroller, by managing smaller "scroll pages" in the parent components, only allowing the scroller to know about small subsets of what we actually have in the cache (maybe 30 or so at a time), so that we can invert based on the loading direction the way the scroller was already working prior to my changes based on reading direction.

### DIFF
--- a/apps/tlon-web/src/chat/ChatScroller/ChatScroller.tsx
+++ b/apps/tlon-web/src/chat/ChatScroller/ChatScroller.tsx
@@ -174,7 +174,6 @@ export interface ChatScrollerProps {
   isLoadingOlder: boolean;
   isLoadingNewer: boolean;
   replying?: boolean;
-  inThread?: boolean;
   /**
    * Element to be inserted at the top of the list scroll after we've loaded the
    * entire history.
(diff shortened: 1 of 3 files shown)


dnbrwstr (approved):
Cool, this looks good to me! I've added a separate issue at [LAND-1630](https://linear.app/tlon/issue/LAND-1630/make-chatscroller-more-robust-to-changes) to track other improvements we talked about for `ChatScroller`.

PR #3293, opened 2024-02-29
channels, chat: faster search ux

Fang-:
Primarily, we give channels and chat a new search api endpoint, which bounds its requests by messages _searched_, rather than the amount of search results _found_.

The existing search api is bounded by the amount of results returned. This makes searching for something with few results a pathological case: we will continue searching _all_ messages in a channel until we've hit the desired amount of results, or exhausted the entire backlog. This results in search being slow in unpredictable ways.

We add a new search api (that is, a new scry endpoint) where the search is bounded (and paginated) by the total amount of messages searched, rather than the subset that gives results. This way, we can bound the amount of computation that happens for any given "page" of search, making for more predictable ux. It also explicates the distance to which the backlog has been searched by including a "last-searched" identifier (accurately: the `time` key in our `mop`s). We use this to make search progress legible to the user, even in cases where no results have been found yet.

image

To maintain a simple api, we bound by the number of top-level messages, rather than messages plus replies. In contexts where replies are expected to make up the bulk of content, the search bounds/window should be adjusted downwardly to compensate.

Note that we leave the old search api in place, as outside clients like the mobile app might still be using it.

The current configuration, which searches 500 messages at a time, takes 1-2 seconds per request. This feels like a fine balance between request overhead and speed of results/search status. Tweaking this number is trivial, if we so desire.

I am also open to different names for the search "scam", the new search result structure encapsulating the previous "scan". (But don't get carried away bikeshedding: ultimately, the names don't matter.)

Remaining work:
- [x] Apply the frontend improvements for channel search to dm search as well.
- [x] Fix (or work around) a behavior in our dev server: it turns urls of the shape `/abc//def` into `/adc/def`, even though empty path segments are valid (and not entirely uncommon in urbit apis).
  - Right now, the backend api has to do something dumb and non-standard to support use through the dev server.

Closes LAND-1612.

### DIFF
--- a/apps/tlon-web/src/channels/ChannelSearch.tsx
+++ b/apps/tlon-web/src/channels/ChannelSearch.tsx
@@ -1,16 +1,30 @@
+import { useEffect } from 'react';
 import { useParams } from 'react-router';
 
 import ChatSearch, { ChatSearchProps } from '@/chat/ChatSearch/ChatSearch';
+import { CHANNEL_SEARCH_RESULT_SIZE } from '@/constants';
 import { useChannelSearch } from '@/state/channel/channel';
 
 export default function ChannelSearch(
   props: Omit<ChatSearchProps, 'scan' | 'query' | 'isLoading' | 'endReached'>
 ) {
   const { query } = useParams<{ query: string }>();
-  const { scan, isLoading, fetchNextPage } = useChannelSearch(
-    props.whom,
-    query || ''
-  );
+  const {
+    scan,
+    isLoading,
+    fetchNextPage,
+    depth,
+    oldestMessageSearched,
+    hasNextPage,
+  } = useChannelSearch(props.whom, query || '');
+
+  useEffect(() => {
+    const numResults = scan.toArray().length;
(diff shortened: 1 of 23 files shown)


mrozanski:
Blazing fast! (compared to what we've gotten used to)
Works great, I found it hard to break.

Got one: If the search term starts with a `%` it works unless I add a space and a second word. Then it seems to add literal `%20` for spaces and fails to find results.

search-with-cen

mrozanski:
Seems like it's not searching inside code blocks.
search-code

Fang-:
> Seems like it's not searching inside code blocks.

I updated search to search into inline code, codeblocks, and urls also.

> If the search term starts with a % it works unless I add a space and a second word.

Something about the `%` was tripping it up. We were just dumping user-provided strings into the router url, instead of encoding them first. Should be fixed.

mrozanski:
The `%tlon app ` case is solved.
There are other cases like `%path /` still causing issues.
cenpath

I have no objection with merging and treating this separately.

arthyn (approved):
lgtm

PR #3296, opened 2024-02-29
notifications: post daily update on rivfur's state to a specific channel

patosullivan:
still needs:
- [x] ~make sure the daily timer actually works and we're actually calling the thread from the agent~ timer should work, no need for the thread anymore either.
- [x] switch out ship ~nalseg-banmud-finned-palmer for ~rivfur-livmet
- [x] point to the homestead channel rather than my empty channel

Tested with local livenet moons

PR Checklist


desk/mar/provider-state.hoon
```diff
@@ -0,0 +1,12 @@
+/-  n=notify
```
arthyn, line 1: if we're only ever using these as nouns we should probably just use the noun mark

desk/app/notify.hoon
```diff
@@ -64,9 +64,11 @@
       pass  pass:io
   ::
   ++  on-init
+    =/  tomorrow=@da  (add now.bowl ~d1)
     :_  this
     :~  (~(watch-our pass:io /hark) %hark /ui)
         (~(wait pass:io /clear) (add now.bowl clear-interval))
+        (~(wait pass:io /daily-timer) tomorrow)
```
arthyn, line 71: we need to also do this in on-load since most people will already have this agent installed. just need to make sure we don't do both. I'm also wondering if we should do the provider check here instead of in the the thread.

patosullivan: move it to a poke and then have the agent poke itself, in other words?

arthyn: yeah move to poke, and then when you receive the poke check to see if we're already waiting, and then have both init and load do the poke.

arthyn:
might want to update your checkboxes

arthyn (approved):
lgtm

PR #3298, opened 2024-03-01
leap: fix crash on open

patosullivan:
Fixes LAND-1640, a new bug that was introduced on develop yesterday.

Just needed to pass an anonymous function here that returns the component, rather than the component itself.

Tested locally on livenet moons.

PR Checklist


jamesacklin (approved):
Whoops! Thank you.

PR #3299, opened 2024-03-01
groups: better data loading flow

arthyn:
This needs to be tested more thoroughly because it might have far reaching implications. Fixes LAND-1639.

`keepPreviousData` was biting us because if you were coming from another group that would be the data in the query, but then we'd check `canRead` and it would be false because we were comparing against a different group. This caused a redirect to occur to the group home. Then from the group home we try to do the channel redirect and wouldn't find a good candidate so we'd redirect to `${group}/channels`.

Similarly, `placeholderData` didn't seem to be updating every time the flag changed, so I removed entirely in favor of just defaulting to existing data in `useGroups`.

Tested by running locally on both desktop and mobile, navigating to different channels in messages tab and also random groups pages to make sure we weren't missing data or seeing anything weird.

PR Checklist

### DIFF
--- a/apps/tlon-web/package.json
+++ b/apps/tlon-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tlon-web",
-  "version": "5.5.0",
+  "version": "5.7.0",
   "private": true,
   "scripts": {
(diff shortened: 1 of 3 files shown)


patosullivan (approved):
Makes sense, we ought to get it on wannec for testing

PR #3301, opened 2024-03-01
deps: upgrade react-router

latter-bolden:
OTT

Just jumps to the latest minor version. Needed for a search fix that adjusts how query params are decoded. Didn't notice any regressions while playing with it locally.

Fixes LAND-1642


patosullivan:
Seems fine, but what's up with CI?

patosullivan (approved):
Lgtm

PR #3303, opened 2024-03-03
mobile: implement groups tab in RN

alecananian:
Updates mobile app to have the Groups list implemented in React Native. Webview visibility is toggled when you navigate between the Home tab ahd the others.

### DIFF
--- a/packages/ui/src/index.ts
+++ b/packages/ui/src/index.ts
@@ -1,14 +1,3 @@
-// import { FontFamilyTokens } from "@tamagui/web";
-import { FontFamilyTokens } from "@tamagui/core";
-import {
-  getTokenValue,
-  SizeTokens,
-  SpaceTokens,
-  Token,
-  getConfig,
-  Variable,
-} from "tamagui";
-
 export { useTheme } from "tamagui";
 export * from "./components/Button";
 export * from "./components/Icon";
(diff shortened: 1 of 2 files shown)


packages/ui/src/index.ts
```diff
@@ -1,6 +1,57 @@
...
+export * from "./components/ListItem";
 export * from "./components/TamaguiProvider";
+export * from "./components/Text";
+export * from "./components/View";
 export * from "./tamagui.config";
-export * from "./components/Icon";
 export * from "./components/UrbitSigil";
-export { ZStack, View } from "tamagui";
+
+export const token = {
```
dnbrwstr, line 22: I didn't end up using these helpers -- I'd remove, I'm not sure this is how we'd want them shaped anyway.

dnbrwstr: (Referring to `token`, `getFont`, and `getFontTokenValue`.

apps/tlon-mobile/src/components/GroupListItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+        imageUrl={model.iconImage}
+      />
+      <ListItem.MainContent>
+        <ListItem.Title>{model.title}</ListItem.Title>
+        {/* <ListItem.Subtitle>{model.latestPost?.text}</ListItem.Subtitle> */}
+      </ListItem.MainContent>
+      <ListItem.EndContent>
+        {model.lastPostAt ? (
+          <ListItem.Time>
+            {new Date(model.lastPostAt).toLocaleDateString()}
```
dnbrwstr, line 36: This should be formatted date, and probably also be memoized -- not critical for this PR though.

apps/tlon-mobile/src/db/queries.ts
```diff
@@ -0,0 +1,140 @@
...
+  Group,
+  SchemaMap,
+  SchemaModel,
+  SchemaName,
+  SchemaValue,
+} from './schemas';
+
+// Model queries
+
+export interface GroupQuerySettings {
```
dnbrwstr, line 13: Just a note: I ultimately want these to query functions to follow a more consistent pattern, but haven't found the right answer yet.

dnbrwstr (approved):
Looks good to me! A few small comments.

Also, what do you think about mocking the api and adding tests for the sync/query methods? Or what's your take on where we add tests as we build up the foundation?

PR #3310, opened 2024-03-05
mobile: revert native groups list for now

latter-bolden:
OTT

Going to hold off on this until it's fully usable as a drop in replacement. This will allow us to cut over to the new mobile app sooner and build out from there.

Fixes LAND-1657

PR #3311, opened 2024-03-05
mobile: pull in subsequent work from the old repo

latter-bolden:
OTT

Think I got everything except the version bumps and most android/ios project file changes which looked innocuous.

Tested what I could on simulator against a local ship.

Fixes LAND-1646


latter-bolden:
Conflicts should evaporate after #3310 lands

PR #3312, opened 2024-03-05
activity: clearing notification indicators from reading unreads

arthyn:
Fixes LAND-1649 by sending reads to %hark any time we read unreads and they contain notifications. Still WIP because I'm trying to add tests.

This behavior can be removed once %activity is integrated, because it should do this for us.

PR Checklist


arthyn:
@Fang- going to need help dealing with negotiate with these tests

patosullivan (approved):
Makes sense, lgtm

PR #3314, opened 2024-03-06
411: add updated strandio

patosullivan:
OTT, we need a new, up-to-date copy of strandio for 411 compatibility.


patosullivan:
The vitals connection check thread seems to break with the new strandio.

arthyn:
going to close this in favor of re-releasing with a lib/strandio that's compatible with both 412 and 411

PR #3315, opened 2024-03-06
[draft] initial idea for how to structure ephemeral deployment moons

dosullivan:
Just trying to structure out how we could potentially start up ephemeral moons for development branches. 

To make this work we need a hoon thread to run through click to generate a moon. We then need to write that output to a file, then extract the moon name and key, and then use that to boot the one-off moon. One problem I see with the current approach is that the bash script copied/ran on the VM host is very much a "one shot" type thing where the script is just ran one time, similar to how our current `deploy.sh` script is ran. Need to determine how we can actually see if the moon is successfully booted before moving forward with the deployment script itself.

### DIFF
--- a/.github/helpers/create_moon.sh
+++ b/.github/helpers/create_moon.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+# this script creates a moon for an ephemeral deployment on a ship.
+# assumes gcloud credentials are loaded and gcloud installed.
+
+ship=$1
+zone=$2
+project=$3
+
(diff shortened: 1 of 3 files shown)


arthyn (approved):
I think this generally makes sense. @Fang- reading this https://github.com/urbit/tools/wiki/conn.c-Usage-Guide it seems like the thread needs to be inlined and stripped of some of the opening imports and stuff since that's included in khan-eval right?

arthyn:
going to just push this through and we can make another PR with actual deploys

PR #3317, opened 2024-03-07
perf: speed up sidebar sorter

patosullivan:
I was getting annoyed with how slow the sidebar can be, so I did a performance profile and root caused that this particular function was pretty heavy.

Saw a marked improvement in performance for this function by just pre-computing the values for comparison before running the comparison and eliminating the use of .reverse().

Note that this isn't a cure-all, I just decided to take a weed wacker to the tallest weed on a cursory look.

tested locally by running the dev server against my livenet planet.

No idea if there's a linear ticket for sidebar perf.

PR Checklist


arthyn (approved):
lgtm

PR #3319, opened 2024-03-07
e2e: remove specs interdependency

mrozanski:
005-mobile-chat-options.spec.ts expects that a chat message from spec 001 is present. This can cause random failures if spec 001 has not finished before 005 starts. Also, specs should be independent so that they can be run individualy.

This change consists in adding one step to spec 005 where the chat message is created. The test then continues from the 2nd ship by reacting to that message.

Tested locally by running `npm run e2e:debug`
We should wait for CI to run off of this branch to confirm.

PR Checklist

### DIFF
--- a/apps/tlon-web/e2e/005-mobile-chat-options.spec.ts
+++ b/apps/tlon-web/e2e/005-mobile-chat-options.spec.ts
@@ -2,31 +2,51 @@ import { expect, test } from '@playwright/test';
 
 import shipManifest from './shipManifest.json';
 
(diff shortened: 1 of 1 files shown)


mrozanski:
Test failing for obscure reason.

```
005-mobile-chat-options.spec.ts:14:5 › Post chat message as group guest
Error: Timed out 5000ms waiting for expect(received).toBeVisible()
Call log:
  - expect.toBeVisible with timeout 5000ms
  - waiting for getByText('hi there')
```

Investigating.

latter-bolden:
This should still be addressed but isn't working as is. Going to close for now, referenced issue will remain open in Linear.

PR #3320, opened 2024-03-08
profile: make sure text in the bio gets escaped

Fang-:
While injecting the bio text as raw html isn't against the spirit of profiles, it is inconsistent with the way the bio text gets rendered in groups.

(Yes, this diff is sufficient for escaping the bio text. Normally, it gets handled by the sail->html conversion, but we are manually constructing sail elements here (to get text-only nodes). In doing so, we made the mistake of adding a string (`$tape`) of "one" "character" (the whole `@t`), rather than something sane.)


arthyn (approved):
lgtm

PR #3321, opened 2024-03-08
web/nav: channel navigator on group row long-press, channel header tap

jamesacklin:
Adding a draft PR that fixes LAND-1664 (feature-complete) and fixes LAND-1666 (in progress). See the linked Linear issues for the full interaction spec.

- Reappropriates GroupInfo (a route-based modal we abandoned) to contain all current group options, plus a ChannelList.
- Adjusts ChannelList to accept a mandatory flag prop, which avoids "group" collision when opening the menu from MobileRoot. (Simply accepting the group in scope resulted in an incomplete list of channels outside the group context.)
- Adjusts ChannelList to return a flat array of channel navigation row elements, unbounded by the Virtuoso scroller. (This was causing lots of layout problems and deemed unnecessary.)

Contains some sketchy useEffect usage to play nice with long-pressing the GroupsSidebarItem row. @latter-bolden and I dug into it and keeping an exhaustive list of dependencies caused re-rendering issues when dismissing the GroupInfo sheet.

Also uses state to properly navigate the user to other overlay routes from GroupInfo without prematurely closing the sheet and preserving background location.

Tested locally with the Vite dev server proxied to a livenet planet. Will test using the native client connected to the dev server.

---

PR Checklist

### DIFF
--- a/apps/tlon-web/src/app.tsx
+++ b/apps/tlon-web/src/app.tsx
@@ -93,6 +93,7 @@ import {
   useTheme,
 } from '@/state/settings';
 
+import ChannelInfo from './channels/ChannelInfo';
 import ChannelVolumeDialog from './channels/ChannelVolumeDialog';
 import MobileChatSearch from './chat/ChatSearch/MobileChatSearch';
 import ReportContent from './components/ReportContent';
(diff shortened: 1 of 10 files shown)


arthyn:
is this still getting worked on?

PR #3322, opened 2024-03-08
mobile: connect notifications, deep links, and account management

latter-bolden:
Adds support for detecting and handling in app navigation from notifications and deep links. Account management, which requires opening an external hosting page, is also wired up.

Tested by triggering testflight builds for the production app. Once our dev environment is fully configured for branch/notifications on the preview build this could also be tested there.

Fixes LAND-1624

PR Checklist


apps/tlon-mobile/src/components/AuthenticatedApp.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { WebviewPositionProvider } from '../contexts/webview/position';
+import { WebviewProvider } from '../contexts/webview/webview';
+import { useDeepLinkListener } from '../hooks/useDeepLinkListener';
+import useNotificationListener from '../hooks/useNotificationListener';
+import { TabStack } from '../navigation/TabStack';
+import WebviewOverlay from './WebviewOverlay';
+
+export interface AuthenticatedAppProps {
+  initialWer?: string;
```
dnbrwstr, line 11: Can we rename all instances of `wer`? To something like `deepLinkPath` or `lurePath` (not totally clear on what it actually is).

apps/tlon-mobile/src/hooks/useNotificationListener.ts
```diff
@@ -0,0 +1,69 @@
...
+              content: { data },
+            },
+          },
+        } = response;
+        if (actionIdentifier === 'markAsRead' && data.channelId) {
+          markChatRead(data.channelId);
+        } else if (actionIdentifier === 'reply' && userText) {
+          // TODO: Send reply
+        } else if (data.wer) {
+          setGotoPath(getPathFromWer(data.wer));
```
dnbrwstr, line 44: We should extract the path from the `wer` when it's initially received so we only need to do it once.

apps/tlon-mobile/src/hooks/useNotificationListener.ts
```diff
@@ -0,0 +1,69 @@
...
+  useEffect(() => {
+    if (gotoPath && webviewContext.appLoaded) {
+      console.debug(
+        '[useNotificationListener] Setting webview path:',
+        gotoPath
+      );
+      webviewContext.setGotoPath(gotoPath);
+      const tab = parseActiveTab(gotoPath) ?? 'Groups';
+      navigation.navigate(tab, { screen: 'Webview' });
+      setGotoPath('');
```
dnbrwstr, line 66: Intent would be clearer if we used null for empty value (unless there's another reason for the empty string?).

apps/tlon-mobile/src/hooks/useDeepLinkListener.ts
```diff
@@ -0,0 +1,64 @@
...
+  }, [ship, lure, clearLure]);
+
+  // If deep link clicked, broadcast that navigation update to the webview and mark as handled
+  useEffect(() => {
+    if (wer && webviewContext.appLoaded) {
+      const gotoPath = getPathFromWer(wer);
+      console.debug('[useDeepLinkListener] Setting webview path:', gotoPath);
+      webviewContext.setGotoPath(gotoPath);
+      const tab = parseActiveTab(gotoPath) ?? 'Groups';
+      navigation.navigate(tab, { screen: 'Webview' });
```
dnbrwstr, line 60: Would it make sense to use `setGotoTab` for this?

dnbrwstr (changes requested):
Looks good to me overall. My main thing is: no inscrutable words in the codebase!! (Looking at you, `wer`). A few small things as well.

apps/tlon-mobile/src/App.tsx
```diff
@@ -54,12 +51,10 @@ const App = ({ wer: initialWer }: Props) => {
   useDevTools({ enabled: DEV_LOCAL, localCode: DEV_LOCAL_CODE });
```
dnbrwstr, line 51: I know this isn't part of your changes, but could you add a comment here indicating 1) What the `wer` is and 2) Where it's being passed in from? Since it's not clear from the context.

apps/tlon-mobile/src/hooks/useNotificationListener.ts
```diff
@@ -0,0 +1,70 @@
...
+  // If notification tapped, broadcast that navigation update to the
+  // webview and mark as handled
+  useEffect(() => {
+    if (gotoPath && webviewContext.appLoaded) {
+      console.debug(
+        '[useNotificationListener] Setting webview path:',
+        gotoPath
+      );
+      webviewContext.setGotoPath(gotoPath);
+      const tab = parseActiveTab(gotoPath) ?? 'Groups';
```
dnbrwstr, line 65: Since it seems like `setGotoTab` doesn't work here, could you add a comment as to why? Or maybe a comment on `setGotoTab` on when to use or not? Since I'd default to assuming this is the use case it handles.

dnbrwstr (approved):
Looking good! Two minor, non-blocking comments.

latter-bolden:
This looks good on the latest testflight build (61) and has already been code reviewed. Going to merge despite failing tests since those are unrelated.

PR #3323, opened 2024-03-12
chat: fix missing reply meta on posts in outline requests

patosullivan:
We were attempting to re-use the `uv-post` arm for `uv-post-without-replies`, but in order to do that we were setting replies to null before calling `uv-post`. This meant that the get-reply-meta arm, when being called from `uv-post`, would show no reply-meta for the post, even if there were replies.

Fixes LAND-1673

Tested locally with livenet moons.

PR Checklist


arthyn (approved):
lgtm

PR #3324, opened 2024-03-12
chat: fix re-render on chat input

patosullivan:
~The onUpArrow function being passed into the useMessageEditor hook was causing the chat input to re-render after send.~

~I'm disabling it for now so that it doesn't block getting the edit feature out. We should investigate why it's happening and re-implement later if possible.~

It turns out the re-render is only happening on mobile when `lastMessageId` changes. Not clear why this isn't effecting desktop, but we can effectively fix the issue (and disable onUpArrow) on mobile by setting lastMessageId to always be `''` on mobile.
 
Fixes LAND-1674

tested locally on livenet moons

PR Checklist


arthyn (approved):
lgtm, not sure if you want to leave the object logging in?

PR #3325, opened 2024-03-12
scroller: fix inverted scroll direction issue

patosullivan:
Two event listeners were accidentally removed in #3288, this PR adds them back in.

This also moves the use of the `useIsEditingMessage` hook into the `useInvertedScrollInteraction` hook, rather than using it in the scroller directly (since we only use it in `useInvertedScrollInteraction` hook anyway).

Fixes LAND-1675.

Tested locally on livenet moon.

PR Checklist


latter-bolden (approved):
Lgtm! worked locally

PR #3326, opened 2024-03-12
profile: work around clash with realm %profile

Fang-:
A different %profile agent has been spotted in the wild. It was not compatible with the kelvin at which _this_ agent was first released, made no %passes beyond eyre binding, and the software suit it was a part of (Realm) has since been discontinued.

So here we detect that case, and simply discard the previous state, starting anew with our %profile agent.

Closes LAND-1671.


arthyn (approved):
lgtm

PR #3328, opened 2024-03-12
mobile: handle webview crash & termination

latter-bolden:
Attempts to gracefully handle webview crashes and termination. This includes scenarios where the app is in the background and the OS kills the webview to conserve memory.

Tested locally by killing the webview from the command line and observing successful recovery.

Fixes LAND-1668

PR Checklist


dnbrwstr:
Found this, helpful? https://gist.github.com/jasonbekolay/dad7c446ae1b02f174dc3eb3a5ea70ee

latter-bolden:
Awesome find! Using those commands to target the webview pid, was able to test this.

Seems to be working as intended. Tested killing it both while the app is active and in the background.

dnbrwstr (approved):
Looks good to me! Down the road we may wanna pull this stuff into a `useWebviewErrorHandler` hook or something.

PR #3329, opened 2024-03-13
chat editing: fix issue with inability to add spaces

patosullivan:
Fixes LAND-1681.

This was introduced by yesterday's fix in #3325. If we use `useIsEditingMessage` within `useInvertedScrollInteraction`, the `isEditing` value never updates and those listeners are still active.

So the space keydown was getting captured in `useInvertedScrollInteraction` when it shouldn't have been.


arthyn (approved):
lgtm

PR #3330, opened 2024-03-13
e2e: fix commit hash diff issue

patosullivan:
Tests were failing because the %groups hash didn't change after we ran commit, attempting to ~fix by updating fake ships to vere 3 (the old fake ships would first update themselves to vere 3 before doing anything else, because we were downloading the latest vere before running the ships).~

Opening this PR is testing the hypothesis, do not merge yet.

Fixes LAND-1682


patosullivan:
Well, that didn't work. Will keep investigating.

PR #3331, opened 2024-03-13
edit: use v1 sub paths for posts and channels updates

patosullivan:
Fixes LAND-1683 ("Edited" not appearing next to posts until we scry again for posts)

Also fixes an annoying issue I found where if you pressed the up arrow key while composing a new message, you would jump to editing your last message and the text you typed into the chat input would disappear temporarily until you returned to the chat input. Instead, we just disable the up arrow key while you're composing a new message.


patosullivan:
I can at least repro the error locally now.

arthyn (approved):
lgtm

PR #3332, opened 2024-03-13
Mobile debug utils

dnbrwstr:
This PR adds some helpful utilities for debugging the native mobile app. Addresses LAND-1680.

### Expo Dev Client

Integrating this adds some nice utilities. Key feature is that it makes it possible to debug network requests via Chrome devtools. More info in the readme, and [here](https://docs.expo.dev/versions/latest/sdk/dev-client/).

### Credential Env Variables

This PR makes it possible to set default values for credentials screens on the both the Tlon login and ship login screens. If the credentials are not set, the fields will still correctly show placeholder values. This saves a lot of time/typing when iterating on login flows.


latter-bolden (approved):
This looks great! Tested locally on device, went smoothly

re @alecananian's note, agree this conflicts with the _DEV_LOCAL_ setup and that we should only have one path for expediting login. This one feels more elegant and would also solve for my workflow. Went ahead and ripped out the other one.

PR #3335, opened 2024-03-14
e2e: fix e2e by fixing state transitions in %notify

patosullivan:
Fixes LAND-1682 by fixing an issue in the state types in %notify.

At one point, the `notification` type changed and was subtly wrong. This went unnoticed because we used to drop old state anytime it didn't match the current state type.

This began to fail when we introduced changes to the %notify agent that included state transitions and there were ships that had not yet had that old state dropped.

This particularly effected the fake ships we use for e2e tests, because they boot with an old version of %groups. This would have also effected any ships that had not yet dropped their old state %3 and replaced it with the "new" state %3 with the bad notification type.

We get around this issue by just accepting any noun for the `notifications` type for base-state-2 and base-state-3, and creating a new base-state-4 with the correct type. Ships on either version of state %3 or already on state %4 should transition without issue now.

### DIFF
--- a/apps/tlon-web/src/state/channel/channel.ts
+++ b/apps/tlon-web/src/state/channel/channel.ts
@@ -1224,7 +1224,7 @@ export function useJoinMutation() {
       channelAction(chan, {
         join: group,
       }),
-      { app: 'channels', path: '/' },
+      { app: 'channels', path: '/v1' },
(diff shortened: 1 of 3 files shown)


Fang- (approved):
Mother of all branch names.

We paired on this, so of course _I_ would say "LGTM!"

PR #3337, opened 2024-03-14
edit: fix the space issue by using a ref for isEditing

patosullivan:
I'm not sure why the previous fix appeared to work. The event listeners weren't actually getting removed, so the issue persisted. Instead, we should keep isEditing in a ref and use that ref within the event listener, updating the ref when isEditing actually changes.

Fixes LAND-1681 (again).


latter-bolden (approved):
Lgtm! Code and reasoning make sense, worked locally

PR #3339, opened 2024-03-15
pwa: fix the scope/url issue in the manifest

patosullivan:
This fixes the issue where navigating to `[your ship]/apps/groups` would appear to be outside the scope of the PWA (you'd see the gray bar at the top, indicating you're at an external link).

Fixes LAND-1631


arthyn:
yessss

PR #3341, opened 2024-03-15
channels: need to invalidate queries when shoving in cache

arthyn:
I think this might have been the cause of LAND-1556, but still investigating

PR Checklist


latter-bolden (approved):

arthyn:
I think I was smoking something

PR #3345, opened 2024-03-18
mobile: pull in API client, bottom nav blue dots & avatar image

latter-bolden:
Pulls in an updated client which allows passing a custom polyfilled `fetch` with SSE support for native mobile use. Uses this api client to subscribe to unreads and display blue dots under groups & messages tab accordingly. Also adds a shared UI component for rendering avatars and uses it for the profile bottom nav tab.

Fixes LAND-1499
Fixes LAND-1659

PR Checklist


apps/tlon-mobile/index.js
```diff
@@ -1,10 +1,13 @@
 import { registerRootComponent } from 'expo';
 import 'expo-dev-client';
+import polyfill from 'react-native-polyfill-globals';
 import { TailwindProvider } from 'tailwind-rn';
 
 import App from './src/App';
 import utilities from './tailwind.json';
 
+polyfill();
```
dnbrwstr, line 9: Maybe add a comment here about how this is necessary in order to use react native's fetch with the Urbit client's server sent events implementation.

packages/shared/src/index.ts
```diff
@@ -9,3 +9,4 @@ export type {
   WebAppCommand,
 } from "./types/native";
 export { parseActiveTab, trimFullPath } from "./logic/navigation";
+export * from "./client";
```
dnbrwstr, line 12: I don't think we should export the client types at root, since they'd conflict if we were to try to export both client + urbit types at root, and they should be symmetrical. But not critical this second. Maybe just add a todo there?

dnbrwstr (approved):
Looks good overall! I don't think any of my comments are blocking.

packages/shared/src/index.ts
latter-bolden: Was thinking that as well. Let's just export them as a namespace for now.

apps/tlon-mobile/src/lib/sync.ts
```diff
@@ -1,8 +1,17 @@
...
 
 export const syncContacts = async () => {
   const contacts = await getContacts();
   db.createBatch('Contact', contacts, db.UpdateMode.All);
   console.log('Synced', contacts.length, 'contacts');
 };
+
+export const syncUnreads = async () => {
+  const channelUnreads = await getChannelUnreads();
+  const dmUnreads = await getDMUnreads();
```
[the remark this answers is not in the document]
latter-bolden: good catch

PR #3346, opened 2024-03-18
scroller: use subset of messages from chat window

patosullivan:
This fixes LAND-1553 by forcing the scroller to invert/uninvert more often by only giving it a small set of already cached messages to start off with, and progressively adding messages as necessary (as the user scrolls up or down).

This should be ready for review/banging-on.


latter-bolden:
Code makes sense. Banging on it locally, 1) I did hit an instance of a scroll loop in _Internet Cafe_ and 2) got it to lock up where I was at the top of the scroll and older messages didn't load in despite there being more in the channel.

Have to pivot off this for now, but will try to pinpoint specific steps to repro.

arthyn:
@patosullivan were you able to repro what @latter-bolden saw?

PR #3352, opened 2024-03-19
vite: working tamagui plugin

latter-bolden:
Was able to get this running. The urbit vite plugin seems to be misbehaving though, we may want to figure that out before merging.

also @dnbrwstr do you have some type/lint fixes on `hybrid-poc` from yesterday that didn't get pushed up? Getting errors there that looked similar to what you were looking at.


dnbrwstr:
Amazing, yeah I do, will add them in this branch.

dnbrwstr (approved):
LGTM :)

PR #3353, opened 2024-03-19
ops: revert staging merge

latter-bolden:
Undoes those avatar changes that were pulled in from staging yesterday.


arthyn (approved):
lgtm

PR #3354, opened 2024-03-19
activity: channels & groups integrations

Fang-:
Adds logic to the `%channels` and `%groups` agents that makes them push events into `%activity` in all places where we currently might notify. We add additional `$event:activity` cases to support all of these.

Draft PR because the `$event:activity` type is so-so. In particular, we are once again suffering from message identifier inconsistencies. We could also stand to finesse the event labels some more, maybe rearrange type contents for easier handling, update unread-related types for consistency with channels, etc.

We'll want to touch all of that up, but we may also want to await further clarification on desired ux, to help inform our decisions here.

Fixes LAND-1518


desk/app/channels.hoon
```diff
@@ -728,6 +759,7 @@
         ==
       ==
     =.  ca-core  ca-give-unread
+    ::TODO  %read activity-action?
```
arthyn, line 762: yep

desk/app/channels.hoon
```diff
@@ -1070,6 +1120,18 @@
     =/  reply  (get:on-v-replies:c replies.post id-reply)
     ?:  ?=([~ ~] reply)  ca-core
     ?:  ?=(%set -.u-reply)
+      =?  ca-core  ?&  ?=(^ reply)
+                       |(?=(~ reply.u-reply) ~!(reply (gth rev.u.reply.u-reply rev.u.u.reply)))
+                   ==
+        ::REVIEW  this might re-submit on edits. is that what we want?
```
arthyn, line 1126: we should probably discuss with crew about UX here

desk/app/groups.hoon
```diff
@@ -1313,6 +1348,8 @@
           ==
       ==
     =.  cor  (emit (pass-hark new-yarn))
+    ::TODO  want to (go-activity %flag), but we would need
```
arthyn, line 1351: yeah hmm, maybe we have `%flag-post` and `%flag-reply`

desk/app/groups.hoon
```diff
@@ -1466,7 +1503,13 @@
...
         =?  cor  go-is-our-bloc
           (emit (pass-hark new-yarn))
-        go-core
+        =+  ships=~(tap in ships)
+        |-
+        ?~  ships  go-core
+        =.  go-core
+          =<  ?>(?=(%shut -.cordon.group) .)  ::NOTE  tmi
+          (go-activity %ask i.ships)
+        $(ships t.ships)
```
arthyn, line 1512: hmm I wonder if we just want to support `(set ship)` for `ask` and similar activity

arthyn (approved):

PR #3355, opened 2024-03-19
groups: fix role deletion DDOS

arthyn:
This PR should fix LAND-1698 by doing all our role removals in one event rather than an event for each member of the group. It also prevents channels from attempting to resubscribe to channels that are no longer associated with the group. This is untested, but should fix the issues we've been seeing.

**todos:**
- follow-up PR with proper tests
- follow-up PR to rip out old agent logic (heap/diary) that isn't involved in migration

As far as we understand the issue with role deletion was as follows:
**host**
delete role in UI (1 event) -> verify group cabal(role) data -> check every member's roles for old roles -> poke self for each member that needs a role change (5k events) -> disseminate events -> each event goes to 5k subscribers ... eventually receive resubscribes for channels 

**subscribers**
receive role change event for each member (5k events) -> recheck permissions of each channel -> for each channel that we can read, try to resubscribe if subscription is missing (5-10 events)

So we can calculate these by multiplying for each step. Also some of these events were ran multiple times:
5 tries * 5000 role change events * 5000 ships to distribute each event to = 125 million facts needing to be sent from nibset

meanwhile each time a ship hears any of those events it sends back 5 more events trying to resubscribe to deleted channels so 125m * 5 = 625 million

so in total facts/messages:
- distributed: 125m + 625 m (watch-acks/nacks) = 750m
- heard: 625m

which means nibset has been chewing through 1.375 b events. This makes me think we should have diagrams for each event flow to make sure we never cascade like this.

PR Checklist

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -772,13 +772,6 @@
   ++  ca-safe-sub
     |=  delay=?
     ?:  ca-has-sub  ca-core
-    =/  =flag:g  group.perm.perm.channel
-    =/  =path
-      %+  weld
-        /(scot %p our.bowl)/groups/(scot %da now.bowl)
-      /groups/(scot %p p.flag)/[q.flag]/v1/group-ui
-    =+  .^(group=group-ui:g %gx path)
-    ?.  (~(has by channels.group) nest)  ca-core
     ?^  posts.channel  (ca-start-updates delay)
     =.  load.net.channel  |
     %.  delay
@@ -1871,6 +1864,17 @@
   ::
   ++  ca-recheck
     |=  sects=(set sect:g)
+    =/  =flag:g  group.perm.perm.channel
+    =/  groups-prefix  /(scot %p our.bowl)/groups/(scot %da now.bowl)
+    =/  exists-path  (weld groups-prefix /exists/(scot %p p.flag)/[q.flag])
+    =+  .^(exists=? %gx exists-path)
+    ?.  exists  ca-core
+    =/  =path
+      %+  weld
+        groups-prefix
+      /groups/(scot %p p.flag)/[q.flag]/v1/group-ui
+    =+  .^(group=group-ui:g %gx path)
+    ?.  (~(has by channels.group) nest)  ca-core
     ::  if our read permissions restored, re-subscribe
     ?:  (can-read:ca-perms our.bowl)  (ca-safe-sub |)
     ca-core
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -767,11 +767,18 @@
   ::
   ++  ca-has-sub
     ^-  ?
-    (~(has by wex.bowl) [ca-sub-wire ship.nest dap.bowl])
+    (~(has by wex.bowl) [ca-sub-wire ship.nest %channels-server])
   ::
   ++  ca-safe-sub
     |=  delay=?
     ?:  ca-has-sub  ca-core
+    =/  =flag:g  group.perm.perm.channel
+    =/  =path
+      %+  weld
+        /(scot %p our.bowl)/groups/(scot %da now.bowl)
+      /groups/(scot %p p.flag)/[q.flag]/v1/group-ui
+    =+  .^(group=group-ui:g %gx path)
+    ?.  (~(has by channels.group) nest)  ca-core
     ?^  posts.channel  (ca-start-updates delay)
     =.  load.net.channel  |
     %.  delay
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -1459,50 +1459,38 @@
     ::
         %del
       =.  cabals.group  (~(del by cabals.group) sect)
-      go-repair-permissions
-    ==
-  ::
-  ++  go-repair-permissions
-    =/  current-sects  ~(key by cabals.group)
-    =.  go-core
-      ::  repair members as needed
-      ::
-      %-  go-fleet-del-sects
-      %+  roll
-        ~(tap by fleet.group)
-      |=  [[s=ship =vessel:fleet:g] [ships=(set ship) sects=(set sect:g)]]
-      ::  traverse through each member to collect all old sects and the
-      ::  ships which had them
-      =/  dif  (~(dif in sects.vessel) current-sects)
-      :-  ?:(=(~ dif) ships (~(put in ships) s))
-      (~(uni in sects) dif)
-    =^  core=_go-core  cor
-      %+  roll
-        ~(tap by channels.group)
-      |=  [[=nest:g =channel:g] [gc=_go-core cr=_cor]]
-      =.  gc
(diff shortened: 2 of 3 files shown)


desk/app/channels.hoon
```diff
@@ -767,11 +767,18 @@
   ::
   ++  ca-has-sub
     ^-  ?
-    (~(has by wex.bowl) [ca-sub-wire ship.nest dap.bowl])
+    (~(has by wex.bowl) [ca-sub-wire ship.nest %channels-server])
```
Fang-, line 770: Should probably say `server` because that's what `+ca-safe-sub` uses.

```suggestion
    (~(has by wex.bowl) [ca-sub-wire ship.nest server])
```

("They should all be `%channels-server`" is a valid response, but also out of scope here.)

desk/app/channels.hoon
```diff
@@ -767,11 +767,18 @@
...
   ::
   ++  ca-safe-sub
     |=  delay=?
     ?:  ca-has-sub  ca-core
+    =/  =flag:g  group.perm.perm.channel
+    =/  =path
+      %+  weld
+        /(scot %p our.bowl)/groups/(scot %da now.bowl)
+      /groups/(scot %p p.flag)/[q.flag]/v1/group-ui
+    =+  .^(group=group-ui:g %gx path)
```
Fang-, line 780: A little bit surprising that the mark here isn't `%group-ui-v1`, even though the `v0` endpoint produces `%group-ui-v0`...

For added safety you may want to do an `/exists/[flag]` scry first. If the group doesn't even exist, `/groups` scries will just block on you.

desk/app/groups.hoon
```diff
@@ -253,16 +253,21 @@
...
-    =/  action  [flag now.bowl %fleet (~(gas in *(set ship)) ~[s]) %del-sects diff]
-    cre(cards [[%pass /groups/role %agent [our.bowl dap.bowl] %poke [act:mar:g !>(action)]] cards.cre])
+    =/  [affected=(set ship) old-sects=(set sect:g)]
+      %+  roll
+        ~(tap by fleet.group)
+      |=  [[s=ship =vessel:fleet:g] [ships=(set ship) sects=(set sect:g)]]
+      ::  traverse through each member to collect all old sects and the
+      ::  ships which had them
+      =/  dif  (~(dif in sects.vessel) current-sects)
+      :-  ?:(=(~(wyt in dif) 0) ships (~(put in ships) s))
```
Fang-, line 267:
```suggestion
      :-  ?:(=(~ dif) ships (~(put in ships) s))
```

desk/app/groups.hoon
```diff
@@ -253,16 +253,21 @@
...
+      (~(uni in sects) dif)
```
Fang-, line 268: Not sure I understand `%del-sects`, cannot tell if this is right or not.

Isn't the old logic sort of right, in that it says, "for this ship, these roles were removed"? Doesn't this new logic say "for these ships, these roles were removed", leaving it ambiguous as to which role was removed from which ship?

arthyn: I think this latest commit should address these concerns, lmk.

desk/app/groups.hoon
```diff
@@ -1638,13 +1630,17 @@
...
+    |=  [ships=(set ship) sects=(set sect:g)]
+    =.  fleet.group
         %-  ~(urn by fleet.group)
         |=  [=ship =vessel:fleet:g]
         ?.  (~(has in ships) ship)  vessel
-        vessel(sects (~(dif in sects.vessel) sects.diff))
+        vessel(sects (~(dif in sects.vessel) sects))
       go-core
-    ==
+  ::
```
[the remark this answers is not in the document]
Fang-: I also suspect there's some affected-ships : group-size ratio for which it becomes more performant to do a couple of `+jab:by`s rather than a full `+urn:by`, but in the common case (small to medium-sized groups) this is probably faster, and always fast enough for it to not matter either way.

desk/app/groups.hoon
```diff
@@ -1496,8 +1459,37 @@
...
+      =.  go-core
+        ::  remove from members as needed
+        ::
+        %-  go-fleet-del-sects
+        :_  old-sect
+        %+  roll
+          ~(tap by fleet.group)
+        |=  [[s=ship =vessel:fleet:g] ships=(set ship)]
+        ?.  (~(has in sects.vessel) sect)  ships
+        (~(put in ships) s)
```
Fang-, line 1472: If we're already iterating over the entire fleet here, do we not just want to do the `vessel(sects (~(dif in sects.vessel) sects))` from `+go-fleet-del-sects` in-line here? ie:

```suggestion
        %-  ~(run by fleet.group)
        |=  [s=ship =vessel:fleet:g]
        vessel(sects (~(dif in sects.vessel) sects))
```

desk/app/groups.hoon
```diff
@@ -1496,8 +1459,37 @@
...
+        ?.  has  [gc cr]
+        :: unsupported channel
+        ?.  ?=(?(%chat %heap %diary) p.nest)  [gc cr]
+        :: not host
+        ?:  !=(our.bowl p.q.nest)  [gc cr]
+        =/  cmd=c-channels:d  [%channel nest %del-writers old-sect]
+        =/  cage  [%channel-command !>(cmd)]
+        :-  gc
+        cr(cards [[%pass /groups/role %agent [p.q.nest %channels-server] %poke cage] cards.cr])
+      core
```
Fang-, line 1492: The double context editing here kinda scares me. I guess you need them both explicitly because you can't do edits on ""`cor.gc`""? In that case, I'd probably prefer just inlining the loop. Lets you avoid doing manual `emit`s also.

(And given that scries have some non-trivial overhead I'd do that as late as reasonable.)

```suggestion
      =/  channels
        ~(tap by channels.group)
      |-
      ?~  channels  go-core
      =*  nest     p.i.channels
      =*  channel  q.i.channels
      ::  delete sects from channel readers
      ::
      =.  go-core  (go-channel-del-sects:gc nest old-sect)
      ::TODO  delete sects from channel writers?
      ::  unsupported channel type
      ::
      ?.  ?=(?(%chat %heap %diary) p.nest)  gc
      ::  we are not the host, so not responsible for role management
      ::
      ?:  !=(our.bowl p.q.nest)  gc
      ::  if the channel still exists in the server,
      ::  tell it about the role change
      ::
      =+  .^(has=? %gu (channel-scry nest))
      ?.  has  go-core
      =/  cmd=c-channels:d  [%channel nest %del-writers old-sect]
      =/  cage  [%channel-command !>(cmd)]
      =.  cor
        %-  emit
        [%pass /groups/role %agent [p.q.nest %channels-server] %poke cage]
      $(channels t.channels)
```

Fang- (commented):
Better already! Some further comments on implementation details, if you'll forgive my nitpicking.

Fang- (approved):
This will do!

PR #3358, opened 2024-03-22
web: fix search bar click focusing

latter-bolden:
Noticed an issue with my search fix from yesterday where if the results ever gain focus, the search bar can never receive it again. This makes it tricky to do multiple searches.

Radix is pretty intense about it's focus guarding and the way we're using it here doesn't seem in line with what it's designed for. Looks like the click outside tracking was the main feature we were leveraging, so after pulling in a hook for that functionality I was able to remove Radix from the component entirely.

Now behaves as expected without the need for hacks and extra event handling. Keyboard result navigation and escape are maintained as well.

Tested against a local ship.

PR Checklist


arthyn (approved):
lgtm

PR #3359, opened 2024-03-25
hidden message: fix issue with react query key

patosullivan:
found this while working on another bug. we were using both ['diary', 'hidden'] and ['channels', 'hidden'] as query keys, this caused an issue where you sometimes were unable to unhide a message.

fixes LAND-1703


arthyn (approved):
lgtm

PR #3360, opened 2024-03-25
edit: fix issue with re-renders on editing reply message

patosullivan:
We were setting the scrollToId in the ChatWindow using the `edit` url param, which could include an ID for a reply message. The message would never be found and we'd end up in a re-render. We needed a different url param for editing a reply, `editReply`.

I also used wydr to try to root cause this originally, and found that we were seeing some unnecessary re-renders that could be fixed by moving the useLocation() hook call into the GroupsRoutes component and that we ought to memoize ChatChannel and ChatWindow.

fixes LAND-1701

### DIFF
--- a/apps/tlon-web/src/app.tsx
+++ b/apps/tlon-web/src/app.tsx
@@ -156,15 +156,16 @@ const SuspendedDiaryAddNote = (
 );
 
 interface RoutesProps {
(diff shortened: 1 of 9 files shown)


arthyn (approved):
lgtm, curious if this will help with sidebar animation

PR #3361, opened 2024-03-25
tests: support loading custom vase in +do-load

Fang-:
Was using this library for my own purposes recently and found that it offered no built-in affordance for modifying state or injecting state wholesale. Since `+on-load` gets called with a `vase`, seems appropriate for `+do-load` to accept that as an argument.

This way, state migrations may be tested, and agent state may be created ex-nihilo instead of needing to come about strictly through the normal lifecycle.

We take a `(unit vase)` so that it remains easy to fall back to `+on-save`'s output.

(Includes some tiny stylistic touch-ups to the library as well.)

### DIFF
--- a/desk/lib/test-agent.hoon
+++ b/desk/lib/test-agent.hoon
@@ -200,7 +200,7 @@
 ::
 ++  do-load
   =+  scry-warn=&
-  |=  =agent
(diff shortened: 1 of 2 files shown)


arthyn (approved):
very nice lgtm

PR #3366, opened 2024-03-26
channels: server persisted pending messages

arthyn:
This PR fixes LAND-1670 by persisting messages to the backend and alerting when pending messages are received. This should help issues where people double post because they "lost" their message because they navigated away. This also means your pending messages will appear cross device. 

To test, I created a group on a moon with all the channel types joined that from a separate moon and then shut the host down. Then from the second moon I sent messages in all the channel types. In each case you should see the message appear with an indicator that it is pending confirmation from host. Then once you've sent the messages, turn on the host and watch them change to their final state.

PR Checklist


arthyn:
putting some more cleanup on this, but I think it's mostly ready to review

arthyn:
@latter-bolden @Fang- @patosullivan I think this is finally ready for review.

Fang- (commented):
We looked at the hoon side of this together last week. Like we said, this will do.

apps/tlon-web/src/state/channel/channel.ts
```diff
@@ -1165,6 +1461,25 @@ export function useSortMode(nest: string): SortMode {
...
+const emptyPending: PendingMessages = {
+  posts: {},
+  replies: {},
+};
+export function usePendingPosts(nest: Nest) {
+  const channel = useChannel(nest);
+  return channel?.pending || emptyPending;
+}
+
+export function useIsSentPost(nest: Nest, cacheId: CacheId) {
```
latter-bolden, line 1473: may be good to keep these around, but doesn't seem like `useIsSentPost` or  `useIsSentReply` are used?

apps/tlon-web/src/state/channel/channel.ts
```diff
@@ -893,9 +1166,9 @@ export function useChannels(): Channels {
   const { data, ...rest } = useReactQueryScry<Channels>({
     queryKey: channelKey(),
     app: 'channels',
-    path: '/channels',
+    path: '/v2/channels',
     options: {
-      refetchOnMount: false,
+      // refetchOnMount: false,
```
latter-bolden, line 1171: should remove if we're gonna leave it out

apps/tlon-web/src/diary/DiaryChannel.tsx
```diff
@@ -79,13 +80,57 @@ function DiaryChannel({ title }: ViewProps) {
   const arrangedNotes = useArrangedPosts(nest);
   const lastArrangedNote = arrangedNotes[arrangedNotes.length - 1];
 
+  useEffect(() => {
+    let timeout: any;
+
+    console.log(nest, awaitingNote, location);
```
latter-bolden, line 86: probably want to remove this

latter-bolden (commented):
Few nits, but frontend code looks good.

`state/channel` logic was a bit complicated to read through, so going to try this out locally before approving.

apps/tlon-web/src/diary/DiaryChannel.tsx
```diff
@@ -79,13 +80,53 @@ function DiaryChannel({ title }: ViewProps) {
...
+            if ('post' in response) {
+              const { 'r-post': postResponse } = response.post;
+              if (
+                'set' in postResponse &&
+                postResponse.set !== null &&
+                postResponse.set.essay.author === window.our &&
+                postResponse.set.essay.sent === parseInt(awaitingNote, 10)
+              ) {
+                const { essay } = postResponse.set;
+                return essay.sent === parseInt(awaitingNote, 10);
```
latter-bolden, line 104:
```suggestion
                return true;
```

latter-bolden (approved):
Worked well locally. Two smaller things I noticed:

Refreshing from undelivered to delivered leads to duplicates. A second refresh resolves the cache correctly.
Screenshot 2024-04-02 at 2 25 15 PM

Gallery message overflows on my screen size at full width.
Screenshot 2024-04-02 at 2 20 45 PM


Out of scope for this PR, but this feels like a concept where a more concise visual representation + expanded info would be useful (similar to sent/read double checkmark + tooltip on other platforms).

latter-bolden:
Fix resolves the dupes, nice work.

PR #3367, opened 2024-03-27
fix react-native-reanimated crash on web

alecananian:
- Upgrades tamagui packages to latest `1.93.2` and removes the patch we had
- Adds patch to `react-native-reanimated` to polyfill the `global` var, which was causing the web to crash because `@tamagui/animations-moti` imports this module on web, where `global` is not available. (workaround suggestion found here: https://github.com/tamagui/tamagui/issues/1747)

Tested and confirmed to resolve the crash by running `npm run build` then `npm run serve` on the web app


patosullivan (approved):
Nice, thank you!

PR #3369, opened 2024-03-27
Mobile SQLite integration

dnbrwstr:
This PR adds sqlite + dev tooling. Still needs to be documented + fully integrated with the GroupsList, I'll continue on with that next.


packages/shared/src/api/subscribe.ts
```diff
@@ -0,0 +1,12 @@
+import type { ClientTypes as Client } from "../client";
+import { subscribeChannelUnreads, subscribeDMUnreads } from "./unreadsApi";
+
+export const subscribeUnreads = async () => {
+  async function handleUnreadUpdate(unread: Client.Unread) {
+    // db.create("Unread", unread, db.UpdateMode.All);
```
patosullivan, line 6: Same question here, why is this duplicated? And why comment this out?

packages/shared/src/db/client.ts
```diff
@@ -0,0 +1,28 @@
+import type { QueryResult } from "@op-engineering/op-sqlite";
+import { BaseSQLiteDatabase } from "drizzle-orm/sqlite-core";
+import { SqliteRemoteResult } from "drizzle-orm/sqlite-proxy";
+import { Schema } from "./types";
+
+export type AnySqliteDatabase = BaseSQLiteDatabase<
```
patosullivan, line 6: Do we know if this works with our anticipated web sqlite implementation?

packages/shared/drizzle.config.ts
```diff
@@ -0,0 +1,7 @@
+import type { Config } from "drizzle-kit";
```
patosullivan, line 1: How will we handle the drizzle config in the shared package with web?

patosullivan (approved):
This looks great! A few small things, but nothing that needs to block getting this in.

packages/shared/src/db/client.ts
dnbrwstr: Yes it will, this type matches the base implementation of all the SQLite drivers. I tried it as just a union of the three, but there were type errors some where in Drizzle's internals. Will add a comment.

apps/tlon-mobile/drizzle-studio.config.ts
```diff
@@ -0,0 +1,9 @@
+import type { Config } from 'drizzle-kit';
+
+export default {
+  schema: '../../packages/shared/src/db/schema.ts',
```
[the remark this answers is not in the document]
dnbrwstr: Unfortunately not, it only accepts a path.

packages/shared/src/db/client.ts
patosullivan: Nice!

packages/shared/src/api/subscribe.ts
dnbrwstr: Yeah this is wonky. Removed the duplicate and added a comment for now, will clean up with the next round of work.

packages/shared/drizzle.config.ts
patosullivan: Ah ok, cool.

PR #3370, opened 2024-03-27
edit: make useMyLastMessage more performant

patosullivan:
Rewrote useMyLastMessage to be used only for chat channels, added useMyLastReply to be used only for replies in channels. Made `myLastMessage` a prop for chatInput, we now call the hook from whatever the parent of ChatInput is.

Noticed/remembered this while looking into perf via the profiler elsewhere.


arthyn (approved):
lgtm

PR #3371, opened 2024-03-27
perf: speed up useRecentSort in useSidebarSort

patosullivan:
Another small perf improvement for the big brain sidebar sorting. Pre-compute and memoize the unreads before sorting. Profiler showed this function was fairly heavy, this reduces the load quite a bit.


arthyn (approved):
lgtm, although we should probably nuke the sorts soon

PR #3372, opened 2024-03-27
cleanup: consistent formatting for shared packages

dnbrwstr:
This PR creates a root `.prettierrc.js` for common formatting settings. I also added new `.prettierrc.js` files for the `shared` and `ui` packages, which were missing them before.

One small future improvement would be to also hoist the import order plugin config -- I didn't do that here because I started seeing import/module format issues when I did.

### DIFF
--- a/.prettierrc.cjs
+++ b/.prettierrc.cjs
@@ -1,0 +1,36 @@
+module.exports = {
+  bracketSpacing: true,
+  printWidth: 80,
+  proseWrap: 'always',
+  singleQuote: true,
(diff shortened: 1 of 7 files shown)


dnbrwstr:
Needs some work, drafting for now.

arthyn (approved):
lgtm

dnbrwstr:
Drafting again, I think we should probably wait and merge this and reformat the shared packages at the same time, but want to coordinate on when with @patosullivan and @alecananian.

PR #3373, opened 2024-03-27
edit: fix issue with editing replies in notebooks and galleries

patosullivan:
Fixes LAND-1731 by making sure we have a nest (we don't in the case of chat, we do in the case of diary/heap, previously we were just appending `chat/` to whatever we had in `whom`).


arthyn (approved):
lgtm, we should target staging though

patosullivan:
Closing in favor of https://github.com/tloncorp/landscape-apps/pull/3374

PR #3376, opened 2024-03-28
WIP: add workers to tlon-web to speed up state updates

patosullivan:
Fixes LAND-1556 (potentially) by speeding up our state updates by moving the computation for new state produced by new facts into workers.

Most things seem to work, only issue I'm aware of at the moment is we're not updating reply count in the main window when new replies are received.


arthyn:
@jamesacklin this is separate, we can probably punt for now because from what I remember it wasn't clear whether this was actually better right @patosullivan ?

PR #3377, opened 2024-03-28
Mobile: add react-cosmos

patosullivan:
Adds react-cosmos so we can do more parallel work on developing our views.

Fixes LAND-1734


dnbrwstr (approved):
Lgtm! I'd double check the formatting now that prettier config is updated.

PR #3382, opened 2024-03-29
native: initial components for channel switcher

alecananian:
Adds initial components and scaffolding for channel switcher screen, to be revisited when designs are completed and the parent screen is available

Simulator Screenshot - iPhone 15 Pro - 2024-03-29 at 15 42 03


dnbrwstr (approved):
Nice! Lgtm.

PR #3383, opened 2024-03-31
Imports + migrations rework

dnbrwstr:
This PR:

- Modifies the `ClientTypes` export -- suggested way to do it now is  `import * as client from '@tloncorp/shared/dist/client'`, which better matches the way we import from the other subpackages.
- Modifies the schema update process so that we're just replacing a single initial migration for now. At some point we'll start actually using migrations, but it'll save us time + complexity if we just drop all data for now -- should be ephemeral anyway.
- Adds a script to watch `schemas.ts` and attempt to regenerate the migration when it changes.
- Modifies the native db initialization so that it will drop + recreate the database if the initial migration fails.

Together, these changes mean that you should be able to make a change in `schemas.ts`, refresh the app, and have your schema changes applied automatically.

Fixes LAND-1737


patosullivan (approved):
Looks great! Adding a tsup config was also sorely needed, ha.

PR #3385, opened 2024-04-01
channel view + scroller

patosullivan:


packages/ui/src/components/ChatMessage/AuthorRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+  author,
+  sent,
+  roles,
+}: {
+  author: client.Contact;
+  sent: string;
+  roles?: string[];
+}) {
+  const date = new Date(sent);
+  const timeDisplay = utils.makePrettyDayAndDateAndTime(date);
```
dnbrwstr, line 32: Should memoize this -- I see date related calcs pop up a lot when debugging performance.

packages/ui/src/components/ChatMessage/AuthorRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const firstRole = roles?.[0];
+
+  return (
+    <XStack gap="$space.l">
```
dnbrwstr, line 36: Space is redundant here, `gap` should automatically pull from space values.

```suggestion
    <XStack gap="$l">
```

packages/ui/src/components/ChatMessage/ChatContent.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+import { Button } from '../Button';
+import ShipName from '../ShipName';
+
+function ShipMention({ ship }: { ship: string }) {
+  return (
+    <Button
+      // TODO: implement this once we have a profile screen or sheet
+      // onPress={() => naivigate('Profile', { ship })}
+      backgroundColor="$blueSoft"
```
dnbrwstr, line 32: This is prob placeholder, so feel free to ignore, but with this many style props for something like this, I'd start thinking about adding a variant.

packages/ui/src/components/ChatMessage/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import AuthorRow from './AuthorRow';
+import ChatContent from './ChatContent';
+
+const ChatMessage = memo(({ post }: { post: client.Post | null }) => {
+  if (!post) {
+    return null;
+  }
+
+  const content = JSON.parse(post.content) as Story;
```
dnbrwstr, line 14: Should memoize this + roles.

packages/shared/src/logic/utils.ts
```diff
@@ -0,0 +1,106 @@
+import { differenceInDays, endOfToday, format } from 'date-fns';
+import emojiRegex from 'emoji-regex';
+import _ from 'lodash';
+
+type App = 'chat' | 'heap' | 'diary';
+
+export function checkNest(nest: string) {
```
dnbrwstr, line 7: Can these be removed or moved to urbit package? Trying to keep all the urbit utils colocated.

packages/ui/src/components/Channel/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  type?: 'chat' | 'gallery' | 'notebook';
+}) {
+  const generateTitleFromMembers = (members: client.GroupMember[]) => {
+    const contacts = useContacts();
+    return members
+      .map((m) => contacts[m.id].nickname || contacts[m.id].id)
+      .join(', ');
+  };
+
+  const title = channel.title
```
dnbrwstr, line 33: I'd memoize this title stuff + maybe extract as a hook or utils function -- i think we'll need it elsewhere too. Might even make sense to cache on db update.

packages/ui/src/components/ChatMessage/AuthorRow.tsx
```diff
@@ -0,0 +1,45 @@
...
+import * as client from '@tloncorp/shared/dist/client';
+import { SizableText, View, XStack } from 'tamagui';
+
+import { Avatar } from '../Avatar';
+import ShipName from '../ShipName';
+
+const RoleBadge = ({ role }: { role: string }) => {
+  return (
+    <View
+      borderRadius="$l"
```
dnbrwstr, line 11: Probably not there yet, but I'm curious if there are wrapper types it makes sense to extract.

packages/ui/src/components/IconButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+  children,
+  onPress,
+  size = '$s',
+}: {
+  children: React.ReactNode;
+  onPress: () => void;
+  size?: SizeTokens;
+}) {
+  return (
+    <Button size={size} onPress={onPress} borderWidth="unset">
```
dnbrwstr, line 15: Why `borderWidth="unset"`?

packages/ui/src/components/ShipName.tsx
```diff
@@ -0,0 +1,64 @@
+import { cite } from '@urbit/aura';
+import { useMemo } from 'react';
+import { FontSizeTokens, FontWeightTokens, SizableText } from 'tamagui';
+
+import { useCalm } from '../contexts/calm';
+import { useContact } from '../contexts/contacts';
+
+export default function ShipName({
```
dnbrwstr, line 8: Would it make sense to call this `AuthorName`? or `ContactName`? Just trying to keep the vocab consistent.

packages/ui/src/components/UrbitSigil.native.tsx
```diff
@@ -1,13 +1,62 @@
 // Note: we must import sigil from @urbit/sigil-js/dist/core
```
dnbrwstr, line 1: Can we extract everything these two files have in common to a `UrbitSigil.base.tsx`? There's a lot of overlap.

dnbrwstr (commented):
Looking good! Some initial, incomplete comments.

packages/shared/src/logic/utils.ts
patosullivan: are you referring to checkNest and nestToFlag in particular?

packages/ui/src/components/ChatMessage/AuthorRow.tsx
patosullivan: Not sure what you mean on this one.

packages/ui/src/components/IconButton.tsx
patosullivan: Because Button has a border and if we leave it on there the icon buttons look pretty strange: 
image

packages/ui/src/components/ShipName.tsx
patosullivan: ContactName probably works.

packages/ui/src/components/ChatMessage/AuthorRow.tsx
dnbrwstr: Haha yeah, not clear -- just wondering out loud if we'll want any `PaddedBox` type components.

packages/ui/src/components/IconButton.tsx
dnbrwstr: Makes sense, could you add a comment?

patosullivan: Sure thing.

dnbrwstr (approved):
Like we talked about, a few things to clean up, but this is a great start. Lgtm!

PR #3389, opened 2024-04-04
native: separate branch.io test/prod apps based on preview/production targets

alecananian:
Updates the Branch.io config so that our test app corresponds to the preview target and the live app corresponds to the production target (i.e., join.tlon.io links will launch the main app and sa96e.test-app.link links will launch the preview app)

Tested by installing the preview app alongside the main app and tapping on a link like https://sa96e.test-app.link/dm-sapdex-diflyx to launch it. Note that deep linking is not currently set up and will need to be revisited when the home tab stack is more complete

I've had trouble running the preview Android app locally since the introduction of `expo-dev-client`, so I tested that out by firing off an Expo build and installing the generated APK

### DIFF
--- a/apps/tlon-mobile/eas.json
+++ b/apps/tlon-mobile/eas.json
@@ -10,8 +10,7 @@
         "APP_VARIANT": "preview"
       },
       "android": {
-        "gradleCommand": ":app:assemblePreviewRelease",
-        "buildType": "apk"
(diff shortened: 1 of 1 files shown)


dnbrwstr:
Nice! Looks good to me.

dnbrwstr (approved):

PR #3390, opened 2024-04-04
sidebar sort: fix issue with recency

patosullivan:
My recent change to useRecentSort() didn't account for the fact that some channels/dms don't have any unreads.

This led to a bizarre looking recent sort order for groups.

This fixes that by defaulting to Number.NEGATIVE_INFINITY if we can't find an unread (rather than only if we can't find recency).

Tested locally on a live ship.

Fixes LAND-1747

PR Checklist


latter-bolden (approved):
lgtm

PR #3391, opened 2024-04-04
Fix vite/e2e issues caused by urbit sigil component

patosullivan:
Had to rejigger the way we use UrbitSigilBase and the native/mobile components for it.

Fixes LAND-1748


patosullivan:
Looking into the ambiguous import issue showing up in vite on CI now.

patosullivan:
> Looks good for now -- I'll try to do another pass on these soon to simplify. I think for web we just import the `urbit-sigil` package and use the custom element? Like in `tlon-web/src/components/Avatar.tsx`

Yep, that's right.

PR #3395, opened 2024-04-05
native: fix missing nav bar in onboarding stack

alecananian:
[This change](https://github.com/tloncorp/tlon-apps/commit/f743da3188a429909e976971f1753807e4909275#diff-ef1a11e46f5301a808634240cd96b8fdf05898ebcf1f4efdfd255da71a56ce8c) for the `HomeStack` ended up removing the top nav bar for the `OnboardingStack` as well. I've updated the `useScreenOptioins` hook to accept overrides to the default values to support the `HomeStack`.


dnbrwstr (approved):
Good catch, works for me.

PR #3396, opened 2024-04-05
Add declaration maps to shared package

dnbrwstr:
This should enable us to use "go to definition" to go directly to the source file, rather than .d.ts file.


dnbrwstr:
Seems to be working, I think it's good to go.

PR #3397, opened 2024-04-05
mobile: channel search

latter-bolden:
https://github.com/tloncorp/tlon-apps/assets/90741358/739662e1-8220-4434-945a-2bfd40810ac4

Should cover everything from the Figma spec except in message search query highlighting and stitching local/remote results together (remote only right now). The former should just be done and I'll file an issue. For the latter, we'll probably want to hold off until we have a better idea of how the local cache will work. Getting a good UX there with consistent ordering might be tricky.

Still getting sporadic errors/warnings I'm trying to track down and want to include a few tests. Added basic fixtures for a generic input, but didn't have time to take a real stab at building out a reusable Ochre lego.

Fixes LAND-1736


apps/tlon-mobile/src/screens/ChannelSearchScreen.tsx
```diff
@@ -0,0 +1,82 @@
...
+      navigation.getParent()?.setOptions({
+        tabBarStyle: undefined,
+      });
+    };
+  }, [navigation]);
+
+  const navigateToPost = useCallback(
+    (post: db.PostWithRelations) => {
+      navigation.navigate('Channel', {
+        group: group.result!, // something is wrong here?
```
dnbrwstr, line 44: I think the issue is just that `result` can be null, right? This is fine for now, since my next chunk of work will make `group` optional entirely on the channel screen, but maybe add a less cryptic comment :P

packages/shared/src/api/channelsApi.ts
```diff
@@ -10,6 +17,73 @@ export const getUnreadChannels = async () => {
...
 
+const searchChatChannel = async (params: {
+  channelId: string;
+  query: string;
+  cursor?: string;
+}) => {
+  const SINGLE_PAGE_SEARCH_DEPTH = 500;
+  const encodedQuery = stringToTa(params.query);
+
+  console.log(`search scry for ${params.query}, cursor: ${params.cursor}`);
```
dnbrwstr, line 28: Remove or use dev logger

packages/ui/package.json
```diff
@@ -22,6 +22,7 @@
     "@tamagui/image": "1.93.2",
     "@uidotdev/usehooks": "^2.4.1",
     "color2k": "^2.0.0",
+    "lodash": "^4.17.21",
```
dnbrwstr, line 25: I think we've removed this?

packages/ui/src/components/ChannelSearch/SearchResults.tsx
```diff
@@ -0,0 +1,104 @@
...
+            <YStack alignItems="center" marginTop="$3xl">
+              <SizableText color="$secondaryText">No results found</SizableText>
+              <SearchStatus search={search} color="$tertiaryText" />
+            </YStack>
+          )}
+
+          {posts.length > 0 && (
+            <>
+              <XStack justifyContent="space-between" marginBottom="$xl">
+                <View>
```
dnbrwstr, line 52: Do we need the wrapper views here?

packages/ui/src/components/ChannelSearch/SearchStatus.tsx
```diff
@@ -0,0 +1,85 @@
...
+}: {
+  search: SearchState;
+  color?: ColorTokens;
+}) {
+  const { searchComplete, errored, numResults, searchedThroughDate } = search;
+
+  return (
+    <XStack justifyContent="center" alignItems="center">
+      {errored ? (
+        <SizableText size="$s" color="$negativeActionText">
```
dnbrwstr, line 20: We should consider breaking some of this into other components -- eg `SmallErrorMessage` or w/e. Might not be quite there yet, but soon.

dnbrwstr (changes requested):
Looking good overall! A few nitpicks.

apps/tlon-mobile/src/screens/ChannelSearchScreen.tsx
```diff
@@ -0,0 +1,82 @@
+import type { NativeStackScreenProps } from '@react-navigation/native-stack';
+import { useChannelSearch } from '@tloncorp/shared/dist';
+import * as db from '@tloncorp/shared/dist/db';
+import { XStack, YStack } from '@tloncorp/ui';
+// TODO: update references to dist
+import { Button, SearchBar, SearchResults } from '@tloncorp/ui/src/index';
```
dnbrwstr, line 6: One more thing: should update this

dnbrwstr (approved):
Nice! looks good, pending the contact bits we discussed.

PR #3403, opened 2024-04-08
tiptap in react-native

patosullivan:
Ended up using tentap via its "Advanced setup" (https://10play.github.io/10tap-editor/docs/setup/advancedSetup).

We won't be able to reuse much of the tiptap config, but we will be able to reuse our parsing code for translating from prosemirror to our AST and back.

Fixes LAND-1652


patosullivan:
Looking into the postinstall script issue.

patosullivan:
I also need to update this now that the channel component can render DMs/group DMs.

dnbrwstr:
Great progress here, two highish-level items:
- Could you move notebook-related stuff to another branch? So it's clear what's baked and what's not.
- Can we use our `@tloncorp/ui` components instead of bringing in tailwind etc.?

patosullivan:
> Great progress here, two highish-level items:
> 
> * Could you move notebook-related stuff to another branch? So it's clear what's baked and what's not.
> * Can we use our `@tloncorp/ui` components instead of bringing in tailwind etc.?

Removed the notebook related stuff and put it on a separate branch. Right now I brought in tailwind just to make sure the inline editor styles match what we had previously. We could set the styles manually, though.

patosullivan:
Working on making it so the input can send a message on any type of chat channel now.

patosullivan:
> > Removed the notebook related stuff and put it on a separate branch. Right now I brought in tailwind just to make sure the inline editor styles match what we had previously. We could set the styles manually, though.
> 
> Ok, nice. I do think it'd be worth seeing what pulling in our ui components / styles would look like as part of this work, since that'd be the cleanest way to share styles, and it might impact how we integrate the editor package.

Right now we're actually not using any other components in the editor package, just the MessageInputEditor. I pulled in tailwind, but it wasn't doing much, went ahead and pulled it out.

packages/ui/src/components/Channel/ChatScroll.tsx
```diff
@@ -29,16 +29,21 @@ export default function ChatScroll({
...
   posts: db.PostWithRelations[];
   unreadCount?: number;
   firstUnread?: string;
+  setInputShouldBlur: (shouldBlur: boolean) => void;
   selectedPost?: string;
 }) {
   const [hasPressedGoToBottom, setHasPressedGoToBottom] = useState(false);
   const flatListRef = useRef<FlatList<db.PostWithRelations>>(null);
   const lastPost = posts[posts.length - 1];
+  const sortedPosts = posts.sort((a, b) => {
```
latter-bolden, line 44: We probably want to memoize this one.

patosullivan:
Found an issue with the webview crashing if you highlight text, trying to figure that one out now.

apps/tlon-mobile/src/screens/ChannelScreen.tsx
```diff
@@ -30,6 +34,23 @@ export default function ChannelScreen(props: ChannelScreenProps) {
...
 
   const { top, bottom } = useSafeAreaInsets();
+  const { ship } = useShip();
+
+  const messageSender = async (content: JSONContent, channelId: string) => {
+    if (!ship) {
+      return;
+    }
+
+    const isDm = whomIsDm(channelId);
```
latter-bolden, line 44: can we pull this off `channel.type`? More broadly, would be nice to constrain backend terminology like `whom` to `shared/api`

patosullivan: Yeah, good idea.

packages/shared/src/api/chatApi.ts
```diff
@@ -1,8 +1,76 @@
...
+      id,
+      diff: {
+        uid: '0v3',
+        delta,
+      },
+    },
+  };
+}
+
+function getActionAndEvent(
```
latter-bolden, line 54: maybe we collapse these into a single action builder since it seems like we never use the event/diff?

patosullivan:
Ok, the issue with the webview crashing after the user selects some text had to do with some lower level issue that was caused (*I think*) by differing versions of prosemirror-view used by the tiptap placeholder extention and tentap. If I switch over to the tentap built-in placeholder bridge, the crash goes away, *but*, it doesn't seem to actually apply the placeholder message set in the extension config, so I had to manually override the CSS for that extension to show our intended placeholder message.

Now the user should be able to select text and apply formatting without the editor crashing.

latter-bolden (approved):
Lgtm! Let's get this in

PR #3404, opened 2024-04-09
groups: fix issue where admins can't delete if they're not the host

patosullivan:
Apparently the `is-admin` util has been broken. It was scrying into the wrong place to check if a user is an admin. It only appeared to work because of an early return where we check if the ship is also the host of the channel.

If an admin was *not* also the host of the channel, then they were not able to delete messages in that channel.

### DIFF
--- a/desk/lib/channel-utils.hoon
+++ b/desk/lib/channel-utils.hoon
@@ -306,7 +306,6 @@
     ;:  weld
         /gx
         groups-scry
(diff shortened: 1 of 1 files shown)


Fang- (approved):
As discussed out of band, this indeed matches the group-wide permission scry endpoint we do implement:

https://github.com/tloncorp/tlon-apps/blob/7230eb68f195fe04cafe28ac657b29c1bef66fda/desk/app/groups.hoon#L1005

Good catch, thank you!

PR #3405, opened 2024-04-09
Native DMs + Clubs

dnbrwstr:
This overstuffed PR adds handling for DMs + Clubs.

Done
- Add `dm` and `groupDm` channel types and syncs those channels from the server
- Combine Groups, DMs, and clubs into a single list
- Modify Channel screen to allow it to display a channel that's not part of a group
- Tidy up list item styles + bring closer to design
- Add parsing + storage for thread unreads (not used yet)
- Add new query for home view that sorts + aggregates groups + channels to render list
- Modify `useQuery` to rerun every time args change, even if a query is in progress. If there's a running query, its result will be discarded (but it won't be cancelled).

Notes
- Group unread counts show number of unread messages, not unread channels
- Navigating to a group loads the mostly recently updated channel; it should load the last channel you visited
- Query updates run way too frequently right now. Probably do want to move to react query with smarter keying.
- Simplified UI contexts, want to refactor soon
- Did some rough tweaks to Avatar to get resizing to work, needs refactoring
- Channel switcher appearance is delayed when navigating to a group channel
- List items getting a little wonky with the introduction of `ChatListItem`, need a refactor. Not clear when to use `GroupListItem` vs `ChannelListItem` vs `ChatListItem`
- Not sure about the way I'm modelling / passing around relations. Types are getting verbose/confusing (Looking at you, `GroupWithLastPostAndMembers`).


jamesacklin:
Screenshot 2024-04-10 at 2 52 46 PM

A few things in the channel/DM view:
- Nicknames are black in dark mode
- Messages appear to now be in reverse-order? Éd sent a message at midnight:31, Zach sent something at 11:14a, I sent a few messages at 1:48p.

IMG_BA843E5BB3DF-1

In the list:
- I have a blank DM? channel? group? at the top of my pinned list called "No title," which isn't present on the web. This might be a me-thing?
- The "testing12345" group has an intentionally busted image. The expo dev console raised absolute hell so I deleted the group from the web client. Deleting the group did not remove it from the list.

Not sure how much of this is in the scope of this PR. Since you touched data fetching, how many of these issues are implicated?

dnbrwstr:
@jamesacklin this is great, thank you for documenting. These are definitely things that need to be fixed, but I'd put them at the same level as my "notes" -- there's a lot here that needs polish. Maybe throw 'em in Linear for now?

patosullivan (approved):
Looks great! Let's get this in.

PR #3407, opened 2024-04-10
mobile: message actions

latter-bolden:
This should be ready for review. Adds long-press message options with animation and some behavior (emoji reacts, copy, hide, delete). Also adds an initial custom EmojiPickerSheet we can build out from.

Some message action behavior is left for once other pieces are in place (threading, replying, editing). Also punted on tracking message hidden state on initial import — these APIs are a bit of a mess since adding that behavior on the Urbit side was rushed.

Still needs some visual love and better modal ergonomics, but those can be addressed down the line.

Fixes LAND-1653


packages/ui/src/components/Channel/ChatScroll.tsx
```diff
@@ -128,6 +162,20 @@ export default function ChatScroll({
...
         onScrollToIndexFailed={handleScrollToIndexFailed}
       />
+      <Modal
+        visible={activeMessage !== null}
+        onDismiss={() => setActiveMessage(null)}
+      >
+        {activeMessage !== null && (
+          <ChatMessageActions
+            currentUserId={currentUserId}
+            post={activeMessage!}
```
dnbrwstr, line 172: Do we need the `!` here? I would think the null check would be enough, and if it's not there might be a case where it's an incorrect value that we should be handling.

latter-bolden:
Working on tracking down CI type issues, but that needn't block review.

apps/tlon-mobile/cosmos.imports.ts
```diff
@@ -16,7 +16,7 @@ import * as fixture1 from './src/fixtures/TlonButton.fixture';
 import * as decorator0 from './src/fixtures/cosmos.decorator';
 
 export const rendererConfig: RendererConfig = {
-  playgroundUrl: 'http://localhost:5000',
+  playgroundUrl: 'http://localhost:5001',
```
dnbrwstr, line 19: Intentional?

packages/ui/src/components/ChatMessage/ChatMessageActions.tsx
```diff
@@ -0,0 +1,279 @@
...
+  shortCode: string;
+  details: ReactionDetails;
+  handlePress: (shortCode: string) => void;
+}) {
+  return (
+    <Button
+      padding="$xs"
+      borderWidth={0}
+      backgroundColor={
+        details.self.didReact && details.self.value.includes(shortCode)
```
dnbrwstr, line 248: Should use `$positiveBackground` and `$positiveBorder` -- might need a better name for that.

packages/ui/src/components/ChatMessage/ReactionsDisplay.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <XStack padding="$m" paddingLeft="$4xl" borderRadius="$m">
+      {reactionDetails.list.map((reaction) => (
+        <XStack
+          justifyContent="center"
+          alignItems="center"
+          backgroundColor={
+            reaction.value === reactionDetails.self.value
+              ? '$blueSoft'
```
dnbrwstr, line 32: Should use `$positiveBackground` etc.

dnbrwstr (changes requested):
This is great -- some really nice stuff here. Dropping a bunch of nitpicks, I was looking as you pushed the last few commits, so ignore anything outdated.

packages/ui/src/components/ChatMessage/ChatMessageActions.tsx
latter-bolden: I think the name is clear, this was just sloppy.

apps/tlon-mobile/cosmos.imports.ts
latter-bolden: Yeah, 5000 is used by the system on Mac if you have AirPlay Receiver enabled ([see here](https://github.com/react-cosmos/react-cosmos/issues/1355))

Fun fact: that feature works well for mirroring your phone screen to your laptop for demo's/Gather screen share

apps/tlon-web/tsconfig.json
```diff
@@ -19,8 +19,7 @@
       "vite/client",
       "vitest/globals",
       "@testing-library/jest-dom",
-      "vite-plugin-pwa/react",
-      "../../packages/ui/src/tamagui.d.ts"
```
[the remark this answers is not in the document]
latter-bolden: Yeah, intentional. Don't think the type issues are related, but may need to adjust.

packages/ui/src/components/Sheet.tsx
```diff
@@ -38,6 +38,7 @@ const Frame = styled(YStack, {
   borderTopLeftRadius: '$3xl',
   borderTopRightRadius: '$3xl',
   backgroundColor: '$background',
+  zIndex: 9999999999999,
```
dnbrwstr, line 41: One more

dnbrwstr (approved):
Looks good to me. Last tiny thing, then merge away.

PR #3410, opened 2024-04-11
GroupActions: don't prevent non-hosting admins from leaving group

jamesacklin:
Removes a erroneous check for non-admin status in the "Leave group" GroupAction. Instead, we just look at the flag (is the current ship the host?).

We probably also need to bolster this to prevent channel hosts from leaving the group and orphaning their hosted channels.

PR Checklist
- [x] Tested locally with the Vite dev server proxied against a livenet planet


patosullivan (approved):
lgtm

PR #3412, opened 2024-04-11
Remove unused webview setup

dnbrwstr:
This PR improves boot time a bit by removing unused webview-related plumbing. It also includes a fix for a minor issue that could cause channel types to be set incorrectly.


patosullivan (approved):
RIP

PR #3413, opened 2024-04-11
Image improvements

dnbrwstr:
- Adds an `AsyncStorage` cache for svg images so that we don't have to refetch every time we render them.
- Reworks `Avatar` and `-Sigil` components to support standard sizing and fix some rendering issues


patosullivan (approved):
Looks good

PR #3417, opened 2024-04-12
mobile: misc fixes for init related data.

patosullivan:
1) we should use the new /v1/init endpoint.
2) DMs have a different message id type (patp with a ud that has decimals in it), our unreadsApi didn't account for that. 
3) users may not always have pins, we were failing to insert because we tried to pass an empty array into insertPinnedItems 
4) since we don't always have pins, hide that section in the ui.


dnbrwstr (approved):
Beautiful, ty.

PR #3418, opened 2024-04-12
web: add missing find route back (fixes desktop lure links)

patosullivan:
Looks like this (user sees the home screen to the app, with the group preview modal):

image

(also note that I cut off half of the screen in the screenshot, the modal shows in the middle like it normally does)


arthyn (approved):

jamesacklin:
Thank you!

PR #3424, opened 2024-04-16
native: add wrapper to all cosmos fixtures

jamesacklin:
Wraps each Cosmos fixture in a FixtureWrapper component.


latter-bolden (approved):
Looking good, will merge once green

PR #3426, opened 2024-04-17
desktop: fix missing groups sidebar metadata

latter-bolden:
There was a recent performance optimization targeting sidebar transition lag that updated the sidebar to avoid re-rendering. The caching there seems too aggressive causing missing group metadata when refreshed or navigated to directly (no title or icon). This small fix makes sure we use the correct data there.

Performance wise, I didn't notice an impact. Though the issue was subtle iirc.

Also smuggling in a small change to avoid `tsc` failing to find the tamagui type declarations.

FIxes LAND-1765


patosullivan (approved):
Makes sense!

PR #3427, opened 2024-04-17
native: render chat message content appropriately

patosullivan:
fixes LAND-1766

Makes sure that we can recursively render formatted text, that the formatted text appears on the same line, and that we add a new row for line breaks.

Also adds a video player for embedded videos. No other embed work yet.

### DIFF
--- a/packages/ui/src/components/ChatMessage/ChatContent.tsx
+++ b/packages/ui/src/components/ChatMessage/ChatContent.tsx
@@ -18,8 +18,10 @@ import {
   isShip,
   isStrikethrough,
 } from '@tloncorp/shared/dist/urbit/content';
+import { ReactElement } from 'react';
+import { Linking } from 'react-native';
 
-import { Image, SizableText, Text, View, YStack } from '../../core';
+import { Image, Text, View, XStack, YStack } from '../../core';
 import { Button } from '../Button';
 import ContactName from '../ContactName';
 
@@ -62,7 +64,7 @@ export function InlineContent({ story }: { story: Inline | null }) {
       <>
         {story.bold.map((s, k) => (
           <Text fontSize="$m" fontWeight="bold" key={k}>
-            {s}
+            <InlineContent story={s} />
           </Text>
         ))}
       </>
@@ -74,7 +76,7 @@ export function InlineContent({ story }: { story: Inline | null }) {
       <>
         {story.italics.map((s, k) => (
           <Text fontSize="$m" fontStyle="italic" key={k}>
-            {s}
+            <InlineContent story={s} />
           </Text>
         ))}
       </>
@@ -86,7 +88,7 @@ export function InlineContent({ story }: { story: Inline | null }) {
       <>
         {story.strike.map((s, k) => (
           <Text fontSize="$m" textDecorationLine="line-through" key={k}>
-            {s}
+            <InlineContent story={s} />
           </Text>
         ))}
       </>
@@ -95,14 +97,14 @@ export function InlineContent({ story }: { story: Inline | null }) {
 
   if (isInlineCode(story)) {
     return (
-      <SizableText
+      <Text
         fontFamily="$mono"
         backgroundColor="$gray100"
         padding="$xs"
         borderRadius="$xl"
       >
-        <InlineContent story={story['inline-code']} />
-      </SizableText>
+        {story['inline-code']}
+      </Text>
     );
   }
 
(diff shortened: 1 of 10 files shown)


patosullivan:
Working on the vite issue now.

packages/ui/src/components/ChatMessage/ChatContent.native.tsx
```diff
@@ -0,0 +1,351 @@
...
+} from '@tloncorp/shared/dist/urbit/content';
+import { AVPlaybackStatus, ResizeMode, Video } from 'expo-av';
+import { ReactElement, useEffect, useRef, useState } from 'react';
+import { Linking } from 'react-native';
+
+import { Image, Text, View, XStack, YStack } from '../../core';
+import { Button } from '../Button';
+import ContactName from '../ContactName';
+
+function ShipMention({ ship }: { ship: string }) {
```
dnbrwstr, line 29: Most of the components here don't need to be native-specific -- can we split it so that the minimum possible is in the native file? For example if `Video` component from `expo-av` is native-only, I'd have a `Video.native.ts`, but still have `BlockContent` with wrapper styles, etc. be shared.

packages/ui/src/components/ChatMessage/ChatContent.native.tsx
```diff
@@ -0,0 +1,351 @@
...
+}
+
+export function BlockContent({ story }: { story: Block }) {
+  // TODO add support for other embeds and refs
+
+  if (isImage(story)) {
+    const isVideoFile = utils.VIDEO_REGEX.test(story.image.src);
+
+    if (isVideoFile) {
+      console.log('is video file');
```
dnbrwstr, line 174: Some stray logs -- would use a `devLogger` here to keep logging under control.

packages/ui/src/components/ChatMessage/ChatContent.native.tsx
```diff
@@ -0,0 +1,351 @@
...
+      })}
+    </>
+  );
+}
+
+export default function ChatContent({ story }: { story: Story }) {
+  const storyInlines = (
+    story.filter((s) => 'inline' in s) as VerseInline[]
+  ).flatMap((i) => i.inline);
+  const storyBlocks = story.filter((s) => 'block' in s) as VerseBlock[];
```
dnbrwstr, line 319: Would be good to memoize this stuff

packages/ui/src/components/ChatMessage/ChatContent.native.tsx
```diff
@@ -0,0 +1,351 @@
...
+import { Image, Text, View, XStack, YStack } from '../../core';
+import { Button } from '../Button';
+import ContactName from '../ContactName';
+
+function ShipMention({ ship }: { ship: string }) {
+  return (
+    <Button
+      // TODO: implement this once we have a profile screen or sheet
+      // onPress={() => naivigate('Profile', { ship })}
+      backgroundColor="$blueSoft"
```
dnbrwstr, line 34: Should use `$positiveBackground` here -- in general we should stick to theme colors (under `dark` or `light` in `tamagui.config.ts`, so that dark mode etc. work as expected. If we need a key not in there, we can expand.

packages/ui/src/components/ChatMessage/ChatContent.tsx
```diff
@@ -187,9 +187,83 @@ export function BlockContent({ story }: { story: Block }) {
...
-    </SizableText>
+    </Text>
+  );
+}
+
+function LineRenderer({ storyInlines }: { storyInlines: Inline[] }) {
+  const inlineElements: ReactElement[][] = [];
+  let currentLine: ReactElement[] = [];
+
+  storyInlines.forEach((inline, index) => {
```
dnbrwstr, line 200: Can anything here be memoized? I'd `React.memo` this component either way.

packages/ui/src/components/ChatMessage/ChatContent.tsx
```diff
@@ -187,9 +187,83 @@ export function BlockContent({ story }: { story: Block }) {
...
+            {inline}
+          </Text>
+        );
+      }
+    } else if (isBlockquote(inline)) {
+      currentLine.push(
+        <YStack
+          key={`blockquote-${index}`}
+          borderLeftWidth={2}
+          borderLeftColor="$gray100"
```
dnbrwstr, line 227: Should be `$borderColor`

packages/ui/src/components/MessageInput/index.native.tsx
```diff
@@ -179,6 +186,11 @@ export function MessageInput({
                   return;
                 }
 
+                if (e.key === 'Enter' && e.shiftKey) {
+                  window.ReactNativeWebView.postMessage('shift-enter');
```
dnbrwstr, line 190: Doesn't hurt, but probably impossible, right? Is shift-enter on mobile a thing?

dnbrwstr (changes requested):
Looks great, a few nitpicks, as usual.

packages/ui/src/components/MessageInput/index.native.tsx
patosullivan: Actually, yeah, we should probably just make enter add the new line rather than have it send, and just do away with shift-enter.

packages/ui/src/components/ChatMessage/ChatContent.tsx
```diff
@@ -18,22 +18,27 @@ import {
...
       // onPress={() => naivigate('Profile', { ship })}
-      backgroundColor="$blueSoft"
+      backgroundColor="$positiveBackground"
       paddingHorizontal="$xs"
       paddingVertical={0}
-      borderRadius="$xl"
+      borderRadius="$m"
     >
-      <ContactName name={ship} showAlias />
+      <Text color="$blue" fontSize="$m">
```
dnbrwstr, line 39: Sorry, one more -- should be `$positiveText`.

PR #3428, opened 2024-04-17
ops: manual build order

latter-bolden:
- move post-install logic at app & package level to build scripts (avoiding post-install lifecycle hook)
- add scripts to top level `package.json` for building sub-packages in acceptable order
- call those build scripts in top level post-install hook

Want to make sure this fixes the issues we're observing before we try to build things concurrently. Anecdotally, it spends most of the time waiting for the initial package install anyhow.

Also adds preinstall hook to remove that troublesome `tsconfig.tsbuildinfo` file in `packages/shared`


patosullivan (approved):
Nice, green check mark

PR #3429, opened 2024-04-17
native: fix scrolling in chat channels

patosullivan:
fixes land-1769 by removing the onPress from View, passing the same handler to onScrollBeginDrag, which achieves the same effect afaict.


dnbrwstr (approved):
Nice, lgtm.

latter-bolden (approved):
Makes sense, worked locally

PR #3430, opened 2024-04-17
[Draft] Thread loading

dnbrwstr:


packages/shared/src/db/queries.ts
```diff
@@ -626,6 +654,7 @@ export const insertChannelPosts = createWriteQuery(
     }
     return client.transaction(async (tx) => {
       const lastPost = posts[posts.length - 1];
+      console.log(lastPost.textContent);
```
patosullivan, line 657:
```suggestion
```

patosullivan (approved):
Two tiny things, otherwise looks good and runs well in the simulator.

PR #3432, opened 2024-04-18
mobile: fix chat message lag

latter-bolden:
Updates the MotiView to only be rendered once the message is active vs always on every chat message.


dnbrwstr (approved):
Nice, lgtm!

PR #3438, opened 2024-04-18
mobile: swipeable channel list items

latter-bolden:
https://github.com/tloncorp/tlon-apps/assets/90741358/9a8701a5-899f-4987-8807-611c45db7d1a

Makes chats in the main list swipeable. Animated the slide in for multiple actions similar to how most native ones feel. Also adds functionality for pinning/unpinning chats and wires that action up.

Wasn't able to get the "swipe further to trigger the last action in the list" interaction feeling right and didn't want to fidget with it any longer (@dnbrwstr would be fun to take a look at that with you). It would be easier to implement it for the case where we only have one action, but I didn't think it made sense to try adding that as a temporary measure.

Hides unimplemented actions behind a flag, so only swipe right to pin/unpin will be visible for now.

Fixes LAND-1809


dnbrwstr (approved):
Very cool. Understand what you mean about the colors now lol. Lgtm!

PR #3439, opened 2024-04-19
mobile: image viewer

latter-bolden:
Adds a lightbox style screen for viewing images in chat on press. Supports zoom, pan, double press to zoom/reset, etc. in a way that should be intuitive from other native experiences.

The navigational changes feel somewhat dramatic, but supposedly breaking a screen out above the tab stack is the proper way to handle nested fullscreen views. Working within the existing stack while messing with bottom tab style (opacity, display, etc.) had drawbacks. Had to pull out the cool "shared element between screens" animation — I think we can work around existing bugs there, but it will take a bit more time to have that feeling stable.

Would like to implement save to camera roll if I have time, but I think this is mergable as is.


packages/ui/src/components/ImageViewerScreenView.tsx
```diff
@@ -0,0 +1,96 @@
...
+            onPinchEnd={handlePinchEnd}
+            onDoubleTap={onDoubleTap}
+            onSingleTap={onSingleTap}
+            minPanPointers={minPanPointers}
+          />
+        </View>
+
+        {/* overlay */}
+        <YStack>
+          <XStack opacity={showOverlay ? 1 : 0}>
```
dnbrwstr, line 71: Not necessary for this pr, but flagging for later: we should break this out as a generic `Overlay` component.

dnbrwstr (approved):
Nice! This lgtm.

PR #3440, opened 2024-04-19
mobile: message actions fixes

latter-bolden:
Branches off the image viewer PR, we shouldn't merge this until that one is in.

- fixes longpress detection on chat images (image viewer press handling was swallowing)
- fires haptic when longpress is detected, not after modal opens
- reworks position calculations so focused message appears to transition from original location more closely
- adjusts message action animations more broadly (@dnbrwstr take a look, may swing too far in the bouncy direction)

still needs action list brightened a bit


latter-bolden:
Going to close this, will be easier to salvage the important bits in other PRs given conflicts.

PR #3444, opened 2024-04-19
native: add ability to upload images

patosullivan:


patosullivan:
Working on getting the animations right.

patosullivan:
Having trouble getting tamagui animations to actually show up, marking this as ready for now. Might still get animations in before merge.

dnbrwstr:
Code is looking good overall. I don't love pulling the full immer+zustand+lodash stack in, but I don't think it makes sense to rewrite it all yet, so let's just keep an eye on it.

I'm going to pull down + build to play with the UI a bit more, but some high level questions:
- Could more of this be moved to the shared package? Seems like most parts of this will be a shared concern. 
- I do think it would be worth looking at image resizing here -- it's important, and there probably won't be a time it makes more sense to get into. Could be in a followup PR though if you prefer.
- I'm curious about video -- how much extra work is it? Not a requirement here, but worth looking into.

apps/tlon-mobile/src/screens/ChannelScreen.tsx
```diff
@@ -64,27 +75,98 @@ export default function ChannelScreen(props: ChannelScreenProps) {
...
+    [ship, channel, resetImageAttachment]
+  );
 
   useEffect(() => {
     if (error) {
       console.error(error);
     }
   }, [error]);
 
+  const fetchImageFromUri = useCallback(
```
dnbrwstr, line 110: Can we pull out most or all of the image upload logic here? Some of it (like this method) could be part of `storage`, some could be hooks. We'll end up needing image upload in a number of places.

packages/shared/src/api/chatApi.ts
```diff
@@ -78,11 +78,17 @@ export const markChatRead = (whom: string) =>
 export const sendDirectMessage = async (
```
dnbrwstr, line 78: Should combine this with `postsApi.sendPost`, following the same basic pattern as `postsApi.getChannelPosts`

packages/shared/src/api/postsApi.ts
```diff
@@ -47,10 +47,14 @@ export function channelPostAction(nest: ub.Nest, action: ub.PostAction) {
 export const sendPost = async (
   channelId: string,
   content: JSONContent,
-  author: string
+  author: string,
+  blocks?: ub.Block[]
```
dnbrwstr, line 51: Should the signature just be `(channelId: string, authorId: string, content: ub.Story)`? Or why pass them in separately?

packages/shared/src/urbit/upload.ts
```diff
@@ -0,0 +1,161 @@
+import type { S3Client } from '@aws-sdk/client-s3';
```
dnbrwstr, line 1: Wdyt about putting this file in `api`, rather than `urbit`?

packages/ui/src/components/IconButton.tsx
```diff
@@ -5,6 +5,7 @@ export function IconButton({
   children,
   onPress,
   size = '$s',
+  ...props
```
dnbrwstr, line 8: Intentional?

packages/ui/src/components/MessageInput/AttachmentButton.native.tsx
```diff
@@ -0,0 +1,124 @@
...
+      requestMediaLibraryPermission();
+    }
+
+    if (showPopover && cameraPermissionStatus?.granted === false) {
+      requestCameraPermission();
+    }
+  }, [mediaLibraryPermissionStatus, showPopover]);
+
+  return (
+    <Popover
```
dnbrwstr, line 62: Let's use our standard action sheet here for now, rather than popover.

packages/ui/src/components/MessageInput/index.native.tsx
```diff
@@ -95,7 +100,22 @@ export function MessageInput({
 
   const sendMessage = useCallback(() => {
     editor.getJSON().then(async (json) => {
-      await send(json, channelId);
+      const blocks: ub.Block[] = [];
+
+      console.log({ uploadedImage });
```
dnbrwstr, line 105: Stray log

packages/ui/src/core/tamagui.ts
```diff
@@ -16,6 +16,8 @@ export {
   TextArea,
   Text,
   Dialog,
+  Popover,
```
dnbrwstr, line 19: Can we scratch these for now? Just trying to limit the higher-level tamagui we stuff we add because not all of it is super baked.

dnbrwstr (changes requested):

packages/shared/src/api/postsApi.ts
patosullivan: Yeah, we should just create the story in the message input and send the whole thing up, just committed that now.

packages/shared/src/urbit/upload.ts
patosullivan: Yeah, good idea, just committed that.

packages/ui/src/components/IconButton.tsx
patosullivan: Nope, removed.

patosullivan:
Addressed @dnbrwstr's main feedback, now looking into simplifying the flow, removing the crop step, and optimistically showing the image in the image preview/chat so we don't wait for it to upload before showing the preview.

patosullivan:
Found an issue with uploading on android while working on the new flow, now looking into that.

patosullivan:
Ok, fixed the android issue, and now we use the local image for preview and skip the crop step.

packages/ui/src/components/Channel/index.tsx
```diff
@@ -1,5 +1,6 @@
-import { JSONContent } from '@tiptap/core';
 import * as db from '@tloncorp/shared/dist/db';
+import { Story } from '@tloncorp/shared/dist/urbit';
+import { Upload } from 'packages/shared/dist/api';
```
dnbrwstr, line 3: Wrong import path (strange)

packages/ui/src/components/MessageInput/AttachmentButton.native.tsx
```diff
@@ -0,0 +1,127 @@
...
+export default function AttachmentButton({
+  setImage,
+  uploadedImage,
+  paddingBottom,
+}: {
+  setImage: (uri: string | null) => void;
+  uploadedImage?: Upload | null;
+  paddingBottom: number;
+}) {
+  const [showPopover, setShowPopover] = useState(false);
```
dnbrwstr, line 20: Since it's not a popover, can we call this `showInputSelector` or something?

apps/tlon-mobile/src/hooks/storage/reducer.ts
```diff
@@ -0,0 +1,155 @@
+/* eslint-disable no-param-reassign */
```
dnbrwstr, line 1: Can the `storage` module be a part of `store`?

apps/tlon-mobile/src/hooks/storage/upload.ts
```diff
@@ -0,0 +1,465 @@
...
+import * as api from '@tloncorp/shared/dist/api';
+import type {
+  FileStore,
+  Status,
+  StorageConfiguration,
+  StorageCredentials,
+  Uploader,
+} from '@tloncorp/shared/dist/api';
+import { deSig, formatDa, unixToDa } from '@urbit/aura';
+// import imageCompression from 'browser-image-compression';
```
dnbrwstr, line 13: Comment

dnbrwstr (commented):
Looking really good -- some small things here, plus looking at thumbnails like we discussed, then I think it's good to go.

packages/shared/src/api/urbit.ts
```diff
@@ -70,6 +70,8 @@ export function subscribe<T>(
     throw new Error('Tied to subscribe, but Urbit client is not initialized');
   }
 
+  logger.debug('subscribing to', printEndpoint(endpoint));
```
dnbrwstr, line 73: 🙏

packages/ui/src/assets/icons/ArrowUp.svg
```diff
@@ -1,3 +1,3 @@
 <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M12 5L6 11M12 5L18 11M12 5V19" stroke="#1A1818" style="stroke:#1A1818;stroke:color(display-p3 0.1020 0.0941 0.0941);stroke-opacity:1;" stroke-width="1.6" stroke-linecap="round" stroke-linejoin="round"/>
```
dnbrwstr, line 2: Let's revert this for now -- `.svgrrc` is supposed to key on that color value to automatically replace color; if that's not working we should fix there.

packages/ui/src/components/MessageInput/AttachmentButton.native.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <Sheet.Overlay animation="quick" />
+        <Sheet.Frame>
+          <Sheet.Handle paddingTop="$xl" />
+          <YStack
+            paddingBottom={paddingBottom}
+            gap="$xl"
+            paddingHorizontal="$xl"
+            paddingTop="$xl"
+          >
+            <Button
```
dnbrwstr, line 98: Can we use our standard `ActionMenu` component for this?

packages/ui/src/core/tamagui.ts
```diff
@@ -16,6 +16,13 @@ export {
   TextArea,
   Text,
   Dialog,
+  Separator,
```
dnbrwstr, line 19: Is this used? Would prefer to remove it.

apps/tlon-mobile/index.js
```diff
@@ -3,6 +3,7 @@ import { addCustomEnabledLoggers } from '@tloncorp/shared';
 import { setStorage } from '@tloncorp/ui';
 import { registerRootComponent } from 'expo';
 import 'expo-dev-client';
+import 'react-native-get-random-values';
```
dnbrwstr, line 6: This ended up being extraneous, right?

packages/ui/src/svg.d.ts
```diff
@@ -1,6 +1,5 @@
 declare module '*.svg' {
-  const content: React.FC<
-    React.SVGAttributes<SVGElement> & React.RefAttributes
-  >;
+  import { SvgProps } from 'react-native-svg';
```
dnbrwstr, line 2: Is this necessary? I had tried to set it up so that the shared package used the more generic type and `svg.d.ts` in tlon mobile overrode it with the native types.

dnbrwstr (approved):
Looks good! Down to nitpicky stuff, then good to go.

packages/ui/src/assets/icons/ArrowUp.svg
patosullivan: Yeah, doesn't appear to be working. I'll revert.

patosullivan: Actually, it is working. Idk why I thought it wasn't earlier. But yeah, I'll revert.

packages/ui/src/components/MessageInput/AttachmentButton.native.tsx
patosullivan: Discussed out of band, ActionMenu doesn't exist but it should (outside of the scope of this PR, will post another PR to create that component and use it here and in GroupOptionsSheet).

packages/ui/src/core/tamagui.ts
patosullivan: Removed

apps/tlon-mobile/index.js
patosullivan: This is necessary for our s3 methods to work.

packages/ui/src/svg.d.ts
patosullivan: Looks like it's not

PR #3445, opened 2024-04-19
Convert to PNPM

dnbrwstr:
Uses PNPM instead of NPM.


package.json
```diff
@@ -2,38 +2,48 @@
...
     "prepare": "husky",
-    "postinstall": "npx patch-package && npm run build:all"
+    "postinstall": "pnpm run build:all"
   },
   "devDependencies": {
     "@trivago/prettier-plugin-sort-imports": "^4.3.0",
     "concurrently": "^8.2.2",
     "husky": "^9.0.10",
-    "patch-package": "^6.4.7",
+    "lint-staged": "^15.0.0",
```
arthyn, line 26: do these do the same thing? I think we still use patch package

arthyn (approved):
lgtm, just the one question

package.json
dnbrwstr: They don't do the same thing -- this is just moving the lint-staged command to top level. PNPM has patching built in, so the patches in `<root>/patches` get applied automatically on install. More info here: https://pnpm.io/cli/patch

PR #3446, opened 2024-04-22
Refactor ESLint

dnbrwstr:
This PR refactors our ESLint config into a single config at root level. We can add custom config to individual packages as we need, but it didn't seem necessary for this pass. 

I also chose a set of rules that seemed like pragmatic defaults, but I'm very much open to suggestions, so see what you all think.


arthyn (approved):
lgtm

latter-bolden (approved):

PR #3448, opened 2024-04-23
android: fix issue with channel view 'wiping away' on keyboard dismiss

patosullivan:
It turns out this was caused by the KeyboardAvoidingView's `keyboardVerticalOffset` prop, which a) positioned the input box too high on iOS b) causes this weird issue on android when used in combination with the `height` behavior. Removing the offset and setting the behavior to `position` when on android fixes the issue, and iOS still looks/works the same.

Note that we also need to set a `contentContainerStyle` when using the `position` behavior on Android (this does not effect the way the component renders when we're using the `padding` behavior on iOS).


dnbrwstr (approved):
Nice, lgtm.

PR #3450, opened 2024-04-23
Post-demo fixes

dnbrwstr:
A couple fixes I made immediately after the demo that got lost in the shuffle.

- Fixes duplicate issue in chat list
- Fixes missing contact names in post previews


dnbrwstr:
Scratch that, I'm going to figure out something better.

PR #3453, opened 2024-04-23
mobile: ensure we always pass along correct user id

latter-bolden:
OTT

Tried this in my already authenticated state and from scratch with a fresh install. No cast fails in sight


patosullivan (approved):
Looks good, makes sense

PR #3454, opened 2024-04-24
mobile ops: automated build numbers & PR previews

latter-bolden:
1) Single click build action

Uses remote, auto-incrementing build numbers on Expo. This means our action becomes single click to generate new TestFlight/Play Store builds without needing to update a bunch of native files. Tested by triggering a Preview app build from the command line, but once this is merged I'll run both profiles end to end to confirm it's working as expected. Note, we'll still need to update user facing versions manually when releasing  (i.e. 4.0.0 -> 4.0.1) .

2) Adds support for attaching demo builds to PRs for checking out features during review. 

All you need to do is add one of the new labels and when the check turns green, you can click through to Expo and scan the QR code to install. Uses internal distribution, so you need to have your device registered in the set for it to work (ping me and I can add you). Right now, it has to overwrite your existing Preview install.

Fixes LAND-1807


patosullivan (approved):
Super nice.

dnbrwstr (approved):
🙏

apps/tlon-mobile/eas.json
```diff
@@ -23,13 +25,28 @@
...
         "scheme": "Landscape"
       }
     },
+    "demo": {
+      "extends": "base",
+      "channel": "demo",
+      "distribution": "internal",
+      "autoIncrement": true,
+      "android": {
+        "gradleCommand": ":app:bundlePreviewRelease",
```
latter-bolden, line 42: @alecananian when I try to build the demo profile locally for Android, it warns me that this command will output the wrong artifact for internal distribution (`.aab` not `.apk`)

do you know what the gradle command should be to build for the preview bundle id but output the correct filetype here?

latter-bolden: Yeah I tried that, but Expo still warns. Haven't registered my Android device yet, so still need to try running it end to end and see if it ends up being fine.

Supporting iOS for these is more important anyhow, so I'll give that a shot after merging.

PR #3455, opened 2024-04-24
native: fix sqlite transaction failures

dnbrwstr:
This PR should fix the transaction failures we've been seeing. It seemed like there were two issues: 

- Drizzle was treating `op-sqlite` transactions as synchronous while calling asynchronous methods, 
- Drizzle doesn't seem to have any mechanism for ensuring that asynchronous transactions don't interrupt each other (at least in any of the sqlite drivers)

I fixed this by patching the drizzle op-sqlite driver to use op-sqlite's native transaction method, and to run it asynchronously. 

Note: this will break nested transactions, which we're not using, and which I don't think will be an issue.


latter-bolden (approved):
Lgtm

PR #3456, opened 2024-04-25
native: add action sheet component

patosullivan:
Simulator Screenshot - iPhone 15 - 2024-04-25 at 10 53 58

Simulator Screenshot - iPhone 15 - 2024-04-25 at 11 07 43

Simulator Screenshot - iPhone 15 - 2024-04-25 at 11 08 07

I added a generic ActionSheet component and implemented it in the two places (that I'm aware of) that we need it. Also added a fixture for it.

Fixes LAND-1811

### DIFF
--- a/apps/tlon-mobile/cosmos.imports.ts
+++ b/apps/tlon-mobile/cosmos.imports.ts
@@ -3,6 +3,7 @@
 import { RendererConfig, UserModuleWrappers } from 'react-cosmos-core';
 
 import * as fixture0 from './src/App.fixture';
+import * as fixture15 from './src/fixtures/ActionSheet.fixture';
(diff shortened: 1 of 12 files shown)


latter-bolden (approved):
This looks good! Was cool playing with the attachments sheet.

I don't think either of these need to block, but two thoughts I had reading through it:

- Press state visual change & haptic would make selecting action items feel more tactile. Simple opacity tweak might be easiest.

- Adopting radix style composition (i.e. `<ActionSheet>`, `<ActionSheet.Title>`, `<ActionSheet.Item>`) might work nicely here. I like how it lets you pass arbitrary attributes down to the child nodes. Might be overkill for now, but Tamagui's helpers make that pattern more straightforward to implement.

latter-bolden (approved):

PR #3459, opened 2024-04-26
native: color fixes

jamesacklin:
- Uses $tertiaryText on ActionSheet item borders and on channel icons in ChannelListItem.
- Adds the useMedia hook to the editor package and utilizes it to show white text on a dark background in dark mode.

Screenshot 2024-04-26 at 12 15 43 PM

Screenshot 2024-04-26 at 12 16 58 PM

Screenshot 2024-04-26 at 12 17 09 PM

Fixes LAND-1802
Fixes LAND-1803
Fixes LAND-1084

### DIFF
--- a/packages/editor/src/useMedia.ts
+++ b/packages/editor/src/useMedia.ts
@@ -1,0 +1,70 @@
+import { useCallback, useEffect } from 'react';
+import create from 'zustand';
+
+interface MediaMatchStore {
+  media: {
+    [query: string]: boolean;
+  };
+  setQuery: (query: string, data: boolean) => void;
+}
+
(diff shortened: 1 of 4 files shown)


packages/editor/src/useMedia.ts
```diff
@@ -0,0 +1,70 @@
+import { useCallback, useEffect } from 'react';
```
patosullivan, line 1: Might want to put this in the shared package, we can pull it into the new web app later too.

Not a blocker necessarily.

patosullivan (approved):
One nit, not a big deal though.

PR #3460, opened 2024-04-26
native: fix urbit color normalization

alecananian:
- Updates contacts sync to use shared `normalizeUrbitColor` utility instead of local `parseUrbitColor` utility which padded the wrong side of the color value
- Pulls in `normalizeUrbitColor` tests from the web repo
- Removes unneeded lodash dependencies

Before
Screenshot 2024-04-26 at 1 17 17 PM

After
Screenshot 2024-04-26 at 1 16 21 PM


latter-bolden (approved):
Nice, lgtm!

PR #3462, opened 2024-04-26
mobile: use platform/theme neutral overlay for image viewer

latter-bolden:
Removes the existing overlay in the ImageViewer in favor of a simpler close icon. Right now it may be a bit too subtle, but I think it'll be more clear when the overlay is open once it's more featureful (save to camera roll action for example)

Also fixes longpress on message images to open the actions. Still has that brief yet annoying resize stutter when we swap in the animated message (I'm coming for you!)

Fixes LAND-1797


dnbrwstr (approved):
Nice

PR #3465, opened 2024-04-29
native: fix weird height issue on input on android

patosullivan:
Apparently 44px of padding was hardcoded into tentap's RichText component when rendered on android (apparently to account for the OS's bottom nav bar?). I set it to zero, seems fine with both gesture based nav and older button based nav, so I patched tentap with that value set to zero.

Fixes LAND-1795


dnbrwstr (approved):
👍

PR #3471, opened 2024-05-01
native: add notebook channel root view

patosullivan:
Adds the basic notebook channel rootview (without the expanded version of the posts for now, would require building out the content rendering component(s) for notebooks unless we just reuse the chat content rendering, which would look pretty weird imo).

Also uses the long press for message actions.

At the moment, the NotebookPost component has a ton of optional boolean props. It probably needs to be split out into separate components, but I'm not sure what those components will be yet. Headline vs summary vs full?

Looks like this:

https://github.com/tloncorp/tlon-apps/assets/1221094/41780c84-19cc-4356-a3bd-685500d4e0eb


dnbrwstr (commented):
This is looking good overall. Few things:
- Could you try to break out the shared pieces between ChatScroll  and NotebookScroll? Seems like there's some overlap there. 
- Could you move the dividers to `NotebookScroll` and use the new `ChannelDivider` component?

packages/ui/src/components/Channel/index.tsx
```diff
@@ -119,7 +117,7 @@ export function Channel({
                       >
                         <Spinner />
                       </View>
-                    ) : (
+                    ) : channel.type === 'chat' ? (
```
latter-bolden, line 120: We have to include `dm` and `groupDm` channel types as well. Same for showing the `MessageInput`

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -42,6 +44,16 @@ export default function ChatScroll({
   onPressReplies,
   showReplies = true,
 }: {
+  inverted: boolean;
+  postComponent: (props: {
```
dnbrwstr, line 48: How bout calling this `renderItem` to mirror the `FlatList` api?

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -131,14 +144,15 @@ export default function ChatScroll({
...
+          ) : isFirstPostOfDay && item.type === 'chat' ? (
             <ChannelDivider unreadCount={0} timestamp={item.receivedAt} />
           ) : null}
           <PressableMessage
             ref={activeMessageRefs.current[item.id]}
             isActive={activeMessage?.id === item.id}
           >
-            <ChatMessage
+            {/* ts may yell at you in the editor, but this is fine */}
+            <PostComponent
```
dnbrwstr, line 155: And just doing `{renderItem(props)}` here

packages/ui/src/components/Channel/index.tsx
```diff
@@ -86,6 +85,13 @@ export function Channel({
   const title = utils.getChannelTitle(channel);
   const groups = useMemo(() => (group ? [group] : null), [group]);
 
+  const chatChannel =
```
dnbrwstr, line 88: Seems like a helper we'll want to reuse -- maybe it could go in logic?

dnbrwstr (approved):
Couple small nits, but I think it's good to go.

latter-bolden (approved):

PR #3472, opened 2024-05-01
native: fix wrapping/linebreak issues in chat content

patosullivan:
Rather than returning XStack in LineRenderer, we return Text, everything gets wrapped in that Text component. This fixes the issues we were seeing with line breaks not being applied in some situations and seemingly being applied when they shouldn't have in others. Also, mentions are now just ContactName, with no wrapping view (the wrapping view causes the mention to appear out of alignment with the other text). The wrapping text was also unnecessary for ContactName, since it's a SizableText.

Also added some chat message fixtures and changed the way our fake data generation works a bit, to make it more flexible.

I am having an issue getting the text background to be rounded. borderRadius, borderWidth, etc, don't seem to work on SizableText. 

Fixes LAND-1823
Fixes LAND-1824

Looks like this now:
Simulator Screenshot - iPhone 15 Pro - 2024-05-01 at 14 57 28
Simulator Screenshot - iPhone 15 Pro - 2024-05-01 at 14 57 42

https://github.com/tloncorp/tlon-apps/assets/1221094/bde26433-c1bd-4cad-a9a1-9152e0844339


packages/ui/src/components/ChatMessage/index.tsx
```diff
@@ -1,119 +1,3 @@
-import { PostContent } from '@tloncorp/shared/dist/api';
-import * as db from '@tloncorp/shared/dist/db';
-import { memo, useCallback, useMemo } from 'react';
+import ChatMessage from './ChatMessage';
```
dnbrwstr, line 1: 🙏

apps/tlon-mobile/src/fixtures/ChatMessage.fixture.tsx
```diff
@@ -0,0 +1,33 @@
...
+  </FixtureWrapper>
+);
+
+const ChatMessageFixtureWithBlockQuote = () => (
+  <FixtureWrapper fillWidth>
+    <ChatMessage post={fakeBlockQuotePost} currentUserId="~zod" />
+  </FixtureWrapper>
+);
+
+export default {
```
dnbrwstr, line 30: Great to have this fixture.

dnbrwstr (approved):
Looks great, let's merge it up.

PR #3473, opened 2024-05-01
perf: init noun response

arthyn:
This implements a cached `init` response which operates on a 5 second debounce. In addition, we use nouns as our serialization format instead of JSON. This speeds up our initial load/reconnection of the app by ~2x.

PR Checklist


latter-bolden:
Going to close this but keep the branch

PR #3475, opened 2024-05-02
native: allow for posting references and inline replies, show previews of references

patosullivan:
Allows the user to paste a reference link to the chat input and send it in their message.

Previews the reference above the chat input.

Allows user to quote reply to a message (just renders as a reference).

Fixes LAND-1831
Fixes LAND-1832
Fixes LAND-1833


latter-bolden (changes requested):
code lgtm! Just those two small things — if ref is from the same channel, don't navigate to it on press & clearing the ref after send.

dnbrwstr (approved):
Looks good to me -- might make sense to try using styled context for `asAttachment`.

PR #3477, opened 2024-05-03
native: add line break after blockquote

alecananian:
- Adds line breaks after `blockquote` renders
- Adds `color` prop support for chat messages to support blockquotes rendering in secondary text color

Before
Screenshot 2024-05-02 at 8 53 00 PM

After
Screenshot 2024-05-02 at 9 03 36 PM


patosullivan (approved):
Looks good, thanks Alec!

PR #3478, opened 2024-05-06
mobile: empty channel placeholder & hide input when no write permissions

latter-bolden:
- adds a placeholder for empty groups channels (leaving out DMs since we'll want that sigil block there instead)
- stores channel write permissions on init and checks them before displaying the message input

Fixes LAND-1827
Fixes LAND-1829

### DIFF
--- a/packages/shared/src/api/groupsApi.ts
+++ b/packages/shared/src/api/groupsApi.ts
@@ -1,7 +1,6 @@
 import * as db from '../db';
 import type * as ub from '../urbit';
 import { getChannelType } from '../urbit';
-import { ChannelInit } from './channelsApi';
 import { toClientMeta } from './converters';
 import { poke, scry } from './urbit';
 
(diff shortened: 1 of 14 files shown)


packages/shared/src/api/groupsApi.ts
```diff
@@ -1,6 +1,7 @@
 import * as db from '../db';
 import type * as ub from '../urbit';
 import { getChannelType } from '../urbit';
+import { ChannelInit } from './channelsApi';
```
patosullivan, line 4:
```suggestion
```

Do we need this here?

patosullivan (approved):
Looks good! Maybe one errant import statement? But that's it.

packages/shared/src/api/groupsApi.ts
latter-bolden: Nice catch

PR #3480, opened 2024-05-06
native: message sending fixes

latter-bolden:
This trivially fixes the issue where replies were showing up in the main channel if they're undelivered. Also attempts to fix the issue where messages are getting stuck in pending/sent states.

Repro'ing the stuck states was tricky, so I mostly tested by sending messages to a group on a local ship that I could start/stop at will.

### Notes on Delivery Tracking
Theoretically we should always be hearing a fact when a message is delivered. If for some reason we miss it, we're more susceptible to getting stuck in the wrong state (compared to Web) since we're using an additional client side cache layer behind `react-query`. We can probably alleviate this with smarter refetching logic + `useChannelPosts` configuration, but refining that will be an ongoing process. 

This PR takes an approach that's more blunt. It adds a `syncChannelMessageDelivery` function which attempts to continuously re-sync a channel with exponential backoff if we have any undelivered messages. It's triggered after sending a message/reply. This seemed to reliably avoid messages getting stuck undelivered, but does feel a bit heavy handed. Curious what you guys think.

Fixes LAND-1847

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -28,6 +28,7 @@
   "dependencies": {
     "@urbit/api": "^2.2.0",
     "big-integer": "^1.6.52",
+    "exponential-backoff": "^3.1.1",
     "sorted-btree": "^1.8.1"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


latter-bolden:
No luck figuring out why posts are getting stuck undelivered to begin with. Going to keep digging...

patosullivan (approved):
Code looks good, blunt approach makes sense to me for now.

latter-bolden:
Going to merge this for now and we can consider removing after subscription hardening.

PR #3482, opened 2024-05-06
native: fix ref navigation

patosullivan:
I should have used `navigation.navigate`, not `navigation.push`. Fixed here.

fixes LAND-1845

Makes it so that we don't transition to a new screen when pressing a ref.

### DIFF
--- a/apps/tlon-mobile/src/screens/ChannelScreen.tsx
+++ b/apps/tlon-mobile/src/screens/ChannelScreen.tsx
@@ -174,7 +174,7 @@ export default function ChannelScreen(props: ChannelScreenProps) {
 
   const handleGoToRef = useCallback(
     (channel: db.Channel, post: db.Post) => {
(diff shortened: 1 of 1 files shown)


latter-bolden:
Maybe we need a `pop` and then a `push` if it's a different channel? Unsure what the idiomatic way to navigate to the same screen with different props is

patosullivan:
> Maybe we need a `pop` and then a `push` if it's a different channel? Unsure what the idiomatic way to navigate to the same screen with different props is

This is what I ended up doing, seems to do what we want.

dnbrwstr:
I think this might be the use case that [`replace`](https://reactnavigation.org/docs/native-stack-navigator#replace) is for.

patosullivan:
🫡 replaced with `replace`

latter-bolden (approved):
Confirmed it now works as expected!

PR #3483, opened 2024-05-06
native: refine ref preview on input

patosullivan:
Fixes LAND-1854 by truncating the text in the chat message when it's within a ref that's attached to the input, also adjusts the position/width of the ref preview.

Simulator Screenshot - iPhone 15 Pro - 2024-05-06 at 16 02 56

### DIFF
--- a/packages/ui/src/components/MessageInput/MessageInputBase.tsx
+++ b/packages/ui/src/components/MessageInput/MessageInputBase.tsx
@@ -63,13 +63,19 @@ export const MessageInputContainer = ({
         >
           {Object.keys(references).map((ref) =>
             references[ref] !== null ? (
-              <XStack position="relative" key={ref} width="100%" height="auto">
+              <XStack
+                left={20}
+                position="relative"
+                key={ref}
+                width="100%"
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/MessageInput/MessageInputBase.tsx
```diff
@@ -63,13 +63,19 @@ export const MessageInputContainer = ({
         >
           {Object.keys(references).map((ref) =>
             references[ref] !== null ? (
-              <XStack position="relative" key={ref} width="100%" height="auto">
+              <XStack
+                left={20}
```
latter-bolden, line 67: I think this may be causing the image input + button to look off center

latter-bolden (approved):
One small nit, otherwise lgtm! These feel a lot better truncated

packages/ui/src/components/MessageInput/MessageInputBase.tsx
latter-bolden: Looks like this may actually be the same on develop. My bad

PR #3484, opened 2024-05-06
native: thread fixes

latter-bolden:
- fixes missing nicknames in threads
- extracts upload image logic into reusable hook
- adds support for uploading images to threads
- updates parent post metadata when sending/receiving thread replies (reply count, reply authors, last reply time)

Fixes LAND-1852
Fixes LAND-1850

### DIFF
--- a/apps/tlon-mobile/src/hooks/useImageUpload.ts
+++ b/apps/tlon-mobile/src/hooks/useImageUpload.ts
@@ -1,0 +1,76 @@
+import type {
+  Upload,
+  UploadInfo,
+  UploadParams,
+} from '@tloncorp/shared/dist/api';
+import { handleImagePicked, useUploader } from '@tloncorp/shared/dist/store';
+import { useCallback, useEffect, useState } from 'react';
+
+import { imageSize, resizeImage } from '../utils/images';
+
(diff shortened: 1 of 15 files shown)


apps/tlon-mobile/src/hooks/useImageUpload.ts
```diff
@@ -0,0 +1,76 @@
+import type {
```
patosullivan, line 1: Nice

patosullivan (changes requested):
Code looks good, but when I run locally I'm noticing that the reply counts can be off:


https://github.com/tloncorp/tlon-apps/assets/1221094/d4aa38d6-a9f7-4370-9b52-b260cc9474e8

latter-bolden:
@patosullivan I think this was happening for sent ones because it was updating the parent reply meta both on send and upon hearing the fact. Should be fixed now

patosullivan:
Might be off by one now?

https://github.com/tloncorp/tlon-apps/assets/1221094/13e639f7-0833-4ba3-bc2d-abf17e8d55df

patosullivan (approved):
I think this is good to go for now, but we'll have to watch to see if the (difficult to reproduce) off-by-one weirdness crops up again.

PR #3486, opened 2024-05-07
native: sync calm state, pass it where appropriate

patosullivan:
Fixes LAND-1838

Also fixes an issue with syncing channel perms, where the db would throw an error if we've previously synced channel perms (added onConflictDoUpdate).

### DIFF
--- a/packages/shared/src/db/queries.ts
+++ b/packages/shared/src/db/queries.ts
@@ -38,6 +38,7 @@ import {
   pins as $pins,
   postReactions as $postReactions,
   posts as $posts,
+  settings as $settings,
   threadUnreads as $threadUnreads,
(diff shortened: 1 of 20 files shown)


packages/shared/src/db/queries.ts
```diff
@@ -329,7 +358,13 @@ export const insertChannelPerms = createWriteQuery(
         roleId: writer,
       }))
     );
-    return client.insert($channelWriters).values(writers);
+    return client
+      .insert($channelWriters)
+      .values(writers)
+      .onConflictDoUpdate({
```
latter-bolden, line 364: 🙏🏻

latter-bolden (approved):
Code looking clean, nice work

PR #3487, opened 2024-05-07
native: pending arrows

latter-bolden:
https://github.com/tloncorp/tlon-apps/assets/90741358/7b659b3e-a550-4e64-a4fb-8e42b7d22b18

Adds back in the arrows instead of plain text and adjusts flex spacing to prevent text colliding with the indicator.

Depending on word breaks, long messages do sometimes adjust their layout when the indicator clears. Ochre seemed to have a pretty narrow right gutter, but if we want to reserve space there to comfortably fit the indicators instead, I can do that. (also fine for now if the indicator may change down the line)

Fixes LAND-1853


jamesacklin (approved):
Looks good!

PR #3488, opened 2024-05-07
native: scroller fixes

dnbrwstr:
- Disables scroll-to-unread for now, since it was causing problems
- Fixes date + unread divider positions -- because the divider + message were wrapped in a dummy element, inverted FlatList transform was being applied to them individually rather than both together, which meant that the divider rendered under the message rather than on top. Resolved by just wrapping both in a view.

### DIFF
--- a/packages/ui/src/components/Channel/Scroller.tsx
+++ b/packages/ui/src/components/Channel/Scroller.tsx
@@ -145,7 +145,7 @@ export default function Scroller({
       // (they are objects, not functions)
       const RenderItem = renderItem;
       return (
-        <>
+        <View>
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):
Looks good, makes sense. Also going to hide the "Scroll to Latest" button for now since we're not attempting to navigate on channel entry.

jamesacklin:
> Looks good, makes sense. Also going to hide the "Scroll to Latest" button for now since we're not attempting to navigate on channel entry.

Ah-ha, so that's where it went! I was looking for it this morning to fix it in dark mode. 😛

PR #3489, opened 2024-05-07
channels, chat: /heads endpoint for latest messages

Fang-:
We add a `/heads` scry endpoint, which returns the latest message for every channel/chat. If the optional date path element is specified, we only include results whose latest message is after that date.

Groups-ui gets a `/heads` endpoint also, which simply merges the channels and chat `/heads` scries. Maybe, arguably, this should go into the init endpoint instead.

For channels, we also implement `/changes`, which gives you all changes made to a channel since the given timestamp. We currently cannot reasonably do this for chats, because those don't track changelogs, so including new replies would require checking all threads in the entire backlog.  
We could track new things in state to make that more performant, but we don't do that right now. Adding that should probably be considered out of scope here.

Fixes LAND-1840

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -617,6 +617,54 @@
       [%u ?(%v0 %v1) =kind:c ship=@ name=@ ~]
     =/  =ship  (slav %p ship.pole)
     ``loob+!>((~(has by v-channels) kind.pole ship name.pole))
+  ::
+    ::TODO  rename to /overview or similar?
+    ::TODO  want to include "any changes at all since" flag?
+      [%x %v2 %latest since=?(~ [u=@ ~])]
+    ::REVIEW  this will miss cases where the then-last msg was deleted (or edited).
+    ::        i suppose we would need to look at the logs also?
+    =/  since=(unit id-post:c)
+      ?~  since.pole  ~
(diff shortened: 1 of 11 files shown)


dnbrwstr (approved):
Lgtm!

desk/app/channels.hoon
```diff
@@ -1639,6 +1683,51 @@
...
+      ::NOTE  would manually walk the tree, but logic gets rather confusing,
+      ::      so we just eat the conversion overhead here
+      =/  posts  (bap:on-v-posts:c (lot:on-v-posts:c posts.channel ~ `before))
+      =/  logs   (bap:log-on:c (lot:log-on:c log.channel ~ `before))
+      =|  s=[=_limit older=_`(unit time)`(some before) out=v-posts:c]
+      =<  [older out]
+      |-  ^+  s
+      ?:  =(0 limit.s)  s
+      ?~  posts
+        ::  cannot have logs if posts already empty ::REVIEW right?
```
arthyn, line 1711: I think this is true

arthyn (approved):
lgtm, the changes endpoint seems like it long logs could really make it slow. if we start relying on this we might need some way of traversing it easier.

PR #3490, opened 2024-05-08
native: harden subscriptions

latter-bolden:
- recovers from subscription `quit` events initiated by the client
- handles fatal client errors by rerunning our sync start logic 
- adds additional logging for client events

This is likely what was biting us if you lost connection randomly, in particular if you left the app open overnight and checked it the next day. If Urbit automatically closes the connection after ~12 hours of inactivity, we'll properly detect that and resub.

Tested by connecting to a local ship I could start/stop as needed. Tested with full & seamless resets.

Fixes LAND-1857


latter-bolden:
After talking with @dnbrwstr, the strategy we want to try for keeping the cache synchronized is:

- rely on the subscription for real time updates while you're using the app or it's in the background
- for ensuring no changes get dropped when the app is closed or unexpected network behavior occurs (client hard reset), we'll leverage the new `/changes` scry and trigger it defensively where needed (upon navigating to a channel for example)

This hopefully gives us a good balance without requiring complicated catchup logic. Once that new scry makes it in we'll try connecting things up and see how it feels.

In the meantime, I think this can be merged. Only change in behavior is ensuring the subscriptions stay alive over time.

apps/tlon-mobile/src/components/AuthenticatedApp.tsx
```diff
@@ -19,11 +19,14 @@ function AuthenticatedApp({ initialNotificationPath }: AuthenticatedAppProps) {
   useDeepLinkListener();
 
   useEffect(() => {
-    configureClient(ship ?? '', shipUrl ?? '');
-    sync.start().catch((e) => {
-      console.warn('Sync failed', e);
-    });
-    useStorage.getState().start();
+    const start = () => {
```
dnbrwstr, line 22: This wrapper function seems unnecessary? Or if it is necessary, I don't understand why, and a comment might help ha.

packages/shared/src/api/urbit.ts
```diff
@@ -29,11 +29,13 @@ export function configureClient({
   shipUrl,
   fetchFn,
   verbose,
+  onReset,
```
dnbrwstr, line 32: Doesn't seem like this is actually used?

packages/shared/src/api/urbit.ts
```diff
@@ -43,8 +45,21 @@ export function configureClient({
     logger.log('status-update', status);
   });
 
+  clientInstance.onReconnect = () => {
+    logger.log('client reconnect');
+  };
+
+  clientInstance.on('reset', () => {
+    logger.log('client reset');
```
dnbrwstr, line 53: Will we ever have to do anything in response to these reset events?

dnbrwstr (commented):
Looking good, a few questions.

dnbrwstr (approved):
Makes more sense after discussion, questions are nonblocking.

PR #3492, opened 2024-05-08
native: author names in group last-message previews

jamesacklin:
Adds the nicknamed author to message previews in the Groups list.

- Wraps ChannelListScreen in a ContactsProvider
- Adds an appropriately-sized ContactName to the message preview
- Restyles the last message author in club previews to match
- Makes prettier text representations of images and refs

image


patosullivan (approved):
Looking good

PR #3493, opened 2024-05-08
usability fixes

latter-bolden:
- hide _Today_ in unread banner if not first posts of the day
- placeholder delayed read while we wait for activity
- wire up reactions for messages & thread replies

Fixes LAND-1862


patosullivan (approved):
Looks good, makes sense.

PR #3494, opened 2024-05-08
groups: previewing issues and fact crash fix

arthyn:
This fixes two issues I found. One being that we don't actually fetch group previews because we were setting initial data, so they were potentially always using stale data (unless you don't have that group yet). The other issue is that editing a cabal that someone doesn't know about will crash because we use jab by without checking for existence of key.

PR Checklist

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -1451,7 +1451,9 @@
       go-core
     ::
         %edit
-      ::  TODO: REFACTOR GROUPS PLZ
+      ::  TODO: we don't know why we could be desynced on cabals, but we
+      ::        need to be safe so we don't enter a loop.
+      ::        REFACTOR GROUPS PLZ
       =?  cabals.group  (~(has by cabals.group) sect)
         %+  ~(jab by cabals.group)  sect
         |=  cabal:g
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -1734,7 +1734,10 @@
   ^+  cor
   =/  =wire  /gangs/index/(scot %p ship)
   =/  =dock  [ship dap.bowl]
-  (emit %pass wire %agent dock %watch `path`wire)
+  %-  emil
+  :~  [%pass wire %agent dock %leave ~]
+      [%pass wire %agent dock %watch `path`wire]
+  ==
 ::
 ++  hi-and-req-gang-index
   |=  =ship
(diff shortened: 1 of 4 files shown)


latter-bolden (approved):
JS changes make sense

desk/app/groups.hoon
```diff
@@ -1451,7 +1451,7 @@
       go-core
     ::
         %edit
-      =.  cabals.group
+      =?  cabals.group  (~(has by cabals.group) sect)
         %+  ~(jab by cabals.group)  sect
         |=  cabal:g
         +<(meta meta.diff)
```
Fang-, line 1457: You kind of want something like the following, except the `+go-fact-init` does too many things right now, so kind of hard to flow into that as-is without surprising side-effects.

```suggestion
      ?.  (~(has by cabals.group) sect)
        ::  we have incomplete data, resubscribe to bring us back in sync
        ::
        (go-sub & blah)
      =.  cabals.group
        %+  ~(jab by cabals.group)  sect
        |=  cabal:g
        +<(meta meta.diff)
```

In any case, this is just a crash that says "we're not in sync", and so we should try to get in sync somehow. Sweeping the problem under the rug may be fine, but I'll want to see a very big `::TODO` here and probably a matching linear ticket (or sub-issue on the groups refactor one).

Fang-: None of the above solves the root issue of _becoming_ out of sync, btw, but we don't have much to go on for that rn.

desk/app/groups.hoon
```diff
@@ -1827,9 +1830,16 @@
...
+      |=  delay=?
+      =/  =wire  (welp ga-area /preview)
+      =/  =dock  [p.flag dap.bowl]
+      =/  =path  /groups/(scot %p p.flag)/[q.flag]/preview
+      ^+  cor
+      =^  caz=(list card)  subs
+        (~(unsubscribe s [subs bowl]) wire dock)
+      =^  crds=(list card)  subs
+        (~(subscribe s [subs bowl]) wire dock path delay)
+      (emil (welp caz crds))
```
Fang-, line 1842: I don't think we need any delays on these, since these are "single read" subscriptions. (Using remote scry for these should go on the groups agent refactor list.)

Fang- (commented):
As discussed:

Fang- (approved):
Looks good to me!

PR #3495, opened 2024-05-09
native: automatically save input drafts

patosullivan:
fixes LAND-1837


latter-bolden:
Seeing this when starting a new thread and typing a message. The draft does still seem to get stored though if I navigate away and then back again.

latter-bolden (approved):
Code looks good, works as expected. 

Only weirdness I saw was that one warning.

PR #3496, opened 2024-05-09
native: hide bottom tab bar when in channels

patosullivan:
fixes LAND-1864

looks like this:

https://github.com/tloncorp/tlon-apps/assets/1221094/a2a3356f-eaa3-40c5-ac23-bd69c6e628cd

I'd rather not have it disappear as we're transitioning to the channel screen, any ideas appreciated there.


jamesacklin (approved):
Perfectly suitable for now.

PR #3502, opened 2024-05-10
native: fix "Start a Thread" on Android

latter-bolden:
OTT

Fixes LAND-1872


patosullivan (approved):
Thanks for fixing the syncGroup issue too.

PR #3503, opened 2024-05-10
native: fix input lag

patosullivan:
fixes LAND-1871 by removing most of the content parsing (and all of the content replacement) from onContentUpdate.

Instead, we listen for a paste event within the editor in the webview and send a message up to the react native app, then handle the content parsing/updating in a handler.

onContentUpdate is still used for triggering a mention or storing a draft, but we don't manipulate the content there anymore.


dnbrwstr (approved):
Lgtm!

PR #3505, opened 2024-05-10
native: chat message editing

patosullivan:
Fixes LAND-1836, adds end to end chat editing support.

Looks like this:

https://github.com/tloncorp/tlon-apps/assets/1221094/8e897ab0-1092-43cb-ad07-1b10d03a2726


packages/shared/src/api/postsApi.ts
```diff
@@ -133,6 +133,42 @@ export const sendPost = async ({
...
+  content,
+}: {
+  channelId: string;
+  postId: string;
+  authorId: string;
+  sentAt: number;
+  content: Story;
+}) => {
+  if (isDmChannelId(channelId) || isGroupDmChannelId(channelId)) {
+    return;
```
dnbrwstr, line 150: Maybe throw an error here? Since this shouldn't be getting called.

dnbrwstr, line 150: Maybe throw an error here? Since this shouldn't be getting called.

packages/ui/src/components/MessageInput/index.native.tsx
```diff
@@ -344,6 +371,59 @@ export function MessageInput({
     sendMessage();
   }, [sendMessage]);
 
+  const handleEdit = useCallback(() => {
+    if (!editingPost) {
+      return;
+    }
+
+    editor.getJSON().then(async (json) => {
+      const blocks: Block[] = [];
```
dnbrwstr, line 380: Could you extract the duplicated bits here to helpers?

dnbrwstr (approved):
Looks good! A couple small things.

Also, is there something going on with the checkmark in the vid? Looks a bit squished.

packages/ui/src/components/ChatMessage/ChatMessageActions/MessageActions.tsx
```diff
@@ -4,6 +4,7 @@ import * as db from '@tloncorp/shared/dist/db';
 import * as logic from '@tloncorp/shared/dist/logic';
 import * as store from '@tloncorp/shared/dist/store';
 import * as Haptics from 'expo-haptics';
+import { isDmChannel } from 'packages/ui/src/utils';
```
latter-bolden, line 7: Looks unused

latter-bolden (commented):
Looking good! Just one tiny nit.

Did we want to add thread reply editing as part of this?

patosullivan:
There's still an issue with persisting edits w/replies, looking into it.

dnbrwstr (approved):
Nice, lgtm.

PR #3506, opened 2024-05-10
channels: dont crash on read

arthyn:
Reported by @patosullivan we potentially can crash when doing a read action because we might have a malformed notification thread. Also adds mark that @Fang- forgot.

PR Checklist


patosullivan (approved):
Thank you!

Fang-:
I had done `chat` instead of `dms` but this is fine too. Otherwise identical. My bad!

PR #3507, opened 2024-05-13
native: create & join DMs and Groups

latter-bolden:
WIP


packages/shared/src/api/urbit.ts
```diff
@@ -51,6 +63,7 @@ export function configureClient({
 
   clientInstance.on('reset', () => {
     logger.log('client reset');
+    subWatchers = {};
```
dnbrwstr, line 66: Should we notify watchers on reset? Have them error out or something?

packages/shared/src/db/queries.ts
```diff
@@ -69,6 +70,8 @@ import {
   Unread,
 } from './types';
 
+const logger = createDevLogger('queries', true);
```
dnbrwstr, line 73: Intentional?

packages/shared/src/db/queries.ts
```diff
@@ -101,10 +104,20 @@ export const getSettings = createReadQuery(
...
+    });
+  },
+  []
+);
+
 export const getGroups = createReadQuery(
   'getGroups',
   async ({
-    includeUnjoined,
+    includeUnjoined = false,
```
dnbrwstr, line 120: Does this do anything?

dnbrwstr (commented):
Few starters, still working through.

packages/ui/src/utils/channelUtils.tsx
```diff
@@ -8,8 +8,8 @@ export function getChannelTitle(channel: db.Channel) {
   if (channel.type === 'dm') {
     const member = channel.members?.[0];
     if (!member) {
-      console.warn('bad dm channel', channel.id, 'missing contact');
-      return 'Untitled DM';
+      console.log('bad dm channel', channel.id, 'missing contact');
```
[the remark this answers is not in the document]
latter-bolden: It was triggering a bunch before I fixed contact passing. It will still happen if we get invited and don't have the contact already. Backend should automatically add it for DMs, but I need to look again at if/how we're subbing to that from the client.

In any case, I left it bumped down to a log vs warning since we have a meaningful fallback with the @p.

packages/shared/src/db/queries.ts
latter-bolden: From Alec's work, prevents unjoined gangs from showing up in the query. Seems defensive. 

Whether it's useful or not will be more clear when I tackle the group reference issue after this lands.

packages/shared/src/store/sync.ts
```diff
@@ -548,9 +577,10 @@ async function persistPagedPostData(
   }
 }
 
-export const start = async () => {
+export const start = async (currentUserId: string) => {
```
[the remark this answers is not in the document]
dnbrwstr: Gonna resolve this one -- think we covered it in conversation.

packages/shared/src/db/queries.ts
latter-bolden: Oh I see what you mean, guessing he was just trying to make it more explicit. Removed the default.

packages/ui/src/hooks/contactSorters.ts
```diff
@@ -0,0 +1,204 @@
...
+  const pals = useMemo(() => new Set([]), []);
+  const sortedContacts = useMemo(() => {
+    if (sortOrder.length === 0) {
+      return contacts;
+    } else {
+      return sortContacts(contacts, sortOrder, pals);
+    }
+  }, [contacts, sortOrder, pals]);
+
+  if (isValidQuery(query)) {
```
dnbrwstr, line 106: Should memoize this bit

packages/shared/src/api/channelsApi.ts
```diff
@@ -303,3 +341,63 @@ function toThreadUnreadStateData(unread: ub.Unread): db.ThreadUnreadState[] {
     };
   });
 }
+
+export function createPendingMultiDmChannel(
```
dnbrwstr, line 345: I think this and `createSingleDmChannel` should either be named differently or go in a different file -- it's a little confusing that they're named similarly to `createDefaultChannel` but don't actually mutate anything on the backend.

packages/shared/src/api/channelsApi.ts
```diff
@@ -245,6 +245,44 @@ export const toChannelsUpdate = (
   return { type: 'unknown' };
 };
 
+export const createDefaultChannel = async ({
```
dnbrwstr, line 248:
```suggestion
export const createDefaultGroupChannel = async ({
```

packages/shared/src/api/chatApi.ts
```diff
@@ -13,43 +16,199 @@ export const markChatRead = (whom: string) =>
...
+        ship: channel.id,
+        ok: accept,
+      },
+    });
+  }
+
+  const action = multiDmAction(channel.id, {
+    team: { ship: currentUserId, ok: accept },
+  });
+  console.log(`action:`, action);
```
dnbrwstr, line 61: Stray log

packages/shared/src/api/chatApi.ts
```diff
@@ -13,43 +16,199 @@ export const markChatRead = (whom: string) =>
...
+      id,
+      diff: {
+        uid: '0v3',
+        delta,
+      },
+    },
+  };
+}
+
+export const respondToMultiDMInvite = {};
```
dnbrwstr, line 136: ?

packages/shared/src/db/types.ts
```diff
@@ -78,3 +78,12 @@ export type Pin = BaseModel<'pins'>;
 export type PinType = schema.PinType;
 export type Settings = BaseModel<'settings'>;
 export type PostWindow = BaseModel<'postWindows'>;
+
+// Type guards
+export function isGroup(obj: Channel | Group): obj is Group {
```
dnbrwstr, line 83: Would it make sense to use one of the methods that gets type from id here? This feels a little fragile.

packages/shared/src/store/sync.test.ts
```diff
@@ -88,7 +88,8 @@ test('syncs contacts', async () => {
   storedContacts.forEach((c) => {
     const original = contactsData[c.id];
     expect(original).toBeTruthy();
-    expect(original.groups?.length ?? 0).toEqual(c.pinnedGroups.length);
+    // TODO: fix contact groups test
+    // expect(original.groups?.length ?? 0).toEqual(c.pinnedGroups.length);
```
dnbrwstr, line 92: Can this be uncommented now that the pinned groups sync works again?

packages/ui/src/components/AddChats/AddDmSheet.tsx
```diff
@@ -0,0 +1,120 @@
...
+  open,
+  onOpenChange,
+  goToDm,
+}: {
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+  goToDm: (participants: string[]) => void;
+}) {
+  const insets = useSafeAreaInsets();
+  const [contentScrolling, setContentScrolling] = useState(false);
```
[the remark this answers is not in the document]
dnbrwstr: Yeah, it's a pretty thin wrapper over the bottom sheet lib. Ok, let's leave for now, but we should consider just using that lib directly in the future.

packages/ui/src/components/ContactRow.tsx
```diff
@@ -0,0 +1,72 @@
...
+  selectable = false,
+  onPress,
+  ...rest
+}: {
+  contact: db.Contact;
+  onPress: (id: string) => void;
+  selectable?: boolean;
+  selected?: boolean;
+} & Omit<ListItemProps, 'onPress'>) {
+  const displayName = useMemo(
```
dnbrwstr, line 23: Is there a helper for this? If not it's a good candidate for one.

packages/ui/src/components/Badge.tsx
dnbrwstr, line 1: This is great; could you replace the `RoleBadge` in `AuthorRow` with it so we're not duplicating?

packages/ui/src/components/AddChats/ViewUserGroupsWidget.tsx
```diff
@@ -0,0 +1,101 @@
...
+  userId: string;
+  onSelectGroup: (group: db.Group) => void;
+}) {
+  const [state, setState] = useState<State>({
+    loading: true,
+    error: null,
+    groups: [],
+  });
+
+  useEffect(() => {
```
dnbrwstr, line 30: Would it make sense to make this a `dbHook` and use react query?

dnbrwstr (changes requested):
Looking good! Besides the few comments above, remaining stuff is:

- Refactor api methods so we don't need to pass in current user
- If there's anyway we can abstract the scroller managements stuff it'd be really nice, as it's kind of threaded through in a number of places

jamesacklin:
Exciting stuff. Do we want an ephemeral build for this?

dnbrwstr (approved):
Code lgtm!

PR #3509, opened 2024-05-13
native: hide send arrow if editor is empty

patosullivan:
fixes LAND-1875 by hiding the send arrow when we have nothing to send.


latter-bolden (approved):
Code looks good, worked locally

PR #3510, opened 2024-05-13
native: ellipsize long contact display names

patosullivan:
fixes LAND-1874

Simulator Screenshot - iPhone 15 Pro - 2024-05-13 at 13 31 51

### DIFF
--- a/packages/ui/src/components/ContactName.tsx
+++ b/packages/ui/src/components/ContactName.tsx
@@ -139,7 +139,13 @@ export default function ContactName({
   }
 
   return (
-    <SizableText aria-label={formattedId.ariaLabel} {...rest}>
+    <SizableText
+      ellipsizeMode="tail"
+      numberOfLines={1}
+      maxWidth="75%"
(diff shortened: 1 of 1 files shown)


dnbrwstr (approved):
Lgtm!

packages/ui/src/components/ContactName.tsx
```diff
@@ -139,7 +139,13 @@ export default function ContactName({
   }
 
   return (
-    <SizableText aria-label={formattedId.ariaLabel} {...rest}>
+    <SizableText
+      ellipsizeMode="tail"
+      numberOfLines={1}
+      maxWidth="75%"
```
latter-bolden, line 145: Will we always want this `maxWidth` here? If not, maybe we pass that in at the call site where needed

patosullivan: I think it's probably ok, it'll just make sure that we keep a reasonable length anywhere that the parent has a set width, afaik

latter-bolden (approved):

PR #3513, opened 2024-05-14
native, ui: move from TabStack to new custom NavBar component

jamesacklin:
Renovates the app navigation to use nested Native Stacks and a custom NavBar component.

- Rather than using createBottomTabNavigator, uses createNativeStackNavigator for all nested "home" screens (ChatList, Channel, Post, ChannelSearch).
- Adds a custom navigation component (NavBar) in packages/ui which seeks to be as generic as possible
- Adds a NavBarView to the app screens themselves, which in turn takes the navigation prop from the parent screen and handles all the navigate() calls itself

Functionality is exactly the same as before but eliminates the gray box on view transition (fixes LAND-1873). Thanks to @patosullivan for the talk-through assist.

Note that this is an absolutely-positioned element, so any screen that uses a scroller that extends below the bottom edge of the screen will need to be padded up at the bottom.

### DIFF
--- a/packages/ui/src/components/ChatList.tsx
+++ b/packages/ui/src/components/ChatList.tsx
@@ -41,6 +41,7 @@ export function ChatList({
       gap: '$s',
       paddingTop: '$l',
       paddingHorizontal: '$l',
+      paddingBottom: '$6xl',
     },
     { resolveValues: 'value' }
   ) as StyleProp<ViewStyle>;
--- a/packages/ui/src/components/ChatList.tsx
+++ b/packages/ui/src/components/ChatList.tsx
@@ -8,12 +8,14 @@ import {
   StyleProp,
   ViewStyle,
 } from 'react-native';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
 import { useStyle } from '../core';
 import ChannelListItem from './ChannelListItem';
 import { GroupListItem } from './GroupListItem';
 import { ListItemProps } from './ListItem';
 import { ListSectionHeader } from './ListSectionHeader';
+import { navHeight } from './NavBar/NavBar';
 import { SwipableChatRow } from './SwipableChatListItem';
 
 export function ChatList({
@@ -25,6 +27,8 @@ export function ChatList({
   onPressItem?: (chat: db.Channel) => void;
   onLongPressItem?: (chat: db.Channel) => void;
 }) {
+  const { bottom } = useSafeAreaInsets();
+
   const data = useMemo(() => {
     if (pinned.length === 0) {
       return [{ title: 'All', data: unpinned }];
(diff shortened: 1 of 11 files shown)


packages/ui/src/components/ChatList.tsx
```diff
@@ -41,6 +41,7 @@ export function ChatList({
       gap: '$s',
       paddingTop: '$l',
       paddingHorizontal: '$l',
+      paddingBottom: '$6xl',
```
dnbrwstr, line 44: Will this value be correct across different sized safe areas? And/or if we tweak navbar size? Would prefer to derive both from a `navBarHeight` constant + `insets.bottom` if possible.

packages/ui/src/components/NavBar/NavIcon.tsx
```diff
@@ -0,0 +1,63 @@
...
+  id: string;
+  focused: boolean;
+  contact: Contact;
+  isLoading: boolean;
+  onPress?: () => void;
+}) {
+  return isLoading && !contact ? null : (
+    <View flex={1} onPress={onPress} alignItems="center" paddingTop={'$s'}>
+      <Avatar
+        width={22}
```
dnbrwstr, line 24: Looks like ochre is size `22`, radius `4`?

packages/ui/src/components/ProfileScreenView.tsx
```diff
@@ -32,7 +32,12 @@ export function Wrapped(props: Props) {
 
   return (
     <ScrollView>
-      <YStack flex={1} paddingHorizontal="$xl" paddingTop={top}>
+      <YStack
+        flex={1}
+        paddingHorizontal="$xl"
+        paddingTop={top}
+        paddingBottom={'$6xl'}
```
dnbrwstr, line 39: Same comment re: bottom padding.

packages/ui/src/tamagui.config.ts
```diff
@@ -54,6 +54,8 @@ export const tokens = createTokens({
     '2xl': 24,
     '3xl': 32,
     '4xl': 48,
+    '5xl': 64,
```
dnbrwstr, line 57: Intentional? Not seeing these used.

dnbrwstr (approved):
Looks great! A few nitpicks.

packages/ui/src/components/NavBar/NavIcon.tsx
dnbrwstr: `20`** oops

packages/ui/src/components/ChatList.tsx
jamesacklin: Makes sense, will do

latter-bolden (approved):

PR #3514, opened 2024-05-14
native: version negotiation for channels

patosullivan:
Fixes LAND-1828. We skip the db and just query the api directly. We'd have to sync regularly and write to/read from the db anyway, this saves a step and squeaks out a little performance.

We assume there is a match until we hear otherwise. Users shouldn't see the message about a mismatch unless we know that one exists.

Reuses/adapts logic from tlon-web.

Looks like this:
Simulator Screenshot - iPhone 15 Pro - 2024-05-14 at 13 37 42


latter-bolden (approved):
Code looks good!

PR #3516, opened 2024-05-14
[draft] native: scroller post loading

dnbrwstr:


packages/shared/src/debug.ts
```diff
@@ -31,3 +31,8 @@ export async function logDuration<T>(
   logger.log('end:time:' + label + '', `(${Date.now() - start}ms)`);
   return result;
 }
+
+// Attempts to slice off rarely significant digits for more legible logging.
+export function shortPostId(postId: string) {
```
patosullivan, line 36: Good idea.

packages/shared/src/store/sync.ts
```diff
@@ -102,10 +103,23 @@ export const handleChannelsUpdate = async (update: api.ChannelsUpdate) => {
...
+          channelId: update.post.channelId,
+          posts: [update.post],
+        });
       }
-
-      // finally, always insert the post itself
-      await db.insertChannelPosts(update.post.channelId, [update.post]);
+      // For non-replies, we skip inserting and sync from the api instead.
+      // This is significantly slower, but ensures that windowing stays correct. Better solutions would be:
+      // 1. Have the server send the previous post id so that we can match this up with an older window if it exists
```
patosullivan, line 114: This would be easy to add to the backend

packages/shared/src/store/useChannelPosts.ts
```diff
@@ -1,37 +1,47 @@
...
-import { useEffect } from 'react';
+import { useEffect, useMemo } from 'react';
 
 import * as db from '../db';
 import { createDevLogger } from '../debug';
 import * as sync from './sync';
 import { useKeyFromQueryDeps } from './useKeyFromQueryDeps';
 
-const postsLogger = createDevLogger('useChannelPosts', false);
+const postsLogger = createDevLogger('useChannelPosts', true);
```
patosullivan, line 9: Do we need this to be true?
```suggestion
const postsLogger = createDevLogger('useChannelPosts', false);
```

packages/ui/src/components/Channel/index.tsx
```diff
@@ -88,14 +91,31 @@ export function Channel({
...
   negotiationMatch: boolean;
+  hasNewerPosts?: boolean;
+  hasOlderPosts?: boolean;
 }) {
   const [inputShouldBlur, setInputShouldBlur] = useState(false);
-  const title = utils.getChannelTitle(channel);
+  const title = channel ? utils.getChannelTitle(channel) : '';
   const groups = useMemo(() => (group ? [group] : null), [group]);
-  const canWrite = utils.useCanWrite(channel, currentUserId);
+  const canWrite = true;
```
patosullivan, line 100: Why set this to true?

patosullivan (approved):
Looks good, just two questions

packages/ui/src/components/Channel/index.tsx
dnbrwstr: Ahh thanks, good catch.

packages/shared/src/store/sync.ts
dnbrwstr: Ya, @Fang- and I discussed some better ideas here -- will update with the changes endpoint.

PR #3517, opened 2024-05-14
native: handle delete post subscription events

patosullivan:
Fixes LAND-1887.

Also, we don't need to specifically handle edit, that's handled via the `addPost` case (it comes in via postResponse.set).


latter-bolden (approved):
Nice, lgtm

Also good to know on edit, thanks for calling out

PR #3519, opened 2024-05-15
native: add group invitations to chat list and handle accept/reject rsvps

alecananian:
- Adds fetching of pending group invitations
- Adds group invitations to the top of the chat list
- Adds subscriptions for groups and gangs updates to determine when invitations are accepted and joined

Marking as a draft because there are a couple parts that could likely use some more work:

- To make the invitations show on the chat list, I'm creating a placeholder channel. If there's another approach you recommend let me know! Otherwise, need to find a way to delete that placeholder when the group is eventually joined
- The subscriptions may be overkill and could instead be merged with the `trackedPoke` changes from #3507 

Invitation in chat list

Screenshot 2024-05-10 at 3 07 32 PM

Invitation sheet and joining status

Screenshot 2024-05-10 at 3 07 32 PM

Screenshot 2024-05-10 at 3 07 32 PM

Invitation joining status in chat list

Screenshot 2024-05-10 at 3 09 31 PM


latter-bolden:
This work has been pulled into #3507,

PR #3520, opened 2024-05-15
native: fix image uploading for hosted users & custom S3s

latter-bolden:
Adds a custom native uploader that uses an Expo package to handle uploads instead of rolling our own `fetch` based implementation. Mostly maintains the existing filestore implementation with the addition of passing the native uploader down where needed.

Tested on simulator & with demo builds using a hosted ship and a local ship that I wired up with Digital Ocean and Native Planet S3 configs.

Fixes LAND-1856


latter-bolden:
Going to do one more demo build to confirm this is working after my last edits, but should be ready for review.

jamesacklin:
Functional stamp from me--I installed the test build and was able to post an image both in a chat channel and in a thread. Good work!

packages/shared/src/store/storage/upload.ts
```diff
@@ -121,31 +126,70 @@ export const useFileStore = create<FileStore>((set, get) => ({
...
-    // // TODO: fix upload compression for Safari and iOS webview
-    // if (isImageFile(file)) {
-    // compressedFile = await imageCompression(file, compressionOptions);
-    // }
-    // } catch (error) {
-    // console.log({ error });
-    // }
+
+    // Native app uses its own uploader which gets passed in
+    if (nativeUploader) {
```
dnbrwstr, line 152: Does this still handle resizing + compressing the image? I think those are important to get in as massive imaages cause all kinds of pain, and photo roll images are pretty massive by default.

packages/shared/src/store/storage/upload.ts
```diff
@@ -121,31 +126,70 @@ export const useFileStore = create<FileStore>((set, get) => ({
...
+        updateStatus(uploader, key, 'error', 'Upload failed');
+      }
+
+      return;
+    }
+
+    // Pre-existing web upload logic
 
     // Logic for uploading with Tlon Hosting storage.
     if (config.service === 'presigned-url' && config.presignedUrl) {
```
dnbrwstr, line 195:
```suggestion
    if (isHostedUpload) {
```

packages/shared/src/store/storage/upload.ts
```diff
@@ -3,7 +3,7 @@ import { getSignedUrl } from '@aws-sdk/s3-request-presigner';
 import { deSig, formatDa, unixToDa } from '@urbit/aura';
```
dnbrwstr, line 3: This file is getting a bit convoluted, especially the main `upload` function. I also wonder whether there's anything that can be factored out of the two upload flows, especially once we're doing resizing etc. Any ideas? Or I'm down to discuss. Nonblocking, but would be nice if we can improve.

packages/shared/src/store/storage/index.ts
```diff
@@ -2,3 +2,4 @@ export * from './reducer';
 export * from './storage';
 export * from './upload';
 export * from './utils';
+// export * from './useStorageUpload';
```
dnbrwstr, line 5: Intentional?

dnbrwstr (changes requested):
Nice, this is looking good overall. Mostly nonblocking stuff, but I am curious about resizing + compression as I think those are important -- once someone uploads a massive image it's basically stuck in the scrollback forever.

If you do get into that, one other thing to consider: we should probably apply some compression/resizing before we even upload the original. Could still keep the original pretty large, but no point in taking the time to upload a 20mb png or whatever.

packages/shared/src/store/storage/upload.ts
latter-bolden: That piece should be unchanged. We resize the image in `useImageUpload` before it ever gets passed to the uploader in file store.

dnbrwstr: Ahh, ok, I see where I tripped up. Not related to your PR, but while you're in there, would you mind renaming `imageSizer` to `getImageSize`? I think that'd be clearer.

packages/shared/src/store/storage/upload.ts
latter-bolden: Yeah strong agree here. I should have added more context in the PR description, but I basically started down this path and then backed away because it felt hard to do without the scope creeping quite a bit.

My main takeaway was that if we're going to spend the time to refactor this, we should try to do so in a way that will set us up for any features we'll want for launch or soon after. The big one being avoiding uploading attachments until after you actually send the message. Doing this in a way where we can still update the UI optimistically and (ideally) handle retries got complicated enough that I didn't wanna bite it off here. This becomes more important if we support multiple attachments or videos which I could see being another thing we do sooner rather than later.

I'll log a ticket laying out some of the areas where I think we need improvement to have good baseline UX  and mention how a refactor here would need to be part of that.

dnbrwstr (approved):
Looks good!

PR #3524, opened 2024-05-16
chat: send more facts to / subscriptions

Fang-:
Chat was already sending some updates (block & msg toggle changes) to the top-level subscription path (and surprisingly isn't sending those anywhere else), but wasn't doing the same for updates about individual chats.

Here, we chuck `/` into the list of paths to update for most response-level subscription updates, to mirror the "firehose" style that channels' `/` endpoint has.

Closes LAND-1880.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -1196,7 +1196,7 @@
     |=  =action:club:c
     =/  =cage  chat-club-action+!>(action)
     =.  cor
-      (emit %give %fact ~[/clubs] cage)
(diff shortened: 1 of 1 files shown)


arthyn:
I think you want to add them for `give-unread` and `di-give-writs-diff` as well

latter-bolden (approved):
Looking good! I'll work on listening for live updates & moving existing subs over once the DM create/invite  PR lands.

re unreads: having them in the firehose makes sense to me, but isn't critical for our needs on the client. Will defer to y'all on that one

PR #3525, opened 2024-05-16
native: handle channel perms updates and all groups subscription updates

patosullivan:
Fixes LAND-1863 and LAND-1891 by 

1) cleaning up channels subscription updates
2) adding in other update types (several of which aren't used yet, not sure if they will be, but for those I didn't add any new queries or calls to existing queries)
3) adding in groups subscription handling (necessary for perms changes, went ahead and added other updates that we'll need to). This includes a lot of db schema changes and some new queries.

I also got rid of the require cycle issues by removing the need for one of the require cycles and adding one to requireCycleIgnorePatterns.


packages/shared/src/db/schema.ts
```diff
@@ -166,6 +168,7 @@ export const groups = sqliteTable('groups', {
   id: text('id').primaryKey(),
   ...metaFields,
   isSecret: boolean('is_secret'),
+  publicOrPrivate: text('public_or_private').$type<'public' | 'private'>(),
```
dnbrwstr, line 171: Maybe this could be `visibility`? Also, can this be combined this with `isSecret`? Or are those separate properties?

packages/shared/src/db/schema.ts
```diff
@@ -176,6 +179,7 @@ export const groupsRelations = relations(groups, ({ one, many }) => ({
   roles: many(groupRoles),
   members: many(chatMembers),
   navSections: many(groupNavSections),
+  flaggedContent: many(groupFlaggedContent),
```
dnbrwstr, line 182: Wdyt about `flaggedPosts` for this?

packages/shared/src/db/schema.ts
```diff
@@ -226,6 +230,137 @@ export const chatMembers = sqliteTable(
...
+      fields: [groupFlaggedContent.flaggedByContactId],
+      references: [contacts.id],
+    }),
+  })
+);
+
+export const groupMemberInvites = sqliteTable(
+  'group_member_invites',
+  {
+    groupId: text('group_id')
```
dnbrwstr, line 278: Is this compatible with @latter-bolden's stuff?

packages/shared/src/store/sync.ts
```diff
@@ -66,9 +66,202 @@ const resetUnreads = async (unreads: db.Unread[]) => {
   });
 };
 
+async function handleGroupUpdate(update: api.GroupUpdate) {
+  logger.log('received group update', update.type);
```
dnbrwstr, line 70: This function is beautiful 😍

packages/ui/src/components/ChatMessage/ChatEmbedContent.tsx
```diff
@@ -59,8 +58,13 @@ export default function ChatEmbedContent({
   }
 
   return (
-    <Text textDecorationLine="underline" lineHeight="$m" onPress={openLink}>
-      <InlineContent story={content} />
+    <Text
```
dnbrwstr, line 61: 🙏

packages/shared/src/db/queries.ts
```diff
@@ -350,6 +363,62 @@ export const insertGroups = createWriteQuery(
   ]
 );
 
+export const updateGroup = createWriteQuery(
```
dnbrwstr, line 366: Optional: might be nice/pretty straightforward to add tests to some of these added queries, or at least any that might be error prone.

packages/shared/src/api/groupsApi.ts
```diff
@@ -103,6 +639,24 @@ export function toClientGroup(
   isJoined: boolean
 ): db.Group {
   const rolesById: Record<string, db.GroupRole> = {};
+  const flaggedContent: db.GroupFlaggedContent[] = group['flagged-content']
```
dnbrwstr, line 642: Might break this up for legibility.

dnbrwstr (approved):
Nice, this looks great. Nitpicks only.

packages/shared/src/db/schema.ts
patosullivan: They're separate properties, on the backend a group could actually be set with a cordon of `shut` and still not be `secret`. This gives us the three types: public, private, secret. It's annoying that it's setup this way.

packages/shared/src/db/schema.ts
patosullivan: Good question, will have to look.

packages/shared/src/db/schema.ts
dnbrwstr: Cool, makes sense.

packages/shared/src/db/schema.ts
patosullivan: With a few changes I've made, this should be fine

PR #3527, opened 2024-05-17
native: add root gallery channel view and gallery channel input

patosullivan:
Adds the root gallery view and the ability to add rich text or image posts. Will also render a "text post" that consists of a single link to an image as an image post.

Also fixes an issue where we weren't rendering similar "text post is just a link to an image" posts in chat.

Looks like this:

https://github.com/tloncorp/tlon-apps/assets/1221094/7d08ea19-af99-452c-a8dd-a69e48db5cfc

fixes LAND-1812
fixes LAND-1821


packages/shared/src/api/postsApi.ts
```diff
@@ -117,12 +121,36 @@ export const sendPost = async ({
     return;
   }
 
+  const kindData = (): ub.KindData => {
```
dnbrwstr, line 124: Could this be extracted into a `toWrit` helper or something? We prob need some similar things for `editPost`, and maybe in other places.

packages/shared/src/store/postActions.ts
```diff
@@ -7,10 +7,12 @@ export async function sendPost({
   channel,
   authorId,
   content,
+  metaData,
```
dnbrwstr, line 10: Extremely minor nit: how about `metadata` here, to match `db.PostMetadata`?

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -187,7 +190,7 @@ export default function Scroller({
       // (they are objects, not functions)
       const RenderItem = renderItem;
       return (
-        <View onLayout={() => handleItemLayout(item, index)}>
+        <View padding="$m" onLayout={() => handleItemLayout(item, index)}>
```
dnbrwstr, line 193: Intentional? This adds extraneous padding around chat posts + would probably be better applied to the actual message or divider component getting rendered.

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -162,11 +162,14 @@ export default function Scroller({
...
 
+  const theme = useTheme();
+
   // Used to hide the scroller until we've found the anchor post.
   const style = useMemo(() => {
     return {
       opacity: hasFoundAnchor ? 1 : 0,
-      backgroundColor: 'white',
+      backgroundColor: theme.background.val,
+      paddingTop: channelType === 'gallery' ? 8 : 0,
```
dnbrwstr, line 172: Could this be applied by the component wrapping this? Either way a comment explaining what we're compensating for here would be helpful.

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -162,11 +162,14 @@ export default function Scroller({
...
     };
   }, [hasFoundAnchor]);
```
dnbrwstr, line 175:
```suggestion
  }, [hasFoundAnchor, channelType, theme.background.val]);
```

packages/ui/src/components/MessageInput/AttachmentButton.native.tsx
```diff
@@ -9,6 +9,7 @@ import { useEffect, useState } from 'react';
 import { Add } from '../../assets/icons';
 import { Spinner, View } from '../../core';
 import { ActionSheet } from '../ActionSheet';
```
dnbrwstr, line 11: Some extraneous deps here now.

packages/ui/src/components/Channel/index.tsx
```diff
@@ -119,6 +136,29 @@ export function Channel({
     return null;
   }, [selectedPostId, channel]);
 
+  const sendGalleryImagePost = useCallback(async () => {
```
dnbrwstr, line 139: Could `messageSender` be generalized to handle this part too?

packages/ui/src/components/Channel/index.tsx
```diff
@@ -135,6 +175,7 @@ export function Channel({
                   justifyContent="space-between"
                   width="100%"
                   height="100%"
+                  position="relative"
```
dnbrwstr, line 178: Probably extraneous? RN views are `position: relative` by default.

dnbrwstr (changes requested):
Looking good -- still working through, but some initial feedback here.

packages/ui/src/components/Channel/Scroller.tsx
patosullivan: Not intentional, was testing something, will fix.

packages/ui/src/components/Channel/Scroller.tsx
patosullivan: Are you referring to the color or the padding? I set the color to just be whatever the theme's background is because setting it to `white` gives a white background when the app is in dark mode.

The padding could be set elsewhere, yeah.

dnbrwstr: Yep the background makes sense, was just looking at the padding.

packages/shared/src/api/postsApi.ts
```diff
@@ -89,13 +89,17 @@ export const sendPost = async ({
...
   sentAt,
   content,
+  channelType,
+  metadata,
 }: {
   channelId: string;
   authorId: string;
   sentAt: number;
   content: Story;
+  channelType: db.ChannelType;
```
dnbrwstr, line 99: Could we skip this param and use `getChannelType` on the id?

packages/ui/src/components/Channel/Scroller.tsx
patosullivan: That padding was totally extraneous afaict, just removed it.

packages/ui/src/components/GalleryInput/index.native.tsx
```diff
@@ -0,0 +1,531 @@
+import {
```
dnbrwstr, line 1: Should extract everything duplicated from `MessageInput` in some sensible way.

packages/ui/src/components/IconButton.tsx
```diff
@@ -1,18 +1,30 @@
...
+  ThemeTokens,
+  useTheme,
+} from '../core';
 
 export function IconButton({
   children,
   onPress,
   size = '$s',
   color = '$primaryText',
+  backgroundColor = '$background',
```
dnbrwstr, line 17: I think it'd make more sense to create a separate component for the floating action button rather than introduce this extra config here.

dnbrwstr (commented):
Few more...

packages/ui/src/components/GalleryInput/GalleryInputBase.tsx
```diff
@@ -0,0 +1,135 @@
...
+  isEditing?: boolean;
+  cancelEditing?: () => void;
+  onPressEdit?: () => void;
+  editorIsEmpty: boolean;
+}>) => {
+  const { references, setReferences } = useReferences();
+
+  return (
+    <YStack position="relative" width="100%" height="100%">
+      {Object.keys(references).length ? (
```
dnbrwstr, line 55: Same question here -- can we extract the duplicated bits?

packages/ui/src/components/GalleryPost/GalleryPost.tsx
```diff
@@ -0,0 +1,222 @@
...
+export default function GalleryPost({
+  post,
+  onPress,
+  onLongPress,
+}: {
+  post: db.Post;
+  onPress?: () => void;
+  onLongPress?: (post: db.Post) => void;
+}) {
+  const content = useMemo(
```
dnbrwstr, line 33: Some of the helpers here look familiar -- could they be pulled in from anything existing?

packages/ui/src/components/ChatMessage/ChatContent.tsx
```diff
@@ -55,9 +55,15 @@ function ShipMention({ ship }: { ship: string }) {
 export function InlineContent({
   story,
   color = '$primaryText',
+  isGalleryPost = false,
```
dnbrwstr, line 58: Proposal for this + the ref `inGalleryPost` param: how about we use `viewMode`? Which could be `'chat' | 'block' | 'note'. Since we'll probably need similar switches for note, and maybe feed etc. as time goes on.

packages/ui/src/components/GalleryPost/GalleryPost.tsx
patosullivan: Will take a look.

packages/ui/src/components/ChatMessage/ChatContent.tsx
patosullivan: Good point

packages/ui/src/components/GalleryInput/index.native.tsx
patosullivan: Combined the inputs in latest commits

dnbrwstr (approved):
Great, lgtm!

dnbrwstr:
Actually, last thing: I'm seeing some layout issues in non-gallery channels, are you encountering that?

Screenshot 2024-05-21 at 12 28 21 PM

patosullivan:
@dnbrwstr Yeah, just noticed that in DMs. Looks normal in chat channels. Did you see it anywhere other than DMs/groupDMs?

dnbrwstr:
I think I was seeing it pretty much everywhere? The second image is from the homestead channel.

patosullivan:
Oh, is it extra vertical padding on the chat channels?

patosullivan:
The other screenshot must have been a dm/group dm, I see what the issue was there.

PR #3530, opened 2024-05-20
activity: follow-up from integration

arthyn:
This goes through and adds everything that was TODO on my previous PR, except activity feed.

- [X] Allow notification settings for threads
- [X] Remove current "full-info" scry on frontend and wrap unreads into a new init scry
- [X] Add migration from old unreads
- [X] Add handling for devs who enabled %activity already
- [X] Show unread counts based on a setting?
- [X] Verify everything is in mark-warmer that needs to be

More follow-up:
- [ ] Fix issue where clearing main chat window makes dot go away, but then it comes back because of unread threads
- [ ] TLON-1689
- [ ] TLON-1556

PR Checklist

Fixes TLON-1924, TLON-1562, TLON-1414


latter-bolden:
I can't get this installed on fresh ships. The other agents bump, but activity fails

```
gall: failed
/sys/vane/gall/hoon:<[2.021 15].[2.021 36]>
/app/activity/hoon:<[31 5].[33 17]>
/app/activity/hoon:<[32 7].[32 20]>
/app/activity/hoon:<[32 12].[32 20]>
/app/activity/hoon:<[73 3].[78 31]>
/app/activity/hoon:<[74 3].[78 31]>
/app/activity/hoon:<[75 3].[78 31]>
/app/activity/hoon:<[75 12].[75 26]>
/app/activity/hoon:<[152 3].[200 23]>
/app/activity/hoon:<[153 3].[200 23]>
/app/activity/hoon:<[154 3].[200 23]>
/app/activity/hoon:<[154 7].[156 7]>
/gx/~ralfyr-sidlyn-rispeg-nocput--fabsel-nacreg-linwyc-marzod/chat/~2024.5.28..15.30.19..8057/full/noun
```

If I comment out the new `++  on-init` logic, it goes through.

arthyn:
@latter-bolden latest push should fix

latter-bolden (approved):
This is looking good, let's get it in!

To the extent that there are any lingering UI/UX todos, we can target them as separate PRs to `activity-release`

desk/app/activity.hoon
```diff
@@ -71,59 +71,192 @@
...
+    |=  [=time =reply:ch]
+    =/  mention
+      (was-mentioned:ch-utils content.reply our.bowl)
+    [time %dm-reply key parent whom content.reply mention]
+  (welp writs replies)
+++  volume-type
+  $:  base=level:v
+      area=(map flag:g level:v)  ::  override per group
+      chan=(map nest:g level:v)  ::  override per channel
+  ==
```
Fang-, line 208: Should use something we import from `/sur/groups`, so that we're sure these two agents stay in sync wrt to this type.

Fang- (commented):
Small request, looks good otherwise.

desk/app/activity.hoon
```diff
@@ -71,59 +71,192 @@
...
+  =*  next  $(entries t.entries)
+  =/  [=whom:ch =unread:unreads:ch]  head
+  =/  =pact:ch
+    ?-  -.whom
+      %ship  pact:(~(gut by dms) p.whom *dm:ch)
+      %club  pact:(~(gut by clubs) p.whom *club:ch)
+    ==
+  =;  events=(list [time incoming-event:a])
+    (weld events next)
+  :-  [*@da %dm-invite whom]
```
Fang-, line 178: We want to use the `recency.unread` here too, like in the channels case, if we didn't add any messages.

Fang-: But apparently even in the channels case it's bugged and doesn't affect the ordering as expected. We should investigate & fix prior to release.

desk/app/activity.hoon
```diff
@@ -190,7 +190,7 @@
       ~(tap by threads.unread)
     |=  [parent=message-key:ch [key=message-key:ch count=@ud]]
     ^-  (unit (list [time incoming-event:a]))
-    =/  writ=(unit writ:ch)  (~(get by wit.pact) time.parent)
+    =/  writ=(unit writ:ch)  (get:on:writs:ch wit.pact time.parent)
```
Fang-, line 193: These are classic, and nasty. Using `+by` on a `mop` will compile and get you results, but potentially the wrong result. Good catch.

desk/lib/channel-utils.hoon
```diff
@@ -29,13 +29,19 @@
...
+      sort   +.sort.channel
+      order  +.order.channel
+      pending  pending.channel
+    ==
+  ?.  full  base
+  %*  .  base
+    posts  (uv-posts posts.channel)
+    net  net.channel
+    remark  remark.channel
   ==
```
Fang-, line 45: This is just `%_  base` (in place of `%*  .  base`). We need it for `*channel:c` because that needs to be evaluated, and we modify the result, but here `base` already is a noun in our subject, so we can just modify that directly.

Fang- (approved):
Changes look good, thank you. A question for confirmation, and a small style nit, but good otherwise.

desk/app/activity.hoon
```diff
@@ -644,9 +644,10 @@
...
   ?~  stream
     [newest total notified ?~(last ~ `[u.last main main-notified]) children]
-  =/  [[* =event:a] rest=stream:a]  (pop:on-event:a stream)
+  =/  [[=time =event:a] rest=stream:a]  (pop:on-event:a stream)
   =/  volume  (get-volume -.event)
   =?  notified  &(notify.volume notified.event)  &
+  =.  newest  time
   ?.  ?&  unreads.volume
           ::TODO  support other event types
           ?=(?(%dm-post %dm-reply %post %reply) -<.event)
```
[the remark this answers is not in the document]
arthyn: right any event should potentially affect the recency although we may want to limit it at some point like you said. mostly did this because `chan-init` and `dm-invite` weren't being taken into account, can address in the follow-up

desk/lib/channel-utils.hoon
```diff
@@ -24,18 +24,24 @@
...
+  =/  base
+    %*  .  *channel:c
+      perm   +.perm.channel
+      view   +.view.channel
+      sort   +.sort.channel
+      order  +.order.channel
+      pending  pending.channel
+    ==
+  ?.  full  base
+  %*  .  base
```
arthyn, line 41:
```suggestion
  %_  base
```

PR #3532, opened 2024-05-22
native: add gallery  and notebook detail view

patosullivan:
Fixes LAND-1805

Also fixes an issue where the user couldn't add an optional message to gallery posts (and fixes an issue where we weren't rendering that message, albeit in a primitive way for now).

Also note that this just uses ChatContent to render the notebook content for now. We probably want to build a separate component for rendering notebook content at some point (for rendering lists, in particular, but probably for other things), but this seems fine for now.

Looks like this:

https://github.com/tloncorp/tlon-apps/assets/1221094/7b0798da-f2ea-496c-91ee-bbdd2538d9d4


apps/tlon-mobile/src/screens/PostScreen.tsx
```diff
@@ -137,6 +137,7 @@ export default function PostScreen(props: PostScreenProps) {
         contacts={contacts ?? null}
         calmSettings={defaultCalmSettings}
         currentUserId={contactId}
+        parentPost={post!}
```
dnbrwstr, line 140: Can we throw an error or something instead of force casting? And/or make it actually optional.

packages/ui/src/components/CommentsScrollerSheet.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  return (
+    <ActionSheet
+      open={open}
+      onOpenChange={setOpen}
+      snapPointsMode="percent"
+      snapPoints={[90]}
+      disableDrag
+    >
+      <View height="100%" paddingBottom={bottom + 24}>
```
dnbrwstr, line 66: Can we use `getTokenValue` here to keep it linked to the theme?

packages/ui/src/components/GalleryPost/GalleryPost.tsx
```diff
@@ -5,28 +5,40 @@ import {
...
 
-import { LinearGradient, View } from '../../core';
+import { LinearGradient, Text, View, YStack } from '../../core';
 import ChatContent from '../ChatMessage/ChatContent';
 import ContentReference from '../ContentReference';
 import Pressable from '../Pressable';
 import GalleryAuthorRow from './GalleryAuthorRow';
 
 const HEIGHT_AND_WIDTH = (Dimensions.get('window').width - 46) / 2;
+const HEIGHT_DETAIL_VIEW = Dimensions.get('window').height - 236;
```
dnbrwstr, line 27: Would be nice to have some comments here to explain what's happening w/ the magic numbers.

packages/ui/src/components/GalleryPost/GalleryPost.tsx
```diff
@@ -53,89 +65,149 @@ export default function GalleryPost({
...
-            </View>
-          </ImageBackground>
-        )}
+        {(postIsJustImage || textPostIsJustLinkedImage) &&
+          (detailView ? (
+            <TouchableOpacity
+              onPress={() =>
+                handleImagePressed(postIsJustImage ? image!.src : linkedImage!)
+              }
+              activeOpacity={0.9}
```
dnbrwstr, line 91: Are we using similar opacity values elsewhere? Might make sense to extract as a constant.

packages/ui/src/components/PostScreenAuthorRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+  parentPost,
+  timeDisplay,
+  setShowComments,
+}: {
+  parentPost: db.Post;
+  timeDisplay: string;
+  setShowComments: (show: boolean) => void;
+}) {
+  return (
+    <Pressable onPress={() => setShowComments(true)}>
```
dnbrwstr, line 19: Just checking -- we've got a few `AuthorRow`s now; anything abstractable?

packages/ui/src/components/PostScreenView.tsx
```diff
@@ -53,15 +63,32 @@ export function PostScreenView({
   negotiationMatch: boolean;
 }) {
   const [inputShouldBlur, setInputShouldBlur] = useState(false);
+  const [showComments, setShowComments] = useState(false);
   const canWrite = utils.useCanWrite(channel, currentUserId);
+  const isChatChannel = channel ? getIsChatChannel(channel) : true;
+
+  console.log('PostScreenView', {
```
dnbrwstr, line 70: Stray log

packages/ui/src/components/PostScreenView.tsx
```diff
@@ -98,21 +140,53 @@ export function PostScreenView({
...
+              {parentPost && (
+                <CommentsScrollerSheet
+                  open={showComments}
+                  setOpen={setShowComments}
                   channelId={channel.id}
+                  currentUserId={currentUserId}
+                  editingPost={editingPost}
+                  setEditingPost={setEditingPost}
+                  editPost={editPost}
+                  posts={posts?.filter((p) => p.id !== parentPost.id) ?? []}
```
dnbrwstr, line 152: Should memoize this.

dnbrwstr (approved):
Looking good, just a bunch of nits, most pretty minor.

packages/ui/src/components/GalleryPost/GalleryPost.tsx
patosullivan: It just looks better at this opacity imo, but we don't use it elsewhere. I can just leave it out for now.

apps/tlon-mobile/src/screens/PostScreen.tsx
patosullivan: We could just return null if we don't have the post, like we do for channel and contactId, will just do that for now I think.

packages/ui/src/components/PostScreenAuthorRow.tsx
patosullivan: They all have some pretty unique layouts (detail view author row vs gallery view vs notebook view [hasn't been extracted to a component yet] vs chat view), so at the moment they seem like they're ok if they're separated imo.

patosullivan: Actually, we could just pass some extra props and make one component.

PR #3535, opened 2024-05-24
dms: broadcast functionality

Fang-:
DM "broadcasting" functionality, initially for internal-only usage. (As such, not developed for or tested on mobile, and gated on manual backend agent enabling.)

Tiny /app/broadcaster agent on the backend keeps track of established "cohorts", the actions taken on them, messages sent to them, and handles the spreading of messages into those cohorts.

Frontend has a special "broadcasts" messages filter, showing the created cohorts and letting you view the aforementioned history.

Implementation-wise, we finesse data in a couple of places to re-use existing components (most notable, pretending the cohort history is normal dm backlog), and most of the code/structure has been cargo-culted from multi-DM flows.  
I've tried to pare it down to just the bare necessities, but during review keep an eye out for things I should or should not have trimmed, or places where a different implementation pattern would be more appropriate. It's been a while since my last big frontend feature, so please be critical.

Remaining work:
- [x] Add cohort creation flow,
- [x] Hide broadcasts filter if backend agent not running,
- [ ] Do smaller post-change refetches,
- [ ] Maybe restructure backend json api to be slightly more normal,
- [x] Test with large amount of ships,
- [x] Fix bugs:
  - [x] Don't crash on-hover of cohort messages (mb render w/ custom component instead of as fake writs),
  - [x] Stay on messages tab when viewing a cohort,
  - [x] Clear dialog input box after adding/removing ships from a cohort,
  - [x] Actually render confirmation dialog prior to deleting a cohort.
  - [ ] Correctly sort cohorts in the sidebar by recency.
- [x] Investigate if we can support mass-deletion easily.
  - [x] ~~If~~ we can, support it.

Closes TLON-1892.


apps/tlon-web/src/components/Sidebar/MessagesSidebar.tsx
```diff
@@ -68,8 +73,17 @@ export default function MessagesSidebar({
...
   ];
+  if (cohorts) filterActions.push({
+    key: 'broadcasts',
+    onClick: () => setFilterMode(filters.broadcasts),
+    content: 'Broadcasts',
+    containerClassName: cn(
+      'flex items-center space-x-2 rounded-none',
+      messagesFilter === filters.broadcasts && 'bg-gray-50 text-gray-800'
+    ),
+  });
```
patosullivan, line 86: This technically works, but it breaks our style conventions. You probably need to run prettier/the linter for all of these changes.

apps/tlon-web/src/dms/BroadcastInviteDialog.tsx
```diff
@@ -0,0 +1,160 @@
...
+                return so.value;
+              })
+          }
+        };
+      }
+      //TODO  refetch just this specific cohort
+      api.poke({
+        mark: 'broadcaster-action', app: 'broadcaster', json,
+        onSuccess: refetchCohorts, onError: refetchCohorts
+      });
```
patosullivan, line 94: Normally all of this should live outside of the component, in this case it would probably go in state/broadcasts.

apps/tlon-web/src/dms/DMOptions.tsx
```diff
@@ -221,7 +221,7 @@ export default function DmOptions({
     actions.push({
       key: 'leave',
       type: 'destructive',
-      onClick: leaveMessage,
+      onClick: () => setDialog(true),
```
patosullivan, line 224: lol, thank you.

apps/tlon-web/src/dms/BroadcastDm.tsx
```diff
@@ -0,0 +1,140 @@
...
+        </span>
+        <span className="line-clamp-1 w-full break-all text-sm text-gray-400">
+          <span>{`${count} ${pluralize('Target', count)}`}</span>
+        </span>
+      </div>
+    </BackButton>
+  );
+}
+
+export default function MultiDm() {
```
patosullivan, line 57:
```suggestion
export default function BroadcastDm() {
```

patosullivan (changes requested):
Overall, looks good. There are a few things I've flagged here. Also need to make sure this all gets formatted properly (`pnpm lint:format` if you don't have prettier in your editor).

apps/tlon-web/src/components/Sidebar/Sidebar.tsx
```diff
@@ -256,6 +269,13 @@ const Sidebar = React.memo(() => {
           <MessagesSidebar searchQuery={searchInput} />
         )}
       </div>
+      <BroadcastInviteDialog
+        mode="add"
+        inviteIsOpen={broadcastIsOpen}
+        setInviteIsOpen={setBroadcastIsOpen}
+        whom=""
```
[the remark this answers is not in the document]
Fang-: Yes, intentional, but I should just make that property optional here.

patosullivan (approved):
This looks good, the only thing I think is maybe missing is the ability to delete a sent broadcast (if feasible). Edit would also be nice, but we don't have edit for DMs yet.

patosullivan (approved):
Looks great, thanks for adding delete!

jamesacklin:
@latter-bolden and @arthyn, can you add your reviews? Since this is feature-flagged out I'm slightly less concerned and would like to close the milestone out.

apps/tlon-web/src/dms/MessagesList.tsx
```diff
@@ -63,9 +69,23 @@ export default function MessagesList({
       : { main: 400, reverse: 400 },
   };
 
+  console.log('filter', filter, filters.broadcasts, broadcasts.data);
```
latter-bolden, line 72:
```suggestion
```

latter-bolden (approved):
JS looks fine to me.

 I think this primarily needs a thumbs up on the hoon code before merging.

arthyn (approved):
lgtm!

jamesacklin:
If we don't have any more blockers, let's merge.

PR #3538, opened 2024-05-25
native: hook up notification tap handler

alecananian:
Hooks up the Android and iOS push notification tap handlers to navigate to the related channel

Notes:
- Added `IGNORE_COSMOS` env var that needs to be set if debugging notification tap handler locally
- I'm not sure how to get from the yarn type's `wer` value to a valid post ID we can use for selected posts or thread navigation


patosullivan (approved):
Nice, thank you!

PR #3541, opened 2024-05-27
native: channel screen refactor

dnbrwstr:
This is an incremental refactor that groups up shared functionality between thread and channel views. Could be cleaned up more but helps start to isolate the reused logic.

### DIFF
--- a/apps/tlon-mobile/src/fixtures/Channel.fixture.tsx
+++ b/apps/tlon-mobile/src/fixtures/Channel.fixture.tsx
@@ -239,15 +239,12 @@ function SwitcherFixture({
 }: {
   switcher: ReturnType<typeof useChannelSwitcher>;
 }) {
-  const { bottom } = useSafeAreaInsets();
-
   return (
     <ChannelSwitcherSheet
       open={switcher.isOpen}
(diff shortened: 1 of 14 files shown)


latter-bolden (changes requested):
Code lgtm. 

But running it locally, channel switching doesn't seem to work properly. Header seems to stay the same and display type doesn't update.

jamesacklin:
I found an issue loading posts "forward" in history; see screencapture. I am testing with two different ships with different unread states to check message history continuity.

- Enter a channel
- Scroll down to latest messages, notice that they are missing w/ no loading indicator
- Tab back to main Groups list, then navigate back into the group, then into the channel
- See latest page of messages

https://github.com/tloncorp/tlon-apps/assets/748181/2d3fb7dc-a00b-448c-ba82-be29b8d5e7c5

latter-bolden (approved):
Looks good! The update fixed my switcher issue.

Reasonably sure the issue @jamesacklin mentioned pre-exists this PR, so I'd say good to merge.

jamesacklin:
Apologies, already exists on `develop`. Logging a separate issue, not blocking merge.

PR #3546, opened 2024-05-29
native: notebook editor

patosullivan:
https://github.com/tloncorp/tlon-apps/assets/1221094/627432fe-4a64-4477-820c-620064fb4678

Adds a new BigInput that we'll use for both rich text input on gallery posts and for the text input on notebook posts. Adds a title input and the ability to attach an image. Adds a toolbar for use on notebook input.

Moves ChatContent to ContentRenderer, and adds the ability to render all notebook content (outside of references to anything other than chat messages).

Updates FloatingActionButton to match designs.

Updates sendPost to account for sending metadata (title, metadata), also cleans up messageSender/sendPost.

I grabbed these icons from a public git repo (https://github.com/tabler/tabler-icons) just because we needed icons for the toolbar, but we can switch them out for whatever we want.


latter-bolden:
Submitting my first test note failed (`poke-as cast fail` on `[%bad-key 'underline']`)

patosullivan:
> Submitting my first test note failed (`poke-as cast fail` on `[%bad-key 'underline']`)

Oh wow, yeah. The backend apparently doesn't support underline 😧

packages/ui/src/components/MessageInput/InputToolbar.tsx
```diff
@@ -0,0 +1,117 @@
...
+    editorState,
+    setToolbarContext,
+    toolbarContext,
+  };
+
+  switch (toolbarContext) {
+    case ToolbarContext.Main:
+    case ToolbarContext.Heading:
+      return (
+        <FlatList
```
dnbrwstr, line 49: This list could use some optimization -- it's rerendering frequently during editing. I'd memoize everything you can, maybe break out the item component. `FlatList` might also be overkill for this use case.

Screenshot 2024-05-30 at 5 12 54 PM

packages/ui/src/components/BigInput.tsx
```diff
@@ -0,0 +1,177 @@
...
+          bigInput
+          showToolbar={channelType === 'notebook'}
+          channelType={channelType}
+          ref={editorRef}
+        />
+      </ScrollView>
+      {channelType === 'notebook' &&
+        editorRef.current &&
+        editorRef.current.editor && (
+          <KeyboardAvoidingView
```
dnbrwstr, line 155: Does this screen wrapper have anything in common with others we're using? I've been looking at trying to simplify how we handle them.

latter-bolden (commented):
Usability wise, I think the only blocker on my end is fixing the list formatting as discussed out of band

Can do a more in depth code review tomorrow if needed

dnbrwstr (commented):
Looking good! First pass down below, still have a bit to go through. The only bit that's blocking for demo imo is the channel layout issue.

- Can the placeholder icons use SVGs so that we can drop Ochre ones in easily when we get ‘em
- Inline code rendering looks strange (see screenshot 1)
- Missing code block toggle (?)
- Could you list out places where you had to invent non-ochre UI? Url input for instance.
- Channel width layout issue is back (see screenshot 2)

Screenshot 1: 
Screenshot 2024-05-30 at 5 23 55 PM

Screenshot 2:
Screenshot 2024-05-30 at 4 59 41 PM

apps/tlon-mobile/src/screens/ChannelScreen.tsx
```diff
@@ -86,15 +86,15 @@ export default function ChannelScreen(props: ChannelScreenProps) {
...
-    async (content: Story, _channelId: string) => {
+    async (content: Story, _channelId: string, metadata?: db.PostMetadata) => {
       if (!channel) {
         throw new Error('Tried to send message before channel loaded');
       }
       store.sendPost({
         channel: channel,
         authorId: currentUserId,
         content,
-        attachment: uploadInfo.uploadedImage,
```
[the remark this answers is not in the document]
patosullivan: Yep, this just happens in the same place as it does for any other post type now (`sendMessage` in MessageInput)

packages/shared/src/store/postActions.ts
```diff
@@ -8,20 +8,12 @@ export async function sendPost({
...
   authorId: string;
   content: urbit.Story;
-  attachment?: api.UploadedFile | null;
   metadata?: db.PostMetadata;
 }) {
-  // replace content with attachment if empty
-  // TODO: what if we have both?
-  if (content.length === 0 && attachment && channel.type === 'gallery') {
-    content = [createVerseFromAttachment(attachment)];
-  }
```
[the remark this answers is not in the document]
patosullivan: See other comment.

packages/ui/src/components/MessageInput/InputToolbar.tsx
patosullivan: Memoized a bunch of stuff and added some perf related changes for FlatList, got it down to between 5 and 8ms per character. Unfortunately we do need to update it on each change from the editor in order to show the current enabled/disabled state for the buttons.

packages/ui/src/components/BigInput.tsx
patosullivan: I replaced the input for gallery rich text with this, not sure where else we would use it other than that though.

packages/ui/src/components/MessageInput/InputToolbar.tsx
```diff
@@ -0,0 +1,177 @@
...
+            style={[
+              editor.theme.toolbar.toolbarBody,
+              hideToolbar ? editor.theme.toolbar.hidden : undefined,
+            ]}
+            renderItem={renderItem}
+            horizontal
+            removeClippedSubviews
+            initialNumToRender={6}
+            maxToRenderPerBatch={6}
+            getItemLayout={(_, index) => ({
```
dnbrwstr, line 138: I'd usually say memoize this, but if the list never gets long it doesn't really matter.

dnbrwstr (approved):
Lgtm!

PR #3547, opened 2024-05-29
native: fix isNotice variant of ChatMessage, ChannelDivider

jamesacklin:
Adjusts the ChannelDivider and the isNotice variant of ChatMessage to be full-width with a pill in the center. Also adds fixtures for both.

image
image
Screenshot 2024-05-29 at 11 41 55 AM


latter-bolden (approved):
Nice!

PR #3548, opened 2024-05-29
native: ChatListScreen renovations

jamesacklin:
- Removes the top border on the bottom tab bar and the bottom rule on the top nav
- Switches the bottom navbar to use a BlurView
- Replaces "Tlon" in the header with a rotated t-block icon
- Adjusts section header padding to spec
- Conditionally renders the section header so that "All" is hidden if we have no pinned items

image


apps/tlon-mobile/src/screens/ChatListScreen.tsx
```diff
@@ -81,7 +81,7 @@ export default function ChatListScreen(
     <ContactsProvider contacts={contacts ?? []}>
       <View backgroundColor="$background" flex={1}>
         <ScreenHeader
-          title={<Icon type="TBlock" size="$xl" flex={1} rotate="-6deg" />}
+          title={<Icon type="TBlock" size="$m" flex={1} rotate="-6deg" />}
```
urcades, line 84: Given the original image I was just about to comment that the block was a little large, this feels like it'll do the trick as long as it's close to designs / proportional to the + icon

urcades (approved):
stamp!

apps/tlon-mobile/src/screens/ChatListScreen.tsx
jamesacklin: Yes, amended. Thank you

jamesacklin:
@dnbrwstr Any notes? Seems uncontroversial but would appreciate a look-see

packages/ui/src/components/NavBar/NavBar.tsx
```diff
@@ -11,17 +13,22 @@ const NavBar = React.memo(function NavBar(props: {
...
       position="absolute"
       width={'100%'}
       bottom={0}
-      backgroundColor={'$background'}
-      borderTopWidth={1}
       paddingTop={'$m'}
       borderTopColor={'$border'}
       height={navHeight + bottom}
     >
+      <BlurView
```
latter-bolden, line 27: This will likely cause issues on Android. Other places where we're using `BlurView` have to be special cased for iOS only.

jamesacklin: Thank you for resolving!

PR #3551, opened 2024-05-29
native: center-aligned FloatingActionButton for galleries/notebooks

jamesacklin:
Modifies the FloatingActionButton slightly:
- Uses a Button component in the FAB
- Uses the Icon component when passing in an icon to actually render the icon the way Button wants it
- Absolutely positions the button in the channel at the bottom/center to conform with latest designs
- Adds a secondaryBackground to pure-text Gallery posts in the channel to help with visual structure (@urcades, feel free to weigh in)

@patosullivan The FAB was also showing up post-upload in the gallery image upload view, maybe the logic there was placeholder? Double-check that for me please. This doesn't show the FAB in the confirm step.

Screenshot 2024-05-29 at 3 55 00 PM

### DIFF
--- a/packages/ui/src/components/Channel/index.tsx
+++ b/packages/ui/src/components/Channel/index.tsx
@@ -11,7 +11,6 @@ import { useCallback, useMemo, useState } from 'react';
 import { KeyboardAvoidingView, Platform } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
-import { Add, ArrowUp } from '../../assets/icons';
 import {
   CalmProvider,
   CalmState,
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Channel/index.tsx
```diff
@@ -285,28 +285,27 @@ export function Channel({
...
-                                  )
-                                }
-                              />
-                            )}
+                          <View
+                            position="absolute"
+                            bottom={bottom}
+                            flex={1}
+                            width="100%"
+                            alignItems="center"
```
dnbrwstr, line 293: I think this is fine, but would you mind verifying that the container won't block touches on elements underneath?

dnbrwstr (approved):
Lgtm!

packages/ui/src/components/Channel/index.tsx
jamesacklin: Can do, I'll follow-up if so

PR #3552, opened 2024-05-29
native: refactor + optimize creation sheets

dnbrwstr:
This PR optimizes the AddGroup/StartDm sheets to reduce jank during initial load.

- Modifies `ContactBook` to render using a single `SectionList`, giving us more control of how items are rendered.
- Adds `Sheet.LazyFrame`, a drop-in alternative to `Sheet.Frame` which only renders its content after the sheet has been opened at least once.
- Extracts `BlockSectionList` (a section list where each section is contained in a visual block) from `ContactBook`
- Contains a tweak to the count of loaded posts.


jamesacklin (approved):
Great. SectionList looks good, let's commit to using it elsewhere.

PR #3553, opened 2024-05-30
native: FAB positioning

jamesacklin:
Follow-up PR to #3551 where the view was blocking touch-events beneath it. Applies a temporary visual trick until we move to a full-on ChannelFooter component. Apologies for the oversight.


dnbrwstr (commented):
Code seems fine; I'm slightly worried abt the android case as there are some differences in overflow handling behavior -- @latter-bolden or @patosullivan, would you verify that it looks right when you have a moment?

latter-bolden (approved):
Confirmed it displays correctly and is clickable on Android/iOS

Not needed rn, but we should avoid having any safe area bottom padding for the underlying channel content in Galleries/Notebooks

PR #3554, opened 2024-05-30
chat: include $whom in %writ-response mark

Fang-:
This lets us specify the chat that the writ response happened in/for, which cannot otherwise be known if you're subscribing to the / "firehose" endpoint as opposed to an individual, specific chat.

We expect to get away with just tacking it on, because nothing on our backend listens for this mark, it's not used in scry endpoints, and the json is "backwards compatible" in the sense that it just has an additional field on the top-level object now.

Closes TLON-1880 again.


arthyn (approved):
seems good

latter-bolden (approved):
Looking good, confirmed they're showing up when subscribed

PR #3556, opened 2024-05-30
native: don't pass empty posts to scroller, fixes gallery grid issue

patosullivan:
Fixes TLON-1945. We were seeing some posts appear as if they were in a "single column" per row in ed's product group because the posts actually had no content, so the GalleryPost component just returned null. We need to filter out posts like this before we pass them into FlatList.

I also added a "Unsupported content type" component we can return if we run into this again, so we'll know what's going on (it shouldn't happen again, but just in case).


dnbrwstr (approved):
Lgtm!

PR #3557, opened 2024-05-31
native: identify members of Tlon group in PostHog

jamesacklin:
Identifies Tlon employees as members of the Tlon group, then flags their anonymous identifier in PostHog. This lets us filter internal usage out in our analytics for TestFlight/production builds going forward.

- Adds a TLON_GROUP environment variable in Expo/.env
- Checks all resolved chats to see if they contain the Tlon group
- Sends a isTlon: true flag to PostHog

Fixes TLON-1955


dnbrwstr (approved):
Lgtm! Only stray thought I had was might be nice to name like `TLON_EMPLOYEE_GROUP`, `identifyTlonEmployee`, `isTlonEmployee`, just to make it clear what's happening if anyone is poring over the analytics source code.

jamesacklin:
No problem, chef - thank you

PR #3558, opened 2024-05-31
native: android fixes, logout, & view build number

latter-bolden:
Bit of a hodgepodge. 

Biggest piece is fixing a few Android bugs. Namely the new create/join sheets crashing due to an issue with Tamagui Sheet's ["modal" property](https://github.com/gorhom/react-native-portal/issues/34) and one where the empty channel message was showing up rotated.

Also adds the ability to log out and display hidden dev info by clicking the _Profile_  page avatar 5 times rapidly

Fixes TLON-1935
Fixes TLON-2042

### DIFF
--- a/apps/tlon-mobile/src/components/AddGroupSheet.tsx
+++ b/apps/tlon-mobile/src/components/AddGroupSheet.tsx
@@ -7,10 +7,12 @@ import {
   NativeStackScreenProps,
   createNativeStackNavigator,
 } from '@react-navigation/native-stack';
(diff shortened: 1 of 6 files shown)


jamesacklin:
Still saving a bit of authentication state, moving to DNM.

latter-bolden:
Updated this to fix the logout not sticking & removed the dev debug popup since that's pulled in/updated on the activity branch.

This was already reviewed, so think I'm just gonna merge it.

PR #3563, opened 2024-06-03
native: sync performance improvements

dnbrwstr:
Mostly the sync work from the demo branch, with a few tweaks.

- Disables re-rendering of chat list query when offscreen
- Disabled refetch of init data every time ChatList is focused
- Re-enable syncing stale channels
- Cache auth cookie in AsyncStorage
- Add devtools for `react-navigation`, `react-query`, and `AsyncStorage` (hit shift+M from the bundler terminal window to open)
- Tweak load order + grouping


apps/tlon-mobile/src/hooks/useFocusNotifyOnChangeProps.ts
```diff
@@ -0,0 +1,31 @@
...
+    React.useCallback(() => {
+      focusedRef.current = true;
+
+      return () => {
+        focusedRef.current = false;
+      };
+    }, [])
+  );
+
+  return () => {
```
latter-bolden, line 20: These default return values are confusing to me. Some comments about what's happening there or more broadly, how this hook is supposed to be used would be helpful

packages/shared/src/db/queries.ts
```diff
@@ -1573,52 +1568,49 @@ async function updatePostWindows(
...
-          eq($postWindows.channelId, window.channelId),
-          lte($postWindows.oldestPostId, newer ?? window.newestPostId),
-          gte($postWindows.newestPostId, older ?? window.oldestPostId)
-        )
-      )
+      .where(overlapsWindow(referenceWindow))
   )[0];
 
+  // Delete intersecting windows.
+  await tx.delete($postWindows).where(overlapsWindow(referenceWindow));
```
latter-bolden, line 1589: Nice, reading through this function is a lot easier now

packages/shared/src/store/sync.ts
```diff
@@ -598,11 +596,24 @@ export const clearSyncQueue = () => {
...
+    ]);
+    return Promise.all([
+      api.subscribeToChannelsUpdates(handleChannelsUpdate),
+      api.subscribeToChatUpdates(handleChatUpdate),
+      api.subscribeToContactUpdates(handleContactUpdate),
+    ]);
+  });
 };
+
+export const start = async () => {};
```
latter-bolden, line 619: if this is no op, should we remove?

packages/shared/src/store/sync.ts
```diff
@@ -598,11 +596,24 @@ export const clearSyncQueue = () => {
...
-  useStorage.getState().start();
+export const initializeStorage = () => {
+  return syncQueue.add('initialize storage', async () => {
+    return useStorage.getState().start();
+  });
+};
+
+export const setupSubscriptions = async () => {
+  return syncQueue.add('setup subscriptions', async () => {
+    await Promise.all([
```
latter-bolden, line 607: maybe a comment as to why you're making this load order explicit

apps/tlon-mobile/src/screens/ChannelScreen.tsx
```diff
@@ -81,7 +81,7 @@ export default function ChannelScreen(props: ChannelScreenProps) {
         }
       : {
           mode: 'newest',
-          firstPageCount: 10,
+          firstPageCount: 50,
```
latter-bolden, line 84: maybe bump around as well? Unless you found it noticeably impacted response time. Iirc post count was less important on average than you'd think

latter-bolden (approved):
Code looks good, few small things but none blocking.

apps/tlon-mobile/src/screens/ChannelScreen.tsx
dnbrwstr: With `around`, it's more about layout than anything else. In order to correctly position the anchored post on screen, we need the anchored post plus a few posts around it to have been rendered. The low initial count allows use to render exactly what we want on first render + scroll to it right away.

apps/tlon-mobile/ios/Landscape/UrbitModule.swift
```diff
@@ -78,7 +78,10 @@ class UrbitModule: NSObject {
         UrbitModule.authCookie = authCookie
         Task {
             try? await UrbitAPI.shared.open(for: shipUrl)
-            try? await PocketUserAPI.fetchSettings()
+            // We don't appear to need these settings anymore, so I've disabled
+            // this in order to save a network request.
+            //
+            // try? await PocketUserAPI.fetchSettings()
```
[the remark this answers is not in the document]
dnbrwstr: Ah that makes sense, will remove for now.

packages/shared/src/store/sync.ts
dnbrwstr: Going to remove as I don't remember why and it doesn't seem to have much of an effect.

packages/shared/src/store/sync.ts
dnbrwstr: I've removed for now, and reworked init a little bit. We should really colocate all the start stuff, possibly in `useInitialSync`.

PR #3564, opened 2024-06-03
native: render interleaved content for notes

patosullivan:
Fixes TLON-1958 by making sure we always interleave rendered content for notes (as opposed to chat where blocks should always appear before the text/inline content, or galleries where the content is either text or an image).

Simulator Screenshot - iPhone 15 Pro - 2024-06-03 at 14 08 51


latter-bolden (approved):
Makes sense!

PR #3566, opened 2024-06-04
native: query refactor

dnbrwstr:
This PR refactors how queries are called in order to allow us to batch query effects. It adds a `ctx` param to all queries that includes a reference to the database client, as well a reference to the set of pending effects. For callers, the `ctx` param is optional -- a default one will be created if one isn't passed in.


packages/shared/src/store/syncQueue.ts
```diff
@@ -12,7 +12,7 @@ interface SyncOperation {
 export class QueueClearedError extends Error {}
 
 class SyncQueue {
-  concurrency = 2;
```
patosullivan, line 15: Does this actually work? I wouldn't expect it to. On a fresh install I have 131 threads, they just synchronize one-by-one.

patosullivan (approved):
Code itself looks fine, but I'm wondering if the `concurrency` property (that existed prior to this PR) actually makes sense since urbit can't respond to requests concurrently

PR #3570, opened 2024-06-05
activity: when migrating, refresh summaries *once*

Fang-:
Migration logic was calling into `+update-index` for each entry we imported, which in turn calls `+refresh-summary`. Of course, we know we'll likely be adding more things, so this is unnecessary (and slow) work.

Here, we add an `importing` flag into the context, and set it to true for the duration of migration logic. The `should-notify` hidden parameter we had added to the `+add` arm previously gets replaced by this, and if it's set to true, we don't run `+refresh-summary` for every invocation of `+update-index`.

This brings upgrade wall-clock time down from "idk, didn't finish in a reasonable timeframe" to a minute and a half, of which a full minute is spent on the activity agent's migration logic.

We could probably bring that down further, or at least reduce variance, by capping the amount of events we add per channel. Unclear whether that's desirable or not. The change here feels like 80% of the solution, at least.

Fixes TLON-1986

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -17,6 +17,11 @@
 =|  state-1
 =*  state  -
 ::
+::NOTE  setting this to true causes some parts of state & update management to
+::      take shortcuts, which we want to do during initial migration/import.
+::      shouldn't be set to true outside of calls to +migrate.
+=/  importing=?  |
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

PR #3571, opened 2024-06-05
ci: disable e2e tests (for now)

Fang-:
As discussed out of band, these "seemingly just started failing due to timeouts at some point". We ignored this and got used to the red X's.

It seems preferable to disable the tests until we have time to re-enable them and investigate their failures properly, rather than get used to CI failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,21 +28,22 @@ jobs:
       - name: Run Tests
         run: pnpm -r test run
 
-      - name: Install Playwright Browsers
-        run: pnpm exec playwright install --with-deps
-        working-directory: ./apps/tlon-web
+      #NOTE  e2e tests disabled temporarily, please investigate why they fail
 
-      #NOTE  disabled temporarily, please investigate why they fail
+      # - name: Install Playwright Browsers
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -32,9 +32,10 @@ jobs:
         run: pnpm exec playwright install --with-deps
```
patosullivan, line 32: We ought to remove this step

.github/workflows/ci.yml
```diff
@@ -32,9 +32,10 @@ jobs:
...
 
-      - name: Run E2E Tests for Groups
-        run: pnpm e2e
-        working-directory: ./apps/tlon-web
+      #NOTE  disabled temporarily, please investigate why they fail
+      # - name: Run E2E Tests for Groups
+      #   run: pnpm e2e
+      #   working-directory: ./apps/tlon-web
 
       - uses: actions/upload-artifact@v3
```
patosullivan, line 40: And this one

patosullivan (changes requested):

patosullivan (approved):

PR #3574, opened 2024-06-06
activity: refactor children tracking

arthyn:
Instead of a list, The backend now sends down a map of child activity summaries, but only a single level deep. This allows us to sum the child counts easier.

In addition, we now track combined "status" more actively by updating parents when childrens' statuses change. Now, the nav bar dots will immediately clear upon being seen.

Also takes into account threads for DM dots and sidebar filtering so that it is more accurately reflected.

Fixes TLON-1983

PR Checklist


patosullivan (commented):
Code looks good. I am noticing that I don't see gray dots on groups with channels that also have gray dots, should I?


https://github.com/tloncorp/tlon-apps/assets/1221094/936c9375-f618-4152-9e44-d5bb19c53ac9

arthyn:
@patosullivan if it's a ship that has already run %activity, you might need to run `:activity &noun %migrate`

desk/app/activity.hoon
```diff
@@ -429,6 +447,7 @@
   |-
   ?~  sources  cor
   =.  cor  (refresh-summary i.sources)
+  =.  cor  (refresh-parent-summaries i.sources)
```
Fang-, line 450: Oh, do we expect the parent sources to not always exist in `indices`? Is that because of the way in which we do "less work" during migration logic?

Fang- (commented):
We discussed this yesterday, but I have some additional questions.

desk/app/channels.hoon
```diff
@@ -2132,9 +2132,20 @@
...
+    ?.  (can-read:ca-perms our.bowl)
+      (send:ca-activity [%adjust source ~] ~)
+    =+  .^(=volume-settings:activity %gx (scry-path %activity /volume-settings/noun))
+    =.  ca-core
+      ::  if we don't have a setting, no-op
+      ?~  setting=(~(get by volume-settings) source)  ca-core
+      ::  if they have a setting that's not mute, retain it otherwise
+      ::  delete setting if it's mute so it defaults
+      ?.  =(setting mute:activity)  ca-core
+      (send:ca-activity ~[[%adjust source ~]])
```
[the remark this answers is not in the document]
arthyn: Because this will run on any permissions changes not just ones where we've regained/lost permissions, so we don't want to clear any settings they might have if the permissions didn't change.

desk/app/activity.hoon
arthyn: It's not that we don't expect them to exist, but that if a child source changes, we need to update the parent because we do some aggregation that should be affected by children.

Fang-: Hmm, makes me wonder then if `+refresh-summary` should recursively call itself for all the affected parent sources.

But, this should accomplish the same, just with superfluous passes, I think.

arthyn: yeah it probably should

arthyn: @Fang- am I good to merge as is though?

Fang- (approved):
Latest change looks good. I think this should be good to go as is!

PR #3575, opened 2024-06-06
native: activity feed (middle tab)

latter-bolden:
WIP

Fixes TLON-1505


latter-bolden:
I'm going to unify both activity issues into a single PR since they'll need to be left open and continually rebased while we wait for activity agent to hit internal.

PR #3576, opened 2024-06-06
native: sync fixes

dnbrwstr:
- Fixes TLON-1900 (optimistic updates)
- Fixes TLON-1972 ("last post missing" weirdness)
- Fixes TLON-1985 (Chat list's initial flickering)

Might fix TLON-1939


packages/shared/src/db/queries.ts
```diff
@@ -1013,19 +1013,13 @@ export const insertChannels = createWriteQuery(
       return;
     }
 
-    logger.log(
```
dnbrwstr, line 1016: Why

packages/shared/src/db/queries.ts
```diff
@@ -615,8 +615,8 @@ export const addChatMembers = createWriteQuery(
     },
     ctx: QueryCtx
   ) => {
-    return ctx.db.transaction(async (tx) => {
-      await tx
+    return withTransactionCtx(ctx, () => {
+      return ctx.db
```
dnbrwstr, line 619: txCtx

packages/shared/src/db/queries.ts
```diff
@@ -1476,15 +1470,6 @@ export const insertStandalonePosts = createWriteQuery(
   async (posts: Post[], ctx: QueryCtx) => {
     withTransactionCtx(ctx, async (txCtx) => {
       await insertPosts(posts, txCtx);
-      for (const post of posts) {
```
dnbrwstr, line 1479: Also rename insertStandalone to insertLatest

packages/shared/src/store/sync.ts
```diff
@@ -518,36 +522,34 @@ export async function syncChannelWithBackoff({
 }: {
   channelId: string;
 }): Promise<boolean> {
+  async function isStillPending() {
+    return (
+      currentPendingMessageSyncs.has(channelId) &&
```
dnbrwstr, line 527: Nop

latter-bolden (approved):
We walked through this together, feedback captured in Dan's comments. Good to merge once those are resolved

PR #3577, opened 2024-06-06
channels: respect the (lack of) /full on scry path

Fang-:
We were testing for null against full (from the hoon.hoon stdlib) rather than full.pole, causing it to always give full results. Of course, in practice, you don't want this in most cases.


patosullivan:
Tested on ~finned-palmer, works!

patosullivan:
Somehow the types are failing, though

patosullivan:
looks like the type issue was a bad merge to develop, ignore

arthyn (approved):

PR #3579, opened 2024-06-06
native: channel switchers

dnbrwstr:
This PR implements two new versions of the channel switcher:
- A default version that's close to the existing mobile app: when you tap a group, you see a list of its channels before entering.
- The next version that James + Ed have been working on, which can be enabled in `Profile > App Settings`

It also adds a really rudimentary feature flag system.
- Module in `tlon-mobile/lib/featureFlags` has basic get/set methods
- Screens read config from module and pass it down into UI components
- There's a new screen for toggling flags, accessible from profile provides editing interface. I think it's fine to keep this public for now.

If we continue to use this feature flag system, it might be nice to just the state/accessors in `context` so we don't need to drill props down, but it should be fine for now.


patosullivan (changes requested):
Code looks good afaict, but if I select a pinned channel I still get taken to the channel list for the group rather than the pinned channel.

jamesacklin:
Instrumented feature flags are _sick!_ Thank you!

patosullivan (approved):
Looks good, works well locally!

jamesacklin:
Fixes TLON-1982

PR #3580, opened 2024-06-07
native: fix chat list order and more

dnbrwstr:
This pr contains a few small fixes/improvements.
- Re-enables the old unreads api for now, which fixes chat list order, unread counts, and a number of other things.
- Refactors sync methods so that we're attempting to load the next request while the last one is persisting to the database.
- Adds a background tint to text icons that don't have a background color set (minor visual thing that's been bugging me).

    
Before:

After:


patosullivan (approved):
Code looks good, works well locally.

PR #3582, opened 2024-06-07
native: show user sigil in bottom nav while contacts is still loading in

patosullivan:
Fixes TLON-1967 by showing the default, unstyled sigil in the bottom nav until we get the contact details for the logged-in user (rather than just not showing the button at all).


jamesacklin (approved):
Gotcha-gotcha, thank you!

PR #3584, opened 2024-06-07
sidebar: unclickable pins and unable to pin chats

arthyn:
This fixes TLON-2001. Pinning/unpinning channels was broken because we were using wrong identifier. Creating message keys every render was causing the `markRead` function to change every render which was causing us to infinitely loop, which made pinned items unclickable.

PR Checklist


patosullivan (approved):

jamesacklin:
Checks out, thanks

PR #3585, opened 2024-06-10
native: fix issues with reactions in DMs

patosullivan:
Fixes LAND-1997. We were using %channels pokes where we should have been using %chat.


latter-bolden (approved):
Looks good, makes sense

PR #3586, opened 2024-06-10
web: fix leap crashing issue

patosullivan:
We were using entry.string, which couldn't split out to [ship, nickname] appropriately anymore. `whom` passed to useCheckDmUnread was always `${ship}{nickname}`, which we'll never find an unread for. We can just pull the ship and the nickname out of the entry.original property instead. I also updated useCheckDmUnread() so that unreads[whom] is optional.

Fixes TLON-2029


arthyn (approved):
lgtm

PR #3587, opened 2024-06-10
native: blur message input when adding an attachment

patosullivan:
Fixes TLON-2032 by blurring the message input when we pop the attachment sheet open.


latter-bolden (approved):
Sweet, looks good

PR #3591, opened 2024-06-10
native: wire up chat sub

latter-bolden:
Ports over my chat sub work from the main activity PR. Hopefully helps solve some of the weirdness we're seeing in DMs more broadly.

Also fixes the problem where the parent post is missing and the input is all the way at the top when starting a new thread.

Fixes TLON-1926


patosullivan (approved):
Looks good, works as expected!

PR #3592, opened 2024-06-10
native: fix keyboard focus issues by relying on the OS

patosullivan:
Fixes TLON-1995
Fixes TLON-1996

This fixes the keyboard focus jankiness we've been seeing by relying on the native OS rather than react state to dismiss the keyboard. On iOS, if the user interacts with the scroller the keyboard will automatically be dismissed. On Android, it won't, but Android has the hide keyboard button built-into the keyboard itself, and this is a normal interaction on Android.


latter-bolden (approved):
Confirmed this works locally for iOS

WOW does that make a difference in perceived usability, great fix

PR #3596, opened 2024-06-11
native: correctly handle channel join updates

patosullivan:
Fixes TLON-2064 by updating the logic around channel join updates on the channel subscription.

Also fixes an issue with initial sync if the ship doesn't know about any channels with writer roles (don't attempt to update unless we actually have something to insert, otherwise we'll fail because we're trying to insert nothing).

Attempted to handle channel leaves, but it looks like there might be a backend issue with leaving channels right now. Chatting with hunter about it.

### DIFF
--- a/packages/shared/src/db/queries.ts
+++ b/packages/shared/src/db/queries.ts
@@ -458,6 +458,8 @@ export const deleteGroup = createWriteQuery(
 export const insertUnjoinedGroups = createWriteQuery(
   'insertUnjoinedGroups',
   async (groups: Group[], ctx: QueryCtx) => {
+    logger.log('insertUnjoinedGroups', groups.length);
+
     return withTransactionCtx(ctx, async (txCtx) => {
       // ensure we never delete metadata if we get a partial for some reason
       // during the join process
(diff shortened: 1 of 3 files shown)


packages/shared/src/db/queries.ts
```diff
@@ -1205,6 +1208,39 @@ export const setJoinedGroupChannels = createWriteQuery(
   ['channels']
 );
 
+export const addJoinedGroupChannel = createWriteQuery(
```
dnbrwstr, line 1211: Could this be combined with `setJoinedGroupChannels`?

packages/shared/src/db/queries.ts
```diff
@@ -1205,6 +1208,39 @@ export const setJoinedGroupChannels = createWriteQuery(
...
+        currentUserIsMember: true,
+      })
+      .where(eq($channels.id, channelId));
+  },
+  ['channels']
+);
+
+export const removeJoinedGroupChannel = createWriteQuery(
+  'removeJoinedGroupChannel',
+  async ({ channelId }: { channelId: string }, ctx: QueryCtx) => {
```
dnbrwstr, line 1233: Could this be combined with `setLeftGroupChannels`?

packages/shared/src/store/sync.ts
```diff
@@ -17,14 +17,23 @@ const logger = createDevLogger('sync', false);
 
 export const syncInitData = async () => {
   const initData = await syncQueue.add('init', () => api.getInitData());
+  const writers = initData.channelPerms.flatMap((chanInit) =>
```
dnbrwstr, line 20: There are a number of places where we check array length inside query write methods -- could this go there too, for consistency?

dnbrwstr (approved):
Looks good, just nits.

packages/shared/src/db/queries.ts
patosullivan: This handles the specific case where we've just gotten an update about a particular channel being added, so we add to the 'default' nav section before running the update for the single channel.

I looked at using setJoinedGroupChannels, but the logic of the query doesn't make a lot of sense to me for this use case. We update all channels and set currentUserIsMember on each one based on whether the channel ID for each row is in the list of IDs we've just passed to the query. This makes sense for the use case of updating channel membership based on unreads, but for a one off channel add this will mark every other channel (other than the new channel) the user has as unjoined.

We could make setJoinedGroupChannels more complex, but I feel like two simple queries here is better than one complex one.

packages/shared/src/db/queries.ts
patosullivan: Same comments apply as in the setJoinedGroupChannels case.

packages/shared/src/store/sync.ts
patosullivan: Yeah. It looked like I couldn't early return within `insertChannelPerms` because the type def wouldn't accept a return type of undefined, but after marking the queryFn as async typescript doesn't seem to mind (kinda weird), pushing fix.

PR #3599, opened 2024-06-11
native: better handle linked images in gallery posts

patosullivan:
It looks like we could have posts with inline links where the total inlines could be one rather than two, we need to account for both.

Fixes TLON-2048


jamesacklin:
This should be good. I'm not sure I understand the situation where we create those, but I understand the fix.

dnbrwstr (approved):
Cool, lgtm.

PR #3601, opened 2024-06-11
native: fix missing author row on DMs and multiDMS

patosullivan:
Fixes TLON-2043. We were just taking the first part of a split string to determine post type on an optimistically created message, which won't work for dms or multiDms

### DIFF
--- a/packages/shared/src/db/modelBuilders.ts
+++ b/packages/shared/src/db/modelBuilders.ts
@@ -23,6 +23,12 @@ export function buildPendingPost({
   const sentAt = Date.now();
   const id = getCanonicalPostId(unixToDa(sentAt).toString());
   const [postContent, postFlags] = api.toPostContent(content);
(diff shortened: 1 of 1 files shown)


jamesacklin:
Interesting—stands to reason why we only saw it in DMs.

packages/shared/src/db/modelBuilders.ts
```diff
@@ -23,6 +23,12 @@ export function buildPendingPost({
   const sentAt = Date.now();
   const id = getCanonicalPostId(unixToDa(sentAt).toString());
   const [postContent, postFlags] = api.toPostContent(content);
+  const isDm = isDmChannelId(channel.id) || isGroupDmChannelId(channel.id);
+  const type = parentId
```
dnbrwstr, line 27: Do we not have a helper for this somewhere?

dnbrwstr (approved):
Looks fine, but would be good to use a helper if possible -- really want all that string splitting stuff encapsulated/done in the fewest places possible.

PR #3602, opened 2024-06-11
native: remove cursor from useChannelPosts query key

patosullivan:
Fixes TLON-2046 by removing the cursor param from the query key in useChannelPosts. Having the cursor in there forced the query to refetch every time we got a new message, so the user would briefly see the channel reload (since the data would be reset each time).


dnbrwstr (changes requested):
It's included there to fix a potential hang when attempting to load the channel at a specified cursor (which isn't always available at the jump). I'm also tweaking this stuff now, so I think let's hold off for a moment.

dnbrwstr:
Closing this since it's addressed in #3607

PR #3604, opened 2024-06-11
activity: feed endpoints

arthyn:
This adds two endpoints specifically for the activity tab with pagination.

`/activity/feed/init/{count}.json` gives latest items limited by count for each tab

`/activity/feed/{type}/{count}/{start}?.json`
- count: `@ud` representing number of items you wish to receive
- start: `@ud` or `@da` optional parameter, will only get events before this time
- type: `all | mentions | replies`
  - `all`: any post, reply, dm, or dm reply that I was notified for
  - `mentions`: any message that mentions me
  - `threads`: any replies for which I was notified for

Also updates the `add` update to include source and the stringified form of source.

Fixes TLON-2036

PR Checklist

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -313,6 +313,15 @@
     =-  ``activity-stream+!>((gas:on-event:a *stream:a -))
     (tab:on-event:a stream:base `(slav %da start.pole) (slav %ud count.pole))
   ::
+      [%x %feed %init count=@ ~]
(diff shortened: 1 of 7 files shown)


Fang- (commented):
As discussed out of band. Mostly just style nits. But looks like it does the right thing.

arthyn:
@Fang- since you're afk packing and I just implemented what you commented on going to take that as soft approval

PR #3605, opened 2024-06-11
native: remove loading spinner from channel view

patosullivan:
Fixes TLON-1913, which asks that we remove all loading spinners from the channel view.


dnbrwstr (approved):
👍

PR #3606, opened 2024-06-12
native: reduce space within/between chat messages, fix keyboard offset in thread view

patosullivan:
Fixes TLON-2055
Fixes TLON-2073

### DIFF
--- a/packages/ui/src/components/MessageInput/index.tsx
+++ b/packages/ui/src/components/MessageInput/index.tsx
@@ -3,6 +3,8 @@ import { useState } from 'react';
 import { TextArea } from '../../core';
 import { MessageInputContainer, MessageInputProps } from './MessageInputBase';
 
+export const DEFAULT_MESSAGE_INPUT_HEIGHT = 44;
+
 export function MessageInput({
   shouldBlur,
   setShouldBlur,
--- a/packages/ui/src/components/MessageInput/index.tsx
+++ b/packages/ui/src/components/MessageInput/index.tsx
@@ -3,8 +3,6 @@ import { useState } from 'react';
 import { TextArea } from '../../core';
 import { MessageInputContainer, MessageInputProps } from './MessageInputBase';
 
-export const DEFAULT_MESSAGE_INPUT_HEIGHT = 44;
-
 export function MessageInput({
   shouldBlur,
   setShouldBlur,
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/MessageInput/index.tsx
```diff
@@ -3,6 +3,8 @@ import { useState } from 'react';
 import { TextArea } from '../../core';
 import { MessageInputContainer, MessageInputProps } from './MessageInputBase';
 
+export const DEFAULT_MESSAGE_INPUT_HEIGHT = 44;
```
dnbrwstr, line 6: Can we actually apply this to the input to enforce it matching?

packages/ui/src/components/PostScreenView.tsx
```diff
@@ -96,7 +96,9 @@ export function PostScreenView({
               <KeyboardAvoidingView
                 //TODO: Standardize this component, account for tab bar in a better way
                 behavior={Platform.OS === 'ios' ? 'padding' : 'height'}
-                // keyboardVerticalOffset={70}
+                keyboardVerticalOffset={
+                  Platform.OS === 'ios' ? -bottom : DEFAULT_MESSAGE_INPUT_HEIGHT
```
dnbrwstr, line 100: Does it make sense to extract this wrapper component and use across multiple screens? I'm guessing we have similar in other places?

dnbrwstr (approved):

packages/ui/src/components/PostScreenView.tsx
patosullivan: Looked for that, afaict this is the only place with this exact offset.

packages/ui/src/components/MessageInput/index.tsx
patosullivan: We do, in /MessageInput/index.native.tsx

dnbrwstr: Ahh I see -- can index.native import the constant from this file then? Nbd if not though.

jamesacklin (approved):
Much denser. We'll refine more later but looks good to me!

packages/ui/src/components/MessageInput/index.tsx
patosullivan: Now using the single definition.

patosullivan: Causes a require cycle, reverting for now.

dnbrwstr: Dang, thanks for giving it a try.

patosullivan: I'm fixing the keyboardavoidingview issue and it'll obviate the need for this anyway

PR #3607, opened 2024-06-12
native: small sync fixes

dnbrwstr:
- Prevents replies from showing up in main scroll
- Fixes situation where conversations could become stuck when loading older posts
- Fixes issue where replies confused our channel cursor tracking, which could cause new posts to fail to appear
- Prevents channel cursor from changing after initially set, preventing some extraneous refreshes

Fixes TLON-2050


latter-bolden (approved):
Lgtm!

PR #3608, opened 2024-06-12
unreads: make all sources fully reactive to children

arthyn:
Previously only DMs and groups were fully reactive to their children, meaning that we took into account the client-side unread status of each child as a sum, so if a child was seen and it was the last source of unreads, the dot would go away. However, we weren't accounting for notify in the channel case which we always want to badge with a blue dot, but when I fixed that in tloncorp/tlon-apps#3589 this made blue dots stickier because channels weren't fully reactive to children.

This PR fixes TLON-2072 by making all sources fully reactive to their children's statuses.

PR Checklist


patosullivan (approved):
Makes sense.

PR #3609, opened 2024-06-12
notify: get notifs from activity, not hark

Fang-:
The change as described above is simple enough. But there's a catch here that pulls in many lines of low-complexity high-nuance code: we want to continue supporting old clients.

Notify passes notification uids through a pipeline where we can't trivially slap on additional markers/details. Current/old-style clients use these uids to look up notification data in the hark agent directly.

Here, we propose a new scheme, of asking the notify agent itself about notification details, creating some separation of concerns, or at least separating us from hark.

Clients may not update to use this scheme right away, so we include code that converts new-style activity events into old-style hark notifications, and injects them into hark. This way, we can support both old and new clients while emitting a single notification uid.

Due to lack of context, this conversion cannot always achieve 100% parity with the old-style notifications. These cases have been marked for `::REVIEW` explicitly.

Smoke-tested, but not yet seen working end-to-end.

Closes TLON-1517.


desk/app/notify.hoon
```diff
@@ -332,22 +348,260 @@
...
+                ::
+                    %dm-invite
+                  ::REVIEW  'author.memo has invited you to a direct message'
+                  ::TODO    maybe want [%dm-invite =whom from=@p]
+                  :~  'Someone'
+                      ' has invited you to a direct message'
+                  ==
+                ::
+                    %dm-post
+                  ::REVIEW  should maybe not be rendered if you haven't accepted
```
jamesacklin, line 483: Correct

arthyn (approved):
lgtm, this may unintentionally produce hark duplicates, but that's ok

PR #3613, opened 2024-06-12
native: fix issue with scroller not moving to newly sent posts

patosullivan:
Fixes TLON-2049. Looks like this ternary was just flipped?


dnbrwstr (approved):
Talked about this offline -- either solution is fine, up to you.

PR #3618, opened 2024-06-13
native: properly track channel membership

latter-bolden:
Right now the way we determine if you're a member of a channel is by seeing if you have an unread for it. This is similar to what we do on web, except that web also looks to see at least one of your roles is in the _readers_ array for that channel if the _readers_ array isn't empty.

We were only tracking channel writers on mobile, not readers. This adds support for storing readers and uses a check against read entries when setting `currentUserIsMember`.

Tested by checking that channels I didn't have access to were absent after the change.

Fixes TLON-2081

### DIFF
--- a/packages/shared/src/api/channelsApi.ts
+++ b/packages/shared/src/api/channelsApi.ts
@@ -100,22 +100,27 @@ export const subscribeToChannelsUpdates = async (
   );
 };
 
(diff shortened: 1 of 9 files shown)


patosullivan (approved):
Makes sense to me!

dnbrwstr (approved):

PR #3542, opened 2024-05-28
native: integrate activity

latter-bolden:
This covers the initial integration of `%activity` into the native app which means:

- moving our unreads infrastructure to pull from there vs the channels/chat agents
- issuing channel/thread/comment reads to `%activity` more consistently (we trigger reads as soon as you see the divider, but it sticks around for a delay or until you navigate away)
- implements volume controls for top level device notifications + granular muting at the DM/group & thread level
- adds in firehose subscription for DMs so we'll get live updates for posts and reactions now

This now also includes the work for the main activity feed (middle tab)

Fixes TLON-1749
Fixes TLON-1505
Fixes TLON-2084
Fixes TLON-2083

PR #3619, opened 2024-06-13
native: Add ability to switch sorting method of channels in list

patosullivan:
Fixes TLON-2067

Adds ability to change sorting preference for the channel list (in either the interstitial screen or the channel switcher sheet), stores the user preference in async storage.

Looks like this:
Simulator Screenshot - iPhone 15 Pro - 2024-06-13 at 12 14 43
Simulator Screenshot - iPhone 15 Pro - 2024-06-13 at 12 15 13

Also fixes the issue with the missing unread counts on channels.


packages/ui/src/components/ChannelSwitcherSheet.tsx
```diff
@@ -23,8 +27,11 @@ export function ChannelSwitcherSheet({
   channels,
   onSelect,
   contacts,
+  sortBy,
+  setSortBy,
 }: Props) {
   const [hasOpened, setHasOpened] = useState(open);
+  // const [sortBy, setSortBy] = useState<'recency' | 'arranged'>('recency');
```
dnbrwstr, line 34: ?

dnbrwstr (changes requested):
Overall looks good to me -- only objection is that you and @latter-bolden are using p different pattern to do pretty similar things in `channelSortPreferences` and `keyVal` https://github.com/tloncorp/tlon-apps/pull/3542/files#diff-e60050fa0b9ae6b31ef3b2930557bc12d2a2bd4e5f5f576ec94079745790fd9c

Could you two sync up on a shared approach for query-invalidate-update for these sorts of settings?

jamesacklin:
Seeing an odd case of mismatched sorting, but this could be just me:

image

patosullivan:
> Seeing an odd case of mismatched sorting, but this could be just me:
> 
> image

yeah, I'm seeing this too, looking into it.

patosullivan:
> > Seeing an odd case of mismatched sorting, but this could be just me:
> > image
> 
> yeah, I'm seeing this too, looking into it.

fixed

latter-bolden (approved):
Thanks for updating it to match my keyValue!

PR #3620, opened 2024-06-13
native: manually update channel posts

dnbrwstr:
This pr refactors the way we update channel data. Instead of refetching the entire query when a relevant table updates, we now listen to database change events + make targeted events.
- When a new post is created, we use `setQueryData` to manually add that post to any matching `channelPosts` queries.
- We now wrap data for each individual post in a query, which we invalidate when that post (or its reactions) are updated. When the post row in the database changes, we rerun the query.


dnbrwstr:
Update: some rough edges, drafting for now.

PR #3621, opened 2024-06-13
native: fix link open issue

patosullivan:
fixes TLON-2085

We didn't need to use `canOpenUrl` here, that's not what it's for, and it was breaking the ability to open links in other apps that support them (i.e., a github link wouldn't open at all if you had github installed on your android phone, now it will open in the app).


dnbrwstr (approved):
👍

PR #3623, opened 2024-06-14
activity: dont set channel defaults if group has setting during migration

arthyn:
This fixes TLON-2087 by checking to see if the group has a setting and if so we don't set the old "default" on channels under that group.

Tested by setting a group volume with `:groups &volume-set [[%group ~nibset-napwyn %tlon] %loud]` and then running the migration

PR Checklist

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -227,6 +227,7 @@
   (welp writs replies)
 ++  set-volumes
   |=  =channels:c
+  =+  .^(=volume:v %gx (scry-path %groups /volume/all/noun))
   ::  set all existing channels to old default since new default is different
   =^  checkers  cor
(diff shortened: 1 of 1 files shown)


desk/app/activity.hoon
```diff
@@ -247,11 +248,12 @@
         .^($-([ship nest] ?) %gx path)
       [test (~(put by checkers) group test)]
     =.  cor
+      ::  don't set channel default if group above it has setting
+      ?^  (~(get by area.volume) group)  cor
```
Fang-, line 252:
```suggestion
      ?:  (~(has by area.volume) group)  cor
```

I would just `has:by` here if we're not using the result.

If the value from this gets set/copied elsewhere, may want to point to where that happens here. But not a big deal.

Fang- (approved):
Simple enough change! Always something to nit, but fine either way.

PR #3624, opened 2024-06-14
unreads: clear lingering unread markers

arthyn:
This cleans up how we were clearing unreads when navigating away because it wasn't firing in all the cases we wanted. Also fixes the unread marker never going away because we weren't actually checking status for it exactly. Also fixed reply meta updating as well as replies not showing up in threads.

Fixes TLON-2096

PR Checklist

### DIFF
--- a/apps/tlon-web/src/state/unreads.ts
+++ b/apps/tlon-web/src/state/unreads.ts
@@ -48,7 +48,7 @@ export interface UnreadsStore {
   update: (unreads: Activity) => void;
 }
 
-export const unreadStoreLogger = createDevLogger('UnreadsStore', true);
+export const unreadStoreLogger = createDevLogger('UnreadsStore', false);
 
 function getUnreadStatus(count: number, notify: boolean): ReadStatus {
   if (count > 0 || notify) {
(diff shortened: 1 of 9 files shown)


apps/tlon-web/src/state/unreads.ts
```diff
@@ -347,12 +341,17 @@ export const useUnreadsStore = create<UnreadsStore>((set, get) => ({
   delayedRead: (key, cb) => {
     const { sources, read } = get();
     const source = sources[key] || emptyUnread();
+    debugger;
```
patosullivan, line 344:
```suggestion
```
Stray debugger.

apps/tlon-web/src/state/unreads.ts
```diff
@@ -48,7 +48,7 @@ export interface UnreadsStore {
   update: (unreads: Activity) => void;
 }
 
-export const unreadStoreLogger = createDevLogger('UnreadsStore', false);
+export const unreadStoreLogger = createDevLogger('UnreadsStore', true);
```
patosullivan, line 51:
```suggestion
export const unreadStoreLogger = createDevLogger('UnreadsStore', false);
```

Do we want this left on?

patosullivan (approved):

PR #3625, opened 2024-06-14
Activity sync tweaks

dnbrwstr:


packages/shared/src/store/dbHooks.ts
```diff
@@ -379,10 +379,23 @@ export const useChannel = (options: { id: string }) => {
   });
 };
 
-export const usePostWithRelations = (options: { id: string }) => {
+export const usePostWithThreadUnreads = (options: { id: string }) => {
   const tableDeps = useKeyFromQueryDeps(db.getPostWithRelations);
   return useQuery({
     queryKey: [['post', options], tableDeps],
+    staleTime: Infinity,
```
latter-bolden, line 386: i get why we want this for the other hook, but do we need this here?

packages/shared/src/store/dbHooks.ts
```diff
@@ -379,10 +379,23 @@ export const useChannel = (options: { id: string }) => {
...
+    queryFn: () => db.getPostWithRelations(options),
+  });
+};
+
+export const usePostWithRelations = (
```
latter-bolden, line 391: maybe a comment about how these two differ/why we set it up this way if you think we'll stick with this approach

packages/shared/src/store/postActions.ts
```diff
@@ -29,12 +29,7 @@ export async function sendPost({
 
   // optimistic update
   const cachePost = db.buildPendingPost({ authorId, channel, content });
-  await db.insertChannelPosts({
-    channelId: channel.id,
-    posts: [cachePost],
-    older: sync.channelCursors.get(channel.id),
-  });
-  sync.updateChannelCursor(channel.id, cachePost.id);
+  sync.handleAddPost(cachePost);
```
latter-bolden, line 32: nice

latter-bolden (approved):
This looks good to me, feel free to merge into my branch at will

PR #3626, opened 2024-06-14
activity: self activity should change recency

arthyn:
Fixes TLON-2088 by changing the way we give recency values in that if the "read" floor is greater than the last message, make that the recency. However to get our %read action to put a value other than the last message time, I had to change the action type slightly. 

PR Checklist


Fang- (approved):
Hoon-side changes here look good, thanks for the fix!

PR #3629, opened 2024-06-17
native: add error boundary, explicitly send js errors to crashlytics

patosullivan:
Fixes TLON-2108

### DIFF
--- a/apps/tlon-mobile/src/ErrorBoundary.tsx
+++ b/apps/tlon-mobile/src/ErrorBoundary.tsx
@@ -1,0 +1,42 @@
+import crashlytics from '@react-native-firebase/crashlytics';
+import React, { Component, ErrorInfo, ReactNode } from 'react';
+import { Text, View } from 'react-native';
+
+interface ErrorBoundaryProps {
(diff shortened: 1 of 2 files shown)


apps/tlon-mobile/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,42 @@
...
+    super(props);
+    this.state = { hasError: false, error: null };
+  }
+
+  static getDerivedStateFromError(error: Error): ErrorBoundaryState {
+    return { hasError: true, error };
+  }
+
+  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+    crashlytics().recordError(error);
```
latter-bolden, line 25: maybe only attempt this if not in dev environment?

latter-bolden (approved):

apps/tlon-mobile/src/ErrorBoundary.tsx
patosullivan: Crashlytics apparently doesn't send anything unless we're running in prod.

PR #3631, opened 2024-06-17
web: fix issue with sending DM replies

patosullivan:
Fixes TLON-2107. We were attempting to use `time` (which is null when we create a new reply) as a key for optimistically adding a new reply on a parent message. We used to use `sent` on `memo` to do this (not sure why this changed), reverting back to using `memo.sent` here fixes the issue.

### DIFF
--- a/apps/tlon-web/src/state/chat/chat.ts
+++ b/apps/tlon-web/src/state/chat/chat.ts
@@ -544,7 +544,7 @@ function infiniteDMsUpdater(queryKey: QueryKey, data: WritDiff | WritResponse) {
               return;
             }
 
-            const newReplyId = unixToDa(memo.sent).toString();
+            const newReplyId = formatUd(unixToDa(memo.sent));
             const newReply: Reply = {
               seal: {
                 id: replyId,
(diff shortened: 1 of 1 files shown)


apps/tlon-web/src/state/chat/chat.ts
```diff
@@ -544,6 +544,7 @@ function infiniteDMsUpdater(queryKey: QueryKey, data: WritDiff | WritResponse) {
               return;
             }
 
+            const newReplyId = unixToDa(memo.sent).toString();
```
arthyn, line 547:
```suggestion
            const newReplyId = formatUd(unixToDa(memo.sent));
```

these keys have the dots

apps/tlon-web/src/state/chat/chat.ts
```diff
@@ -555,7 +556,7 @@ function infiniteDMsUpdater(queryKey: QueryKey, data: WritDiff | WritResponse) {
 
             const newReplies = {
               ...prevReplies,
-              [formatUd(bigInt(reply.delta.add.time!))]: newReply,
```
arthyn, line 558: we need to use this in cases where we're hearing replies, otherwise if optimistic use memo.sent

arthyn (changes requested):

arthyn (approved):
lgtm

PR #3633, opened 2024-06-18
web: fix bug with new dm creation

patosullivan:
Fixes TLON-2114. We were finding the first unread that matched `ship/${theShip}` and then returning that unread key as the existing DM. This got inserted into the route as `/new/ship/~sampel-palnet`, which puts `new` in the `:ship` parameter place. the `Message` component then checks if the `:ship` parameter is a patp. If it's not, it assumes multidm. We then routed the user to the MultiDm component, and since they're not in a MultiDM with that ID we show the invite accept/reject page (which won't work because this isn't a mutidm and that isn't a valid multidm ID).

The fix is just to return only the patp from the unread key, chopping off the `ship/` portion.

### DIFF
--- a/apps/tlon-web/src/logic/useMessageSelector.ts
+++ b/apps/tlon-web/src/logic/useMessageSelector.ts
@@ -52,11 +52,13 @@ export default function useMessageSelector() {
     }
 
     return (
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):
Makes sense!

PR #3635, opened 2024-06-18
refactor volume settings handling

dnbrwstr:


packages/shared/src/db/queries.ts
```diff
@@ -987,168 +1008,40 @@ export type ChannelVolume = {
...
-      );
-    }
-    noisyChunks.push(sql`end)`);
-    const isNoisy: SQL = sql.join(noisyChunks, sql.raw(' '));
-
-    return ctx.db
-      .update($groups)
-      .set({ isMuted, isNoisy })
-      .where(inArray($groups.id, groupIds));
+export const getVolumeSettings = createReadQuery(
```
latter-bolden, line 1027: maybe name this singularly since it just grabs one record

packages/shared/src/db/queries.ts
```diff
@@ -987,168 +1008,40 @@ export type ChannelVolume = {
...
-    noisyChunks.push(sql`(case`);
-    for (const volume of validVolumes) {
-      noisyChunks.push(
-        sql`when ${$posts.id} = ${volume.postId} then ${volume.isMuted}`
-      );
-    }
-    noisyChunks.push(sql`end)`);
-    const isNoisy: SQL = sql.join(noisyChunks, sql.raw(' '));
-
+export const clearVolumeSettings = createWriteQuery(
```
latter-bolden, line 1037: likewise

packages/shared/src/db/queries.ts
```diff
@@ -987,168 +1008,40 @@ export type ChannelVolume = {
...
-      .set({ isMuted, isNoisy })
-      .where(inArray($posts.id, postIds));
+      .delete($volumeSettings)
+      .where(eq($volumeSettings.itemId, itemId));
   },
-  ['posts']
+  ['volumeSettings']
 );
 
 export const getGroupUnreadCount = createReadQuery(
```
latter-bolden, line 1047: another slop on my part, this & its corresponding hook can be removed

packages/shared/src/store/activityActions.ts
```diff
@@ -6,93 +6,52 @@ import * as ub from '../urbit';
...
-    } else {
-      db.setChannelVolumes([
-        {
-          channelId: channel.id,
-          isMuted: channel.isMuted ?? undefined,
-          isNoisy: channel.isNoisy ?? undefined,
-        },
-      ]);
+    if (initialSettings) {
+      await db.setVolumes([initialSettings]);
```
latter-bolden, line 54: We should talk quick. I _think_ we need to handle group vs channel conditionals when rolling back to existing values

latter-bolden (changes requested):
Few small things

latter-bolden (approved):
Good to merge once that local change we looked at is pushed

PR #3640, opened 2024-06-19
web: fix for 'include all children, as appropriate, in combined unrea…

patosullivan:
Sometimes child/grandchild doesn't exist, we needed to account for that with optional chaining.


patosullivan:
Closed in favor of https://github.com/tloncorp/tlon-apps/pull/3647

PR #3645, opened 2024-06-20
native: explicit initial sync behavior

latter-bolden:
This PR moves our initial fetching logic into an explicit `startSync` function which we call once the app is authenticated and we have the Urbit client setup. The previous hook we were using refetched more than it should and used sync queues that could be accidentally cleared before completion.

Manually triggering failures in the underlying sync handlers, the only one that would cause the chat list to be blank was _syncInitData_, so now we explicitly call that first and wait for it to complete before attempting any other fetch/sync/subscribe.

Hoping some combination of the two will resolve the issues we're seeing. In case it doesn't, this also adds a new _ErrorReporter_ class with robust logging that should let us know in Crashlytics if and how something went wrong.

Fixes TLON-2118


latter-bolden:
Test runner not liking the _Crashlytics_ import, investigating now. Should otherwise be ready for code review.

packages/shared/src/store/sync.ts
```diff
@@ -746,14 +761,46 @@ export const initializeStorage = () => {
...
+        reporter.log(`finished initializing storage`)
+      ),
+      syncPushNotificationsSetting().then(() =>
+        reporter.log(`finished syncing push notifications setting`)
+      ),
+    ]);
+    reporter.log('sync start complete');
+  } catch (e) {
+    reporter.report(e);
+    logger.warn('INITIAL SYNC FAILED', e);
```
jamesacklin, line 794: Good, thank you

dnbrwstr (approved):
Nice, code lgtm!

PR #3646, opened 2024-06-20
native: catch issues w/input webview and force reload

patosullivan:
Fixes TLON-2130.

This is an attempt to catch all the ways that the editor webview could fail and then force a reload. The particular issue that was happening with TLON-2130 seems to have been a race condition with setting up tiptap within tentap (not all extensions were loaded in time, which seems like it shouldn't happen, but is). I patched tentap to catch any errors with tiptap and send a message to the native side if it happens, the native side then reloads the webview.

We also reload for other various failure modes, and inject some JS in the webview to try catch any errors that might be happening in the JS running within the webview.


latter-bolden (approved):
Lgtm!

PR #3649, opened 2024-06-21
native: a few fixes

latter-bolden:
Bit of a hodgepodge, stumbled upon these while investigating missing unreads. I was testing joining/leaving groups and noticing we could get stuck in inconsistent states.

- fixes empty info list item for invites to groups with missing metadata (falls back to displaying the group ID)
- ensures query powering the chat list only considers group channels where you're joined to the group
- adds behavior to reconcile group and channel joins & leaves in cases where we don't explicitly hear the fact
- adds more early returns for `inArray` based queries if the array is empty

Fixes TLON-1946


patosullivan (approved):
Looks good, makes sense

PR #3652, opened 2024-06-21
native: only reload input webview if we know we have an issue

patosullivan:
fixes TLON-2130 (again). Previously we were reloading anytime the editor wasn't ready, but this could cause extraneous reloads that then caused other issues. Now we'll only reload the webview if we *know* it has crashed.

Also changes the font used in the system font, happened upon this issue while looking at the xcode console (the app was endlessly spamming the console with: `Unrecognized font family 'System, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Ubuntu, 'Helvetica Neue', sans-serif'`)


latter-bolden (approved):
Nice, looks good

PR #3653, opened 2024-06-21
native: fix initial post author

dnbrwstr:
This fixes the issue where the post author is incorrectly displayed (missing contact) when a just-sent post is initially rendered.


patosullivan (approved):
Makes sense!

PR #3657, opened 2024-06-24
native: fix crash on navigating to gallery with one unread

patosullivan:
Fixes TLON-2148, we need to make sure we have an index great than zero, otherwise we'll try to scroll to -1 when the firstUnread is at the 0 position.


latter-bolden (approved):
Nice catch!

PR #3658, opened 2024-06-24
native: hide input when channel is a DM invite

latter-bolden:
Micro PR to fix a small regression.

Fixes TLON-2151


dnbrwstr (approved):
Lgtm!

PR #3659, opened 2024-06-24
native: fix off by 1 DM reply counts

latter-bolden:
OTT

Fixes TLON-2155


patosullivan (approved):
Makes sense, code looks good.

PR #3661, opened 2024-06-24
activity: correct bad activity keys 

arthyn:
Fixes TLON-2103 by adding some cleanup to the on-load arm. This is some of the most wild juggling, but I'm not sure how to do it any better.

The basic idea is go through all our sources that come from clubs that are dms or dm threads. For DMs we just need to comb through the stream and correct any keys we find. 

For threads we have to do more work since the source identifiers themselves have bad keys. So first we gather all the source index pairs that share the same message ID, then we go through the entire list of threads and look to see what was gathered under the ID they shared. If we don't have an entry for that source yet we create one and merge all the data collected under that same ID. Then we keep going until we have the real threads. Finally once we come back around to actually update the indexes, we make sure that we delete any data associated with old sources.

Tested by installing from binnec, then sending some multi-dms replies to generate bad data, then installed these changes and saw that dm threads successfully cleared.

PR Checklist

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -97,7 +97,7 @@
   =?  old  ?=(%1 -.old)  (state-1-to-2 old)
   ?>  ?=(%2 -.old)
   =.  state  old
-  correct-dm-keys
+  (emit %pass /clean-keys %agent [our.bowl dap.bowl] %poke noun+!>(%clean-keys))
   +$  versioned-state  $%(state-2 state-1)
   +$  state-2  current-state
   +$  state-1
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -97,7 +97,7 @@
   =?  old  ?=(%1 -.old)  (state-1-to-2 old)
   ?>  ?=(%2 -.old)
   =.  state  old
-  cor
+  correct-dm-keys
   +$  versioned-state  $%(state-2 state-1)
   +$  state-2  current-state
   +$  state-1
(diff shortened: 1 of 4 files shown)


latter-bolden (approved):
JS change lgtm

desk/app/activity.hoon
```diff
@@ -108,6 +108,123 @@
...
+    ::  cleanup old bad keys
+    =?  cor  ?=(%dm-thread -.source)
+      =/  old-source  source(key [id.key.source q.id.key.source])
+      %=  cor
+        indices  (~(del by indices) old-source)
+        activity  (~(del by activity) old-source)
+        volume-settings  (~(del by volume-settings) old-source)
+      ==
+    ::  update source + index, if new key create new index
+    =.  cor  (update-index source index.i.indxs &)
```
Fang-, line 150: We do processing on the `%dm-thread` sources here exclusively, because those sources (source ids, indices, w/e) themselves have the incorrect keys, right?

Also, don't we want to do the `+update-index` call only once at the end, in the `?~ indxs` case?

Fang- (commented):
Ok yeah, this is hairy. I think it's mostly good, but I do have a correction and some questions. Also, we may benefit from doing a pair review still.

desk/app/activity.hoon
arthyn: yes both the sources and the indexes need correction. and no we have to call update-index for each one since they are disparate indexes and each one needs to update.

desk/app/activity.hoon
```diff
@@ -97,7 +97,7 @@
   =?  old  ?=(%1 -.old)  (state-1-to-2 old)
   ?>  ?=(%2 -.old)
   =.  state  old
-  cor
+  correct-dm-keys
```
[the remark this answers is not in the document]
arthyn: I'll switch it to be on state transition in my next pr which changes state, but yeah I will make it poke here.

desk/app/activity.hoon
```diff
@@ -108,6 +108,123 @@
     [%2 %all +.old]
   --
 ::
+++  correct-dm-keys
+  |^  ^+  cor
```
Fang-, line 112:
```suggestion
::  at some time in the past, for clubs activity, %dm-post and %dm-reply events
::  with bad message/parent identifiers (respectively) got pushed into our
::  streams. for the %dm-reply case, this made it impossible for clients (that
::  use and expect the correct identifiers) to work with the sources these
::  events got put into. for both cases, they were Wrong.
::
::  here, we iterate over all clubs and, for each of the known sources for a
::  club, re-constitute the contents of that source's stream from scratch. this
::  means _moving_ dm-thread streams to be under corrected source identifiers,
::  and updating top-level streams _in-place_. for both, we update the contents
::  of the stream by rewriting %dm-post and %dm-reply events to use corrected
::  message identifiers.
::
::  to correct the message identifiers, we simply do the reverse time-id lookup
::  in dex.pact, as is done across the chat codebase.
::
++  correct-dm-keys
  |^  ^+  cor
```

Fang- (commented):
As discussed, some suggestions for clarity. And also yes, let's put it near the bottom of the file lol.

Fang- (approved):
Changes look good, thank you!

PR #3663, opened 2024-06-25
native: handle subscription discontinuity

latter-bolden:
The Urbit channel we use to maintain our subscriptions gets periodically reset; even long running connections while the app is open in the background are automatically closed after 12 hours. Once this happens, we can't expect to have heard every update event and won't be able to guarantee the validity of our local cache. Ensuring we keep data comprehensively up to date will require more work, but the most pressing symptom we need to address is our channel cursors getting stale which leads to new messages not showing up correctly in channels.

We updated the Urbit client to properly emit a signal when the underlying channel cycles. This PR adjusts our logic to listen for those signals and when they occur, run a discontinuity recovery procedure. Right now that's limited to clearing stale cursors and re-running our init behavior, but this is something we can flesh out over time.

Also bumps the heads query higher up in the start fetch ordering to ensure we're setting those cursors as early as possible.

Fixes TLON-2158


latter-bolden:
@dnbrwstr wondering if we should also explicitly invalidate the `channelPosts` queries when this happens?

packages/shared/src/store/sync.ts
```diff
@@ -113,12 +113,10 @@ function checkForNewlyJoined({
 }
 
 export const syncLatestPosts = async (reporter?: ErrorReporter) => {
-  const result = await syncQueue.add('latest-posts', async () =>
-    Promise.all([
-      api.getLatestPosts({ type: 'channels' }),
-      api.getLatestPosts({ type: 'chats' }),
-    ])
-  );
+  const result = await Promise.all([
```
dnbrwstr, line 116: Was sync queue causing an issue?

apps/tlon-mobile/src/lib/api.ts
```diff
@@ -34,10 +34,18 @@ export function configureClient({
   shipName,
   shipUrl,
   onReset,
```
dnbrwstr, line 36: Should we remove `onReset`? Not clear on what it's actually doing for us after the convo yesterday.

dnbrwstr (approved):
Lgtm! Couple minor things.

dnbrwstr:
> @dnbrwstr wondering if we should also explicitly invalidate the `channelPosts` queries when this happens?

Probably yes, I think so.

packages/shared/src/store/sync.ts
latter-bolden: I want it to run immediately when called and didn't want to have to worry about the queue getting cleared.

I think until we move things over to a priority based system, it's gonna be simpler to avoid the queue for anything mission critical.

apps/tlon-mobile/src/lib/api.ts
latter-bolden: We can if you'd prefer, I don't feel strongly. It's not there to handle anything explicit right now, at the moment it's only relevant if we ourselves manually attempt to reset it. 

Would rather leave it in defensively though since if the client ever does emit that, we can be certain we need to resub.

packages/shared/src/store/sync.ts
dnbrwstr: How about this -- we set the concurrency for the `syncQueue` to `Infinity` so that it runs everything immediately. That way we can maintain the bottleneck + logging for when we move to priority queue without adding overhead.

dnbrwstr: Actually whatever, this is gonna get complicated. Remove it + we'll hash it out later lol.

latter-bolden: haha ya, lets chat tomorrow. I think as we start mapping out the background sync stuff, updating the queue to work the way we want could be a reasonable first step

apps/tlon-mobile/src/lib/api.ts
dnbrwstr: My take would be remove the param, but pipe all the "potential reset" events through one handler. Up to you tho, not blocking! If that doesn't make sense, merge away.

PR #3665, opened 2024-06-25
native: tee up group settings work

patosullivan:
- Adds the long press sheet back for groups, updating it to match current wireframes.
- Adds a new stack navigator for group settings and new stub screens (some of these screens could end up being navigators themselves).
- Adds useGroupContext where we'll pull data or functions for setting data, with probably nearly all of the functions stubbed out.

Fixes TLON-2181

### DIFF
--- a/apps/tlon-mobile/src/screens/ChatListScreen.tsx
+++ b/apps/tlon-mobile/src/screens/ChatListScreen.tsx
@@ -161,11 +161,11 @@ export default function ChatListScreen(
   );
 
   const handleGoToGroupMeta = useCallback(
-    (group: db.Group) => {
+    (groupId: string) => {
       // @ts-expect-error TODO fix nested navigator types
       navigation.navigate('GroupSettings', {
         screen: 'GroupMeta',
-        params: { group },
+        params: { groupId },
       });
 
       setLongPressedGroup(null);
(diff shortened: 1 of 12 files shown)


apps/tlon-mobile/src/screens/ChatListScreen.tsx
```diff
@@ -162,6 +240,8 @@ export default function ChatListScreen(
     setScreenTitle(title);
   }, []);
 
+  console.log({ pinned });
```
latter-bolden, line 243:
```suggestion
```

apps/tlon-mobile/src/types.ts
```diff
@@ -41,12 +41,34 @@ export type RootStackParamList = {
...
     uri?: string;
   };
+  GroupSettings: {
+    group: db.Group;
+  };
+};
+
+export type GroupSettingsStackParamList = {
+  GroupMeta: {
+    group: db.Group;
```
latter-bolden, line 51: Somewhat random, but I've found it occasionally annoying to always need a full model when routing. If we're usually going to be pulling the group off `useGroupContext` anyhow, wonder if we could get away just passing the `groupId`

Fine as is if you'd prefer, just a stray thought

latter-bolden (approved):
This is looking good to me!

apps/tlon-mobile/src/types.ts
patosullivan: Yeah, agreed. I'll change it.

PR #3666, opened 2024-06-25
native: provisional welcome card

jamesacklin:
Adds an AsyncStorage key `splash` to either show or hide the welcome sheet. Also introduces a Welcome sheet with provisional icons and a rasterized graphic.

This will need to get in before our Wednesday testing cohort installs the app. I'd appreciate the expedited review / glossing-over of graphic details.

I will create follow-up tasks to replace the icons with the selections in the Figma design and replace the raster graphic with an SVG (or at least a 2x-resolution image).


packages/ui/src/components/WelcomeSheet.tsx
```diff
@@ -0,0 +1,105 @@
...
+}) {
+  const handleDismiss = useCallback(() => {
+    onOpenChange(false);
+  }, [onOpenChange]);
+
+  return (
+    <Sheet
+      open={open}
+      onOpenChange={handleDismiss}
+      snapPoints={[85]}
```
latter-bolden, line 23: do we need to explicitly set this vs using `fit` mode? Felt like a lot of white space at the bottom on my device

latter-bolden (approved):
Lgtm given time constraints

packages/ui/src/components/WelcomeSheet.tsx
jamesacklin: Wasn't aware of that! Will fix.

PR #3667, opened 2024-06-25
native: sync behavior improvements p1

dnbrwstr:
- Create post windows when inserting new posts
- Prevent updates to `useChannelPosts` data while fetching.


dnbrwstr:
Note: type error is on current develop, not related to this code.

latter-bolden (approved):
After chatting, this makes sense & should be low risk. Feedback on the build seems positive as well.

Let's get this in!

PR #3670, opened 2024-06-26
update tamagui dep

dnbrwstr:
Updates Tamagui packages from 1.93.2 to 1.101.3 (latest). No breaking changes, tested in sim and via Expo build. 

Details here: https://github.com/tamagui/tamagui/releases


patosullivan (approved):
Cool. I also checked it out on android, works fine.

PR #3672, opened 2024-06-26
add cache control header to tlon-hosted uploads

waiyaki:
I am investigating how we can add some caching to tlon-hosted assets. The easiest/most straightforward way is to add the cache-control header to the asset while uploading it. This needs to be done [when creating the upload URL](https://github.com/tloncorp/ylem/pull/1985) and when uploading the asset from the client, which is what this PR attempts to do.

I found the places in this PR that reference upload headers, but I need another set of eyes to guide me here &mdash; in the approach as well as development process.

PR Checklist


latter-bolden (approved):
This looks great! Approach feels straightforward, updates the correct codepaths for native and web.

Confirmed hosted uploads work in the simulator and saw the cache control applied to the uploaded asset. Will check that it works for web once it's on wannec.

@waiyaki unless there's anything else you wanted to add, feel free to move it out of _draft_ status and we can get it merged.

waiyaki:
That's great! Thanks for taking a look!

waiyaki:
@latter-bolden I'm marking this as do-not-merge for now. I need to think a little more about this &mdash; deploying memex with the cache changes is instant while rolling out app changes is gradual, and this will break uploads for any clients without the corresponding cache headers.

latter-bolden:
Confirmed after the latest hosting-side changes that this continues to work as expected in the staging environment

latter-bolden:
Going to wait to merge this until after we cut for the next native app tester release.

PR #3673, opened 2024-06-26
native: handle deleted unread

dnbrwstr:
Currently, when the first unread of a channel is a deleted message, the channel screen will appear blank. This PR fixes that by:
- Modifying the `getChannelPosts` query so that it can retrieve posts around a given post, even if that post has been deleted, as long as it's covered by a window. If the post is deleted, it will get posts around where the post would have been.
- Adding logic to the `Scroller` that will make it visible once we've rendered the first page of posts, even if we haven't found our anchor post.

Fixes TLON-2142

### DIFF
--- a/packages/shared/src/db/queries.ts
+++ b/packages/shared/src/db/queries.ts
@@ -1601,23 +1601,27 @@ export const getChannelPosts = createReadQuery(
         .as('posts');
 
       // Get the row number of the cursor post
-      const cursorRow = await ctx.db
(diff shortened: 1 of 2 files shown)


dnbrwstr:
I said TLON-2142

latter-bolden (approved):
Makes sense, works locally

jamesacklin:
@wca4a For your edification. Should resolve the issue you saw with Tlon Local not loading.

PR #3674, opened 2024-06-26
post loading ux

dnbrwstr:
Improvements to post loading behavior + performance
- Debounce loader changes so that they don't trigger repeatedly
- Actually load newest page of posts when entering a channel with no unreads.
- Increase page size of posts when loading from the api


latter-bolden (approved):
Lgtm, made sense in our walkthrough

jamesacklin:
Excited to see!

PR #3677, opened 2024-06-27
native: sync start robustness

latter-bolden:
Adds retry logic to the individual pieces of our fetching startup sequence. If after retrying we still fail, we'll pop the error boundary to prevent app use in an unknown state.

Once we support offline mode (where failure here would make sense), we'll need to be more careful about deciding when exactly to throw. But for now, we lock the screen if there's no network connection anyhow.

Fixes TLON-2183


dnbrwstr (approved):
Lgtm, will get deeper into this this afternoon.

PR #3678, opened 2024-06-27
native: temporarily disable sigil rendering in push notifications

dnbrwstr:
This PR fixes a crash that happens frequently on receipt of a push notification. May need @alecananian's advice to get to a real fix.


latter-bolden:
I'll copy this over to the main branch. Thanks for figuring this out!

PR #3681, opened 2024-06-27
groups-ui: fix init types

arthyn:
PR Checklist


Fang- (approved):
Good fix. We should roundtable at some point to think of ways to catch these things earlier. A "I just changed some types, what's next?" checklist isn't gonna cut it, we need automation/compiler failures somehow...

PR #3682, opened 2024-06-27
activity: refine combined count for channels

arthyn:
After discussion with @latter-bolden, we decided to change how `count` in activity summaries gets tallied to better reflect what expect UX-wise, aka ignoring threads that don't involve us. 

We refresh all summaries on load since that shouldn't take very long and likely should be done anytime the agent changes, and in this case we have to for those counts to get updated immediately.

PR Checklist

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -100,7 +100,7 @@
   =?  old  ?=(%2 -.old)  (state-2-to-3 old)
   ?>  ?=(%3 -.old)
   =.  state  old
-  cor
+  refresh-all-summaries
   +$  versioned-state  $%(state-3 state-2 state-1)
   +$  state-3  current-state
   +$  state-2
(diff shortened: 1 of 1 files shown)


desk/app/activity.hoon
```diff
@@ -899,7 +899,11 @@
       notify-count  (^add notify-count.sum notify-count.as)
     ==
   =/  newest=time  :(max newest.cs floor.reads top)
-  =/  total  count.cs
+  =/  total
+    ::  if we're a channel, we only want thread notify counts, not totals
+    ?:  ?=(%channel -.source)
```
Fang-, line 904:
```suggestion
    ::  if we're a channel, we only want thread notify counts, not totals
    ::
    ?:  ?=(%channel -.source)
```

(;

Fang- (approved):
I can debate the expectation and its implications here, but the code seems to reflect the description at least, so looks good to me.

arthyn:
Fixes TLON-2202

PR #3683, opened 2024-06-27
various: add resource-id-ing ~| into common paths

Fang-:
As discussed out of band with @arthyn, we should have some more of these. This is a start.

We see crash traces for all kinds of things. Currently, we print poke and watch nack traces to the user's terminal. These by themselves only point at the codepath, but don't explicate the crash reason (which could also be deduced from the trace) or the resource that was operated on (which might not be recoverable in cases where we don't have/print the wire).

Here, we add a couple of trace hints, to print subscription paths and resource identifiers in common crash pathways.

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -279,7 +279,8 @@
 ++  watch
   |=  =(pole knot)
   ^+  cor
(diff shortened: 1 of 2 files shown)


arthyn (approved):
very nice!

PR #3684, opened 2024-06-27
activity: allow clearing unreads for deleted messages

arthyn:
This fixes TLON-2188 by showing a deleted message placeholder so that the unread marker can be displayed. Additionally this stops the %channels agent from sending edits as activity events. If we want this in the future, we'll probably want to handle it differently.

image

PR Checklist


Fang- (commented):
Hoon-side change looks fine, but we should note that this does mean message contents as stored in/retrieved from activity will be stale in the case of edits. Probably important for "activity view" use cases.

PR #3685, opened 2024-06-27
native: include involved threads in group/channel count

latter-bolden:
OTT

The way @arthyn updated the backend, changes needed here are actually pretty light!

Fixes TLON-2157


dnbrwstr (approved):
Great, lgtm!

PR #3686, opened 2024-06-28
activity: increment time-ids in "better" steps

Fang-:
Instead of incrementing the atom, add a proper fraction to it. This way, we don't end up with crazy long `@da` renderings all over the place.


arthyn (approved):
lgtm

PR #3687, opened 2024-06-28
native: priority syncing

dnbrwstr:
Initial, naive implementation of priority syncing. It touches a lot of sync code, and the initial approach is fairly naive, but it already yields some benefits + I think it probably makes sense to get this chunk in then keep working.

### Priority-based `syncQueue`

- Add `priority` param to sync manager operations.
- Higher priority tasks will always be executed before lower priority ones. 
- We leave "threads" open for high priority operations, so they can't get blocked by low priority ones. 

We could get a lot fancier with this, but it already feels decent. 

### Changed posts syncing
- We now track a `syncedAt` field on posts, updated each time the post is fetched from the
- Inside a channel, we monitor the `posts` array for posts whose `syncedAt` field is from before the current session.
- When we find a post that needs to be synced, we request it be synced at medium priority, which should mean it will never block actually loading additional pages. 

The approach is naive, but surprisingly doesn't feel too bad. Next step will be to proactively sync changes so that we can hopefully do less of this in-channel.


apps/tlon-mobile/src/screens/ChatListScreen.tsx
```diff
@@ -69,8 +69,7 @@ export default function ChatListScreen(
 
   useFocusEffect(
     useCallback(() => {
-      store.syncUnreads();
-      return () => store.clearSyncQueue();
+      store.syncUnreads(store.SyncPriority.High);
```
jamesacklin, line 72: !!!

latter-bolden (approved):
Lgtm! Demo build is working smoothly minus the missing row swiping we talked about.

This will be nice to have both immediately and as something to build off as we start doing more prefetching/pre-emptive change syncing

dnbrwstr:
Fixed a couple remaining things, should be good to go, but I'll wait a couple hours to see if anyone sees anything wonky.

PR #3689, opened 2024-06-28
notify: track client request statistics

Fang-:
We want to have some insights into how push notifications propagate to
client devices. But, client devices currently request notification
details through the scry endpoints, leaving us no way to write metadata
about the request into state. To support this, we make two changes:

First, we add new http endpoints under `/apps/groups/~/notify`, that
expose the same data that the current scry endpoints do; they support
both `%activity-event` and `%hark-yarn` shaped responses.

Second, we include a "first-req" timestamp alongside each notification
in state. Whenever an http request for a notification comes in, we set
that timestamp for that notification (if it wasn't already).

To easily digest that data, we add a `+groups!notify-delivery-report` generator. Running that generates outputs that looks like the following:

```
Between ~2024.6.27 and ~2024.6.28,
Served 3 out of 4 notifications (75%).
Average time-to-request was ~s15.
```

Additionally, instead of wiping all notification state periodically, we
now only wipe older notifications, leaving more-recent ones in place.
This way we should always keep around a relevant backlog of notification
details.

Not included in this PR: an update to the mobile client to use the new http endpoint (`/apps/groups/~/notify/note/[uid]/hark-yarn.json`) instead of the scry endpoint. This should be a drop-in replacement, though we'll eventually want to use `activity-event.json` requests instead.

~~(Given that backend and mobile client won't update in lockstep, maybe we want to add a debug poke for clearing notify's state, so we don't leave seemingly-unrequested notifications around until cleanup?)~~ Added.

Fixes TLON-2159


arthyn (approved):
lgtm

PR #3690, opened 2024-06-28
native: optimize list rendering

dnbrwstr:
This PR makes rendering our lists a million times more perfomant lol. Will probably want to keep tuning.


latter-bolden (approved):
Amazing, love to see it, lgtm, makes sense, let's get it in

PR #3694, opened 2024-06-28
native: unreads hotfix

latter-bolden:
I missed versioning the new unreads scry, so we were getting correct data on init but pulling incorrect data on updates. This fixes it to consistently use the new API.

Also leaves unread thread indicator dots gray if you're uninvolved (not just muted)

Fixes TLON-2212


patosullivan (approved):
Looks good!

dnbrwstr (approved):

PR #3695, opened 2024-06-28
native: account for changes in thread unreads when using live post

latter-bolden:
We were invalidating on changes to the post and post reactions tables, but not the unread. Symptom was you'd have to leave the channel to see the thread marked as read.

Fixes TLON-2214


dnbrwstr (approved):
V nice.

PR #3696, opened 2024-06-28
native: channel management screens

patosullivan:
Fixes TLON-1716.
Fixes TLON-1712
Fixes TLON-1708

Fixes TLON-2266
Fixes TLON-2265


packages/ui/src/tamagui.config.ts
```diff
@@ -67,6 +67,7 @@ export const tokens = createTokens({
     '3xl': 32,
     '4xl': 48,
     '5xl': 64,
+    '6xl': 72,
```
dnbrwstr, line 70: Is this shown in design somewhere? Would rather not add here unless we really need it.

patosullivan:
FYI, drag and drop is still a bit janky on android, but functionally works. Could follow up later to clean that up.

packages/shared/src/store/dbHooks.ts
```diff
@@ -227,6 +227,7 @@ export const useChannel = (options: { id: string }) => {
   return useQuery({
     queryKey: [['channel', options]],
     queryFn: () => db.getChannelWithLastPostAndMembers(options),
+    enabled: !!options.id,
```
dnbrwstr, line 230: Might as well make the param optional in this case

packages/shared/src/store/groupActions.ts
```diff
@@ -182,3 +185,365 @@ export async function deleteGroup(group: db.Group) {
     await db.insertGroups({ groups: [group] });
   }
 }
+
+export async function addNavSection(
+  group: db.Group,
+  navSectionMeta: db.ClientMeta
+) {
+  const idParts = formatUv(bigInt(Date.now())).split('.');
```
dnbrwstr, line 193: Can this happen in `urbit/utils`? Just trying to keep encapsulation.

dnbrwstr (approved):
Looking good -- some comments on code, mostly on the api stuff. 

Playing with this, I'm wondering whether we should just do arrows for drag and drop? It's pretty difficult to work with moving the larger sections around, and I can tell it's going to be a good chunk of work to get all the interactions feeling polished.

dnbrwstr (changes requested):
Oops wrong status haha.

packages/ui/src/tamagui.config.ts
patosullivan: Yeah, the channel items are shown as 72px: https://www.figma.com/design/mUIcOGeELxWswuusYUnEiW/2024-Q2-Tlon.app?node-id=4001-64568&t=FhFpsvOxwFgP6jzV-4 

Could hardcode as 72px?

dnbrwstr: Ah got it. Yeah, I think hardcode for now -- let's see if it gets designed into the system.

packages/shared/src/store/dbHooks.ts
patosullivan: This is an artifact of an earlier implementation on this PR, removing.

jamesacklin:
Heard on the section-draggging usability.

Here's an alternative approach: https://www.figma.com/design/mUIcOGeELxWswuusYUnEiW/2024-Q2-Tlon.app?node-id=4001-64564&t=Mlu2XTVUcydU1IET-4

image

@patosullivan Does this seem amenable?

patosullivan:
@jamesacklin I updated to use the arrows

jamesacklin:
Raised TLON-2265 and TLON-2266 in experiential testing. Code otherwise looks good.

patosullivan:
Fixes TLON-2266

jamesacklin (approved):
Thanks for the turnaround!

PR #3697, opened 2024-06-28
list item folder refactor

dnbrwstr:
A Friday afternoon kind of refactor -- have been wanting to do this for a while, but it touches a lot. Our list items are pretty all over the place in how they're imported, where the main content is, etc.

This PR:
- Creates a `ListItem` folder where all ListItems live.
- Standardizes imports / exports: all substantial `ListItems`  have been moved to standalone files that are named `<X>ListItem` and that export `<X>ListItem` as a named export.

Not that this PR doesn't include any modifications to how code works, it just swaps things around. Will have a followup later that refactors/cleans up the files, but didn't want to obscure that work with the bigger move changes.


latter-bolden (approved):
Always nice to see things getting a little cleaner

Tried locally, clicked around. Everything appears in order

PR #3698, opened 2024-06-29
activity: recursive unreads

arthyn:
This fixes TLON-2187 and TLON-2206 by changing the backend to keep unread state synced across parent and children sources. It also adds a new way to mark read that marks read recursively down through the children which we now use on the frontend.

PR Checklist


desk/app/activity.hoon
```diff
@@ -770,50 +596,85 @@
...
   ::
       %item
-    |=  =index:a
-    =-  index(items.reads -)
-    %+  put:on-read-items:a  items.reads.index
-    [id.action ~]
+    =/  new-read  [id.action ~]
+    =/  read-items  (put:on-read-items:a items.reads.index new-read)
+    =.  cor  (update-parents source ~[new-read])
+    (update-index source index(items.reads read-items) &)
```
Fang-, line 623: Here and in `+read`, it looks like we're processing parents first, and then the index/source itself. But I thought we want to do children first, and then the parents? Or is that only relevant during summary calculation? iow: any reason we don't always go local->up?

desk/app/activity.hoon
```diff
@@ -770,50 +596,85 @@
...
+    (update-index source new &)
+  ::
+      %recursive
+    =/  new-floor=time
+      ?^  time.action  u.time.action
+      =/  latest=(unit [=time event:a])
+        (ram:on-event:a stream.index)
+      ?~(latest now.bowl time.u.latest)
+    =/  new=index:a  index(reads [new-floor ~])
+    ~&  [source floor.reads.index new-floor]
```
Fang-, line 645:
```suggestion
```

desk/app/activity.hoon
```diff
@@ -838,6 +699,21 @@
...
+  ?<  ?=(%base -.source)
+  :-  [%base ~]
+  ?+  -.source  ~
+    %channel  ~[[%group group.source]]
+    %dm-thread  ~[[%dm whom.source]]
+  ::
+      %thread
+    :~  [%channel channel.source group.source]
+        [%group group.source]
+    ==
```
Fang-, line 715: Since we're always _pre_-pending `%base`, the `%thread` case should be `~[%group %channel]` in that order, right? Otherwise the `%thread` case isn't fully ordered.

(And then, arguably, if we want to process bottom-up instead of top-down, maybe the entire order should be reversed.)

Fang- (commented):
Not immediately obvious what has changed about the big arms you moved around, and I may have misunderstood something about the processing order for these things, so in addition to the questions here we should probably go over the logic on call again.

desk/app/activity.hoon
```diff
@@ -596,9 +419,12 @@
     (give-update update ~)
   =?  cor  &(!importing notify (is-allowed inc))
     (give %fact ~[/notifications /v0/notifications] activity-event+!>([time-id event]))
-  =.  indices
-    =/  =stream:a  (put:on-event:a stream:base time-id event)
-    (~(put by indices) [%base ~] [stream reads:base])
+  ::  we always update sources in order, so make sure base is last
```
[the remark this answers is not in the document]
arthyn: going to try making a generic helper to see how bad it gets. do you think I should add a preamble to the agent?

desk/app/activity.hoon
```diff
@@ -619,49 +619,57 @@
...
-    =/  new  index(items.reads (malt new-reads))
-    =?  cor  !from-parent
-      (update-parents source new-reads)
-    (update-index source new &)
-  ::
-      %recursive
-    =/  new-floor=time
+    =/  new=index:a
+      ::  if we're only marking at our level, then we only want to grab
+      ::  read items from our level, because we can't move the floor in
```
Fang-, line 667:
```hoon
::  take every event between the floor and now, and put it into
::  the index's items.reads. this way, the floor can be moved without
::  "losing" any unreads, and the call to +refresh-index below will clean
::  up unnecessary items.reads entries.
::
```

Here or slightly further down.

Fang-: Also we should move the `?:  deep.action` branching to only branch once at the top of this function. More clear/coherent wrt guarantees about the changes we're making to the index.

Fang- (commented):
As discussed.

Fang- (approved):
Sweet, seems all good now!

PR #3700, opened 2024-07-01
native: refactor list items and avatars.

dnbrwstr:
Hoping to get this in while it doesn't conflict with much. Refactors our handling of `ListItem`s and avatars to reduce redundancy + make things more consistent.

For avatars, adds a suite of `Avatar` components separate from list icons, that we should use everywhere we need that sort of thing.

High level:
- `GroupAvatar`
- `ChannelAvatar`
- `ContactAvatar`

Low level: 
- `SystemIconAvatar`
- `ImageAvatar`
- `TextAvatar`
- `SigilAvatar`

Also made it so that `ContactAvatar` automatically pulls the contact, we don't have to pass it in, and implemented background color awareness for `TextAvatar`s.

Should have only very minor visual impact; I've gone through the call sites I could find and verified things look sane. May need additional tuning in some spots, but will give us unified spots to that.

Before:

After:


packages/ui/src/components/Activity/ChannelActivitySummary.tsx
```diff
@@ -90,84 +86,21 @@ export function ChannelIndicator({
...
-        <SizableText marginLeft="$s" fontSize="$s" color="$secondaryText">
+        <SizableText fontSize="$s" color="$secondaryText">
           {logic.makePrettyTime(new Date(sentTime))}
         </SizableText>
       )}
     </XStack>
   );
 }
-
-// TODO: we dont really have a sizable channel/group icon, we use hardcoded ListItem everywhere?
```
latter-bolden, line 108: 🙏🏻

latter-bolden (approved):
Code looks good. 

Aside from verifying activity looked okay, have not thoroughly reviewed changes in app.

PR #3705, opened 2024-07-02
native: more empty array checks on `insertGroups`

latter-bolden:
More prevention against running array based DB operations when the array is empty. In particular, this targets `insertGroups` which is a spot where we're seeing crashlytics start sync issues. Not positive this will resolve those, but I think this is worth including in the next build and should be a low risk change.

Fixes TLON-2224 (hopefully)


jamesacklin:
Could be the move. Thanks for tracking this down.

patosullivan (approved):
Makes sense

PR #3710, opened 2024-07-03
native: update crashlytics setup

patosullivan:
I added the react-native-firebase plugins to the expo config and some statements to the Podfile. Kind of a mix of the instructions available on rnfirebase.io, but in local testing (by adding a firebase.json with `crashlytics_debug_enabled` set to true and calling crashlytics().crash() in the App.main.tsx file), this is consistently logging crashes in crashlytics. 

Not 100% certain that this is a magic fix but it does seem to consistently work.

You can test by doing the same thing I mentioned here, just set that property to true in firebase.json and call crashlytics().crash() somewhere in the app, then run ios:preview or ios:release, you should see a new crash logged in the release version of the app in crashlytics (for either release or preview builds, not clear why it always goes to release in crashlytics).


dnbrwstr (approved):
Lgtm! Let's give it a try.

PR #3711, opened 2024-07-03
native: pop profile sheet for ShipMentions in inline content

jamesacklin:
Matches current desktop behavior. Tap someone's mentioned ContactName, see their profile. (This was a point of early feedback from userland, no Linear issue.)

Not convinced this is completely safe and may cause some crashes. If so, we can close and move on.

https://github.com/tloncorp/tlon-apps/assets/748181/97753947-5c28-4781-a57f-799a174bd624

### DIFF
--- a/packages/ui/src/components/ContentRenderer.tsx
+++ b/packages/ui/src/components/ContentRenderer.tsx
@@ -1,4 +1,8 @@
-import { extractContentTypesFromPost, utils } from '@tloncorp/shared';
+import {
+  extractContentTypesFromPost,
+  useContact,
+  utils,
+} from '@tloncorp/shared';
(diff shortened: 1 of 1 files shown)


dnbrwstr (commented):
Wdyt about sticking the profile sheet at the screen root level, and passing methods down in context to pop it, a la `useRequests`? That way we're not re-rendering the wrapper everywhere.

jamesacklin:
> Wdyt about sticking the profile sheet at the screen root level, and passing methods down in context to pop it, a la `useRequests`? That way we're not re-rendering the wrapper everywhere.

Cool, that sounds like a better approach. I'll try that.

PR #3712, opened 2024-07-03
activity: moving helpers to lib

arthyn:
OTT, makes reading through the agent much nicer

PR Checklist


desk/app/activity.hoon
```diff
@@ -866,63 +726,22 @@
...
-    ^-  update-0:old:a
-    ?+  -.update  update
-        %read
-      [%read source.update (activity-summary-0 activity-summary.update)]
-    ==
-  --
+::
+::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+::  repent ye, for the sins of the past have come to haunt us         ::
+::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
```
Fang-, line 732: Such comment blocks are technically illegal, you know. (;

Most of the arms below already have their own paragraph(s) for each specific sin. If we really want a separator, may be more appropriate to use `+|` chapters.

Fang- (commented):
Nice, looks better indeed. How would you feel about further splitting the library up into scope-separated cores? ie having `+get-parents:source:lib` instead of `+get-parents:lib`, `+get-reads:stream:lib`, etc? (Though probably `+src:lib` and `+stm:lib` or w/e to avoid shadowing too hard.)

May or may not make sense for some of those to be written as doors, too, not sure how the ergonomics of that would shake out.

desk/app/activity.hoon
arthyn:
```suggestion
::
```

Fang- (approved):
As discussed out of band, thanks for your changes. Approved!

PR #3714, opened 2024-07-05
native: fix reply references

patosullivan:
We weren't passing the replyId into ChannelReference or ChatReferenceWrapper, previously.

Fixes TLON-2213.


dnbrwstr (approved):
Nice, lgtm!

PR #3720, opened 2024-07-08
native: implement notebook refs

patosullivan:
Fixes TLON-2255

Looks like this:
Simulator Screenshot - iPhone 15 Pro - 2024-07-08 at 15 54 37
Simulator Screenshot - iPhone 15 Pro - 2024-07-08 at 15 54 45
Simulator Screenshot - iPhone 15 Pro - 2024-07-08 at 15 54 48

Will cause some merge conflicts with my other PR, but nbd.

### DIFF
--- a/packages/ui/src/components/ContentReference/ChannelReference.tsx
+++ b/packages/ui/src/components/ContentReference/ChannelReference.tsx
@@ -2,6 +2,7 @@ import { getChannelType } from '@tloncorp/shared/dist/urbit';
 
 import { PostViewMode } from '../ContentRenderer';
 import ChatReferenceWrapper from './ChatReferenceWrapper';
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/ContentReference/ChannelReference.tsx
```diff
@@ -34,9 +35,12 @@ export default function ChannelReference({
   if (channelType === 'notebook') {
     // TODO: Implement notebook reference
```
latter-bolden, line 36:
```suggestion
```

latter-bolden (approved):
Code looks great!

jamesacklin (approved):
Looks good. Eventually I think we'll want to start stripping out the channel names themselves in favor of things like "SingleInline" or "SingleBlock." That's a discussion for another day, thanks for getting this in.

PR #3721, opened 2024-07-08
notify/chat: correct dm thread paths and prep for activity cutover

arthyn:
Fixes TLON-2269 by changing the paths we send to hark. Also includes the `@da` version of the activity ID that we send to Twilio so that we can use it instead of the old `@uv` version. Either version is supported in notify so this is just some cleanup.

PR Checklist


patosullivan (approved):

latter-bolden (approved):
This is working as expected while testing against a modified client app

PR #3724, opened 2024-07-09
native: notification fixes

latter-bolden:
- account for updated DM thread yarns
- send report to Crashlytics if we fail to navigate after clicking a push notif (with extra context if user is a Tlon employee)


latter-bolden:
This is good to merge once reviewed thanks to the latest internal deploy

jamesacklin:
@dnbrwstr all you my friend

dnbrwstr (approved):
Great, lgtm!

PR #3725, opened 2024-07-09
native: prevent URL and +code login for Tlon-hosted ships

jamesacklin:
Prevents users from logging in to the app with a Tlon-hosted ship URL and +code. This means we don't have to do suspension detection and revival in this login flow as well.

Also adds "Back" buttons to all login screens to allow the user to flip between methods without force-killing the app (!).

Fixes TLON-2285

image


jamesacklin:
Making this slightly more robust, please hold.

patosullivan (approved):
Nice! While you're in here, would you mind throwing a `.trim()` on `accessCode`? As of now (in production and the new app), if you accidentally add a space the login will fail.

jamesacklin:
@patosullivan Done, chef.

PR #3727, opened 2024-07-10
native: app settings screen

latter-bolden:
Adds _App Settings_ to the profile page and builds out the contained screens for unblocking users, managing your account, and viewing app info. Fixed the Horizon auth issue James and I were looking at and updated things to match the nav architecture in Figma. I can try to grab a screen recording in a bit.

Few lingering items but should otherwise be ready for review:
- need to fix logout borking things once you log back in
- should probably shuffle things around to follow the pattern of core display logic being in _ScreenView_ files in packages/ui

Fixes TLON-1722


latter-bolden:
@patosullivan i think the logout but is fixed, can you take another look whenever you have a chance?

PR #3729, opened 2024-07-10
native: support hiding/showing posts

patosullivan:
Fixes TLON-2236

Ideally we'd send along hidden status with the posts themselves when we fetch them, but at the moment we scry for them separately (and they're store in separate state in the agents).

It would also be good to get hidden posts added to our init scry.

Note that I also tried to make some existing code more readable as I went along.


packages/shared/src/api/postsApi.ts
```diff
@@ -102,23 +104,33 @@ export async function getPostReference({
...
+async function toPostReference(said: ub.Said) {
   const channelId = said.nest;
+  const hiddenPosts = await db.getHiddenPosts();
   if ('reply' in said.reference) {
     return toPostReplyData(
       channelId,
       said.reference.reply['id-post'],
-      said.reference.reply.reply
+      said.reference.reply.reply,
+      hiddenPosts.some(
```
dnbrwstr, line 122: Is this stuff in here to prevent overwriting the `hidden` field on posts? I'm wondering if we could skip doing it here, and just prevent `hidden` from being overwritten on insert (via `onConflictDoUpdate`). If that doesn't work, can we move the db wrangling either into sync or into the query?

packages/shared/src/api/postsApi.ts
```diff
@@ -597,9 +616,54 @@ export async function hidePost(channelId: string, postId: string) {
     },
   };
 
-  return await poke(action);
+  return poke(action);
+}
+
+export async function hideDMPost(authorId: string, postId: string) {
+  const writId = `${authorId}/${postId}`;
```
dnbrwstr, line 623: Can we collapse the type-specific methods like we do in `sendPost`?

packages/shared/src/db/schema.ts
```diff
@@ -725,6 +725,17 @@ export const posts = sqliteTable(
   })
 );
 
+export const hiddenPosts = sqliteTable(
```
dnbrwstr, line 728: Would it simplify things to just have an `is_hidden` field on the posts table? Or do we need this.

dnbrwstr (changes requested):
Looks good overall, but I'm wondering if we can simplify a bit.

packages/shared/src/db/schema.ts
patosullivan: I think we can just use the is_hidden field and make sure it doesn't get overwritten

latter-bolden (approved):
Fixes lgtm

PR #3730, opened 2024-07-10
activity: reduce transfer and quantity of updates

arthyn:
This updates activity to use a new set of types. We revert back to only send children as source strings. We also add a new type of update `%activity` which batches together the source and and parent updates. We also empty out children for base since everything can be considered it's children. Additionally we move threads out of the main activity scry into their own scry per channel/dm. This all should reduce data over the wire significantly. Fixes TLON-2097 mostly, but there are a few things that have not been implemented here, mainly around "debouncing" activity updates.

To fully use the types we had to refactor to the way that the mobile app does "sticky" unreads aka fixes TLON-2185. This made the unread management much cleaner and more resilient since now we always mark on the server as soon as you see it. 🫡 @latter-bolden 

Fixes TLON-2088 by adding a new action type `[%bump =source]` so that we can tell activity when we've last interacted with a source and that will set its recency to `now.bowl`

Finally since we were updating init anyway I added a few things that @patosullivan requested, hidden posts/messages and blocks for the new groups-ui init scry.

PR Checklist


desk/app/chat.hoon
```diff
@@ -1693,8 +1711,12 @@
         =.  recency.remark.dm  now.bowl
         =?  cor  &(!=(old-unread di-unread) !=(net.dm %invited))
           (give-unread ship/ship di-unread)
-        ?:  =(our.bowl author.memo)  (di-give-writs-diff diff)
         ?~  entry  (di-give-writs-diff diff)
+        =/  top-con  [id.writ.u.entry time.writ.u.entry]
+        =/  concern  [%reply [id.q.diff now.bowl] top-con]
+        =/  mention  (was-mentioned:utils content.memo our.bowl)
+        =.  di-core  (di-activity concern content.memo mention)
+        ?:  =(our.bowl author.memo)  (di-give-writs-diff diff)
```
Fang-, line 1719: Why was the "we are the author" conditional moved down (such that we do `+di-activity` unconditionally now)? We still want to not register activity for things we did ourselves, right?

desk/app/activity.hoon
```diff
@@ -424,21 +491,58 @@
...
-  (give %fact v0-paths v0-cage)
+  ~?  verbose  [update dist]
+  =?  cor  ?!(?=(%activity -.update))
+    =?  dist  ?=(%read -.update)  [%both /unreads]
+    =/  v0-paths
+      ?-  -.dist
+        %hose  ~[/ /v0 /v2]
+        %only  ~[path.dist [%v0 path.dist] [%v2 path.dist]]
+        %both  ~[/ /v0 /v2 path.dist [%v0 path.dist] [%v2 path.dist]]
+      ==
```
Fang-, line 523: For these and the others, to prevent flubs around path changes/inclusion etc, might be worth it to write the "both" case as being the `+weld` of `%hose` and `%only`. Probably not worth doing a proper loop, but something like the following would be fine.

```suggestion
    =/  v0-paths
      =*  hose  ~[/ /v0 /v2]
      =*  only  ~[path.dist [%v0 path.dist] [%v2 path.dist]]
      ?-  -.dist
        %hose  hose
        %only  only
        %both  (weld hose only)
      ==
```

Fang-: (Doing the version prefixes based on the kind of update being emitted here is a neat approach.)

desk/app/activity.hoon
```diff
@@ -502,8 +612,10 @@
   (refresh-index source index(stream new) |)
 ++  refresh-index
   |=  [=source:a new=index:a new-floor=?]
+  ~?  verbose  "refeshing index: {<source>}"
```
Fang-, line 615: You may consider prefixing all the verbose prints with `%{(trip dap.bowl)}`, for log reading clarity.

(Arguably at that point there's enough boilerplate to want to do a `+log` arm such that you `%-  (log "refreshing index")  rest-of-the-hoon`, but the `~?` are also fine for now.)

desk/app/activity.hoon
```diff
@@ -527,14 +639,31 @@
...
-    %channel  (refresh-summary [%group group.source])
-    %dm-thread  (refresh-summary [%dm whom.source])
-  ::
-      %thread
-    =.  cor  (refresh-summary [%channel channel.source group.source])
-    (refresh-summary [%group group.source])
-  ==
+  =.  cor
+    ?+  -.source  cor
+      %channel  (refresh-summary [%group group.source])
```
Fang-, line 644: A little bit surprised to see this wasn't using `+get-parents:src`. Could be worth refactoring lightly (now or later) so that we only ever get the source hierarchy from that utility function.

Fang- (commented):
Hoon-side changes look good for the most part! We already talked about most of the stuff here, so nothing too surprising. Some questions and notes below.

desk/app/chat.hoon
arthyn: because `di-activity` now handles the "self" case by bumping and not sending any activity

desk/app/activity.hoon
arthyn: Had to add a face to the `hose` case to do this, and weld wanted more type information, but I got it lol

desk/app/activity.hoon
```diff
@@ -115,6 +115,10 @@
 ++  emil  |=(caz=(list card) cor(cards (welp (flop caz) cards)))
 ++  give  |=(=gift:agent:gall (emit %give gift))
 ++  from-self  =(our src):bowl
+++  log
+  |=  msg=tape
+  ~?  verbose  "%{(trip dap.bowl)} {msg}"
+  same
```
Fang-, line 121: Consider:

```suggestion
++  log
  |=  msg=tape
  ?.  verbose  same
  (slog leaf+"%{(trip dap.bowl)} {msg}" ~)
```

Fang- (approved):
Changes look good, thank you!

Small suggestion about prettier logging, but not blocking.

patosullivan (approved):
Frontend code looks good, excited to get this in.

PR #3731, opened 2024-07-11
native: profile edit screen

latter-bolden:
Adds screen for editing your profile. Adds a new component that can be used in editing icon/cover images for both group and user profiles.

Fixes TLON-1721


jamesacklin:
Beautiful work, thank you!

Logging out and then back in with the same credentials still results in an unrecoverable state. Maybe we hide the log out button for today's demo build?

jamesacklin (approved):
.

latter-bolden:
This branched off the other PR before the logout fix went in. Should be resolved on `develop`

PR #3732, opened 2024-07-11
native: visual fix omnibus

jamesacklin:
- Updates the copy of WelcomeSheet (fixes TLON-2164)
- Brings gallery index view up to spec (fixes TLON-2251)
- Brings gallery detail views up to spec (fixes TLON-2161)
- Brings notebook index and detail views up to spec (fixes TLON-2162)

Does this by way of:

- Adding a check to usePostMeta() to determine if the text post is a non-image URL
- Adding a shortened text-only return to ContentRenderer to show preview text of a Notebook post in the channel
- Adding day suffixes to pretty dates (st, nd, rd, th)


packages/shared/src/logic/utils.ts
```diff
@@ -98,9 +99,29 @@ export function makePrettyDay(date: Date) {
...
+    case 1:
+      return 'st';
+    case 2:
+      return 'nd';
+    case 3:
+      return 'rd';
+    default:
+      return 'th';
+  }
+}
```
patosullivan, line 124: Why are we using this function? Doesn't `do` give us this? Is there some corner case we need covered that date-fns doesn't cover?

packages/ui/src/components/ChatMessage/ChatMessageReplySummary.tsx
```diff
@@ -10,9 +10,11 @@ export const ChatMessageReplySummary = React.memo(
   function ChatMessageReplySummary({
     post,
     onPress,
+    paddingLeft = true,
   }: {
     post: db.Post;
     onPress?: () => void;
+    paddingLeft?: boolean;
```
patosullivan, line 17: Might be good to leave a comment here explaining why this is necessary.

packages/ui/src/components/ContentReference/ChatReference.tsx
```diff
@@ -44,6 +44,7 @@ export default function ChatReference({
           <ContactName
             color="$tertiaryText"
             size="$s"
+            maxWidth={230}
```
patosullivan, line 47: Would probably be good to leave a comment explaining why this specific number is used.

packages/ui/src/components/ContentReference/GroupReference.tsx
```diff
@@ -24,14 +31,15 @@ export function GroupReference({ groupId }: { groupId: string }) {
...
-    <Reference onPress={onPress}>
+    <Reference onPress={onPress} viewMode={viewMode}>
       <Reference.Header>
         <Reference.Title>
           <ContactAvatar contactId={host} size="$xl" />
           <ContactName
             color="$tertiaryText"
             size="$s"
             userId={host}
+            maxWidth={230}
```
patosullivan, line 42: Same here

patosullivan: And if this is going to be used in a few places, maybe we define it in one place and import it where we need it.

packages/ui/src/components/GalleryPost/GalleryPost.tsx
```diff
@@ -64,11 +70,13 @@ export default function GalleryPost({
...
       ? 'image'
-      : isText && !isLinkedImage && !isRefInText
+      : isText && !isLinkedImage && !isRefInText && !isLink
         ? 'text'
         : isReference || isRefInText
           ? 'reference'
-          : 'unsupported';
+          : isLink
+            ? 'link'
+            : 'unsupported';
```
patosullivan, line 79: Would probably be good to rewrite this without the ternaries now that it's gotten so complex.

packages/ui/src/components/NotebookPost/NotebookPost.tsx
```diff
@@ -67,15 +64,14 @@ export default function NotebookPost({
             }}
             width="100%"
             height={smallImage ? IMAGE_HEIGHT / 2 : IMAGE_HEIGHT}
-            borderRadius="$m"
+            borderRadius="$s"
           />
         )}
         {post.title && (
           <Text
+            fontWeight={'500'}
```
patosullivan, line 72: Should use one of the tamagui tokens here, and if we don't have one maybe we should add one.

patosullivan (approved):
Looks good, just a few nits

packages/shared/src/logic/utils.ts
jamesacklin: Absolutely nothing. Thanks

patosullivan (approved):
Looks great!

PR #3733, opened 2024-07-11
wip: native: allow users to flag posts and admins to see the flag posts in activity

patosullivan:
For now, I've wired up the flagging from the user side and started on getting it to show up in the activity feed we return from %activity, but it's going to involve some more changes to the way the %all mode works for the feed. @Fang- is going to make those changes and then I'll wire up the rest of the frontend part if anything new needs to be done there (just need to navigate to the channel by pressing on the activity item).


dnbrwstr (approved):
Nice, lgtm!

PR #3734, opened 2024-07-11
channels & chat: mark activity for deleted msgs as read

Fang-:
Whenever we see a deletion, mark the corresponding activity event as read. It doesn't make sense to still leave that up to the client, since they won't be rendering those messages anymore.

Draft because we need to do this for chat also.

Closes TLON-2274.


arthyn:
so far so good

arthyn (approved):
lgtm ty!

PR #3735, opened 2024-07-12
web: activity tab using activity agent

arthyn:
Fixes TLON-2189 by changing notifications to now use the same source and (mostly the same) rendering as the mobile.

PR Checklist


apps/tlon-web/src/notifications/ActivitySummary.tsx
```diff
@@ -0,0 +1,215 @@
...
+    const message = `replied in a thread you're involved in`;
+    return (
+      <SummaryMessageWrapper>
+        {Authors}
+        {` ${message}`}
+      </SummaryMessageWrapper>
+    );
+  }
+
+  if (relevancy === 'postToChannel') {
```
patosullivan, line 170: We'll need to cover flag-post and flag-reply cases as well

patosullivan (changes requested):
Looks good, only missing piece, I think, is rendering flagged posts/replies.

desk/app/activity.hoon
```diff
@@ -857,8 +864,9 @@
   =/  volume  (get-volume:evt volume-settings -.event)
   ::TODO  support other event types
   =*  is-msg  ?=(?(%dm-post %dm-reply %post %reply) -<.event)
-  =*  supported
-    |(is-msg ?=(?(%dm-invite %chan-init) -<.event))
+  =*  inits  ?=(?(%dm-invite %chan-init) -<.event)
+  =*  flags  ?=(?(%flag-post %flag-reply) -<.event)
```
Fang-, line 868: Consider `is-init` and `is-flag` for consistency and clarity. (Arguably, `is-supported` too, but that one was already there.)

Fang- (commented):
Hoon side looks good! A nit, but not a blocker.

latter-bolden (approved):
New virtualization stuff lgtm!

jamesacklin:
Overall looks good and matches the content on mobile.

Clicking "Mark everything as read" unfortunately results in a spinner that hangs out forever with no meaningful dojo or js-console output. Is there something else I should be inspecting to see progress there?

image

PR #3738, opened 2024-07-15
native: add button for showing chat list filters

patosullivan:
Fixes TLON-2378 by adding a button to show the filters rather than relying entirely on the animation. Left the animation in since we may want to keep that functionality on iOS and fix the issue with Android later.


latter-bolden (approved):
Code lgtm

jamesacklin (approved):

PR #3744, opened 2024-07-15
native: update notifications to use instrumented endpoint

latter-bolden:
Updates the Swift and Java code to use the new endpoint with Mark's delivery instrumentation. 

Tested on iOS with posts and threads for groups and DMs. I don't have local testing setup for Android notifications, but will confirm it's working on that platform after merging.

Fixes TLON-2226


dnbrwstr (approved):
Cool, lgtm!

PR #3745, opened 2024-07-15
native: clear message input immediately on send

patosullivan:
We don't want to `await` the `send` because we need to clear the editor content and the draft immediately.

fixes TLON-2386


dnbrwstr (approved):
Lgtm!

PR #3747, opened 2024-07-16
native: dynamic channel welcome notice

davidisaaclee:
fixes TLON-1970

- Added `channels` sqlite columns: `last_viewed_at`, `is_default_welcome_channel`
- Only show welcome notice when all are true: (1) visiting the channel for the first time, and (2) we know that this is the channel we autocreated on this client during a group creation, and (3) user is admin

### DIFF
--- a/apps/tlon-mobile/src/screens/ChannelScreen.tsx
+++ b/apps/tlon-mobile/src/screens/ChannelScreen.tsx
@@ -29,6 +29,14 @@ export default function ChannelScreen(props: ChannelScreenProps) {
       }
     }, [props.route.params.channel.group])
   );
+  useFocusEffect(
+    useCallback(
+      () => () => {
+        store.markChannelVisited(props.route.params.channel);
+      },
+      [props.route.params.channel]
+    )
+  );
 
   const [channelNavOpen, setChannelNavOpen] = React.useState(false);
   const [currentChannelId, setCurrentChannelId] = React.useState(
--- a/apps/tlon-mobile/src/screens/ChannelScreen.tsx
+++ b/apps/tlon-mobile/src/screens/ChannelScreen.tsx
@@ -31,9 +31,11 @@ export default function ChannelScreen(props: ChannelScreenProps) {
   );
   useFocusEffect(
     useCallback(
-      () => () => {
-        store.markChannelVisited(props.route.params.channel);
-      },
+      () =>
+        // Mark the channel as visited when we unfocus/leave this screen
+        () => {
+          store.markChannelVisited(props.route.params.channel);
+        },
       [props.route.params.channel]
     )
   );
--- a/packages/shared/src/test/helpers.ts
+++ b/packages/shared/src/test/helpers.ts
@@ -2,7 +2,7 @@ import Database from 'better-sqlite3';
 import { BetterSQLite3Database, drizzle } from 'drizzle-orm/better-sqlite3';
 import { migrate } from 'drizzle-orm/better-sqlite3/migrator';
 import tmp from 'tmp';
-import { MockedFunction, beforeAll, beforeEach } from 'vitest';
+import { MockedFunction, beforeAll, beforeEach, vi } from 'vitest';
 
 import { scry } from '../api/urbit';
 import { setClient } from '../db';
@@ -47,13 +47,9 @@ export function setupDatabaseTestSuite() {
 }
 
 export function setScryOutput<T>(output: T) {
-  (scry as MockedFunction<() => Promise<T>>).mockImplementationOnce(
-    async () => output
-  );
+  vi.mocked(scry).mockImplementationOnce(async () => output);
 }
 
 export function setScryOutputs<T>(outputs: T[]) {
-  (scry as MockedFunction<() => Promise<T>>).mockImplementation(
-    async () => outputs.shift()!
-  );
+  vi.mocked(scry).mockImplementation(async () => outputs.shift()!);
 }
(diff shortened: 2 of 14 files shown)


davidisaaclee:
what kind of tests would you recommend adding? I looked but didn't find anything that I could extend to exercise this feature (I think I'd want a React-level test for checking that navigating away from a channel marks it as visited; and that going through `store.createGroup` ends up with a channel that's marked as the default welcome channel in SQLite).

apps/tlon-mobile/src/screens/ChannelScreen.tsx
```diff
@@ -29,6 +29,16 @@ export default function ChannelScreen(props: ChannelScreenProps) {
       }
     }, [props.route.params.channel.group])
   );
+  useFocusEffect(
+    useCallback(
+      () =>
+        // Mark the channel as visited when we unfocus/leave this screen
+        () => {
+          store.markChannelVisited(props.route.params.channel);
```
davidisaaclee, line 37: is this the only place that we'd need to mark a channel as visited? (on that note, can you think of a more specific term than "visited?" we are not considering any view other than the main `ChannelScreen` as a visit. I can't!)

dnbrwstr: Nope, that makes sense to me! `ChannelScreen` really is the only context in which you view a channel. And visited works for me, or `viewed` to match the sqlite column.

dnbrwstr (approved):
Lgtm! Notes re: tests below, mostly for the sake of information transfer: there's a good chance the setup lift for this feature at this moment isn't worth it, and it'd be nice to start getting into notifications, which is more critical. Up to you though!

> what kind of tests would you recommend adding? I looked but didn't find anything that I could extend to exercise this feature (I think I'd want a React-level test for checking that navigating away from a channel marks it as visited; and that going through `store.createGroup` ends up with a channel that's marked as the default welcome channel in SQLite).

We don't yet have react-level testing; I think that's ok to skip for now. I could see a test for `createGroup` that checks that the correct flags end up getting set. There are some structurally similar examples in `sync.test.ts` -- basically we've been using the `setScryOutputs` function to mock out requests we know we'll perform (for `createGroup`, that'd be the scries that get called via `syncGroup` and `syncUnreads`. For this application, you'd also need to mock out the `poke` or `trackedPoke` functions to handle the `api.create...` calls.

Also, vocab: 
- `scry` ~= `GET` request
- `poke` ~= `PUT` request
- `trackedPoke` ~= send `poke` and wait for a corresponding subscription event

jamesacklin:
Awesome, this looks great! Very easy to reason about. Thank you.

davidisaaclee:
@dnbrwstr thanks for review – I'm timeboxing adding a test now for education purposes.

davidisaaclee:
Cutting myself off of writing a test, as mocking out `subscribe` was taking too long; but I added some commits while starting that work:

- [ce2e2f5](https://github.com/tloncorp/tlon-apps/pull/3747/commits/ce2e2f5b931bf91918053b42d08c54806c15929a) – while looking through vitest docs, I saw code like this and thought it'd be a little nicer. adding comments inline...
- [7e031e3](https://github.com/tloncorp/tlon-apps/pull/3747/commits/7e031e30fe05708bd2ddeb07e17fa34c91e8b163) – when starting to write a new test file, I saw that this code was often repeated, so I factored it into a helper

packages/shared/src/test/helpers.ts
```diff
@@ -47,13 +47,9 @@ export function setupDatabaseTestSuite() {
...
-    async () => output
-  );
+  vi.mocked(scry).mockImplementationOnce(async () => output);
 }
 
 export function setScryOutputs<T>(outputs: T[]) {
-  (scry as MockedFunction<() => Promise<T>>).mockImplementation(
-    async () => outputs.shift()!
-  );
+  vi.mocked(scry).mockImplementation(async () => outputs.shift()!);
```
davidisaaclee, line 54: [`mocked`](https://vitest.dev/api/vi.html#vi-mocked) is a no-op but gives us the type we want (and preserves the signature of the mocked function)

packages/shared/src/test/setup.ts
```diff
@@ -20,10 +20,14 @@ vi.mock('@react-native-firebase/crashlytics', () => {
 });
 
 export function mockUrbit() {
-  vi.mock('../api/urbit', async () => {
-    return {
+  vi.mock('../api/urbit', async (importOriginal) => {
+    const mod = await importOriginal<typeof import('../api/urbit')>();
```
davidisaaclee, line 24: this form preserves the type of the original import, so mocked implementations here are sure to match the original ones.

it's notable that this changes behavior of tests a bit – since we splat `...mod`, we now include all of the original members of `../api/urbit`. This is different from the previous code, which would error if a test called a member of that module that was not mocked. I see the benefit of that approach, but I wasn't sure if it was intentional. I think we can restore that functionality by just deleting line 26. wdyt?

packages/shared/src/test/helpers.ts
```diff
@@ -2,7 +2,7 @@ import Database from 'better-sqlite3';
 import { BetterSQLite3Database, drizzle } from 'drizzle-orm/better-sqlite3';
 import { migrate } from 'drizzle-orm/better-sqlite3/migrator';
 import tmp from 'tmp';
-import { MockedFunction } from 'vitest';
+import { MockedFunction, beforeAll, beforeEach } from 'vitest';
```
davidisaaclee: I'm not 💯 on the impact of this on the production bundle (will vitest now get bundled?) – I figured we can fix this later if it's a problem

dnbrwstr: Yeah, I think this should be fine.

dnbrwstr (approved):
Nice, good stuff! Feel free to merge.

PR #3748, opened 2024-07-16
activity: only render logs if we need them

Fang-:
Classic case of rendering nouns that we won't print unless a flag is set. Noun renders can be quite slow for bigger nouns, and turns out we're printing a good handful of those. This should resolve the worst of the performance problems people were seeing.

(Activity still isn't fast by any means, summarizing unreads may still take over a second, but the impact of noun rendering was just egregious.)


arthyn (approved):
makes sense I had a tickle in the back of my mind about this but I wasn't sure

jamesacklin:
Any objections to merging? Suffering on wannec atm

PR #3750, opened 2024-07-16
native: reset image attachment after sending

latter-bolden:
Should fix the issue Pat reported where hitting send with an image attached left the image preview open even though the message _was_ actually sent. Believe this is related to the recent async changes there.

Fixes TLON-2392

### DIFF
--- a/apps/tlon-mobile/src/screens/ChannelScreen.tsx
+++ b/apps/tlon-mobile/src/screens/ChannelScreen.tsx
@@ -120,9 +120,7 @@ export default function ChannelScreen(props: ChannelScreenProps) {
 
       // clear the attachments immediately so consumers know the upload state is
       // no longer needed
-      setTimeout(() => {
-        uploadInfo.resetImageAttachment();
-      }, 20);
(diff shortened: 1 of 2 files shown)


dnbrwstr (commented):
Maybe missing something, but I think this is the same as what we do in `ChannelScreen` on line 128? I'm wondering what Pat was seeing was the send event failing to complete at all, possibly because of slowness yesterday. In either case, we probably don't want to await the send to clear the input.

latter-bolden:
@dnbrwstr yes it does the same thing as what's happening in _ChannelScreen_, I suppose if we bump that call above the await we can rely solely on that instead. Removing the `setTimeout` as well to make sure it feels responsive and doesn't delay itself a tick.

Can confirm it wasn't just send failure because of ship weirdness. It repros consistently, awaiting `store.sendPost` was only added yesterday. If you leave the input open after sending for long enough, it does eventually close once delivery of the message is confirmed.

dnbrwstr (approved):
Lgtm!

PR #3751, opened 2024-07-16
activity: only direct kids from +get-children

Fang-:
Previously, we were giving all sources as children of base. But this doesn't fit with our source hierarchy, and is not how `+get-children` is used in `/app/activity`.

Now, we only return `%group` and `%dm` sources as children of %base, and let the code iterate further on those for deeper processing.

With this change, we no longer do repeated processing of `%thread` and `%dm-thread` sources, resulting in (on ~palfun) a 20% speedup in unread summarization.

### DIFF
--- a/desk/lib/activity.hoon
+++ b/desk/lib/activity.hoon
@@ -19,17 +19,18 @@
       ==
     ==
   ::
-  ++  get-children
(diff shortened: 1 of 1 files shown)


arthyn (approved):
🫡

PR #3752, opened 2024-07-16
native: fix issue with non-inline content clearing on edit

patosullivan:
Fixes TLON-2338. We need to keep track of any attachments (images, references) to messages we're editing and make sure they're not overwritten when we insert the existing post content into the editor.


dnbrwstr (approved):
Cool, lgtm.

PR #3753, opened 2024-07-16
native: avoid erroneous logout after managing account

latter-bolden:
After you manage your account, we need to make sure you didn't delete it and log you out if you did. False positives (where you get logged out when you shouldn't) were happening too frequently, so this adds retry logic to avoid that situation.

Fixes TLON-2396


dnbrwstr (approved):
Nice, lgtm.

PR #3754, opened 2024-07-16
activity: reduce data transfer and processing further

arthyn:
Addresses the first part of TLON-2390 by removing `reads` and `children` from summaries sent to clients since none of them use those keys and they were a significant chunk of the data. Also reduces the amount of work done to summarize unreads and corrects base's events as being "children".

PR Checklist


Fang- (approved):
This looks like what you did on-call, and what I've already run with great success on ~palfun. The /lib/activity-json changes are probably superfluous here though. But: approved!

PR #3755, opened 2024-07-16
channel: don't unconditionally scry activity, check running

arthyn:
OTT, let's us suspend activity on hosts

PR Checklist


Fang- (commented):
We probably don't need the same scry in `+send` now anymore.

arthyn:
@Fang- we do because we use send elsewhere

Fang- (approved):
You're right. Once again outing myself as a blind man.

Good change!

PR #3756, opened 2024-07-16
native: fix lingering blue dots on activity items after clicking into them

latter-bolden:
This was happening because of a stale reference to the unread. Avoids this by way of a new `useLiveUnread` hook to avoid refetching the full activity query anytime unreads change. 

Fixes TLON-2234


patosullivan (approved):
Code looks good!

PR #3757, opened 2024-07-16
activity: stop using read items

arthyn:
We were previously not using individual item reads, but because we were trying to keep floors in sync between children and parents we were incurring huge costs from having to juggle lots of "read" items. This PR removes those from use and returns us to simply just moving floors.

Also included is the updated frontend types based on tloncorp/tlon-apps#3754

PR Checklist

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -647,13 +647,10 @@
   =/  =index:a  (~(gut by indices) source *index:a)
   =/  new=_stream.index
     (put:on-event:a stream.index time-id event)
-  (refresh-index source index(stream new) |)
+  (refresh-index source index(stream new))
 ++  refresh-index
-  |=  [=source:a new=index:a new-floor=?]
+  |=  [=source:a new=index:a]
   %-  (log |.("refeshing index: {<source>}"))
-  =?  new  new-floor
-    (update-reads:idx new)
-  %-  (log |.("new reads: {<reads.new>}"))
   =.  indices
     (~(put by indices) source new)
   ?:  importing  cor  ::NOTE  deferred until end of migration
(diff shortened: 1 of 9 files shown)


desk/app/activity.hoon
```diff
@@ -747,43 +711,21 @@
...
-    ::  since we're marking deeply we need to recursively read all
+    ::  if we're marking deeply we need to recursively read all
     ::  children
-    =.  cor
+    =?  cor  deep.action
       =/  children  (get-children:src indices source)
       |-
       ?~  children  cor
       =/  =source:a  i.children
       =.  cor  (read source action &)
```
Fang-, line 723: We don't use the `from-parent` arg anymore, right? Can take it out.

desk/app/activity.hoon
```diff
@@ -813,27 +755,20 @@
...
+    ::  all base's events are from children so we can ignore
+    ?:  ?=(%base -.source)  ~
+    ::  we don't need to take child events into account when summarizing
+    ::  the activity, so we filter them out
+    =-  +:-
+    %^  (dip:on-event:a @)  stream.index  ~
+    |=  [st=@ =time-id:a =event:a]
+    :_  [%.n st]
+    ?.  &(!child.event (gth time-id floor.reads.index))  ~
+    `event
```
Fang-, line 770: We use a weird construction with `+dip` here because the alternative is tapping, murning and gassing, like the old code did, right? Is this new implementation measurably faster?

Previously we used `+lot` to limit the amount of items we processed. Do we want to early-out once we reach a `time-id` that's beyond the `floor.reads.index`?

Fang- (commented):
So far so good.

desk/app/activity.hoon
```diff
@@ -861,7 +865,14 @@
   ::  with new reads, update our index and summary
   =.  cor  (refresh-index source index)
   =/  new=(unit activity-summary:a)  (~(get by activity) source)
-  ~?  !=(old new)  "%sync-reads: WARNING old and new summaries differ {<old>} {<new>}"
+  =/  old-sum  ?~(old ~ %=(u.old reads ~))
+  =/  new-sum  ?~(new ~ %=(u.new reads ~))
```
Fang-, line 869: Might be the first time I've seen short-form `%=` instead of the sugar. (^:

```suggestion
  =/  old-sum  ?~(old ~ u.old(reads ~))
  =/  new-sum  ?~(new ~ u.new(reads ~))
```

You might also consider moving these to be inline in the `~&` below, since that's the only place they're used. In general I'd be weary of introducing stuff into the subject ahead/outside of the one branch of the code where it gets used.

Fang- (approved):
Good changes. As discussed yesterday, we'll want to see this behave in practice before calling it good, but this feels like a slightly simpler model, and there's tests, so I'm cautiously optimistic.

PR #3758, opened 2024-07-17
native: add group members screen

patosullivan:
Fixes TLON-1718

Adds the group members screen, including a "Banned Users" section. Adds the ability to kick, ban, and unban users.

Refactors the ProfileSheet so we can use it in this context (and potentially give admins the ability to kick or ban directly from the channel view later).

Also creates a new "AppDataContext" that currently just takes contacts and currentUserId (optionally), but we should probably move more things into it. Open to ideas on what all should go in there, but I also feel like adding much more to it blows out the scope of this PR even more than I already have.

Also fixes TLON-2209


packages/shared/src/store/groupActions.ts
```diff
@@ -609,3 +609,186 @@ export async function deleteNavSection(group: db.Group, navSectionId: string) {
...
+    // rollback optimistic update
+    await db.addChatMembers({
+      chatId: groupId,
+      type: 'group',
+      contactIds: [contactId],
+    });
+  }
+}
+
+export async function getBannedUsers(groupId: string) {
```
latter-bolden, line 660: unused? If we want this, seems like it should be a db hook?

packages/ui/src/components/AddChats/AddDmSheet.tsx
```diff
@@ -64,33 +62,31 @@ export function StartDmSheet({
       <Sheet.Overlay />
       <Sheet.LazyFrame paddingTop="$s" paddingHorizontal="$2xl">
         <QueryClientProvider client={queryClient}>
-          <ContactsProvider contacts={contacts ?? null}>
```
latter-bolden, line 67: Confirmed removing the contact provider here breaks it on Android. Can't remember the exact specifics, but there's some bug with modal'd components not correctly getting context from further up in the tree

latter-bolden (changes requested):
Only blocking thing is the `AddDmSheet` being broken on Android. Code otherwise looks quite good — hero status for updating all those stale prop/context references

On the UX side, my only comment is that it was non-obvious that the members count on the group sheet was clickable

latter-bolden:
Got an Urbit error trying to ban someone, but it seems like it did somehow still go through? have you seen this at all?

packages/ui/src/components/AddChats/AddDmSheet.tsx
patosullivan: Oof, ok. Was wondering why it was there. Will add it back w/a comment.

patosullivan:
> Got an Urbit error trying to ban someone, but it seems like it did somehow still go through? have you seen this at all?
> 
> 

Hadn't seen that, the bans were going through for me without an error. What did the trace point to?

latter-bolden:
oh whoops, looks like cordon on that group was actually `%shut`. It was failing on an `%open` typecheck.

guess we probably wanna hide the button in that case? Curious that it did eventually resolve to being in the ban set, maybe a failed rollback?

packages/shared/src/store/groupActions.ts
patosullivan: I'll just remove

patosullivan:
> oh whoops, looks like cordon on that group was actually `%shut`. It was failing on an `%open` typecheck.
> 
> guess we probably wanna hide the button in that case? Curious that it did eventually resolve to being in the ban set, maybe a failed rollback?

Will take a look

latter-bolden (approved):
Looking good!

PR #3759, opened 2024-07-17
native: better emoji picker

latter-bolden:
I think this is at a good stopping place for now.

Fixes the sheet drag catching when you scroll the emojis and adds search. Should also be a step up in performance, but there's certainly more work to be done on that front.

Fixes TLON-2333

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -37,6 +37,7 @@
     "expo-blur": "^12.9.2",
     "expo-haptics": "^12.8.1",
     "expo-image-picker": "~14.7.1",
+    "fuse.js": "^7.0.0",
     "fuzzy": "^0.1.3",
     "lodash": "^4.17.21",
     "moti": "^0.28.1",
--- a/packages/ui/src/components/ChatMessage/ChatMessageActions/EmojiToolbar.tsx
+++ b/packages/ui/src/components/ChatMessage/ChatMessageActions/EmojiToolbar.tsx
@@ -36,11 +36,11 @@ export function EmojiToolbar({
 
   const lastShortCode =
     details.self.didReact &&
-    !['+1', 'heart', 'cyclone', 'seedling'].some((code) =>
+    !['+1', 'heart', 'laughing', 'cyclone'].some((code) =>
       details.self.value.includes(code)
     )
       ? details.self.value
-      : 'seedling';
+      : 'cyclone';
 
   return (
     <>
@@ -64,7 +64,7 @@ export function EmojiToolbar({
         />
         <EmojiToolbarButton
           details={details}
-          shortCode="cyclone"
+          shortCode="laughing"
           handlePress={handlePress}
         />
         <EmojiToolbarButton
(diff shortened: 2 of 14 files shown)


packages/ui/package.json
```diff
@@ -37,6 +37,7 @@
     "expo-blur": "^12.9.2",
     "expo-haptics": "^12.8.1",
     "expo-image-picker": "~14.7.1",
+    "fuse.js": "^7.0.0",
```
patosullivan, line 40: good idea, we should probably switch from fuzzy to fuse elsewhere. looks like fuzzy is no longer maintained.

patosullivan (approved):
Code looks good!

I know this is a different issue, but would you mind fixing the emoji size issue on android while you're in there?

Screenshot_1721304890

It shows up in the sheet, too.

packages/ui/src/components/Emoji/data.ts
```diff
@@ -1,14 +1,87 @@
...
   }
 }
+
+export function searchEmojis(query: string): EmojiObject[] {
+  if (!query) return Object.values(EMOJI_MAP);
+  return fuse.search(query).map((result) => result.item);
+}
+
+function processEmojiData() {
+  console.log('building emoji index');
```
patosullivan, line 53:
```suggestion
  console.log('building emoji index');
```

packages/ui/src/components/Emoji/EmojiPickerSheet.tsx
```diff
@@ -1,59 +1,154 @@
...
+  const [searchResults, setSearchResults] = useState<string[]>([]);
   const { onEmojiSelect, ...rest } = props;
+  const ALL_EMOJIS = usePreloadedEmojis();
+
+  const listData = useMemo(() => {
+    return query ? searchResults : ALL_EMOJIS;
+  }, [query, searchResults, ALL_EMOJIS]);
+
+  const handleQueryChange = useCallback((query: string) => {
+    console.log(`bl: setting query to ${query}`);
```
patosullivan, line 59:
```suggestion
```

packages/ui/src/components/Emoji/data.ts
latter-bolden: This one I want to leave in, should only be output once

packages/ui/package.json
patosullivan: (not in this PR tho)

packages/ui/src/components/ChatMessage/ChatMessageActions/EmojiToolbar.tsx
```diff
@@ -22,22 +22,25 @@ export function EmojiToolbar({
...
+      await Haptics.notificationAsync(Haptics.NotificationFeedbackType.Success);
+      setTimeout(() => onDismiss(), 50);
+    },
+    [currentUserId, details.self.didReact, details.self.value, onDismiss, post]
+  );
 
   const lastShortCode =
     details.self.didReact &&
-    !['+1', 'heart', 'cyclone', 'seedling'].some((code) =>
+    !['+1', 'heart', 'laughing', 'cyclone'].some((code) =>
```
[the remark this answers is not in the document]
latter-bolden: I figured we could bikeshed on the last slot during standup 😆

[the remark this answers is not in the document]
patosullivan: also doesn't really matter a whole lot since it will update based on what you've used before.

PR #3763, opened 2024-07-18
native: show 'delete message' when post has been deleted, clear out content in db

patosullivan:
Fixes TLON-2339

Also disables the ability to long press on the message once it is marked as deleted.

Looks like this in chat:

Simulator Screenshot - iPhone 15 Pro - 2024-07-18 at 10 40 56

### DIFF
--- a/packages/shared/src/db/queries.ts
+++ b/packages/shared/src/db/queries.ts
@@ -2055,6 +2055,22 @@ export const deletePost = createWriteQuery(
   ['posts']
 );
 
+export const markPostAsDeleted = createWriteQuery(
+  'markPostAsDeleted',
+  async (postId: string, ctx: QueryCtx) => {
+    return ctx.db
+      .update($posts)
+      .set({
+        isDeleted: true,
+        content: null,
+        textContent: null,
+        authorId: undefined,
+      })
+      .where(eq($posts.id, postId));
+  },
+  ['posts']
+);
+
 export const getPostReaction = createReadQuery(
   'getPostReaction',
   async (
--- a/packages/shared/src/store/postActions.ts
+++ b/packages/shared/src/store/postActions.ts
@@ -166,8 +166,11 @@ export async function showPost({ post }: { post: db.Post }) {
 }
 
 export async function deletePost({ post }: { post: db.Post }) {
+  const existingPost = await db.getPost({ postId: post.id });
+
   // optimistic update
-  await db.deletePost(post.id);
+  await db.markPostAsDeleted(post.id);
+  await db.updateChannel({ id: post.channelId, lastPostId: null });
 
   try {
     await api.deletePost(post.channelId, post.id);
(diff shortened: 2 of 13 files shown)


packages/shared/src/db/queries.ts
```diff
@@ -2055,6 +2055,22 @@ export const deletePost = createWriteQuery(
...
+export const markPostAsDeleted = createWriteQuery(
+  'markPostAsDeleted',
+  async (postId: string, ctx: QueryCtx) => {
+    return ctx.db
+      .update($posts)
+      .set({
+        isDeleted: true,
+        content: null,
+        textContent: null,
+        authorId: undefined,
```
latter-bolden, line 2067: nice touch

packages/shared/src/store/postActions.ts
```diff
@@ -166,16 +166,23 @@ export async function showPost({ post }: { post: db.Post }) {
 }
 
 export async function deletePost({ post }: { post: db.Post }) {
+  const existingPost = await db.getPost({ postId: post.id });
+
   // optimistic update
-  await db.deletePost(post.id);
+  await db.markPostAsDeleted(post.id);
+  await db.updateChannel({ id: post.channelId, lastPostId: null });
```
latter-bolden, line 173: trying to think if there's any weird side effects here. Guess you just lose the chatlist preview until it refetches heads or gets another new post?

packages/shared/src/store/sync.ts
```diff
@@ -579,7 +579,7 @@ export const handleChannelsUpdate = async (update: api.ChannelsUpdate) => {
       });
       break;
     case 'deletePost':
-      await db.deletePosts({ ids: [update.postId] });
+      await db.markPostAsDeleted(update.postId);
```
latter-bolden, line 582: would we want to clear the `lastPostId` here as well?

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -505,6 +505,33 @@ const BaseScrollerItem = ({
       <ChannelDivider unreadCount={0} post={post} index={index} />
     ) : null;
 
+  if (post.isDeleted) {
```
latter-bolden, line 508: could we make the press handlers no ops instead of duplicating the whole `View`? If not for some reason, maybe a dynamic wrapper component that we set above

latter-bolden (approved):
Lgtm, just a few nits

jamesacklin:
Can we get a timestamp and italicized content to match the web treatment?

Screenshot 2024-07-18 at 1 29 11 PM

packages/shared/src/store/postActions.ts
patosullivan: I think so, yeah.

patosullivan:
> Can we get a timestamp and italicized content to match the web treatment?
> 
> Screenshot 2024-07-18 at 1 29 11 PM

We never show timestamps in that spot on native, we only show it next to the author name (on the right rather than the left). Should we special case this?

patosullivan:
@jamesacklin about like this?
Simulator Screenshot - iPhone 15 Pro - 2024-07-18 at 13 00 55

jamesacklin:
Visually approved.

PR #3767, opened 2024-07-18
native: fix searching within filters on chat list

patosullivan:
Fixes TLON-2388 by:

- Making the search more performant using fuse.js
- Making sure the search re-runs when we change tabs
- Using a single `displayData` object for data in the section list.


latter-bolden (approved):
Looks good!

Might want this extracted into a hook in its own file, but that doesn't need to happen right this minute.

PR #3771, opened 2024-07-18
activity: don't mark read if tab not focused and automatically mark read at chat bottom

arthyn:
This didn't have an issue, but it was old behavior we had that I think is better than what's currently in there. Basically there's three situations:
- tab is out of focus, so don't do any read marking and let them accumulate
- tab is in focus, and we're at the bottom of the channel, automatically mark read w/o showing markers
- tab is in focus, but we're not at the bottom of the channel, don't automatically mark read, let the normal "seeing" trigger mark read

PR Checklist

### DIFF
--- a/apps/tlon-web/src/app.tsx
+++ b/apps/tlon-web/src/app.tsx
@@ -678,6 +678,23 @@ function RoutedApp() {
   const isDarkMode = useIsDark();
 
   useEffect(() => {
+    const onFocus = () => {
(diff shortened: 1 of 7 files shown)


jamesacklin:
Cool, this makes sense. No need to flash the blue separator if you're actively watching the channel in focus.

patosullivan (approved):
Looks good, makes sense.

PR #3774, opened 2024-07-19
native: batch add post sub updates

latter-bolden:
Trying to prevent that interaction where you open the app after a long while and see the last message update iteratively as it processes the events you missed.


dnbrwstr:
This is a great idea -- I wonder if we can push some of the complexity into `api/urbit` subscribe method? And throttle there. Could also be worth looking into RxJS (or something similar, but more lightweight) to manage these event streams.

PR #3775, opened 2024-07-19
native: a few ChatList fixes

latter-bolden:
- pad the bottom so no groups at the end of the list are obscured by the bottom nav
- stop refetching unreads anytime you return to the chatlist. Instead, only rerun that scry defensively whenever you foreground the app


dnbrwstr (approved):
Lgtm

PR #3777, opened 2024-07-19
native: allow user to retry or delete posts that failed to send

patosullivan:
Fixes TLON-2407 by notifying a user when the post failed to send and allowing them to tap the post that will show a sheet which allows them to retry or delete the post.

### DIFF
--- a/packages/ui/src/components/ChatMessage/ChatMessage.tsx
+++ b/packages/ui/src/components/ChatMessage/ChatMessage.tsx
@@ -2,10 +2,12 @@ import { utils } from '@tloncorp/shared';
 import * as db from '@tloncorp/shared/dist/db';
 import { Story } from '@tloncorp/shared/dist/urbit';
 import { isEqual } from 'lodash';
-import { memo, useCallback, useMemo } from 'react';
+import { memo, useCallback, useMemo, useState } from 'react';
 
 import { Text, View, XStack, YStack } from '../../core';
+import { ActionSheet } from '../ActionSheet';
 import AuthorRow from '../AuthorRow';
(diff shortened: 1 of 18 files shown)


packages/ui/src/components/ChatMessage/ChatMessage.tsx
```diff
@@ -186,6 +215,15 @@ const ChatMessage = ({
       post.replyContactIds ? (
         <ChatMessageReplySummary post={post} onPress={handleRepliesPressed} />
       ) : null}
+      <ActionSheet open={showRetrySheet} onOpenChange={setShowRetrySheet}>
```
latter-bolden, line 218: interesting that we're including a sheet with every post, did you notice any performance impact? Wondering if it might be worth pulling up and handling as a singleton.

May be fine as is, just know we've run into issues with rendering actions per item in a list before on web

latter-bolden (approved):
Code looks good to me! 

One question around approach, but otherwise g2g

packages/ui/src/components/ChatMessage/ChatMessage.tsx
patosullivan: Haven't noticed any performance impact on android or iOS, but I'll run the profiler and see if anything pops out.

latter-bolden: Feel free to merge if it seems fine

patosullivan: After opening channels, threads, and scrolling around, not seeing any perf issues in the profiler.

PR #3779, opened 2024-07-22
native: fix joined channel detection

latter-bolden:
After moving to activity v4, there was an issue with the partial unread updates we get from activity causing a full reconfiguration of which channels we're joined to.

This just adds an optional flag to specify when you do and do not have the full set of unreads and accordingly branches on that when deciding to update channel joins.

Fixes TLON-2420


patosullivan (approved):
Code looks good, tried it in the sim and it works.

PR #3780, opened 2024-07-22
activity: web match mobile bell tab

arthyn:
Fixes TLON-2419 by mimicking what mobile does, basically clearing the bell any time you visit and bringing it back upon new relevant activity. This also fixes optimistic reads which were not happening because we defaulted `variables.action` in the mutation function and didn't account for that later on.

Button moved to settings but not sure what language we want here:
image

PR Checklist


arthyn:
diff is coming up weird because it's including the base branch changes for some reason

latter-bolden (approved):
Makes sense to me, works locally.

PR #3781, opened 2024-07-22
native: handle failed authentication

latter-bolden:
We now check the response for the authentication request. Bad access code returns `400`, so we just look to see that the response falls within the `200`s. The throw causes red error text to be displayed on the login screen and the user can try again.

Fixes TLON-2262


dnbrwstr (approved):
Beautiful

PR #3787, opened 2024-07-23
misc: cleanup activity layout and fix groups preview fetch

arthyn:
We had some weird spacing on mobile here so I removed some of the padding to help it have more room. Also found the issue with group preview fetching (although I still think there are other problems with it). Also fixed an issue where the author row wouldn't show up after a notice.

PR Checklist


patosullivan (approved):
Looks good

PR #3788, opened 2024-07-23
native: fix issue with scrolling to refs in the same channel

patosullivan:
fixes TLON-2413 by handling a situation where the anchor changes after we've mounted the component.

In working on this, I also found a new bug on develop where following a ref to a different chat channel, where the referenced message is far enough back in the back log that we'll see it on the initial render of the scroller *but* the after we load new posts in the scroller jumps to another spot and it's difficult to find the post you were looking for.

I thought I had introduced that issue here, but I verified that it exists in develop. I logged that issue in triage.


dnbrwstr (changes requested):
Afaict, this doesn't work unless the reffed post has already been loaded? I think the easiest way to get this to work right is going to be to force a full refresh for refs of posts that we haven't loaded in yet -- basically we need the same behaviors as when loading things in for the first time, including resetting the cursor, `userHasScrolled`, etc.

dnbrwstr (changes requested):
This is looking good, really just one remaining issue:

Running locally, I noticed that in some cases `scrollToIndexFailed` gets called when attempting to scroll to a post farther back in the scrollback. I think it's triggered by jumpiness, and then `scrollToOffset` causes more stuff to load in, causing more jumpiness. 

To fix this we can either:
- Prevent `scrollToOffset` being called when we're refreshing, or 
- Remove it entirely and make sure that we only call `scrollToIndex` when the message has actually been rendered (using `viewableItems`?).

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -50,13 +51,14 @@ type RenderItemFunction = (props: {
...
   onPressDelete: (post: db.Post) => void;
+  isHighlighted?: boolean;
 }) => ReactElement | null;
 
 type RenderItemType =
   | RenderItemFunction
   | React.MemoExoticComponent<RenderItemFunction>;
 
-const logger = createDevLogger('scroller', false);
+const logger = createDevLogger('scroller', true);
```
dnbrwstr, line 61:
```suggestion
const logger = createDevLogger('scroller', false);
```

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -72,375 +74,419 @@ export type ScrollAnchor = {
...
-    console.log('scroll to index failed');
-  }, []);
+    const handleScrollToIndexFailed = useCallback(
+      (info: {
+        index: number;
+        highestMeasuredFrameIndex: number;
+        averageItemLength: number;
+      }) => {
+        logger.log('scroll to index failed', { info });
+        // setShouldScrollToAnchor(false);
```
dnbrwstr, line 300:
```suggestion
```

dnbrwstr (approved):
Nice, lgtm!

PR #3789, opened 2024-07-23
activity: fix bump hiding unreads

arthyn:
OTT, this was happening occasionally when you reply to a thread and then the OP or someone else responds. This was putting an invalid source in our activity map which was overriding the activity sent to the frontend making it look "read".

PR Checklist


Fang- (approved):
"Yeah sure it's fine", approved. (;

PR #3792, opened 2024-07-23
native: add DM Lure links

jamesacklin:
Currently feature-flagged behind a switch in PostHog.

- Puts the Branch.io URL and key into .env (also set in Expo secrets)
- Reimplements the branch.ts util in the tlon-mobile source tree
- Adds a call in ProfileScreen out to branch to fetch or create the user's unique link
- Passes this down into a feature-flagged React.Share() button, which utilizes the system share sheet

https://github.com/user-attachments/assets/7911d08a-c508-468f-a977-72594f9b7f0d


jamesacklin:
I'll open the release aperture for the feature flag so we can test this, then close afterward.

apps/tlon-mobile/src/utils/branch.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (!fallbackUrl || !path) {
+    return undefined;
+  }
+
+  if (type === 'lure' && !urbit.whomIsFlag(path)) {
+    return undefined;
+  }
+
+  if (type === 'wer') {
+    const [location, ship] = path.split('/');
```
dnbrwstr, line 58: I think we could use `api.getPostInfoFromWer` for this.

apps/tlon-mobile/src/screens/ProfileScreen.tsx
```diff
@@ -23,6 +24,16 @@ export default function ProfileScreen(props: Props) {
     props.navigation.navigate('EditProfile');
   }, [props.navigation]);
 
+  const [dmLink, setDmLink] = useState('');
+
+  useEffect(() => {
```
dnbrwstr, line 29: Maybe this goes in `utils/branch` as `useDmLink`?

apps/tlon-mobile/src/utils/branch.ts
dnbrwstr, line 1: I think this file could go in `shared`, either under `api` or under `logic`.

dnbrwstr (approved):
Overall looks good! Some minor organizational nits.

apps/tlon-mobile/src/utils/branch.ts
jamesacklin: Cool. Since we can't use a hook in a normal JS function, is it still OK to pass the ship in?

apps/tlon-mobile/src/screens/ProfileScreen.tsx
jamesacklin: Totally. Let me move.

apps/tlon-mobile/src/utils/branch.ts
jamesacklin: Well, that's handy :)

apps/tlon-mobile/src/utils/branch.ts
dnbrwstr: Ya, that sounds fine.

dnbrwstr (approved):
Nice, lgtm!

PR #3793, opened 2024-07-23
native: a few fixes

latter-bolden:
Bit of a hodgepodge to start knocking out some of the smaller _Production Readiness_ items.

- removes unused _Pin Post_ notebook action
- prevents showing _Group Preview_ sheet for DMs
- Closes _Group Preview_ sheet upon clicking leave

Also found a bug where we let you try to leave groups you yourself host. Added logic to avoid that and tacked on `hostUserId` and `currentUserIsHost` to the groups table.

Fixes TLON-2427
Fixes TLON-2321
Fixes TLON-2428

### DIFF
--- a/packages/ui/src/components/GroupOptionsSheet.tsx
+++ b/packages/ui/src/components/GroupOptionsSheet.tsx
@@ -74,13 +74,29 @@ export function ChatOptionsSheet({
     [currentUser, groupData?.members]
   );
 
-  const adminActions = useMemo(
-    () => [
+  const actions = useMemo(() => {
+    const actions = [];
+    actions.push(
       {
+        title: 'Copy group reference',
+        action: () => {},
+        icon: 'ArrowRef',
+      },
+      {
+        title: isPinned ? 'Unpin' : 'Pin',
+        action: () => {},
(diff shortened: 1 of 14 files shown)


packages/ui/src/components/GroupOptionsSheet.tsx
```diff
@@ -74,13 +74,29 @@ export function ChatOptionsSheet({
...
+    actions.push(
       {
+        title: 'Copy group reference',
+        action: () => {},
+        icon: 'ArrowRef',
+      },
+      {
+        title: isPinned ? 'Unpin' : 'Pin',
+        action: () => {},
+        icon: 'Pin',
```
patosullivan, line 88: Wanna hook this up with store.pinItem and store.unpinItem?

patosullivan (approved):
Looks good. One note about hooking up pin/unpin, but we can just do that later if you wanna.

packages/ui/src/components/GroupOptionsSheet.tsx
latter-bolden: Good call, wired up pin and copy ref

PR #3795, opened 2024-07-25
web: fix routing from group reference to group preview modal

latter-bolden:
This was an issue with navigation logic. The path in our router for modals requires `location.state.backgroundLocation` to be present. There was a bug in a navigation helper that causes that prop not to get set _if_  there's pre-existing location state. When navigating to the gallery detail view, we set the preloaded block to `location.state` causing that bugged conditional to override the prop we care about. This just tweaks the logic to avoid that override. 

After tracing around a bit, I'm still unsure why this is only happening now. Seemed like none of the chokepoints have been touched in the last ~3 months.

Also fixes a tiny issue I noticed while testing. If you click a group reference from a DM, it was failing to navigate if you're already joined.

Fixes TLON-2441
Fixes TLON-2445

### DIFF
--- a/apps/tlon-web/src/app.tsx
+++ b/apps/tlon-web/src/app.tsx
@@ -163,11 +163,13 @@ interface RoutesProps {
   isSmall: boolean;
 }
 
-const GroupsRoutes = React.memo(({ isMobile, isSmall }: RoutesProps) => {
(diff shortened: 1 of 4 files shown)


patosullivan (approved):
Good catch!

PR #3796, opened 2024-07-25
native: access GroupOptions from channel list

jamesacklin:
- Recycles GroupOptionsSheet from ChatList in GroupChannelsScreen. I had to do a fair amount of click-handler-passing, so the diff looks scarier than it is. I also needed to get the current group in scope from navigation, rather than relying on long-press (user selection). Hopefully this makes adding a new channel to a group slightly less insane.
- Pads the ChannelFooter up for bauble users; fixes TLON-2125
- Adds an icon for the "Leave group" option in GroupOptionsSheet
- Makes the button to view members in GroupOptionsSheet look like a button

For **non-bauble users,** the UI now looks like this:
Screenshot 2024-07-25 at 2 27 21 PM
Screenshot 2024-07-25 at 2 31 21 PM

### DIFF
--- a/apps/tlon-mobile/src/screens/GroupChannelsScreen.tsx
+++ b/apps/tlon-mobile/src/screens/GroupChannelsScreen.tsx
@@ -1,10 +1,13 @@
+import { useIsFocused } from '@react-navigation/native';
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
 import * as db from '@tloncorp/shared/dist/db';
 import * as store from '@tloncorp/shared/dist/store';
 import { AppDataContextProvider, GroupChannelsScreenView } from '@tloncorp/ui';
-import { useCallback } from 'react';
+import { useCallback, useMemo } from 'react';
 
-import type { RootStackParamList } from '../types';
+import { useCurrentUserId } from '../hooks/useCurrentUser';
+import type { GroupSettingsStackParamList, RootStackParamList } from '../types';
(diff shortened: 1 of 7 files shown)


latter-bolden (approved):
Works locally!

jamesacklin:
@dnbrwstr ping-ping? Would be nice for today

apps/tlon-mobile/src/screens/GroupChannelsScreen.tsx
```diff
@@ -31,13 +44,84 @@ export function GroupChannelsScreen({
 
   const contactsQuery = store.useContacts();
 
+  const { leaveGroup, togglePinned } = useGroupContext({
+    groupId: groupParam.id,
+  });
+
+  const navigateToGroupSettings = useCallback(
```
dnbrwstr, line 51: Could this stuff go in a context, rather than getting drilled down? `NavigationContext` or `AppDataContext` could both be good candidates. `AppDataContext` already includes the `currentUserId`.

dnbrwstr (approved):
Overall looks good, just a q about context; I'm trying to tamp down prop drilling a little.

apps/tlon-mobile/src/screens/GroupChannelsScreen.tsx
jamesacklin: Absolutely, let me refactor later today.

PR #3797, opened 2024-07-25
activity: fix init unreads

arthyn:
Fixes TLON-2447 by adding a new state transition that will run a fix. The fix looks for any DM or channel sources with only one event which should be either `%dm-invite` or `%chan-init` and sets the floor to the `last-read` time from the old agents. If this is actually an invite pending, then it will stay unread which is what we want. If it's not pending and they had in fact previously interacted and read that channel, then it will be marked read.

Unfortunately, to call these migration methods I had to factor them out to be pure so that we could call them on old state. Sync reads tests still pass, going to write some for this fix scenario.

PR Checklist

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -148,13 +148,7 @@
         state-1
     ==
   +$  state-6  current-state
-  +$  state-5
-    $:  %5
-        allowed=notifications-allowed:a
-        =indices:a
-        =activity:a
-        =volume-settings:a
-    ==
+  +$  state-5  _%*(. *state-6 - %5)
   ++  state-5-to-6
     |=  old=state-5
     ^-  state-6
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -927,7 +927,7 @@
       %gx  (scry-path %chat /full/noun)
     ==
   =.  indices  (fix-dm-init-unreads indices dms clubs)
-  cor
+  refresh-all-summaries
 ++  fix-channel-init-unreads
   |=  [=indices:a =channels:c]
   %-  ~(urn by indices)
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -48,7 +48,7 @@
   +$  card  card:agent:gall
   ::
   +$  current-state
-    $:  %5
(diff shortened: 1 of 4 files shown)


desk/app/activity.hoon
```diff
@@ -135,11 +135,37 @@
...
+        state-1
+    ==
+  +$  state-6  current-state
+  +$  state-5
+    $:  %5
+        allowed=notifications-allowed:a
+        =indices:a
+        =activity:a
+        =volume-settings:a
+    ==
```
Fang-, line 157: Might be able to say something like this. Or too cute? Does make it clear that nothing besides the tag changed.

```suggestion
  +$  state-5  _%*(. *state-6 - %5)
```

desk/app/activity.hoon
```diff
@@ -797,102 +826,64 @@
...
+  ?~  channel=(~(get by channels) nest.source)  index
+  index(floor.reads last-read.remark.u.channel)
+++  fix-dm-init-unreads
+  |=  [=indices:a dms=(map ship dm:ch) clubs=(map id:club:ch club:ch)]
+  ^-  indices:a
+  %-  ~(urn by indices)
+  |=  [=source:a =index:a]
+  ::  if we're a DM with only the %dm-invite event, we need to set
+  ::  the floor to the last-read time
+  ~!  stream.index
```
Fang-, line 864: Straggler.

```suggestion
```

desk/app/activity.hoon
```diff
@@ -908,13 +899,19 @@
     [~ %.n ?:((gth time-id st) time-id st)]
   =.  reads.index  [new-floor ~]
   ::  with new reads, update our index and summary
-  =.  cor  (refresh-index source index)
-  =/  new=(unit activity-summary:a)  (~(get by activity) source)
-  ?:  !=(?~(old ~ u.old(reads ~)) ?~(new ~ u.new(reads ~)))
+  :: =.  cor  (refresh-index source index)
```
Fang-, line 902: Straggler.

```suggestion
```

Fang- (approved):
This looks like what we reviewed out of band yesterday. Nothing serious, just some small clean-up things.

PR #3798, opened 2024-07-25
native: search fixes

latter-bolden:
- display normal chat message affordances in search results (contact info, date, etc.)
- include and route to thread search results
- make search results clickable again
- fix `selectedPostId` updates not carrying through to the `useChannelPosts` hook
- improve rendering performance for the search result list

There's more optimization that could be done around avoiding jitter in the results, but anecdotally it's less noticeable for real world searches. The scroller thrashing while anchoring on the selected post after clicking remains the dominating pain point, as is the case elsewhere.

Fixes TLON-2052

### DIFF
--- a/apps/tlon-mobile/src/screens/ChannelSearchScreen.tsx
+++ b/apps/tlon-mobile/src/screens/ChannelSearchScreen.tsx
@@ -9,7 +9,6 @@ import {
   SearchResults,
   XStack,
   YStack,
-  useContacts,
 } from '@tloncorp/ui';
 import { useCallback, useState } from 'react';
 import { SafeAreaView } from 'react-native-safe-area-context';
--- a/apps/tlon-mobile/src/screens/ChannelSearchScreen.tsx
+++ b/apps/tlon-mobile/src/screens/ChannelSearchScreen.tsx
@@ -35,10 +35,20 @@ export default function ChannelSearch({
 
   const navigateToPost = useCallback(
     (post: db.Post) => {
-      navigation.navigate('Channel', {
-        channel,
-        selectedPostId: post.id,
-      });
+      if (post.parentId) {
+        navigation.replace('Post', {
+          post: {
+            id: post.parentId,
+            channelId: channel.id,
+            authorId: post.authorId,
+          },
+        });
+      } else {
+        navigation.navigate('Channel', {
+          channel,
+          selectedPostId: post.id,
+        });
+      }
     },
     [channel, navigation]
   );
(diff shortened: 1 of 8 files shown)


jamesacklin:
Would also be good to review + get in today.

apps/tlon-mobile/src/screens/ChannelSearchScreen.tsx
```diff
@@ -16,49 +26,66 @@ export default function ChannelSearch({
...
-        <XStack gap="$l">
-          <SearchBar
-            onChangeQuery={setQuery}
-            placeholder={`Search ${channel.title}`}
-          />
-          <Button minimal onPress={() => navigation.pop()}>
-            <Button.Text>Cancel</Button.Text>
-          </Button>
-        </XStack>
+      <AppDataContextProvider
```
dnbrwstr, line 58: Nice, like seeing the provider moved to this side of the ui boundary.

packages/ui/src/components/AuthorRow.tsx
```diff
@@ -50,13 +55,25 @@ export default function AuthorRow({ onPress, ...props }: AuthorRowProps) {
   return (
     <>
       {props.detailView ? (
-        <DetailViewAuthorRow {...props} onPress={handlePress} />
+        <DetailViewAuthorRow
+          {...props}
+          onPress={nonPressable ? undefined : handlePress}
```
dnbrwstr, line 60: Would just setting a `disabled` prop on `AuthorRow` allow us to skip this prop + the duplication here? Also, depending on what you're trying to achieve `pointerEvents="none" ` would probably work.

packages/ui/src/components/ChannelSearch/SearchResults.tsx
```diff
@@ -23,10 +23,37 @@ export function SearchResults({
...
     }
-  }, [search.loadMore, search.loading, search.hasMore]);
+  }, [search]);
 
   const isInitial = search.query === '';
 
+  const renderItem = useCallback(
+    ({ item: post }: { item: db.Post }) => {
+      return (
+        <View
```
dnbrwstr, line 33: Probably want to break out this component and memoize the item and its function props. Also, the double handlers here tell me something funky is going on here -- would suggest disabled or pointerEvents none like above if that's what you're trying to achieve.

packages/ui/src/components/ChatMessage/ChatMessage.tsx
```diff
@@ -74,6 +78,18 @@ const ChatMessage = ({
     onPressReplies?.(post);
   }, [onPressReplies, post]);
 
+  const shouldHandlePress = useMemo(() => {
```
dnbrwstr, line 81: Same comments, just wondering if we can simplify.

dnbrwstr (approved):
Overall looks good, just nitpicking about press handler management -- would be nice if we had a standard/clean way of working with it.

PR #3801, opened 2024-07-26
activity: use unreads to determine which sources should change

arthyn:
This changes the init fix to use unreads. It turns out the last-read/recency comparisons we were doing are invalid because of the way the old agents calculate unreads, so instead we just the unreads themselves.

PR Checklist


arthyn:
going to have to ask forgiveness for this one

PR #3803, opened 2024-07-29
native: more search fixes

latter-bolden:
A couple lingering search issues from the PR last week:

- Fixes search for DM's (1:1 and group)
- Resolves the issue with message presses overeagerly navigating you to the thread
  - restores previous behavior of needing to click the reply summary or long press to start a new thread


dnbrwstr (approved):
Nice, lgtm.

PR #3806, opened 2024-07-30
activity: allow showing reply deletes and make sure deletes don't cause unreads

arthyn:
We add a new set of types so that replies can be null so we have a marker in the chat. We previously were relying on this to allow people to mark deleted messages read, however in chat a much larger refactor would have to occur to allow the same system. So instead we've created a new action `%del-event` which allows deleting an activity event from the index. Doing so will refresh the activity summary and make it look as if the message never happened.

Fixes other cases of TLON-2274 that weren't captured

Fixes TLON-2274

PR Checklist


desk/app/activity.hoon
```diff
@@ -696,10 +697,39 @@
...
+  ?~  events  cor
+  =/  new-index  index(stream +:(del:on-event:a stream.index time-id:(head events)))
+  =.  indices
+    (~(put by indices) source new-index)
+  =.  cor  (refresh source)
+  =/  new-activity=activity:a
+    %+  roll
+      (snoc (get-parents:src source) source)
+    |=  [=source:a out=activity:a]
+    (~(put by out) source (~(gut by activity) source *activity-summary:a))
```
Fang-, line 730: Can this not be a `got:by`? We _just_ refreshed this source, and so we also refreshed its parents, right?

desk/app/chat.hoon
```diff
@@ -779,18 +779,19 @@
...
           ==
           content=story:d
           mention=?
       ==
   ^+  cor
+  ?.  .^(? %gu (scry-path %activity /$))
+    cor
   =;  actions=(list action)
-    ?.  .^(? %gu (scry-path %activity /$))
-      cor
```
Fang-, line 793: Hah, better. Thank you.

desk/app/chat.hoon
```diff
@@ -1345,7 +1357,9 @@
       ::
           %del
         =?  cu-core  ?=(^ had)
-          (cu-activity [%delete p.diff.delta time.u.had] *story:d |)
+          =*  content  content.writ.u.had
```
Fang-, line 1360: If we ever add edits to chat this might start failing, but for now we should consider ourselves lucky...

Fang- (approved):
Some notes, but nothing blocking. Mostly about 9b3ba5f, everything else we already worked on together. Thanks!

desk/app/chat.hoon
arthyn: ha

PR #3808, opened 2024-07-30
native: sync start fixes

latter-bolden:
- no more flashing pins, no more invalid pins model column name
- only refetch pins when app is foregrounded
- move foregrounded actions to only run after first initialization (i.e. subsequent app opens)
- latest posts show up consistently (were missing after last sync changes on fresh download or logout/login)
- fixed ordering for `syncInit` db writes
- concurrent fetches for `init` and `latestPosts` while enforcing init always gets written first
- fix straggler bug with setting joined channels
- add sample heads data and test for syncing latest posts

Fixes TLON-2444


packages/shared/src/store/sync.test.ts
```diff
@@ -329,20 +338,18 @@ test('syncs init data', async () => {
...
-  //   //   ['heap/~bolbex-fogdys/design-1761'],
-  //   //   ['chat/~bitpyx-dildus/interface'],
-  //   //   ['chat/~bolbex-fogdys/ops'],
-  //   //   ['heap/~dabben-larbet/fanmail-3976'],
-  //   //   ['diary/~bolbex-fogdys/bulletins'],
-  //   //   ['chat/~nocsyx-lassul/bongtable'],
-  //   // ]);
+});
+
+test('syncs last posts', async () => {
```
dnbrwstr, line 343: 😍

packages/shared/src/store/sync.ts
```diff
@@ -17,7 +17,7 @@ import { addToChannelPosts, clearChannelPostsQueries } from './useChannelPosts';
 
 export { SyncPriority, syncQueue } from './syncQueue';
 
-const logger = createDevLogger('sync', false);
+const logger = createDevLogger('sync', true);
```
dnbrwstr, line 20:
```suggestion
const logger = createDevLogger('sync', false);
```

dnbrwstr (approved):
Nice, lgtm!

PR #3809, opened 2024-07-30
ops: getting started edits

latter-bolden:
Just a few small edits to help with getting started on the mobile repo:
- clarify what's needed in sample .env
- fix instructions for dev package signing on android
- troubleshooting for Android version mismatch fix


arthyn (approved):
lgtm

PR #3810, opened 2024-07-31
docket-proxy: mitm docket to inject custom headers

Fang-:
Some of the upcoming frontend code needs access to functionality the browser only enables in the presence of specific headers on the responses serving the source code files.

Docket serves our frontend files, but does not have any affordances for specifying custom headers. But it serves from globs, a non-trivial task to take over, even if we were to run a custom fork of it.

Instead, here we introduce a local proxy that watches on the endpoint docket would normally serve on. It passes requests through unaltered, but injects the desired headers into responses before sending them back to the client.

This approach will stop working if docket ever injects responses on this endpoint into the eyre cache, which means we can't trivially capture them in userspace anymore. However, vere 3.1 never ever, so we should be good for the time being.

### DIFF
--- a/desk/app/groups-docket-proxy.hoon
+++ b/desk/app/groups-docket-proxy.hoon
@@ -1,0 +1,114 @@
+::  groups-docket-proxy: mitm glob responses
+::
+::NOTE  this will stop working when docket starts putting responses into
+::      the eyre response cache. at that point we'll probably need to do
+::      something more involved (or just wipe the cache entries)...
+::
+|%
+++  extra-headers
+  ^-  header-list:http
+  :~  ['cross-origin-embedder-policy' 'require-corp']
+      ['cross-origin-opener-policy' 'same-origin']
+  ==
+::
+++  give-response
+  |=  [eyre-id=@ta simple-payload:http]
+  ^-  (list card)
+  =+  head=http-response-header+!>(response-header)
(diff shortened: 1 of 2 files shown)


desk/app/groups-docket-proxy.hoon
```diff
@@ -0,0 +1,114 @@
...
+        %fact
+      ?.  ?=(%http-response-header p.cage.sign)
+        [%give %fact [wire]~ cage.sign]~
+      =+  !<(head=response-header:http q.cage.sign)
+      ::NOTE  debatable whether this needs to use +set-header or not
+      =.  headers.head  (weld headers.head extra-headers)
+      [%give %fact [wire]~ p.cage.sign !>(head)]~
+    ::
+        %kick
+      [%give %kick [wire]~ ~]~
```
arthyn, line 96: so if I understand this correctly, for anything besides `%http-response-header` we just relay to the http response subscription that came to us from eyre. if it is `%http-response-header` we add the extra headers we need and then give that fact to the pending http response.

arthyn (approved):
lgtm

desk/app/groups-docket-proxy.hoon
Fang-: Correct.

patosullivan (approved):
Tested locally with the sqlite web frontend, it works.

PR #3812, opened 2024-07-31
native: relocate group options context + actions

jamesacklin:
Move from in-app context/drilled props to GroupOptionsContext + provider.

Uses group option sheet + new provider + hook in ChannelScreen -> BaubleHeader.

Fixes TLON-2451


dnbrwstr (approved):
Great, lgtm!

PR #3813, opened 2024-07-31
native: add iOS notification service extension

davidisaaclee:
Fixes TLON-2393

## Summary
Most configuration changes are duplicated across (1) a main target (like `Landscape`) and (2) a preview target (like `Landscape-preview`).

- Adds `Notifications` and `Notifications-preview` targets to Xcode project
- Add `group.io.tlon.groups` and `group.io.tlon.groups.preview` app groups
- Moves iOS code around so that we include a minimal amount of code in the new targets. This includes splitting some code files up; adding existing files to the new targets; and moving code files so that all files which are assigned to multiple targets are in a new `Shared` directory/group
- Updates some iOS code to use app group shared resources (`UserDefaults`, `HTTPCookieStorage`), including changing some singletons implemented using static methods (`PocketAPI.fetchDecodable`) to be instance methods on a shared instance (`PocketAPI.shared.fetchDecodable`)
- Remove storage of `authCookie` in keychain, as it did not appear to be used
- Adds `TlonDefaultAppGroup` field to each target's `Info.plist`, with the id of the default app group for that app (e.g. `group.io.tlon.groups`)

There's a big diff in the Xcode project file – majority here is moving files and assigning files to the new targets, but there are some important changes (mainly around changes to the `Info.plist`s). I'd maybe just open the project in Xcode and look through the new targets.

[Avoid cycle in Landscape target build](https://github.com/tloncorp/tlon-apps/pull/3813/commits/de96cbb1eca765c5d323affa45303d01ac7c12c7) and [Reorder build phases to avoid false circular dependency detection](https://github.com/tloncorp/tlon-apps/pull/3813/commits/80d54ff9351e52268637bfe8e0d09b995f03bfeb) reorder build phases to avoid Xcode complaining about a circular dependency – I don't understand this but it works.

## Decisions
- When accessing `shipURL`, read from UserDefaults each time instead of going through UrbitModule, and instead of configuring API instances with a delegate (e.g. from `PocketAPI`, calling `self.urbitProvider?.shipURL`)
  - Using `UrbitModule` was a no-go for use outside of host app
  - UserDefaults intelligently caches its reads, so reading directly seems fine to me (instead of having an in-memory cache as before)
- One NSE target per host target (a `Notifications` target and a `Notifications-preview` target, instead of a single target that dynamically changes its config based on the host app it is being embedded into)
  - Significantly simpler configuration than trying to dynamically change config
  - This approach requires adding new files to both targets :( and mirroring config changes to both targets
- One app group per target (instead of a shared app group for all Tlon builds)
  - This avoids needing to namespace keys in `UserDefaults`, which would be a significant footgun
  - I can't think of a reasonable feature for sharing data between targets; if we do need this later on, we can add a third shared app group
- Remove `authCookie` from login storage
  - afaict, we can't access the host app's keychain from the NSE, so we can't read the `authCookie` from the host app
  - `authCookie` doesn't appear to be doing anything - removing it completely from keychain does not affect the ability to fetch notification content
- Mirror changes in cookies in host app's HTTPCookieStorage.shared -> app group cookie storage
  - I tried moving all cookie storage read/writes to app group store (i.e. in our iOS code; when setting up an AF session; and in the RCTNetworking session provider), but the cookie kept on getting written to the default store. I see a lot of references to `HTTPCookieStorage.shared` in the RCTNetworking code (not using the caller's session provider), so I'm guessing the write is somewhere in there

## How did I test?
I tested exclusively on the preview build, as we don't have notifications set up for the main debug build. I don't know which target is used for the Testflight build – **if Testflight is not `io.tlon.groups.preview`, there is a decent chance of notifications breaking when we deploy**, as there are a number of places where configuration needs to change based on which target is being used. (I think I got all of them but I am human.)

Between a moon and my hosted ship...
- Sent DMs while app in background/killed; banner looked as it did before changes (with message presentation, still broken sigil); populated with display name if available, else ship name
- Sent DMs while app in foreground; banner did not show
- Sent group message  while app in background, while notifications enabled; banner showed with group display name + sender name
- Sent group message while app in background, while notifications disabled; banner showed with group display name + sender name

**What other flows should I test?**

## How to deploy?
- We need to update the production Twilio service to provide `mutable-content: 1` in the apns request (implemented in [`tlon-preview-debug`](https://console.twilio.com/us1/develop/functions/api/start/ZSa265ba09f28ce9a228e4a9e25f6dafc2)).
- We should also add a more _srs bizness_ default alert to the apns payload, which will be shown to users if the NSE fails


jamesacklin:
Beautiful work and excellent write-up. I'll comb through the finer points tomorrow morning; wanted to give you a "seen." Thank you!

dnbrwstr (approved):
Went through; this looks great to me! TestFlight indeed uses the preview target, so that shouldn't be an issue. 

Only request is let's make sure we have a deploy plan before we merge in so this doesn't cause any issues with the build we'll do tomorrow. I'll start a thread in the group...

PR #3814, opened 2024-08-01
native: enhanced crash/bug reports

latter-bolden:
Some updates to error reporting. Ideally, this will give us more context when investigating crashes and allow easier handoffs of relevant info for issues like TLON-2442.

### Logger based debug breadcrumbs
To get debug info into crashlytics, previously you had to explicitly instantiate a new ErrorReporter and add any relevant logging. You would then have to manually detect relevant failures for when to send the report. Now you can use normal dev loggers to tack on anything you want to show up in crashlytics. In addition to `logger.log` you now have:

`logger.crumb` — attach a debug line that will always be included in the report
`logger.sensitiveCrumb` — same thing but for information that we may want to include optionally for privacy reasons

It operates like a normal log, you'll still get output in the console during dev. As implemented, the last 100 crumbs are stored and always uploaded to Crashlytics. Currently using a naive Array push/shift to track those, but we can do something fancier if performance dictates.

### Manual bug reporting
I updated our error boundary to always include the breadcrumbs. In addition to relying on crashes to get debug information, users can now manually submit bug reports. When a report is submitted, we send the debug breadcrumbs as well as any additional notes about the issue the user may have added.

https://github.com/user-attachments/assets/2f383016-756f-4292-a2de-2e6502be3f6d

### What Information is now included
Previously, sync start + the stack trace for the crashing error was the only info we got in Crashlytics. We now have breadcrumbs wired up for:
- sending posts and replies + tracking delivery
- uploading attachments
- cursor context when viewing a channel
- screen navigations
- `%groups` sync ship and current hash
- network disconnects and reconnects
- battery level, whether the device is on low power mode

happy to wire it up for more mutations and interactions if we like the pattern.


latter-bolden:
Need to fix an android display issue with the manual logging, but should otherwise be ready for review.

apps/tlon-mobile/src/screens/UserBugReportScreen.tsx
```diff
@@ -0,0 +1,193 @@
...
+              </SizableText>
+            </YStack>
+          )}
+          <View paddingTop="$2xl">
+            {state === 'initial' ? (
+              <>
+                <SizableText
+                  marginHorizontal="$m"
+                  marginBottom="$l"
+                  fontSize="$s"
```
dnbrwstr, line 122: `SizableText` only requires `size` prop set -- corresponding `fontSize` and `lineHeight` will be set automatically.

apps/tlon-mobile/src/hooks/useNavigationLogger.ts
```diff
@@ -0,0 +1,36 @@
+import { useNavigation } from '@react-navigation/native';
+import { useEffect, useRef } from 'react';
+
+export function useNavigationLogging(
+  logger: (from: string, to: string) => void
```
dnbrwstr, line 5: Why inject here instead of using the same pattern as `useNetworkLogger`?

dnbrwstr (approved):
Lgtm overall!  A few nits.

apps/tlon-mobile/src/screens/UserBugReportScreen.tsx
latter-bolden: so I thought! But it looked off to me, explicitly setting to `$s` does seem to have a visual change...

latter-bolden: Whoops, was surprised line height wasn't getting set. Forgot to use the right prop

PR #3815, opened 2024-08-01
native: revert UrbitModule signature to match Android

davidisaaclee:
#3813 changed `UrbitModule.setUrbit`'s signature in JS + iOS because iOS does not use one of the params, but the change was not reflected in the Android native module. This change makes the signature the same for both Android/iOS, and just ignores the unused param in iOS.

I built this on iOS, but I have not built it on Android


latter-bolden (approved):
Looks good! I can log in on Android without a crash

PR #3816, opened 2024-08-01
join-requests: show up in activity and allow management on mobile

arthyn:
This fixes TLON-2152 by adding join requests to the feed endpoint and also accounting for them in the summaries. Additionally, mobile had no way of acting upon these so that's now available in the Members screen as another section similar to bans. The profile sheet was getting crowded so I made a sheet specifically for join requests. Still testing to make sure this all works. 

Also cleaned up the behavior in groups so that if you previously asked to join and you receive an invitation, you just automatically start the join process instead of having to accept an invite.

PR Checklist


packages/ui/src/components/Activity/ActivityScreenView.tsx
```diff
@@ -78,6 +83,14 @@ export function ActivityScreenView({
             console.warn('No parent found for reply', event);
           }
           break;
+        case 'group-ask':
+          if (event.group) {
+            markGroupRead(event.group);
```
latter-bolden, line 88: if we're only going to run both maybe pass a single handler that worries about that itself

packages/ui/src/components/Activity/GroupActivitySummary.tsx
```diff
@@ -0,0 +1,133 @@
...
+              {` ${plural ? 'have' : 'has'} requested to join the group`}
+            </SizableText>
+          </View>
+        </YStack>
+      </XStack>
+    </View>
+  );
+}
+
+export function GroupIndicator({
```
latter-bolden, line 106: do we use this elsewhere? Can we reuse the same header across both feed types?

latter-bolden (approved):
This looks awesome! Way to hit the ground running on the native codebase

Some nits, but they're all minor and non-blocking. This worked successfully end to end on both native and web.

One thing we may want to monitor once it's in the wild: since we modified bundling logic for activity events (dupe filtering for example), we should ensure the final composite feeds continue to look correct.

desk/app/groups.hoon
```diff
@@ -1940,6 +1950,8 @@
     --
   ++  ga-start-join
     ^+  ga-core
+    %-  (log |.("claim: {<cam.gang>}"))
+    =.  cam.gang  `[& %adding]
```
Fang-, line 1954: This refactoring feels like it should've gone into a separate PR.

Either way, previous logic had `join-all.join` instead of hard-coded `&`. Even if it's always `&` in practice, are we sure we want to hard-code it? Should probably just make this arm a gate, right?

Fang- (approved):
Better, thanks!

PR #3818, opened 2024-08-02
native: favorite groups selection flow

latter-bolden:
OTT

https://github.com/user-attachments/assets/66bf4613-2c05-4001-962a-95ba0b055ea5

Fixes TLON-1726


dnbrwstr:
Two questions on the UX first:
- In the video, the initial screen says "None selected" for favorite groups -- seems like a mistake?
- Could we just pop the sheet when you tap "favorite groups"? Having a whole second screen seems unnecessary.

latter-bolden:
> "None selected" for favorite groups -- seems like a mistake?

Was indeed a mistake, it's fixed on the latest commit

> Could we just pop the sheet when you tap "favorite groups"?

I think you do wanna preview which ones you have set. We haven't really come up with an elegant way to do that within the existing contact book style multiselect. Likely feels cramped if you just shove it above the selector? In which case we're probably looking at a multi pane sheet kinda thing

All that to say, sure I can refactor it that way if you think the dedicated screen UX is meaningfully rough, but I'd prefer not to :~)

latter-bolden:
We chatted out of band. Think we landed on a good UX that avoids the separate screen without going full sheet maximalism. Moving back into draft until that's in.

latter-bolden:
Given the edits needed, was easier to just combine this with the other PR.

Closing in favor of #3819

PR #3820, opened 2024-08-02
native: render optimization

dnbrwstr:
Lots of changes here, so I'd recommend viewing by commit:
- 0a39f8937d784e0c198232462b90c73e545ca657 Enables Tamagui flattening of components in the `@tloncorp/ui` module, which was previously not working at all. This was failing when the tamagui compiler's inner compiler attempted to use esbuild to require the module and I had to tweak package.json a bit to get it to import correctly.
- 331e6c5d1078ee5bd1621e86a41f2849b75e745f Biggest raft of file changes -- getting rid of `ui/core`. Tamagui was failing to flatten components re-exported by core, so I've just removed that and modified components so that they import from tamagui directly.
- 66ffe3f4677802cb2b4721eb12dd90fa95f3bd1e Refactors `ChatList`, removing scroll animation and adding the new reanimated-based swipeable component, among other things.

Fixes TLON-2343
Fixes TLON-2387


packages/ui/src/components/ListItem/ListItem.tsx
```diff
@@ -123,27 +133,34 @@ export const ListItemTimeText = styled(SizableText, {
...
+      }
+    }, [time]);
+    return (
+      <ListItemTimeText {...props} ref={ref}>
+        {formattedTime ?? ''}
+      </ListItemTimeText>
+    );
+  },
+  {
+    staticConfig: {
```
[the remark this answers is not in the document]
dnbrwstr: It's not critical, but it is nice to have -- basically just makes sure React Devtools displays the correct name.

packages/ui/src/index.tsx
```diff
@@ -59,7 +59,40 @@ export * from './components/TamaguiProvider';
...
 export * from './contexts';
 export * from './tamagui.config';
 export * from './types';
 export * from './utils';
+
+export {
+  Circle,
+  useTheme,
+  useStyle,
+  View,
```
davidisaaclee, line 72: When should we do `import { View } from 'ui'` instead of `import { View } from 'tamagui'`?

packages/shared/src/debug.ts
```diff
@@ -194,3 +194,27 @@ export function useSendPosts(
...
 }
+
+export const useLogChange = (label: string, value: unknown) => {
+  const isFirst = useRef(true);
+  useEffect(() => {
+    if (isFirst.current) {
+      isFirst.current = false;
+      return;
+    }
+    console.log(label, 'changed', value);
```
davidisaaclee, line 205: here's a nitpick – neither of my solutions seem worth it to me, but mentioning in case you think of something:

It'd be nice to protect these `console.log`s from adding to noise (e.g. someone commits a `useLogChange`, now other dev needs to either undo that change or comment out this `console.log` to silence).

1. Add third param `logger: ReturnType<typeof createDevLogger>` (although someone can just pass `console` and we'd have same issue of noise)
2. Add lint rule to avoid committing code that calls these functions

packages/ui/src/components/ChatList.tsx
```diff
@@ -242,164 +150,273 @@ export function ChatList({
...
+  return (
+    <>
+      <ChatListFilters
+        query={searchQuery}
+        onQueryChange={setSearchQuery}
+        activeTab={activeTab}
+        onPressTab={setActiveTab}
+        isOpen={showFilters}
+      />
+      {searchQuery !== '' && !displayData[0]?.data.length ? (
```
davidisaaclee, line 170: Should this be `searchQuery.trim() !== ''`?

If that's a bug, I think `useFilteredChats` could give more info here to avoid similar bugs – e.g. return `{isSearching: boolean, hasResults: boolean, displayData: ReturnValueOfCurrentImpl}` then just check `isSearching` here

packages/ui/src/components/ChatList.tsx
```diff
@@ -242,164 +150,273 @@ export function ChatList({
...
+  return performSearch;
+}
+
+function useDebouncedValue<T>(input: T, delay: number) {
+  const [value, setValue] = useState<T>(input);
+  const debouncedSetValue = useMemo(
+    () => debounce(setValue, delay, { leading: true }),
+    [delay]
+  );
+  useLayoutEffect(() => {
```
davidisaaclee, line 388: why `useLayoutEffect` here? I think of `useLayoutEffect` as like "make this change blocking a repaint because it will likely change the outcome of the repaint" – I think `debounce` will always call `setValue` asynchronously, which means we'll never get a change to `value` synchronously before the repaint; so a `useEffect` should work as well and prevent adding work that blocks a render. (i'm not convinced that i understand useLayoutEffect, lmk if that's wrong)

packages/ui/src/components/ChatList.tsx
```diff
@@ -242,164 +150,273 @@ export function ChatList({
...
+  activeTab,
+}: {
+  pinned: Chat[];
+  unpinned: Chat[];
+  pending: Chat[];
+  searchQuery: string;
+  activeTab: TabName;
+}) {
+  const performSearch = useChatSearch({ pinned, unpinned });
+  const debouncedQuery = useDebouncedValue(searchQuery, 200);
```
davidisaaclee, line 317: this seems like a great use case for `useTransition`, if you feel like it! could probably get snappier performance.

davidisaaclee (commented):
whoops, I'm too late! here's a review anyways. I obviously don't know the previous state of the code, so hopefully I'm not commenting on irrelevant stuff.

patosullivan:
Cool, I wonder if this will fix some HMR issues on the web side

PR #3822, opened 2024-08-02
queries: do less

arthyn:
The way that 'keyFromQueryDeps' generated a key was to create a `Set` out of the dependency array from the query's table dependencies. However, doing so seemed to create a bad value for the key at least when logged, maybe the actual value was fine, but just using the array by itself seems to increase performance pretty significantly and I see way less DB reads. 

Also added a stale time of 5 minutes because we always invalidate queries based on table changes so we shouldn't see stale data.

PR Checklist


arthyn:
The tests failed locally but didn't show what test failed or any exceptions it threw so I may need help Monday figuring that out.

arthyn:
closing for now

PR #3823, opened 2024-08-05
activity: dont use now.bowl when reading

arthyn:
This was posted in Homestead. The Bulletins channel was randomly getting bumped to the top. We use a null `time` in the `read` action to signify "latest", however if we came up with no events we were erroneously saying set it to `now.bowl`. Since the floor gets taken into account for the recency, this was bumping sources to the top that were empty themselves or contained empty sources. This instead leaves the floor unchanged if there are no events.

PR Checklist


Fang- (approved):
Makes sense. Approved!

PR #3824, opened 2024-08-06
activity: fix notification paths

arthyn:
Fixes TLON-2423 by adjusting the paths we send people to, reflecting the actual routes instead of all trying to be chats. Also fixed two issues encountered when testing.

- Creating a post and then commenting on it yourself was crashing because the source didn't exist
- Sometimes the feed would produce bundles where the latest time didn't match any event in the bundle, which makes no sense. So I changed it to always pull the "latest" value from the list of events it sends with the bundle (instead of the event on which we built this bundle which may not be present in the bundle itself)

PR Checklist

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -548,8 +548,9 @@
   =/  collapsed
     (~(gas in collapsed.acc) (turn top head))
   :-  (~(put by sources.acc) source src-info(added &))
-  =/  happening  [source (head (head top)) top]
-  [(sub limit.acc 1) (snoc happenings.acc happening) collapsed]
+  ?~  top  +.acc
+  =/  latest=^time  -<.top
+  [(sub limit.acc 1) (snoc happenings.acc [source latest top]) collapsed]
   +$  out
     $:  sources=(map source:a [latest=time-id:a added=?])
         limit=@ud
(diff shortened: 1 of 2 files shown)


latter-bolden (approved):
js change makes sense

desk/app/activity.hoon
```diff
@@ -548,7 +548,8 @@
   =/  collapsed
     (~(gas in collapsed.acc) (turn top head))
   :-  (~(put by sources.acc) source src-info(added &))
-  [(sub limit.acc 1) (snoc happenings.acc [source time top]) collapsed]
+  =/  happening  [source (head (head top)) top]
+  [(sub limit.acc 1) (snoc happenings.acc happening) collapsed]
```
Fang-, line 552: > the event on which we built this bundle (...) may not be present in the bundle itself

That sounds wild, not sure I understand why that would be the case.

Either way the `top` here is a `list`, I don't think it's strictly guaranteed to contain at least one item right? So this may runtime crash in those cases. Are we ok with that?

(`+<.top` would probably be more idiomatic, too.)

Fang- (approved):
A remaining nit, but looks good otherwise.

desk/app/activity.hoon
```diff
@@ -548,8 +548,9 @@
   =/  collapsed
     (~(gas in collapsed.acc) (turn top head))
   :-  (~(put by sources.acc) source src-info(added &))
-  =/  happening  [source (head (head top)) top]
-  [(sub limit.acc 1) (snoc happenings.acc happening) collapsed]
+  ?~  top  +.acc
+  =/  latest=^time  -<.top
```
Fang-, line 552: I realize I may seem to be moving the goalposts, but: `time.i.top` please. (At which point do you really need the intermediate assignment?)

PR #3826, opened 2024-08-07
notify: only handle authenticated http requests

Fang-:
`/app/notify` serves a custom endpoint for grabbing notification data. It handles raw http requests for this, but we weren't checking them for authentication. Now we do.

Closes TLON-2486.


arthyn (approved):
lgtm

PR #3829, opened 2024-08-08
native: fix issue with notebook input being obscured

patosullivan:
Fixes TLON-2525

It looks like the View component that we were using from core was slightly different (?) than the View component from the tamagui package, and the properties set in the `style` prop were just being ignored, which caused a bunch of weirdness. Switched over to setting them all directly as props (the tamagui way) and now it's fine.

Also fixed a few other small issues in this component (deprecated prop on the image, unnecessary condition).

Also had to run `pod install --repo-update` when I attempted a pod install, got this change to the project.pbxproj file.


jamesacklin (approved):
Works for me locally. I think you were out—Dan flattened the Tamagui library such that we don't use the `core` components any more. It makes sense that we were passing things along to the core components differently than the Tamagui base components.

PR #3830, opened 2024-08-08
various: %egg-any import support

Fang-:
Adds support for restoring agent state from `$egg-any:gall` backups. (That's the backup format [phoenix](https://github.com/urbit/phoenix) supports.)

This restores both locally-hosted and remote data. We do our best to re-establish subscriptions based on the data we have, generally picking up where we left off. It restores state for DMs (chat), groups (groups, channels-server, channels, activity) and public profiles (profile). Contacts data (at least our own profile) should probably support this, but it outside of this repo's scope.

I tested this pretty extensively prior to the addition of restoring non-locally-hosted data, and that seemed to work well. After the latest batch of changes, I'll want to do another serious round of testing.  

(Specifically, I may have introduces some additional ordering concerns that I'll need to address. I suspect we want to do both channels _and_ channels-server before groups now, and maybe activity before all of those... And, need to make sure subscriptions get picked up at the right places under all circumstances. ...To be continued.)

One missing piece here is actually on the "everyone else" side of this code: if a ship that hosts a group breaches, comes back, then re-imports the group, those subscriptions don't get re-established automatically. We may or may not be able to solve this from the group host side. (This may be more complicated for the "hosting a channel in a group I don't host myself" cases, too.) So, some pending work here, but may not be considered strictly blocking.

Realizing now that this was tested only on fully-joined states, not on weird in-between "trying to join the group" states or stuff like that, or on invites for that matter. Those are probably of secondary concern, but we should know what the behavior there is at least.

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -290,7 +290,7 @@
         (load -:!>(*versioned-state:load) +>.old-state.egg-any)
       ::  restore volume settings, but keep any we've explicitly set ourselves
       ::
-      =.  allowed  allowed:bak  ::REVIEW
+      =.  allowed  allowed:bak
       =.  volume-settings  (~(uni by volume-settings:bak) volume-settings)
       cor
     ==
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -286,12 +286,13 @@
         ~&  [dap.bowl %egg-any-not-live]
         cor
       =/  bak
-        ::TODO  test
         (load -:!>(*versioned-state:load) +>.old-state.egg-any)
-      ::  restore volume settings, but keep any we've explicitly set ourselves
+      ::  restore volume settings, but keep any we've explicitly set ourselves,
+      ::  and restore activity summaries  ::REVIEW  for unintended side-effects
       ::
       =.  allowed  allowed:bak
       =.  volume-settings  (~(uni by volume-settings:bak) volume-settings)
+      =.  activity         (~(uni by activity:bak) activity)
       cor
     ==
   ==
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -125,8 +125,8 @@
   (emit %pass /migrate %agent [our.bowl dap.bowl] %poke noun+!>(%migrate))
 ::
 ++  load
-  |=  =vase
-  |^  ^+  cor
+  |^  |=  =vase
+  ^+  cor
   ?:  ?=([%0 *] q.vase)  init
   =+  !<(old=versioned-state vase)
   =?  old  ?=(%1 -.old)  (state-1-to-2 old)
@@ -276,6 +276,23 @@
       %read     (read source.action read-action.action |)
       %adjust   (adjust +.action)
       %allow-notifications  (allow +.action)
+    ==
+  ::
+      %egg-any
+    =+  !<(=egg-any:gall vase)
+    ?-  -.egg-any
+        ?(%15 %16)
+      ?.  ?=(%live +<.egg-any)
+        ~&  [dap.bowl %egg-any-not-live]
+        cor
+      =/  bak
+        ::TODO  test
+        (load -:!>(*versioned-state:load) +>.old-state.egg-any)
+      ::  restore volume settings, but keep any we've explicitly set ourselves
+      ::
+      =.  allowed  allowed:bak  ::REVIEW
+      =.  volume-settings  (~(uni by volume-settings:bak) volume-settings)
+      cor
     ==
   ==
 ::
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -290,7 +290,7 @@
       ::  restore volume settings, but keep any we've explicitly set ourselves,
       ::  and restore activity summaries  ::REVIEW  for unintended side-effects
       ::
-      =.  allowed  allowed:bak
+      =.  allowed          allowed:bak
       =.  volume-settings  (~(uni by volume-settings:bak) volume-settings)
       =.  activity         (~(uni by activity:bak) activity)
       cor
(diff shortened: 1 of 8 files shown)


desk/sur/activity.hoon
```diff
@@ -5,7 +5,9 @@
 ::  $stream: the activity stream comprised of events from various agents
 +$  stream  ((mop time event) lte)
 ::  $indices: the stream and its read data split into various indices
-+$  indices  (map source index)
++$  indices
```
arthyn, line 8: good change

desk/app/activity.hoon
```diff
@@ -277,6 +278,24 @@
...
+    =+  !<(=egg-any:gall vase)
+    ?-  -.egg-any
+        ?(%15 %16)
+      ?.  ?=(%live +<.egg-any)
+        ~&  [dap.bowl %egg-any-not-live]
+        cor
+      =/  bak
+        (load -:!>(*versioned-state:load) +>.old-state.egg-any)
+      ::  restore volume settings, but keep any we've explicitly set ourselves,
+      ::  and restore activity summaries  ::REVIEW  for unintended side-effects
```
arthyn, line 292: they need these sources to exist in the index because if they have unreads here and try to read them before getting new events it will just crash. the least we can do here is grab all the new sources and bunt the index and place them in the indices. maybe even give them a floor:

```hoon
=.  indices
  %.  indices
  %~  uni  by
  %-  ~(urn by activity:bak)
  |=  [=source:a as=activity-summary:a]
  ?~  unread.as  *index:a
  index(floor.reads (sub time.u.unread.as (div ~s1 100)))
```

although I'm inclined to say even with this it might not be enough, may just be better to import indices same way you do activity.

desk/app/channels.hoon
```diff
@@ -386,6 +405,22 @@
   |=  [=mark =vase]
   ^+  cor
   ?+    mark  ~|(bad-poke+mark !!)
+      %noun
+    ?+  q.vase  !!
+        [%channel-wake @ @]
+      =+  ;;([=kind:c name=term] +.q.vase)
+      =/  =nest:c  [kind src.bowl name]
+      ?.  (~(has by v-channels) nest)  cor
+      ca-abet:(ca-safe-sub:(ca-abed:ca-core nest) |)
```
arthyn, line 414: should we maybe do a leave first here? or do we not need to?

desk/app/channels.hoon
```diff
@@ -475,6 +519,27 @@
...
+    ?.  ?=(%live +<.egg-any)
+      ~&  [dap.bowl %egg-any-not-live]
+      cor
+    =/  bak
+      ::TODO  test
+      (load -:!>(*versioned-state:load) +>.old-state.egg-any)
+    ::  we only restore miscellanea. groups' import should prompt us to
+    ::  re-join channels, we won't bother pre-loading their content for now.
+    ::
+    =.  v-channels    (~(uni by v-channels:bak) v-channels)
```
arthyn, line 537: is this comment still correct given that we now do import `v-channels`?

desk/app/channels.hoon
```diff
@@ -872,8 +938,13 @@
...
     |=  [n=nest:c group=flag:g]
     =.  nest  n
-    ?<  (~(has by v-channels) nest)
     ?>  |(=(p.group src.bowl) from-self)
+    ?:  (~(has by v-channels) nest)
+      ::  we should already be in, but make sure our subscriptions still exist
+      ::  just in case
+      ::
+      =.  channel  (~(got by v-channels) nest)
+      (ca-safe-sub |)
```
arthyn, line 947: this is a good change to join. just maybe something vestigial here:

```suggestion
    ?:  (~(has by v-channels) nest)
      ::  we should already be in, but make sure our subscriptions still exist
      ::  just in case
      ::
      (ca-safe-sub |)
```

desk/app/groups.hoon
```diff
@@ -116,6 +117,21 @@
...
     ?+  q.vase  !!
       %reset-all-perms  reset-all-perms
+    ::
+        [%group-wake flag:g]
+      =+  ;;(=flag:g +.q.vase)
+      ?.  |(=(our src):bowl =(p.flag src.bowl))
+        cor
+      ?~  g=(~(get by groups) flag)
+        cor
+      go-abet:(go-safe-sub:(go-abed:group-core:cor flag) |)
```
arthyn, line 127: same question here, do we need a leave?

desk/lib/negotiate.hoon
```diff
@@ -772,7 +780,30 @@
...
+      ::  strip the negotiate state out of the egg, tidy up its bitt/boat,
+      ::  and pass it on to the inner agent
+      ::
+      =/  =cage  !:
+        :-  mark
+        !>  ^-  egg-any:gall
+        =+  !<(=egg-any:gall vase)
+        ?>  ?=(%live +<.egg-any)
+        %_  egg-any
+          +.old-state  [-:!>(**) q:(slot 7 +.old-state.egg-any)]
```
arthyn, line 801: scary!

arthyn (commented):
Overall changes look good, a few small changes are warranted, and I might need you to walk me through the negotiate diff.

desk/app/channels.hoon
Fang-: Not sure why we'd need to. If the subscription is already there, it's fine, it can just continue existing. We just want to (re-)establish it if it's not there, which is what `+ca-safe-sub` does.

desk/app/channels.hoon
Fang-: Not vestigial, `+ca-join` gets called directly from the core, instead of initializing the core through `+ca-abed` first. This is also why this arm does the `=.  nest` a couple lines above.

desk/app/channels.hoon
arthyn: what if we think it's there, but they do not? will gall reject a duplicate sub going out like it does coming in?

arthyn (approved):
lgtm!

PR #3831, opened 2024-08-09
native: hide menu button on detail views

patosullivan:
Hides the three dots menu in detail views (also hides it from the channel view, where it shouldn't have been visible anyway).


jamesacklin (approved):
LGTM!

PR #3832, opened 2024-08-09
native: chat options sheets + screens

dnbrwstr:
Changes
- Added dm + group dm options sheets
- Added routes + screens for dm info editing + members view
- Implemented leave dm / modify dm actions + routes
- Memoized some components
- Refactored ActionSheet + modified to better match design
- Updated most icons from figma

This is part 1 of 2. Next chunk of work will be
- Convert remaining ActionSheets to new format
- Move modal rendering to new app directory + set up to be controlled via context
- Further optimize sheet rendering + possibly look at alternate renderers

Fixes TLON-2438
Fixes TLON-2437


patosullivan (approved):
Looks great, nice work.

PR #3835, opened 2024-08-12
native: avoid crash on missing notification payload

davidisaaclee:
fixes TLON-2551
fixes TLON-2495

pasted context from 2495:

> for an unknown reason, we're not sending a proper payload from the NSE to the host app; the host app assumes we are getting a well-formed payload, and crashes when trying to access a property that doesn't exist.
> 
> 1. We should not crash on badly-formed payloads – since we don't have any data here to point the user to the notification's subject, I think we just do nothing and log an error with all the data we do have, so we can investigate what's going on
> 2. We should figure out why we're not getting a payload.

This PR addresses (1) by avoiding access on `data`, and logs the full payload to PostHog – this should avoid crashes, but will not navigate anywhere in the app.

**🚨 There's a chance of logging message content + sender to PostHog with this change:** There's a chance that the notification from `expo-notifications` _does_ have the payload we fetched, but it's just in an unexpected place in the `response.notification.request` object we're passing to PostHog.


davidisaaclee:
I'd love to do a proper validation of the payload to avoid any future bugs here – do we have anything like zod / typebox that we could use to validate against a schema? (looked but didn't see anything, looks like we just assert types on API calls)

dnbrwstr:
We don't have anything like that currently, but I'd be open to it. Are you thinking to use it to catch errors earlier in development? Or how would you handle invalid payloads at runtime?

dnbrwstr:
Also, wdyt about using the `isTlonEmployee` check  (`queries.getIsTlonEmployee`) to gate this logging? I've seen the bug, so I'd imagine we'd still get useful data out of it.

davidisaaclee:
> Are you thinking to use it to catch errors earlier in development? Or how would you handle invalid payloads at runtime?

Runtime – the theoretical bug is if the notification payload changes (on native side or on backend) and we don't change the JS implementation, so we're trying to e.g. access `data.wer` when `wer` has moved to `data.payload.wer` – we'd catch a validation error and just avoid access.

> isTlonEmployee check

makes sense, I'll add this now!

davidisaaclee:
>isTlonEmployee

added in [bde3c81](https://github.com/tloncorp/tlon-apps/pull/3835/commits/bde3c81819f017df3723042a653ac895c9ae3afd)

dnbrwstr (approved):
Lgtm! Re: typebox, I think maybe we hold off for now -- the particular call site is kind of a one-off, and I'd wanna talk through what the best solution for the whole api module would be.

jamesacklin:
Awesome, thank you for the hustle @davidisaaclee!

PR #3837, opened 2024-08-13
render new web app w/new app structure

patosullivan:
This adds a new vite mode, `new`, that allows for us to build the new app for web. Most screens outside of settings/profile are covered. Those still need to be added. There are some  residual issues with reanimated that we need to figure out, and an issue with the initial load being blank until interaction (didn't happen on the old branch, still need to figure that out).

This seemed like a good place to cut and merge back in before it diverges too much from develop.

The changes don't seem to effect the mobile app, but it would be a good idea to pull it down and test for yourself.

We also need to make sure the new wasm mark makes it onto the globber fakezod in gcp.

To run the new web app locally, from tlon-web, run `pnpm dev:new`.


packages/ui/src/components/ChatList.tsx
```diff
@@ -28,7 +28,7 @@ import Animated, {
   useAnimatedStyle,
```
patosullivan, line 28: try tamagui ignore on this file, remove prop comparison func.

packages/ui/src/components/MessageInput/MessageInputBase.tsx
```diff
@@ -148,22 +155,23 @@ export const MessageInputContainer = ({
         ) : (
           <View paddingBottom="$xs">
             {disableSend ? null : (
-              <IconButton
+              <Button
                 disabled={isSending}
-                color={'$primaryText'}
+                // color={'$primaryText'}
```
patosullivan, line 160: Why?

packages/ui/src/components/WelcomeSheet.tsx
```diff
@@ -41,7 +41,7 @@ function WelcomeSheetComponent({
               <Image
                 style={{ width: '100%', height: 188 }}
                 resizeMode={'cover'}
-                source={require('../assets/raster/welcome_flowers.jpg')}
+                // source={require('../assets/raster/welcome_flowers.jpg')}
```
patosullivan, line 44: Wat?

packages/ui/src/contexts/attachment.tsx
```diff
@@ -109,7 +110,7 @@ export const AttachmentProvider = ({
   useEffect(() => {
     attachments.forEach((a) => {
       if (a.type === 'image' && !a.uploadState) {
-        uploadAsset(a.file);
+        uploadAsset(a.file, isWeb);
```
patosullivan, line 113: We should just use an env var within storageActions for this.

apps/tlon-mobile/src/screens/BlockedUsersScreen.tsx
```diff
@@ -1,5 +1,5 @@
 import { NativeStackScreenProps } from '@react-navigation/native-stack';
-import { useCurrentUserId } from '@tloncorp/app/hooks/useCurrentUser';
+import { useCurrentUserId } from '@tloncorp/app/hooks/useCurrentUser.native';
```
patosullivan, line 2: Switch back to normal import

apps/tlon-web/src/controllers/ActivityScreenController.tsx
```diff
@@ -0,0 +1,57 @@
+// import { NativeStackScreenProps } from '@react-navigation/native-stack';
+import { ActivityScreen } from '@tloncorp/app/features/top/ActivityScreen';
+import * as db from '@tloncorp/shared/dist/db';
+import * as store from '@tloncorp/shared/dist/store';
+import { useCallback } from 'react';
+import { useNavigate } from 'react-router';
+
+// import { RootStackParamList } from '../types';
```
patosullivan, line 8: Remove

packages/app/hooks/useChannelContext.native.ts
dnbrwstr, line 1: A lot duplicated between this and the web version -- could we do something like this?

Current:
- `useChannelContext.native.ts` 
- `useChannelContext.ts`

Next
- `useChannelContext.ts` 
- `useChannelNavContext.native.ts` 
- `useChannelNavContext.ts`

packages/ui/src/components/FloatingAddButton.native.tsx
dnbrwstr, line 1: Let's kill `ContextMenu` and use a sheet to avoid splitting this file.

packages/ui/src/contexts/attachment.tsx
```diff
@@ -82,7 +83,7 @@ export const AttachmentProvider = ({
   children,
 }: PropsWithChildren<{
   canUpload: boolean;
-  uploadAsset: (asset: ImagePickerAsset) => Promise<void>;
+  uploadAsset: (asset: ImagePickerAsset, isWeb?: boolean) => Promise<void>;
```
dnbrwstr, line 86: Possible to skip this arg? We could prettily easily add an `isWeb` to the shared package just by having an `isWeb.ts` and `isWeb.native.ts`.

dnbrwstr (approved):
Mostly nits here, none that have to be blocking. Would also be good to add the Reanimated eslint rule as we discussed, and I'd really love if we could create the new package if at all doable.

package.json
```diff
@@ -51,7 +51,9 @@
       "react-native-reanimated@3.8.1": "patches/react-native-reanimated@3.8.1.patch",
       "drizzle-orm@0.30.9": "patches/drizzle-orm@0.30.9.patch",
       "@10play/tentap-editor@0.4.55": "patches/@10play__tentap-editor@0.4.55.patch",
-      "any-ascii@0.3.2": "patches/any-ascii@0.3.2.patch"
+      "any-ascii@0.3.2": "patches/any-ascii@0.3.2.patch",
+      "react-native-gesture-handler@2.18.1": "patches/react-native-gesture-handler@2.18.1.patch",
```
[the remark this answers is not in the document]
patosullivan: Unfortunately gesture-handler is already at the newest version and image-zoom's newest version still uses useAnimatedStyle without a dep array.

PR #3838, opened 2024-08-13
Visual tweaks for Activity screen

jamesacklin:
- Correctly references `props.viewMode` in ContentReference to correctly size reference text in the activity context (we were relying on a similarly-named `useReferenceContext` hook which only gives us the immediate parent `viewMode`, not the one we're passing down through the Activity screen).
- Renames a prop for ContactName (we use `size` rather than `fontSize`) so we don't have medium-sized ship nicknames in the activity item content (which is small-sized)
- Cleans up the inter-item layout of the Activity screen to prevent unsightly wrapping and  margin inconsistencies
- Restores the faint blue background for unread items
- Left-aligns the screen title

Before & after:
image

### DIFF
--- a/packages/ui/src/components/Activity/ActivitySummaryMessage.tsx
+++ b/packages/ui/src/components/Activity/ActivitySummaryMessage.tsx
@@ -26,7 +26,12 @@ function SummaryMessageRaw({
 
   const NewestAuthor = useMemo(() => {
     return (
-      <ContactName fontSize="$s" userId={newest.authorId ?? ''} showNickname />
+      <ContactName
+        fontSize="$s"
+        lineHeight={18}
+        userId={newest.authorId ?? ''}
+        showNickname
+      />
     );
   }, [newest.authorId]);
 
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/Activity/ActivitySummaryMessage.tsx
```diff
@@ -35,19 +40,30 @@ function SummaryMessageRaw({
       <>
         <ContactName
           fontSize="$s"
+          lineHeight={18}
```
jamesacklin, line 43: @dnbrwstr We don't have a value for this in the Tamagui config  and we only use it in this view.

packages/ui/src/components/Activity/ChannelActivitySummary.tsx
```diff
@@ -44,23 +44,22 @@ export function ChannelActivitySummary({
 
   return (
     <View
-      padding="$l"
-      marginBottom="$l"
+      padding="$m"
       backgroundColor={
-        newestPost.timestamp > seenMarker && unreadCount > 0
+        newestPost.timestamp > seenMarker || unreadCount > 0
```
jamesacklin, line 49: @latter-bolden check my logic here?

dnbrwstr (approved):
Lgtm!

PR #3840, opened 2024-08-13
native: reduce overfetching pt 1

arthyn:
This starts TLON-2543, a follow-up will address thread fetching on return. This PR switches out many instances where we used `useCurrentChats` which is a very expensive query for much less expensive queries. Additionally this does some work to make sure that we don't fetch screens that are out of view in the stack. 

One thing we may want to consider here is every screen having `isFocused` in a context so that any queries running could respect it with `enabled: isFocused`.

PR Checklist


arthyn:
working on the merge conflicts

packages/app/hooks/useGroupContext.ts
```diff
@@ -110,6 +121,7 @@ export const useGroupContext = ({ groupId }: { groupId: string }) => {
         title,
         channelType,
         currentChatData,
+        pendingChats,
```
dnbrwstr, line 124: Missing dep below: `pendingChats`

dnbrwstr (approved):
Lgtm! Just nits. Agree on focus state too -- maybe we want a `ScreenContext` or something. Would be good to get @latter-bolden eyes on this too, since he's more familiar with pending chats.

PR #3841, opened 2024-08-13
native: RNTL UI test setup

davidisaaclee:
fixes TLON-2538

- Adds two passing Jest tests which mount components and does some basic checks
  - `HeaderButton.test.tsx` tests `HeaderButton` in isolation (added this as a first step and just kept it around)
  - `App.test.tsx` mounts the main component of `App.main.tsx`, enters self-hosted login flow, enters text into form, submits, accepts EULA, and checks that an `AuthenticatedApp` component was mounted

### Compromises
I had to make a lot of compromises to get these tests to pass – the majority of these push the behavior of the test away from the behavior of the real app, and some were one-off changes to config:
- **Adding Jest instead of using the existing Vitest setup**: here's my rationale pasted from a ticket update (which are all still valid):
  - The majority of getting these tests to work is mocking all of the native modules: using [jest-expo](https://github.com/expo/expo/tree/main/packages/jest-expo) gives us a big step up here, and that is sadly only supported on Jest.
  - RN has issues with Vitest: https://github.com/vitest-dev/vitest/issues/1321 (https://github.com/sheremet-va/vitest-react-native attempts to solve but doesn't seem ready for use)
  - I'm looking to use [@testing-library/react-native](https://testing-library.com/docs/react-native-testing-library/intro/) because it is the most supported, and will be less flaky than "real device" UI testing. RNTL is largely set up for Jest.
- **Lots of mocks!** Native modules are just not loaded when running through Jest, so every dependency touched by the tests that has a native module needed to get mocked. A lot of these significantly change the behavior of these modules.
  - **DB mock**: Since we can't use opsqlite in Jest environment, I had to refactor `nativeDb.ts` to use better-sqlite3 for the app DB. No code is _really_ exercising this, but it appears to work.
  - **`AuthenticatedApp` mock**: When `AuthenticatedApp` mounts, it starts doing a lot of API calls, and had other hard-to-solve Jest issues. To kick the can, I mocked `AuthenticatedApp` to be a simple component, and just checked that it was mounted at the end of the login flow.
  - Some mocks are in `tlon-mobile/__mocks__/`, which I preferred to keep things atomic; but it was easier to write others in a Jest setup file (when needing to export a specific provided mock module). I think more of those could be moved to `__mocks__`, but I didn't want to spend too much time on it.
- **Enabling package exports**: Importing `@urbit/sigil-js/dist/core` was failing through Jest (similarly, `react-tweet/dist/api`). To fix, I:
  - changed the import to `@urbit/sigil-js/core` (which is a defined `exports` in `sigil-js`)
  - enabled `unstable_enablePackageExports` in Metro (which allows referencing these alternative `exports`)
  - set TS `moduleResolution: bundler` in the `ui` package (reacting to `tsc` error after those changes)


apps/tlon-mobile/package.json
```diff
@@ -100,7 +101,7 @@
     "react-native-phone-input": "^1.3.7",
     "react-native-polyfill-globals": "^3.1.0",
     "react-native-reanimated": "^3.8.1",
-    "react-native-safe-area-context": "4.8.2",
+    "react-native-safe-area-context": "^4.9.0",
```
davidisaaclee, line 104: We had `react-native-safe-area-context: ^4.9.0` in another package (`ui`?), which required pnpm to install two separate versions – this broke mocks.

packages/app/lib/betterSqlite3Connection.ts
```diff
@@ -0,0 +1,64 @@
...
+          operation: 'INSERT' | 'DELETE' | 'UPDATE';
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          row?: any;
+          rowId: number;
+        }) => void)
+      | null
+  ): void {
+    console.warn(
+      'BetterSqlite3$SQLiteConnection::updateHook() is not implemented'
+    );
```
davidisaaclee, line 30: We're not testing code that response to subscriptions, so this isn't an issue – but I figured it'd be annoying to try to add a test that relies on `updateHook` and not get a big warning that it's not yet implemented.

jamesacklin:
> To kick the can, I mocked AuthenticatedApp to be a simple component, and just checked that it was mounted at the end of the login flow.

I think this is actually fine—if the auth API call passes, then that means the auth flow hasn't changed substantially (A) on the server or (B) in the JS API. Seems like a totally acceptable solve right now.

Nice work!

dnbrwstr (approved):
Nice, this lgtm. Only bummer is that the tamagui compilation slows things down a bit, which I'm torn on. On the one hand, there's a real possibility that the compilation itself could cause issues, on the other it's really significantly slower. Wdyt?

davidisaaclee:
hm, I might've pushed too soon – I ran the test and saw a much shorter duration, so I thought it was good, but I think Jest was just using a cache. Running from a clear cache, first run of `test-ui` under [Conditionally disable Tamagui optimizing compiler for tests](https://github.com/tloncorp/tlon-apps/pull/3841/commits/26166c2efd4a3af898745d1d65ac1cb3e2aeb46b) is 44s, vs the base of 52s at [268c722](https://github.com/tloncorp/tlon-apps/pull/3841/commits/268c722350d82c4208b5de196dc971505d2d800c). I guess that's still a good gain, but I don't know enough about the reliability of the Tamagui optimizer to say whether this is worth it – wdyt?

davidisaaclee:
> i think let’s skip for now then, since I have seen real issues from the compilation step

-- dan, from chat (2024)

PR #3843, opened 2024-08-14
webapp/sidebar: add Landscape icon + link

jamesacklin:
Screenshot 2024-08-14 at 3 41 53 PM

Adds a way to get out to Landscape from the webapp in anticipation of redirecting through the hosting signup flow directly into the Tlon/%groups app.


arthyn (approved):
lgtm!

apps/tlon-web/src/components/Sidebar/AppNav.tsx
```diff
@@ -357,6 +358,20 @@ function ProfileTab(props: { isInactive: boolean }) {
   );
 }
 
+function LandscapeTab() {
+  return (
+    <div
```
arthyn, line 363: I think you can just make this an `<a>` with `target="_blank"`

PR #3844, opened 2024-08-15
native: fix channel write permission check

latter-bolden:
This was an issue with permissions not getting initially configured correctly. There was a bugged early return preventing writer role insertion. Also fixes a bug I noticed while working on this with the empty channel message still getting displayed inverted.

Fixes TLON-2561


patosullivan (approved):
lgtm!

PR #3846, opened 2024-08-16
Add dev menu items to log sqlite query durations

davidisaaclee:
Not sure if we want to merge this, although I think it'd be safe to do so (we add a call to `QueryLogger#log` in all queries, but that exits early if the logger is paused).

Adds shake menu items to (1) start capturing query durations, and (2) print out (or otherwise share) a CSV representation of the captured query durations for easy import into a visualization spreadsheet.

Exporting the CSV from this tool into a spreadsheet lets us visualize our query load easily (the following was navigating from home screen -> channel):
Screenshot 2024-08-16 at 11 14 41 AM

### DIFF
--- a/packages/app/lib/devMenuItems.ts
+++ b/packages/app/lib/devMenuItems.ts
@@ -1,6 +1,18 @@
 import { registerDevMenuItems } from 'expo-dev-menu';
-import { NativeModules } from 'react-native';
+import {
+  Alert,
+  Clipboard,
+  DevSettings,
+  NativeModules,
+  Share,
+} from 'react-native';
 
+// Since we're relying on state local to this module, we need to ensure that we
+// import from the same location that other callers import it from (i.e. don't
+// import from `@tloncorp/shared/src/db` because `tlon-mobile` largely imports
+// from `@tloncorp/shared/dist/store`, which transitively hits `QueryLogger`
+// via the post-bundler `dist`).
+import { QueryLogger } from '../../../packages/shared/dist/db';
 import { getDbPath, purgeDb } from './nativeDb';
 
 let metroBundlerURL: string | null = null;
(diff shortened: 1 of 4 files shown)


davidisaaclee:
@dnbrwstr @arthyn here's the code that I used to measure those queries

davidisaaclee:
I think some commit made queries significantly better: here's another set of measurements from idly tapping around in the app on a7b100ae191ffdb60035972d9f0c13beada4c091 https://docs.google.com/spreadsheets/d/1AhBdRMn5HWbclZiYl9OBNfpEfn8zoGDWR_ggv6VBozg/edit?usp=sharing
Screenshot 2024-08-19 at 9 37 21 AM

dnbrwstr:
Wow interesting, that's pretty surprising -- the diff is just this? https://github.com/tloncorp/tlon-apps/compare/dil/csv-query-logs...develop ? I guess it could just be the removal of all the sheet rendering overhead...

dnbrwstr:
It'd be interested to see stacked results for this too -- like what's the time spent in all queries per slice of time.

davidisaaclee:
@dnbrwstr yes, the diff is surprising! maybe the measurement from last week had some irrelevant work going on?

re: stacked – I just gave you edit access to the sheet, but I'll give that a shot in a minute

davidisaaclee:
Screenshot 2024-08-19 at 10 32 54 AM
https://docs.google.com/spreadsheets/d/1AhBdRMn5HWbclZiYl9OBNfpEfn8zoGDWR_ggv6VBozg/edit?gid=251621322#gid=251621322

pretty sure this is right (i'm no spreadsheet wiz) - 2sec buckets

insertChannelPosts, getPostsWithRelations, and getGroup seem to be the biggest offenders

PR #3848, opened 2024-08-19
native: use zustand for feature flags to enable reactivity

davidisaaclee:
- Changes backing store of feature flags to a Zustand store (publishes changes easily, can be accessed from outside of React, and was already installed)
- Flag changes are written to local storage whenever Zustand store is mutated
- Added `useFeatureFlag` hook that acts like `useState`, but for feature flags
- Update call sites to use reactive API

I tested by toggling the experimental channel switcher flag, checking that it took effect in the app, toggled it back off, did the same and quit the app, checked my pref was saved.

I also inserted a console.log in the subscription and checked that it doesn't appear to double-trigger after navigating around the app. Hypothetical bug here (which I did not see, but not convinced it's a non-issue): since we're registering a listener in the top-level scope of the module, it's possible that importing the containing module multiple times could run that top-level `subscribe` multiple times, ending up with redundant subscriptions.

### DIFF
--- a/packages/app/lib/featureFlags.ts
+++ b/packages/app/lib/featureFlags.ts
@@ -1,11 +1,15 @@
+import { mapValues } from 'lodash';
+import create from 'zustand';
+
 import storage from './storage';
 
+// Add new feature flags here:
 export const featureMeta = {
   channelSwitcher: {
     default: false,
     label: 'Experimental channel switcher',
   },
-};
+} satisfies Record<string, { default: boolean; label: string }>;
 
 export type FeatureName = keyof typeof featureMeta;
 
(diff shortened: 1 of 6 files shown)


packages/app/lib/featureFlags.ts
```diff
@@ -1,33 +1,54 @@
...
-const featureState: { [K in FeatureName]: boolean } = {
-  channelSwitcher: false,
-};
+interface FeatureStateStore {
+  flags: FeatureState;
+  setEnabled: (name: FeatureName, enabled: boolean) => void;
+}
+
+export const useFeatureFlagStore = create<FeatureStateStore>((set) => ({
+  flags: mapValues(featureMeta, (meta) => meta.default),
```
davidisaaclee, line 26: @dnbrwstr "fixed" default types by deferring to `mapValues` – I think in theory this is unsound (e.g. someone could import `featureMeta` and add an entry, which would break a bunch of these types), so w/e, it works lol

packages/app/package.json
```diff
@@ -13,5 +13,9 @@
   "devDependencies": {
     "@types/react": "^18.2.55",
     "@types/react-native": "0.73.0"
+  },
+  "peerDependencies": {
+    "lodash": "^4.17.21",
+    "zustand": "^3.7.2"
```
davidisaaclee, line 19: we aren't listing a lot of the dependencies we import from `packages/app/src/` here – idk best practice but it seems best to be explicit about what's needed (e.g. let's say we add a new package `apps/tlon-vr` and we want to import `@tloncorp/apps` – having `peerDependencies` listed would let us know what we need in that new package)

dnbrwstr (approved):
Great! Lgtm.

PR #3849, opened 2024-08-19
native: fix stale unreads

latter-bolden:
fixes two issues:
#### stale unread count after viewing a channel when you return to the chatlist
 We now invalidate any queries like before instead of specifically targeting active ones. This was tripping things up and is the least risky way to resolve this in the short term.

#### incorrect unreads when you enter a channel subsequently after viewing it once
This was an issue with us getting stuck on the cached value in react query before it refetched. A simpler approach of just manually fetching the initial unread one time on mount is used instead.

Fixes TLON-2566


dnbrwstr:
Lgtm to me overall, one question: how does this interact with the channel switcher? That's the only situation I can think of where we might want to refetch unread state.

latter-bolden:
> how does this interact with the channel switcher?

good thought! Updated. This should clean up a pre-existing bug with the cursor logic in that context as well

arthyn (approved):
lgtm

PR #3851, opened 2024-08-20
onboarding: convert screens to use @tloncorp/ui components

jamesacklin:
Fixes TLON-2585

- Converts (almost) all UI components in the onboarding flow to use `@tloncorp/ui` equivalents
  - Form fields are all wrapped in the Field component and use the standard TextInput component
  - The PhoneNumber component in RequestPhoneVerify is complex enough that if we want to factor it out in a separate PR, we should
  - We should probably pull in the CountrySelector or wrap it in some way. We only use it here
  - We are still using the react-native Switch on SetTelemetryScreen, but we have plans to remove that screen altogether anyway
- Improves validation on some form fields (including hiding the "Next" button in the header until the entire form on each screen is valid)
- Removes all uses of `tailwind-rn` and converts to native styling where appropriate
- Removes LoadingSpinner, TlonButton, and HeaderButton from `@/apps/tlon-mobile` as they have been superseded by versions from `@tloncorp/ui`

Preview videos are in the Homestead channel on the network since they contains my phone number and email address(es).


apps/tlon-mobile/src/App.main.tsx
```diff
@@ -61,11 +59,19 @@ const App = ({
...
+
   const { isLoading, isAuthenticated } = useShip();
   const [connected, setConnected] = useState(true);
   const { lure, priorityToken } = useBranch();
   const screenOptions = useScreenOptions();
+
+  const onboardingScreenOptions = {
+    ...screenOptions,
+    headerShown: false,
+    animation: 'simple_push' as const,
```
dnbrwstr, line 71: Curious why we need animation settings here, since it doesn't seem like we're setting them on other stacks?

dnbrwstr (changes requested):
Overall looks good! one nit, and two questions:
- Can we extract any of the screen wrapper bits into shared components? Seems like there's some duplication there.
- Can we use the form elements defined in `Form.tsx` for field labels, errors, etc.? If not, I'd be curious why they don't work, since I'd like them to be fully general.

jamesacklin:
@dnbrwstr Absolutely. Want do do this stepwise with another PR, or should I do this now?

dnbrwstr:
Probably just do now? But either way works; the only thing I'd say is blocking is possibly the screen animation piece -- just want to make sure we're handling consistently.

apps/tlon-mobile/src/App.main.tsx
jamesacklin: No good reason, left it accidentally while I was playing around. Fixed.

jamesacklin:
I'll only push back on the screen wrapper components slightly because the structure mirrors those in, say, `EditProfileScreenView`—we have a flex-view, a GenericHeader, a keyboard-avoiding view, a scroll view, and a Y-stack. This is also the case in any of the `@tloncorp/ui` "screen views," we have a lot of duplication there as well. Not all that much different than these screens.

Happy to take care of the form stuff in this PR.

dnbrwstr:
> I'll only push back on the screen wrapper components slightly because the structure mirrors those in, say, `EditProfileScreenView`—we have a flex-view, a GenericHeader, a keyboard-avoiding view, a scroll view, and a Y-stack. This is also the case in any of the `@tloncorp/ui` "screen views," we have a lot of duplication there as well. Not all that much different than these screens.

Ultimately we should extract from all those screens (if styling/mechanics match) 😈 , but totally fine to punt for now, especially if there's not an obvious/easy way to do it here.

dnbrwstr (approved):
Nice, this is looking good. If you have time, I'd suggest converting a number of the `TextInputs` to the versions in `Form/controlledFields`, but that doesn't need to block -- I know you're out in a little bit, and we can always grab on another pass.

PR #3853, opened 2024-08-21
channel: improve return loading performance

arthyn:
This fixes the most egregious case of TLON-2543 which was that we made a request for every post in a channel who's sync time was less than the current session time when returning to a channel. To prevent this, we made a change in tloncorp/tlon-apps#3847 to a previously written but unused scry endpoint "changes" so that we could hand it a start and end post cursor and then an "after" time. This lets us hand it a range of posts and say give us back any posts that have changed since this time. 

This seems to speed things up when running it, but hard to tell with a local moon. Could use some help on how to get a timing here that we could measure.

Also I did a small drive-by and refactored @urbit/api out of the shared lib, so now it only uses @urbit/aura

### DIFF
--- a/apps/tlon-mobile/ios/Podfile.lock
+++ b/apps/tlon-mobile/ios/Podfile.lock
@@ -1130,7 +1130,7 @@ PODS:
     - React-Core
   - react-native-netinfo (11.1.0):
     - React-Core
-  - react-native-safe-area-context (4.9.0):
+  - react-native-safe-area-context (4.8.2):
     - React-Core
   - react-native-webview (13.6.4):
     - glog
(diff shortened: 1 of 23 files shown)


packages/shared/src/api/urbit.ts
```diff
@@ -79,7 +78,7 @@ export function configureClient({
   clientInstance.on('fact', (fact) => {
     logger.log(
       'received message',
-      runIfDev(() => escapeLog(JSON.stringify(fact)))
+      runIfDev(() => escapeLog(JSON.stringify(fact)))()
```
dnbrwstr, line 81: Curious why the second set of parens is necessary here -- afaik `runIfDev` should execute the function immediately in dev environment?

apps/tlon-mobile/ios/Podfile.lock
```diff
@@ -1130,7 +1130,7 @@ PODS:
     - React-Core
   - react-native-netinfo (11.1.0):
     - React-Core
-  - react-native-safe-area-context (4.9.0):
```
dnbrwstr, line 1133: Likely unintentional change?

packages/shared/src/api/postsApi.ts
```diff
@@ -429,8 +431,9 @@ export const getChangedPosts = async ({
     app: 'channels',
     path: formatScryPath(
       `v1/${channelId}/posts/changes`,
-      afterCursor ? formatCursor(afterCursor) : null,
-      count
+      formatCursor(startCursor),
+      formatCursor(endCursor),
+      formatUd(unixToDa(afterTime.valueOf()).toString())
```
dnbrwstr, line 436: Could be `formatDateParam` (I think)

dnbrwstr (approved):
Lgtm! A few little nitpicks. Have we rolled out the new api to internal? Would be good to have that in before merging.

PR #3854, opened 2024-08-21
native: fix push notif routing

latter-bolden:
We weren't fully resetting the nav stack when handling push routing. We now fully clear navigation before pushing on the notif path for all cases which fixes the situation reported in the issue convo.

Fixes TLON-2232


patosullivan (approved):
Looks good!

PR #3857, opened 2024-08-21
native: dim header when session inactive

dnbrwstr:
Visually indicates whether we have an active session using screen header (grays it out if we haven't initialized session). Fixes TLON-2254; more advanced introspection soon.


latter-bolden (approved):
Nice!

PR #3858, opened 2024-08-22
native: hotfixes

latter-bolden:
Grab bag of small fixes that came up in feedback with the last build.

Fixes TLON-2586 (channels freeze on white screen)
Fixes TLON-2580 (missing close overlay on image viewer)
Fixes TLON-2589 (reactions not inserted when adding posts)


arthyn (approved):
makes sense lgtm

PR #3859, opened 2024-08-23
native: forward errors from notification extension to posthog

davidisaaclee:
- NSE doesn't have error logging to any remote target (Posthog, Crashlytics, etc)
- To hack this in, I added a `notificationServiceExtensionErrors` field on the user payload of the notification we present via the NSE, and use that to hold any errors we want to log
- When user taps on the notification, we check if that field is available, and log it to Posthog if so

My guess right now is that we are failing to forward the latest cookies from host to extension – we should see an auth error in Posthog if that's the case. (If we do see that, we won't have enough information to have a concrete fix...)

Tested by inserting a fake error in the NSE code, and checked that tapping on the failed notification ends up with the following in Posthog events dashboard:
Screenshot 2024-08-22 at 5 56 51 PM


apps/tlon-mobile/src/hooks/useNotificationListener.ts
```diff
@@ -20,23 +23,68 @@ type RouteStack = {
...
+
+  if (payload == null || typeof payload !== 'object') {
+    return null;
+  }
+
+  const baseNotificationData: BaseNotificationData = {
+    meta: { errorsFromExtension: payload.notificationServiceExtensionErrors },
+  };
+
+  // welcome to my validation library
```
latter-bolden, line 66: 🤭

latter-bolden (approved):
Code looks great!

PR #3861, opened 2024-08-26
channels: bad mark

arthyn:
OTT, this mark should have changed when we did all our new types


Fang- (approved):
"Are you sure this mark is immutable?"
"It's a good mark sir."
Look inside the mark. Its type changed.

Good catch, thank you.

PR #3862, opened 2024-08-26
old web: fix backspace issue in input

patosullivan:
Fixes TLON-2600

Turns out this was caused by a version mismatch for prosemirror-state that was causing the history plugin in tiptap not to load/work.

Tested to make sure everything still works w/the mobile input and in the new app, all works fine.


latter-bolden (approved):

patosullivan:
The type check fail issue is weird. `pnpm -r tsc` runs fine locally. Looking into it.

patosullivan:
Ok, *this* seems to be the right incantation. Everything works in local testing, no more tsc fails.

PR #3864, opened 2024-08-26
native: editable group privacy

latter-bolden:
Adds an option to the _Group Options Sheet_ to modify the group's privacy if you're an admin. Looks like this:

https://github.com/user-attachments/assets/78f75d1e-893c-4959-9a5f-066223786d5e

Note, in the mocks there was also group invite functionality on this screen, but that has its own ticket (TLON-1717).

Fixes TLON-1713

### DIFF
--- a/apps/tlon-mobile/src/screens/GroupSettings/InvitesAndPrivacyScreen.tsx
+++ b/apps/tlon-mobile/src/screens/GroupSettings/InvitesAndPrivacyScreen.tsx
@@ -1,6 +1,9 @@
 import { NativeStackScreenProps } from '@react-navigation/native-stack';
-import { Text } from '@tloncorp/ui';
-import { SafeAreaView } from 'react-native-safe-area-context';
+import { GroupPrivacy } from '@tloncorp/shared/dist/db/schema';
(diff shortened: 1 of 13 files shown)


apps/tlon-mobile/src/screens/GroupSettings/InvitesAndPrivacyScreen.tsx
```diff
@@ -1,6 +1,14 @@
 import { NativeStackScreenProps } from '@react-navigation/native-stack';
```
patosullivan, line 1: Maybe we rename this file?

packages/ui/src/components/ChatOptionsSheet.tsx
```diff
@@ -121,6 +121,7 @@ export function GroupOptions({
     onPressGroupMembers,
     onPressGroupMeta,
     onPressManageChannels,
+    onPressInvitesAndPrivacy,
```
patosullivan, line 124: Probably also rename

patosullivan (approved):
Looks good, just a few nits about names.

PR #3866, opened 2024-08-28
native: account for new hosting auth

latter-bolden:
Hosting is changing the way they handle cookies. All current ones will become invalid and the new system will use a token that lasts a full year but expires after 30 days of no use. We want to make sure this doesn't immediately impact app users.

This PR adds a check whenever the app returns from background that pings hosting once a day to make sure we continue refreshing the cookie and track whether our hosting session is expired. 

At the moment the only place where we require hosting auth is on the _Manage Account_ screen. For that case, if your hosting session is expired we pop an alert informing the user that they'll need to log back in if they want to manage their account (in which case we'll have a fresh session).

Fixes TLON-2605


patosullivan (approved):
Looks good!

jamesacklin:
@latter-bolden Feel free to merge and start baking another internal build if you want to get this out and unblock Hosting.

PR #3867, opened 2024-08-28
native: on-network and lure invites 

patosullivan:
Fixes TLON-2526, using mostly the same zustand store for lure we used on web, requires https://github.com/tloncorp/landscape/pull/277 to go in as well if we want it to be a bit quicker (not having to wait on a timeout for the `token-link/${flag}` subscribeOnce call).

Also fixes:
- an issue where our our subscribeOnce in the api/urbit module was attempting to use subscribeOnce on `client` rather than `clientInstance` (this was causing the `true` to appear to return for the `token-link` subscribeOnce call, it was actually the response from the `group-enabled` subscribeOnce)
- an issue where we weren't passing contacts into the AppDataContext for the AddGroupSheet, so you couldn't actually search by contact to find a group.
- lack of verbosity on logging from the urbit api module (we turn on verbose if we're in dev mode now, will only be seen if you turn on the logger in the urbit module).

*Note*: this automatically enables lure invites any time you attempt to invite someone to a group. We need a way for admins to override this somewhere else in the app (cc: @jamesacklin)

Also note that this can't be tested on android at the moment because no chat list options work on Android in develop for some reason.


patosullivan:
FYI, I didn't use zustand persist w/AsyncStorage because of this issue: https://stackoverflow.com/questions/78262236/import-meta-is-currently-unsupported-in-eas-build

We're already on 3.7.2, so I'm not sure why that worked for them. It looks like we need to update to the latest zustand, and I thought that might open up a can of worms with our other uses of zustand.

latter-bolden (approved):
Worked on device! Can send invites and generate lure links

I'm noticing that same issue you mentioned on the context fix where certain option actions are shown inconsistently, but I think we can handle that in a followup.

PR #3868, opened 2024-08-28
post content overhaul

dnbrwstr:
This PR overhauls the display of post content. I expect there to be some minor glitchiness as it settles; I've added a "copy json" option to the message long press menu when running in dev mode. If you see anything weird,  please grab a screenshot + the associated json and let me know. 

Other changes of note
- Adds a video player component
- Makes codeblock scroll work
- Correctly renders lists and notebook posts
- Reworks notebook detail view to match design
- Patches UIFont to allow us to use the system monospace font with react native
- Adds a `Text` fixture which automates most of precomputing margin values for text trimming.


packages/ui/src/components/TextV2/Text.tsx
jamesacklin, line 1: Oh boy!

PR #3869, opened 2024-08-29
new web: update globber

patosullivan:
Fixes TLON-2590 by adding the wasm mark to the globber. Tested locally by creating a glob, uploading it to my own s3, then pointing the desk.docket-0 file at the glob and committing it on a local moon.

Also needed to add a jpg.hoon mark (which apparently has never existed? wat) because the new-web-app includes `welcome_flowers.jpg` in the dist (because it's referenced in `WelcomeSheet.web.tsx`).


arthyn (approved):
lgtm

PR #3871, opened 2024-08-29
onboarding: less-intrusive EULA

jamesacklin:
- Adds a CheckboxInput to the Form components
- Adds an optional boolean to GenericHeader to disable the session check
- Moves all onboarding screens in apps/tlon-mobile to their own directory
- Removes all setShip and onboarding flow logic from the EULA screen
- Adds a checkbox to the set password screen, the login via email screen, and the self-hosted ship connection screen that the user can check to agree to the EULA
- Adds ListItem (with a right-chevron) to let the user read the EULA
- Updates the EULA to reflect this checkbox change
- Adds new copy and formatting to the first screen in the email signup flow
- Explicitly sets a text color for text inputs

Fixes TLON-2172, fixes COMMS-351, fixes TLON-2621

Screenshots:

Screenshot 2024-08-29 at 7 25 28 PM

Screenshot 2024-08-29 at 7 14 03 PM

Screenshot 2024-08-29 at 7 14 18 PM

Screenshot 2024-08-29 at 7 15 24 PM

Screenshot 2024-08-29 at 7 15 27 PM


packages/ui/src/components/Form/inputs.tsx
```diff
@@ -242,3 +242,101 @@ function ListItemInputRow<T>({
...
+          <LabelText size="$m" color="$secondaryText">
+            {option.description}
+          </LabelText>
+        ) : null}
+      </YStack>
+    </RadioInputRowFrame>
+  );
+}
+
+const CheckboxControlFrame = styled(View, {
```
latter-bolden, line 311: `CheckboxControlFrame` and `CheckboxControl` look pretty similar to the radio ones. Could we avoid the duplication by allowing style override on the frame via props? Looks like border radius is the only difference?

latter-bolden (approved):
Looks good! Would like to avoid component creep in `Form/inputs`, but otherwise no issues.

I like the orientation info above the email. Could see it being it's own dedicated screen before we prompt for account info. Seems like a common pattern elsewhere.

packages/ui/src/components/Form/inputs.tsx
jamesacklin: Thanks, took care of that!

jamesacklin:
Not sure why CI is failing...

```
packages/ui lint: /home/runner/work/tlon-apps/tlon-apps/packages/ui/src/components/Form/inputs.tsx
packages/ui lint:   313:10  error  'LabelText' is not defined  react/jsx-no-undef
packages/ui lint:   317:12  error  'LabelText' is not defined  react/jsx-no-undef
```

PR #3872, opened 2024-08-30
native: fix chat actions sheet on Android

latter-bolden:
This was another instance of the Android modal context bug (see TLON-2618 for details). The pattern of using context to grab these handlers feels good, so I wanted to avoid removing it in hopes we can return to normal use once the bug gets squashed (sounded like Facebook RN core team has it on their radar).

To that end, the workaround introduced is to update the sheets to take a new prop that matches the missing context value. The context file itself exports a hook for returning the needed value in addition to its existing provider based API. My thinking was this way we can avoid modifying the business logic as much as possible and instead just remove the new prop threading once the normal usage is restored.

Open to better ideas on solutions if y'all have any!

Fixes TLON-2618

### DIFF
--- a/packages/app/features/top/ChatListScreen.tsx
+++ b/packages/app/features/top/ChatListScreen.tsx
@@ -18,6 +18,7 @@ import {
   StartDmSheet,
   View,
   WelcomeSheet,
+  useChatOptionsContextValue,
 } from '@tloncorp/ui';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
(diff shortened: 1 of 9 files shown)


patosullivan (approved):
Solution seems fine to me, and it works on both platforms now.

One other thing I noticed (and this issue exists on develop as well), we don't see any admin options on groups where we should on Android. Might be outside the scope of this PR.

dnbrwstr:
Lgtm, but wanna grab me a sec before you merge in?

latter-bolden:
Going to ship a patch in the demo branch and close this in favor of other approaches.

PR #3875, opened 2024-09-02
native: dump/restore SQLite database via dev menu

davidisaaclee:
| Simulator | Physical device |
| --- | --- |
|  |  |

- Restores / adds dev menu items to dump / restore SQLite database, relying on database paths on Metro host
- Adds `.env` entries (illustrated in `.env.sample`) for dump / restore paths
- Adds check to dev menu setup to only add simulator-exclusive items when `DeviceInfo.isEmulator`

### DIFF
--- a/packages/app/lib/devMenuItems.ts
+++ b/packages/app/lib/devMenuItems.ts
@@ -1,6 +1,7 @@
 import * as api from '@tloncorp/shared/dist/api';
 import { registerDevMenuItems } from 'expo-dev-menu';
 import { Alert, DevSettings, NativeModules } from 'react-native';
+import * as DeviceInfo from 'react-native-device-info';
 
 import { getDbPath, purgeDb } from './nativeDb';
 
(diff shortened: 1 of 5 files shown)


packages/app/lib/devMenuItems.ts
```diff
@@ -9,27 +11,77 @@ if (__DEV__) {
...
       const path = getDbPath() ?? '';
       sendBundlerRequest('open-sqlite', { path });
     },
   },
   {
-    name: 'Dump sqlite database',
+    name: 'Dump SQLite',
     callback: async () => {
-      const path = getDbPath() ?? '';
-      sendBundlerRequest('dump-sqlite', { path });
```
davidisaaclee, line 28: I didn't see any way that this was functional before, lmk if there's something I missed

packages/app/lib/devMenuItems.ts
```diff
@@ -9,27 +11,77 @@ if (__DEV__) {
...
+            },
+          },
+          { text: 'Cancel', style: 'cancel' },
+        ]
+      );
     },
   },
 ];
 
+const devMenuItems: Promise<ExpoDevMenuItem[]> = DeviceInfo.isEmulator().then(
```
davidisaaclee, line 75: there is a synchronous form of `isEmulator`, but I chose async to avoid possibly blocking initial app load

dnbrwstr (approved):
Great! Lgtm.

PR #3877, opened 2024-09-02
Disable newlines in ListingContent

neonfuz:
Fixes bug where lists are too spaced out because there's an extra line break added to the end of every item.

The data retrieved from the backend includes line breaks at the end of every item in a list. Rendering this line break as a <br> makes the items in a list too spaced out, because <li> elements already have their own vertical spacing.

This PR filters line breaks at the end of list items on the front end, instead of changing the back end code. I'm not sure if this is the preferred solution, so merge at your discretion. 

Before:
Screenshot 2024-09-02 at 3 12 25 PM

After:
Screenshot 2024-09-02 at 3 07 40 PM

The state with said line breaks:
Screenshot 2024-09-02 at 2 52 00 PM

Tested locally with fakeship

PR Checklist


latter-bolden (approved):
Looks good to me! Confirmed the line breaks get removed.

PR #3878, opened 2024-09-03
native: increase channel scroll position stability when loading to anchor

davidisaaclee:
| Description | Before | After |
| --- | --- | --- |
| **Simulated load `around` anchor** – here's the worst case I could find for the "before" case - most of the time it's relatively stable with a small blip in position. | <video src="https://github.com/user-attachments/assets/0bef08a1-9de0-41e6-9830-a059e29d704c" /> | <video src="https://github.com/user-attachments/assets/19fb6b3b-07e9-469b-944f-96ea7e3f520d" /> |
| **Load "blitz"** – load posts out of order, with anchor not in first page. This is an unlikely scenario, but illustrates what things look like in a worst-case, possibly buggy release. | <video src="https://github.com/user-attachments/assets/c73e8d7e-52c5-4ab0-a014-b6f7d38e56f1" /> | <video src="https://github.com/user-attachments/assets/fee5e769-ca6d-4da6-b03d-8db1eac1ae1d" /> |
| **Load newer posts while locked to anchor** – after loading channel with anchor locked to top, I pressed the `Load newer` button that adds posts to the "new" end of the timeline | <video src="https://github.com/user-attachments/assets/3615d66d-6676-4a6d-9e2a-c2719781125e" /> | <video src="https://github.com/user-attachments/assets/e6ff584a-9e87-46f1-9ce1-26cf8952ee23" /> |

- [extra] Replace all `Math.random` with a seeded random in `fakeData.ts` – this ensures that Cosmos fixtures always look the same (unless you change the seed).
- Added two `Channel` fixtures: one that simply loads a chat with an unread anchor, and one that has a toolbar for interactively loading posts in different ways.
- Reorganization of anchor scroll lock code – hidden in this reorg are the following key changes:
  - We previously attempted to scroll to anchor whenever _any list item_ got laid out (and we believed we had an `anchorIndex`) – **removed attempts to scroll when non-anchor list items are laid out**. (My guess is that this used to be the case to emulate `maintainVisibleContentPosition` as list items moved the anchor around, but we can rely on mvcp now.)
  - Previously, when a scroll-to-anchor failed, we scrolled to a "best guess" of the anchor and rested (probably relying on the "did layout" scroll attempt to fix stuff). Now, **on a scroll-to-anchor failure, we mark ourselves as explicitly `needsScrollToAnchor = true`, and wait for an anchor layout to trigger a scroll.**

I kept the existing code for `autoscrollToTopThreshold`, but I believe that that feature is broken in the build of RN that we are using – we should apply https://github.com/facebook/react-native/pull/38245 (tested in a quick-n-dirty patch and appears to work)

Content is still shifting a little bit as things load in: I believe this is due to images in messages resizing themselves, and thus their `ChatMessage` containers.

I've logged into this client on my live ship and browsed around to channels with lots of unreads, and some with no unreads – things seem fairly stable. (Did the same with current release client and it is less scroll-stable.)


galenwp:
Heroic

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -169,62 +169,27 @@ const Scroller = forwardRef(
...
+      channelType,
+    });
 
     const theme = useTheme();
 
     // Used to hide the scroller until we've found the anchor post.
     const style = useMemo(() => {
       return {
-        opacity: hasFoundAnchor ? 1 : 0,
+        opacity: readyToDisplayPosts ? 1 : 0,
```
latter-bolden, line 189: I think this is firing too frequently. When searching for results far in scrollback and scrolling down, I'm noticing a blank screen/scroll thrash immediately after loading newer posts.

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -717,3 +634,205 @@ const PressableMessage = React.memo(
...
+      channelType === 'dm' ||
+      channelType === 'groupDm'
+      ? {
+          minIndexForVisible: 0,
+          // If this is set to a number, the list will scroll to the bottom (or top,
+          // if not inverted) when the list height changes. This is undesirable when
+          // we're starting at an older post and scrolling down towards newer ones,
+          // as it will trigger on every new page load, causing jumping. Instead, we
+          // only enable it when there's nothing newer left to load (so, for new incoming messages only).
+          autoscrollToTopThreshold:
```
latter-bolden, line 777: potentially related to the previous comment?

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -717,3 +634,205 @@ const PressableMessage = React.memo(
...
+            !userHasScrolled || hasNewerPosts ? undefined : 0,
+        }
+      : undefined;
+  }, [userHasScrolled, hasNewerPosts, channelType]);
+
+  const handleScrollToIndexFailed = useCallback(
```
latter-bolden, line 783: in theory, should this only ever be called once per unique anchor?

latter-bolden (changes requested):
This is much easier to read through! Appreciate the copious amount of comments <3 

The posts disappearing during search scrollback feels like a regression we should address before merging, but otherwise this feels like a much better base to build off of.

packages/ui/src/components/Channel/Scroller.tsx
davidisaaclee: I *think* there is an edge case where this is called multiple times:
1. Attempt to scroll to anchor index
2. Scroll to index fails, causing a "best effort" scrollToOffset
3. We later think we can scroll correctly, so we do scrollToAnchorIfNeeded
4. but that scroll to index fails again, looping back to step 2

I'm not sure if this would happen in practice (one hypothetical situation is if extra list content gets loaded in between the second invocation of scrollToAnchorIfNeeded and the actual scrollToIndex which changes the layout of the anchor)

packages/ui/src/components/Channel/Scroller.tsx
davidisaaclee: chatted in DM – I'm seeing this as well when jumping to an older message via search -> tap, then scrolling down towards newer messages – I agree that this is a blocker, and it's proving to be tricky, so I thought I'd do a lil check in + notes:

- This doesn't appear to be related to `autoscrollToTopThreshold`, as `maintainVisibleContentPositionConfig` does not change after user scrolls for the first time.
- Posts are loading in as expected (+50 posts to the appropriate side with no stray null `posts`)
- We hide the scroller until it's ready to show, so I thought there might be a bug there – that appears to be working as expected

My current hypothesis is that this is due to `FlatList`'s async rendering combined with `maintainVisibleContentPositionConfig`. During the bug, the scroll view is technically scrolling >5000pts instantaneously:
1. We're adding 50 new posts once we hit the bottom of the scroll
2. which grows the scroll content size
3. which naturally changes what content the viewport is looking at
4. `maintainVisibleContentPositionConfig` jumps that 5000px to move the viewport back to the content we were originally looking at

Since `FlatList` discards offscreen content to make way for onscreen content, I think we rapidly discarding old content for new content, then immediately jumping back to that "older" content and redrawing it.

I don't know why this wasn't happening before (confirmed that the current released client does not have this issue).

davidisaaclee: total gamer moment, looks like this is happening on `develop`, and it is also happening on release client (although it's faster on release).

davidisaaclee:
In the before case, I didn't record jumping to the selection anchor from search page – it anchored correctly to the item, but only after scroll position jumped up and down a few times.

| Current Testflight client | 58177658f5db9100b21b7897f1a726f45c63f847 |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/4f398b12-84a4-4daf-b32e-bc6aa5c95d3d" /> | <video src="https://github.com/user-attachments/assets/31a775a4-5415-4b6e-9859-46795951d420" /> |

There are still some white flashes at 58177658f5db9100b21b7897f1a726f45c63f847, which I tried to capture but could not – in any case, I think the flashes are much briefer.

packages/app/features/top/ChannelScreen.tsx
```diff
@@ -147,7 +147,7 @@ export default function ChannelScreen({
   } = store.useChannelPosts({
     enabled: !!channel && !channel?.isPendingChannel,
     channelId: currentChannelId,
-    count: 50,
+    count: 15,
```
davidisaaclee, line 150: cc @dnbrwstr – I saw that the previous value here was explicitly set in one of your commits – do you see any bad effect of using a smaller value here? (more context in main PR thread but gist is that loading too much alongside `maintainVisibleContentPosition` was causing white flashes (i think!)

packages/app/features/top/ChannelScreen.tsx
```diff
@@ -247,6 +247,16 @@ export default function ChannelScreen({
...
   );
+  const loadNewerUnlessAlreadyLoading = useCallback(() => {
+    if (!isLoadingPosts) {
+      loadNewer();
+    }
+  }, [loadNewer, isLoadingPosts]);
+  const loadOlderUnlessAlreadyLoading = useCallback(() => {
+    if (!isLoadingPosts) {
+      loadOlder();
+    }
```
davidisaaclee, line 258: i chose not to queue loads here – if we queued loads, then we'd have close to the same behavior as _not_ guarding on `isLoadingPosts`:
1. user scrolls to bottom of list
2. load starts
3. user is still at bottom of list (nudges scroll maybe)
4. `onScrollEndReached` called a second time – we queue the request to load more
5. load from step 2 completes; we immediately start the queued load, even though the user already has posts in that direction

With the current simple implementation,
1. user scrolls to bottom of list
2. load starts
3. user is still at bottom of list (nudges scroll maybe)
4. `onScrollEndReached` called a second time – we discard the request to load more
5. load from step 2 completes, user has items
and worse case:
1. user scrolls to top of list
2. load starts for older posts
3. user changes mind, scrolls to bottom of list
4. load request for newer posts is discarded since load is active
5. load from step 2 completes, but they are not in user's view
6. if user still wants newer posts, they'll probably be rage-scrolling down :/ so they'll eventually get the posts

not great, but i think workable for now?

latter-bolden: Make sense, definitely workable for now. 

I'd expect worst case direction switch is somewhat infrequent. Anecdotally, I only really do that if the first load is super slow.

latter-bolden (approved):
Changes make sense to me. Likewise curious if you'd have feedback on the modified post fetch count @dnbrwstr.

Excellent work on the fixes! Was feeling _really_ good on my dev build.

packages/app/features/top/ChannelScreen.tsx
dnbrwstr: Nope, whatever works, works! Just trying to tune for max perf.

PR #3880, opened 2024-09-03
profile: allow saving avatar/cover changes

jamesacklin:
Adds `shouldDirty: true` to onSetCoverURL and onSetIconURL in the profile editor, which enables the Save button (which only does anything if the form itself is dirty).

Also hides the Save button until the form is dirty to stay on-pattern w/ elsewhere in the app.


latter-bolden (approved):
Nice!

PR #3881, opened 2024-09-03
onboarding: ask for notification access during nickname-setting

jamesacklin:
- Moves a couple onboarding screens I missed during the first pass
- Moves the notification access prompt to when the user enters the SetNickname screen
- Removes the notification access request screen altogether
- Passes the notification token to the next screen in flow (telemetry)


dnbrwstr (approved):
Nice, lgtm!

PR #3884, opened 2024-09-04
native: fix context passing for Sheets

latter-bolden:
Uses a new strategy for Sheets that maintains the modal like feel while avoiding portal'ing the component to another location in the tree. This means context passing works as expected on both platforms.

I updated existing components that were using context instantiation hacks. The one exception is _AddGroupSheet_ which mounts its own navigation container making a refactor to use the now standard _ActionSheet_ non-trivial. I don't think we should use that pattern elsewhere (we should mimic that effect in Reanimated where needed) , but rewriting it entirely seemed outside the scope of this PR.

I tried to audit all the existing sheets to make sure this didn't cause any display weirdness, but let me know if anyone sees any regressions during review!

Fixes chat actions and reaction viewing on Android as a sideffect.

Fixes TLON-2657

### DIFF
--- a/packages/ui/src/components/ActionSheet.tsx
+++ b/packages/ui/src/components/ActionSheet.tsx
@@ -8,6 +8,7 @@ import {
   useMemo,
   useRef,
 } from 'react';
+import { Modal } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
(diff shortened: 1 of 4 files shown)


dnbrwstr (approved):
Great, lgtm!

PR #3885, opened 2024-09-04
native: fix editing messages

latter-bolden:
On edit, we weren't stringifying the JSON content before writing to the DB as part of the optimistic update. This broke things after the recent content render changes due to an unprotected JSON parse.

Fixes TLON-2627


jamesacklin:
Makes total sense. Thanks for the fix!

dnbrwstr (approved):
Lgtm!

PR #3889, opened 2024-09-05
detail view: allow navigating to profile

jamesacklin:
As it turns out, we had all the click-handling set up in AuthorRow but weren't passing anything down through the view (from the controller) to handle navigating to a profile from a detail view. Fixes TLON-2208.

Video evidence:

https://github.com/user-attachments/assets/bfddcab7-e08c-4138-98dc-2d5a2a427a17


dnbrwstr (approved):
Lgtm! Really need that content refactor...

PR #3890, opened 2024-09-06
AttachmentSheet: wait 300ms before popping image picker/camera

jamesacklin:
Ugly hotfix for an issue where the pickImage and takePicture functions are attempting to access the PHPickerViewController in the current view hierarchy, but the AttachmentSheet detaches itself (hides) before the controller can fire.

Will look into moving the image picker logic higher up into the component tree so we can ensure it's always attached. I can do this in a follow-up PR, or amend this one (or abandon if you have better ideas, @dnbrwstr).

Fixes TLON-2665


dnbrwstr (approved):
Lgtm! Would just add a comment for future reference.

PR #3892, opened 2024-09-06
lure: individualized links

arthyn:
This updates lure from the %groups perspective, changing %grouper and the web frontend to work with the new %reel flow. Fixes TLON-2517

PR Checklist


arthyn:
need to add changes to mobile as well

patosullivan (approved):
Frontend changes look good

desk/app/grouper.hoon
```diff
@@ -36,39 +36,50 @@
...
     =/  baseurl  .^(cord %gx /(scot %p our.bowl)/reel/(scot %da now.bowl)/service/noun)
-    =/  target=ship  (slav %p i.t.path)
-    =/  group=cord  i.t.t.path
+    ::  it really is necessary to double-encode this, make sure we strip
+    ::  the leading slash before encoding
+    =/  end  (en-urlt:html (en-urlt:html +:(spud rest.pole)))
+    =/  url
+      ?.  =(baseurl 'https://tlon.network/lure/')
+        (crip "{(trip baseurl)}{end}")
+      (crip "https://tlon.network/v1/policies/lure/{end}")
```
Fang-, line 74: Here and in urbit/landscape#280, it's surprising to me that the base url is expected to be given with a trailing slash, leading you into the annoying pattern of needing to strip the leading slash off the "actual" path... Unclear if that's worth correcting now, or punting to later.

(Also I maintain that double-encoding (or needing to encode at all for that matter) is a smell, but I get that it's finicky so we can just hold our noses for now.)

desk/app/grouper.hoon
```diff
@@ -104,30 +126,48 @@
...
-      %kick       
+      %kick
     :_  this
     ~[(bite-subscribe bowl)]
     ::
       %fact
     =+  !<(=bite:reel q.cage.sign)
-    ~?  dev-mode  [bite (~(has in enabled-groups) token.bite)]
-    ?>  (~(has in enabled-groups) token.bite)
-    ?>  ?=([%bite-1 *] bite)
```
Fang-, line 115: I don't see any changes here that stop producing `%bite-1` or start producing `%bite-2`. Am I missing something?

Also did `%bite-1` ever make it out into production-production? I thought that was your recent addition. If not, could be worth just updating the `%bite-1` case and calling it a day.

desk/app/grouper.hoon
```diff
@@ -104,30 +126,48 @@
...
+    |.
+    ?~  group=(~(get by fields.metadata.bite) 'group')
+      ~|("no group field" !!)
+    =/  =flag:groups  (flag:dejs:gj s+u.group)
+    ~?  dev-mode  [bite (~(has in enabled-groups) q.flag)]
+    ?>  (~(has in enabled-groups) q.flag)
+    ~?  dev-mode  'inviting'
+    =/  =invite:groups  [flag joiner.bite]
+    =/  =path  /(scot %p our.bowl)/groups/(scot %da now.bowl)/groups/(scot %p p.flag)/[q.flag]/noun
     =+  .^(=group:groups %gx path)
```
Fang-, line 161: Are you virtualizing here to save yourself if the scry crashes? Unfortunately, that won't save you. You'll want to do an explicit existence check.

Usually that'd be a `%gu` scry, but groups appears to implement a `%gx /exists/~some/group/noun`.

desk/app/grouper.hoon
```diff
@@ -104,30 +126,48 @@
...
+      =/  wir=^wire  /dm/(scot %p joiner.bite)
+      =/  =dock  [our.bowl %chat]
+      =/  =id:c  [our now]:bowl
+      =/  =memo:c
+        [~[[%inline ~[[%ship joiner.bite] ' has joined the network']]] id]
+      =/  =action:dm:c
+        :-  joiner.bite
+        [id %add memo [%notice ~] ~]
+      =/  =cage  chat-dm-action+!>(`action:dm:c`action)
+      (welp caz ~[[%pass wir %agent dock %poke cage]])
```
Fang-, line 147: For a single item, you can just `+snoc` it onto the end.

```suggestion
      (snoc caz [%pass wir %agent dock %poke cage])
```

Fang- (commented):
Some comments. Primarily, you can't save yourself from scry crashes by virtualizing.

428f4e6 is a weird inclusion for this particular pr, and on its face I don't immediately like it because it (lightly) obscures the extent to which chat and channel structures are identical. I don't feel super strongly though.

desk/app/grouper.hoon
arthyn: yeah it's bad just going to hold nose until we make a simpler lure service

desk/app/grouper.hoon
```diff
@@ -1,8 +1,8 @@
-/-  reel, groups
-/+  default-agent, verb, dbug
+/-  reel, groups, c=chat
+/+  gj=groups-json, default-agent, verb, dbug
 ::
 |%
-++  dev-mode  |
+++  dev-mode  &
```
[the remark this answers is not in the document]
arthyn:
```suggestion
++  dev-mode  |
```

desk/app/grouper.hoon
arthyn: those changes went through on the landscape side here: https://github.com/tloncorp/landscape/blob/develop/desk/app/bait.hoon#L136-L150

desk/app/grouper.hoon
arthyn: Was virtualizing here for the early return if we don't have a group, but I guess I could just return empty list in both crash/assertions

desk/app/grouper.hoon
```diff
@@ -138,7 +138,7 @@
       =/  wir=^wire  /dm/(scot %p joiner.bite)
       =/  =dock  [our.bowl %chat]
       =/  =id:c  [our now]:bowl
-      =/  =memo:c
+      =/  =memo:ch
```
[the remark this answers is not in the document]
arthyn: 🤯

Fang- (approved):
Thanks for the changes, hoon code looks good!

PR #3893, opened 2024-09-09
native: fix broken autoscrollToTopThreshold

davidisaaclee:
Fixes these bugs:

<details>

<summary>Inconsistent autoscroll-to-bottom behavior</summary>

We've specified that chats should scroll to the bottom whenever new content arrives.
In reality, the client decides whether to scroll to the bottom of the channel using some inscrutable logic.

</details>
<details>

<summary>Newly-sent message is placed below the bottom of scroll</summary>

1. Enter a chat channel
2. Don't scroll
3. Send a text message

Expected: Chat scrolls down to show your newly-sent message
Actual: Message is placed below bottom of scroll – offscreen until scrolled to

</details>

### Changes

- Applied https://github.com/facebook/react-native/pull/38245 to make autoscroll-to-bottom via `autoscrollToTopThreshold`.
- #3878 changed the [`autoscrollToTopThreshold`](https://reactnative.dev/docs/scrollview#maintainvisiblecontentposition) prop to disable autoscroll until the user has triggered an interactive scroll. I think the idea here was to guard against incorrect autoscrolls when loading posts in[0]. This caused the _Newly-sent message is placed below the bottom of scroll_ bug. Removed this condition.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/4bec2e70-f8fd-4955-ae52-3db3a0c2d22b"/> | <video src="https://github.com/user-attachments/assets/3442deb7-3353-481d-8c03-d3c5cd5d2efb" /> |

Fixes TLON-2675
Fixes TLON-2676

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,8 @@
       "@10play/tentap-editor@0.4.55": "patches/@10play__tentap-editor@0.4.55.patch",
       "any-ascii@0.3.2": "patches/any-ascii@0.3.2.patch",
       "react-native-gesture-handler@2.18.1": "patches/react-native-gesture-handler@2.18.1.patch",
-      "@likashefqet/react-native-image-zoom@3.0.0": "patches/@likashefqet__react-native-image-zoom@3.0.0.patch"
+      "@likashefqet/react-native-image-zoom@3.0.0": "patches/@likashefqet__react-native-image-zoom@3.0.0.patch",
+      "react-native@0.73.4": "patches/react-native@0.73.4.patch"
     },
     "allowNonAppliedPatches": true,
     "overrides": {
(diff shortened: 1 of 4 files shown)


davidisaaclee:
With https://github.com/facebook/react-native/pull/38245 applied, we can now use a real value for `autoscrollToTopThreshold`: ad0a5c1 adds code to scroll to bottom when within 400pt of the bottom of chat and you send a message or new messages load in.

I removed that commit since I don't think autoscrolling on new messages is 100% a good feature; maybe could be made a good feature by e.g. checking for message input focus

dnbrwstr (approved):
Great, lgtm!

PR #3897, opened 2024-09-10
native: make inverted empty FlatList resilient to upstream fixes 

davidisaaclee:
https://github.com/facebook/react-native/issues/21196 seems like it has regressed a few times. We previously manually flipped the broken empty list view to counteract the bugged transform; but this ends up with a bugged flipped view if the bug is fixed (which may likely happen without being marked as a breaking release).

To avoid Tlon breaking whenever the upstream bug gets fixed, disabled FlatList's `inverted` when we're showing an empty list – i.e. avoid dealing with any transforms.

I tested by running the new empty chat channel fixture on iOS and Android simulators. I also temporarily changed the `chatWithSimulatedLoad` to show an empty indicator when no posts are loaded, and checked that going from empty -> non-empty had no visible jank.

### DIFF
--- a/apps/tlon-mobile/src/fixtures/Channel.fixture.tsx
+++ b/apps/tlon-mobile/src/fixtures/Channel.fixture.tsx
@@ -362,6 +362,7 @@ function ChannelWithControlledPostLoading() {
             channel: baseProps.channel,
             post: anchorPost,
           }),
+          hasNewerPosts: true,
         })}
       />
       <FixtureToolbar>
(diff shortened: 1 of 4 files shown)


apps/tlon-mobile/src/fixtures/Channel.fixture.tsx
```diff
@@ -362,6 +362,7 @@ function ChannelWithControlledPostLoading() {
             channel: baseProps.channel,
             post: anchorPost,
           }),
+          hasNewerPosts: true,
```
davidisaaclee, line 365: without this line, this fixture would continuously load new posts, scroll to bottom, load more posts, scroll to bottom, etc... I'm guessing #3893 made the bug apparent

latter-bolden (approved):

PR #3899, opened 2024-09-11
Create %tm-alpha desk, update workflows/scripts/moons so we can deploy it

patosullivan:
Adds a new desk directory, `tm-alpha-desk`.
Adds a vite mode for the new `/apps/tm-alpha` base.
Updates github workflows/helper scripts.

I also installed %tm-alpha on both wannec and binnec (and then ran `|install our %tm-alpha` on each), so this should be good to go.

There's still some frontend bugs in the new app, going to break those out into their own issues.

fixes TLON-2535


latter-bolden (approved):
Lgtm!

PR #3900, opened 2024-09-11
build: remove iPad, visionOS, macOS build targets

jamesacklin:
Gets the builds running for production.


jamesacklin:
@dnbrwstr Something happened with the tsc types after merging #3894, can you take a look?

PR #3903, opened 2024-09-11
native: fix link pasting

latter-bolden:
This was a weird one to nail down the repro for since it was working inconsistently in testing. Turns out if your draft is empty when you enter the channel, we don't mark `setHasSetInitialContent` as true. This means once you do paste something in, it gets saved as a draft and loaded back into the input as a raw text overwrite. This PR just tweaks that logic to always mark initialized after entering the draft/edit content conditional.

I tried to make sure this doesn't break things elsewhere, but @patosullivan you might have more context there. Let me know if you foresee any issues.

Fixes TLON-2660


patosullivan (approved):
Makes sense to me, and I don't think this should break anything.

PR #3904, opened 2024-09-11
native: fix issue with pasted ref strings appearing in input

patosullivan:
Fixes TLON-2120

This one is difficult to reproduce, but my theory is that the editor within the webview is rerendering and we're sticking the ref string back in via drafts, this should fix that by removing ref strings from draft content before inserting it.

### DIFF
--- a/packages/ui/src/components/MessageInput/index.native.tsx
+++ b/packages/ui/src/components/MessageInput/index.native.tsx
@@ -226,8 +226,22 @@ export const MessageInput = forwardRef<MessageInputHandle, MessageInputProps>(
         try {
           getDraft().then((draft) => {
             if (draft) {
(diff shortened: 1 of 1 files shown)


latter-bolden:
@patosullivan was thinking about this as well, I think it might be fixed as a side effect of marking the draft/edit status as set.

If you pull latest develop, do you still see the ref text staying in there?

patosullivan:
> @patosullivan was thinking about this as well, I think it might be fixed as a side effect of marking the draft/edit status as set.
> 
> If you pull latest develop, do you still see the ref text staying in there?

yeah, had the same thought after looking at your PR, testing it now

latter-bolden (approved):
We talked out of band. Sounds like this will pair well with the other change.

PR #3909, opened 2024-09-12
ops: fix incorrect kelvin version

arthyn:
Fixes TLON-2652

PR Checklist


Fang- (approved):
Approved. For the record, we break compat with 412, because the import logic we added (in #3830) depends on the `$egg-any:gall` type in lull, which is an addition from 411.

PR #3911, opened 2024-09-12
new web: omnibus db-related fixes

patosullivan:
Fixes TLON-2691
Fixes TLON-2692
Fixes TLON-2716

We need to wait to render the AppRoutes until we know that we have data in the db, otherwise the app may appear blank before the user starts interacting with it. To fix this, I added a loop to poll whether the database size is zero. If it's not zero, we consider that "loaded" and we can render the AppRoutes.

This issue also effected loading in contact/calm settings data via the AppDataContext, and even if that data had loaded in we weren't getting it throughout useContact/useCalmSettings in the app. This was somewhat fixed by moving the context provider into AppRoutes, but it seems like a race condition could cause that data to still not be loaded in, so I also added a useEffect to refetch contacts/calm settings.

This also fixes the blank-screen-on-reload issue when reloading from a nested route (the route was actually rendering, but no data was loaded before it was rendered and there was nothing for the user to interact with that could cause a re-render).

I also fixed some issues with bad IDs in triggers (some tables use a combination of fields as primary keys), and enabled the change polling so we can get functionality similar to what we were getting with op-sqlite's updateHook (calling handleChanges every time an update is made).

### DIFF
--- a/apps/tlon-web-new/src/app.tsx
+++ b/apps/tlon-web-new/src/app.tsx
@@ -1,9 +1,11 @@
 // Copyright 2024, Tlon Corporation
 import { TooltipProvider } from '@radix-ui/react-tooltip';
+import { useCurrentUserId } from '@tloncorp/app/hooks/useCurrentUser';
 import { Provider as TamaguiProvider } from '@tloncorp/app/provider';
 import { AppDataProvider } from '@tloncorp/app/provider/AppDataProvider';
 import { sync } from '@tloncorp/shared';
 import * as api from '@tloncorp/shared/dist/api';
+import * as store from '@tloncorp/shared/dist/store';
 import cookies from 'browser-cookies';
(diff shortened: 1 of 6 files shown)


apps/tlon-web-new/src/app.tsx
```diff
@@ -97,99 +99,128 @@ function handleGridRedirect(navigate: NavigateFunction) {
   }
 }
 
-function AppRoutes() {
+function AppRoutes({ isLoaded }: { isLoaded: boolean }) {
+  const contactsQuery = store.useContacts();
+  const currentUserId = useCurrentUserId();
+  const calmSettingsQuery = store.useCalmSettings({ userId: currentUserId });
+
+  useEffect(() => {
```
latter-bolden, line 107: A comment for this section of logic would be helpful.

latter-bolden (approved):
Makes sense!

PR #3912, opened 2024-09-12
rework activity rendering

dnbrwstr:
I think there's more that can be to improve this, but cutting it for now.

| Before  | After |
| ------------- | ------------- |
|  |  |


apps/tlon-mobile/src/fixtures/Activity.fixture.tsx
```diff
@@ -0,0 +1,92 @@
...
+        currentUserId="~zod"
+        contacts={Object.values(exampleContacts)}
+      >
+        {children}
+      </AppDataContextProvider>
+    </FixtureWrapper>
+  );
+}
+
+const ActivityFixture = ({
```
latter-bolden, line 39: 🙏🏻

latter-bolden (approved):
Looking good

apps/tlon-mobile/src/fixtures/Activity.fixture.tsx
jamesacklin: You have to be fixturemaxing

PR #3914, opened 2024-09-12
bump react native to 0.73.9

dnbrwstr:
This PR upgrades react native from 0.73.4 to 0.73.9 in order to fix copying / logging of JSON in the chrome debugger.

The privacy manifest is autogenerated by the new version of React Native from our dependency pods. It looks appropriately conservative/accurate (write files in app group, use timers, that kind of thing), and I believe is required, so I'm leaving it, but we should research the ins + outs -- I'm not super familiar.


latter-bolden (approved):
Likewise not super familiar, but looks reasonable to me.

davidisaaclee:
I was curious about some of the keys in the manifest (e.g. are we really checking the user's available space?), which led me to https://github.com/Wooder/ios_17_required_reason_api_scanner – running this on my currently checked-out version of the project, I get https://gist.github.com/davidisaaclee/d1bf532361f010544ecd2a096bc07391

davidisaaclee (approved):

dnbrwstr:
Thanks for digging in @davidisaaclee, this is useful!

PR #3923, opened 2024-09-16
Edit group: navigate to ChatList (home) after group deletion

jamesacklin:
Navigates the user to ChatList once they delete a group. Fixes TLON-2680


patosullivan (changes requested):
We can't use react-navigation within the app package, the navigation action would need to be defined externally in the web and native controllers and then passed into GroupMetaScreen as a prop.

PR #3927, opened 2024-09-17
Posts, Channels: relocate overflow menu

jamesacklin:
Fixes an inconsistency where post detail views have Channel-level action menus (which have nothing to do with the post itself), and Channels have no menu whatsoever (so muting a channel involves navigating into a child post and accessing the menu in the incorrect place).

Note that this adds a '...' button to the channel header. This is probably correct and not reflected in design.

Screenshot 2024-09-17 at 9 59 42 AM

Screenshot 2024-09-17 at 10 00 03 AM


dnbrwstr (approved):
Nice, lgtm.

PR #3930, opened 2024-09-17
native: onboarding flow coherence with new lure setup

latter-bolden:
- use correct token in branch link & post signup bite to hosting
- wait for valid lure token before showing _Share with friends not on Tlon_ and refetch until we have it
- fix post-signup boot failure (don't let bad ship status kill the polling)


packages/app/lib/hostingApi.ts
```diff
@@ -141,7 +141,7 @@ export const signUpHostingUser = async (params: {
       email: params.email,
       password: params.password,
       lure: params.lure || DEFAULT_LURE,
-      priorityToken: params.priorityToken || DEFAULT_PRIORITY_TOKEN,
```
arthyn, line 144: might not want to commit this

packages/shared/src/logic/branch.ts
```diff
@@ -97,9 +97,9 @@ export const createDeepLink = async ({
   if (!fallbackUrl || !path) {
     return undefined;
   }
-  if (type === 'lure' && !urbit.whomIsFlag(path)) {
-    return undefined;
-  }
+  // if (type === 'lure' && !urbit.whomIsFlag(path)) {
+  //   return undefined;
+  // }
```
arthyn, line 102:
```suggestion
```

arthyn (approved):
lgtm

PR #3934, opened 2024-09-18
native: fix onboarding nickname

latter-bolden:
The code that tries to thread through nickname via screen props wasn't working. It was also calling out to a non-standard legacy helpers for setting profile data.

I extracted signup configuration into its own context and pulled post-signup logic up into _AutheticatedApp_ so we can work with our normal api client & store methods. I think this will be more agile moving forward compared to passing along screen props and bundling the logic into the final onboarding screen.

Confirmed setting the nickname worked by running through a full signup.

Fixes TLON-2744


arthyn (approved):
lgtm

PR #3936, opened 2024-09-19
expose + profile: moves all shared styles, reformats widgets

jamesacklin:
Moves styles around a bit to unify public-facing pages in style, type, colors, etc.

Barebones preview:
image

Kitted-out preview:
image


desk/app/profile.hoon
```diff
@@ -134,6 +137,12 @@
     ::
     :-  [%pass /eyre/cache %arvo %e %set-response '/profile' `[| %payload payload]]
     (spout:rudder id payload)
+  ?:  =('/expose/style/shared.css' url.request)
+      :-  [200 ['content-type' 'text/css']~]
+      `(as-octs:mimes:html style-shared)
```
Fang-, line 142: The %profile agent only serves requests for `/profile` and stuff underneath that path, so this is never going to get hit. %expose will handle it though.

desk/app/profile.hoon
```diff
@@ -348,22 +237,19 @@
...
-          ;path
-            =d  "M15.4151 0.259814L0.497261 1.82774C0.222631 1.85661 0.0233995 2.10264 0.0522642 2.37727L0.391982 5.60946C0.420847 5.88409 0.666877 6.08332 0.941507 6.05446L5.41686 5.58408C5.96612 5.52635 6.45818 5.92482 6.51591 6.47407L6.79029 9.08469C6.84081 9.5653 6.49215 9.99585 6.01155 10.0464C5.53095 10.0969 5.10039 9.74822 5.04988 9.26762L4.85389 7.40289C4.82502 7.12826 4.57899 6.92903 4.30436 6.95789L1.07217 7.29761C0.797538 7.32648 0.598306 7.57251 0.627171 7.84714L1.56793 16.7978C1.62566 17.3471 2.11772 17.7456 2.66698 17.6878L16.5903 16.2244C17.1395 16.1667 17.538 15.6746 17.4803 15.1254L16.5395 6.17468C16.5107 5.90005 16.2646 5.70082 15.99 5.72968L12.7578 6.0694C12.4832 6.09827 12.2839 6.3443 12.3128 6.61893L12.5088 8.48366C12.5593 8.96426 12.2107 9.39481 11.73 9.44533C11.2494 9.49584 10.8189 9.14718 10.7684 8.66658L10.494 6.05596C10.4363 5.5067 10.8347 5.01464 11.384 4.95691L15.8593 4.48653C16.134 4.45767 16.3332 4.21164 16.3043 3.93701L15.9646 0.70481C15.9357 0.430181 15.6897 0.230949 15.4151 0.259814Z";
+      ;div.tlon-badge
+        ;a(href "https://tlon.io")
+          ;img@"https://tlon.io/icon.svg"(alt "Tlon logo", width "18");
```
Fang-, line 248: Any reason to hotlink this instead of embedding the svg directly? Loading external resources is somewhat lame imo, we should be able to easily avoid it here right?

desk/app/profile.hoon
```diff
@@ -134,6 +137,12 @@
     ::
     :-  [%pass /eyre/cache %arvo %e %set-response '/profile' `[| %payload payload]]
     (spout:rudder id payload)
+  ?:  =('/expose/style/shared.css' url.request)
+      :-  [200 ['content-type' 'text/css']~]
+      `(as-octs:mimes:html style-shared)
+  ?:  =('/profile/page.css' url.request)
+      :-  [200 ['content-type' 'text/css']~]
+      `(as-octs:mimes:html style-page)
```
Fang-, line 145: You might consider serving this at `/profile/style/page.css` instead, for consistency with the `/expose/style/*.css` paths. Or maybe we should take out the `/style` in both?

(It doesn't strictly need to match the file/folder structure in the desk, but I'd update those to match if you make such changes.)

Fang- (commented):
Some comments!

desk/app/profile.hoon
Fang-: Oh and we'll want to make sure this gets served from cache at some point, but, fine for now.

jamesacklin: Moved!

desk/app/profile.hoon
jamesacklin: Embedded!

desk/app/profile.hoon
jamesacklin: Removed!

Fang- (approved):
Changes look good, thank you!

PR #3937, opened 2024-09-19
native: fix invite sheet on Android

latter-bolden:
The invite sheet was using the old format that caused missing React context on Android. This was causing it to crash to the error boundary when opened. Moves the sheet to use the new style which fixes that issue. Looks like @patosullivan may have done something similar on his branch after merging this with latest develop.

Note, we still have an issue with subscriptions not working on Android sim or dev builds. To test this is now working, I had to do a local production build via eas.

Also removes the sometimes-failing metadata scry from Lure fetching since it's not used anywhere.

Fixes TLON-2749


arthyn (approved):
lgtm

PR #3938, opened 2024-09-19
native: remove skip button in invite sheet

latter-bolden:
Now that the updated group creation flow is in and we display the invite in the channel itself, we don't need the skip button in the sheet. This will help avoid confusion since it was always shown and doesn't make sense outside the group creation context.


arthyn (approved):
lgtm

PR #3939, opened 2024-09-19
native: more lure fixes

latter-bolden:
- fix stale Lure detection
- only consider Lure status _ready_ if we have a deeplink
- more error logging
- shorten timeout for subscribe once fetches

Should resolve TLON-2753, but I was only seeing that present on the actual build. Will cut a new one and confirm resolved before closing.


arthyn (approved):
lgtm

PR #3940, opened 2024-09-19
group creation: make sure back button goes back to chat list after successful group creation

patosullivan:
Fixes TLON-2754


latter-bolden (approved):
Looks good!

Not sure if we'd want to push the  _GroupChannels_ screen as well (like we do [here](https://github.com/tloncorp/tlon-apps/blob/55c93d953e37fc2b5b9854cc67599c9065e216a4/apps/tlon-mobile/src/hooks/useNotificationListener.ts#L177)). Might make sense to skip since we know they'll just have the one channel.

PR #3941, opened 2024-09-19
native: add comment about `hasCachedNewest`

davidisaaclee:
I was confused by this, so I thought I'd add a comment for future readers. There's a small code change, but I think it's equivalent.

### DIFF
--- a/packages/app/features/top/ChannelScreen.tsx
+++ b/packages/app/features/top/ChannelScreen.tsx
@@ -113,9 +113,48 @@ export default function ChannelScreen({
       return false;
     }
     const { syncedAt, lastPostAt } = channel;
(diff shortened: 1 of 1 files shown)


packages/app/features/top/ChannelScreen.tsx
```diff
@@ -113,9 +113,48 @@ export default function ChannelScreen({
       return false;
     }
     const { syncedAt, lastPostAt } = channel;
-    if (syncedAt && session.startTime < syncedAt) {
+
+    if (syncedAt == null) {
+      return false;
```
davidisaaclee, line 118:
```suggestion
+      // I think we could return `true` here if we've received at least one post through the subscription
      return false;
```

dnbrwstr (approved):
V nice.

PR #3942, opened 2024-09-19
native: handle Lure link click when already signed in

latter-bolden:
Adds logic to handle the case where you click a lure link after already being authenticated. Instead of triggering the automatic join + dm, we treat it as a deeplink to that group and pop a preview sheet.

Fixes TLON-2758

### DIFF
--- a/packages/app/contexts/branch.tsx
+++ b/packages/app/contexts/branch.tsx
@@ -17,7 +17,7 @@ import { useShip } from './ship';
 
 interface LureData extends DeepLinkMetadata {
   id: string;
-  clickedPreAuth: boolean;
+  shouldAutoJoin: boolean;
 }
 
 type Lure = {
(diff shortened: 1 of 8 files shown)


packages/shared/src/store/groupActions.ts
```diff
@@ -50,6 +50,17 @@ export async function createGroup({
   }
 }
 
+export async function getGroupPreview(groupId: string) {
```
dnbrwstr, line 53: Seems largely the same as `sync.syncGroupPreviews`, can we combine the two? I'm agnostic as to which we keep.

packages/app/contexts/branch.tsx
```diff
@@ -13,9 +13,11 @@ import branch from 'react-native-branch';
 import { DEFAULT_LURE, DEFAULT_PRIORITY_TOKEN } from '../constants';
 import storage from '../lib/storage';
 import { getPathFromWer } from '../utils/string';
+import { useShip } from './ship';
 
 interface LureData extends DeepLinkMetadata {
   id: string;
+  clickedPreAuth: boolean;
```
dnbrwstr, line 20: Either a comment or a different variable name would be helpful here -- not obvious to me what "clickedPreAuth" means.

dnbrwstr (approved):
Overall lgtm! Couple nits.

PR #3943, opened 2024-09-20
native: use null pagination token to indicate end of posts list

davidisaaclee:
fixes TLON-2729 (or at least one cause of it!)

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/1f330280-1e99-4202-b759-64049d67e0b7" /> | <video src="https://github.com/user-attachments/assets/d85e97ba-851c-41a9-ac27-199e2d9b7acd" /> |

<details>

<summary>Those videos use the following patch to make the bug reproducible by simulating a slow call to sync endpoint:</summary>

```diff
diff --git a/packages/shared/src/store/sync.ts b/packages/shared/src/store/sync.ts
index 9f403e053..d412a600d 100644
--- a/packages/shared/src/store/sync.ts
+++ b/packages/shared/src/store/sync.ts
@@ -901,6 +901,10 @@ export async function syncPosts(
       newer: response.newer,
       older: response.older,
     });
+  } else {
+    console.log('Simulating slow sync...');
+    await new Promise((resolve) => setTimeout(resolve, 5000));
+    console.log('Resuming sync.');
   }
   if (!response.newer) {
     await db.updateChannel({
```

</details>

_This was a tricky one, excuse my wordcount!_

### Background context
- When sending a message, we only show the remote echo of the message – we don't show the "pending" message.
- We use `hasNewerPosts` in this logic, which is equivalent to the result of `useInfiniteQuery`'s `getPreviousPageParam(...) === undefined`. (This is distinct from `hasCachedNewest`: `hasCachedNewest` is about caching new messages locally, and `hasNewerPosts` is about paginating posts into memory.)
- https://linear.app/tlon/issue/TLON-2249/native-full-chat-log-doesnt-load informed some of the current implementation
- We prevent autoscroll-to-bottom in chat when there are newer posts to load (`hasNewerPosts`)

### What was the issue?
The problem case is sending a message after we've received all most-recent posts, but before we've fetched an empty page of posts to confirm that we are up-to-date on sync.

- In the problem case, we flipped `hasNewerPosts` after rendering the newly-sent message:
  - Render 1: We get sync response with the new message's remote echo, and we haven't fetched an empty page confirming done – we render the post set including the new message, and `hasNewerPosts = true`
  - Render 2: Sync with empty page returns, confirming done; we render the same set of posts as last time, but with `hasNewerPosts = false`
- Since we render the new post while `hasNewerPosts = true` and `hasNewerPosts` prevents autoscroll, we don't scroll to it

### What's the fix?
Switch `hasNewerPosts` to false at the same time that we get the remote echo of our sent message:

- afaict, `get posts` API call returns `newer: null` when there are no newer posts to get.
- We get this signal at the same time that we receive the latest message (i.e. the remote echo of our newly-sent message).
- This PR uses that signal in `getPreviousPageParam` to stop pagination, causing `hasNewerPosts` to flip to false on the same render as passing the last set of posts to `Scroller`.

Downsides:
- We block on API response to show the sent message in chat – this is our existing behavior
- We still hit the API once at the end of every backfill, in worst case getting an empty page – again, existing behavior afaict. We could do better by checking our cached messages against the received head of channel.

### What about the code in `getPreviousPageParam` fixing https://linear.app/tlon/issue/TLON-2249/native-full-chat-log-doesnt-load ?
This was the condition that, when true, led to `hasNewerPosts = false` previously: https://github.com/tloncorp/tlon-apps/blob/cbe9d1d113c6770e6a08bdff4a64d0f19cc7017f/packages/shared/src/store/useChannelPosts.ts#L116-L121

The bolded part of **`firstPageParam.mode === 'newest'`** `&& options.hasCachedNewest` was added in https://github.com/tloncorp/tlon-apps/pull/3708 to f-i-x TLON-2249.

What was going on here? Does this PR's change break that fix?
- On first load, simply checking `firstPageIsEmpty || options.hasCachedNewest` is sufficient for `around` / `newer` backfills
- But once you have the chat cached, close app, and return to chat via `around` (e.g. via search into chat history), paginating back towards present stopped working
- This is because `hasCachedNewest` was returning true before loading all posts in the chat, since we had the latest post cached **from network into DB, but not from DB into runtime use-query memory** https://github.com/tloncorp/tlon-apps/blob/adde000f4330af7e0a2e19bdfcb295f5eb9fe3da/packages/app/features/top/ChannelScreen.tsx#L118-L119

Theoretically, this change preserves that distinction between DB-cached vs query-cached; and attempting repro by entering a chat with unreads, scrolling towards present, and triggering pagination does work with this change; as does navigating to a search result and scrolling towards present. **Is this the correct repro?**

---

### Remaining work
- I've tested the problem cases thoroughly, but I still need to check all other cases to ensure no regression
- The `getNextPageParam` callback has a similar issue where it requires one more fetch of an empty page before confirming that pagination is done – I think we should use a check on `null` cursor here too, but I don't think there's much user-facing benefit.

### DIFF
--- a/packages/shared/src/store/useChannelPosts.ts
+++ b/packages/shared/src/store/useChannelPosts.ts
@@ -19,8 +19,17 @@ import { SyncPriority } from './syncQueue';
 
 const postsLogger = createDevLogger('useChannelPosts', false);
 
+type PostQueryPage = {
+  posts: db.Post[];
+  /**
+   * False when a sync page reports that there are no more newer posts.
+   * Obviously, new posts can be made after this is set: in practice, we
+   * should have switched over to a subscription by then.
+   */
+  canFetchNewerPosts: boolean;
+};
 type UseChannelPostsPageParams = db.GetChannelPostsOptions;
-type PostQueryData = InfiniteData<db.Post[], unknown>;
+type PostQueryData = InfiniteData<PostQueryPage, unknown>;
 type SubscriptionPost = [db.Post, string | undefined];
 
 type UseChanelPostsParams = UseChannelPostsPageParams & {
@@ -52,7 +61,7 @@ export const useChannelPosts = (options: UseChanelPostsParams) => {
       count: firstPageCount,
     } as UseChannelPostsPageParams,
     refetchOnMount: false,
-    queryFn: async (ctx): Promise<db.Post[]> => {
+    queryFn: async (ctx): Promise<PostQueryPage> => {
       const queryOptions = ctx.pageParam || options;
       postsLogger.log('loading posts', queryOptions);
       // We should figure out why this is necessary.
@@ -66,7 +75,7 @@ export const useChannelPosts = (options: UseChanelPostsParams) => {
       const cached = await db.getChannelPosts(queryOptions);
       if (cached?.length) {
         postsLogger.log('returning', cached.length, 'posts from db');
-        return cached;
+        return { posts: cached, canFetchNewerPosts: true };
       }
 
       postsLogger.log('no posts found in database, loading from api...');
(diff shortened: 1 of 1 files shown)


dnbrwstr (approved):
Lgtm! @latter-bolden -- would be great to get this in the build today.

PR #3944, opened 2024-09-20
native: handle pasting Lure deeplinks in chat

latter-bolden:
We've seen users get confused about how to properly link to a group. Dropping an off-network share link into chat is a reasonable instinct that we should support. This PR adds the ability to extract group refs from Lure deeplinks in the same way we do if you paste a copied cite ref. Handles links at the branch level (_join.tlon.io_) and after fallback redirect (_tlon.network/lure_).

I tried to add support in a way that would make it easier to add more of these paste handlers down the road, but I could use eyes on the extracted logic — @patosullivan I think you're most familiar?

Fixes TLON-2760

### DIFF
--- a/packages/shared/src/store/lure.ts
+++ b/packages/shared/src/store/lure.ts
@@ -78,14 +78,14 @@ export const useLureState = create<LureState>((set, get) => ({
       },
     });
 
-    return get().fetchLure(flag, branchDomain, branchKey);
+    // return get().fetchLure(flag, branchDomain, branchKey);
   },
   toggle: async (flag, meta, branchDomain, branchKey) => {
     const { name } = getFlagParts(flag);
     const lure = get().lures[flag];
     const enabled = !lure?.enabled;
     if (!enabled) {
-      lureLogger.log('not enabled, poking reel-undescribe', flag);
+      lureLogger.crumb('not enabled, poking reel-undescribe', flag);
       await poke({
         app: 'reel',
         mark: 'reel-undescribe',
@@ -112,7 +112,7 @@ export const useLureState = create<LureState>((set, get) => ({
       json: name,
     });
 
-    return get().fetchLure(flag, branchDomain, branchKey);
+    // return get().fetchLure(flag, branchDomain, branchKey);
   },
   start: async () => {
     const bait = await scry<Bait>({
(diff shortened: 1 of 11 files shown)


jamesacklin:
Highly reasonable expectation. Does this paste-action work in other inputs (Gallery rich text, Notebook BigInput)?

latter-bolden:
> Does this paste-action work in other inputs (Gallery rich text, Notebook BigInput)

Yeah, uses the same component under the hood. 

Definitely optimized for the chat case though. As with the preexisting cite pastes there, the ux is lackluster since we don't preview inserted attachment until after you save

packages/ui/src/components/MessageInput/helpers.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const logger = createDevLogger('processReference', true);
+
+export async function processReferenceAndUpdateEditor({
+  editor,
+  pastedText,
+  matchRegex,
+  processMatch,
+}: {
+  editor: EditorBridge;
```
patosullivan, line 20: We don't use tentap on web, so we should probably do something other than passing the editor bridge here

patosullivan (changes requested):
Looks good overall, but this will only work for mobile, we should probably make sure that the helpers module can work for both mobile and web.

packages/ui/src/components/MessageInput/helpers.ts
latter-bolden: Added the ability to pass either web or native version, how's that feel? Could also see an argument for the function accepting/returning full editor content instead

packages/shared/src/store/lure.ts
```diff
@@ -78,14 +78,14 @@ export const useLureState = create<LureState>((set, get) => ({
       },
     });
 
-    return get().fetchLure(flag, branchDomain, branchKey);
+    // return get().fetchLure(flag, branchDomain, branchKey);
```
patosullivan, line 81:
```suggestion
```

packages/shared/src/store/lure.ts
```diff
@@ -112,7 +112,7 @@ export const useLureState = create<LureState>((set, get) => ({
       json: name,
     });
 
-    return get().fetchLure(flag, branchDomain, branchKey);
+    // return get().fetchLure(flag, branchDomain, branchKey);
```
patosullivan, line 115:
```suggestion
```

patosullivan (approved):
lgtm, just a couple commented out lines that can go

PR #3947, opened 2024-09-23
lib-negotiate: set src.bowl in +on-load when simulating kicks

mikolajpp:
`lib-negotiate` does not properly set `src.bowl` when issuing kicks in `+on-load`. This PR fixes this, by properly setting the kicks source according to the `gill` from `wex.bowl`


Fang- (approved):
Oops, sorry for the wait. Just ping me if I leave review requests hanging!

This is a good catch, thank you.

mikolajpp:
This already went in, and was reviewed here https://github.com/tloncorp/tlon-apps/commit/ae805939b5d8921eb3ab47411540853c6af0fbbf#comments

PR #3949, opened 2024-09-24
emojimart: transform "shortcodes" back into emoji

Fang-:
Instead of the raw unicode, we've stored reacts as ascii-textual representations, which are largely impractical, require a library for rendering, and end up with rather ridiculous representations, such as `:hand_with_index_finger_and_thumb_crossed::skin-tone-2:` in place of `🫰🏻`.

Here, we port part of the aforementioned rendering library into hoon. Specifically, the mappings of those "shortcode" identifiers to their unicode characters, and a helper function for looking those up from a potentially with-modifier shortcode.

Note the auraless atoms produced here, because there exists no aura for "unicode character". (`@c` is a UTF-32 codepoint, not a semantic character.)

At some point, we may want a more full-fledged unicode wrangling library in hoon. But for now, this suffices.

(For future reference, the `+data:emojimart` was codegen'ed from the missive/emojimart repo's `packages/emoji-mart-data/sets/15/all.json` file with `jq --raw-output '[.emojis | .[] | { id: .id, ch: ([.skins[].native] | join("\' \'")) }] | .[] | "[\'\(.id)\' \'\(.ch)\']"'`... which could probably be golfed down.)

Includes 80% of migration code, which was used for performance testing. Building the library file takes half a minute, running the migration takes a mere second.

Not included: rest of migration code for channels, all of the migration code for chat, related frontend changes.

Going ahead and pushing this as a draft PR, because we'll want to roll this into message type changes whenever they happen. Proposed new react type is as follows, but up for discussion:

```hoon
+$  react
  $@  @          ::  direct unicode character representation
  $%  [%any @t]  ::  any string representation (for backcompat)
  ==
```


mikolajpp:
Closing, as emoji migration has become part of the channels changes implemented in https://github.com/tloncorp/tlon-apps/pull/4105

PR #3952, opened 2024-09-25
Switch to react-navigation on web, consolidate screens, remove controllers

patosullivan:
Also moved the stacks into packages/app and moved lib/webDb and its dependencies in there too.


jamesacklin:
A couple things:

The "Blocked users" screen is missing a back-caret in the header.
image

I'm unable to test group settings; clicking the "..." icon here (or on the ChatList screen) doesn't open a modal:
image

The "+ Add" button (now deprecated?) on the ChatList screen does nothing:

https://github.com/user-attachments/assets/44628256-c7fb-46fc-8b45-fe06f3a3aa31

packages/app/hooks/useChannelNavigation.old.ts
```diff
@@ -0,0 +1,76 @@
+import { isGroupChannelId } from '@tloncorp/shared/dist';
```
dnbrwstr, line 1: If we're not actually referencing this file anyway, I think it's fine to discard

packages/app/hooks/useGroupNavigation.old.ts
```diff
@@ -0,0 +1,23 @@
+import * as db from '@tloncorp/shared/dist/db';
```
dnbrwstr, line 1: Remove

packages/app/hooks/useIsFocused.old.ts
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState } from 'react';
```
dnbrwstr, line 1: Remove

packages/app/hooks/useIsFocused.ts
```diff
@@ -1,34 +1 @@
...
-    document.addEventListener('visibilitychange', handleVisibilityChange);
-
-    return () => {
-      document.removeEventListener('visibilitychange', handleVisibilityChange);
-    };
-  }, []);
-
-  return isFocused;
-}
+export { useIsFocused } from '@react-navigation/native';
```
dnbrwstr, line 1: Do we need this file?

dnbrwstr (changes requested):
Overall lgtm! Mostly just comments about pruning more files.

One optional: passing the navigation methods into the screens as raw functions will mean that the screen components are "changed" on every render, which is potentially expensive -- might be worth trying to memoize them.

dnbrwstr (approved):
Nice, lgtm!

PR #3954, opened 2024-09-26
expose: make author link to local /profile

Fang-:
when we can. Right now, we only know this for the local ship, where it's simply /profile, assuming the public profile agent is running.

Still pending some styling adjustments from @jamesacklin.

### DIFF
--- a/desk/app/expose/page.hoon
+++ b/desk/app/expose/page.hoon
@@ -123,7 +123,11 @@
     ^-  marl
     =/  main=manx
       ;div.author-row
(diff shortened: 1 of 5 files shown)


jamesacklin:
Changes are in.

- Realigns author row block, inline elements; restyles link
- Locks profile header and diary cover images to 1.6 aspect ratio
- Fixes an issue where the border-radius on the note/post page was rounder than on the profile page

jamesacklin (approved):
LGTM

Fang- (commented):
Style changes look good!

PR #3958, opened 2024-09-26
expose: refresh for local profile changes

Fang-:
This is a detail that appears on most of the pages that we publish, that the user has control over, so we should reflect is accurately in a timely fashion.

We don't care too much right now about cache busting in response to other people's profile detail changes, because we don't display comments/replies, and publishing someone else's content isn't a primary/supported use-case. If we ever do want that, probably better to just periodically refresh the cache.

Also makes some small adjustments to how the contacts scry helper works, and makes /app/profile use that instead of its own code.

(Doesn't account for doing the contacts subscription on existing instances of exposé, @jamesacklin if you or anyone internal wants the reactivity there you'll need to nuke & restart the agent.)

### DIFF
--- a/desk/sur/contacts.hoon
+++ b/desk/sur/contacts.hoon
@@ -89,8 +89,10 @@
 ::
 ++  get-contact
   |=  [=bowl:gall who=@p]
-  ^-  (unit contact)  ~+
-  =/  base=path  /(scot %p our.bowl)/contacts/(scot %da now.bowl)
+  =>  :_  ..get-contact
+      [who=who our=our.bowl now=now.bowl]
+  ~+  ^-  (unit contact)
+  =/  base=path  /(scot %p our)/contacts/(scot %da now)
   ?.  ~+  .^(? %gu (weld base /$))
     ~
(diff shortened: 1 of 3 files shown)


jamesacklin (approved):
Approved! Works as advertised.

desk/sur/contacts.hoon
```diff
@@ -89,8 +89,10 @@
 ::
 ++  get-contact
   |=  [=bowl:gall who=@p]
-  ^-  (unit contact)  ~+
-  =/  base=path  /(scot %p our.bowl)/contacts/(scot %da now.bowl)
+  =>  :_  ..get-contact
```
arthyn, line 92: what does this do?

Fang-: We want to memoize (with `~+`) the computations below. For that, we need a "fixed" subject, as it becomes the memoization key. We need some stdlib functions as well as some types from the file/core we're currently in, so instead of including those by hand with `[contact=contact ..zuse]` or similar, we just say `..get-contact` for "the context of this arm".

arthyn: interesting TIL

arthyn (approved):
lgtm!

PR #3960, opened 2024-09-27
ChatOptionsSheet: fixes for notifications, leave

jamesacklin:
- Guards against channel hosts being able to leave channels they host. This would fail to complete on the back-end but would be reflected in the client and in SQLite. (Fixes TLON-2785)
- Moves "Back" in notification level options to sheet header, adjusts title (fixes TLON-2737)
- Adds active selection icons for notification level selections (fixes TLON-2737)
- Changes ActionSheet to render (optional) accent styling for individual actions within a group (which can have its own accent)
- Provides a "positive" accent to the active selection for notification level in ChatOptionsSheet

I can now no longer leave a channel I host:
image

Changes to the notification selection menu(s):
image


patosullivan (approved):
Looks good! I do wonder if the caret in the sheet header is a little too far to the left, but that's not my call.

PR #3961, opened 2024-09-30
Only show lure invite button if we're an admin in a new empty group

patosullivan:
fixes TLON-2804


jamesacklin (approved):
Certainly more conditional than before!

PR #3962, opened 2024-09-30
ChatList: remove native scroll events and title-swapping

jamesacklin:
Removes the flapping between "Pinned" and "Home" on ChatList. Should help with performance a bit.


patosullivan (approved):
Looks fine to me, hope it makes a difference on Android (seems like it might).

jamesacklin:
@latter-bolden told me to merge.

PR #3963, opened 2024-09-30
input: take keyboard height into account when setting maxInputHeight

patosullivan:
Fixes TLON-2730.

We need to update the maxInputHeight when the keyboard is open, subtracting the keyboard height.


jamesacklin:
Now it's a party. Checking it out now.

latter-bolden (approved):
Worked in testing on both iOS and Android.

PR #3966, opened 2024-09-30
fix type issue so CI can start passing again

patosullivan:
we need to check that `session.startTime` actually exists before we check to see if `syncedAt` is greater than or equal to it.

this is currently causing type checks in CI to fail.

### DIFF
--- a/packages/app/features/top/ChannelScreen.tsx
+++ b/packages/app/features/top/ChannelScreen.tsx
@@ -1,3 +1,5 @@
+import { useFocusEffect } from '@react-navigation/native';
+import { useIsFocused } from '@react-navigation/native';
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
 import { createDevLogger } from '@tloncorp/shared/dist';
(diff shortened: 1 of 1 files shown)


davidisaaclee:
lgtm

@dnbrwstr https://github.com/tloncorp/tlon-apps/commit/a51eb5259c2f84f74e4847df4169ef776ba4f501 why was `startTime` changed to be optional? it's not clear what it'd be mean to have a non-null Session without a startTime

dnbrwstr:
> lgtm
> 
> @dnbrwstr [a51eb52](https://github.com/tloncorp/tlon-apps/commit/a51eb5259c2f84f74e4847df4169ef776ba4f501) why was `startTime` changed to be optional? it's not clear what it'd be mean to have a non-null Session without a startTime (maybe it's a "currently reconnecting" state? still not clear why we'd keep a `Session` instead of setting it to undefined)

Yeah, I think that's just an oversight and that you're right that it should be non-optional. I'll change back.

Also, I made this same change while i was fixing the expo build, so I'm going to close. Sorry Pat!

PR #3968, opened 2024-09-30
native: factor draft inputs into isolated components

davidisaaclee:
### Motivation
Custom channels requires swapping input views out dynamically. To make this easy, I wanted to factor the existing `Channel` component so that showing a specific input is close to `if inputType === 'chat' { return <ChatInput> }`.

This PR is _only inputs_ – dynamically changing post content renderers is already nicely encapsulated, and post collections (switching between grid, list, etc) is a future PR.

### Summary
- Moved notebook, gallery, chat inputs into similarly-typed React components
- Flattened conditionals as much as possible in `Channel` component
- We reuse the `Channel` nav bar's back button to exit out of a fullscreen "big input." I think this is wonky and we should change the implementation, but to reduce this PR's scope, I hacked this in using a ref handle on the draft inputs 441b63625578e3bab77d91f9c262b0d15a505fed
- Changed how we use safe area insets and keyboard-avoiding views w.r.t. inputs:
  - We want the following features:
    - inline input is stacked with chat scroller (i.e. input has ability to resize scroller)
    - `KeyboardAvoidingView` should shrink the stack of both chat scroller + input
    - safe area inset should not add whitespace below scroll (e.g. when input is FAB; scroll should extend to bottom of screen)
    - safe area inset should collapse when keyboard is up (naturally handled when KAV is nested inside view with safe area inset)
  - Previously, we had one bottom-safe-area-inset view with an implementation-specific condition https://github.com/tloncorp/tlon-apps/blob/e239da85db2bb01771670c9276273126f486831f/packages/ui/src/components/Channel/index.tsx#L265-L267 near the root of `Channel`, and a [`KeyboardAvoidingView`](https://github.com/tloncorp/tlon-apps/blob/e239da85db2bb01771670c9276273126f486831f/packages/ui/src/components/Channel/index.tsx#L289) inside that inset view. All `Channel` contents (posts + inputs) shared those containers.
  - The input-specific condition for the inset `paddingBottom` is less maintainable with plug-n-play draft inputs. I figured that, while most inputs will want a KAV and safe area insets, it's nicer to just let each input decide what it wants: i.e. move `KeyboardAvoidingView` + `paddingBottom: safeAreaInsets.bottom` into the specific `ChatInput` / `NotebookInput`, etc
  - This exposed a bug in `KeyboardAvoidingView`: when KAV is inside a parent that is vertically offset from the top of the screen (e.g. `paddingTop > 0`), KAV doesn't offset its content as expected. To fix, I added [`ParentAgnosticKeyboardAvoidingView`](https://github.com/tloncorp/tlon-apps/blob/dil/factor-draft-input-methods/packages/ui/src/components/ParentAgnosticKeyboardAvoidingView.tsx), which should work as an almost-drop-in for `KeyboardAvoidingView` (just without some customization props to reduce scope).
  - With `ParentAgnosticKeyboardAvoidingView`, draft inputs can decide which parts they want to have avoid keyboard without worrying about where the input view will be presented.
  - **I care not for the boy,** I tried to make this work with the stock KAV but failed, and `ParentAgnosticKeyboardAvoidingView` was pretty quick to throw together. We should try to revert to a stock solution – writing a custom component just seemed quicker in the moment. (example of why we should try to go to stock solution: apparently "prefer cross fade transitions" on iOS will be buggy with my naive implementation: https://github.com/facebook/react-native/blob/513e9669e78a4bfd9b0380335c61f581343c4009/packages/react-native/Libraries/Components/Keyboard/KeyboardAvoidingView.js#L87-L95 )

### Behavior changes
- Previously, we were only showing the "negotiation mismatch" banner in chat channels. I changed this to show in any channel type if we have a mismatch.
- Technically, KeyboardAvoidingView behavior has changed on Android – I haven't checked the current client's behavior yet, but I think this change has no apparent effect, since we are using [`adjustResize` on `windowSoftInputMode`](https://github.com/tloncorp/tlon-apps/blob/9f69866bf51eb4f31402e7008d562732739f755b/apps/tlon-mobile/android/app/src/main/AndroidManifest.xml#L33), which means the window will always resize to stack with the keyboard, and we never really need to have views avoid the keyboard on Android.

### Future work
- I avoided the internals of `MessageInput` and `BigInput` – I think both of these have internal checks about the channel type (at least `BigInput` does), and we'll need to factor this yet; but I tried to make this PR as small as possible.

| Scenario | Video |
| --- | --- |
| Chat | <video src="https://github.com/user-attachments/assets/f29974be-2975-4cfb-85df-b875515bcdf3" /> |
| Notebook | <video src="https://github.com/user-attachments/assets/58fa97a1-b039-4311-a072-6f18837bb047" /> |
| Gallery photo (before fixing animation issue) | <video src="https://github.com/user-attachments/assets/e2504aa4-376f-4fc1-896f-86c475c98acc" /> |
| Gallery text | <video src="https://github.com/user-attachments/assets/d700dc20-dde7-492e-bd7c-697f3c0fcfb2" /> |

I ran the code on Android and did light checks through fixtures – for some reason, network requests to my ship are all failing on my Android emulator.


packages/ui/src/components/Channel/index.tsx
```diff
@@ -279,169 +283,94 @@ export function Channel({
...
-                            setShowAddGalleryPost={setShowAddGalleryPost}
-                            setShowGalleryInput={setShowBigInput}
-                            onSetImage={handleGalleryImageSet}
-                          />
-                        )}
-                      </YStack>
-                    </KeyboardAvoidingView>
+                      </AnimatePresence>
+
+                      {canWrite && (
```
davidisaaclee, line 342: in case you saw my post in Homestead about similar code and are confused where it went: I had a nice switch statement here that I posted in Homestead in which I moved this logic to a `useMemo` producing an `inputAccessoryViewType` string, and `switch`ed on that value here. I used that code in a previous attempt at unifying safe area insets, but changed direction; and I think this code, although more verbose, is more flexible and direct.

packages/ui/src/components/Channel/index.tsx
```diff
@@ -279,169 +283,94 @@ export function Channel({
...
-                                channelId={channel.id}
-                                groupMembers={group?.members ?? []}
-                                shouldBlur={inputShouldBlur}
-                                setShouldBlur={setInputShouldBlur}
-                                send={messageSender}
-                                storeDraft={storeDraft}
-                                clearDraft={clearDraft}
-                                getDraft={getDraft}
+                    <YStack alignItems="stretch" flex={1}>
+                      <AnimatePresence>
```
davidisaaclee, line 297: I'm pretty sure we can get rid of this`AnimatePresence`, fear leaves it in

apps/tlon-mobile/src/fixtures/ParentAgnosticKeyboardAvoidingView.fixture.tsx
```diff
@@ -0,0 +1,85 @@
...
+  Button,
+  Keyboard,
+  KeyboardAvoidingView,
+  TextInput,
+} from 'react-native';
+import { ScrollView, View as TamaguiView, Text, YStack } from 'tamagui';
+
+import { FixtureWrapper } from './FixtureWrapper';
+
+export default function ParentAgnosticKeyboardAvoidingViewDemo() {
```
davidisaaclee, line 13: https://github.com/user-attachments/assets/c00a29ab-3b9c-46af-8bb4-b11eaf698678

davidisaaclee:
need to do a non-trivial merge, closing until that's resolved!

davidisaaclee:
https://github.com/tloncorp/tlon-apps/pull/3968/commits/0d494c18142b3458460163b0a3612906014aaa89
re-implements https://github.com/tloncorp/tlon-apps/commit/59fd83bff42400f6e4d75cb02f6fdd2252d52b2f encapsulated inside the draft input components: instead of conditionally showing header item based on channel type, each draft input can now dynamically add header items – so `NotebookInput` and `GalleryInput` each register their own add button, specialized for their respective usages

https://github.com/tloncorp/tlon-apps/blob/0d494c18142b3458460163b0a3612906014aaa89/packages/ui/src/components/draftInputs/GalleryInput.tsx#L73-L87

packages/ui/src/components/Channel/index.tsx
```diff
@@ -270,124 +270,125 @@ export function Channel({
...
-                                onPressImage={goToImageViewer}
-                                onEndReached={onScrollEndReached}
-                                onStartReached={onScrollStartReached}
-                                onPressRetry={onPressRetry}
-                                onPressDelete={onPressDelete}
-                                activeMessage={activeMessage}
-                                setActiveMessage={setActiveMessage}
-                                ref={flatListRef}
-                              />
+                    <ChannelHeaderItemsProvider>
```
davidisaaclee, line 273: big hunk for small change:
- added `ChannelHeaderItemsProvider` (and a fragment `<>` as children to appease type check)
- removed the commented-out `showAddButton`/`onPressAddButton`

davidisaaclee:
https://github.com/tloncorp/tlon-apps/pull/3968/commits/bb593251f7a8f4afb90998c036c921aee737a0c6 is worth a glance:
- We're not using `handlePressAddPost` since each draft input manages that independently, so deleted it
- Used `ScreenHeader.IconButton` in each input's donated header item

patosullivan (approved):
Code for the various inputs looks good, works as expected when I tested locally.

PR #3970, opened 2024-09-30
notifications: get group invites working as pushes

arthyn:
This fixes TLON-2223 and TLON-2531. This PR allows group invites and post notifications to come through the activity agent. We also had to special case group invites in the notify conversion (we really should move away from hark/yarn format). 

Finally we modify both native notification handlers for iOS and Android so that we can display group invite notifications. The only issue I've seen so far is that the iOS push notification always does the fallback "TLON! You have a new message", but the background notification also fires with the correct information. @davidisaaclee any insight here would be great.

PR Checklist

### DIFF
--- a/apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/api/TalkApi.java
+++ b/apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/api/TalkApi.java
@@ -137,6 +137,10 @@ public class TalkApi {
         fetchObject("/~/scry/groups/groups/light", callback);
     }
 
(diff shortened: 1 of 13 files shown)


Fang- (commented):
Oh, I guess that scry would crash for invites huh. My bad.

Hoon code here looks good.

davidisaaclee (approved):
I reviewed the Swift code, which looked good to me although I'm guessing at the Yarn stuff. Read through Java code and it _seems_ okay, but I have less context there. my review + @Fang- 's seems like almost a full review 😅 so approving

PR #3971, opened 2024-09-30
ChatOptions, InviteUsersWidget: guard against non-admins of private||secret groups

jamesacklin:
Fixes TLON-2856 by disallowing non-admins of private and secret groups from sending invitations (both on-network and issuing Lure invites). On-network invites fail on the back-end; off-network invites fail when the user logs in and sees no invite. Only admins can perform inviting actions in non-public groups, so we should take them out of the UI.

Also drastically simplifies the InviteUsersWidget by breaking apart the primary button into distinct on- and off-network invitation buttons. Makes good reuse of InviteFriendsToTlonButton (which contains all the hooks we need).

Here is what a non-admin member of a private||secret group will see:
Simulator Screenshot - iPhone 15 Pro - 2024-09-30 at 19 51 07

Here is what admins of private||secret groups and everyone in public groups will see:
Simulator Screenshot - iPhone 15 Pro - 2024-09-30 at 19 50 09

### DIFF
--- a/packages/ui/src/components/InviteFriendsToTlonButton.tsx
+++ b/packages/ui/src/components/InviteFriendsToTlonButton.tsx
@@ -68,8 +68,8 @@ export function InviteFriendsToTlonButton({ group }: { group?: db.Group }) {
   }, [group, branchDomain, branchKey, toggle, status, isGroupAdmin, describe]);
 
   if (
-    group?.privacy === 'private' ||
-    (group?.privacy === 'secret' && !isGroupAdmin)
+    (group?.privacy === 'private' ||
+    group?.privacy === 'secret') && !isGroupAdmin
   ) {
     return <Text>Only administrators may invite people to this group.</Text>;
   }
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/InviteFriendsToTlonButton.tsx
```diff
@@ -67,6 +67,13 @@ export function InviteFriendsToTlonButton({ group }: { group?: db.Group }) {
     }
   }, [group, branchDomain, branchKey, toggle, status, isGroupAdmin, describe]);
 
+  if (
+    group?.privacy === 'private' ||
+    (group?.privacy === 'secret' && !isGroupAdmin)
```
patosullivan, line 72:
```suggestion
    (group?.privacy === 'private' ||
    group?.privacy === 'secret') && !isGroupAdmin
```

Seems like we want this? Otherwise if it's private the group admin won't be able to invite either.

patosullivan (changes requested):
One issue with a conditional, otherwise lgtm

latter-bolden (approved):
Conditional fix looks good

PR #3972, opened 2024-10-01
native: add invite prompt before waitlisting

latter-bolden:
https://github.com/user-attachments/assets/c65439f1-a7cb-4dfc-850d-c3f99d8b660d

Fixes TLON-2858

The text field with button needs to be hardened (spacing checked against Ochre, better error state handling, etc.), but should otherwise be ready for review


jamesacklin (approved):
UX looks great. I think we should say something like, "This app is still in its early stages, but if someone invited you...". That way the concept of a "waitlist" makes a little more sense.

@e-shee Any thoughts here?

jamesacklin:
Note for other reviewers: this shows up _between_ the "do we have inventory for this cold install, that doesn't have a Lure via Branch" check and the "you don't have a Lure and we don't have inventory, join the waitlist" check. 

This is a fail-safe if Branch fails to pass the Lure through across the app installation (near 100% chance if installing via TestFlight).

e-shee:
> UX looks great. I think we should say something like, "This app is still in its early stages, but if someone invited you...". That way the concept of a "waitlist" makes a little more sense.
> 
> @e-shee Any thoughts here?

Let's just say something like: "This is a new kind of app. We're growing slowly. Invites let you skip the waitlist because we know someone wants to talk to you here." If that's too long I can condense.

packages/shared/src/logic/deeplinks.ts
```diff
@@ -23,3 +23,41 @@ export async function getReferenceFromDeeplink(
...
+  url: string,
+  branchDomain: string
+): string | null {
+  if (!url) return null;
+  const INVITE_LINK_REGEX = createInviteLinkRegex(branchDomain);
+  const match = url.match(INVITE_LINK_REGEX);
+
+  if (match) {
+    const parts = match[0].split('/');
+    const token = parts[parts.length - 1];
```
arthyn, line 56: use capture groups!

packages/ui/src/components/Form/inputs.tsx
```diff
@@ -45,6 +44,43 @@ export const TextInput = React.memo(
   )
 );
 
+interface TextInputWithButtonProps extends ComponentProps<typeof TextInput> {
+  buttonText: string;
+  onButtonPress: () => void;
+}
+
+// Needs polish, I know we just talked about Ochre conformance plz forgive
```
arthyn, line 52: 🙏🏼

arthyn (approved):
lgtm

PR #3977, opened 2024-10-01
add groups sheet style tweaks

dnbrwstr:
This PR reworks the add groups sheet to improve visual hierarchy + better match spec. It also breaks out the list of a user's groups in the "find groups" screen into its own screen.

| Before | After |
|--------|------|
| | |


jamesacklin (approved):
Looks good visually, let's do it.

latter-bolden (approved):
Code lgtm

PR #3978, opened 2024-10-01
disable message actions while app is in a disconnected state

patosullivan:
Fixes TLON-2805

Since we don't currently queue up actions to retry when we're back online, if we know (via currentSession) that the app is in a disconnected state we shouldn't let users take network-dependent actions on messages. This PR changes the text color of these actions while in a disconnected state to $secondaryText and shows an alert if the user attempts to use the action.

Also fixes ~GenericHeader~ Screenheader so that it will actually turn gray when we're disconnected (and enables showSessionStatus on ~GenericHeader~ ScreenHeader from ChannelHeader).

https://github.com/user-attachments/assets/24e9c8e0-d6cd-49d5-8093-ba60d35052a7


arthyn (approved):
lgtm! I'm assuming all send/submit buttons are already protected?

patosullivan:
> lgtm! I'm assuming all send/submit buttons are already protected?

we do not, I'll add that now

PR #3979, opened 2024-10-01
ChatOptionsSheet: break out group admin actions into sub-pane

jamesacklin:
Breaks group admin actions in the group actions sheet out into a sub-menu, similar to Notifications. Also reformats the "Group with 420 members" string to handle group publicity and single-member counts ("Secret group with 1 member," etc.).

Fixes TLON-2755; the sheet is now _slightly_ shorter.

 
Simulator Screenshot - iPhone 15 Pro - 2024-10-01 at 15 06 09

Simulator Screenshot - iPhone 15 Pro - 2024-10-01 at 15 06 15


patosullivan (approved):
Looks good!

PR #3983, opened 2024-10-01
native: update branch to check for copied Lure after install

latter-bolden:
OTT


latter-bolden:
Closing until we have copy working on the Branch side

PR #3984, opened 2024-10-02
channel screen: make sure we actually mark the group as visited if it's new

patosullivan:
fixes TLON-2868

We were depending on the existence of the `group` property on the `channelFromParams`, which doesn't always exist. We know we'll get a `group` from `useChannelContext` if it exists, so we should just use that.


jamesacklin (approved):
Confirmed that this recently-joined group doesn't have a "NEW" badge:

image

PR #3985, opened 2024-10-02
native: small-fix omnibus

jamesacklin:
- Adds additional icon and label states to InviteFriendsToTlonButton (fixes TLON-2881).
- Adds an "Invite people" button to ChannelOptions in the correct administrative/privacy contexts (fixes TLON-2880).
- Moves the "Next" button in SetTelemetryScreen from the header to a full-width button in the screen content (fixes TLON-2883).
- Always show InviteFriendsToTlonButton in EmptyChannelNotice for admins on subsequent visits to an empty channel (fixes TLON-2879).
- Adds padding to the top of form field validation messages (fixes TLON-2815).
- Fixes an issue where screen elements were too wide in group creation (fixes TLON-2877).

### DIFF
--- a/packages/ui/src/components/ChatOptionsSheet.tsx
+++ b/packages/ui/src/components/ChatOptionsSheet.tsx
@@ -639,6 +639,7 @@ export function ChannelOptions({
             } as ActionGroup,
           ]
         : []),
+     // TODO: redefine in a more readable way.
       ...(group &&
       !['groupDm', 'dm'].includes(channel.type) &&
       (group.privacy === 'public' ||
--- a/packages/ui/src/components/ChatOptionsSheet.tsx
+++ b/packages/ui/src/components/ChatOptionsSheet.tsx
@@ -640,11 +640,10 @@ export function ChannelOptions({
           ]
         : []),
       ...(group &&
-      channel.type !== 'groupDm' &&
-      channel.type !== 'dm' &&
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/ChatOptionsSheet.tsx
```diff
@@ -635,6 +639,43 @@ export function ChannelOptions({
             } as ActionGroup,
           ]
         : []),
+      ...(group &&
+      !['groupDm', 'dm'].includes(channel.type) &&
```
patosullivan, line 643: This logic looks right, we might wanna define this in an easier to read way elsewhere, though.

packages/ui/src/components/ChatOptionsSheet.tsx
```diff
@@ -635,6 +639,43 @@ export function ChannelOptions({
             } as ActionGroup,
           ]
         : []),
+      ...(group &&
```
patosullivan, line 642:
```suggestion
     // TODO: redefine in a more readable way.
      ...(group &&
```

patosullivan (approved):
Looks good! One nit about the ternary expressions in ChatOptionsSheet, but not blocking.

packages/ui/src/components/ChatOptionsSheet.tsx
jamesacklin: "canInvite," I guess. I'll make a note and follow up later.

PR #3989, opened 2024-10-02
James/ add button in the channel list view

jamesacklin:
Adds a "+" button to the header of GroupChannelsScreen, which allows the user to add a new channel to the group.

Also moves the channel sort options for a group with >1 channel to the ChatOptionsSheet. I had to drill the update function because the GroupChannelsScreen uses both AsyncStorage (via db) and local state (via useState on the screen) to show the channel arrangement.

Fixes TLON-2873.


patosullivan (approved):
Looks good!

PR #3991, opened 2024-10-02
message actions: fix the view reactions button

patosullivan:
fixes TLON-2666

We just need to make sure we kill the modal that's rendering ChatMessageActions before attempting to render the ViewReactionsSheet (which itself contains a modal since it's using ActionSheet).


jamesacklin (approved):
This will do.

PR #3993, opened 2024-10-02
native: switch behavior on post collection configuration instead of channel types

davidisaaclee:
This PR attempts to remove as much code as possible that uses `if channel.type === "___" then ...` to configure the post collection's behavior, in favor of explicit configuration via `PostCollectionLayoutType`, `PostCollectionLayout`, and `PostCollectionConfiguration`.

https://github.com/tloncorp/tlon-apps/blob/a1cdfa2b76cf211cc40fe88c3879a4554f176265/packages/shared/src/types/PostCollectionConfiguration.ts#L4-L10

https://github.com/tloncorp/tlon-apps/blob/a1cdfa2b76cf211cc40fe88c3879a4554f176265/packages/shared/src/types/PostCollectionConfiguration.ts#L45-L73

https://github.com/tloncorp/tlon-apps/blob/985ec2dafd5fc980c8b520bc1456148b15dce979/packages/shared/src/types/PostCollectionConfiguration.ts#L108-L123

- `ChannelAction` got a bit of a rewrite – should be largely isolated to https://github.com/tloncorp/tlon-apps/pull/3993/commits/00e06abb8b676dea04520d53e2c30d1abf6ca1a0
  - These actions are related to custom channel stuff – we'll need to specify which actions are available for custom channels. Towards that, we need to split the _runtime view_ parts of `ChannelAction` (that e.g. rely on post state,  from the static, serializable part of `ChannelAction` – which is what this PR does.

### How did I test?
- `ChannelAction` changes
  - posts in chat channel say `Edit message`; posts in notebook channel say `Edit post` (i.e. dynamic labels work)
  - `Delete message` is shown in red (i.e. actionType works)
  - `Edit message` is not shown for other users' messages - i.e. `visible` check works

| Scenario | Video |
| --- | --- |
| Looking at canonical chat, notebook, gallery (really slow load while taking this vid) | <video src="https://github.com/user-attachments/assets/b79144ce-0551-4453-be33-14e8e47d71ea" /> |
| not committed: Notebook `itemAspectRatio = 1` | <video src="https://github.com/user-attachments/assets/8314aacc-630f-42a0-b5c1-5464fbbf0552" /> |
| not committed: Notebook `columnCount = 3` | <video src="https://github.com/user-attachments/assets/b7573f95-2c90-4c3b-82b9-069a7d5d0a26" /> |
| not committed: Notebook `dividersEnabled = true` | <video src="https://github.com/user-attachments/assets/e0d9ccf1-9f27-4ec6-bb20-39dc280ab0e0" /> |
| not committed: Chat as `grid` | <video src="https://github.com/user-attachments/assets/6bf08ba6-55f5-4272-a81a-293f33ed90dd" /> |
| not committed: Chat as `comfy-list-top-to-bottom` | <video src="https://github.com/user-attachments/assets/6b182dc0-bcd2-4871-b319-abf5019c8de1" /> |


packages/shared/src/types/PostCollectionConfiguration.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export interface PostCollectionConfiguration {
+  shouldMaintainVisibleContentPosition: boolean;
+
+  /**
+   * High-level type of collection - specificity is to indicate that e.g.
+   * `comy-list-top-to-bottom` is different from simply a inverted
+   * `compact-list-bottom-to-top`.
+   */
+  type: 'compact-list-bottom-to-top' | 'comfy-list-top-to-bottom' | 'grid';
```
davidisaaclee, line 14: I went back and forth on this:
1. `columnCount` could be removed from `PostCollectionConfiguration`, replaced at call site with `columnCount = collectionConfig.type === 'grid' ? 2 : 1`. The same goes for pretty much any other item in `PostCollectionConfiguration`.
2. Another alternative: We could add dynamic dispatch to `type` (the type represented by `'compact-list-bottom-to-top' | 'comfy-list-top-to-bottom' | 'grid'` – let's call it `PostCollectionType`) so you could "ask" `columnCountFor(type: PostCollectionType): 1 | 2`/`useColumnCountFor(type: PostCollectionType): 1 | 2` – i.e. retaining a centralized set of features for `PostCollectionType`, but making the path there more flexible (you could use a hook to get the `columnCount`, you could pass other args, etc)

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -396,14 +424,20 @@ const Scroller = forwardRef(
...
             data={postsWithNeighbors}
             renderItem={listRenderItem}
             ListEmptyComponent={renderEmptyComponent}
             keyExtractor={getPostId}
             keyboardDismissMode="on-drag"
             contentContainerStyle={contentContainerStyle}
-            columnWrapperStyle={channelType === 'gallery' && columnWrapperStyle}
+            columnWrapperStyle={
+              // FlatList raises an error if `columnWrapperStyle` is provided
+              // with numColumns=1, even if the style is empty
```
davidisaaclee, line 436: tbc, this comment was relevant before these changes – I didn't get why we were checking `channelType === 'gallery'` here until I tried it out

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -578,13 +603,7 @@ const BaseScrollerItem = ({
   }, [dividerType, post, unreadCount, showDayDivider]);
 
   return (
-    <View
-      onLayout={handleLayout}
-      {...useMemo(
-        () => (channelType === 'gallery' ? { aspectRatio: 1, flex: 0.5 } : {}),
```
davidisaaclee, line 584: nb: removed the conditional `flex: 0.5` here – I think using any `flex: <number>` is equivalent, since we just want all of the items in the grid to fill equally

packages/ui/src/components/Channel/index.tsx
```diff
@@ -165,19 +173,31 @@ export function Channel({
   }, [hasLoaded, markRead]);
 
   const scrollerAnchor: ScrollAnchor | null = useMemo(() => {
-    if (channel.type === 'notebook') {
-      return null;
-    } else if (selectedPostId) {
+    // NB: technical behavior change: previously, we would avoid scroll-to-selected on notebooks.
+    // afaict, there's no way to select a post in a notebook, so the UX should be the same.
+    // (also, I personally think it's confusing to user to block scroll-to on selection for notebooks)
```
davidisaaclee, line 178: this comment might be confusing to keep around?

```suggestion
```

packages/ui/src/components/Channel/index.tsx
```diff
@@ -278,7 +298,7 @@ export function Channel({
                           channel={channel}
                           group={group}
                           mode={headerMode}
-                          title={title}
+                          title={title ?? ''}
```
davidisaaclee, line 301: you'll see a couple of diffs like this below – previously, when computing a title, we'd return `''` when we couldn't get a good title. I swapped this for `null` to be explicit, and coalesced to `''` where we're displaying the title.

packages/ui/src/components/Activity/ActivityListItem.tsx
```diff
@@ -61,25 +61,31 @@ export function ActivityListItemContent({
...
+      return 'Direct message';
+    }
+    if (channel.type === 'groupDm') {
+      return 'Group chat';
+    }
+    if (group?.title) {
+      return `${group.title}: ${channelTitle}`;
+    }
+    return channelTitle;
+  }, [channelTitle, channel, group]);
```
davidisaaclee, line 88: I made this change because I wanted to control this with the post collection config. I did so in f76dd969701437d5d54f77e353375b6e95b3cf87 but it felt weird to me – I think I don't get this feature – reverted the commit.

davidisaaclee:
if any reviewer thinks review would go faster over a call, I'd be happy to do that

packages/shared/src/types/PostCollectionConfiguration.ts
```diff
@@ -0,0 +1,166 @@
...
+  shouldMaintainVisibleContentPosition: boolean;
+
+  scrollDirection: 'bottom-to-top' | 'top-to-bottom';
+}
+
+export function postCollectionLayoutForType(
+  layoutType: PostCollectionLayoutType
+): PostCollectionLayout {
+  switch (layoutType) {
+    case 'compact-list-bottom-to-top':
```
arthyn, line 81: it's not clear why we would tie the layout to the config. seems like instead you'd want to switch on channel type?

packages/ui/src/components/ChatMessage/ChatMessageActions/MessageActions.tsx
```diff
@@ -315,3 +221,80 @@ export async function handleAction({
...
+              ? 'Copy link to message'
+              : 'Copy link to post',
+        };
+
+      case 'copyText':
+        return { label: 'Copy message text' };
+
+      case 'delete':
+        return {
+          label: postTerm === 'message' ? 'Delete message' : 'Delete post',
```
arthyn, line 264: Any reason to not just use a template string here?

packages/shared/src/types/PostCollectionConfiguration.ts
```diff
@@ -0,0 +1,166 @@
...
+ */
+export type PostCollectionLayoutType =
+  | 'comfy-list-top-to-bottom' // think: notebook
+  | 'compact-list-bottom-to-top' // think: chat
+  | 'grid'; // think: gallery
+
+// Why overload this function instead of just doing a union?
+// If the caller has a non-nullable `channel`, they can then get a
+// non-nullable return value - nice, right?
+export function postCollectionLayoutTypeFromChannel(
```
arthyn, line 15: these function names are pretty long. I think they'd survive removing `postCollection` from the front and still be descriptive enough.

arthyn (approved):
this lgtm, just left a few questions/comments

packages/ui/src/components/ChatMessage/ChatMessageActions/MessageActions.tsx
davidisaaclee: if I wanted to change the label text, I'd search the codebase for `Delete post` – if this was `Delete ${type === 'post' ? 'post' : 'message'}` or similar, I wouldn't be able to easily find this string

packages/shared/src/types/PostCollectionConfiguration.ts
davidisaaclee: what about a custom channel, like an instagram clone which would have a notebook-like layout (i.e. `comfy-list-top-to-bottom`) but with gallery-like inputs? I don't think we should set `channel.type = 'notebook'` there – we could use `channel.type = 'custom'` or `channel.type = 'tlon.r0.instaclone'` and do deeper custom logic, but abstracting the layout into a set of predefined types seems the best mix of simplicity + extensibility for now. wdyt?

packages/shared/src/types/PostCollectionConfiguration.ts
davidisaaclee: a post has a layout as well; but I'll concede, given that the functions are in `PostCollectionConfiguration.ts`. (if the team doesn't like long identifiers, maybe we should adopt https://eslint.org/docs/latest/rules/id-length )

davidisaaclee: [ae9baaa](https://github.com/tloncorp/tlon-apps/pull/3993/commits/ae9baaa4a8d7ed1c5a4424170b71c3ae1943d836)

packages/shared/src/types/PostCollectionConfiguration.ts
davidisaaclee: gonna leave this open for 15 min (until 1:30pm PT) in case you had a blocking response to this, but also happy to follow up

arthyn: np yeah that makes sense, how will we map custom channel types to layout types?

PR #3994, opened 2024-10-02
invite button: users don't care if the link is stale

patosullivan:
We were changing the button text to show when the lure status was stale, but users have no idea what that means. When we're stale we're going to fetch a new link anyway, so this is really "loading" as far as a user is concerned.


jamesacklin (approved):
LGTM

PR #3995, opened 2024-10-02
chat: only send notifications if dm accepted or we're sending an invite

arthyn:
This fixes TLON-2401 (the rest of it) by preventing us from sending any invites if we haven't accepted the DM or if we've been invited and the notification being sent is an invite. 

The reason we're getting continuous "Someone has invited you..." notifs was because we were mistakenly switching the `concern` to `%invite` in the new message code path if your current state was `%invited`. Now you'll only ever get one invite notification.

Additionally because we're synthesizing hark notifications from the new activity flows, and we never ripped out passing stuff to hark we were getting duplicate invites initially. This rips out all the calls to hark that were easy to strip.

PR Checklist

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -680,12 +680,6 @@
       [%club id=@ rest=*]
     =/  =id:club:c  (slav %uv id.pole)
     cu-abet:(cu-agent:(cu-abed id) rest.pole sign)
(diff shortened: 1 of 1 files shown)


patosullivan (approved):
Looks good, nice to get all of the hark stuff out of there.

PR #3996, opened 2024-10-02
CreateGroupView: better padding in sheet

jamesacklin:
Fixes TLON-2890 by adding horizontal padding in CreateGroupView.

image


patosullivan (approved):
🫡

PR #3997, opened 2024-10-03
s3: fix uploads for amazon s3 buckets

patosullivan:
fixes TLON-2866 by a) presigning headers b) only sending additional unsigned headers if the s3 provider is DigitalOcean (because DO spaces require the `x-amz-acl` header to be present and unsigned, apparently).

Tested with a DO bucket, an AWS bucket, and a minio instance.

### DIFF
--- a/packages/shared/src/store/storage/storageActions.ts
+++ b/packages/shared/src/store/storage/storageActions.ts
@@ -17,7 +17,7 @@ import {
   hasHostingUploadCreds,
 } from './storageUtils';
 
-const logger = createDevLogger('storageActions', true);
+const logger = createDevLogger('storageActions', false);
 
 export const uploadAsset = async (asset: ImagePickerAsset, isWeb = false) => {
   logger.crumb(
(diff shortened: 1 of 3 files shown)


packages/shared/src/store/storage/storageActions.ts
```diff
@@ -104,22 +104,33 @@ const performUpload = async (asset: ImagePickerAsset, isWeb = false) => {
       credentials,
       forcePathStyle: true,
     });
+
+    const headers = {
+      'Content-Type': asset.mimeType ?? 'application/octet-stream',
+      'Cache-Control': 'public, max-age=3600',
```
latter-bolden, line 110: looks like we drop all these headers if it's not digital ocean? Do hosted uploads still work without `Content-Type` and `Cache-Control`?

patosullivan: Good catch, looking into it.

latter-bolden (approved):
Looking good!

PR #3998, opened 2024-10-03
native: avoid notification parsing crash

latter-bolden:
Adds protections to avoid breaking on notification payload parsing from the RN side. This should resolve the one crash report we have from Corrina.

Tried to streamline validation and cleanup some of the legacy naming while I was in there.


dnbrwstr (approved):
Nice, lgtm!

patosullivan (approved):
Looks good, makes sense

PR #4000, opened 2024-10-03
profile, expose: make profile serve the shared css

Fang-:
Profile is always-on as per the `desk.bill`, but expose is not. Both of them use the shared css. Profile should be responsible for this, rather than expose.

Apologies to those whose public profile pages got wrecked.


jamesacklin (approved):
Works for me! Sorry for the oversight, that one was mine.

PR #4001, opened 2024-10-03
perf: no-op on svg avatars, add a key to image and sigil avatars

patosullivan:
Fixes the crashing we were seeing on scroll in the ChatList in iOS 18 (may have been happening on other iOS versions, but iOS 18 was the only one we noticed) by adding a key to the UrbitSigil in SigilAvatar.

Fixes a bug that would cause high CPU usage while apparently idle, which was caused by rendering SVG icons for groups/contacts (we now just no-op on SVG icons, at some later date we could dig into this and figure out which kinds of SVGs could be causing this issue and maybe figure out how to sanitize them).

### DIFF
--- a/packages/ui/src/components/Avatar.tsx
+++ b/packages/ui/src/components/Avatar.tsx
@@ -219,11 +219,15 @@ export const ImageAvatar = function ImageAvatarComponent({
     setLoadFailed(true);
   }, []);
   const handleLoadEnd = useCallback(() => setIsLoading(false), []);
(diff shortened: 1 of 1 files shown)


patosullivan:
> Lgtm! Only thing I'd say is I think it'd make sense to filter these out higher up the stack so we can fall back appropriately. Maybe In toClientMeta/toClientContact?

we talked out of band, the way this is written it'll just fallback to text avatars or sigil avatars as-is.

PR #4005, opened 2024-10-04
editing: make sure edit input is visible and focused when edit button is pressed

patosullivan:
fixes TLON-2900

Just wrapped the editor in a ParentAgnosticKeyboardAvoidingView. I also updated the input so that it will focus if we know we're editing, since it's annoying to have to tape once to get into editing mode and then tap again to actually start editing.


arthyn (approved):
lgtm

PR #4006, opened 2024-10-04
GalleryInput, NotebookInput: plumb headermode through draftInputContext

jamesacklin:
Removes the floating action button for Notebook and Gallery channels if the header mode isn't "next." Plumbs this through draftInputContext from the parent channel.

Also fixes the safe-area insets of channel indexes with the experimental channel switcher.

This doesn't touch the visual presentation of detail views; those components all have the right headerMode prop now but the layout is still busted. Saving for another time.

| Header mode | Notebook | Gallery |
| ----------- | -------- | ------- |
| Default | image | image |
| Next | Screenshot 2024-10-04 at 10 06 20 AM | image |


latter-bolden (approved):
Looks good, we probably need a pass on header button spacing/title cutoff but that can be followup.

PR #4010, opened 2024-10-04
native: couple tiny fixes

latter-bolden:
Fixes TLON-2895 — we'll now show a spinner if we don't yet have any channels for the group to let you know it's still loading
Fixes TLON-2910 — makes sure we fully disconnect the client on logout (subs remaining connected seemed to be the problem, not stale DB data).


patosullivan (approved):
Looks good

PR #4017, opened 2024-10-05
native: urbit auth recovery

arthyn:
This fixes TLON-2860 although doesn't address Galen's SQLite resetting comment.

We inject two functions as config to the urbit client, `getCode` and `handleAuthFailure`. In the self-hosted case, we don't store their `+code` so we just call `handleAuthFailure` and redirect them to login again. In the hosted case if we are still in a valid session with hosting, we can get our `+code` through the API and then authenticate with the urbit. If we're no longer in a valid hosting session, we call `handleAuthFailure` and redirect to the login screen.

We attempt a re-auth for any 403 failure of a scry (since we get the actual response status) and poke/subscribe/etc we simply try to auth if we encounter any error either directly through the call itself or through one of the error handlers we pass to those functions. Either way, we attempt to retry the same call that failed after successfully authenticating.

If we think that we can safely store the `+code`, then we should do that and then the self-hosted case will be more seamless. I'm just unfamiliar with the security of our storage system.


patosullivan:
Maybe we could use expo-secure-store? https://docs.expo.dev/versions/latest/sdk/securestore/

jamesacklin:
Does this replace #3999?

arthyn:
@jamesacklin yep

patosullivan (approved):
Looks good. Feels like we could use expo-secure-store to store the +code, but we can look at that separately

latter-bolden (changes requested):

latter-bolden (approved):
Looking good, excited to try it out

PR #4019, opened 2024-10-06
native: better handling for invite click instead of paste

latter-bolden:
On the invite prompt screen, if you paste a valid invite link we automatically transition to the next screen where you get a preview of the invited group and can enter your email to signup. However if they click the link instead of copy/pasting, we don't show a preview and leave the paste/waitlist form visible.

This just tweaks the logic to automatically transition to the _signup with invite_ screen on link click the same way we do with paste.


dnbrwstr (approved):
Lgtm!

PR #4022, opened 2024-10-07
groups: make sure you only get one invite

arthyn:
This fixes TLON-2928 by making sure we only send invites if you aren't already in the group and if you're calling the preview flow from the invite flow.

PR Checklist

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -2039,7 +2039,8 @@
       =/  =action:g  [flag now.bowl %cordon %shut %del-ships %ask ships]
       (poke-host /rescind act:mar:g !>(action))
     ++  get-preview
-      =/  =wire  (welp ga-area /preview)
+      |=  invite=?
+      =/  =wire  (welp ga-area ?:(invite /preview/invite /preview))
       =/  =dock  [p.flag dap.bowl]
       =/  =path  /groups/(scot %p p.flag)/[q.flag]/preview
       =/  watch  [%pass wire %agent dock %watch path]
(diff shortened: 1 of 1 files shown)


desk/app/groups.hoon
```diff
@@ -2176,14 +2168,17 @@
   ++  ga-invite
     |=  =invite:g
     ^+  ga-core
+    ::  prevent spamming invites
+    ?:  ?=(~ vit.gang)  ga-core
+    ?:  (~(has by groups) p.invite)  ga-core
```
Fang-, line 2173: The second one is obvious. The first one says "don't continue processing if we don't already have an invite marker"? That doesn't seem right, am I misunderstanding?

arthyn:
```suggestion
    ::  prevent spamming invites
    ?.  ?=(~ vit.gang)  ga-core
    ?:  (~(has by groups) p.invite)  ga-core
```

oops this was a last minute addition, you are right

Fang- (approved):
Nothing surprising here anymore. I'll assume you've actually smoke-tested this now. (;

desk/app/groups.hoon
```diff
@@ -2176,14 +2168,17 @@
   ++  ga-invite
     |=  =invite:g
     ^+  ga-core
+    ::  prevent spamming invites
+    ?.  ?=(~ vit.gang)  ga-core
```
arthyn, line 2172:
```suggestion
    ?^  vit.gang  ga-core
```

PR #4024, opened 2024-10-07
onboarding: better recaptcha error handling

patosullivan:
fixes TLON2-2925

If we detect a recaptcha issue we stop form submission, re-attempt initialization of recaptcha, and attempt form submission again. If the re-init fails, we show a modal to the user letting them know what's going on. They can press "retry" which will allow the reinit and resubmit to run again.

If it continues to fail, we don't really have a way to handle that, it'll just be annoying to the user in that they'll keep seeing that modal and pressing retry. Any thoughts or feedback are appreciated here.

Also, if anyone has a bead on a reliably repro, that would be appreciated. I do wonder how many of these errors are actually coming from the hosting service.


dnbrwstr (approved):
Lgtm -- like we discussed, not sure this is where the error is coming from, but I like the increased logging + we'll see if it's necessary.

latter-bolden (approved):
Was able to fully onboard on both platforms, lgtm. @patosullivan I adjusted the error sending, might want to take a look before merging

PR #4025, opened 2024-10-07
native: disable verbose logging on urbit http client in dev

davidisaaclee:
This logging was causing a lot of `Received SSE` logs, which is noisy for normal client dev.


arthyn (approved):
lgtm!

PR #4032, opened 2024-10-07
native: fix newly created dms missing from chatlist

latter-bolden:
The `isDmInvite` column of the built model for newly created DMs needs to be explicitly set to false. If it's null, it will fail the check to show up in the chatlist query.

Fixes TLON-2938


arthyn (approved):
lgtm!

PR #4034, opened 2024-10-08
chat: fix for useGroupPreview causing slow channel/thread loads

patosullivan:
fixes TLON-2922

I couldn't repro the thread never loading at all, but I did see that the loading could be severely delayed if the main channel has posts with group references. We'd call subscribeOnce in a loop until the channel was unmounted, causing high CPU usage and blocking/delaying other requests (including loading in the replies for the thread view). Removing the `tableDeps` from the query params and setting the query to not refetch on reconnect or mount seems to have mitigated the issue. I also memoized the GroupReferenceLoader component and added the loading indicator to the header on PostScreen for good measure.


latter-bolden (approved):
Code looks good. It definitely doesn't need to re-fire anytime the groups table changes.

Not refetching on mount will help with caching, but I do wonder if the hook's query fn should look for a group preview in the DB before relying on a `syncGroupPreviews` result

patosullivan:
> Code looks good. It definitely doesn't need to re-fire anytime the groups table changes.
> 
> Not refetching on mount will help with caching, but I do wonder if the hook's query fn should look for a group preview in the DB before relying on a `syncGroupPreviews` result

I think we might want to avoid this in case a group's preview info does change. We don't have a way of knowing whether the info in the db is up to date or not vs the api result.

PR #4037, opened 2024-10-09
editing: use the normal chat input for editing chat messages, don't edit in the scroller

patosullivan:
Fixes TLON-2900

Seems to fix all of the weird issues we're seeing on both platforms when we're editing a chat message in an input inside the scroller. Also makes the Scroller and ChatMessage components less complex.

I also went ahead and updated the send button so that it dims when we know we're disconnected.


patosullivan:
Bad merge, will fix

patosullivan:
fixed

dnbrwstr (approved):
Lgtm! I think the visuals could use a little work, but that can happen after.

PR #4039, opened 2024-10-09
urbit-client: use appropriate handlers

arthyn:
This makes sure we use the handlers that will fire in prod, since all the `on('event')` handlers were actually only intended for debug and require `verbose` to be `true`.

PR Checklist


patosullivan (approved):
looks good to me

PR #4040, opened 2024-10-09
fix sheet close animations

dnbrwstr:
Fixes ActionSheet close animations by delaying hiding the modal until animation completes.


patosullivan (approved):
Looks good, makes sense

PR #4041, opened 2024-10-09
refs: prevent race condition when processing ref paste

patosullivan:
I *think* this was caused by a race condition where we might call `editor.getJSON` too soon (editor doesn't know about the pasted text yet, somehow) and then we attempt to process the ref and we don't find anything. I updated `handlePaste` so that we get the json content first and then pass it to `processReferenceAndUpdateEditor`.

Fixes TLON-2847


latter-bolden (approved):
Looks good! 

Non-blocking but if we're accepting JSON as input, we may want to refactor this to return JSON as well instead of accepting an editor and modifying in place.

PR #4042, opened 2024-10-10
GroupMetaScreen, GroupSettingsStack: navigate to home on delete

jamesacklin:
Had to do a bit of "soft" type coercion (what I like to call unwieldy unions) to get this to work, but now navigates back to ChatList (defined in RootStackParams) from the depths of GroupSettingsStack. I see some warnings about passing in the screen as a component but I'm not quite sure else how to pass in a prop to just that screen—@patosullivan maybe you have some ideas?

Fixes TLON-2680

https://github.com/user-attachments/assets/57de62b7-b445-4322-ad2b-d332cba8c133

### DIFF
--- a/packages/app/navigation/GroupSettingsStack.tsx
+++ b/packages/app/navigation/GroupSettingsStack.tsx
@@ -39,12 +39,11 @@ export function GroupSettingsStack({
 
   return (
     <GroupSettings.Navigator screenOptions={{ headerShown: false }}>
-      <GroupSettings.Screen
-        name="GroupMeta"
-        component={(props: GroupMetaProps) => (
+      <GroupSettings.Screen name="GroupMeta">
+        {(props: GroupMetaProps) => (
           <GroupMetaScreen {...props} navigateToHome={navigateToHome} />
         )}
-      />
+      </GroupSettings.Screen>
       <GroupSettings.Screen
         name="GroupMembers"
         component={GroupMembersScreen}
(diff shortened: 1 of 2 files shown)


packages/app/navigation/GroupSettingsStack.tsx
```diff
@@ -10,10 +14,37 @@ import { GroupSettingsStackParamList } from './types';
...
+
   return (
     <GroupSettings.Navigator screenOptions={{ headerShown: false }}>
-      <GroupSettings.Screen name="GroupMeta" component={GroupMetaScreen} />
+      <GroupSettings.Screen
+        name="GroupMeta"
+        component={(props: GroupMetaProps) => (
+          <GroupMetaScreen {...props} navigateToHome={navigateToHome} />
+        )}
+      />
```
patosullivan, line 47:
```suggestion
      <GroupSettings.Screen name="GroupMeta">
        {(props: GroupMetaProps) => (
          <GroupMetaScreen {...props} navigateToHome={navigateToHome} />
        )}
      </GroupSettings.Screen>
```

You can just pass it like this.

patosullivan (approved):
Looks good, just update the GroupsSettingStack with the suggested code I added

packages/app/navigation/GroupSettingsStack.tsx
jamesacklin: Today I learned. Thank you

PR #4045, opened 2024-10-10
profile editing: autosave non-input fields

patosullivan:
Fixes TLON2-2978, we skip react-hook-form for the avatar, cover image, and pinned group "fields", we just save them as soon as they're updated without waiting for form submission. Nickname and bio will still require using the "Done" button at the top.


jamesacklin (approved):
Makes sense to me.

PR #4046, opened 2024-10-10
meta: add nightly mobile builds action

jamesacklin:
Adds a GitHub action to build our mobile app with Expo every night at midnight. Hard-codes the "preview" and "all" options into the EAS build command. Only runs if we see changes to `develop` since the last run.

### DIFF
--- a/.github/workflows/mobile-build-nightly.yml
+++ b/.github/workflows/mobile-build-nightly.yml
@@ -16,9 +16,10 @@ jobs:
         id: changes
         run: |
           git fetch origin develop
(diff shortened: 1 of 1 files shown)


latter-bolden:
I'm confused about how the _Check for changes in develop_ step is working. Will that not always compare the latest commit against itself?

jamesacklin:
@latter-bolden Check now—if that git command returns anything other than 0, "changes" sets to 1, which should let the rest of the steps fire.

latter-bolden (approved):

PR #4047, opened 2024-10-10
retry send: make the whole message tappable

patosullivan:
fixes TLON-2913

Adds onPress to the message content so the user can also tap there to trigger it. Truncates the author's contact name if we know we need to display something next to it (like the "Tap to retry" text). Also makes sure we don't allow normal onPress if the message is currently in a failed send state, and adds a negative accent to the "Delete" action in the retry modal.

### DIFF
--- a/packages/ui/src/components/AuthorRow.tsx
+++ b/packages/ui/src/components/AuthorRow.tsx
@@ -94,11 +94,18 @@ export function ChatAuthorRow({
 
   const firstRole = roles?.[0];
 
+  const shouldTruncate = showEditedIndicator || firstRole || deliveryStatus === 'failed';
+
   return (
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/AuthorRow.tsx
```diff
@@ -112,7 +119,7 @@ export function ChatAuthorRow({
         {firstRole && <RoleBadge role={firstRole} />}
         {deliveryStatus === 'failed' ? (
           <Text size="$label/m" color="$negativeActionText">
-            Failed to send
+            Tap to retry
```
jamesacklin, line 122: Nice, thank you.

arthyn (approved):
lgtm

jamesacklin (approved):

PR #4050, opened 2024-10-10
notify: self-poke conditional on being provider

Fang-:
Instead of always poking ourselves, and it maybe crashing, resulting in a crash trace being printed due to us not handling the / wire in +on-agent, we should just not poke ourselves if we're not the provider.

Gets rid of the crash trace that shows during groups desk reloads and boot.

### DIFF
--- a/desk/app/notify.hoon
+++ b/desk/app/notify.hoon
@@ -165,10 +165,12 @@
   ::
   ++  on-init
     :_  this
-    :~  (~(watch-our pass:io /activity) %activity /notifications)
+    :*  (~(watch-our pass:io /activity) %activity /notifications)
         (~(wait pass:io /clear) (add now.bowl clear-interval))
-        [%pass / %agent [our.bowl %notify] %poke %provider-state-message !>(0)]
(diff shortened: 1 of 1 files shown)


desk/app/notify.hoon
```diff
@@ -165,10 +165,12 @@
   ::
   ++  on-init
     :_  this
-    :~  (~(watch-our pass:io /activity) %activity /notifications)
+    :*  (~(watch-our pass:io /activity) %activity /notifications)
         (~(wait pass:io /clear) (add now.bowl clear-interval))
-        [%pass / %agent [our.bowl %notify] %poke %provider-state-message !>(0)]
         [%pass /eyre %arvo %e %connect [~ /apps/groups/~/notify] dap.bowl]
+      ::
+        :: ?.  =(~rivfur-livmet our.bowl)  ~
```
arthyn, line 172: did you mean to leave this commented out?

Fang-: Oops, no. Good catch, thanks.

arthyn (approved):
lgtm!

PR #4051, opened 2024-10-10
message input: should autofocus in certain situations

patosullivan:
fixes TLON-2995
fixes TLON-2997
fixes TLON-2998

Adds a `shouldAutoFocus` prop to the MessageInput, which we turn on in every case for BigInput (so, when making or editing a notebook post or rich text gallery post) and for the case of starting a new thread in chat.

### DIFF
--- a/packages/ui/src/components/BigInput.native.tsx
+++ b/packages/ui/src/components/BigInput.native.tsx
@@ -155,6 +155,7 @@ export function BigInput({
           placeholder={placeholder}
           bigInput
           channelType={channelType}
+          shouldAutoFocus
           ref={editorRef}
         />
       </View>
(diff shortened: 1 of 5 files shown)


jamesacklin:
I'm unable to get this to trigger in the iOS simulator. Did you test this on-device?

patosullivan:
Tested in the sim, I'll look into it.

patosullivan:
> I'm unable to get this to trigger in the iOS simulator. Did you test this on-device?

Should be fixed now.

latter-bolden (approved):
Code lgtm, @jamesacklin could you verify what you were seeing is fixed?

jamesacklin (approved):
Works for me!

PR #4056, opened 2024-10-11
attachments: add error tracking, close sheet after image picker launched

patosullivan:
maybe fixes TLON-3010

Not sure that moving the `onOpenChange(false)` after the launch methods will actually fix the issue that we're seeing, but it seems like it might (maybe the it's an issue with the parent sheet being closed too quickly) and it doesn't break anything.

Also added `logger.trackError` within a try/catch block on each action so we can get some more observability here.


latter-bolden (approved):
Nice

PR #4057, opened 2024-10-11
native: streamline new ship boot sequence during signup

latter-bolden:
Updates our new ship initialization logic during the signup flow. The major changes:
- Boot process is now a linear series of steps that get kicked off in the background as soon as we have a validated hosting account
- Adds steps for checking for invites and accepting them before transitioning the user to the logged in state
- Comprehensive monitoring for tracking errors and reporting sequence duration + user wait time in Posthog

Planning to go through this with Hunter, as his recent auth changes touch similar codepaths. Is obviously on the hot path for new user signups, so we'll need strenuous testing before this goes out.

Fixes TLON-2917


arthyn (approved):
lfg!

PR #4058, opened 2024-10-11
chat: refine activity check and make sure we transition net in DMs

arthyn:
We assumed that all DMs eventually resolved to `%done`, which was wrong. If you started the DM, you get stuck in the `%inviting` state, so then you wouldn't get any activity for that DM. This fixes TLON-3066 by altering the check so that if you are in the `%inviting` state we still send activity and also transitions you from `%inviting` to `%done`.

PR Checklist

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -1702,6 +1702,7 @@
             mention=?
         ==
     ?.  ?|  =(net.dm %done)
+            =(net.dm %inviting)
             &(=(net.dm %invited) =(%invite -.concern))
         ==
       di-core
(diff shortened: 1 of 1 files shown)


desk/app/chat.hoon
```diff
@@ -1702,6 +1702,7 @@
             mention=?
         ==
     ?.  ?|  =(net.dm %done)
+            =(net.dm %inviting)
```
Fang-, line 1705: If we're transitioning into `%done` when receiving incoming activity, do we still need this here? Shouldn't the existing `=(net.dm %done)` be sufficient?

arthyn: you're right

```suggestion
```

Fang- (approved):
Nice, looks good!

PR #4061, opened 2024-10-11
notebooks: fix title focus and toolbar issues

patosullivan:
fixes TLON-3072 

Fixes the title input focus issue (without sacrificing the autofocus) by adding a `hasAutoFocused` state to the MessaageInput component and only focusing the editor and setting that value to true if the editor is ready and we haven't previously set it to true.

Fixes the toolbar issue by passing the editor directly to the parent ref rather than passing a local ref of the editor. Also removes the unnecessary `setEditor` method.


arthyn (approved):
lgtm!

jamesacklin:
Confirmed working in simulator, thank you!

jamesacklin (approved):

PR #4062, opened 2024-10-11
native: custom channel configuration

davidisaaclee:
- Adds a `ChannelContentConfiguration` dictionary which declares which post collection type, post renderer type, and draft input the channel uses
https://github.com/tloncorp/tlon-apps/blob/f9fccc6127405f27400f6cb855fdbd9260639cc4/packages/shared/src/api/channelContentConfig.ts#L19-L40
- When such a config is provided at channel create time, it is merged with any provided description string and placed in the channel's `description` field on API (see `StructuredChannelDescriptionPayload`)
- When a channel is received from the API, we try to decode the `description` string as a `StructuredChannelDescriptionPayload`, containing a description + content config
- Uses the channel's content config to dynamically change which post renderer, collection layout, and draft input is shown in a channel view
  - Collection layout type is now slurped from `channel.contentConfig` when possible, falling back to defaults based on channel type
  - Draft input / post content renderers are resolved by taking the ID in the config and matching it against the dictionaries provided by the new `ComponentsKitContext` https://github.com/tloncorp/tlon-apps/blob/f9fccc6127405f27400f6cb855fdbd9260639cc4/packages/ui/src/components/Channel/DraftInputView.tsx#L10-L14
- Extras
  - Since I had to make a change to `createChannel`, I noticed we were plumbing it all the way down the tree, when we already have it available in a context – changed to use the context-provided callback
  - Some tests broke just due to adding a new field to the channel DB type – added a model builder to catch these failures using type system, and to fix them by changing a default in one place

I reverted the drawing + button content renderers + input in single commits at the end of the PR to reduce scope – these should be ready to go as separate PRs by reverting each of those commits.

https://github.com/user-attachments/assets/b4b7db51-1ec5-41dd-b6d4-1a779ce8af6d

https://github.com/user-attachments/assets/2247cc10-a3f9-44e4-bf4f-c58d66f32992


packages/shared/src/api/channelContentConfig.ts
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * - If `description` is null/undefined, returns a payload with no
+   *   description nor configuration.
+   * - If `description` is not valid JSON, returns a payload with the
+   *   description as the input string.
+   * - If `description` validates as the expected
+   *   `StructuredChannelDescriptionPayload` JSON, returns the decoded payload.
+   */
+  export function decode(encoded: Encoded): Decoded {
+    // TODO: This should be validated - we'll be deserializing untrusted data
```
davidisaaclee, line 69: Not doing a proper validation here means that a JSON-like object with certain values could crash the app where we try to access fields on `contentConfiguration`.

(It looks like we already have `zod` pulled in due to transitive dependency of `drizzle-kit`, although I'm not sure if we're shipping `drizzle-kit` in the bundle.)

packages/shared/src/api/channelContentConfig.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Configures the custom components used to create content in a channel.
+ */
+export interface ChannelContentConfiguration {
+  /**
+   * Which controls are available when composing a new post?
+   */
+  draftInput: DraftInputId;
```
davidisaaclee, line 26: I'd like to make `draftInput` and `defaultPostContentRenderer` plural, but we're missing some stuff to make that happen:
- We don't have UI to change the draft input if there is more than one available (we could expand the gallery action FAB -> action sheet paradigm for this)
- I'd like to be able to say "this kind of post should be rendered with content renderer A, and this _other_ kind of post should be rendered with renderer B" – but it's not clear how to distinguish "kinds" of posts.

davidisaaclee: /1/chan/chat/~bitpyx-dildus/interface/msg/170141184507040415784022802543350382592

Screenshot 2024-10-21 at 9 02 39 AM
Screenshot 2024-10-21 at 9 02 52 AM

davidisaaclee: I'm not sure what form of this API we want to ship as a prototype (e.g. should we ship the structured description payload, should we namespace stuff as prototype, etc); but if we want some kind of stability here, I'd say we should change these fields to allow for plurality, but only handle values as singles:

```ts
defaultPostContentRenderers: { '*': 'tlon.r0.content.chat' },
draftInputs: ['tlon.r0.input.chat'], // or the dictionary approach i mentioned in chat above
```

davidisaaclee:
I created two demonstration PRs with custom components:
- #4090 
- #4091

dnbrwstr (changes requested):
Awesome! Honestly looks great, and I think that this is the right level of development for starting to play with. We won't really know eg. how we want to specify channel inputs until we start trying out more template types. Also some v nice bits of cleanup in here. 

I think code looks great; the one issue that I did find was that posting to a newly created custom channel fails the first time you enter it, but works the second -- possibly related to some of the work around optimistic updates @patosullivan was doing last week.

dnbrwstr:
Oh, and noticed one other thing: currently, when there's a single gallery item on a line, the item fills the full width, doubling in size.

davidisaaclee:
> single gallery item

I think this was an existing bug – this is on Testflight:
image

davidisaaclee:
@dnbrwstr I can't reproduce the failure on first post that you mentioned – attached video is attempting to reproduce it, which appears to succeed. Do you have a more precise repro?

https://github.com/user-attachments/assets/3ce8b9be-82b4-456b-8b97-f441e4c01826

dnbrwstr:
Interesting, here's what I'm seeing -- it's not consistent, probably some kind of race.

https://github.com/user-attachments/assets/c6474915-bf3b-4c43-a38e-8367cb773840

davidisaaclee:
@dnbrwstr thanks, will look into it more

davidisaaclee:
@dnbrwstr I did not get a reliable repro, but I saw the bug happening relatively frequently when using a hosted account, and saw the bug correlating with an message about an unknown custom channel type `custom`. From that, I found that we were passing `channelType = custom` to `assembleNewChannelIdAndName`, which created a "nest"(?) with a `custom/` prefix, which we don't support. I don't get why this was causing a backend error, but applying [d7e49d5](https://github.com/tloncorp/tlon-apps/pull/4062/commits/d7e49d5ec41bee57145193ac3c67f369280b0f75) appears to fix the issue.

patosullivan:
> @dnbrwstr I did not get a reliable repro, but I saw the bug happening relatively frequently when using a hosted account, and saw the bug correlating with an message about an unknown custom channel type `custom`. From that, I found that we were passing `channelType = custom` to `assembleNewChannelIdAndName`, which created a "nest"(?) with a `custom/` prefix, which we don't support. I don't get why this was causing a backend error, but applying [d7e49d5](https://github.com/tloncorp/tlon-apps/pull/4062/commits/d7e49d5ec41bee57145193ac3c67f369280b0f75) appears to fix the issue.

nest is the id for the channel on the backend, and it's hardcoded to three different `kind`s, `%diary` (notebook), `%heap` (gallery), and %chat.

Types for channels live here:
https://github.com/tloncorp/tlon-apps/blob/develop/desk/sur/channels.hoon
https://github.com/tloncorp/tlon-apps/blob/develop/desk/sur/groups.hoon

For the `meta` type that we use throughout:
https://github.com/tloncorp/tlon-apps/blob/develop/desk/sur/meta.hoon

and here for dms/groupdms, but not really relevant for this:
https://github.com/tloncorp/tlon-apps/blob/develop/desk/sur/chat.hoon

Maybe we should add a new `kind` for custom channels? And/or, add new fields to `meta` (or an extension of meta for channels) for PostContentRendererId, DraftInputId, CollectionRendererId? Could still just be arbitrary strings. Seems like a better option than using the description field. cc: @Fang- @arthyn 

In general, though, I really like where this is going, it looks great.

dnbrwstr (approved):
Fix works for me, looking good! We can figure out backend stuff in the next pass.

jamesacklin:
Develop is now clean to merge into. GLHF!

davidisaaclee:
thanks,didn't mean to delay -  I was getting pumpkins image

PR #4064, opened 2024-10-15
notebook/gallery: scroll to bottom of detail view when reply input is focused

patosullivan:
fixes TLON-2907

https://github.com/user-attachments/assets/dfbf3a23-a82f-4530-abef-369a73a8c515


latter-bolden (approved):
Very nice

PR #4066, opened 2024-10-15
gallery: insert old image link posts as image posts in local db

patosullivan:
fixes TLON-2961

At one point, in the old web app, we allowed users to post image urls that would be rendered as image posts on the frontend (in galleries and chat). The new app doesn't allow this, but we need to make sure we can still render old image posts in galleries that were formatted this way.

Rather than changing the content renderer, we can detect posts like this before we insert them into the DB and insert them as if they were normal image posts.

The only drawback to this method, afaik, is that the user would need to log out and log back in for posts they already know about (that fit this criteria) to be rendered correctly.

### DIFF
--- a/packages/shared/src/api/postsApi.ts
+++ b/packages/shared/src/api/postsApi.ts
@@ -3,6 +3,7 @@ import { Poke } from '@urbit/http-api';
 
 import * as db from '../db';
 import { createDevLogger } from '../debug';
+import { IMAGE_URL_REGEX } from '../logic';
 import * as ub from '../urbit';
 import {
   ClubAction,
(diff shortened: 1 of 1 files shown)


packages/shared/src/api/postsApi.ts
```diff
@@ -824,16 +825,12 @@ export function toPostData(
   channelId: string,
   post: ub.Post | ub.Writ | ub.PostDataResponse
 ): db.Post {
-  const getPostType = (
-    channelId: string,
-    post: ub.Post | ub.PostDataResponse
-  ) => {
+  const channelType = channelId.split('/')[0];
```
latter-bolden, line 828: I think we have a helper for this

latter-bolden (approved):
Looks good!

I think the tradeoff of not working with existing cache is workable. We've been talking about dumping local DB on update anyhow.

PR #4067, opened 2024-10-15
delete: only show when appropriate

patosullivan:
Fixes TLON-3076

Only allow posts to be deleted by their authors or the group admin.

### DIFF
--- a/packages/ui/src/components/ChatMessage/ChatMessageActions/MessageActions.tsx
+++ b/packages/ui/src/components/ChatMessage/ChatMessageActions/MessageActions.tsx
@@ -33,6 +33,16 @@ export default function MessageActions({
   const currentUserId = useCurrentUserId();
   const { addAttachment } = useAttachmentContext();
   const channel = useChannelContext();
+  const { data: group } = store.useGroup({ id: post.groupId ?? '' });
+  const currenUserIsAdmin = useMemo(
+    () =>
+      group?.members?.some(
+        (member) =>
+          member?.contactId === currentUserId &&
+          member.roles.some((role) => role.roleId === 'admin')
+      ),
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/ChatMessage/ChatMessageActions/MessageActions.tsx
```diff
@@ -33,6 +33,16 @@ export default function MessageActions({
   const currentUserId = useCurrentUserId();
   const { addAttachment } = useAttachmentContext();
   const channel = useChannelContext();
+  const { data: group } = store.useGroup({ id: post.groupId ?? '' });
+  const currenUserIsAdmin = useMemo(
```
latter-bolden, line 37: We should really pull this out into something reusable. Looks like we do this same thing in 3 other places

latter-bolden (approved):
Makes sense! Up to you if ya wanna refactor the admin check into a helper func.

packages/ui/src/components/ChatMessage/ChatMessageActions/MessageActions.tsx
patosullivan: Agreed, will do that now.

patosullivan: Wait, we *do* have a hook, we just don't use it consistently.

patosullivan: Ok, fixed here and in those other places.

latter-bolden: Awesome, ty!

PR #4068, opened 2024-10-15
notebooks: represent hidden and deleted states for posts

patosullivan:
fixes TLON-2981
fixes TLON-2982

We weren't representing hidden or deleted states on notebook posts.

Edit: no longer rendering deleted posts. Note that the issue with deleted posts previously was that we would render an "Untitled", empty note.


latter-bolden (approved):
Looking good!

PR #4070, opened 2024-10-15
members: add 'groups' table to deps for removeChatMembers query

patosullivan:
fixes TLON-2831

also fixes TLON-2971, same thing.


arthyn (approved):
lgtm

PR #4076, opened 2024-10-16
channels: add ability to leave group channels on mobile

patosullivan:
fixes TLON-2987

We actually didn't have any logic related to leaving a group channel in the mobile app before, and the `leave` action in ChatOptions just assumed we were leaving a dm, so it makes sense that it was failing.


latter-bolden (approved):
Sweeet, looks good!

jamesacklin (approved):
Cool, works as intended.

I'm navigated back to ChatList; would expect that I'm navigated back to GroupChannelsList but I won't block merge over it.

PR #4077, opened 2024-10-16
native: vendor http api

arthyn:
This fixes TLON-3092 by changing how we detect when to shove you to login page. We now only do so if we receive a 403 in a scry or if we receive an `AuthError` bubbled up from the http-api.

PR Checklist

### DIFF
--- a/packages/app/features/top/PostScreen.tsx
+++ b/packages/app/features/top/PostScreen.tsx
@@ -64,7 +64,7 @@ export default function PostScreen(props: Props) {
   }, [post, threadPosts]);
 
   const markRead = useCallback(() => {
-    if (channel && post && threadPosts) {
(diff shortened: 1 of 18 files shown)


latter-bolden (approved):

packages/app/features/top/PostScreen.tsx
```diff
@@ -64,7 +64,7 @@ export default function PostScreen(props: Props) {
   }, [post, threadPosts]);
 
   const markRead = useCallback(() => {
-    if (channel && post && threadPosts) {
+    if (channel && post && threadPosts && threadPosts.length > 0) {
```
jamesacklin, line 67: There we have it!

latter-bolden: indeed! Though we also found repeated valid channel reads occasionally error for unknown reasons

PR #4078, opened 2024-10-16
lure, onboarding: improved metrics

jamesacklin:
Improves our onboarding metrics in PostHog.

On SignUpEmailScreen, if you enter a @tlon.io email address and submit it to go to the next step, we immediately identify you as a Tlon employee (and can discount you from any metrics in PostHog itself).

The remainder of the PR is focused on passing onShareInvite events from ChannelScreen, GroupChannelsScreen, and ChatListScreen all the way down to InviteFriendsToTlonButton, which then waits until the user has finished interacting with the share sheet to tell PostHog to log an "Invite Link Shared" event with the Lure token.

We'll then use these Lure tokens to correlate who is signing up via the "Invite Link Added" onboarding action.

Fixes TLON-3093


latter-bolden (changes requested):
Hold tight, need to do some branch surgery. There's a new method on our testbed branch that makes this kind of thing much easier.

latter-bolden (approved):
@jamesacklin gimme a thumbs up if change looks fine. Changed the base so history looks weird on GH, but the diff is correct if you look at files.

Tested with staging posthog, confirmed the events show up.

jamesacklin:
Great, much cleaner. Thanks!

PR #4082, opened 2024-10-18
group channel list: fix long press on arranged channels

patosullivan:
OTT, we just weren't passing `onLongPress` to ChannelNavSection and then into ChannelListItem within it.
fixes TLON-3071


arthyn (approved):
lgtm

PR #4083, opened 2024-10-18
handle scrolling in view reactions sheet

patosullivan:
OTT, fixes TLON-2732

### DIFF
--- a/apps/tlon-mobile/cosmos.imports.ts
+++ b/apps/tlon-mobile/cosmos.imports.ts
@@ -62,7 +62,7 @@ import * as fixture54 from './src/fixtures/ActionSheet/AddGalleryPostSheet.fixtu
 import * as decorator0 from './src/fixtures/cosmos.decorator';
 
 export const rendererConfig: RendererConfig = {
-  "playgroundUrl": "http://localhost:5001",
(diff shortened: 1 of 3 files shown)


apps/tlon-mobile/cosmos.imports.ts
```diff
@@ -62,7 +62,7 @@ import * as fixture54 from './src/fixtures/ActionSheet/AddGalleryPostSheet.fixtu
 import * as decorator0 from './src/fixtures/cosmos.decorator';
 
 export const rendererConfig: RendererConfig = {
-  "playgroundUrl": "http://localhost:5001",
+  "playgroundUrl": "http://localhost:5000",
```
latter-bolden, line 65: intentional?

latter-bolden (approved):
LGTM!

apps/tlon-mobile/cosmos.imports.ts
patosullivan: happened automatically when running cosmos, afaik 5000 is the default. I'm guessing someone previously ran cosmos while also running another service that was bound to port 5000, then cosmos.imports.ts was updated with that value and it got committed that way.

PR #4085, opened 2024-10-18
native: update signup waiting screen to reflect design

latter-bolden:
OTT

https://github.com/user-attachments/assets/e61b2537-a668-4943-b461-9f97db864bd3


jamesacklin (approved):

e-shee:
Let's stay "unboxing your new cloud computer" or "unboxing your cloud computer" instead of peer in the first point.

PR #4087, opened 2024-10-18
detail views: fix bad layouts for comments

patosullivan:
OTT, fixes TLON-3099

### DIFF
--- a/packages/shared/src/types/PostCollectionConfiguration.ts
+++ b/packages/shared/src/types/PostCollectionConfiguration.ts
@@ -13,13 +13,16 @@ export type PostCollectionLayoutType =
 // If the caller has a non-nullable `channel`, they can then get a
 // non-nullable return value - nice, right?
 export function layoutTypeFromChannel(
-  channel: db.Channel
+  channel: db.Channel,
+  detailView?: boolean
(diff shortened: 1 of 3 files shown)


latter-bolden (approved):

davidisaaclee:
do you know how this worked before?

davidisaaclee:
ah, we were asserting that the `channelType` was `chat` in detail view https://github.com/tloncorp/tlon-apps/blob/1a6d219db1e49daaa0f682f4ae54260582467358/packages/ui/src/components/DetailView.tsx#L57

davidisaaclee (approved):
thanks for fixing this – I would prefer to hard-code `compact-list-bottom-to-top` in `DetailView` instead of putting it in `layoutTypeFromChannel`, since I'm not sure we want to maintain that being dynamic, but I'm fine with this change as a fix (I can make this change as a follow-up if it sounds okay).

to hardcode, I'd think to:
1. add scroller `layoutType` prop
2. move `layoutTypeForChannel` up a level, so we pass a `layoutType` to all `Scroller`s
3. set `layoutType="compact-bottom-to-top"` on the `Scroller` in `DetailView`

PR #4088, opened 2024-10-21
android: fix issue with notifications service

patosullivan:
fixes TLON-3104

It looks like we already have everything we need to receive notifications while the app has been killed, the issue was that our `isInForeground` check was always returning true while the app was closed. Afaik, we want notifications while the app is in the foreground anyway (according to another issue galen filed), so I think we can safely remove this (or at least comment it out for now).

Also added some logs around the TalkNotificationManager for debugging in the future.

Tested on a physical device by running a build, allowing the app to open and register a push token, killing the app, then sending a DM to the ship that was logged-in to the app.


arthyn (approved):
lgtm ty for finding!

PR #4090, opened 2024-10-21
custom channels: add button input

davidisaaclee:


packages/ui/src/contexts/componentsKits.tsx
```diff
@@ -68,6 +69,13 @@ const BUILTIN_DRAFT_INPUTS: { [id: string]: DraftInputRendererComponent } = {
   [DraftInputId.chat]: ChatInput,
   [DraftInputId.gallery]: GalleryInput,
   [DraftInputId.notebook]: NotebookInput,
+  [DraftInputId.yo]: ({ draftInputContext }) => (
+    <ButtonInput
+      draftInputContext={draftInputContext}
+      messageText="Yo"
+      labelText="Yo"
```
davidisaaclee, line 76: In a perfect world, this input configuration would be stored on the channel somehow, preferably in the timeline – but this is not a perfect world.

This could be some new kind of channel state, or it could be implemented with Hoon snippets.

PR #4094, opened 2024-10-21
debug: tools for prod logging

arthyn:
This PR adds the ability for any user in production to start capturing logs in their app. The logs themselves are not persisted but the setting to do logging is. This allows a user to restart the app and catch every log involved.

Once the logs are captured, they can return to the app info screen and hit the 'Upload logs' button which will attempt to open an email app if they have it to send support@tlon.io an email. If they don't, it simply prints the UUID to the page so that they can copy and paste it to wherever they were communicating with us.

PR Checklist


packages/app/features/settings/AppInfoScreen.tsx
```diff
@@ -75,6 +147,34 @@ export function AppInfoScreen(props: Props) {
...
+          {enabled && logs.length > 0 && (
+            <Stack>
+              <Button onPress={onUploadLogs}>
+                <Text>Upload logs ({logs.length})</Text>
+              </Button>
+            </Stack>
+          )}
+          {enabled && logId && !hasClients && (
+            <YStack padding="$l">
+              <Text>Please email support@tlon.io with this log ID:</Text>
```
latter-bolden, line 173: out of scope, but would be cool to dm Tlon Team with the ID

packages/app/features/settings/AppInfoScreen.tsx
```diff
@@ -1,38 +1,110 @@
...
+}
+
 export function AppInfoScreen(props: Props) {
   const { data: appInfo } = store.useAppInfo();
+  const { enabled, logs, logId, uploadLogs } = useDebugStore();
   const easUpdateDisplay = useMemo(() => getEasUpdateDisplay(Updates), []);
+  const [hasClients, setHasClients] = useState(true);
+
+  useEffect(() => {
+    async function checkClients() {
```
latter-bolden, line 58: nit, but might make sense to pull the email handling out into a hook rather than inlined on the screen

packages/app/features/settings/UserBugReportScreen.tsx
```diff
@@ -56,14 +58,11 @@ export function UserBugReportScreen({ navigation }: Props) {
...
-        'User manually submitted a bug report'
-      );
       if (submission.additionalNotes) {
-        reporter.log(`User attached notes:`);
-        reporter.log(submission.additionalNotes);
+        logger.crumb(`User attached notes:`);
+        logger.crumb(submission.additionalNotes);
       }
-      reporter.report(null);
+      logger.trackError('User manually submitted a bug report');
```
latter-bolden, line 65: 🙏🏻

packages/app/lib/debug.ts
```diff
@@ -0,0 +1,23 @@
...
+
+storage
+  .load({
+    key: DEBUG_STORAGE_KEY,
+  })
+  .then((enabled) => {
+    useDebugStore.getState().toggle(enabled);
+  });
+
+export const toggleDebug = async (enabled: boolean) => {
```
latter-bolden, line 15: nit, `setDebug` may be clearer since you're passing the value you want and not flipping curr

latter-bolden (approved):
Looking good!

packages/app/features/settings/AppInfoScreen.tsx
arthyn: yeah that's not a bad idea

packages/app/features/settings/AppInfoScreen.tsx
arthyn: if we start using this more yeah we should

PR #4097, opened 2024-10-22
ios: open channel using shortcut action

davidisaaclee:
Not sure if we want to ship this, but thought I'd throw it up for review and get the conversation going.

Adds an iOS intent + app shortcut to open the app to a specific channel, with the option to start a new draft in that channel (if input supports it).

- Extended `useGroupNavigation` to be able to navigate to a channel by specifying only a channel ID, and to take any other params for the channel route
- Added `startDraft: boolean` channel route param to start a new draft on open if possible
- Added `startDraft?: () => void` to draft input ref handle for inputs to define how to start a draft programmatically (e.g. open modal input, focus text box)
- Added deeplink handler of the form `io.tlon.groups://channel/open?id=0v4.00000.qd4mk.d4htu.er4b8.eao21&startDraft=true` to open to a channel and optionally start a draft
- Registered app intent + shortcut action to open channel by specifying channel ID

| Creating a shortcut | Adding shortcut to homescreen |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/165e4aa1-454f-4f6c-8f8f-0004e13ef23d" /> | <video src="https://github.com/user-attachments/assets/0571ffa3-7776-4a48-8cbe-5f1e11283dc8" /> |


davidisaaclee:
I'm still learning about these APIs – I think ideally we'd register this shortcut via an `AppShortcutProvider`, but Apple is doing some very unique things with the build once you add that, and it doesn't seem to work with our Xcode project for some reason. This code still works, but we don't get Siri integration.

apps/tlon-mobile/ios/Intents/OpenAppIntent.swift
```diff
@@ -0,0 +1,26 @@
+import AppIntents
+
+struct OpenAppIntent: AppIntent {
+  static var title: LocalizedStringResource = "Open channel"
+  static var description = IntentDescription("Opens the app to the specified channel.")
+  static var openAppWhenRun = true
+  
+  @Parameter(title: "Channel ID", description: "The ID of the channel to open the app to.")
+  var channelId: String?
```
davidisaaclee, line 9: there's a much better way to expose channels here: you can define a "query" which could let people search the app's store e.g. by channel title. idk if we actually wanna ship this so I didn't go into that.

that said, replacing this param in the future with a searchable channel entity means people's shortcuts will break. idk how you do things in a backwards compatible way (without just being verbose and exposing all old variables).

davidisaaclee:
@patosullivan is there a better URL to "open channel and optionally start draft" that I could use to unify with web navigatoin?

dnbrwstr:
I think this is cool! Any reason not to merge it? 

Also curious whether this is the same plumbing we'd need for the share-to-app flow (eg share a photo to a channel).

davidisaaclee:
>Any reason not to merge it?

Changing the parameters of the intent (i.e. going from raw channel ID to more friendly search/query-based channel title) will necessarily break people's shortcuts: so if someone is relying on this shortcut, they'd need to rebuild their workflow if we change that param.

> share-to-app flow
I haven't looked at this in a while, but that'd be separate – those are implemented on iOS as extensions (like the notification service extension)

davidisaaclee:
btw, it looks like we can't dynamically enable/disable app intents (e.g. put them behind a feature flag) – app intents are slurped up statically during build

jamesacklin:
I'm not terribly worried about the possibility of breaking a shortcut workflow. We can always make a medium-sized stink about it if we change the intent parameter. I think it's cool and worth having even in the short-term.

davidisaaclee:
@jamesacklin makes sense – I'm doing a quick spike on getting the query-based parameter in since (1) it'd be cool and relatively future-proof, and (2) it opens up more cool stuff we can do once we can read from SQLite in native context

davidisaaclee:
I added minimal code to interact with SQLite from iOS – we should consider pulling in a dependency that makes that easier if we add to this code.

Searching for a channel works now! Not shown in video, but running the workflow after selecting a channel does bring you to the correct channel.

https://github.com/user-attachments/assets/49d9e9c8-ce4d-4da4-aa06-3457dc0e9ccb

pinging @dnbrwstr for re-review

davidisaaclee:
mind dump: I was intrigued by this line in apple's docs:

> Apps implementing either the EnumerableEntityQuery or EntityPropertyQuery protocols automatically add a Find intent in the Shortcuts app.

https://developer.apple.com/documentation/appintents/integrating-custom-types-into-your-intents#Enable-Find-intents

I tried implementing an EntityPropertyQuery, but (1) without an SQL library, I had to build up SQL manually which felt fragile, and (2) we only get a "Find Channels" action, which is cool and possibly helpful for some super users, but not all that useful. The find action lets you define any number of conditions+sort+limit, like "channel title equals ____ and/or group title contains _____; sort by group title Z-A; limit to 10" - then you can theoretically use the results in a subsequent action (e.g. open the top match using the "Open Channel" action).

I adopted the EntityPropertyQuery protocol on the same struct that had EntityStringQuery, and this ended up removing the nice modal search for channels in the "Open Channel" action. It's possible this wouldn't happen if we just split into two structs.

I was hoping this would magically put our stuff into the Spotlight index, but that's a whole different API (under Core Spotlight in docs) - which seems pretty easy and maybe we don't even need to add a new target.

PR #4099, opened 2024-10-22
native: onboarding resiliency

latter-bolden:
Add support for reviving abandoned onboarding sessions if the user is interrupted. Should be resilient to network disconnects and app closes. Also fixes our logic for the case where you've already signed up but haven't been assigned a ship — attempting to login will correctly direct you to the onboarding flow. Was able to simplify `signupContext` quite a bit in the process.

Few other honorable mentions:
- fixes system SMS paste
- adds dev env vars for prefilling onboarding values
- introduces a new pattern for working with keyval store (only used for the new onboarding stuff, will file another issue to move over existing values)

Fixes TLON-3065


apps/tlon-mobile/src/screens/Onboarding/ReserveShipScreen.tsx
```diff
@@ -32,10 +32,11 @@ export const ReserveShipScreen = ({ navigation }: Props) => {
...
 
   useEffect(() => {
-    if (!signupContext.didCompleteSignup) {
-      signupContext.setDidCompleteSignup(true);
+    if (!signupContext.didCompleteOnboarding) {
+      signupContext.setOnboardingValues({ didCompleteOnboarding: true });
     }
-  }, [signupContext]);
+    signupContext.kickOffBootSequence();
+  }, []);
```
arthyn, line 39: Should this be empty?

arthyn (approved):
lgtm just the one question

apps/tlon-mobile/src/screens/Onboarding/ReserveShipScreen.tsx
latter-bolden: Yeah, we only want it to fire once when the screen mounts

PR #4100, opened 2024-10-22
native: hardcode detail view layout

davidisaaclee:
Hardcodes detail view layout instead of adding to `layoutTypeFromChannel`. We may want to revert this in the future to customize detail view layouts, but I would prefer to avoid the complication until we need it – we don't have enough usage yet to know if this is the config "shape" we want yet, and by hardcoding, we can be more agile with the base collection layout type code.

This is a followup from #4087 – [context](https://github.com/tloncorp/tlon-apps/pull/4087#pullrequestreview-2378872029)

Screenshots proving that this still works:
| Notebook detail view | Gallery detail view |
| --- | --- |
| Simulator Screenshot - iPhone 16 Pro - 2024-10-22 at 12 07 01 | Simulator Screenshot - iPhone 16 Pro - 2024-10-22 at 12 06 55 |


patosullivan (approved):
makes sense!

PR #4101, opened 2024-10-22
new web: omnibus build/organization/perf fixes

patosullivan:
- Resolved build failures in web environment caused by:
  - Incorrect import of `react-native-branch` through `useHandleLogout`
  - Direct usage of `nativeDb.resetDb` instead of platform-specific implementation in `useConfigureUrbitClient`
  - Build failure due to top-level await causing `regeneratorRuntime` errors
- Moved `signupContext` to mobile-specific directory as it's only used in mobile flows
- Resolved CPU-intensive performance issue in ChatList:
  - Replaced `useStyle` hook with static style object
  - Now using `getTokenValue` helper to access tamagui's `$l` padding value
  - Prevents FlashList from freezing app due to excessive re-renders


latter-bolden (approved):
Looks good to me!

PR #4105, opened 2024-10-23
channels: update the $seal; update $react

mikolajpp:
Two new fields are introduced to $v-seal (and $seal). seq=@ud is the post sequence number. $mod-at=@da is the timestamp of last modification of the post structure.

$react type now defaults to unicode emoji, while preserving backward-compatibility with its %any variant. This lays the groundwork for future custom reactions.

Fixes TLON-2686


desk/app/channels-server.hoon
```diff
@@ -655,7 +777,7 @@
         =/  post  (get:on-v-posts:c posts.channel now.bowl)
         ?~  post  now.bowl
         $(now.bowl `@da`(add now.bowl ^~((div ~s1 (bex 16)))))
-      =/  new=v-post:c  [[id ~ ~] 0 essay.c-post]
+      =/  new=v-post:c  [[id +(count.channel) id ~ ~] 0 essay.c-post]
       :-  `[%post id %set ~ new]
       (put:on-v-posts:c posts.channel id ~ new)
```
Fang-, line 782: This count gets reflected in the update that gets sent after inserting this new post. The way (as currently written) we then increment the count in channel state is that `+ca-update` looks at the update being sent and extracts the new count. This is a confusing mix of responsibilities.

I would update `+ca-c-post` to produce a modified `channel` (instead of a modified `posts.channel`), and then simply do the modification in-line here.

Shouldn't be too big a change, `+ca-c-post` only has one callsite (in `+ca-c-channel`).

desk/app/channels-server.hoon
```diff
@@ -92,27 +92,149 @@
...
+    =.  count  +(count)
+    :-  count
+    ?~  post  posts
+    ::  insert seq and mod-at into seal
+    ::
+    =/  new-post=v-post:c
+      =/  new-seal=v-seal:c
+        =+  mod-at=(~(got by mod) id-post)
+        =*  seal  -.u.post
+        =+  seal(|1 [+(count) mod-at |1.seal])
```
Fang-, line 152: You already incremented above.

```suggestion
        =+  seal(|1 [count mod-at |1.seal])
```

desk/app/channels-server.hoon
```diff
@@ -92,27 +92,149 @@
...
+    |=  vc=v-channels:v7:old:c
+    ^-  v-channels:c
+    %-  ~(run by vc)
+    |=  v=v-channel:v7:old:c
+    ^-  v-channel:c
+    =/  [log=log:v-channel:c mod=(map id-post:c time)]
+      (log-6-to-7 log.v)
+    =/  [count=@ud =v-posts:c]
+      (v-posts-6-to-7 posts.v mod)
+    =-  -(|1.- [count |1.-.-])
```
Fang-, line 126:
```suggestion
    =-  -(|1.- [count |1.-<])
```

desk/app/channels-server.hoon
```diff
@@ -92,27 +92,149 @@
...
+  ++  v-channels-6-to-7
+    |=  vc=v-channels:v7:old:c
+    ^-  v-channels:c
+    %-  ~(run by vc)
+    |=  v=v-channel:v7:old:c
+    ^-  v-channel:c
+    =/  [log=log:v-channel:c mod=(map id-post:c time)]
+      (log-6-to-7 log.v)
+    =/  [count=@ud =v-posts:c]
+      (v-posts-6-to-7 posts.v mod)
```
Fang-, line 125: @arthyn are we fine with only sequence-number-ing the top-level posts? The purpose here is to make it easier for the client to detect & fetch missing chunks of backlog, so I'm assuming it's Probably Fine™, but maybe there's still concerns around large numbers of replies or something like that.

desk/app/channels-server.hoon
```diff
@@ -92,27 +92,149 @@
...
+    ?~  post.u-post.u-channel
+      :-  [count seq-log]
+      :_  (~(put by mod) id.u-channel time)
+      (put:log-on:c log time %post id.u-channel %set ~)
+    ::  increment .seq only for a new post
+    ::
+    =^  seq=@ud  count
+      ?~  seq=(~(get by seq-log) id.u-channel)
+        =.  count  +(count)
+        [count count]
```
Fang-, line 210:
```suggestion
        [. .]:+(count)
```

(^:

desk/app/channels-server.hoon
```diff
@@ -92,27 +92,149 @@
...
+        (~(put by seq-log) id.u-channel count)
+    :_  (~(put by mod) id.u-channel time)
+    (put:log-on:c log time %post id.u-channel u-post)
+  ::
+  ++  future-6-to-7
+    |=  f=future:v-channel:v7:old:c
+    ^-  future:v-channel:c
+    ::  channel future is defunct for now
+    ::
+    *future:v-channel:c
```
Fang-, line 226: Little bit unprincipled to nuke the data just so you don't have to write the migration for it. Would be good to write the conversion still, even if in the wild the future is always empty.

desk/app/channels-server.hoon
```diff
@@ -767,6 +892,15 @@
...
     =/  =update:c  [time u-channel]
+    ::  increase count for every new post
+    ::
+    =?  count.channel  ?&  ?=(%post -.u-channel)
+                           ?=(%set -.u-post.u-channel)
+                           ?=(^ post.u-post.u-channel)
+                       ==
+      =+  seq.u.post.u-post.u-channel
+      ?>((gth - count.channel) -)
+    ::
```
Fang-, line 903: As per the comment above, this should disappear.

desk/app/channels.hoon
```diff
@@ -157,20 +176,141 @@
...
+      (v-posts-7-to-8 posts.v mod)
+    =-  -(|1.- [count |1.-.-])
+    %=  v
+      posts    v-posts
+      log      log
+      future   (future-7-to-8 future.v)
+    ==
+  ++  v-posts-7-to-8
+    |=  [vp=v-posts:v7:old:c mod=(map id-post:c time)]
+    ^-  [@ud v-posts:c]
```
Fang-, line 201: I would move the bodies of these arms into /lib/channel-utils, instead of one-to-one duplicating them between channels and channels-server. Just guarantees they'll be consistent/equivalent/the same.

desk/app/channels.hoon
```diff
@@ -474,8 +615,28 @@
       (toggle-post toggle.a-channels)
     ?:  ?=(%join -.a-channel.a-channels)
       ca-abet:(ca-join:ca-core [nest group.a-channel]:a-channels)
+    ::  upconvert emoji
+    ::
+    =?  a-channel.a-channels  ?=([%post %add-react *] a-channel.a-channels)
+      %_  a-channel.a-channels
+        q.c-post
+        =*  react  q.c-post.a-channel.a-channels
+        ?>  ?=(@t react)
```
Fang-, line 624: If it's an old action, you wouldn't need this assertion, right?

Also I'd patch up the indentation like so:

```suggestion
      %_  a-channel.a-channels
          q.c-post
        =*  react  q.c-post.a-channel.a-channels
```

Same thing for the block below.

desk/app/channels.hoon
```diff
@@ -474,8 +615,28 @@
...
+        q.c-reply.c-post
+        =*  react  q.c-reply.c-post.a-channel.a-channels
+        ?>  ?=(@t react)
+        ?~  react=(kill:em react)
+          [%any ^react]
+        u.react
+      ==
     ca-abet:(ca-a-channel:(ca-abed:ca-core nest.a-channels) a-channel.a-channels)
-  ::
+    ::
```
Fang-, line 639: This was fine as-is, note how it's "behind" the code body of the cases.
```suggestion
  ::
```

Fang- (commented):
Haven't completed review of /app/channels yet (need to continue at `+watch`), but here's the comments of what we've covered so far.

desk/app/channels.hoon
```diff
@@ -622,14 +782,22 @@
...
-    =.  cor  (give %kick ~[path v0+path v1+path] ~)
+    =.  cor  (give %fact ~[v2+path] cage.sign)
+    ::  send legacy said:v7 response
+    ::
+    =.  cor  %^  give  %fact
+        ~[path v0+path v1+path]
+      ?.  ?=(%channel-said-2 p.cage.sign)
+        cage.sign
+      =+  !<(=said:c q.cage.sign)
+      channel-said+!>((to-said-1:utils said))
```
Fang-, line 796: I would make sure to specify the different cases that we handle here. Optionally, you can consider unifying this with the `?+` below.

```suggestion
    =.  cor
      %^  give  %fact
        ~[path v0+path v1+path]
      ?+  p.cage.sign  ~|(funny-mark+p.cage.sign !!)
        %channel-denied  cage.sign
      ::
          %channel-said-2
        =+  !<(=said:c q.cage.sign)
        channel-said+!>((to-said-1:utils said))
      ==
```

desk/app/channels.hoon
```diff
@@ -730,75 +898,63 @@
...
       `(slav %ud u.since.pole)
     :^  ~  ~  %channel-heads
+    !>  ^-  channel-heads:v7:old:c
+    (murn ~(tap by v-channels) (cury channel-head-1:utils since))
+    ::
+      [%x %v3 %heads since=?(~ [u=@ ~])]
+    =/  since=(unit id-post:c)
+     ?~  since.pole  ~
+      ?^  tim=(slaw %da u.since.pole)  `u.tim
+      `(slav %ud u.since.pole)
```
Fang-, line 954:
```suggestion
    =/  since=(unit id-post:c)
      ?~  since.pole  ~
      ?^  tim=(slaw %da u.since.pole)  `u.tim
      `(slav %ud u.since.pole)
```

desk/app/channels.hoon
```diff
@@ -587,7 +747,7 @@
       [?(%v0 %v1) ~]                        ?>(from-self cor)
       [?(%v0 %v1) %unreads ~]               ?>(from-self cor)
       [?(%v0 %v1) =kind:c ship=@ name=@ ~]  ?>(from-self cor)
-      [?(%v0 %v1) %said =kind:c host=@ name=@ %post time=@ reply=?(~ [@ ~])]
+      [?(%v0 %v1 %v2) %said =kind:c host=@ name=@ %post time=@ reply=?(~ [@ ~])]
```
Fang-, line 750: The `%v2` case will give you a different initial result (from `+ca-said`, which should become `+ca-said-1` and `+ca-said-2` separately) right?

desk/app/channels.hoon
```diff
@@ -1778,18 +1950,22 @@
   ::  handle scries
   ::
   ++  ca-peek
-    |=  [=(pole knot) ver=?(%v0 %v1 %v2)]
+    |=  [=(pole knot) ver=?(%v0 %v1 %v2 %v3)]
     ^-  (unit (unit cage))
     ?+    pole  [~ ~]
+      ::
```
Fang-, line 1956:
```suggestion
    ::
```

Would dedent all of these separators by one step.

desk/app/channels.hoon
```diff
@@ -2014,13 +2225,23 @@
...
-        ?:  ?=(%v2 version)
-          =/  =paged-posts:c
-            [(uv-posts-2:utils posts) newer older (wyt:on-v-posts:c posts)]
-          ``channel-posts-2+!>(paged-posts)
+        ?-  version
+        ::
+          %v1
         =/  =paged-posts:v1:old:c
           [(uv-posts:utils posts) newer older (wyt:on-v-posts:c posts)]
         ``channel-posts+!>(paged-posts)
```
Fang-, line 2233:
```suggestion
        ?-  version
        ::
            %v1
          =/  =paged-posts:v1:old:c
            [(uv-posts:utils posts) newer older (wyt:on-v-posts:c posts)]
          ``channel-posts+!>(paged-posts)
```

And then for the others just indent the tag/pattern to match.

desk/app/chat.hoon
```diff
@@ -1078,27 +1079,40 @@
...
       (put:on-v-replies:d old-replies time `(convert-quip time writ))
     %+  gas:on-v-posts:d  *v-posts:d
-    %+  murn  writs
-    |=  [=time =writ:t]
-    ^-  (unit [id-post:d (unit v-post:d)])
-    ?^  replying.writ  ~
+    =|  posts=(list [id-post:d (unit v-post:d)])
+    =<  +
+    %+  roll  writs
+    |=  [[=time =writ:t] count=@ud =_posts]
```
Fang-, line 1085: Shouldn't define `posts` outside of the scope where it's being used. (It's not the same posts as the one you use below, anyway.) Confusing.

Consider the following, noting that the faces used in the sample type remain accessible after the accumulation product comes out of `+roll`.

```suggestion
    =<  posts
    %+  roll  writs
    |=  $:  [=time =writ:t]
            [count=@ud posts=(list [id-post:d (unit v-post:d)])]
        ==
```

Same thing in diary and heap.

Fang- (commented):
Last couple comments included. Thank you for your service!

desk/app/channels-server.hoon
```diff
@@ -3,7 +3,7 @@
 ::    this is the server-side from which /app/channels gets its data.
 ::
 /-  c=channels, g=groups
-/+  utils=channel-utils, imp=import-aid
+/+  utils=channel-utils, imp=import-aid, em=emojimart
 /+  default-agent, verb, dbug, neg=negotiate
 ::
 %-  %-  agent:neg
```
Fang-, line 9: Since we're changing the protocol, you need to bump the version here, to `~.channels^%2`.

Then in the channels agent, bump the _expected_ number for channels-server agents to match.

desk/app/channels.hoon
```diff
@@ -11,10 +11,15 @@
...
 ::  performance, keep warm
 /+  channel-json
 ::
+::  compile all marks
+::XX this is not actually compiling??
+:: /~  marks  *  /mar/channel
+::
 =/  verbose  |
 %-  %-  agent:neg
     :+  notify=&
```
Fang-, line 25: So, under this, `[%channels-server^[~.channels^%2 ~ ~] ~ ~]`.

(And arguably the self-protocol should just go, we don't use that as intended here.)

Fang- (commented):
Almost forgot: protocol version bumps!

desk/lib/mop-extensions.hoon
```diff
@@ -155,4 +155,20 @@
...
+  ?~  a  +<+.b
+  $(a r.a, +<+.b $(a l.a, +<+.b (b n.a +<+.b)))
+::  +req: reduce to product, right to left
+::
+++  req
+  |=  a=(tree item)
+  |*  b=_=>(~ |=([* *] +<+))
+  |-
+  ?~  a  +<+.b
+  $(a l.a, +<+.b $(a r.a, +<+.b (b n.a +<+.b)))
```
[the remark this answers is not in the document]
mikolajpp: I have removed these, and used `+roll` with `+tap:on` instead.

desk/app/channels-server.hoon
arthyn: yeah I think it's fine for top level only

desk/app/channels.hoon
```diff
@@ -17,14 +17,14 @@
 /+  channel-json
 ::
 ::  compile all marks
-::XX this is not actually compiling??
+::XX  /~ rune is broken
 :: /~  marks  *  /mar/channel
```
Fang-: Might be useful to point to the accompanying issue here, [`urbit/urbit#5242`](https://github.com/urbit/urbit/issues/5242), so that we know when we can sanely/fully enable this again. (Don't think I had linked that to you before, so see also that issue for further deets.)

desk/app/channels.hoon
```diff
@@ -617,26 +506,31 @@
...
     ::
-    =?  a-channel.a-channels  ?=([%post %add-react *] a-channel.a-channels)
-      %_  a-channel.a-channels
-        q.c-post
+    =/  new-a-channels=a-channels:c
+      a-channels
+    ?>  ?=([%channel *] new-a-channels)
+    =?  new-a-channels  ?=([%post %add-react *] a-channel.a-channels)
+      ?>  ?=([%post %add-react *] a-channel.new-a-channels)
+      %_  new-a-channels
```
Fang-: Not sure I understand why you do the `=/  new-a-channels` here. In this new/current code, we never use `a-channels` again, so it's not like we're keeping a copy of the original data around for reference, right?

The code from before your refactor commit, where it just modified the `a-channels` in-place, was fine.

Fang- (commented):
A couple further comments, but most of the changes here look good. Thank you!

desk/app/channels.hoon
mikolajpp: This has actually been rewritten: the current version does not look like that anymore (I am looking at `bb24466`). Can you please take a look again?

Fang- (approved):
Changes look good, thank you! (:

desk/lib/channel-utils.hoon
```diff
@@ -676,4 +1055,159 @@
...
+  ::  number
+  ::
+  =.  count  +(count)
+  :-  count
+  ?~  post  posts
+  ::  insert seq and mod-at into seal
+  ::
+  =/  new-post=v-post:c
+    =/  new-seal=v-seal:c
+      =+  mod-at=(~(got by mod) id-post)
```
arthyn, line 1093: this crashes currently because we don't capture logs in `%channels` agent only `%channels-server`, so when we try to get the time here it just bails and then fails to load. I think for now we can just default to either the post id or the latest reply id. maybe we want to modify `ca-u-channels` to capture and keep the logs, because we'll likely need it for changes endpoints later on.

arthyn (changes requested):
need to adjust `v-posts-7-to-8` before we can merge this. found by trying to commit these changes to my moon which already had channels state.

desk/lib/channel-utils.hoon
mikolajpp: As we talked, we default to post id in the absence of corresponding log event.

arthyn (approved):
We probably should start saving logs so we can do changes endpoint down the road, but we can do that later.

PR #4109, opened 2024-10-23
channels: custom extension, configuration and post types

arthyn:
PR Checklist


packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+  nickname?: string;
+  avatar?: string;
+  origin?: string;
+  agent: string;
+}
+
+interface PostMetadataSchemaV1 {
+  type: string;
+  bot?: Bot;
+  rendererOverride?: string;
```
davidisaaclee, line 384: do we have an immediate use for this? i think it'll definitely be nice to have, but maybe we should defer until we need it

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+  defaultContentRenderers?: ContentRenderer[];
+  defaultPostCollectionRenderer?: string;
+}
+
+type ChannelMetadataSchema = ChannelMetadataSchemaV1;
+
+interface Hook {
+  version: number;
+  name: string;
+  src: string;
```
davidisaaclee, line 410: any point in naming this `hoon` instead of `src` (or otherwise indicating the "type" of this field is a hoon snippet)? (i.e. if we could conceivably accept another kind of `src` in the future)

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+  compiled: boolean;
```
davidisaaclee, line 411: what does this mean?

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+  messageInputs: MessageInput[];
+  defaultContentRenderers?: ContentRenderer[];
+  defaultPostCollectionRenderer?: string;
+}
+
+type ChannelMetadataSchema = ChannelMetadataSchemaV1;
+
+interface Hook {
+  version: number;
+  name: string;
```
davidisaaclee, line 409: nit: it sounds like `name` needs to be unique – as a non-urbit user, `name` feels like "display name" instead of a machine id. `id` or `key` is more communicative of the required uniqueness

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+  version: 1;
+  messageInputs: MessageInput[];
+  defaultContentRenderers?: ContentRenderer[];
+  defaultPostCollectionRenderer?: string;
+}
+
+type ChannelMetadataSchema = ChannelMetadataSchemaV1;
+
+interface Hook {
+  version: number;
```
davidisaaclee, line 408: nit: why use a version number instead of a created timestamp? seems like potentially one less thing to keep track of

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+  name: string;
+  src: string;
+  compiled: boolean;
+}
+
+interface Hooks {
+  allowed: Hook[];
```
davidisaaclee, line 415: nit: `validate` matches the "first person present verb" form better, and avoids ambiguity of "these are the hooks that are allowed"

packages/shared/src/urbit/channel.ts
```diff
@@ -172,14 +172,23 @@ export type KindDataChat = {
   chat: null | { notice: null };
 };
 
-export type KindData = KindDataDiary | KindDataChat | KindDataHeap;
-export type Kind = 'heap' | 'diary' | 'chat';
+export type KindDataCustom = {
+  custom: string;
```
davidisaaclee, line 176: nit: in an extreme of the custom channels world, everything will end up with kind `custom` – aesthetically, I wonder if there's another term that we could use to support that. ("user"?)

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+}
+
+interface ContentRenderer {
+  render: string;
+  using: string;
+}
+
+interface ChannelMetadataSchemaV1 {
+  version: 1;
+  messageInputs: MessageInput[];
```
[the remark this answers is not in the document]
davidisaaclee: I don't have a strong opinion here because I don't know what the "multiple inputs" UX is going to be. I guess the no-duplicates enforcement isn't that important because we can just do a runtime check.

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+  defaultContentRenderers?: ContentRenderer[];
```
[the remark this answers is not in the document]
davidisaaclee: not a strong opinion though

packages/shared/src/urbit/channel.ts
arthyn: I think at that point we'd just change our API so that kind-data goes away and we just have a post-type field or w/e.

packages/shared/src/urbit/channel.ts
arthyn: agreed

packages/shared/src/urbit/channel.ts
arthyn: I would think that you'd want to know which iteration of this particular hook is running

packages/shared/src/urbit/channel.ts
arthyn: yeah this is good feedback will change

packages/shared/src/urbit/channel.ts
arthyn: I was thinking we might need some indicator of whether this source code actually compiles or not so we could surface an error that something is wrong with your code.

packages/shared/src/urbit/channel.ts
arthyn: that's probably going to require other changes to the API anyway so I'd put it off until we get to the point of multiple languages

packages/shared/src/urbit/channel.ts
arthyn: I don't I just remember us talking about it, we can remove

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+  agent: string;
+}
+
+interface PostMetadataSchemaV1 {
+  type: string;
+  bot?: Bot;
+  rendererOverride?: string;
+}
+
+interface MessageInput {
```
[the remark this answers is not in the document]
arthyn: yep

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+interface ContentRenderer {
+  render: string;
+  using: string;
+}
+
+interface ChannelMetadataSchemaV1 {
+  version: 1;
+  messageInputs: MessageInput[];
+  defaultContentRenderers?: ContentRenderer[];
+  defaultPostCollectionRenderer?: string;
```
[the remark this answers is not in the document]
arthyn: yeah I don't think we'll have multiple

packages/shared/src/urbit/channel.ts
davidisaaclee: seems useful - i read this as "is this code precompiled" - maybe "valid" would be more intuitive?

packages/shared/src/urbit/channel.ts
```diff
@@ -357,6 +371,53 @@ export interface Perm {
...
+interface PostMetadataSchemaV1 {
+  type: string;
+  bot?: Bot;
+  rendererOverride?: string;
+}
+
+interface MessageInput {
+  type: string;
+  postType: string;
+  configuration: Record<string, string>;
```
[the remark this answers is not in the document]
arthyn: right yeah, what would the hoon snippets do here?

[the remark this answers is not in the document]
davidisaaclee: example: button input, snippet calls out to an LLM API, which gives back a fortune cookie response, which is used as the post body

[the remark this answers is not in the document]
davidisaaclee: another one: audio message, hoon applies a user-defined audio effect over the wavetable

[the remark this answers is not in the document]
davidisaaclee: (tbc, i think this means client-side hoon execution to have a decent UX...)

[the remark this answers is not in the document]
davidisaaclee: ftr, I changed my mind on this – I still think that Hoon snippets would be great here, but user configuration through parameters seems pretty important.

(Is there any chance that JSON values look enough like Hoon expressions? like, if we accept values like `3` and `'blue'`, could we change in the future to say "those were all actually simple Hoon expressions that didn't have any operators")

arthyn:
Closing in favor of tloncorp/tlon-apps#4165 and tloncorp/tlon-apps#4161

PR #4110, opened 2024-10-23
ops: update Expo OTA script to account for build environment

latter-bolden:
Updates the mobile update script from the old repo to account for needed environment variables. Note: whenever we add new ones, we'll need to do so both here and on Expo.

Also switches posthog to switch projects for preview vs prod.

Tested by using the action to trigger a preview app OTA.


arthyn (approved):
lgtm!

PR #4112, opened 2024-10-24
native: dynamic collection components

davidisaaclee:
We already have the API functionality to set a channel's collection layout at creation time, but we didn't have an ergonomic way to add new collection layouts to the app. This PR pulls out a component responsible purely for laying out posts in a collection, so that we can quickly add new collection layouts.

[f27d449](https://github.com/tloncorp/tlon-apps/pull/4112/commits/f27d449ae478862ecaaf289b375ac3c081e223ce) adds an example new layout, which displays one post at a time; tapping on the left goes back one post, tapping on the right goes forward. (Loading more posts is not implemented but possible.)


davidisaaclee:
closing since this is in the megabranch #4113

PR #4113, opened 2024-10-24
custom channels megabranch

davidisaaclee:
> Big branch that touches a lot – I think the best way to verify this is to just merge it and keep an eye out for bugs. A lot of this has already been "tested" by release to custom channels tests. I've merged `develop` and checked viewing and drafting in each of the three legacy channel types, which all seem good (save for issues 3440-3442, which exist in current black app anyways).
 
Some missable changes in this branch (I'd like to make fixes in follow-up PRs wherever possible, but mentioning in case you see something blocking):
- Added new (verbosely-named) route `PostUsingContentConfiguration` https://github.com/tloncorp/tlon-apps/blob/38ad895bbeba3e365be47f1265283af12185711e/packages/app/navigation/types.ts#L66-L69 – this is similar to the `Post` route, except that it uses [`DetailPostUsingContentConfiguration`](https://github.com/tloncorp/tlon-apps/blob/38ad895bbeba3e365be47f1265283af12185711e/packages/ui/src/components/DetailPostUsingContentConfiguration/DetailPostUsingContentConfiguration.tsx) which respects the post/channel's content configuration, instead of using channel-type-based switching. (`PostScreenView` had a lot of ad-hoc switching based on channel type, e.g. whether to show an input. I feel like there were other good reasons...) I added this to enable forum threads.
- In order to access feature flags in packages "below" `app`, I added a hacky "feature flag mirroring" context that can be accessed anywhere above `shared`: see diff here https://github.com/tloncorp/tlon-apps/pull/4113/files#diff-6db28889c6a4847ffe10e62d7244a48d5ff482bd93decf3a4b400bc75888bf00 – consensus in Tlon chat was that we should move the base feature flag code to `shared`, but I'd prefer to do that in a follow-up.
- Added `shared/utils` subpackage, which contains utilities with no other Tlon dependencies https://github.com/tloncorp/tlon-apps/blob/38ad895bbeba3e365be47f1265283af12185711e/packages/shared/src/utils/index.ts

This branch puts custom channels data in a JSON blob in channel `description` (to be moved to use API in https://github.com/tloncorp/tlon-apps/tree/release-channels-updates), and flags custom channel editing (the "configurator") behind the `customChannelCreation` flag.

---

Fixes TLON-3276


davidisaaclee:
@.reviewers – I'd like to merge this for tomorrow's build if possible

jamesacklin:
I think given team constraints we may want to wait until next week—@arthyn is out, @dnbrwstr is out, and reviewing would take @latter-bolden the better part of a day.

I'm 100% OK with frankenbranching the Preview app today with this branch + develop, is that acceptable in the 2-3 day term?

davidisaaclee:
@jamesacklin makes sense, didn't know about outages

packages/ui/src/contexts/postCollection.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export const PostCollectionContext =
+  createContext<PostCollectionContextValue | null>(null);
+
+export function usePostCollectionContext() {
+  return useContext(PostCollectionContext);
+}
+
+export function usePostCollectionContextUnsafelyUnwrapped() {
```
davidisaaclee, line 42: just get rid of `UnsafelyUnwrapped`, and change `usePostCollectionContext` to unwrap and throw an error when null

packages/ui/src/contexts/componentsKits.tsx
```diff
@@ -63,11 +113,58 @@ const BUILTIN_CONTENT_RENDERERS: { [id: string]: RenderItemType } = {
...
+  [CollectionRendererId.notebook]: ListPostCollection,
+  [CollectionRendererId.carousel]: CarouselPostCollection,
+  [CollectionRendererId.cards]: CardsPostCollection,
+  [CollectionRendererId.sign]: SingleCardPostCollection,
+  [CollectionRendererId.boardroom]: BoardroomPostCollectionView,
+  [CollectionRendererId.strobe]: StrobePostCollectionView,
+  [CollectionRendererId.summaries]: PostSummaryCollectionView,
 };
 
 export function ComponentsKitContextProvider({
```
davidisaaclee, line 170: it'd be nice to remove the need for an explicit provider and just put this as a default context

packages/shared/src/utils/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './object';
```
davidisaaclee, line 1: add comment forbidding imports from other packages

packages/ui/src/components/ManageChannels/CreateChannelSheet.tsx
```diff
@@ -144,78 +167,38 @@ export function CreateChannelSheet({
...
+      title: displayName,
+      value: id,
+    })
+  ),
+};
 
-const CustomChannelConfigurationForm = forwardRef<{
-  getFormValue: () => ChannelContentConfiguration;
-}>(function CustomChannelConfigurationForm(_props, ref) {
+const CustomChannelConfigurationForm = forwardRef<
```
davidisaaclee, line 187: remove custom channel stuff from this sheet

davidisaaclee: 617b02977c039a168c4c28ac7e215fc7c553638b

packages/shared/src/utils/index.ts
davidisaaclee: 240893b87f7c6ff4b62fd23fc3153f816b9f5924

packages/ui/src/contexts/componentsKits.tsx
davidisaaclee: 84d6a0a1c712876b838ca1d4e240d65f871cdc72

packages/ui/src/contexts/postCollection.ts
davidisaaclee: 207889fb7b9429889de72cd463813f0e297b9397

dnbrwstr (approved):
Looks good to me!

PR #4114, opened 2024-10-24
analytics: track invite added on welcome screen

latter-bolden:
Right now we only fire the _Invite Link Added_ event if they add it on the paste invite screen. If branch properly captured the invite after install OR you clicked the link while still on the welcome screen, we won't hear about it. This just updates that event to also get sent from the welcome screen.


jamesacklin (approved):
Correct

PR #4115, opened 2024-10-24
onboarding: add show/hide to password fields

jamesacklin:
Does what it says on the tin! This was reported a few times by users; I personally find it annoying when testing.

image


arthyn (approved):
beautiful soul

PR #4118, opened 2024-10-25
groups-ui: suggested contacts and pal import

arthyn:
This grabs both our pals and our DMs and uses those to make contacts. For pals we take all `targets` which are people we've asked to be pals. For DMs, we take anyone you've DM'd in the last 30 days that has more than 10 messages. This will only run once upon receiving the update.

Fixes TLON-3146

PR Checklist

### DIFF
--- a/desk/app/groups-ui.hoon
+++ b/desk/app/groups-ui.hoon
@@ -8,6 +8,7 @@
   +$  card  card:agent:gall
   +$  current-state
     $:  %2
+        hidden-contact-suggestions=(set ship)
         pins=(list whom:u)
         first-load=?
     ==
@@ -88,8 +89,6 @@
       =+  !<(old=versioned-state vase)
       =?  old  ?=(~ old)     *current-state
       =?  old  ?=(%0 -.old)  (state-0-to-1 old)
-      =?  cor  ?=(%1 -.old)
-        (emit %pass /preload-contacts %agent [our dap]:bowl %poke noun+!>(%preload-contacts))
       =?  old  ?=(%1 -.old)  (state-1-to-2 old)
       ?>  ?=(%2 -.old)
       =.  state  old
@@ -104,7 +103,7 @@
     ==
   ::
   ++  state-1-to-2
-    |=(state-1 [%2 pins first-load])
+    |=(state-1 [%2 ~ pins first-load])
   +$  state-0  [%0 first-load=?]
   ++  state-0-to-1
     |=(state-0 [%1 ~ first-load])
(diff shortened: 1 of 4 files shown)


desk/app/groups-ui.hoon
```diff
@@ -238,4 +254,28 @@
   ?+  wire  !!
     [%build ~]  cor
   ==
+++  preload-contacts
+  =+  .^(chat-running=? (scry %gu %chat /$))
+  =?  cor  chat-running
+    =+  .^  [dms=(map ship dm:c) *]
+        (scry %gx %chat /full/noun)
+      ==
```
mikolajpp, line 262: Usually you would align `==` with the opening rune, but perhaps here it is ok? @Fang-

mikolajpp, line 262: Usually you would align `==` with the opening rune, but perhaps here it is ok? @Fang-

desk/app/groups-ui.hoon
```diff
@@ -238,4 +254,28 @@
...
+    %-  emil
+    %+  murn
+      ~(tap by dms)
```
mikolajpp, line 265:
```suggestion
    %-  emil  %+  murn  ~(tap by dms)
```
feels more readable to graps at a glance this way. also deferring to @Fang-

desk/app/groups-ui.hoon
```diff
@@ -238,4 +254,28 @@
...
+    =/  count  (wyt:on:writs:c wit.pact.dm)
+    =/  cutoff  (sub now.bowl ~d30)
+    ?.  &((gth count 10) (gth -.u.latest cutoff))  ~
+    `[%pass /contact %agent [our.bowl %contacts] %poke contact-action-1+!>([%page ship ~])]
+  =+  .^(pals-running=? (scry %gu %pals /$))
+  =?  cor  pals-running
+    =+  .^(targets=(set ship) (scry %gx %pals /targets/noun))
+    %-  emil
+    %+  turn
+      ~(tap in targets)
```
mikolajpp, line 277: Same nitpick as commented above.

mikolajpp (changes requested):
Looks good, some styling comments.

Do note that this migration will generate poke crashes when attempting to add someone as an existing contact. It seems it should not be a problem, since we haven't exposed the "add as contact" functionality through UI yet.

desk/app/groups-ui.hoon
arthyn:
```suggestion
    =+  .^  [dms=(map ship dm:c) *]
      (scry %gx %chat /full/noun)
    ==
```

Fang-: @arthyn the `==` is closing the `.^`, so aligning with the opening rune here would look like such:

```suggestion
    =+  .^  [dms=(map ship dm:c) *]
          (scry %gx %chat /full/noun)
        ==
```

Fang- (approved):
Style nits aside, this seems fine. Heuristics here should be easy to tweak too, if we ever change our minds about them.

arthyn:
Based on our discussions, this PR is changing so that instead of automatically preloading contacts, we offer a scry instead so the UI can display who we think might be someone you want to add. We also offer pokes to hide(or show) suggestions and to import your pals wholesale as contacts.

arthyn:
@latter-bolden lmk if changes above seem good

latter-bolden (approved):
Approach sounds good to me

mikolajpp (commented):
Looks good to go.

PR #4119, opened 2024-10-27
native: dismiss image viewer on swipe down

davidisaaclee:
Trick here is to only enable the gesture when at min zoom:
- We can't reliably observe zoom level from `react-native-image-zoom` without forking that package
- Instead, manually track `isAtMinZoom` in a state hook, setting it whenever the user pinches, zooms by double-tap, or resets zoom by double-tap
- Disable gesture when `!isAtMinZoom`

Tested that:
- Swipe down dismisses when at min zoom
- Swipe down does not dismiss after zooming in by pinching
- Swipe down does not dismiss after zooming in by double-tapping
- Swipe down dismisses after returning to min zoom after either of those zoom methods

https://github.com/user-attachments/assets/1ea69275-4a5a-4238-8741-230054d82a03


patosullivan (approved):
Looks good!

PR #4120, opened 2024-10-27
native: import directly from shared src instead of bundle

davidisaaclee:
- `pnpm run dev:shared` is no longer necessary
- While developing app, changing code in `shared` will now trigger a (working) reload of the app – no more `unexpected undefined`
- Importing from `@tloncorp/shared/dist/*` is no longer allowed: imports of those subpackages must now be (1) declared in the `exports` in `shared/package.json`, and (2) imported using the alias from `exports`, e.g. `@tloncorp/shared/db`.

All packages typecheck and build, but some tests are failing – e.g. `packages/shared/src/db/queries.test.ts`.
- These failures manifest as a cryptic Vitest Rollup error: `Error: Expected ';', '}' or <eof>`.
- I ended up debugging by searching for the offending lines, which didn't show any sort of pattern, except that they were imports from a file in the same package.
- Based on a hunch, I ran a dependency cycle checker: Screenshot 2024-10-27 at 12 42 55 PM
and those are only the cycles reachable from that file. [gist](https://gist.github.com/davidisaaclee/34d1f88f50afc7e91864a41262c83194)
- At least one of these cycles (`packages/shared/src/urbit/channel.ts -> packages/shared/src/urbit/content.ts`) requires large careful code movement to fix (or maybe just merge the two files? I think I don't have context on the organization here)

I don't have a great understanding of why this is newly a problem: my guess is that we were kind of "dynamically importing" the bundled code when running tests previously (i.e. each file in the bundle had a dynamic `require` to another file in the bundle): since we're now doing "static imports" and need to resolve dependencies up front (to make an esm bundle?), Vitest's Rollup is blowing up at the cycles.

---

I think the main change in DX around skipping bundling `shared` was good enough here to open this PR; I'm going to look for workarounds on the dependency cycle issue; but we should really try to remove these cycles, as I think they are preventing more efficient bundling and potentially causing ghostly bugs.


davidisaaclee:
- Running `dpdm --transform` (iiuc, transform TS first, then look for cycles, then report based on sourcemap) makes the cycle list much smaller (just two easily-fixed entries); however, applying this patch (https://github.com/tloncorp/tlon-apps/commit/5aba3bffe2c9b1cf8c5d3c32f7dfccd6f4ded02b) to this branch does not fix the test failures
- I don't know why `--transform` is only reporting this small number of cycles: tracing manually, one can see cycles like `api/postsApi.ts` -> `db` -> `db/modelBuilders.ts` -> `api/postsApi.ts`

This will probably require a lot of code movement to fix, but I'm trying to minimize the movement to get this PR in. Going forward, I think we should enforce the following dependency layers using `@typescript-eslint/no-restricted-imports` rules (which I think will also require us to migrate our eslint config to a modern format):

```
(higher directories can import from lower directories, but not the other way around)

store
api
db
urbit
logic
```

lmk if that looks right, I obviously have the least experience with this code

dnbrwstr:
Wow I've wanted this forever, tysm so for finally setting it up. I poked at the test failures a bit, and found that it was actually the import of the migration sql file that was failing. I was able to fix the failures by adding `assetsInclude: ['**/*.sql']` to the `vitest.config.ts` in the shared package. 

That said, the tests for web are still failing for the same reason. We can fix by applying the same thing to the web `vite.config.ts`, or by imports around, since web should be able to load what it needs without accidentally including the migrations.

davidisaaclee:
@dnbrwstr what a relief, i can try applying that patch to web after the meeting. curious how you found it!

dnbrwstr:
Literally just commenting out imports until the bundle step stopped failing lmao. Traced it to db/index, then the migrations file, then the sql.

davidisaaclee:
@dnbrwstr `pnpm -r test run` and `pnpm run build:all` both succeed with latest changes – any other checks you think I should do?

dnbrwstr:
Nothing I can think of! Looks g2g to me.

dnbrwstr (approved):

patosullivan:
looks like this broke the web-new build

PR #4122, opened 2024-10-28
add build:new-web to root package scripts

patosullivan:
should help us avoid situations where changes to the mobile app or other packages break the build of the web-new app.


davidisaaclee:
lgtm (double-checked that yes, [we are running `build:all` in ci](https://github.com/tloncorp/tlon-apps/blob/c0ed29c058e4e4b2269af0538bf4d4972fbc0c0f/.github/workflows/ci.yml#L22-L23))

davidisaaclee (approved):

PR #4123, opened 2024-10-28
web-new: bundle *.sql imports as `export default {filecontent}`

davidisaaclee:
This fixes the build breakage introduced by #4120.

The issue was that `import './MIGRATION_FILE.sql'` in `shared` was not being interpreted correctly by tlon-web-new's Vite setup:
- Without any change to `vite.config`, it was being interpreted as like, JS i guess? and raising syntax errors about SQL code not being valid JS code
- Adding .sql files to `assetsInclude` so that these are treated as assets by Vite, we get the sql file import resolving as a URL (on dev server, this is a weird URL with a path that includes `@fs` and an absolute local path on my computer's filesystem)

Adding a `?raw` suffix (`import mig from './file.sql?raw'`) fixed the issue, but broke the non-Vite builds (i.e. `tlon-mobile`).

I tried the function-based [`build.assetsLimitInline`](https://v3.vitejs.dev/config/build-options.html#build-assetsinlinelimit) to override and inline the .sql assets, but that didn't work.

Landed on writing a small Vite plugin to transform `.sql` files into their file content + `export default` so we can use it like a regular ES module.

Tested by `pnpm run --filter tlon-web-new dev` (migrations worked),  `pnpm run --filter tlon-web dev` (doesn't have any errors in browser), and  `pnpm run --filter tlon-mobile ios` (runs and runs migrations correctly)

### DIFF
--- a/apps/tlon-web-new/vite.config.mts
+++ b/apps/tlon-web-new/vite.config.mts
@@ -3,6 +3,7 @@ import { tamaguiPlugin } from '@tamagui/vite-plugin';
 import { urbitPlugin } from '@urbit/vite-plugin-urbit';
 import basicSsl from '@vitejs/plugin-basic-ssl';
 import react from '@vitejs/plugin-react';
+import fs from 'fs';
 import analyze from 'rollup-plugin-analyzer';
 import { visualizer } from 'rollup-plugin-visualizer';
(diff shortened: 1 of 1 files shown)


patosullivan (approved):
Looks good, thank you!

PR #4124, opened 2024-10-29
native: catch storage key not found when initializing debug

latter-bolden:
Just adds a catch statement to prevent this warning I've been getting when the sim starts:


patosullivan (approved):

arthyn (approved):
sorry, ty!

PR #4125, opened 2024-10-29
web-new: add sidebar nav for desktop

patosullivan:
fixes TLON-3143

Edit: also includes some fixes around height/padding for the BareChatInput on web.

This just adds basic sidebar navigation for desktop. Right now, we just reuse/recreate the same components currently used on mobile. It is responsive, the sidebars will disappear and it will render the mobile version of the app if you adjust the screen width to < 768.

I needed a patch for @react-navigation/drawer because the Drawer module uses useAnimatedStyle without providing a dep array (a no-no for web use).

I didn't use ChannelListScreen and ChatListScreen for ChannelListDrawerContent and ChatListDrawerContent because the props differ and I assume we'll actually want to change these components pretty significantly very soon (for instance, switching away from the sheets).

I pulled @react-navigation/drawer into the mobile app package.json as well because we may render these sidebars on iPad.

Looks like this:

https://github.com/user-attachments/assets/95beb088-44b5-4c2d-bc8d-f1d0a1d63888

Copilot summary:

This pull request includes several changes aimed at enhancing navigation and improving the user interface for both mobile and web platforms. The most significant updates involve adding a drawer navigation component, conditionally rendering components based on device type, and improving the handling of group and channel data.

### Navigation Enhancements:
* Added `@react-navigation/drawer` to `package.json` files for both mobile and web platforms to support drawer navigation. (`apps/tlon-mobile/package.json`, `apps/tlon-web-new/package.json`, `package.json`) [[1]](diffhunk://#diff-8deea517552ee81bbb5b9943b366a4b19324af6c3ba8d60ee96aee46c8cc7fcbR56) [[2]](diffhunk://#diff-bbb45f66c81ab8935a1089904a9bf4614f52b4c0611f849b46f4c4bdc262f2d8R51) [[3]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L55-R56)
* Implemented `TopLevelDrawer` for desktop navigation, which includes a permanent drawer with navigation icons for Home, Activity, and Profile screens. (`packages/app/navigation/desktop/TopLevelDrawer.tsx`)
* Created `HomeNavigator` to manage the drawer content and handle navigation between groups and channels. (`packages/app/navigation/desktop/HomeNavigator.tsx`)

### Conditional Rendering Based on Device Type:
* Modified `RootStack` and `AppRoutes` to conditionally render components based on whether the device is mobile or not. (`apps/tlon-web-new/src/app.tsx`, `packages/app/navigation/RootStack.tsx`) [[1]](diffhunk://#diff-04a4100ed3a6fd5cebc024cfc89fb8331418edddcf3d3eae6aac61aec32a75f4R14) [[2]](diffhunk://#diff-04a4100ed3a6fd5cebc024cfc89fb8331418edddcf3d3eae6aac61aec32a75f4L29-R30) [[3]](diffhunk://#diff-04a4100ed3a6fd5cebc024cfc89fb8331418edddcf3d3eae6aac61aec32a75f4R89) [[4]](diffhunk://#diff-04a4100ed3a6fd5cebc024cfc89fb8331418edddcf3d3eae6aac61aec32a75f4L101-R103) [[5]](diffhunk://#diff-2d017028089577e606a614cca1c2addfe51b6fa580c8a6d22c6ffa229e0e5055L3-R3) [[6]](diffhunk://#diff-2d017028089577e606a614cca1c2addfe51b6fa580c8a6d22c6ffa229e0e5055R45-R55)
* Introduced `useIsMobile` hook to determine the device type and adjust the UI accordingly. (`packages/ui/src/components/Channel/ChannelHeader.tsx`) [[1]](diffhunk://#diff-0bf9dd95a7f33e9fe18c2fefc035c218082876e5e6e448bff4a17e06bbc092a1R11) [[2]](diffhunk://#diff-0bf9dd95a7f33e9fe18c2fefc035c218082876e5e6e448bff4a17e06bbc092a1R104)

### Group and Channel Data Handling:
* Added `ChannelListDrawerContent` and `ChatListDrawerContent` components to manage and display group and channel data within the drawer navigation. (`packages/app/navigation/desktop/ChannelListDrawerContent.tsx`, `packages/app/navigation/desktop/ChatListDrawerContent.tsx`) [[1]](diffhunk://#diff-1ddcd4bb25018286596435ead7d50c18a8eeb3d0cb6eed3ac70199677259d676R1-R66) [[2]](diffhunk://#diff-df6ace81f252b77704393a8a402c46f2964a23e6b00bba20b71b61b8ac5cbef8R1-R236)

These changes collectively enhance the navigation experience and ensure the application adapts to different device types, providing a more seamless and intuitive user interface.


patosullivan:
Fyi that weird space in the video when the ChatList is filtered to Messages was caused by the inclusion of the group invite items. Removing them fixed that (we were also showing these on native).

patches/@react-navigation__drawer@6.7.2.patch
```diff
@@ -0,0 +1,441 @@
+diff --git a/lib/module/views/modern/Drawer.js b/lib/module/views/modern/Drawer.js
+index 3aa1baa2efd65bab22da56416c90e3a294dab266..08fe2866924f233fa4bd349258a4b35ff51c7fc7 100644
+--- a/lib/module/views/modern/Drawer.js
++++ b/lib/module/views/modern/Drawer.js
+@@ -1,7 +1,38 @@
+-function _extends() { _extends = Object.assign ? Object.assign.bind() : function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; }; return _extends.apply(this, arguments); }
```
davidisaaclee, line 6: it looks like this patch got prettier'd, making the diff larger than it needed to be. if we disabled prettier for this, (1) it'd be easier to extend in the future, and (2) it'd be easier to merge upstream updates into this patch without conflict.

i can't tell the extent of the patch – is it small enough that redoing without prettier would be quick?

davidisaaclee:
> I needed a patch for @react-navigation/drawer because the Drawer module uses useAnimatedStyle without providing a dep array (a no-no for web use).

I was curious if we could fix this in a way that we wouldn't need to patch any other third-party code that has a similar issue in the future: I saw https://docs.swmansion.com/react-native-reanimated/docs/guides/web-support/#web-without-the-babel-plugin that says that Reanimated's Babel plugin should fix this issue – did you give that a try?

patosullivan:
> > I needed a patch for @react-navigation/drawer because the Drawer module uses useAnimatedStyle without providing a dep array (a no-no for web use).
> 
> I was curious if we could fix this in a way that we wouldn't need to patch any other third-party code that has a similar issue in the future: I saw https://docs.swmansion.com/react-native-reanimated/docs/guides/web-support/#web-without-the-babel-plugin that says that Reanimated's Babel plugin should fix this issue – did you give that a try?
> 
> I think this diff should do it (not tested; i'm applying this on a different commit so I don't think you can `git apply`):
> this is super cool btw, love sidebar

yep, we've been using it for a while, it's in the vite config, but it's not getting applied to dependencies like @react-navigation/drawer.

patches/@react-navigation__drawer@6.7.2.patch
patosullivan: good catch! yeah, it's a tiny patch, will fix.

davidisaaclee:
oh, sad, makes sense though!

latter-bolden (approved):
Ran it locally and was able to navigate around as expected. 

Got it to lock up once, but couldn't repro. I'm guessing it was a routing issue. Diagnosing that kind of thing will be easier once the URL bar is reflective of navigation state.

UX wise, I think this is good to merge as an initial step.

patches/@react-navigation__drawer@6.7.2.patch
```diff
@@ -1,441 +1,22 @@
 diff --git a/lib/module/views/modern/Drawer.js b/lib/module/views/modern/Drawer.js
-index 3aa1baa2efd65bab22da56416c90e3a294dab266..08fe2866924f233fa4bd349258a4b35ff51c7fc7 100644
+index 3aa1baa2efd65bab22da56416c90e3a294dab266..8737778293408588026d6efab2c9070ed1def4f4 100644
```
davidisaaclee, line 2: beautiful :')

packages/ui/src/hooks/useIsMobile.ts
```diff
@@ -0,0 +1,6 @@
+import { useWindowDimensions } from 'tamagui'
+
+export default function useIsMobile() {
```
davidisaaclee, line 3: nit: can we change this to something like `useIsWindowNarrow`? not all mobile clients match this, and narrow "desktop" clients (with cursor, etc) will (and should) match here – so I'm worried that this function could be misused in the future and require cleanup.

patosullivan: good point, fixed!

PR #4126, opened 2024-10-29
input: show ref previews and allow attachments in chat threads

patosullivan:
Looks like we didn't pass these props with the old input either.

### DIFF
--- a/packages/ui/src/components/PostScreenView.tsx
+++ b/packages/ui/src/components/PostScreenView.tsx
@@ -223,6 +223,8 @@ export function PostScreenView({
                   editPost={editPost}
                   channelType="chat"
                   getDraft={getDraft}
+                  showAttachmentButton={channel.type === 'chat'}
+                  showInlineAttachments={channel.type === 'chat'}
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):
Good find

packages/ui/src/components/PostScreenView.tsx
```diff
@@ -223,6 +223,8 @@ export function PostScreenView({
                   editPost={editPost}
                   channelType="chat"
                   getDraft={getDraft}
+                  showAttachmentButton={channel.type === 'chat'}
+                  showInlineAttachments={channel.type === 'chat'}
```
davidisaaclee, line 227: this isn't worth fixing (might be doing a breaking change to this type soon), but leaving as a note in case someone wants it: this will not fix custom channels (I think you'd want to check `channel.contentConfiguration != null ? channel.contentConfiguration.defaultPostContentRenderer === 'tlon.r0.content.chat'` : channel.type === 'chat'`)

davidisaaclee (approved):

PR #4129, opened 2024-10-30
Centralize components configuration

davidisaaclee:
I started work on component parameterization, then realized that we'd be growing the volume of code necessary to configure a new component.

This code moves almost all config into `channelContentConfig.ts` (everything except the mapping to a React component, since that wouldn't make sense in the dependency graph). It changes the `ContentRendererId` etc enums into a backwards-compatible type + dictionary – this was so that we can centralize a richer configuration (with label, eventually parameter schema, etc) and derive the "enum" off of that.


dnbrwstr (approved):
Nice, lgtm!

PR #4130, opened 2024-10-31
scroller: adds the beloved scroll-to-bottom button

jamesacklin:
Fixes TLON-2956

Conveniently located in the lower-right corner of chat channels (or, I should say, "compact bottom-to-top message lists" 😉). Appears if you aren't at the bottom of the channel or there are unreads.

Thanks for doing the hard work, I just had to un-comment it.

https://github.com/user-attachments/assets/f10f2252-c9e1-4d76-86d0-d05245e68177

### DIFF
--- a/packages/ui/src/components/Channel/Scroller.tsx
+++ b/packages/ui/src/components/Channel/Scroller.tsx
@@ -39,6 +39,8 @@ import { useLivePost } from '../../contexts/requests';
 import { useScrollDirectionTracker } from '../../contexts/scroll';
 import { ChatMessageActions } from '../ChatMessage/ChatMessageActions/Component';
 import { ViewReactionsSheet } from '../ChatMessage/ViewReactionsSheet';
+import { FloatingActionButton } from '../FloatingActionButton';
+import { Icon } from '../Icon';
 import { Modal } from '../Modal';
 import { ChannelDivider } from './ChannelDivider';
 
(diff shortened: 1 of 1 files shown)


dnbrwstr:
Have you tested in a channel with bunch of unreads? I'd be curious if this still works when content isn't already loaded in.

jamesacklin:
It does, but it also doesn't disappear when you press it, since we're relying on "has unreads" and that doesn't get cleared until you exit the channel.

https://github.com/user-attachments/assets/966a580d-f731-46f5-9bc3-6aecd0307762

latter-bolden:
> it also doesn't disappear when you press it, since we're relying on "has unreads" and that doesn't get cleared until you exit the channel

maybe the cheapest hack is just have a state variable that hides it after click? Can use at most once anytime you enter a channel

jamesacklin:
@latter-bolden and @dnbrwstr, can you take a look now? I'm juggling a bunch of cases but this mostly works. It hides when pressed then reappears when you're paginating through lots and lots of unreads (99+) but works great for un-paginated unreads.


https://github.com/user-attachments/assets/5d705dab-692d-4d3b-9513-1f5119a2f176

jamesacklin:
@latter-bolden and @dnbrwstr, feel free to take another look. I am now passing in the channel's loading state to show a spinner in the button itself. I also changed the conditional to show button for scroll if either:

- We haven't pressed it yet
- We're loading
- There are newer posts to load

This way, the button will stay visible with the spinner until we've loaded all newer posts and reached the bottom of the channel.

I think it feels pretty good but may have missed an edge case or two.

latter-bolden (approved):
Let me know how that latest commit is feeling. If okay on your end, I'm game to try merging for the build today.

jamesacklin:
> Let me know how that latest commit is feeling. If okay on your end, I'm game to try merging for the build today.

Works fine for me. I want to try to get that flash-of-white out of there, otherwise good to merge.

https://github.com/user-attachments/assets/8311817b-20db-4f58-8da0-64e28db6c25d

PR #4131, opened 2024-10-31
web-new: fix build issue, clean up vite config

patosullivan:
fixes TLON-3159 by patching @react-navigation/drawer *again*. The DrawerView module was using inline requires (within an esm module). Attempted to fix by using the `build.commonjsOptions.transformMixedEsModules` option in the vite config, but that led to other issues (somehow we ended up importing commonjs modules instead of the es modules). 

Fortunately we're not using the legacy implementation at all, so we can just hardcode these values here.

I also moved the RN related aliases over to reactNativeWebPlugin just to clean things up a bit in the vite config.


latter-bolden (approved):
Makes sense, nice workaround

PR #4133, opened 2024-10-31
custom channels: add parameters to component configurations

davidisaaclee:
> I'm not super happy with how this PR turned out, it feels like way more code than it should be. I think this is because I didn't use a strict pattern when implementing inputs vs collections vs post content config – these all ended up looking almost identical, but we reference them in slightly different ways, so I needed to add adapters whenever trying to operate over all of them.
> But in the interest of having a demo, here's the diff:

- Changed component configs in `ChannelContentConfiguration` (`draftInput` / `defaultCollectionRendererId` / `defaultPostContentRendererId`) from type `ID` -> `ID | { id: ID, configuration?: Record<string, JSONValue> }` – this was to avoid breaking existing channels (which I think would be okay to do, but it wasn't _that_ hard to maintain backwards compatibility)
- Added a `parametersSchema` field to each component spec, which the `ChannelConfigurationBar` uses to populate a sheet with available customizable parameters.
- Provide any available parameters to the post / collection / input components as a loosely-typed `Record<string, JSONValue>`
- Added some example uses of parameters, including a small set of standard parameters available on all collections (I wrote this as opt-in per collection, but enabled on all existing collections)

| New ChannelConfigurationBar | Tapping on one of those cogs |
| --- | --- |
| Simulator Screenshot - iPhone 15 - 2024-10-31 at 09 48 41 | Simulator Screenshot - iPhone 15 - 2024-10-31 at 09 48 12 |

### DIFF
--- a/apps/tlon-mobile/src/fixtures/ChannelConfigurationBar.fixture.tsx
+++ b/apps/tlon-mobile/src/fixtures/ChannelConfigurationBar.fixture.tsx
@@ -8,15 +8,16 @@ import {
 } from '@tloncorp/shared';
 import { UnconnectedChannelConfigurationBar as ChannelConfigurationBar } from '@tloncorp/ui/src/components/ManageChannels/CreateChannelSheet';
 import { useState } from 'react';
-import { View } from 'tamagui';
+import { SafeAreaView } from 'react-native-safe-area-context';
+import { TextArea, View } from 'tamagui';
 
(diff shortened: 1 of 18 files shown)


dnbrwstr (approved):
This lgtm! I imagine we'll want to do a cleanup pass once we have a basic capability set we're happy with. Kind of unavoidable the have some awkward transitional stages when we're moving this incrementally.

PR #4134, opened 2024-10-31
desktop: make action sheet adaptive, display a modal on desktop

patosullivan:
fixes TLON-3148

Also cleaned up some names in ChatListDrawerContent.

I created a new ActionSheetMainContent component because we want a different rule here for padding than we do in the ListItemMainContent component, I think.


dnbrwstr (approved):
Nice, lgtm!

PR #4135, opened 2024-10-31
BareChatInput: typography fixes

jamesacklin:
- Retains focus on the input after the user has sent a message; un-focuses if the user taps outside the input or otherwise blurs the field.
- Removes the TextInput's line-height styling by default, which eliminates (A) a noticeable discrepancy between the placeholder text's vertical position and the inputted text, and (B) the need for corrective padding measures.
- Wraps the TextInput in a View, which we measure with each layout change. If the measured height exceeds the initial height set via props (44px), we can safely assume that the user has entered enough to have 2+ lines of text.
- Applies a line-height to inputted text for 2+ lines of text that matches messages in the scroller.
- Wraps the output of renderTextWithMentions as a child of TextInput, which eliminates the need for the absolute-positioning trick we were employing to render styled mentions (thanks @dnbrwstr: https://docs.expo.dev/guides/editing-richtext/#building-on-top-of-react-native-textinput)

Note that due to an underlying bug in how react-native renders line-height _and_ the fact that `<Text>` and `<TextInput>` render text differently, we do end up with a noticeable additional bottom margin when the user has inputted 4+ lines of text. I think we can tolerate this for the time being, but it's up to us to decide if the line-height matching is worth the trade-off. 

Demo video:

https://github.com/user-attachments/assets/94e80652-c63c-45c2-8e8e-afea668be34f


dnbrwstr (approved):
Nice, lgtm!

PR #4136, opened 2024-10-31
desktop: show updated button for service worker in nav

patosullivan:
fixes TLON-3144

also removes a bunch of old deps that were carried over from the old web app


dnbrwstr (approved):
Nice, lgtm!

PR #4137, opened 2024-10-31
web routing / paths

dnbrwstr:
Initial implementation of web paths/routing. There may be some rough edges, but overall it seems to work. 

A couple notes:
- To get fully correct desktop linking behavior (open in new tab, accessible attributes, etc.) we'll need to use react-navigation's `Link` component. That's not yet implemented here.
- The actual mapping of urls to paths is controlled by the config in `linking.ts`
- I had to wrap all web navigators in the new `BasePathNavigator` in order to get correct behavior when using a non-root path.


latter-bolden:
Code makes sense. Going to try running it locally on mobile and web quick

packages/app/navigation/linking.ts
```diff
@@ -0,0 +1,141 @@
...
+              },
+            },
+          },
+        },
+      },
+    },
+  },
+});
+
+function parsePathParams(...keys: string[]) {
```
latter-bolden, line 131: Can we account for `/` characters here so they don't get URL encoded?

packages/app/navigation/linking.ts
```diff
@@ -0,0 +1,141 @@
...
+                  },
+                  ChannelMeta: {
+                    path: 'group/:groupId/channel/:channelId/meta',
+                    parse: parsePathParams('channelId'),
+                    exact: true,
+                  },
+                  GroupSettings: {},
+                  ChannelSearch: {},
+                  Post: {
+                    path: 'post/:postId',
```
latter-bolden, line 118: Still trying to figure out what's happening here, but I seem to occasionally get a `:postId` of`undefined` and then that path component gets stuck in my URL bricking any other channels I try to navigate to

latter-bolden (commented):
Super cool to see! Code looks good to me. Swapping to the id + DB lookup routing on mobile feels mostly seamless.

Two notes from desktop testing, but I'll leave it up to you on whether those are blocking.

packages/app/navigation/linking.ts
dnbrwstr: I thought about this -- basically we'd choose a less ugly character that can never be part of a path to replace it with. I was thinking about a `+`? I didn't implement because I wasn't sure what the valid character set for flags actually is -- do you?

packages/app/navigation/linking.ts
patosullivan: yep, I see this when starting/opening a thread, URL path looks like this:
`https://localhost:3000/apps/tm-alpha/group/~rabpur-datpel-finned-palmer%2Ftesting-group-admin/channel/chat%2F~rabpur-datpel-finned-palmer%2Fwelcome-7767/post/undefined?post=%5Bobject%20Object%5D`

dnbrwstr: Yeah, I see this too, fix incoming.

patosullivan: looks like navigate to post still passes the whole post as a param:
```
  const navigateToPost = useCallback(
    (post: db.Post) => {
      navigation.push('Post', { post });
    },
    [navigation]
  );
```

packages/app/navigation/linking.ts
latter-bolden: Will leaving the slash intact mess up how it parses the route?

Flag is always `${ship}/${term}`, should be URL safe (https://docs.urbit.org/glossary/term)

dnbrwstr: Yes, it unfortunately will. We could probably hack around, but I'd rather keep it simple for now -- gonna try with pluses.

packages/app/navigation/linking.ts
dnbrwstr: Should be fixed

packages/app/navigation/linking.ts
dnbrwstr: Went a little farther down this road, but found some funky bugs with customizing that behavior -- I think we should punt for now as it's mostly cosmetic.

latter-bolden: Got it, thanks for digging in. I'm okay punting

patosullivan (approved):
Looks good! If the url encoding bothers us enough we can always revisit that later. Imho it's not a deal breaker.

PR #4140, opened 2024-11-01
custom channels: fix handling of omitted standard parameters

davidisaaclee:
Without this PR, older custom channels have authors and replies hidden by default.

- Set param defaults when parsing `StructuredChannelDescriptionPayload`
- In same code block, add any missing fields to `ChannelContentConfiguration` (missing fields here was causing a crash in one of my test channels)
- Hide `Configure view` menu item when config is missing (this would cause a crash otherwise)
- Apply standard params (`Show replies`, `Show authors`) on legacy post types (we were not doing this before)

### DIFF
--- a/packages/shared/src/api/channelContentConfig.ts
+++ b/packages/shared/src/api/channelContentConfig.ts
@@ -14,12 +14,12 @@ export interface ComponentSpec<EnumTag extends string = string> {
 
 function standardCollectionParameters(): Record<string, ParameterSpec> {
   return {
-    showAuthor: {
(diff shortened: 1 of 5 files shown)


davidisaaclee:
i'm using the code from [dd808dd](https://github.com/tloncorp/tlon-apps/pull/4140/commits/dd808dd2987cc3eead84818b35e5cb283782ceaa) to add defaults to positive-language params, will reopen soon

latter-bolden (approved):
Makes sense!

PR #4144, opened 2024-11-01
native: manually close database instead of using deinit

davidisaaclee:
Apparently `~Copyable` protocol can't be used with generics (at least in the Swift version in our CD).

I had used `~Copyable` so that we could implement a `deinit` to close the DB on GC – without being able to handle the case that opening the DB results in an error, I had to remove `~Copyable`, replacing the `deinit` with a manually-called `close()` method.


davidisaaclee:
I don't know how to resolve this against develop, moving to #4145

PR #4151, opened 2024-11-05
Fix threads on web

dnbrwstr:
We weren't using the react query client from the shared package, so invalidations weren't being applied correctly. Fixing that fixes thread loading as well as some of the initial load issues we were seeing.

Fixes TLON-3187


patosullivan (approved):
Looks good!

PR #4153, opened 2024-11-06
profile: pass channelId when navigating to DMs

jamesacklin:
Clicking the "Message" button on a user profile would crash the app with a "Cannot read property 'split' of undefined" error. This occurred because we were passing the full channel object to navigation instead of just the channel ID that the Channel screen expects.

- Adds validation to ensure channel ID exists before navigation.
- Also adds error handling around DM channel creation and navigation

https://github.com/user-attachments/assets/8a3a153b-8780-4e7e-a593-7c1c50191266

Fixes TLON-3197


patosullivan (approved):
Looks good!

PR #4155, opened 2024-11-07
desk: migrate Tlon infrastructure from %landscape

mikolajpp:
The following agents are proprietary, and thus belong in the %groups desk: %bait, %bark, %contacts, %genuine, %growl, %reel and %settings.

This PR must be coordinated with [#tloncorp/landscape/296](https://github.com/tloncorp/landscape/pull/296)


latter-bolden:
Going to close this for now since we're blocked on kelvin release without a clear timeline. Will make sure we address when the time comes.

PR #4157, opened 2024-11-07
tests: various /lib/test-agent improvements

Fang-:
Most importantly, failing to initialize/set the agent core prior to running your tests could've resulted in false successes. Here we make sure that there's always a crashing skeleton core there, and that crashes with empty traces still result in failing tests.

Additionally, we add a `+do-as` helper for temporarily changing the `src.bowl`. Very convenient for tests that simulate multiple parties interacting with the same agent.

### DIFF
--- a/desk/lib/test-agent.hoon
+++ b/desk/lib/test-agent.hoon
@@ -69,7 +69,12 @@
   |=  f=form:m
   ^-  tang
   =/  res  (f %*(. *state agent skeleton))
-  ?-(-.res %& ~, %| p.res)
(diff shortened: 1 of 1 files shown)


desk/lib/test-agent.hoon
```diff
@@ -327,6 +347,17 @@
   |=  s=state
   &+[~ s(scry f)]
 ::
+++  do-as  ::  temporary src.bowl
```
arthyn, line 350: 🙏

arthyn (approved):
lgtm

PR #4159, opened 2024-11-08
native: show custom channels flag only if user is tlon employee

davidisaaclee:
- Hardcode custom channels flag to show only if user is a Tlon employee
- Change custom channels flag key to invalidate for all existing installs (to avoid case where flag was enabled by non-Tlon-employee, disappears for them from this PR, but remains logically/invisibly enabled)
- [extra] I had a hard-to-see bug in my first pass where I did something like
```ts
const isTlonEmployee = useIsTlonEmployee();
if (isTlonEmployee) {
  // do something
}
```
This would always run the if-block because `isTlonEmployee` is actually a react-query query result – i.e. a non-null object, which is truthy. 6d928f46d1327baf0fd2f0342338f1651cd9b1dc changes this to just return an optional boolean.
- [extra] [8666ec8](https://github.com/tloncorp/tlon-apps/pull/4159/commits/8666ec84d5d290986d87d6ac7980c12dae1139d3) Setting `isTlonEmployee` would get skipped if PostHog creds were not set – always set that

I tested by:
1. Load `develop` client on a non-Tlon-employee account
2. Enable custom channels flag
3. Checkout the head of this branch
4. Check that flag is not visible in flags screen
5. Check that channel creation sheet does not show the "custom" option

### DIFF
--- a/apps/tlon-mobile/src/hooks/useNotificationListener.ts
+++ b/apps/tlon-mobile/src/hooks/useNotificationListener.ts
@@ -76,7 +76,7 @@ function payloadFromNotification(
 
 export default function useNotificationListener() {
   const navigation = useNavigation<NavigationProp<RootStackParamList>>();
-  const { data: isTlonEmployee } = store.useIsTlonEmployee();
+  const isTlonEmployee = store.useIsTlonEmployee();
   const [channelSwitcherEnabled] = useFeatureFlag('channelSwitcher');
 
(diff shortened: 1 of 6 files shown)


jamesacklin (approved):
Slick, thanks for renovating that hook. I wrote that in 0.2s to be able to filter us out of the PostHog event stream; thanks for making it load-bearing.

PR #4163, opened 2024-11-08
native: fix user info on profile screen

latter-bolden:
User row's width got collapsed as part of the lonpress/pressable refactor. Just resets that so it displays normally.

That's the only regression I'm seeing, but trying to check for other similar cases


jamesacklin (approved):
Approve

patosullivan (approved):

PR #4164, opened 2024-11-08
pressable: fix onPress issue in ContactRow

patosullivan:
Fixes the issue where pressing a ContactRow does nothing.

The problem was that we were passing `{...rest}` into ListItem, which included `pressStyle`, which ended up making ListItem the pressable element rather than Pressable.

To fix, I pulled `pressStyle` out of props and passed it to Pressable instead, along with some style related props.


latter-bolden (approved):
Good catch

PR #4165, opened 2024-11-12
channels: custom hooks

arthyn:
PR Checklist


Fang- (commented):
Partial comments. We should continue at some point (from `$result` and onward).

desk/app/channels-server.hoon
```diff
@@ -276,7 +292,18 @@
...
+  |=  [[=id:h schedules=(map origin:h cron:h)] cr=_cor]
+  %+  roll  ~(tap by schedules)
+  |=  [[=origin:h cron:h] co=_cr]
+  ?~  hook=(~(get by hooks.hooks) id)  co
+  ?~  compiled.u.hook  co
+  ?^  delay=(~(get by delayed.hooks) delay-id)  co
+  ::  only start timers for crons that haven't already been started
+  =/  fires-at  (add now.bowl schedule)
+  =/  =wire  (hook-cron-wire id origin)
+  (schedule-hook wire origin delay-id id !>(~) fires-at)
```
Fang-, line 306: "Kill this. Just don't write bugs. It's simple."

Fang- (commented):
More comments from out out-of-band discussion. Let's continue again later! (To look at agent implementation.)

mikolajpp:
A comment about the folder structure: normally mark folders do not user plural names. Thus we have `/mar/contact`, `/mar/channel` etc. We should stick with this convention for hooks as well. 

For the same reason it should be `/ted/hook` not `/ted/hooks`, since when these are invoked from dojo plurals do not make sense in the context of the command (we are adding a single hook, not many)

desk/app/channels-server.hoon
```diff
@@ -836,4 +997,283 @@
...
+  ::
+  ++  ho-abed
+    |=  i=id:h
+    ho-core(id i, hook (~(got by hooks.hooks) i))
+  ::
+  ++  ho-add
+    |=  [name=@t src=@t]
+    ^+  ho-core
+    ~&  "adding hook {<name>}"
+    =.  id  (rsh [3 48] eny.bowl)
```
Fang-, line 1048: 128 bits is a lot (arguably too much for a local identifier), but doesn't _strictly_ guarantee uniqueness. You may want to loop inside the `=.` here until `!(~(has by hooks.hooks) id)`.

And, super stupid nitpick, but if you want consistent/maximum size here regardless of what happens to be put into `eny.bowl`, you want `+end` in place of `+rsh`.

```suggestion
    =.  id
      =+  i=(end 7 eny.bowl)
      |-(?:((~(has by hooks.hooks) i) $(i +(i)) i)
```

desk/app/channels-server.hoon
```diff
@@ -836,4 +997,283 @@
...
+  ++  ho-edit
+    |=  [name=(unit @t) src=(unit @t) meta=(unit data:m)]
+    =?  src.hook  ?=(^ src)  u.src
+    =?  name.hook  ?=(^ name)  u.name
+    =?  meta.hook  ?=(^ meta)  u.meta
+    =/  result=(each vase tang)
+      (compile:utils src.hook)
+    =.  compiled.hook
+      ?:  ?=(%| -.result)  ~
+      `p.result
```
Fang-, line 1072: Wondering if this should follow more clay-like behavior instead, where edits that fail to compile don't actually apply.

desk/app/channels-server.hoon
```diff
@@ -836,4 +997,283 @@
...
+    =/  config  ?@(origin ~ (~(gut by config.hook) origin ~))
+    =/  args  [[%wake +.u.delay] "delayed hook" origin config]
+    ho-abet:(ho-run-single:(ho-abed:ho-core hook.u.delay) args)
+  ::
+      [%cron id=@ kind=?(%chat %diary %heap) ship=@ name=@ ~]
+    =/  =id:h  (slav %uv id.pole)
+    =/  =origin:h  [kind.pole (slav %p ship.pole) name.pole]
+    ::  if unscheduled, ignore
+    ?~  crons=(~(get by crons.hooks) id)  cor
+    ?~  cron=(~(get by u.crons) origin)  cor
```
Fang-, line 1197: Also ignore if it's a "misfire", right?

```suggestion
    ::  if unscheduled, ignore
    ?~  crons=(~(get by crons.hooks) id)  cor
    ?~  cron=(~(get by u.crons) origin)  cor
    ?:  (lth now.bowl fires-at.u.cron)  cor
```

Fang-: Likewise for the `%delayed` case.

Fang- (commented):
More comments below.

desk/lib/channel-utils.hoon
```diff
@@ -684,22 +684,23 @@
...
-  ~&  "{(trip name.hook)} hook run:"
-  ~&  outcome
+  %-  (slog (crip "running hook: {<name.hook>} {<id.hook>}") ~)
+  %-  ?~  channel.context.args  same
+      (slog (crip "on channel: {<nest.u.channel.context.args>}") ~)
+  ?~  compiled.hook  ~
+  =/  gate  [p.u.compiled.hook .*(q:subject q.u.compiled.hook)]
+  =+  !<(=outcome:h (slam gate !>(args)))
+  %-  (slog (crip "{(trip name.hook)} {<id.hook>} hook run:") ~)
+  %-  (slog (crip "{<outcome>}") ~)
```
Fang-, line 703: Not sure I understand/see why the way hooks get compiled/run changed here.

Also, instead of `(slog (crip "{<outcome>}") ~)` consider just `(slog >outcome< ~)`. (:

Fang- (commented):
Changes look mostly good! One more question, and a small suggestion.

desk/app/channels-server.hoon
```diff
@@ -842,4 +1004,287 @@
...
+    ^+  ho-core
+    =.  config.hook  (~(put by config.hook) nest config)
+    =.  cor  (give-hook-response [%config id nest config])
+    ho-core
+  ++  ho-wait
+    |=  [=origin:h schedule=$@(@dr schedule:h) =config:h]
+    ^+  ho-core
+    =/  schedule
+      ?:  ?=(@ schedule)  [now.bowl schedule]
+      schedule
```
mikolajpp, line 1120: Look like this should use a `=?`

desk/app/channels-server.hoon
```diff
@@ -842,4 +1004,287 @@
...
+      (~(put by crons) origin cron)
+    =.  cor  (schedule-cron origin cron)
+    =.  cor  (give-hook-response [%wait id origin schedule config])
+    ho-core
+  ++  ho-rest
+    |=  =origin:h
+    ^+  ho-core
+    =/  crons  (~(got by crons.hooks) id)
+    =/  cron  (~(got by crons) origin)
+    =.  crons.hooks
```
mikolajpp, line 1134: The usage of `crons` and `cron` here and elsewhere make this quite confusing. Normally getting an item out of a collection should make it singular. One alternative would be to replace `crons` with `crontab`. Then what is now `crons.hooks` would become `crontabs.hooks`, and an item inside that would be `crontab`, which is a collection of `cron` items. 

Another way to do that (and avoid the not-so-nice _crontab_) would be to employ the names `job` (`cron-job`), `cron`, `crons`.

desk/app/channels-server.hoon
```diff
@@ -842,4 +1004,287 @@
...
+    =.  hook  hook(state new-state.u.return)
+    =.  cor  (run-hook-effects effects.u.return origin)
+    ho-core
+  --
+++  run-hooks
+  |=  [=event:h =nest:c default=cord]
+  ^-  [(each event:h tang) _cor]
+  =;  [result=(each event:h tang) effects=(list effect:h)]
+    [result (run-hook-effects effects nest)]
+  =/  current-event  event
```
mikolajpp, line 1161: Any reason not to put `current-event` directly in the sample?

desk/app/channels-server.hoon
```diff
@@ -842,4 +1004,287 @@
...
+    (run-hook:utils [current-event ctx] hook)
+  ?~  return  next
+  =*  result  result.u.return
+  =.  effects  (weld effects effects.u.return)
+  =.  hooks.hooks  (~(put by hooks.hooks) i.order hook(state new-state.u.return))
+  ?:  ?=(%denied -.result)
+    [|+~[(fall msg.result default)] effects]
+  =.  current-event  new.result
+  next
+++  wakeup-hook
```
mikolajpp, line 1182: I think `+wake-hook` would be better. Alternatively,`+wake-up-hook`.

desk/app/channels-server.hoon
```diff
@@ -842,4 +1004,287 @@
...
+  |=  =(pole knot)
+  ^+  cor
+  ?+  pole  ~|(bad-arvo-take/pole !!)
```
mikolajpp, line 1185: There is an unwritten convention to use `bad-arvo-take+pole` in such cases. In general I think we prefer `+` for a single use, `/` with a double tag use.

desk/app/channels-server.hoon
```diff
@@ -842,4 +1004,287 @@
...
+      [%delayed id=@ ~]
+    =/  =id:h  (slav %uv id.pole)
+    ?~  delay=(~(get by delayed.hooks) id)  cor
+    ::  make sure we clean up
+    =.  delayed.hooks  (~(del by delayed.hooks) id)
+    ::  ignore premature fires
+    ?:  (lth now.bowl fires-at.u.delay)  cor
+    =*  origin  origin.u.delay
+    =/  hook  (~(got by hooks.hooks) hook.u.delay)
```
mikolajpp, line 1194: Another double plural, more understandable than with `cron`, but maybe we could find a new name for either.

desk/app/channels-server.hoon
```diff
@@ -842,4 +1004,287 @@
...
+    =/  =origin:h  [kind.pole (slav %p ship.pole) name.pole]
+    ::  if unscheduled, ignore
+    ?~  crons=(~(get by crons.hooks) id)  cor
+    ?~  cron=(~(get by u.crons) origin)  cor
+    ::  ignore premature fires
+    ?:  (lth now.bowl next.schedule.u.cron)  cor
+    =.  next.schedule.u.cron
+      ::  we don't want to run the cron for every iteration it would
+      ::  have run 'offline', so we check here to make sure that the
+      ::  next fire time is in the future
```
mikolajpp, line 1210: A breathing `::` below would be good.

desk/app/channels-server.hoon
```diff
@@ -842,4 +1004,287 @@
...
+    =.  cor
+      (schedule-cron origin u.cron)
+    =/  args  [[%cron ~] "cron job" origin config.u.cron]
+    ho-abet:(ho-run-single:(ho-abed:ho-core hook.u.cron) args)
+  ==
+++  schedule-cron
+  |=  [=origin:h =cron:h]
+  ^+  cor
+  =/  wire
+    %+  welp  /hooks/cron/(scot %uv hook.cron)
```
mikolajpp, line 1227: Suggestion: `id` or `id-hook` in $cron

mikolajpp (commented):
Substantial effort! Everything looks good. There a few nitpicks, and a whole bunch of naming suggestions.

arthyn:
@mikolajpp going to mark resolved here as I do these, but will post a PR with the actual changes later.

PR #4169, opened 2024-11-12
native: don't leave group when leaving channel

latter-bolden:
OTT, looks like we were just mistakenly calling both group and channel leave here.

Fixes TLON-3222


jamesacklin (commented):
Does this still leave (multi-)DMs and groups? I think we conflate the use with these sheets, which is where the problem crops up.

latter-bolden:
> Does this still leave (multi-)DMs and groups?

Yeah, they're separate actions in the sheet for leave group and leave channel (both are chat/group agnostic)

jamesacklin (approved):
Approved in that case!

PR #4170, opened 2024-11-12
native: fix unban action

latter-bolden:
OTT

May have been resolved with the pressable fixes, because it looks like unbans are going through fine. Weren't reflecting changes to the banned users table in the groups query however, so this just fixes the optimistic UI update.

Fixes TLON-3221


jamesacklin (approved):
LGTM

PR #4172, opened 2024-11-12
native: fix clearing avatar from profile

latter-bolden:
OTT

Just need to be careful and intentional about handling undefined vs empty assignment here.

Fixes TLON-3217

### DIFF
--- a/packages/ui/src/components/EditProfileScreenView.tsx
+++ b/packages/ui/src/components/EditProfileScreenView.tsx
@@ -46,14 +46,27 @@ export function EditProfileScreenView(props: Props) {
     defaultValues: {
       nickname: userContact?.nickname ?? '',
       bio: userContact?.bio ?? '',
-      avatarImage: userContact?.avatarImage ?? undefined,
+      avatarImage: userContact?.avatarImage ?? '',
     },
   });
 
   const handlePressDone = useCallback(() => {
     if (isDirty) {
       handleSubmit((formData) => {
-        props.onSaveProfile(formData);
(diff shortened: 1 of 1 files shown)


patosullivan (approved):

dnbrwstr (commented):
Could you leave a comment about what's happening here? Not clear on why we need to do this.

packages/ui/src/components/EditProfileScreenView.tsx
```diff
@@ -46,14 +46,27 @@ export function EditProfileScreenView(props: Props) {
...
+        const resolvedValues = {
+          nickname: formData.nickname
+            ? formData.nickname
+            : userContact?.nickname
+              ? ''
+              : undefined,
+          bio: formData.bio ? formData.bio : userContact?.bio ? '' : undefined,
+          avatarImage: formData.avatarImage
+            ? formData.avatarImage
+            : userContact?.avatarImage
```
dnbrwstr, line 65: Is this ternary correct? Why do we pass empty string if the contact had an avatar image and undefined if not?

latter-bolden: When calling out to the backend, we try to avoid setting keys you haven't modified. This means there's a difference between `avatarImage = undefined` and `avatarImage = ''` (latter explicitly clears the field, former is a noop).

Added a comment to make intention clearer. Worth noting this is fixed deeper down on my contacts feature branch so this codepath should be short lived.

dnbrwstr (approved):
Lgtm!

PR #4174, opened 2024-11-13
channel, Scroller: don't show scroll-to-bottom button if sending

jamesacklin:
In channel/index: 
- Sets an isSending flag on message send
- Awaits messageSender in draftInputContext to complete
- Then flips the isSending flag again to false 

In channel/scroller: 
- Consumes the isSending flag reflected in scroller and further downstream in shouldShowScrollButton()

Fixes TLON-3205 by hiding the scroll-to-bottom button if the user is sending a message.


latter-bolden:
@jamesacklin check out #4176

jamesacklin:
> @jamesacklin check out #4176

I like that much better. Thanks!

PR #4179, opened 2024-11-13
nav: unset pressStyle on navicons

patosullivan:
afaict there is no difference in height here, the height is just more noticeable because we were using the default pressStyle of the Pressable component. This PR just removes the pressStyle.

fixes tlon-3202


jamesacklin (approved):
Thanks, that was more the symptom I was noticing.

PR #4180, opened 2024-11-13
native: avoid calling branch from client

latter-bolden:
For creating invite links we phone out to a new external service that talks to Branch itself. For looking up metadata associated with an invite link, we hit the provider instead of Branch.

For now, I still have it calling Branch on lookup since the provider is lacking some useful metadata (namely inviter nickname and avatar). But we fallback gracefully if that request fails, it's purely an enhancement. Going forward, I also updated it to add richer metadata to the provider (shoutout @arthyn for pointing me in the right direction there).

Also proxies our analytics to increase the odds they're able to make it into PostHog.

Note, you'll need to update your `.env` with the invite service vars:
```
# Invite service
INVITE_SERVICE_ENDPOINT="https://serverless-infra.vercel.app/api/inviteLink"
INVITE_SERVICE_IS_DEV="true"
```

Fixes TLON-3138


jamesacklin:
We love you Brian

jamesacklin:
This works in production so...

latter-bolden:
Yeah reviewers are busy, I'm gonna merge it. 

At this point, any issues we find with the code will need to be promptly addressed and deployed anyhow. If any of y'all notice anything weird just hmu.

PR #4181, opened 2024-11-13
fix load freeze

dnbrwstr:
Text streaming requests perform much worse than normal ones in react native, which was causing freezes during loading of init data. This data modifies the urbit client so that only the `eventSource` request uses text streaming mode.

Fixes TLON-3201.


arthyn (approved):
lgtm!

PR #4183, opened 2024-11-13
lure backend: keep local enabled set in sync with what we hear remotely

latter-bolden:
There was a bug with how we were tracking Lure enabled status for remote groups in `%grouper`. Before allowing a Lure to be created, the client app performs a check to see if Lure's are enabled for that group. If you're not the group host, the backend would query that from the host and pass it along to the client. If set, the client can create the Lure. 

However we weren't updating `%grouper`'s local view of which Lures are enabled, meaning when the Lure is actually claimed it will fail a check against its list of enabled groups. This small change just tweaks the logic so that when we hear about the enabled status of remote groups, we propagate that info to our local store. 

Tested by creating a secret group on Ship A, joining it from Ship B, and running through a signup with an invite link from Ship B to confirm the group is successfully joined.

### DIFF
--- a/desk/app/grouper.hoon
+++ b/desk/app/grouper.hoon
@@ -58,8 +58,10 @@
   ::
       %grouper-answer-enabled
     =/  [name=cord enabled=?]  !<([cord ?] vase)
-    :_  this
-    ~[[%give %fact ~[[%group-enabled (scot %p src.bowl) name ~]] %json !>(b+enabled)]]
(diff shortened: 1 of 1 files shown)


mikolajpp (commented):
looks good!

Fang- (approved):
This looks like it fixes the issue you found, yeah. You might want a quick look from @arthyn too, just to make sure this isn't too divergent from the intended architecture here. But this would be fine to ship even if it is, easy to revert/change later.

PR #4184, opened 2024-11-14
desktop: fix profile screen navigation, leave chat button

patosullivan:
fixes TLON-3224

Adds a new stack navigator for the profile screen on desktop, fixes the "Leave" button on channels (which used the react-native `Alert` dialog, which doesn't work on web).


latter-bolden (approved):
Code looks good, working locally!

PR #4160, opened 2024-11-08
desktop: fix image rendering and image preview

patosullivan:
Fixes TLON-3171

Added a new WebImage component that will reliably render on web (using just the `img` tag), on web it is now the default export for Image.

Updated react-native-image-zoom to the latest version, which includes the new Zoomable component which allows you to wrap another image component rather than use their built-in image component (just using Zoomable on web for now). `minPanPointers` was deprecated, but it looks like we can switch to `maxPanPointers` without issue.

Copilot summary:
This pull request includes updates to the `@likashefqet/react-native-image-zoom` package, refactoring of the `Image` component to support web platforms, and adjustments to the `ImageViewerScreenView` component to handle these changes. Additionally, it includes updates to the `pnpm-lock.yaml` file to reflect these changes.

### Dependency Updates:
* Updated `@likashefqet/react-native-image-zoom` from version 3.0.0 to 4.2.0 in `package.json` and `packages/ui/package.json` [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L55-R57) [[2]](diffhunk://#diff-1ae5a5e8c23bd3bc31ea590a9cbe48eb9ad3bbddc0fb5732d08c04a53c8ad51dL29-R29).

### Component Refactoring:
* Refactored `Image` component to support web platforms by adding a `WebImage` component and conditionally exporting it based on the platform in `packages/ui/src/components/Image.tsx`.
* Updated `ImageWithFallback` to use `StyledBaseImage` instead of `Image` in `packages/ui/src/components/Image.tsx`.

### Component Adjustments:
* Modified `ImageViewerScreenView` to use `Zoomable` component for web and updated related logic in `packages/ui/src/components/ImageViewerScreenView.tsx` [[1]](diffhunk://#diff-861c497936c346df53ceab9580c9e52db6bf37774df9adb78a14da0d95460189L1-R1) [[2]](diffhunk://#diff-861c497936c346df53ceab9580c9e52db6bf37774df9adb78a14da0d95460189L12-R23) [[3]](diffhunk://#diff-861c497936c346df53ceab9580c9e52db6bf37774df9adb78a14da0d95460189L38-R50) [[4]](diffhunk://#diff-861c497936c346df53ceab9580c9e52db6bf37774df9adb78a14da0d95460189L69-R108).
[...]

PR #4187, opened 2024-11-14
BlockRenderer: avoid setting aspect ratio if unset

jamesacklin:
Fixes a regression introduced in #4160 where Gallery blocks (in 2-column grid format) had their aspect ratio explicitly set and were not square-cropped as before.

This looks to the 'unset' aspect set in SmallContentRenderer (which is used in Gallery blocks) and passed down through props. If the aspect ratio is explicitly unset, we set nothing in the style prop for `Pressable`.

Proof: 
Screenshot 2024-11-14 at 6 06 45 PM

### DIFF
--- a/packages/ui/src/components/PostContent/BlockRenderer.tsx
+++ b/packages/ui/src/components/PostContent/BlockRenderer.tsx
@@ -244,6 +244,8 @@ export function ImageBlock({
     });
   }, []);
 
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):
Checked locally, it works.

I'm kind of confused about how aspect ratio is supposed to flow downward here and why it's unset on `SmallContentRenderer`. Maybe we can take a look tomorrow

PR #4188, opened 2024-11-15
custom channels: add notes.app-like components

davidisaaclee:
- Adds _Summaries_ collection
- Adds _Scratchpad_ content type (notably not a draft input) which allows the author of a text post to edit it inline
- Adds `PostUsingContentConfigurationScreen` screen in root navigator which shows a post detail view, respecting the channel's configuration when displaying the post – _Summaries_ is currently the only collection that uses this route
- Enables editing posts from `ConnectedPostView` by passing `editPost`. 
  - ~~This prop was included in the original `RenderItemType` (i.e. `Component` of `BaseScrollerItem`), but doesn't appear to be used: [this element is missing the `editPost` prop](https://github.com/tloncorp/tlon-apps/blob/6061c8fd1a66a02c2e29bec2534c655d44564623/packages/ui/src/components/Channel/Scroller.tsx#L578-L590)~~
  - ~~This means that `Scratchpad` won't work in collections using `LegacyPostView` unless we connect `editPost` on that linked `Scroller` code – this is shown in the demo video below when I switch the collection to chat and edits don't work~~
  - Added ability to edit `LegacyPostView` in [c8f075b](https://github.com/tloncorp/tlon-apps/pull/4188/commits/c8f075bd4d506bf6b236cc041d3ee11b51a3419f)

There's no new input for this preset, but the notebook input should work perfectly.

https://github.com/user-attachments/assets/f692f9ad-1480-4948-8d96-e4a93607bf3b


apps/tlon-mobile/src/fixtures/Channel.fixture.tsx
```diff
@@ -514,4 +593,17 @@ export default {
...
+      initialContentConfiguration={{
+        ...ChannelContentConfiguration.defaultConfiguration(),
+        defaultPostCollectionRenderer: {
+          id: CollectionRendererId.boardroom,
+          configuration: {
+            showAuthors: true,
+          },
+        },
+      }}
+    />
```
davidisaaclee, line 607: I added this to test presets of configurations as fixtures – you can just pass an initial config here, and then you can tap the floating `Toggle configurator` button in the fixture to open the config bar. (I had a notes app example here, but it doesn't work because of navigation)

packages/ui/src/utils/react.ts
```diff
@@ -0,0 +1,25 @@
...
+ * >) {
+ *   return <Button {...forwardedProps}> ... </Button>;
+ * }
+ * ```
+ */
+export type ForwardingProps<
+  E extends ElementType,
+  CustomProps extends Record<string, unknown>,
+  OmitKeys extends keyof ComponentPropsWithRef<E> = never,
+> = CustomProps & Omit<ComponentPropsWithRef<E>, keyof CustomProps | OmitKeys>;
```
davidisaaclee, line 25: I was doing this a couple of times so decided to make it a shared thing.

I use [this wrapper](https://gist.github.com/davidisaaclee/fdd1baa1d5fed4463923ea27bfc04522) in my personal projects to do similar stuff with stronger type guarantees (ensures you don't expose "connected" props; automatically handles refs), which would look like:

```tsx
export const DetailPostView = connectComponent(
  typeof YStack,
  function ({
    children: BaseComponent,
    ownProps: { post, navigateBack },
  }: ConnectorProps<
    { post: db.Post; navigateBack?: () => void },
    PartialProps<typeof YStack, "backgroundColor" | "children">
  >) {
    return (
      <BaseComponent backgroundColor="...">
        <ImplementationBlahBlah post={post} onPress={navigateBack} />
      </BaseComponent>
    );
  }
);
```

`ForwardingProps` approach is definitely simpler!

packages/ui/src/components/postCollectionViews/SummaryCollectionView.tsx
```diff
@@ -0,0 +1,131 @@
...
+  onPress?: () => void;
+}) {
+  const post = useLivePost(item);
+
+  const { title, subtitle, authorId } = useMemo(
+    () => ({
+      key: post.id,
+      authorId: post.authorId,
+      ...(post.title
+        ? {
```
davidisaaclee, line 60: When testing this against `createFakePosts` creating notes, I see title / subtitle here – not sure how to get that from one of the app inputs. (I thought gallery input had title, but I don't see it anymore...)

packages/ui/src/utils/react.ts
dnbrwstr: Nice! This is helpful. In some cases tamagui's styleable can be useful for this too, but only works for tamagui components, and may have a little overhead. Looks like: 

```ts
export const TextInputWithIcon = React.memo(
  BaseTextInput.styleable<{ icon: IconType }>(({ icon, ...props }, ref) => {
    return (
      <XStack
        borderRadius="$l"
        borderWidth={1}
        borderColor="$border"
        alignItems="center"
        paddingLeft="$xl"
        gap="$l"
      >
        <Icon type={icon} customSize={['$2xl', '$2xl']} />
        <BaseTextInput
          paddingLeft={0}
          borderWidth={0}
          borderRadius={0}
          flex={1}
          ref={ref}
          {...props}
        />
      </XStack>
    );
  })
);
```

dnbrwstr (approved):
Looks great!

packages/ui/src/utils/react.ts
davidisaaclee: oh yeah – I was using that for `Carousel` but iirc, I couldn't figure out how to add a custom ref type...

packages/ui/src/components/postCollectionViews/SummaryCollectionView.tsx
davidisaaclee: aha, that's a bug in the custom channels branch then, looks like it always uses the gallery big input

davidisaaclee:
the detail view here is buggy but going to merge and just fix forward

PR #4192, opened 2024-11-17
ChatOptionsSheet: mark group/channel/DM/groupDM as read

jamesacklin:
Adds a "mark as read" option to the channel actions menu, which handles whole groups, channels, DMs, and group-DMs. 

Video:

https://github.com/user-attachments/assets/e1f8b81c-137c-4ddc-92a2-684bc69f037d

Fixes TLON-2720

### DIFF
--- a/packages/ui/src/components/ChatOptionsSheet.tsx
+++ b/packages/ui/src/components/ChatOptionsSheet.tsx
@@ -1,3 +1,4 @@
+import { useQuery } from '@tanstack/react-query';
 import { sync } from '@tloncorp/shared';
 import * as db from '@tloncorp/shared/db';
 import * as logic from '@tloncorp/shared/logic';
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):

jamesacklin:
@latter-bolden Expanded the scope slightly and included recursive channel unread-marking from the group options sheet.

latter-bolden:
> Expanded the scope slightly and included recursive channel unread-marking

@jamesacklin it would be simpler if we could use `markGroupRead`, did you try using that?

jamesacklin:
@latter-bolden I did. Unfortunately that works optimistically/locally, but the fact that there are unreads on child channels brings the count back when it re-queries. Open to ideas there though?

latter-bolden (changes requested):
Blocked pending backend convo

jamesacklin:
Sweet, if it's supposed to be handled on the back-end to make the client simpler then I'm all for it.

jamesacklin:
@latter-bolden Scoped it back down. I'll open another ticket to support the same thing for all channels in a group from ChatList.

PR #4193, opened 2024-11-17
native: admin note edit + note edit button

latter-bolden:
We were locking the edit action to only post authors. Sounds like we make a special case for admins of notebook posts, so this just updates the logic to account for that.

Also fixes the member roles query to live update as roles do and adds an _Edit_ button to the _Notebook_ post view.

Fixes TLON-3233

### DIFF
--- a/packages/shared/src/store/dbHooks.ts
+++ b/packages/shared/src/store/dbHooks.ts
@@ -357,7 +357,10 @@ export const useGroupByChannel = (channelId: string) => {
 
 export const useMemberRoles = (chatId: string, userId: string) => {
   const { data: chatMember } = useQuery({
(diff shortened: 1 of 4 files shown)


jamesacklin:
Tested in Simulator, this does work. Thanks!

patosullivan (commented):
Works! Only thing I noticed was that the note content got loaded into the reply input after the edit was submitted and I was routed to the detail view:

Simulator Screenshot - iPhone 15 - default - 2024-11-18 at 05 56 32

dnbrwstr (changes requested):
Permissions changes look good, but can we just use the standard overflow menu here?

dnbrwstr (approved):
Never mind for now. We should standardize detail view overflow menus though!!

PR #4194, opened 2024-11-18
channel manager: prevent deleting last channel in group

jamesacklin:
Does what it says on the tin; fixes TLON-3134, which is a bit of an edge-case.

If the user tries to delete the last channel in a group, we pop a warning and tell them to create another channel (or delete the entire group).

This is a particularly interesting bug—it was possible to enter the channel manager _from_ a channel, delete that channel, then .goBack() your way out of it only to be dumped into a channel that doesn't exist (and therefore get a blank screen). 

To complicate matters, you get to the channel manager from ChatOptionsSheet, which is either (A) in the channel itself via the "..." menu, or (B) from GroupChannelsScreen via the "..." menu, or (C) from long-pressing the group on the ChatList screen.

I brute-forced this and just made all the back-navigation explicit for these screens. It would be nice to rely on the native stack order but, as illustrated above, we have multiple ways into the same screen and I think it's better to be much more explicit.

From the channel:

https://github.com/user-attachments/assets/2d4e03ba-0a19-429a-b61a-6b5ba737a231

From long-press on ChatList:

https://github.com/user-attachments/assets/68b1c6c8-b0c2-49ce-b19e-e5d1ec22c284

### DIFF
--- a/packages/app/features/groups/EditChannelScreen.tsx
+++ b/packages/app/features/groups/EditChannelScreen.tsx
@@ -21,9 +21,9 @@ export function EditChannelScreen(props: Props) {
     const prevChannel = data;
     if (prevChannel) {
       deleteChannel(prevChannel.id);
-      props.navigation.goBack();
+      props.navigation.navigate('ManageChannels', { groupId });
     }
-  }, [data, deleteChannel, props.navigation]);
+  }, [data, deleteChannel, props.navigation, groupId]);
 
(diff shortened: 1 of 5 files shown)


jamesacklin:
Extra commits are from me branching off a work branch, my fault :(

dnbrwstr (approved):
Lgtm!

PR #4195, opened 2024-11-18
native: reconfigure telemetry defaults

latter-bolden:
See Linear issue for context.

Fixes TLON-3237


latter-bolden:
Needs to be reworked, closing for now.

PR #4196, opened 2024-11-18
desktop: adjust font sizes and line heights down on web

patosullivan:
Adds two new functions that adjust font sizes and line heights down by 2 on web. No idea if we actually want to adjust down by 2 exactly, but it looks better imo.

This follows the pattern that uniswap uses: https://github.com/Uniswap/interface/blob/main/packages/ui/src/theme/fonts.ts

Note that this isn't based on screen size, just whether or not we're rendering on web. Tamagui doesn't seem to provide a way to use media breakpoints to set font size in the config.

### DIFF
--- a/packages/ui/src/tamagui.config.ts
+++ b/packages/ui/src/tamagui.config.ts
@@ -1,7 +1,7 @@
 import { createAnimations } from '@tamagui/animations-moti';
 import { createMedia } from '@tamagui/react-native-media-driver';
 import { Platform } from 'react-native';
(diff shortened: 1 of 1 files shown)


dnbrwstr (changes requested):
This looks useful, but I think we should hold off for now. Last week we talked about first getting to visual parity between web and desktop, then starting to tweak desktop specific stuff -- that way we can adjust all the theme params at once instead of being in an in-between state.

PR #4197, opened 2024-11-18
ChatOptionsSheet: mark group as read

jamesacklin:
Adds an optional prop to readGroup in activityApi to either mark the group read "deeply" or not (we badge it for new members requesting approval and clear the dot when navigating to it via Activity—this has nothing to do with unread messages in channels).

Adds a menu item to the ChatOptionsSheet for groups with unreads to deep-mark as read.

Fixes TLON-3240

### DIFF
--- a/packages/app/features/top/ChatListScreen.tsx
+++ b/packages/app/features/top/ChatListScreen.tsx
@@ -208,7 +208,11 @@ export function ChatListScreenView({
       if (item.pin?.type === 'channel' || !item.group) {
         chatOptionsSheetRef.current?.open(item.id, item.type);
       } else {
-        chatOptionsSheetRef.current?.open(item.group.id, 'group');
+        chatOptionsSheetRef.current?.open(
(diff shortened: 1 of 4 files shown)


latter-bolden (approved):
Passing deep accordingly does it. 

Only note is the UX is a little weird since it has to finish grabbing the unread before it shows the option. It lags a bit before popping in. Always including it and disabling if no unreads might be less jarring.

jamesacklin:
Sweet, I'll proceed. Thanks!

> Always including it and disabling if no unreads might be less jarring.

I was thinking maybe I should pass it in _from_ the chat list item itself, then check to be sure once it's open.

latter-bolden:
> I was thinking maybe I should pass it in from the chat list item itself, then check to be sure once it's open.

Even better :) ChatList data should be fresh

latter-bolden (approved):
Looks good! Thanks for the edit

PR #4202, opened 2024-11-19
chat input: fix infinite loop on ref

patosullivan:
fixes TLON-3234 by forcing the TextInput to update using setNativeProps *without* setting any `text`, which is apparently additive, and which also causes onChangeText to fire again, which was causing the loop.

I also renamed `text` to `controlledText` for clarity's sake.


jamesacklin:
Works as intended, thank you!

dnbrwstr (approved):
Nice, lgtm!

PR #4203, opened 2024-11-19
native: fix issue with dark/light mode switching

patosullivan:
fixes tlon-3111 (the symptom, anyway).

Looks like this is related to this issue https://github.com/tamagui/tamagui/issues/2856

I attempted updating to the latest tamagui, no dice. I tried setting `fastSchemeChange:false` in the tamagui config, didn't work.

It seems like the issue is happening only with the Channel/index.tsx and and MessageInputBase.tsx components in particular. The backgrounds just don't update when the theme context changes.

If we pull the background variable out manually with getVariableValue(useTheme().background)) it always updates appropriately. I did that here.

We could go with this for now and figure out why this is happening later, or I could keep digging to try to figure out the root cause and hopefully fix that. Fine with me either way.

### DIFF
--- a/packages/ui/src/components/Channel/index.tsx
+++ b/packages/ui/src/components/Channel/index.tsx
@@ -16,7 +16,14 @@ import { ImagePickerAsset } from 'expo-image-picker';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { FlatList } from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
-import { AnimatePresence, SizableText, View, YStack } from 'tamagui';
+import {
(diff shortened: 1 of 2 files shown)


latter-bolden (approved):
Lgtm. I'm okay with a symptom fix for now, especially if we think the true issue is an underlying bug in Tamagui.

Can you confirm unread counts look normal as well when toggling theme?

patosullivan:
> Lgtm. I'm okay with a symptom fix for now, especially if we think the true issue is an underlying bug in Tamagui.
> 
> Can you confirm unread counts look normal as well when toggling theme?

Yep, they do!

PR #4204, opened 2024-11-20
native: simulate pressable event bubbling

davidisaaclee:
_To preface: I don't know if this is a great thing to add to the `Pressable` primitive – opening this for discussion. Alternatives include (1) plumbing a `disabled` prop to `ChatMessage` for my one use case, (2) adding a flag to `Pressable` to opt-in to this behavior, or (3) creating a new `Pressable`-like component with this behavior (`OptionallyPressable`? 🤷 )._

---

I was trying to add a long-press to a container view holding a `ChatMessage` using `Pressable`. My handler wouldn't get called – swapping out `ChatMessage` for a blank `View` worked.

`ChatMessage` has a `Pressable` which has no bound handlers in this scenario, and I expected events to bubble up the stack like on web – but it looks like RN doesn't have that behavior. Here's a recording of the fixture from this PR demonstrating this – the specific case is when `Enable inner press handler` is **off** and I'm tapping on the inner view – I expect that the inner view does not activate (because no handler) and the outer view's handler does activate (since I'm tapping over the outer view):

https://github.com/user-attachments/assets/b9eb2607-b0ab-4fbb-8c17-55d6c1fb0cf3

To solve my initial problem:
1. If I could "reach across" to the `ChatMessage`'s `Pressable`, setting `pointerEvents="none"` would work
2. Setting `disabled` on the inner `Pressable` passes events through to ancestors

(1) seems violent to do for every use case – I'd prefer a component that can be optionally interactive, and when the component is configured to be non-interactive, events naturally pass through to ancestors.

684721376ce54eba6318f56976672baffdc2dbeb implements (2) – setting `disabled={true}` when there are no bound handlers.

https://github.com/user-attachments/assets/f7a552e3-4c14-405d-83ac-4a2225fbab14

---

I think there's a solution here if we were to move to react-native-gesture-handler instead of the Tamagui pressable mixin, but I don't think it'd be very different in implementation – I think we'd still be disabling interaction when missing handlers. (There are [APIs for composing gestures](https://docs.swmansion.com/react-native-gesture-handler/docs/fundamentals/gesture-composition) but they require references to the relevant gestures, which would be a lot of plumbing.)

### DIFF
--- a/packages/app/fixtures/Pressable.fixture.tsx
+++ b/packages/app/fixtures/Pressable.fixture.tsx
@@ -1,0 +1,93 @@
+import { Pressable, Text, View } from '@tloncorp/ui';
+import { useValue } from 'react-cosmos/client';
+import { Alert, SafeAreaView, Switch } from 'react-native';
+
+export default {
+  Nested() {
+    const [innerPressableEnabled, setInnerPressableEnabled] = useValue(
+      'Enable inner press handler',
+      { defaultValue: true }
+    );
+    const [siblingPressableEnabled, setSiblingPressableEnabled] = useValue(
+      'Enable sibling press handler',
+      { defaultValue: true }
+    );
+    return (
+      <SafeAreaView>
+        <View>
+          <Pressable
+            padding={20}
+            backgroundColor="$negativeBackground"
+            onPress={() => {
+              Alert.alert('Outer: onPress fired');
+            }}
+            onTouchEndCapture={(event) => {
+              event.stopPropagation();
+            }}
+          >
+            <Text padding={20}>Outer view (has onPress handler)</Text>
+            <Pressable
+              padding={20}
+              backgroundColor="$positiveBackground"
+              onPress={
+                innerPressableEnabled
+                  ? (event) => {
+                      event.persist();
+                      console.log(event);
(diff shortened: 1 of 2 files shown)


latter-bolden (approved):
This was definitely a point of confusion for me when first working with RN. The approach you implemented feels sensible, I think it will be less surprising to have the clicks bubble by default.

We should however make sure this doesn't behave strangely anywhere in existing code. Would prefer to avoid merging until after we cut this week's release.

dnbrwstr:
Makes sense to me -- @patosullivan could you take a look? As you've spent time in pressable most recently.

davidisaaclee:
wanted to add more context that I realized was missing:
- This is not really bubbling, since we can't receive the press event on _both_ the outer and inner views (i.o.w. we can't optionally `preventDefault`) – this is more about treating an inert `Pressable` as any other non-interactive view
- I'd be open to any alternative solutions to my initial issue – I just couldn't think of any apart from the ones I posted in OP

patosullivan (approved):
Makes sense to me!

davidisaaclee:
great – I'll merge this on Friday afternoon after the build review

dnbrwstr (approved):
Lgtm when you're ready to merge!

PR #4205, opened 2024-11-20
native: update telemetry initialization

latter-bolden:
Moves the set telemetry screen to the self hosted login flow and makes sure we initialize things correctly for authenticated users. Adds a generic telemetry interface that unifies the PostHog API so it's easier to use from components in the _app_ package.

Fixes TLON-3239


dnbrwstr (approved):
Would be nice if we could simplify the helper logic, but not blocking. Lgtm!

PR #4206, opened 2024-11-20
group previews: fix go to groups button

patosullivan:
fixes TLON-3007

This was happening because weren't passing an `onGroupAction` function to the GroupPreviewSheet from the user profile. `onGroupAction` was set as an optional prop, which seems like a mistake, so I made it not optional.

After setting it as not optional, I found that we were pulling the GroupPreviewSheet into ContentReference unnecessarily (rather than using some local state for control and rendering the sheet there, we can just use the GroupPreviewSheet that's already rendered in the Channel view and PostScreenView and use the existing onPressGroupRef method in the navigation context).

I also noticed that we were navigating users directly to the `lastChannel` of a group when going to a group, which seems like it might not be what users expect here, so instead I've got the `goTo` action setup to reset the navigation stack to the GroupChannelsScreen for that particular group.

I also added a new eslint rule to catch any instance where we're accidentally importing from `packages/*`.


latter-bolden (approved):
This is awesome! 

I bumped into this on my contacts branch as well and implemented something similar. I think what you have here is more elegant, let's get it in.

PR #4209, opened 2024-11-20
groups: auto-join new channels in all cases

Fang-:
Problems with channel auto-join were two-fold:

First, channels-server was always poking the local groups agent with the creation command, instead of the groups agent on the host of the group. Here we make sure to poke the appropriate ship.

Second, groups agent was explicitly no-opping when a channel was added to a group it hosts itself. Here, we remove that condition, letting that case proceed into channel joining logic unhindered.

Closes TLON-3151.


arthyn (approved):
🫡

PR #4211, opened 2024-11-20
app: theme switcher

jamesacklin:
**Super-extra fun, absolutely not essential or time-sensitive.**

Adds a theme switcher to the app and provides a set of base16 themes.

Wraps TamaguiProvider with a ThemeContext provider with hooks to set and clear the theme. The user's theme selection is stored in AsyncStorage. We use either `light` or `dark` based on the user's device theme selection if they set to "auto" (default behavior). The themes themselves do not encode any kind of 'lightness' or 'darkness,' so if we are calling upon `useIsDarkMode()` anywhere in the app, we might see some odd color choices.

Also touches up a couple places that were non-reactive to theme changes and got placeholderTextColor working in BareChatInput.

Video demo: https://s3.rilfun-lidlen.startram.io/bucket/rilfun-lidlen/2024.11.20..20.41.48..7a5e.353f.7ced.9168-Simulator-Screen-Recording---iPhone-16---2024-11-20-at-15.38.26.mp4


packages/app/provider/index.tsx
```diff
@@ -1,13 +1,92 @@
...
   children,
   ...rest
 }: Omit<TamaguiProviderProps, 'config'>) {
+  const isDarkMode = useIsDarkMode();
+  const [activeTheme, setActiveTheme] = useState<ThemeName>(
+    isDarkMode ? 'dark' : 'light'
+  );
+
+  useEffect(() => {
+    const loadStoredTheme = async () => {
```
dnbrwstr, line 25: I think boilerplate in this file could be reduced significantly using `keyValue`.

packages/app/features/settings/ThemeScreen.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  return (
+    <>
+      <ScreenHeader
+        title="Theme"
+        backAction={() => props.navigation.goBack()}
+      />
+      <ScrollView>
+        <YStack flex={1} padding="$l" gap="$s">
+          {themes.map((theme) => (
```
dnbrwstr, line 79: Could we use `ListItemInput` here?

dnbrwstr (approved):
A couple nits, but I think this is great, and I'm all for merging. Light/dark/auto is strictly an improvement, the ui is helpful for debugging, and it really doesn't add any more complexity than we're already managing.

packages/app/features/settings/ThemeScreen.tsx
jamesacklin: Yes chef

PR #4212, opened 2024-11-21
native: move useChannelContext to `shared`

davidisaaclee:
I wanted to use `useChannelContext` in a new location for a custom channel layout – issue is that `useChannelContext` was in `app`, while custom channel components are in `ui` – so I would need to pass every one of the props down through to my component.

**Why is `useChannelContext` in `app`?** The best answer is that it relies on [the `storage` singleton in `app`](https://github.com/tloncorp/tlon-apps/blob/524dbe76841d49e6c9526ca91b2ba8de7aae0f41/packages/app/lib/storage.ts#L23). There's also a feature flag access.

To work around this, I moved `useChannelContext` into `shared` by:
1. Passing the feature flag state into the hook as an argument. This can be made better when/if we merge https://github.com/tloncorp/tlon-apps/pull/4113/commits/84841b2d4b67a28f57a46d5015a9dfb34aed5de0
2. Add a definition for a storage React context in `shared` – this structurally matches the `react-native-storage` object we've been using, but without pulling that in as a dependency to `shared`. Since this is in `shared`, anything in `shared` and above can reference it.
3. Add a provider for that context in `app`, which uses the existing `storage` singleton.
4. Move `useChannelContext` into `shared`, and access `storage` through the context (assuming that someone will provide the `StorageContext`).

I tested by building for iOS and checking that:
- I can draft and send posts
- I can draft and send replies
- If I leave a draft or reply and then return to the relevant screen, the draft is preserved


dnbrwstr:
Question -- what about removing this storage altogether? Seems redundant with `keyValue`. @latter-bolden wdyt?

latter-bolden (changes requested):
Agree that storage module is redundant, looks like that's a holdover from the legacy mobile app we inherited. Should prefer `db/keyValue` (which is already in _shared_).

@davidisaaclee could you either update drafts to use `db/keyValue` or constrain the new provider to be draft focused rather than a generic storage mechanism? 

I'm fine with whatever's most expedient there given this is a side quest from the actual custom channels change you're trying to effect.

latter-bolden:
For further context/to rabbit hole a bit:

The new pattern of key instantiation at the bottom of `db/keyValue` is what we're trying to move to, but it's not built for dynamic keys like we're using for drafts. If you go that route, could either try to wrap that or lean on the old-style _react query_ pattern above. 

That said, I'm not sure I understand why drafts are stored as key values to begin with — seems like they map 1:1 with channels/posts. Might make more sense as a column on the respective table

davidisaaclee:
@latter-bolden I agree that putting the drafts into SQLite is nicer, but I went with the "storage item" implementation to reduce diff.

I implemented dynamic keys on the `postDraft` storage item by constructing a new storage item on every access – afaict, there's nothing in the `createStorageItem` implementation that would break there.

Tested again by making some drafts in different channels / groups, exiting app, returning to channels and seeing that the drafts were persisted.

dnbrwstr (approved):
This lgtm now, and I believe @latter-bolden is out for a bit, so I'd say go ahead and merge in when ready.

latter-bolden (approved):
Looks good with the change! Thanks for knocking that out

dnbrwstr:
Correction: @latter-bolden is not out

PR #4213, opened 2024-11-21
custom channels: fixup new thread view

davidisaaclee:
Expands on the new post detail / thread screen added in https://github.com/tloncorp/tlon-apps/pull/4188:
- Allow scrolling
- Show replies
- Add context menu for all posts in screen (enabling edit + delete)
- Add reply input (hardcoded as `chat` input)

This branch includes https://github.com/tloncorp/tlon-apps/pull/4212 and https://github.com/tloncorp/tlon-apps/pull/4204.


dnbrwstr (approved):
Overall looks good to me, only question is re: `overrideChannelType` -- could it make sense to use more specific props instead? So, eg `showMetaEditor={true}` instead of `overrideChannelType='notebook'`. No worries if that'd overcomplicate, but I do want to  reduce dependence on `channel.type` where possible.

davidisaaclee:
@dnbrwstr makes sense + i agree. looking at the downstream uses of `channelType`, I think I'd want to put more thought into how to remove it from `MessageInput` – here are some features that I'd have a hard time naming:

https://github.com/tloncorp/tlon-apps/blob/093aabdfab84ec893c551927fd233c620a67282f/packages/ui/src/components/MessageInput/index.tsx#L574-L576

https://github.com/tloncorp/tlon-apps/blob/093aabdfab84ec893c551927fd233c620a67282f/packages/ui/src/components/MessageInput/index.tsx#L465-L471

PR #4214, opened 2024-11-21
logs: new agent to gather and submit app logs

mikolajpp:
The %logs agent gathers log events from other agents in the groups desk and submits them as "Backend Log" events to PostHog. Currently the only event supported is the `%fail` event usually generated in agent `+on-fail` arm.

As a trial, in this PR we enable logging of crashes in %channels, %channels-server, %groups and %activity.

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -41,7 +41,7 @@
 /-  a=activity, c=channels, ch=chat, g=groups
 /+  *activity, ch-utils=channel-utils, v=volume, aj=activity-json,
     imp=import-aid
-/+  default-agent, verb, dbug
+/+  default-agent, verb, dbug, logs
 ::
 =/  verbose  |
 =>
@@ -107,7 +107,11 @@
   ++  on-agent   on-agent:def
   ++  on-peek    peek:cor
   ++  on-leave   on-leave:def
-  ++  on-fail    on-fail:def
+  ++  on-fail
+    |=  [=term =tang]
+    ^-  (quip card _this)
+    :_  this
+    [(log-fail:logs /logs our.bowl (fail-event:logs term tang))]~
   --
 |_  [=bowl:gall cards=(list card)]
 ++  abet  [(flop cards) state]
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -417,6 +417,7 @@
   |=  [=(pole knot) =sign:agent:gall]
   ^+  cor
   ?+    pole  ~|(bad-agent-wire+pole !!)
+    [%logs ~]  cor
     [%pimp ~]  cor
     [%wake ~]  cor
   ::
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -4,7 +4,7 @@
 ::
 /-  c=channels, g=groups
 /+  utils=channel-utils, imp=import-aid
-/+  default-agent, verb, dbug, neg=negotiate
+/+  default-agent, verb, dbug, neg=negotiate, logs
 ::
 %-  %-  agent:neg
     [| [~.channels^%1 ~ ~] ~]
@@ -58,7 +58,12 @@
   ::
   ++  on-peek    on-peek:def
   ++  on-leave   on-leave:def
-  ++  on-fail    on-fail:def
+  ++  on-fail
+    |=  [=term =tang]
+    ^-  (quip card _this)
+    :_  this
+    [(log-fail:logs /logs our.bowl (fail-event:logs term tang))]~
+  ::
   ++  on-agent
     |=  [=wire =sign:agent:gall]
     ^-  (quip card _this)
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -640,6 +640,7 @@
   ?+    pole  ~|(bad-agent-take/pole !!)
       ~   cor
       [%epic ~]  (take-epic sign)
+      [%logs ~]  cor
       [%helm *]  cor
       [%activity %submit *]  cor
       [%groups %role ~]  cor
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -8,7 +8,7 @@
 /-  meta
 /-  e=epic
 /+  default-agent, verb, dbug
-/+  v=volume, s=subscriber, imp=import-aid
+/+  v=volume, s=subscriber, imp=import-aid, logs
 /+  of
 /+  epos-lib=saga
 ::  performance, keep warm
@@ -69,7 +69,11 @@
   ++  on-peek   peek:cor
   ::
   ++  on-leave   on-leave:def
-  ++  on-fail    on-fail:def
+  ++  on-fail
+    |=  [=term =tang]
+    ^-  (quip card _this)
+    :_  this
+    [(log-fail:logs /logs our.bowl (fail-event:logs term tang))]~
   ::
   ++  on-agent
     |=  [=wire =sign:agent:gall]
--- a/desk/app/logs.hoon
+++ b/desk/app/logs.hoon
@@ -61,8 +61,7 @@
     =+  !<(=a-log:l vase)
     ?-    -.a-log
         %log
-      =/  =log-item:l  [now.bowl +.a-log]
-      (send-posthog-event sap.bowl log-item)
+      (send-posthog-event sap.bowl now.bowl +.a-log)
     ==
   ==
 ::
--- a/desk/app/logs.hoon
+++ b/desk/app/logs.hoon
@@ -5,7 +5,7 @@
 =>
   |%
   +$  card  card:agent:gall
-  +$  current-state  [%0 =logs:l]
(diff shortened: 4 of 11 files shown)


desk/lib/logs.hoon
```diff
@@ -0,0 +1,57 @@
...
+  ++  tang
+    |=  t=^tang
+    ^-  $>(%a json)
+    ?~  t  a+~
+    =/  tame=(list tape)
+      %+  turn  t
+     (cork (cury wash [0 80]) zing)
+    ::XX posthog does not display newlines properly
+    :: s+(crip (zing `(list tape)`(join "\0a" tame)))
+    a+(turn tame tape:enjs)
```
Fang-, line 26: You want to preserve the newlines here, so I don't think you want to `+zing`. Or rather, you probably want to do that _later_:

```suggestion
    =/  tame=(list tape)
      %-  zing
      %+  turn  t
      (cury wash [0 80])
    ::XX posthog does not display newlines properly
    :: s+(crip (zing `(list tape)`(join "\0a" tame)))
    a+(turn tame tape:enjs)
```

desk/lib/logs.hoon
```diff
@@ -0,0 +1,57 @@
...
+    =+  date=(yore id)
+    =+  frac=(head f.t.date)
+    ::  convert urbit fractosecs to millisecs
+    ::
+    =+  mili=(div (mul frac 1.000) 65.563)
+    :-  %s
+    %-  crip
+    "{((d-co:co 4) y.date)}-{(d-co-2 m.date)}-{(d-co-2 d.t.date)}".
+    "T{(d-co-2 h.t.date)}:{(d-co-2 m.t.date)}:{(d-co-2 s.t.date)}".
+    ".{((d-co:co 3) mili)}"
```
Fang-, line 41: As per [posthog docs](https://posthog.com/docs/data/timestamps#recognized-formats) they also support ingesting unix seconds and millisecond timestamps. So you can just use `+time:enjs:format` here. (:

desk/ted/posthog.hoon
```diff
@@ -0,0 +1,61 @@
+::  posthog: submit an log event to PostHog
+::
+/-  spider
+/+  io=strandio, l=logs
+::
+=+  posthog-key='phc_GyI5iD7kM6RRbb1hIU0fiGmTCh4ha44hthJYJ7a89td'
+=+  posthog-url='https://eu.i.posthog.com/capture/'
+=+  posthog-retry=3
+=+  posthog-retry-delay=~s30
```
Fang-, line 9: Could tune this down if you wanted, ~s5 is probably plenty if we're only doing three retries.

desk/ted/posthog.hoon
```diff
@@ -0,0 +1,61 @@
...
+::      partial responses
+?.  ?&  ?=(%finished -.client-response)
+        =(200 status-code.response-header.client-response)
+    ==
+  ?>  ?=(%finished -.client-response)
+  =*  status-code  status-code.response-header.client-response
+  %-  (slog leaf+"posthog request failed: status {<status-code>}" ~)
+  ;<  ~  bind:m  (sleep:io posthog-retry-delay)
+  $(posthog-retry (dec posthog-retry))
+(pure:m !>(`client-response))
```
Fang-, line 61: The assert here reads a little bit weird after the conditional. Maybe the assert should come before it, and not show up in the conditional at all?

Also this `?.` wants to be a `?:`.

desk/app/groups.hoon
```diff
@@ -69,7 +69,11 @@
   ++  on-peek   peek:cor
   ::
   ++  on-leave   on-leave:def
-  ++  on-fail    on-fail:def
+  ++  on-fail
+    |=  [=term =tang]
+    ^-  (quip card _this)
+    :_  this
+    [(log-fail:logs /logs our.bowl (fail-event:logs term tang))]~
```
Fang-, line 76: You want to handle poke-acks on this wire in `+on-agent`. No-opping is fine, we'll just submit and forget about it.

desk/app/channels-server.hoon
```diff
@@ -58,7 +58,12 @@
   ::
   ++  on-peek    on-peek:def
   ++  on-leave   on-leave:def
-  ++  on-fail    on-fail:def
+  ++  on-fail
+    |=  [=term =tang]
+    ^-  (quip card _this)
+    :_  this
+    [(log-fail:logs /logs our.bowl (fail-event:logs term tang))]~
```
Fang-, line 65: Again, handle in `+on-agent`.

desk/app/activity.hoon
```diff
@@ -107,7 +107,11 @@
   ++  on-agent   on-agent:def
   ++  on-peek    peek:cor
   ++  on-leave   on-leave:def
-  ++  on-fail    on-fail:def
+  ++  on-fail
+    |=  [=term =tang]
+    ^-  (quip card _this)
+    :_  this
+    [(log-fail:logs /logs our.bowl (fail-event:logs term tang))]~
```
Fang-, line 114: Again, handle in `+on-agent`. (;

desk/app/logs.hoon
```diff
@@ -0,0 +1,113 @@
...
+  =/  =id-event:l
+    |-
+    ?.  (has:on-log:l log now.bowl)  now.bowl
+    $(now.bowl (add now.bowl tick))
+  :-  id-event
+  %+  ~(put by logs)  agent
+  (put:on-log:l log id-event log-event)
+::
+++  send-posthog-event
+  |=  [agent=@tas =id-event:l =log-event:l]
```
Fang-, line 62:
```suggestion
  |=  [agent=dude:gall =id-event:l =log-event:l]
```

desk/tests/app/logs.hoon
```diff
@@ -0,0 +1,24 @@
...
+    [%fail %test-fail leaf+"test stacktrace" ~]
+  ;<  caz=(list card)  bind:m
+    (do-poke log-action+!>([%log ev-fail]))
+  =/  fard=(fyrd:khan cage)
+    [q.byk.bowl %posthog noun+!>(`[%test ^-(log-item:l [now.bowl ev-fail])])]
+  ::  expect log submission -posthog
+  ::
+  %+  ex-cards  caz
+  :~  (ex-arvo /posthog/test/(scot %da now.bowl) %k %fard fard)
+  ==
```
Fang-, line 23: As discussed out of band, the fact that the card here contains a vase is troublesome, since this logic as written ends up doing a noun comparison, which isn't sane for types (their contexts might be different even if they otherwise nest).

Short-term you probably just want to test the card explicitly here, unpacking it as you go, and then doing an `+expect-eq:test` on the vase it contains.

Longer-term, it might be good for `/lib/test-agent`'s card comparison functions to know where vases might be, and do an extraction internally, just like `+ex-fact` and `+ex-poke` already do. That's out of scope for here though.

Fang- (changes requested):
Some comments below as we reviewed together. This is gonna be nice to have available, thanks for this!

desk/app/groups.hoon
```diff
@@ -1,2195 +1,58 @@
...
-    =.  cor  ga-give-update
-    ga-core
-  --
---
+/-  g=groups
+/+  *test-agent
+/=  groups-agent  /app/groups
+|%
+++  dap  %groups-test
+++  the-wire  /groups/(scot %p ~zod)/test
```
Fang-, line 6: I think you accidentally a file!

desk/app/logs.hoon
```diff
@@ -77,21 +61,17 @@
     =+  !<(=a-log:l vase)
     ?-    -.a-log
         %log
-      ?>  ?=([@ @ ~] sap.bowl)
-      =+  agent=i.t.sap.bowl
-      =^  id-event  logs  (add-log-event agent +.a-log)
-      (send-posthog-event agent id-event +.a-log)
+      =/  =log-item:l  [now.bowl +.a-log]
+      (send-posthog-event sap.bowl log-item)
```
Fang-, line 65: Given that the `+send-posthog-event` sample also specifies time and event separately, it'd be fine to not explicitly construct a `$log-item` here.

```suggestion
      (send-posthog-event sap.bowl now.bowl +.a-log)
```

desk/lib/logs.hoon
```diff
@@ -19,27 +19,11 @@
     ^-  $>(%a json)
     ?~  t  a+~
     =/  tame=(list tape)
+      %-  zing
       %+  turn  t
-     (cork (cury wash [0 80]) zing)
-    ::XX posthog does not display newlines properly
-    :: s+(crip (zing `(list tape)`(join "\0a" tame)))
+     (cury wash [0 80])
```
Fang-, line 24:
```suggestion
      (cury wash [0 80])
```

(;

Fang- (commented):
Something's up with /app/groups, but mostly fine otherwise!

Fang- (approved):
Looks good now, thank you!

jamesacklin:
Merge at will, this will be great to have ahead of Contacts in the client.

PR #4215, opened 2024-11-21
desktop: fixes omnibus

patosullivan:
Fixes TLON-3248 by including the EditProfile screen in the ProfileScreenNavigatorStack.

Fixes TLON-3249 by removing the camera option and adjusting the copy on web for the attachment sheet/modal.

Fixes TLON-3250 by removing an unnecessary flex style on ListItemEndContent

~Fixes TLON-3251 by downgrading react-native-web to 0.18.12 (latest 0.18), which fixes an issue with react-native-reanimated where it can't animate at least some components (see this issue: https://github.com/software-mansion/react-native-reanimated/issues/4335, the fix is included in our version of react-native-reanimated, but the issue still occurs for us unless we downgrade to 0.18).~

Fixes TLON-3251 by patching react-native-reanimated. A distributed JS module (`react-native-reanimated/lib/module/reanimated2/js-reanimated/index.js`) included some require statements, this caused `createReactDOMStyle` to be undefined when it shouldn't have been, which meant that we didn't fall into the correct branch of `_updatePropsJS` in that file, and jumped to the next conditional where we checked for a property that can't exist on a dom node on web, which threw the "Cannot convert undefined or null to object" error we were seeing. This was preventing some (maybe all?) animations that included updates to a style prop from working, such as ChatListSearch.


jamesacklin:
Awesome, thanks. I'll pull this down and take a look today.

patosullivan:
Ok, we didn't need to downgrade to rnw 18, we just needed to patch a file in reanimated. See updated description.

dnbrwstr (approved):
Lgtm!

PR #4216, opened 2024-11-22
GroupChannelsScreen: show unjoined channels, tap to join

jamesacklin:
Fixes TLON-3257

We need to show users the channels they aren't joined to and give them a way to join them.

- Adds a query for channels that are in the group but you aren't joined to.
- Adds channelActions and a tracked poke in the channelsApi for joining a channel in a group.
- Modifies the addJoinedGroupChannel query to avoid collision.
- Adds the result of the unjoined channel query to GroupChannelScreen to create a list of unjoined channels visible to the user.
- Modifies changeListener to account for changes in the db to the channels table and live-update the unjoined channels list.

This produces the following effect:

https://github.com/user-attachments/assets/a1c0fc7b-12a6-4e5a-a8fc-364d7191c2d0

### DIFF
--- a/packages/app/features/top/GroupChannelsScreen.tsx
+++ b/packages/app/features/top/GroupChannelsScreen.tsx
@@ -5,7 +5,6 @@ import {
 } from '@react-navigation/native';
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
 import { useQuery } from '@tanstack/react-query';
-import * as api from '@tloncorp/shared/api';
 import * as db from '@tloncorp/shared/db';
 import * as store from '@tloncorp/shared/store';
 import {
--- a/packages/app/features/top/GroupChannelsScreen.tsx
+++ b/packages/app/features/top/GroupChannelsScreen.tsx
@@ -4,6 +4,8 @@ import {
   useNavigation,
 } from '@react-navigation/native';
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
+import { useQuery } from '@tanstack/react-query';
+import * as api from '@tloncorp/shared/api';
 import * as db from '@tloncorp/shared/db';
 import * as store from '@tloncorp/shared/store';
 import {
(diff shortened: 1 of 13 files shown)


packages/app/features/top/GroupChannelsScreen.tsx
```diff
@@ -4,6 +4,7 @@ import {
   useNavigation,
 } from '@react-navigation/native';
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
+import { useQuery } from '@tanstack/react-query';
```
patosullivan, line 7:
```suggestion
```

patosullivan (approved):
Looks good! Tried on mobile and web, checked against channel join status on the old app. Everything works!

The one nit is just an unused import.

packages/shared/src/db/queries.ts
```diff
@@ -1711,17 +1765,33 @@ export const addJoinedGroupChannel = createWriteQuery(
...
       .where(eq($channels.id, channelId));
+
+    // Then try to add to nav section if not already there
+    const existing = await ctx.db
+      .select()
+      .from($groupNavSectionChannels)
+      .where(
+        and(
+          eq($groupNavSectionChannels.channelId, channelId),
+          eq($groupNavSectionChannels.groupNavSectionId, 'default')
```
latter-bolden, line 1783: I'm kinda confused about this part, when would we expect the channel to not already be in a section? How does this still work if it's in a section that isn't `default`?

latter-bolden (approved):
Lgtm, just that one question. 

They're pretty heavy as is, draws your focus as much as normal joined channels in the list. If there's any quick fixes there, would be worthwhile

packages/shared/src/db/queries.ts
jamesacklin: You're right—we should check if the channel exists in any section, then only add it to 'default' if it truly is unassigned. Let me adjust.

jamesacklin:
@latter-bolden Addressed your concern in queries.ts. Here it is working (before I applied the aesthetic fixes):

https://github.com/user-attachments/assets/2cf52f7e-452d-4513-a6af-bbea77458628

And here are the cosmetic changes:

| Light | Dark |
|-------|------|
| Screenshot 2024-11-22 at 11 06 14 AM | Screenshot 2024-11-22 at 11 06 09 AM |

latter-bolden:
Awesome, appreciate the change. That feels better

PR #4217, opened 2024-11-22
profile & expose: fix integration with contacts

mikolajpp:
There were some outdated type references in %expose that were preventing the upgrade. Additionally, both %profile and %expose were using the removed `/contact` API endpoint. For now we just migrate to subscribing to `/news` and filtering on our own profile updates. We should migrate these two agents to use new %contacts API soon.

### DIFF
--- a/desk/app/expose.hoon
+++ b/desk/app/expose.hoon
@@ -5,7 +5,7 @@
 ::    then visit in the browser:
 ::    /expose/that/reference/as/copied/123456789
 ::
-/-  c=cite, d=channels, t=contacts-0
+/-  c=cite, d=channels, co=contacts-0
 /+  u=channel-utils, hutils=http-utils,
     dbug, verb
 ::
@@ -13,8 +13,8 @@
 /=  widget  /app/expose/widget
 ::
 |%
-+$  state-0
-  $:  %0
++$  state-1
+  $:  %1
       open=(set cite:c)
   ==
 ::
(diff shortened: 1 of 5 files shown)


mikolajpp:
Fixes TLON-3253

desk/app/expose.hoon
```diff
@@ -5,7 +5,7 @@
 ::    then visit in the browser:
 ::    /expose/that/reference/as/copied/123456789
 ::
-/-  c=cite, d=channels, co=contacts
+/-  c=cite, d=channels, t=contacts-0
```
Fang-, line 8: Can we keep the old import names, here and in other files, please?

Fang- (commented):
The import name change here seems unwarranted/out of scope.

Also, don't both of these agents need `+on-load` logic, to tear down the old subscription and set up the new one?

desk/app/expose.hoon
mikolajpp: We can – although we avoid the name clash with `+co` arm in this way.

desk/app/expose.hoon
```diff
@@ -254,10 +279,10 @@
       ==
     ==
   ::
-      [%contacts ~]
+      [%contacts %news ~]
     ?-  -.sign
       %poke-ack  !!
-      %kick      [[%pass /contacts %agent [our.bowl %conacts] %watch /contact]~ this]
```
Fang-, line 260: lol, oops, good catch

Fang- (approved):
Looks good, thank you!

PR #4220, opened 2024-11-22
native: quick-n-dirty app badge management

davidisaaclee:
Fixes TLON-3149
Fixes TLON-3241

- Adds clientside badge management
- Dismiss read notifications using clientside logic

Both of these features will be faster (i.e. no perceptible delay between action/receipt and UI update), simpler to maintain, and more flexible when implemented on backend (as mentioned in tickets), but here's a quick fix.

There's some latency between opening a channel and the notifications going away – I _think_ this is latency from marking stuff as read and getting the db update. Going in and out of the channel quickly still gets the notifications cleared – I think...

We update the badge in two places:
1. When receiving a notification while app is in background, the NSE simply updates the badge to `number of currently-presented notifications` + 1. I think this can drift into the wrong count when receiving notifications in quick succession, but should always at least show _some_ badge if you have an unread; and will be corrected by...
2. When app is in foreground and the database believes unreads to have changed, the app attempts to remove all presented notifications in channels which are fully read; and then updates the badge to match the remaining notifications.

---

⚠️  **NB: When app is in foreground, badge does not get incremented by new messages** – this is a little weird because you can have unreads without a badge. These do not get added by the code from item (2) above – that code only removes from the set of presented notifications.


packages/shared/src/store/dbHooks.ts
```diff
@@ -200,7 +200,7 @@ export const useContacts = () => {
 
 export const useUnreadsCount = () => {
   return useQuery({
-    queryKey: ['unreadsCount'],
+    queryKey: ['unreadsCount', useKeyFromQueryDeps(db.getUnreadsCount)],
```
davidisaaclee, line 203: this works but causes a lot of seemingly irrelevant query updates – would love to know if there's a way to reduce

packages/app/lib/notifications.ts
```diff
@@ -69,3 +74,62 @@ export const connectNotifications = async () => {
...
+  const stillPresented = await Notifications.getPresentedNotificationsAsync();
+  await Notifications.setBadgeCountAsync(stillPresented.length);
+}
+
+export function useUpdatePresentedNotifications() {
+  const query = store.useUnreadsCount();
+
+  // `useUnreadsCount` updates with a lot of false positives - debounce so we
+  // don't run the updater too frequently
+  const debouncedQueryKey = useDebouncedValue(query.dataUpdatedAt, 500);
```
dnbrwstr, line 128: Since the `query.data` is just `number | undefined`, why not use that number as the key?

dnbrwstr (approved):
Lgtm overall!

packages/shared/src/store/dbHooks.ts
dnbrwstr: Ya, I mean this is true in a number of places, mostly because our query invalidation is very coarse. I think fine for now + we should address at a higher level in a perf pass sometime.

packages/app/lib/notifications.ts
davidisaaclee: I was thinking about this case: This query's data could remain the same but badge could need a change if there is an equal number of +/- read/unreads in different rooms in one transaction, causing no change in the total unread count. (_If that's the case, wouldn't the badge also reflect no change in value?_ Not necessarily, because the badge only represents non-muted channels - I thought useUnreadsCount contains muted channels.)

I'll leave a comment inline if that makes sense

dnbrwstr: You're right that it does -- fwiw `useUnreadsCount` does not appeared to be use anywhere else besides here, so feel free to modify if helpful.

davidisaaclee:
@dnbrwstr I changed `useUnreadsCount` -> `useUnreadsCountWithoutMuted` and use that count directly to trigger the notification change, which does simplify things.

I tried using the value of that query for the badge count instead of counting the presented notifications, but it looks like that query counts the number of unread channels, not unread messages (and it doesn't seem trivial to change that).

Going to merge, happy to do a follow-up if you find issue with the recent changes

PR #4221, opened 2024-11-23
native: gracefully handle hearing about channels you don't have access to

jamesacklin:
Fixes TLON-3259. Handles role-based channel permissions in the groupChannelsList to (A) prevent users without read-roles from seeing channels they're not supposed to, (B) live-update the list of available channels when their role-based access changes, and (C) fixes a bug where we mistakenly assumed you were a member of every channel when inserting them.

- groupsApi now tracks readerRoles and writerRoles for every channel so we have something to check against.
- getUnjoinedGroupChannels now filters out channels that do not have readerRoles or writerRoles that line up with the current user's.
- Finally, in the case where we hear about a role change from the group subscription, we re-check the available channels, run them through role-based permission checks, and insert or drop the requisite channels.

Video demonstration (sound up, fam):

https://github.com/user-attachments/assets/12bc9f73-914a-4ea4-b205-4bbacc4fad37

## Copilot summary 
* Added `writers` property to mock group data in `apps/tlon-web-new/src/mocks/groups.ts` and `apps/tlon-web/src/mocks/groups.ts` to support writer roles in mock group definitions. [[1]](diffhunk://#diff-123583aa72f9d495c4492edc27c5bdc52f61e95fb358cff843d238f36818d5c7R211) [[2]](diffhunk://#diff-123583aa72f9d495c4492edc27c5bdc52f61e95fb358cff843d238f36818d5c7R278) [[3]](diffhunk://#diff-123583aa72f9d495c4492edc27c5bdc52f61e95fb358cff843d238f36818d5c7R291) [[4]](diffhunk://#diff-123583aa72f9d495c4492edc27c5bdc52f61e95fb358cff843d238f36818d5c7R341)
* Updated `useGroupContext` hook to include writer roles in the context data in `packages/app/hooks/useGroupContext.ts`.
* Enhanced `toClientChannel` function to map writer roles in `packages/shared/src/api/groupsApi.ts`. [[1]](diffhunk://#diff-5c683842d51c76808df0f83979922aa02c0b45a7ac0303aa948000552378191aR1480-R1490) [[2]](diffhunk://#diff-5c683842d51c76808df0f83979922aa02c0b45a7ac0303aa948000552378191aR1500-R1501)
* Modified `updateChannel` function to handle writer roles and use the `upsertChannel` function for database updates in `packages/shared/src/store/channelActions.ts`. [[1]](diffhunk://#diff-d9e7a242a1239a7bd1fa45014fe300585d6fc78823f8e00981d38614025c77f5R101-R129) [[2]](diffhunk://#diff-d9e7a242a1239a7bd1fa45014fe300585d6fc78823f8e00981d38614025c77f5L134-R150)
* Refactored `getUnjoinedGroupChannels` to improve logging and streamline the filtering process for user roles in `packages/shared/src/db/queries.ts`.
* Added `upsertChannel` query to handle channel upserts, including reader and writer roles, in `packages/shared/src/db/queries.ts`.
* Improved `insertGroups` query to handle reader roles separately and added logging for better traceability in `packages/shared/src/db/queries.ts`. [[1]](diffhunk://#diff-efcb4cf7d0bc50169d5f497604e75d495ff54a546aaee16e0bf48a7841bab03dR472-R481) [[2]](diffhunk://#diff-efcb4cf7d0bc50169d5f497604e75d495ff54a546aaee16e0bf48a7841bab03dR498-R516)


jamesacklin:
Updated to react to the 'addChannel' event to avoid showing all users role-gated channels as they're being added.

https://github.com/user-attachments/assets/6d228af6-024a-4545-8338-1e53b85d4610

apps/tlon-web-new/src/mocks/groups.ts
```diff
@@ -335,6 +338,7 @@ export function createChannel(title: string) {
     added: 1657774188151,
     join: false,
     readers: [],
+    writers: [], // Add this required field
```
patosullivan, line 341:
```suggestion
    writers: [],
```

packages/shared/src/store/sync.ts
```diff
@@ -512,25 +523,40 @@ async function handleGroupUpdate(update: api.GroupUpdate) {
...
+      await db.setJoinedGroupChannels({
+        channelIds: update.ships.includes(currentUser) ? [update.groupId] : [],
+      });
+      api.queryClient.invalidateQueries({ queryKey: ['group'] });
+      api.queryClient.invalidateQueries({ queryKey: ['channels'] });
+      api.queryClient.invalidateQueries({ queryKey: ['unjoinedChannels'] });
       break;
-    case 'addChannel':
-      await db.insertChannels([update.channel]);
-      break;
```
patosullivan, line 531: Why are we not handling the `addChannel` case anymore?

packages/shared/src/store/sync.ts
```diff
@@ -393,6 +393,17 @@ export const syncPushNotificationsSetting = async (ctx?: SyncCtx) => {
...
+  await db.setJoinedGroupChannels({
+    channelIds: [channel.id],
+  });
+  const channelInDb = await db.getChannel({ id: channel.id });
+  // Only insert the channel if the user has permission to see it
+  if (channelInDb?.currentUserIsMember) {
+    await db.insertChannels([channel]);
+  }
+}
+
```
patosullivan, line 406: I'm trying to understand this one. Is this intended to handle both channel adds and updates? 

If we've got a new channel to add, we first call `setJoinedGroupChannels`, which updates an existing channel in the db, and then we call `getChannel`, before inserting the channel to the db? It seems like this would always fail for inserting new channels, since those first checks will fail.

If it's an update, why are we calling insertChannels rather than updateChannel?

patosullivan: I also think we should handle the permissions checks within the `updateChannel` query to avoid potential race conditions.

jamesacklin: I was attempting to insert all new channels through a permission check. I'll ping you when I get to a desktop, but I'll take a look at your suggestion first.

packages/shared/src/store/sync.ts
jamesacklin: My mistake! Added it back in.

packages/shared/src/store/sync.ts
jamesacklin: Alright-take a look again when you get a sec. I put everything for when we hear about during addChannel and updateChannel in queries (as well as the updateChannel action itself) through a new function (`upsertChannel`), which should avoid any inconsistencies. Same behavior as before.

We still use `insertChannels` elsewhere (for DM invites, creating group DMs, etc). Not sure if we want to clean that function up or consolidate, happy to hear thoughts.

latter-bolden:
@jamesacklin can you try on the latest commit?

Thinking it might be less risky to just resync the group when we hear about any of these changes. Avoids having to think about codepaths for patching our local view in place & should be fine on perf since these updates are infrequent.

jamesacklin:
Works on my machine!


https://github.com/user-attachments/assets/cc542562-2c2f-4b1a-b97c-931c28c542d8

latter-bolden (approved):

PR #4222, opened 2024-11-25
desktop: fix issue with ListItem counts getting occluded by the overflow buttons

patosullivan:
OTT

image

fixes TLON-3270


jamesacklin:
We should probably account for this in Ochre, if we don't already @dnbrwstr

jamesacklin (approved):

PR #4223, opened 2024-11-26
desktop: fix ChatList/ChannelList height/space weirdness

patosullivan:
edit: now also fixes TLON-3279
fixes TLON-3271

We're now using the new Text component to render text in the chat list, this fixes the line height issue that required setting a specific (larger) height for ListItem on web. Also updates the $label/l value to something less heavy (Dan's recommendation) and updates the margin/position of the overflow menu button (and the count pill next to it).

Looks much more "normal" now:
image

Also looks good on iOS:
Simulator Screenshot - iPhone 15 - default - 2024-11-26 at 15 03 16

and Android:
Screenshot_1732655065

### DIFF
--- a/packages/ui/src/components/GalleryPost/GalleryPost.tsx
+++ b/packages/ui/src/components/GalleryPost/GalleryPost.tsx
@@ -17,6 +17,7 @@ import {
   PostContent,
   usePostContent,
 } from '../PostContent/contentUtils';
+import Pressable from '../Pressable';
 import { SendPostRetrySheet } from '../SendPostRetrySheet';
 import { Text } from '../TextV2';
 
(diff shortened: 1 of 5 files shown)


jamesacklin:
Indistinguishable on mobile from the native components. Nice work!

patosullivan:
Also tacked on TLON-3279, I missed a spot w/pressables apparently.

dnbrwstr (changes requested):
ChatList changes look good!

Let's skip the ListItem padding changes for now -- again, our lives are going to be a lot easier if we hold off on conditional rendering as much as possible until we get parity.

jamesacklin (approved):

PR #4227, opened 2024-11-27
desktop: fix routing/crash issues in settings, add contacts, add themes

patosullivan:
fixes TLON-3280
fixes TLON-3283
fixes TLON-3284

The crash in the AppInfoScreen was caused by the use of `getCurrentUser` in that screen and the import of the native version of useCurrentUser in the useTelemetry hook.

Updated desktop to now point to the Contacts screen when pressing the profile avatar in the nav, and the settings are now behind the cog like they are on mobile.

I also made desktop work with our themes (and made sure we could navigate to the ThemeScreen on desktop), including adding `backgroundColor` to a few places where we needed it.


dnbrwstr (approved):
Nice, lgtm!

PR #4230, opened 2024-11-27
contacts: quick fixes

jamesacklin:
Omits the current user from the suggested contacts query. Turns out, if you DM with yourself, you're eligible as a contact!

Furthermore, our condition for showing "You" vs "Add" vs just a caret in the ContactListItem endContent was a little shaky. Made that a little more explicit.

Things are looking better now IMO:
image


latter-bolden (approved):
Good catch! Looks good

PR #4231, opened 2024-11-27
native: + button on contacts screen for quick/bulk adding

latter-bolden:
OTT

Fixes TLON-3263


patosullivan (approved):
Looks good. Could we add it to ProfileScreenNavigator as well so it'll work on web?

latter-bolden:
Good callout, thanks! Need to get better at remembering that kind of thing

jamesacklin (approved):
Conditionally approved pending displaying @ps in lieu of nicknames.

image

We should also fix the color-switching of sigils, but feel free to tack that on as a separate PR (if you have time before EOD—otherwise let's leave it for next week).

image

latter-bolden:
@jamesacklin can you pull latest and check those two issues?

PR #4232, opened 2024-11-27
Fix usage of react-native-skia for web

davidisaaclee:
Looks like [`makeImageSnapshotAsync`](https://github.com/tloncorp/tlon-apps/blob/d3c1586ecbd31c958056e68c57ba05965de6542c/packages/ui/src/components/DrawingInput/DrawingInput.tsx#L185) is not implemented on web, which makes DrawingInput kind of useless. This is definitely possible to implement.

Not sure if this is the worth the mess, especially with the missing features (and the fact that we're not trying to ship drawing input in the custom channels framework). The alternative would be to rip out `DrawingInput` so that we can merge the custom channels branch without failing CI.

You can see what other features are missing [here](https://shopify.github.io/react-native-skia/docs/getting-started/web/#unsupported-features) and [in this file](https://github.com/Shopify/react-native-skia/blob/b6a848f55e50470292d16a40cbadf01ca496d37f/packages/skia/src/skia/web/JsiSkia.ts)

supposedly unsupported:
    PathEffectFactory.MakeSum()
    PathEffectFactory.MakeCompose()
    PathFactory.MakeFromText()
    ShaderFilter

---
## Context
- `@shopify/react-native-skia` does not work out of the gate on web
- [They have support for web](https://shopify.github.io/react-native-skia/docs/getting-started/web/)
- We can't follow those instructions exactly, since we aren't using Expo Web nor Metro(?) for web
## Summary
- Skia views will be powered by `canvaskit-wasm` on web
- `canvaskit-wasm`'s .wasm file is placed in the web app's `public/` directory (automatically updated on `postinstall`)
- The Skia-dependent view `DrawingInput` will be lazily-loaded on web, which avoids loading the wasm binary on web app launch
- The shader feature of `DrawingInput` is disabled (at callsite) on web, since runtime shaders are not implemented on web

https://github.com/user-attachments/assets/fa3abc99-d53f-4969-85bd-ec6859f5f53d


davidisaaclee:
CI is failing, but running the same command locally succeeds for me :(

davidisaaclee:
- I did [4c9e43e](https://github.com/tloncorp/tlon-apps/pull/4232/commits/4c9e43e52989ab1a2d0aa340cba19f16bfd86cc0) to see if removing a circular dependency would fix, but the tests still fail.
- `dpdm --transform --no-warning --no-tree  --context packages/shared/src packages/shared/src/api/postsApi.test.ts` shows no other circular dependencies (which could be wrong).
- I'm going to stop trying to fix this for the moment, because I'm not sure we even want to merge this?

@dnbrwstr @patosullivan Do you think we should try to merge this and get Skia into the app (which requires these workarounds and possibly more to fix the weird CI issue), or should we do this later and rip pictochat stuff out for now?

davidisaaclee:
sounds good, thanks

PR #4236, opened 2024-12-02
hooks: remaining cleanup

arthyn:
This adds some fixes/cleanup suggested by @mikolajpp after tloncorp/tlon-apps#4165 was merged

PR Checklist


desk/sur/hooks.hoon
```diff
@@ -23,62 +25,67 @@
...
+      hook=id-hook
       data=vase
       fires-at=time
   ==
-::
+::  $config: configuration data for a hook instance
 +$  config  (map @t *)
+::
+::  $action: what we can do with a hook
 +$  action
```
mikolajpp, line 56: Perhaps we should already follow the convention here and elsewhere of `a-*` for actions and `r-*` for responses?
```suggestion
+$  a-hook
```

mikolajpp: This can wait, but eventually when hooks are integrated we will have to do the renaming.

desk/sur/hooks.hoon
```diff
@@ -23,62 +25,67 @@
...
+      [%edit id=id-hook name=(unit @t) src=(unit @t) meta=(unit data:m)]
+      [%del id=id-hook]
+      [%order =nest seq=(list id-hook)]
+      [%config id=id-hook =nest =config]
+      [%cron id=id-hook =origin schedule=$@(@dr schedule) =config]
+      [%rest id=id-hook =origin]
   ==
+::
+::  $response: the result of an action on a hook
 +$  response
```
mikolajpp, line 67: If we want to follow the convention already:
```suggestion
+$  r-hook
```

desk/lib/hooks-json.hoon
```diff
@@ -93,11 +93,11 @@
       %gone  (id id.r)
       %order  (order-rsp +.r)
       %config  (config-rsp +.r)
-      %wait  (wait-rsp +.r)
+      %cron  (cron-rsp +.r)
       %rest  (rest-rsp +.r)
     ==
   ++  set-rsp
```
mikolajpp, line 99: If by convention we rename hook response to `$r-hook`, this function and others below would be adjusted accordingly
```suggestion
  ++  r-set
```

mikolajpp (commented):
A few more nitpicks regarding JSON naming conventions we use elsewhere, and a consideration whether we should follow action/response naming convention used in chanels.

desk/sur/hooks.hoon
arthyn: yeah I will leave this for later when we move hooks into channel actions

desk/lib/hooks-json.hoon
arthyn: will leave for later

desk/sur/hooks.hoon
arthyn: will leave for later

desk/app/channels-server.hoon
```diff
@@ -360,16 +360,16 @@
       =/  seq
         %+  skim
           seq.action
-        |=  =id:h
-        (~(has by hooks.hooks) id)
+        ~(has by hooks.hooks)
       =.  order.hooks  (~(put by order.hooks) nest.action seq)
-      (give-hook-response %order nest.action seq)
+      =/  =response:h  [%order nest.action seq]
+      (give %fact ~[/v0/hooks] hook-response-0+!>(response))
```
Fang-, line 366: This risks diverging from `+ho-give-response`. It might've been fine to keep a `+give-hook-response` at the top level here if `+ho-give-response` was written to invoke that instead of doing a `+give` (which turns into a top-level `+give` modification of `cor` anyway).

Not a huge deal, I'm not opposed to hand-written cards, but this code may get more complex in the presence of additional protocol versions.

desk/app/channels-server.hoon
```diff
@@ -1134,116 +1113,118 @@
...
+    ?~  job=(~(get by u.cron) origin)  cor
     ::  ignore premature fires
-    ?:  (lth now.bowl next.schedule.u.cron)  cor
-    =.  next.schedule.u.cron
+    ?:  (lth now.bowl next.schedule.u.job)  cor
+    =.  next.schedule.u.job
+      ::
       ::  we don't want to run the cron for every iteration it would
       ::  have run 'offline', so we check here to make sure that the
       ::  next fire time is in the future
```
Fang-, line 1191: I'll note for the record that upright flag comments (with the spacer at the bottom) generally have the preference. Omitting spacers isn't canonical style but sneaks in often enough that I'll let it be. Spacer exclusively at the top is strange in my eyes.

Fang- (approved):
Some comments for your consideration, but looks fine either way. Approved!

desk/app/channels-server.hoon
arthyn, line 1191:
```suggestion
      ::  we don't want to run the cron for every iteration it would
      ::  have run 'offline', so we check here to make sure that the
      ::  next fire time is in the future
      ::
```

desk/app/channels-server.hoon
arthyn: yeah I will leave for now, but yeah we should probably have an overarching `give-hook-response`

mikolajpp (approved):
Looks good to go! Many thanks for the improvements.

PR #4237, opened 2024-12-02
ProfileStatusSheet: add send button

jamesacklin:
Adds a send button to the ProfileStatusSheet and makes the input a single line.

Screenshot 2024-12-02 at 3 07 03 PM

### DIFF
--- a/packages/ui/src/components/ProfileStatusSheet.tsx
+++ b/packages/ui/src/components/ProfileStatusSheet.tsx
@@ -86,7 +86,7 @@ export default function ProfileStatusSheet({
               hero
               onPress={handleSave}
               disabled={!isValid}
-              padding={11.5}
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/ProfileStatusSheet.tsx
```diff
@@ -57,27 +59,38 @@ export default function ProfileStatusSheet({
...
+                  value: 50,
+                  message: 'Your status is limited to 50 characters',
+                },
+              }}
+            />
+            <Button
+              hero
+              onPress={handleSave}
+              disabled={!isValid}
+              padding={11.5}
```
jamesacklin, line 89: Not stoked about this, but I had to do this to make the height of the button match the height of the field.

latter-bolden (approved):
Works for me!

PR #4239, opened 2024-12-02
channels: more metadata work

arthyn:
This was a request from @davidisaaclee. We add the new metadata field to the create payload so that we can immediately pass custom channels data when creating a channel. Also adds the types from tloncorp/tlon-apps#4109 into the channels file.

PR Checklist


arthyn:
Notably the frontend type changes here need to be fully realized across the codebase before we can deploy.

davidisaaclee:
Thanks for this - it looks like my other request from last night got lost: We need an unstructured JSON config dict in each of content renderer / draft input / collection, like this diff: https://gist.github.com/davidisaaclee/097c480318d6a9a8367faaaf8c447360

I don't understand why `Stringified` is needed - and if we truly do need to stringify, why can't we do something like JS's JSON.stringify (instead of "_ value" wrapping)? (Not 100% sure that I'm reading this right.)

mikolajpp:
@arthyn I think there is a problem with state migration in %channels-server. The already deployed hooks increased the state number to `%7` already, so the type changes in `release-channel-updates` branch have to go with `%8`. I have integrated the hooks stuff to this branch already, perhaps you can take a look and rebase your changes here.

arthyn:
@davidisaaclee yeah `ChannelMetadataSchemaV1` was just a suggestion based on some old conversations we had, but it can be whatever we want it to be because it's just an open field on the backend. The reason I have stringified here is because the field on the backend is just plaintext or null. So I thought having a more type safe way of doing the json parse/stringify might be nice. we could store it as an actual JSON type hoon-side I just wasn't sure we would want that since it incurs extra structure that we're not really using on the backend.

davidisaaclee:
great, I can work with that! maybe we could just change the metadata type to an opaque nominal type (TS hack for this [here](https://www.typescriptlang.org/play/?#example/nominal-typing)) - I'm afk now but I can make an inline suggestion comment later

mikolajpp (approved):
The Hoon part looks good. Typescript part should have another look by someone else to be sure. Thanks!

One question: is adjusting the channel metadata after creation not in the picture? I can only see it supplied in `%create` poke. Do we need a `%meta` command in `$c-channel` as well?

arthyn:
@mikolajpp that was already there in the original meta PR, we just missed creation so I added it here

davidisaaclee:
talked with Hunter offline, TS looks good – [here's the change I mentioned in my comment](https://github.com/tloncorp/tlon-apps/blob/f8b4f2cafd242378adcfbca87791783c9c65763f/packages/shared/src/urbit/channel.ts#L338-L346) but it's going to be messy to propose as a commit to this and shouldn't block, so I'll just fix forward once this PR is merged

PR #4240, opened 2024-12-03
contacts: use contacts for populating contact book list, display @p if nickname set

latter-bolden:
OTT

Fixes TLON-3293
Fixes TLON-3299

### DIFF
--- a/packages/ui/src/components/ContactBook.tsx
+++ b/packages/ui/src/components/ContactBook.tsx
@@ -48,10 +48,18 @@ export function ContactBook({
   width?: number;
 }) {
   const contacts = useContacts();
+  const contactsForBook = useMemo(() => {
+    const markedContacts =
+      contacts?.filter((contact) => contact.isContact) ?? [];
+    if (markedContacts.length) {
+      return markedContacts;
+    }
+    return contacts;
+  }, [contacts]);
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/ContactBook.tsx
```diff
@@ -48,10 +48,18 @@ export function ContactBook({
...
+    const markedContacts =
+      contacts?.filter((contact) => contact.isContact) ?? [];
+    if (markedContacts.length) {
+      return markedContacts;
+    }
+    return contacts;
+  }, [contacts]);
   const immutableSet = useMemo(() => new Set(immutableIds), [immutableIds]);
   const contactsIndex = useContactIndex();
   const segmentedContacts = useAlphabeticallySegmentedContacts(
```
jamesacklin, line 61: Should we use this on the contact screen itself?

jamesacklin (approved):

packages/ui/src/components/ContactBook.tsx
latter-bolden: I think it would be weird UX to show the list of peers there as a fallback. 

If we wanted to improve the _I have no contacts or contact suggestions_ state, I think something like what you had in the mocks with a visual hero element + description would be more valuable

PR #4241, opened 2024-12-03
web: remove persist client

dnbrwstr:
This PR disables persisting react-query state in the the new web client. Immediately improves performance pretty dramatically.


patosullivan (approved):
lgtm!

PR #4244, opened 2024-12-03
native: fix contact activity showing up in feed

latter-bolden:
OTT

Fixes TLON-3345 by always permitting contact update events to be included while bundling (irrespective of notify status).

Also updates the logic to include contact events in separate "activity bundles". Since we don't roll them up on the client, we need to treat them as distinct sources to avoid misrepresenting the timeline (similar to how we treat mentions).


desk/app/activity.hoon
```diff
@@ -590,7 +590,7 @@
   ::  we only care about posts/replies events that are notified, and we
   ::  don't want to include events from sources whose latest event is
   ::  after the start so we always get "new" sources when paging
-  ?.  ?&  notified.event
+  ?.  ?&  ?|(notified.event ?=(%contact -<.event))
```
mikolajpp, line 593: Seems like this circumvents the intended logic, but @arthyn approved this for now.

desk/app/activity.hoon
```diff
@@ -613,6 +613,14 @@
   ?^  mention
     :-  sources.acc
     [(sub limit.acc 1) (snoc happenings.acc u.mention) collapsed.acc]
+  =/  contact-bundle=(unit activity-bundle:a)
+    ?.  ?=(%all type)  ~
+    =/  is-contact-event  ?=(%contact -<.event)
+    ?.  is-contact-event  ~
```
mikolajpp, line 619: This should be written as one line as the variable is not used anywhere
```suggestion
    ?.  ?=(%contact -<.event)  ~
```

mikolajpp (changes requested):
One nit, otherwise is good to go.

desk/app/activity.hoon
arthyn: yeah I think this is an ok compromise for now. what we really want are volume settings related to "where" an activity item should show up like unreads, feed, push notifications, etc. we had talked about it when first building activity, but it had already gone long so we decided to go with what we have.

PR #4245, opened 2024-12-04
native: find contact suggestions

latter-bolden:
Some simplistic logic for finding contact suggestions, focuses on generating a handful for new signups who came on via invite.

Fixes TLON-3286


desk/app/groups-ui.hoon
```diff
@@ -227,6 +237,12 @@
     =.  hidden-contact-suggestions
       (~(put in hidden-contact-suggestions) ship)
     cor
+  ::
+      %ui-add-contact-suggestions
+    =+  ship-list=!<((list @p) vase)
+    =.  manual-contact-suggestions
+      (~(uni in manual-contact-suggestions) (sy ship-list))
```
Fang-, line 244: `+gas:in` exists for adding a list of entries into a set. Should be completely equivalent, except more idiomatic and maybe very slightly more performant.

```suggestion
    =.  manual-contact-suggestions
      (~(gas in manual-contact-suggestions) ship-list)
```

desk/app/groups-ui.hoon
```diff
@@ -288,6 +304,8 @@
   =?  suggestions  pals-running
     =+  .^(targets=(set ship) (scry %gx %pals /targets/noun))
     (~(uni in suggestions) targets)
+  =.  suggestions
+    (~(uni in suggestions) manual-contact-suggestions)
```
Fang-, line 308: Alternatively you can change the `=|  suggestions=(set ship)` from above into `=/  suggestions=(set ship)  manual-contact-suggestions`. It's fine either way.

Fang- (commented):
Very small hoon nits, but backend changes look fine otherwise! (:

packages/shared/src/db/queries.ts
```diff
@@ -153,6 +153,32 @@ export const getGroupPreviews = createReadQuery(
   ['groups']
 );
 
+export const getJoinedGroupsCount = createReadQuery(
+  'getJoinedGroupCount',
+  async (ctx: QueryCtx) => {
+    const joinedGroups = await ctx.db.query.groups.findMany({
+      where: eq($groups.currentUserIsMember, true),
```
dnbrwstr, line 160: It'd be more efficient to just grab the count directly, rather than returning the full results array -- check out `getContactsCount`

packages/shared/src/db/queries.ts
```diff
@@ -153,6 +153,32 @@ export const getGroupPreviews = createReadQuery(
...
+    });
+    return joinedGroups.length;
+  },
+  ['groups']
+);
+
+export const getGroupsWithMemberThreshold = createReadQuery(
+  'getGroupsWithMemberThreshold',
+  async (threshold: number, ctx: QueryCtx) => {
```
dnbrwstr, line 169: This could also be done a lot more efficiently -- check out `unreadCounts` filtering in `getGroups`.

desk/app/groups-ui.hoon
latter-bolden: Ahh, good one to know!

apps/tlon-mobile/src/hooks/useFindContactSuggestions.ts
dnbrwstr, line 1: This file probably belongs in the `app` package.

apps/tlon-mobile/src/components/AuthenticatedApp.tsx
```diff
@@ -15,6 +15,7 @@ import { AppStateStatus } from 'react-native';
 
 import { useCheckAppUpdated } from '../hooks/analytics';
 import { useDeepLinkListener } from '../hooks/useDeepLinkListener';
+import useFindContactSuggestions from '../hooks/useFindContactSuggestions';
```
dnbrwstr, line 18: Would it be possible to do this initialization in the `app` or `shared` package? If not we should duplicate the logic in web.

desk/app/groups-ui.hoon
latter-bolden: That feels cleaner! Switched it over

packages/shared/src/store/contactActions.ts
```diff
@@ -71,6 +72,145 @@ export async function removeContactSuggestion(contactId: string) {
...
+  const optimisticUpdates = toUpdate.map((contact) =>
+    db.updateContact({ id: contact.id, isContactSuggestion: true })
+  );
+  await Promise.all(optimisticUpdates);
+
+  try {
+    await api.addContactSuggestions(contactIds);
+  } catch (e) {
+    // Rollback the update
+    const rolbacks = toUpdate.map((contact) =>
```
dnbrwstr, line 90:
```suggestion
    const rollbacks = toUpdate.map((contact) =>
```

packages/shared/src/store/contactActions.ts
```diff
@@ -71,6 +72,145 @@ export async function removeContactSuggestion(contactId: string) {
...
+      db.updateContact({ id: contact.id, isContactSuggestion: false })
+    );
+    await Promise.all(rolbacks);
```
dnbrwstr, line 93:
```suggestion
    await Promise.all(rollbacks);
```

packages/shared/src/store/contactActions.ts
```diff
@@ -71,6 +72,145 @@ export async function removeContactSuggestion(contactId: string) {
   }
 }
 
+export async function addContactSuggestions(contactIds: string[]) {
+  // optimistic update
```
dnbrwstr, line 76: I would just skip persisting this -- adds complexity + doesn't really net us anything.

dnbrwstr (commented):
Looking good -- some nits here. Additionally:
- Would consider removing server-side persistence since unnecessary
- Could the logic for triggering updates be part of `sync`?

dnbrwstr (approved):

arthyn (approved):
lgtm

PR #4247, opened 2024-12-04
native: pre-release fix omnibus

latter-bolden:
- always show your own profile on _Contacts Screen_, even if we somehow don't have an entry
- ding the bell for contact update activity, even if it's non-notifying

a few more to come...

### DIFF
--- a/packages/ui/src/components/ContactsScreenView.tsx
+++ b/packages/ui/src/components/ContactsScreenView.tsx
@@ -40,11 +40,9 @@ export function ContactsScreenView(props: Props) {
   const sections = useMemo(() => {
     const result: Section[] = [];
 
-    if (userContact) {
-      result.push({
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/ContactsScreenView.tsx
```diff
@@ -40,11 +40,9 @@ export function ContactsScreenView(props: Props) {
   const sections = useMemo(() => {
     const result: Section[] = [];
 
-    if (userContact) {
-      result.push({
-        data: [userContact],
-      });
-    }
+    result.push({
+      data: [userContact ?? db.getFallbackContact(currentUserId)],
```
jamesacklin, line 44: Thanks

jamesacklin (approved):

PR #4251, opened 2024-12-04
desktop: navigation fixes

patosullivan:
fixes tlon-3307 - highlight current channel in the sidebar
fixes tlon-3344 - top level sidebar (home/bell/avatar) buttons should reset navigation
fixes tlon-3305 - should navigate user to last visited channel when navigating to a group

### DIFF
--- a/packages/app/features/top/ChannelScreen.tsx
+++ b/packages/app/features/top/ChannelScreen.tsx
@@ -74,9 +74,9 @@ export default function ChannelScreen(props: Props) {
         if (group) {
           // Update the last visited channel in the group so we can return to it
           // when we come back to the group
-          db.updateLastVisitedChannelInGroup({
-            groupId: group.id,
-            channelId: channel.id,
+          db.updateGroup({
+            id: group.id,
+            lastVisitedChannelId: channel.id,
           });
         }
       }
--- a/packages/app/features/top/ChannelScreen.tsx
+++ b/packages/app/features/top/ChannelScreen.tsx
@@ -71,6 +71,14 @@ export default function ChannelScreen(props: Props) {
         store.syncChannelThreadUnreads(channel.id, {
           priority: store.SyncPriority.High,
         });
+        if (group) {
+          // Update the last visited channel in the group so we can return to it
+          // when we come back to the group
+          db.updateLastVisitedChannelInGroup({
+            groupId: group.id,
+            channelId: channel.id,
+          });
+        }
       }
       // Mark the channel as visited when we unfocus/leave this screen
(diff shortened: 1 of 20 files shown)


packages/app/features/top/ChannelScreen.tsx
```diff
@@ -71,14 +71,22 @@ export default function ChannelScreen(props: Props) {
         store.syncChannelThreadUnreads(channel.id, {
           priority: store.SyncPriority.High,
         });
+        if (group) {
+          // Update the last visited channel in the group so we can return to it
+          // when we come back to the group
+          db.updateLastVisitedChannelInGroup({
```
dnbrwstr, line 77: Could this call be part of `markChannelVisited`?

packages/shared/src/db/queries.ts
```diff
@@ -646,6 +646,20 @@ export const deleteGroup = createWriteQuery(
   ['groups', 'channels']
 );
 
+export const updateLastVisitedChannelInGroup = createWriteQuery(
```
dnbrwstr, line 649: Could use `updateGroup` for this instead of adding a new query.

packages/app/features/top/ChatListScreen.tsx
```diff
@@ -45,8 +45,10 @@ export default function ChatListScreen(props: Props) {
 
 export function ChatListScreenView({
   previewGroupId,
+  focusedChannelId,
```
dnbrwstr, line 48: Could we create (or use an existing) navigation context that this is part of? Trying to avoid creating more prop drilling where possible.

dnbrwstr (approved):
Overall looks good! A couple nits.

packages/app/features/top/ChatListScreen.tsx
patosullivan: yeah, I'll stick it in a context.

packages/app/features/top/ChannelScreen.tsx
patosullivan: The issue with using markChannelVisited is that we only call it when the user navigates away from it, I think we should update last visited as soon as a user navigates into a channel to avoid any weirdness that might crop up if the cleanup function in useFocusEffect doesn't run for some reason.

dnbrwstr: Ok skip it -- actually I think the channel visitedAt is redundant in my branch anyway (it's only used for the welcome message, and I'm not going to hide the welcome message until there are actually posts), so I'll probably remove it.

PR #4252, opened 2024-12-05
ops: use deploy keys in all the places

arthyn:
I think we were just missing a few places

PR Checklist


latter-bolden (approved):
Good catch

PR #4253, opened 2024-12-05
flat group creation

dnbrwstr:
This pr simplifies group creation and allows us to use generated titles for untitled groups. Touches quite a lot, so any qa-ing would be appreciated.

Fixes TLON-2794


packages/ui/src/hooks/groupsSorters.ts
```diff
@@ -50,60 +51,29 @@ export function useAlphabeticallySegmentedGroups({
...
-
-  try {
-    const aTitleClean = a.id.split('/')[1].replace(/[~-]/g, '');
-    const bTitleClean = b.id.split('/')[1].replace(/[~-]/g, '');
-    return aTitleClean.localeCompare(bTitleClean);
-  } catch (e) {
-    return 0;
-  }
+function groupsSorter(a: GroupSegment, b: GroupSegment): number {
+  return a.sortable.localeCompare(b.sortable);
```
latter-bolden, line 78: doesn't seem like this would do what's expected?

packages/ui/src/utils/channelUtils.tsx
```diff
@@ -13,15 +14,15 @@ export function getChannelMemberName(
   if (disableNicknames) {
     return member.contactId;
   }
-  return member.contact?.nickname ?? member.contactId;
+  return member.contact?.nickname ?? formatUserId(member.contactId)?.display;
```
latter-bolden, line 17: we've run into null coalesce issues in these kinds of comparisons if nickname is set but it's an empty string — can we make sure we defend against that?

packages/ui/src/components/Channel/EmptyChannelNotice.tsx
```diff
@@ -13,33 +17,42 @@ export function EmptyChannelNotice({
   channel: db.Channel;
   userId: string;
 }) {
-  const isGroupAdmin = useIsAdmin(channel.groupId ?? '', userId);
+  const { onPressGroupMeta } = useChatOptions() ?? {};
```
latter-bolden, line 20: When do you expect `useChatOptions` to not return anything? Only when the provider is uninitialized?

Trying locally on a freshly created multi-dm, this seems to no-op

latter-bolden (commented):
Let's chat through some of the backcompat implications tomorrow

packages/ui/src/components/Avatar.tsx
```diff
@@ -110,9 +110,15 @@ export const GroupAvatar = React.memo(function GroupAvatarComponent({
...
+  const { disableNicknames } = useCalm();
+  const fallbackTitle = useMemo(() => {
+    return isGroupImageShim(model)
+      ? model.title
+      : utils.getGroupTitle(model, disableNicknames);
+  }, [disableNicknames, model]);
   const fallback = (
     <TextAvatar
-      text={model.title ?? model.id.replace('~', '')}
+      text={fallbackTitle ?? 'G'}
```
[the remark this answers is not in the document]
dnbrwstr: Lmao, got another suggestion?

packages/ui/src/components/Channel/EmptyChannelNotice.tsx
dnbrwstr: Ya, never, the hook should actually throw an error if that's the case -- good catch.

packages/ui/src/components/Avatar.tsx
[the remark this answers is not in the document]
jamesacklin: ~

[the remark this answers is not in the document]
latter-bolden: No that's fine lol it just made me giggle

jamesacklin:
A few things of note, ordered in reverse-severity (most urgent stuff first).

The "Customize" button on the initial channel creation view is inactive for quite some time, until I open the "..." menu and navigate back. Furthermore, long-pressing the group in the ChatList and selecting "Edit Group" does nothing.

https://github.com/user-attachments/assets/82a1ab03-810c-4cd9-87bb-b59d7b3d8f50

---

Both variants of the second step in the creation flow need some extra copy. Why am I seeing a list of people here? For the DM case, we need to add rightward-carets to show that choosing an option will navigate the user.

Screenshot 2024-12-11 at 3 34 05 PM

Screenshot 2024-12-11 at 3 36 32 PM

---

We are missing safe-area padding on the "Edit group metadata" screen and the inviting-members screen in the creation flow.

Screenshot 2024-12-11 at 3 35 02 PM

Screenshot 2024-12-11 at 3 34 05 PM

---

In the GroupOptions sheet only accessible from inside the channel, the "Mark as Read" option appears even though there are no messages in the group.

Screenshot 2024-12-11 at 3 36 07 PM


---

The initial bifurcated creation sheet has some text truncation issues in the description of the second item. I'd recommend a copy change.

Screenshot 2024-12-11 at 3 33 51 PM

dnbrwstr:
Re-requesting review; I've added a placeholder title for backcompat + worked through most of the ux changes.

The only thing I didn't add was right arrows for contacts in the new dm sheet -- looking around at other apps, that's usually used to indicate that the row will navigate within the sheet, not when selecting the row is a primary action.

jamesacklin:
> looking around at other apps, that's usually used to indicate that the row will navigate within the sheet, not when selecting the row is a primary action.

Rationale checks out, thank you chef

latter-bolden (approved):
Changes lgtm, let's get it in!

PR #4254, opened 2024-12-05
fix pending group display

dnbrwstr:
Ensure that group invitations + requests are show in chat list. Also modifies labelling logic so that joining / error statuses take precedence over invite status. 

Fixes TLON-3362


latter-bolden (approved):

jamesacklin:
Thanks! Checking it out now.

jamesacklin (approved):
Screenshot 2024-12-05 at 2 40 32 PM

PR #4255, opened 2024-12-05
native: fix onboarding nickname

latter-bolden:
Looks like this was a bug in our key value logic. When relying on the `current` prop in `setValue(current: value) => nextValue`, there was a race condition if you called it rapidly causing stale `current` reads and thus overwrites of any previous pending updates. 

Just adds a basic lock around kv writes, but let me know if you guys can think of a cleaner way to do this?

Fixes TLON-3363


arthyn (approved):
seems fine to me

PR #4259, opened 2024-12-06
key value: add option to keep or clear storage items after logout

latter-bolden:
Today I got bit by a key value item not getting cleared after logout. Thus far we've been ad hoc adding reset calls to `useHandleLogout`, but I realized this should probably just be a property of `createStorageItem` itself.

Can now pass a `persistAfterLogout` flag in the config, defaults to false.


jamesacklin:
Nice, this will certainly help with on-device testing.

patosullivan (approved):
Looks good, good catch!

PR #4261, opened 2024-12-06
desktop: set a max height on the invite people modal

patosullivan:
Ensures that the modal fits within the height of the viewport, with some padding.

fixes TLON-3321


dnbrwstr (changes requested):
Looks good -- could we do this globally for all modals?

dnbrwstr (approved):
Lgtm!

PR #4262, opened 2024-12-06
images: provide default fallback for errored images

patosullivan:
fixes TLON-3323

The fact that we didn't provide a default fallback for errored images was most noticeable on web because:

a) the browsers shows a broken image icon
b) we just weren't accounting for broken images at all, no fallback was ever returned there.

This PR adds a new default image fallback that will be returned if an image fails to load. That fallback can be overridden by the `fallback` prop on Image.


latter-bolden (approved):
Looks good!

PR #4265, opened 2024-12-06
store: add provider for accessing within ui components

latter-bolden:
Adds a new `StoreProvider` that can be accessed via a `useStore` hook from within components. Allows us to stop passing down wrapped mutation methods and instead call them directly within relevant components while maintaining the ability to fall back to no-ops in tests/fixtures.

As discussed in dev sync, moved over `EditProfileScreenView` as an example.

Fixes TLON-3368


dnbrwstr (approved):
Great!!! Lgtm.

PR #4266, opened 2024-12-06
gallery: fix item size issues on desktop

patosullivan:
fixes tlon-3316

This makes grid items size correctly by setting their width based on the number of columns (50% for 2 columns, 100% for 1 column) while using aspectRatio to maintain the desired shape.

This was necessary because without an explicit width constraint, the grid items were sizing themselves based on their content's intrinsic dimensions (the full-size images), and even though we had aspectRatio set to create square shapes, the items were becoming too large and creating layout issues. By explicitly setting the width to a percentage of the container and letting aspectRatio determine the height, we make sure that grid items maintain the right proportions while fitting into the grid layout.

Setting the width of the items to 50% in two column mode also requires adding some right padding to the scroller container itself while in that mode, because otherwise the items on the right would be partially cut off (50% on the item itself doesn't account for padding/gap).

### DIFF
--- a/packages/ui/src/components/Channel/Scroller.tsx
+++ b/packages/ui/src/components/Channel/Scroller.tsx
@@ -32,7 +32,7 @@ import {
 } from 'react-native';
 import Animated from 'react-native-reanimated';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
-import { View, styled, useStyle, useTheme } from 'tamagui';
+import { View, getTokenValue, styled, useStyle, useTheme } from 'tamagui';
 
 import { RenderItemType } from '../../contexts/componentsKits';
 import { useLivePost } from '../../contexts/requests';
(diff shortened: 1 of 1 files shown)


jamesacklin:
Thanks. Does this affect the mobile client too? We've had an outstanding issue where odd-numbered rows have a 100%-wide item in them.

patosullivan:
> Thanks. Does this affect the mobile client too? We've had an outstanding issue where odd-numbered rows have a 100%-wide item in them.

Yeah, it'll handle the mobile client too. It should force 50% width on every item.

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -268,6 +271,7 @@ const Scroller = forwardRef(
             messageRef={activeMessageRefs.current[post.id]}
             dividersEnabled={collectionLayout.dividersEnabled}
             itemAspectRatio={collectionLayout.itemAspectRatio ?? undefined}
+            columnCount={collectionLayout.columnCount}
```
dnbrwstr, line 274: Needs to be added to deps

packages/ui/src/components/Channel/Scroller.tsx
```diff
@@ -201,13 +201,16 @@ const Scroller = forwardRef(
...
 
-    // Used to hide the scroller until we've found the anchor post.
     const style = useMemo(() => {
       return {
         backgroundColor: theme.background.val,
+        // Used to hide the scroller until we've found the anchor post.
         opacity: readyToDisplayPosts ? 1 : 0,
+        // Necessary to prevent content from flowing off the right side of the
+        // screen when the scroller is in two-column mode.
+        paddingRight: collectionLayoutType === 'grid' ? getTokenValue('$l') : 0,
```
dnbrwstr, line 211: Could this be added to `columnWrapperStyle` instead? Would be a little cleaner.

dnbrwstr (approved):
Lgtm to me -- just a couple nits.

PR #4267, opened 2024-12-06
profile: use new contacts type & endpoint

Fang-:
This was still using the old contacts types and scry endpoint. Here, we use the latest instead, and move contact data unpacking up to the top of the file while we're at it.

Resulting widgets confirmed to be exactly equivalent.

(Also sneaked in a small refactoring. Best to review per commit.)

### DIFF
--- a/desk/app/profile/widgets.hoon
+++ b/desk/app/profile/widgets.hoon
@@ -148,63 +148,41 @@
   :~
     ;style:"{(trip style)}"
   ::
-    ?~  cover
-      ;div.profile-without-header
-        ;div.profile-headline
-          ;+  ?^  avatar
-            ;img.profile-headline-avatar
-              =src  "{(trip u.avatar)}"
-              =alt  "Avatar";
-          =/  color=tape  ((x-co:^co 6) color)
-          ;div.profile-headline-avatar-sigil(style "background-color: #{color}")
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm!

desk/app/profile/widgets.hoon
```diff
@@ -1,21 +1,23 @@
...
-        /gx/(scot %p our.bowl)/contacts/(scot %da now.bowl)/all/contact-rolodex
-      ==
-  =/  =foreign-0  (~(gut by rolodex) our.bowl *foreign-0)
-  ?:  ?=([[@ ^] *] foreign-0)
-    `con.for.foreign-0
-  ~
+=+  .^  =contact:co
+      /gx/(scot %p our.bowl)/contacts/(scot %da now.bowl)/v1/self/contact-1
+    ==
+::NOTE  can't quite make a nice helper for this, wetness not wet enough...
```
mikolajpp, line 10: This pattern seems like a good candidate to include as a safe, non-empty variant of `+get` in the `+cy` engine from `/lib/contacts`. Making a mental note to revisit this when I work on new %contacts features.

The code itself is good.

mikolajpp (approved):
Looks good!

PR #4268, opened 2024-12-09
channels: implement custom posts support

mikolajpp:
post $essay is modified to support: (1) post path kind, (2) metadata, and (3) opaque post payload. post $memo is modified to allow custom authorship data
$bot-meta. 

Resolves TLON-3153


desk/lib/channel-utils.hoon
```diff
@@ -429,13 +429,35 @@
...
+++  essay-1
+  |=  =essay:c
+  ^-  essay:v7:old:c
+  *essay:v7:old:c
+::
+++  rev-essay-1
+  |=  essay=(rev:c essay:c)
+  ^-  (rev:c essay:v7:old:c)
+  [rev.essay *essay:v7:old:c]
+::
```
arthyn, line 455: seems like these need to be filled in

arthyn (changes requested):
everything looks good to me, just need the few gates filled in that are defaulting empty

desk/lib/channel-utils.hoon
mikolajpp: For now I have implemented down-conversion of existing post kinds, that in the new scheme correspond to kind paths `/chat`, `/chat/notice`, `/diary` and `/heap`. For posts that are not of this kind we just crash, which means such updates won't make it to an old client.

jamesacklin: Understood. As long as we maintain some kind of basic backcompat until we release the new desktop client, I'm OK with some signal degradation between clients. 

When you say crash—will the old client stop receiving updates to the post? E.g. edits, reactions, replies, etc

mikolajpp: > We can always hide channels that are not of those kind on the front-end of the old client.
To be sure we are on the same page here: this is about the post kind. A channel in principle can contain a mixture of different post kinds. 

There is also another https://github.com/tloncorp/tlon-apps/pull/4239 that introduces the channel meta field which carries arbitrary channel metadata.

desk/lib/dm.hoon
```diff
@@ -1,5 +1,5 @@
-/-  c=chat, d=channels
-/+  mp=mop-extensions
+/-  c=chat, d=channels, meta
+/+  mp=mop-extensions, cu=channel-utils
```
Fang-, line 2: Surprising that this file needs to be changed even though chat continues using the old types. Not sure I understand why.

And as far as I can tell, you don't actually use the `meta` import, do you?

desk/app/groups-ui.hoon
```diff
@@ -285,7 +301,7 @@
   ==
 ++  get-suggested-contacts
   =+  .^(chat-running=? (scry %gu %chat /$))
-  =|  suggestions=(set ship)
+  =/  suggestions=(set ship)  manual-contact-suggestions
```
Fang-, line 304: This guy seems out of place in this PR. (^:

Fang- (commented):
That's a _lot_ of author wrangling! Guess we really do use it all over the place...

Some questions below.

desk/sur/channels.hoon
```diff
@@ -98,13 +98,23 @@
...
-+$  essay  [memo =kind-data]
+::  $memo: post data
+::  .kind: post kind
+::  .meta: post metadata
+::  .blob: custom payload
+::
++$  essay
+  $:  memo
+      kind=path
+      meta=data:m
```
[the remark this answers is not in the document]
mikolajpp: For `/chat` content this is gonna be almost always empty, at least with existing usage patterns.
For `/diary` and `/heap`, `meta` now stores `title` and `image`. Previously, `$essay` had `$kind-data` which stores diary metadata without a unit, but heap has a unitized title field. 

We could make it into a unit without a problem. It seems more reasonable given the overhead on chat messages.

desk/sur/channels.hoon
```diff
@@ -437,7 +458,7 @@
 ::
 +$  r-pending
   $%  [%post =essay]
-      [%reply top=id-post =reply-meta =memo]
+      [%reply top=id-post =reply-meta =memo]  ::XX author present in both fields
```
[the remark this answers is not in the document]
mikolajpp: Indeed, only `$memo` has it. We also keep a `(set author)` in `reply-meta`, but this has different semantics.

desk/app/groups-ui.hoon
mikolajpp: Without this the PR generates merge conflicts. It is not great, I know, but in the end you still need to push those exact changes. I will be sure to push an explicit commit next time.

desk/sur/chat.hoon
```diff
@@ -59,8 +59,7 @@
   ==
 ::
 ::  $essay: a chat message with metadata
-+$  essay  [memo:d %chat =kind]
-::  $kind: whether or not the chat is a system message
```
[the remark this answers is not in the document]
mikolajpp: We definitely want to, it is just that this PR was supposed to land quickly. Since we are still at it, I might as well go and dig it.

We will also want to handle bot authors in activities, but this would require an overhaul of %activity types and we decided to postpone it for now.

desk/sur/channels.hoon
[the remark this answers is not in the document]
mikolajpp: `.meta` in $essay is now a unit.

mikolajpp:
This PR now includes a new %chat version that is compatible with new %channels types.

desk/app/chat.hoon
```diff
@@ -877,8 +1056,28 @@
     |=  [=ship text=cord]
     ^-  delta:writs:c
     =/  =story:d  ~[[%inline ~[[%ship ship] text]]]
-    =/  =memo:v7:old:d  [story our.bowl now.bowl]
-    [%add memo notice/~ `now.bowl]
+    =/  =memo:d  [story our.bowl now.bowl]
+    [%add [memo chat/notice/~ ~ ~] `now.bowl]
```
Fang-, line 1038: Same here as the other comment, `[%chat /notice]`.

Fang- (commented):
As discussed synchronously, some comments below. Thank you!

Fang- (approved):
Changes look good, I think this should be fine to go in now. Thank you!

PR #4270, opened 2024-12-09
desktop: set 50% minWidth for modals

patosullivan:
OTT, fixes TLON-3313


latter-bolden (changes requested):
Could we add a cap on here? 50% can be pretty large depending on the display

Maybe target 50% with explicit fixed min/maxes?

latter-bolden (approved):

PR #4272, opened 2024-12-09
invites: support personal invite links

latter-bolden:
Generate and display personal invite links that aren't tied to joining a specific group. Surfaced via cta header button on home screen

TODO
- fix env var juggling
- sort out backend quirks
- visual polish/copy pass

Fixes TLON-3370

### DIFF
--- a/packages/shared/src/logic/deeplinks.ts
+++ b/packages/shared/src/logic/deeplinks.ts
@@ -1,11 +1,6 @@
 import { ContentReference } from '../api';
 import { citeToPath } from '../urbit';
-import {
-  AppInvite,
-  DeepLinkMetadata,
-  getBranchLinkMeta,
-  isLureMeta,
-} from './branch';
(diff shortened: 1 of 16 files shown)


packages/shared/src/logic/deeplinks.ts
```diff
@@ -23,6 +19,7 @@ export async function getReferenceFromDeeplink({
   });
 
   if (linkMeta && typeof linkMeta === 'object') {
+    // TODO: handle personal invite links
```
arthyn, line 22: do you need to do this before deploy?

arthyn (approved):
lgtm

packages/shared/src/logic/deeplinks.ts
latter-bolden: Na, we can address later. Handling these pastes isn't critical, it'll bail when it sees the missing `invitedGroupId` anyhow

PR #4273, opened 2024-12-09
fix desktop chatlist jank

dnbrwstr:
This PR fixes initial chatlist animation / rendering jank by more accurately estimating item sizes pre-render.

The jank is a known issue for `FlashList` on web, if it continues to crop up we could switch to `FlatList` on desktop instead.

Fixes TLON-3332


patosullivan (approved):
Ran it locally, looks good!

PR #4274, opened 2024-12-10
ChannelListItem, channelUtils: fix bare patp display in ChatList

jamesacklin:
Corrects the nullish check pattern in channelUtils / getChannelMemberName. Tries to get the nickname from the contact, but if that's not available, it falls back to using the contactId (patp).

Also implements a "hasNickname" channelUtil to hide the duplicitous patp subtitle in ChannelListItems.

Here's a DM with someone who has a nickname set:
Screenshot 2024-12-09 at 7 21 31 PM

And without:
image


latter-bolden (approved):
Drats, there was another helper we missed. Good find

PR #4276, opened 2024-12-10
desktop: harden navigation between home/non-home screens and nested screens

patosullivan:
fixes tlon-3306

This adds some new hooks to our navigation utils, `useNavigateToChannel` and `useNavigateToPost`, that allow us to navigate from the ActivityScreen on desktop into the relevant nested screens within the HomeNavigator.

It also updates our navigation types to account for the fact that Channel in HomeNavigator is a Stack and not just a route, and updates Screen components where appropriate to use the new hooks.

This gets us a bit closer to the abstraction layer for navigation that we've been talking about.


dnbrwstr (approved):
Nice, lgtm!

PR #4278, opened 2024-12-12
fix message action height issue

patosullivan:
fixes tlon-3375

Our fixes for getting rid of the FlashList jank involved setting a definite height on ListItem of `$6xl`. This worked fine for most use cases, but not for our message actions. We just need to override the height in this case with `'auto'`.


dnbrwstr (approved):
Lgtm!

PR #4279, opened 2024-12-12
db: use triggers and change table for reactivity on native

patosullivan:
Fixes reactivity issue we were seeing on both platforms by:

1) fixing an issue with a field name in triggers
2) using the triggers and change table/polling method on native to get around the issues we found with op-sqlite's `updateHook`.

This also creates a new `BaseDb` class where shared code for both platforms can live.

I tested this out in the sim on native and watched the CPU usage in xcode, didn't notice any difference over the current baseline (which I found to be pretty bad on initial channel load, fwiw).

fixes tlon-3317


latter-bolden (approved):
Let's try it out!

jamesacklin:
Fire at will

PR #4280, opened 2024-12-12
native: hand off post and DM data to app from notification

davidisaaclee:
fixes TLON-2649 (lots of context in thread)

### Not included
This change only affects iOS – Android should only require (1) requesting activity-event (same as iOS), (2) passing the data (same as iOS), and (3) getting the expo-notifications callback to run (which docs say is possible).

Thread replies are not implemented here.

### Terms
- "Hand off" is what happens when the app grabs data from a notification payload and inserts it into the database
- Unconfirmed posts are posts from notification hand-off; they get confirmed via a subsequent sync. A post is unconfirmed if `synced_at = null`. An unconfirmed window is a window containing an unconfirmed post (which is usually not a row in the database, just an implicit single-post window)
  
### Changes
- iOS notification extension fetches the `activity-event` for each notification alongside the `hark-yarn`. We could move to only use the `activity-event` – I started doing that, but it requires a lot of client code to match our current feature set.
  - I needed to switch behavior on native side based on the content of `activity-event`, so I added Swift types, which was a lot of code! [Here's how I did it](https://linear.app/tlon/issue/TLON-2649/background-fetching-to-improve-channel-freshness-after-a-notif-is#comment-3cd429a4) – I ended up having to edit stuff by hand, so I didn't check in the workflow to generate types.
- For `dm-post` and `post` activities, the relevant payload is attached to the notification `userInfo` for the JS to consume
- Unconfirmeds are picked up from presented and tapped notifications in notification listener
- Unconfirmeds are marked in DB as `synced_at = null`, and are not part of any explicit post window
- `useChannelPosts` fetches unconfirmeds separately from the main `getChannelPosts`; always fetches all unconfirmeds for the channel, and inserts into query list based on post ordering
- `getChannelPosts` is programmed to only fetch posts in a single post window (i.e. the one containing the channel's cursor) to avoid rendering a timeline with gaps in it
  - if this is an unconfirmed window, then `getChannelPosts` will only fetch the single post, which would be a bad UX
  - To avoid using an unconfirmed window: if the cursor is an unconfirmed message, choose the closest confirmed window


apps/tlon-mobile/ios/Shared/Notifications/PushNotificationManager.swift
```diff
@@ -224,3 +64,71 @@ enum NotificationCategory: String {
...
+   // `mutableContent` should no longer be used; use `content` instead
+   if let msgIntent {
+     // donate intent here
+   }
+   ```
+   */
+  func render(to content: UNMutableNotificationContent) async -> (UNNotificationContent, INSendMessageIntent?)
+}
+
+extension Yarn: UNNotificationRenderable {
```
davidisaaclee, line 88: sorry about code movement here – I made this change when I was implementing activity event stuff so I could just adopt `UNNotificationRenderable` on `ActivityEvent`, then I backed out of that – figured it's a fine change to keep

apps/tlon-mobile/src/hooks/useNotificationListener.ts
```diff
@@ -233,3 +283,50 @@ export default function useNotificationListener() {
...
+        }
+      },
+      [handoffFromPresentedNotifications]
+    )
+  );
+
+  return handoffDataFrom;
+}
+
+function getReceivedAtFromId(postId: string) {
```
davidisaaclee, line 330: I copied this from another part of the codebase – I don't get it

packages/shared/src/db/client.ts
```diff
@@ -21,6 +21,22 @@ let clientInstance: AnySqliteDatabase | null = null;
...
+    const execSimple = (strings: TemplateStringsArray, ...values: any[]) => {
+      const result = exec(strings, ...values);
+      return Array(result.rows.length)
+        .fill(0)
+        .map((_, i) => result.rows.item(i));
+    };
+    Object.assign(global, {
+      __db: client,
+      __sql: execSimple,
+      __sqlRaw: exec,
```
davidisaaclee, line 37: you can use this in console like 

```
_sql`select * from posts where id = 'abc' limit ${1 + 3}`
```

packages/shared/src/db/queries.ts
```diff
@@ -1867,6 +1874,32 @@ export const setLeftGroups = createWriteQuery(
   ['groups', 'channels']
 );
 
+// Includes latest post as well as unconfirmed posts
+export const getUnconfirmedPosts = createReadQuery(
```
davidisaaclee, line 1878: why did I include latest post here?

- When doing `useChannelPosts`, we union these posts alongside the main query so that we always have unconfirmed posts, regardless of post window
- So it looks like `main query` + `unconfirmeds`
- naively, `unconfirmeds` does not include the latest post – this is because we fetch latest post for each channel on init, which means it is no longer unconfirmed!
- to get the expected behavior, just throw the latest post in as well

packages/shared/src/db/queries.ts
```diff
@@ -2151,7 +2210,60 @@ export const insertLatestPosts = createWriteQuery(
...
+    return withTransactionCtx(ctx, async (txCtx) => {
+      // insertPosts does multiple queries internally, so we need to wrap it in a transaction
+      await insertPosts(posts, txCtx);
+    });
+  },
+  ['posts']
+);
+
 async function insertPosts(posts: Post[], ctx: QueryCtx) {
+  // HACK: I can't get onConflictDoUpdate to work - manually manage conflicts.
```
davidisaaclee, line 2231: any ideas? even with `onConflictDoUpdate` configured in a pretty simple way, I had `UNIQUE` constraint errors when replacing the unconfirmed posts with confirmed ones.

I think the code was like

```ts
      .onConflictDoUpdate({
        target: $posts.id
        set: conflictUpdateSetAll($posts),
        // with or without
        setWhere: sql`excluded.synced_at IS NOT NULL OR synced_at IS NULL`
      });
```

apps/tlon-mobile/src/hooks/useNotificationListener.ts
```diff
@@ -65,12 +72,51 @@ function payloadFromNotification(
...
   // welcome to my validation library ;)
   if (payload.wer != null && payload.channelId != null) {
+    const postInfo = api.getPostInfoFromWer(payload.wer);
+    const handoffPost: db.Post | undefined = (() => {
+      const dmPost = payload.dmPost as ub.DmPostEvent['dm-post'] | undefined;
+      if (dmPost != null) {
+        // key.id looks like `dm/000.000.000.mor.eme.ssa.gei.dxx`
+        const id = dmPost.key.id.split('/')[1];
+        const receivedAt = getReceivedAtFromId(id);
+        return {
```
davidisaaclee, line 82: would definitely like a second look here to ensure I'm connecting the API resource to db.Post correctly

davidisaaclee:
Here's how I'm testing the change:

1. Apply patch to show confirmed/unconfirmed status in chat message view, and to slow down post fetches
2. Force-quit app on phone, and send more than one message to phone client
3. After receiving at least two notifications, open app
4. Check that all messages appear + are marked as unconfirmed, then resolve as confirmed after 5 seconds

https://github.com/user-attachments/assets/827161a1-fb12-4377-bb5b-121b601ccec8

<details>

<summary>Patch</summary>

```diff
diff --git a/apps/tlon-mobile/ios/Podfile.lock b/apps/tlon-mobile/ios/Podfile.lock
index 716c600b8..06412313e 100644
--- a/apps/tlon-mobile/ios/Podfile.lock
+++ b/apps/tlon-mobile/ios/Podfile.lock
@@ -1880,7 +1880,7 @@ SPEC CHECKSUMS:
   sqlite3: f163dbbb7aa3339ad8fc622782c2d9d7b72f7e9c
   tentap: 2cf2e387dd284bf867010eb7d0f91618fb35b673
   UMAppLoader: 5df85360d65cabaef544be5424ac64672e648482
-  Yoga: fb61b2337c7688c81a137e5560b3cbb515289f91
+  Yoga: 4dbfeceb9bb0f62899d0a53d37a1ddd58898d3f2
 
 PODFILE CHECKSUM: 0cb7a78e5777e69c86c1bf4bb5135fd660376dbe
 
diff --git a/packages/shared/src/api/postsApi.ts b/packages/shared/src/api/postsApi.ts
index 3b0168022..fada2ea3a 100644
--- a/packages/shared/src/api/postsApi.ts
+++ b/packages/shared/src/api/postsApi.ts
@@ -337,6 +337,8 @@ export interface GetChannelPostsResponse {
   totalPosts?: number;
 }
 
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
 export const getChannelPosts = async ({
   channelId,
   cursor,
@@ -344,6 +346,7 @@ export const getChannelPosts = async ({
   count = 20,
   includeReplies = false,
 }: GetChannelPostsOptions) => {
+  await sleep(5_000);
   const type = getChannelIdType(channelId);
   const app = type === 'channel' ? 'channels' : 'chat';
   const path = formatScryPath(
diff --git a/packages/shared/src/store/sync.ts b/packages/shared/src/store/sync.ts
index e3546cf02..6edce6dcb 100644
--- a/packages/shared/src/store/sync.ts
+++ b/packages/shared/src/store/sync.ts
@@ -280,6 +280,8 @@ export async function syncPostReference(options: {
   });
 }
 
+const sleep = (ms: number) => new Promise((res) => setTimeout(res, ms));
+
 export async function syncUpdatedPosts(
   options: GetChangedPostsOptions,
   ctx?: SyncCtx
@@ -288,6 +290,7 @@ export async function syncUpdatedPosts(
     'syncing updated posts',
     runIfDev(() => JSON.stringify(options))
   );
+  await sleep(5000);
   const response = await syncQueue.add('syncUpdatedPosts', ctx, async () =>
     api.getChangedPosts(options)
   );
diff --git a/packages/shared/src/store/useChannelPosts.ts b/packages/shared/src/store/useChannelPosts.ts
index e040c7b69..9742156b1 100644
--- a/packages/shared/src/store/useChannelPosts.ts
+++ b/packages/shared/src/store/useChannelPosts.ts
@@ -39,6 +39,8 @@ type UseChanelPostsParams = UseChannelPostsPageParams & {
   hasCachedNewest?: boolean;
 };
 
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
 export const clearChannelPostsQueries = () => {
   queryClient.invalidateQueries({ queryKey: ['channelPosts'] });
 };
@@ -71,6 +73,7 @@ export const useChannelPosts = (options: UseChanelPostsParams) => {
         queryOptions.mode === 'newest' &&
         !options.hasCachedNewest
       ) {
+        await sleep(5000);
         await sync.syncPosts(queryOptions, { priority: SyncPriority.High });
       }
       const cached = await db.getChannelPosts(queryOptions);
@@ -80,6 +83,8 @@ export const useChannelPosts = (options: UseChanelPostsParams) => {
       }
 
       postsLogger.log('no posts found in database, loading from api...');
+
+      await sleep(5000);
       const res = await sync.syncPosts(
         {
           ...queryOptions,
diff --git a/packages/ui/src/components/ChatMessage/ChatMessage.tsx b/packages/ui/src/components/ChatMessage/ChatMessage.tsx
index 4b9133b16..6148bda9b 100644
--- a/packages/ui/src/components/ChatMessage/ChatMessage.tsx
+++ b/packages/ui/src/components/ChatMessage/ChatMessage.tsx
@@ -171,6 +171,7 @@ const ChatMessage = ({
             isNotice={post.type === 'notice'}
             onPressImage={handleImagePressed}
           />
+          <Text>{post.syncedAt == null ? 'Unconfirmed' : 'Confirmed'}</Text>
         </View>
 
         <ReactionsDisplay
```

</details>

davidisaaclee:
I'm seeing now that this doesn't work for posts in channels. The issue is that [`getPostInfoFromWer`](https://github.com/tloncorp/tlon-apps/blob/eb4f13cf6d63773cc9ef16c30b4b0b59823a7515/packages/shared/src/api/harkApi.ts#L6) doesn't get info for non-reply channel posts; and even for replies, it doesn't get a real `authorId`. It's not clear how to get an author ID without another API request – we don't have it in the payload so far (i.e. not in hark-yarn nor in activity-event for posts, unless it's in the `wer`?)

jamesacklin:
I think we need to address TLON-3387 before merging this, which I would very much like to do. Do you need @patosullivan's help/pointers?

davidisaaclee:
@jamesacklin nope, working on it

davidisaaclee:
@.reviewers, [b621506](https://github.com/tloncorp/tlon-apps/pull/4280/commits/b621506c1c8a96dda820c6198e6e6e08ac0f5c02) should fix TLON-3387 – explanation comment inline.

Here's proof:
https://github.com/user-attachments/assets/b38fcc06-178a-46dd-bda5-541f7ab327a4

davidisaaclee:
looks like I have some real test errors :( fixing now

davidisaaclee:
tests passed for me locally with latest change

jamesacklin:
Thank you @davidisaaclee!

apps/tlon-mobile/src/hooks/useNotificationListener.ts
dnbrwstr: Looks right to my eye -- would it make sense to move this logic to modelBuilders?

apps/tlon-mobile/src/hooks/useNotificationListener.ts
```diff
@@ -233,3 +283,50 @@ export default function useNotificationListener() {
...
+    const handoffPosts = notifications.flatMap((notification) => {
+      const data = payloadFromNotification(notification);
+      if (data == null || data.type === 'unrecognized' || data.post == null) {
+        return [];
+      }
+      return [data.post];
+    });
+
+    if (handoffPosts.length > 0) {
+      console.log('Handing off', handoffPosts);
```
dnbrwstr, line 298: stray log

packages/shared/src/db/queries.ts
dnbrwstr: I'm not sure what the best solution is, but the likely issue is that afaict drizzle doesn't support multiple `onConflictDoUpdate` clauses (open issue here: https://github.com/drizzle-team/drizzle-orm/issues/2276)

packages/shared/src/store/useChannelPosts.ts
```diff
@@ -177,6 +190,27 @@ export const useChannelPosts = (options: UseChanelPostsParams) => {
     return newestQueryPostId ? [...newerPosts, ...dedupedQueryPosts] : newPosts;
   }, [query.data, query.hasPreviousPage, newPosts]);
 
+  const rawPosts = useMemo(() => {
```
dnbrwstr, line 193: Feels like this could be combined with the above dedupe logic? But maybe I'm missing something.

dnbrwstr (approved):
Lgtm overall! Added a few comments

packages/shared/src/store/useChannelPosts.ts
davidisaaclee: I can combine the two `useMemo`s, which might save cognitive load

but if you're saying that we could merge `newPosts` with `unconfirmedPosts` and use the same dedupe logic, then I think this early exit breaks it: https://github.com/tloncorp/tlon-apps/blob/26429d11f34aa476bb0bcbe2c073bc9e0da1a457/packages/shared/src/store/useChannelPosts.ts#L172-L174
My reading of that code is that we only want to show posts coming in through subscription once the channel is completely backfilled; which conflicts with unconfirmed behavior, which we want to show as soon as possible.

will add a comment about this – I don't see a clean way to unify these, lmk if you do

davidisaaclee: https://github.com/tloncorp/tlon-apps/blob/859bd4f92bdcd8d78859455dcdc76b87a9d22fb4/packages/shared/src/store/useChannelPosts.ts#L200-L208

apps/tlon-mobile/src/hooks/useNotificationListener.ts
davidisaaclee: [3af4e2d](https://github.com/tloncorp/tlon-apps/pull/4280/commits/3af4e2d958dc584dd4bc459ce2cca266de6f756c)

apps/tlon-mobile/src/hooks/useNotificationListener.ts
davidisaaclee: done! [9ad3193](https://github.com/tloncorp/tlon-apps/pull/4280/commits/9ad31934220b8183f240003a6eab9cae44c9665c)

packages/shared/src/db/queries.ts
davidisaaclee: very sad, added reference to that issue inline: [Add Drizzle issue reference](https://github.com/tloncorp/tlon-apps/pull/4280/commits/db48efb5607057185addc58694ff57d81b9b1410)

davidisaaclee:
@dnbrwstr addressed all comments – will merge on CI success unless you wanna block to look

PR #4281, opened 2024-12-13
profile: read contacts fields safely

Fang-:
types. It checked for the tags on values found in the map. However, it's possible for values to be null, in which case trying to grab the head would crash at runtime.

Here, we update the typecheck to check on the full value as opposed to just the head.

Since this crashing code might have run in response to a contacts subscription update, killing the subscription, we add a bit of logic to +on-load to ensure that the contacts subscription gets set up again if needed.


arthyn (approved):
lgtm

arthyn:
we can get it out quicker if you merge to staging

Fang-:
The change that was dangerous was in #4267, which is only on develop, hasn't made it to staging yet.

PR #4287, opened 2024-12-13
channels: add copy as template flow

arthyn:
This adds the ability to copy the way any channel currently has hooks setup so that they can run in a new channel of our choosing. This manifests as an entry in the channel options sheet and brings you to a screen where you can see what hooks are running, give the new channel a name, and select what group to add it to. Once confirmed, it brings you immediately in the channel.

Required backend changes to support grabbing templates from other ships and setting them up within our own %channels-server. 

Fixes TLON-3267

PR Checklist


desk/app/channels.hoon
```diff
@@ -136,6 +136,9 @@
   =?  old  ?=(%6 -.old)  (state-6-to-7 old)
   ?>  ?=(%7 -.old)
   =.  state  old
+  =/  =wire  /v1/hooks/preview/chat/~simtyc-tirter-nocsyx-lassul/test
+  =/  =dock  [~simtyc-tirter-nocsyx-lassul %channels-server]
+  =.  cor  (emit %pass wire %agent dock %leave ~)
```
Fang-, line 141: Is this here intentionally? Cleaning up default sub from alpha release?

desk/app/channels-server.hoon
```diff
@@ -1269,4 +1286,74 @@
...
+    %+  turn  ids
+    |=  =id-hook:h
+    ^-  [id-hook:h hook:h]
+    =/  hook  (~(got by hooks.hooks) id-hook)
+    =/  config  (~(gut by config.hook) nest ~)
+    =/  config-map  (~(put by *(map nest:c config:h)) nest config)
+    :-  id-hook
+    hook(compiled ~, state !>(~), config config-map)
+  :*  nest
+      *data:m
```
Fang-, line 1311: We don't have template metadata anywhere I don't think. Should it just be removed?

desk/app/channels-server.hoon
```diff
@@ -1269,4 +1286,74 @@
...
   ==
+++  get-hook-template
+  |=  =nest:c
+  ^-  template:h
+  =/  order=(list id-hook:h)  (~(got by order.hooks) nest)
+  =/  crons=(list [id-hook:h job:h])
+    %+  roll  ~(tap by crons.hooks)
+    |=  [[=id-hook:h =cron:h] cr=(list [id-hook:h job:h])]
+    ?~  job=(~(get by cron) nest)  cr
+    (snoc cr [id-hook u.job])
```
Fang-, line 1297: Might be more clear to write as a `+murn` call, if it's just accumulating into a list.

Fang- (commented):
Primarily just a question about template metadata.

desk/app/channels.hoon
arthyn:
```suggestion
```

desk/app/channels-server.hoon
arthyn: yeah I was thinking we'd want it when you can make a static template, but we can save for then

Fang- (approved):
Changes look good, thank you!

latter-bolden (commented):
Client code looks good, just a few small comments/questions.

Would also request we wire up analytics event for copy hooks success/failure (super easy to drop in on loggers, see logic/analytics.ts for event examples)

Let's talk through functional review when you have a sec

packages/shared/src/api/channelsApi.ts
```diff
@@ -94,12 +101,28 @@ export const createChannel = async (channelPayload: ub.Create) => {
       return (
         'create' in event.response &&
         event.nest ===
-          `${channelPayload.kind}/${channelPayload.group}/${channelPayload.name}`
+          `${channelPayload.kind}/${getCurrentUserId()}/${channelPayload.name}`
```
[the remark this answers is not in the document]
arthyn: I don't know lol

packages/shared/src/api/urbit.ts
```diff
@@ -241,7 +242,12 @@ export async function subscribeOnce<T>(
   }
   logger.log('subscribing once to', printEndpoint(endpoint));
   try {
-    return config.client.subscribeOnce<T>(endpoint.app, endpoint.path, timeout);
+    return config.client.subscribeOnce<T>(
+      endpoint.app,
+      endpoint.path,
+      ship,
+      timeout
```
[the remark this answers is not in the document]
arthyn: it seemed ok. it's only used on the preview. if it turns out to be an issue I can take it off.

latter-bolden (approved):
Nice work!

PR #4290, opened 2024-12-16
native: fix navigating to tapped notification

davidisaaclee:
fixes TLON-3384

Bug was caused by not passing "is desktop" to `getMainGroupRoute` – so we were trying to navigate to a desktop-only route on mobile, which failed. Fixed this by making that argument required, so everyone needs to pass it.

While here, I noticed we're using the `useIsWindowNarrow` for this "is desktop" check – I think `useIsWindowNarrow` makes sense for doing responsive layout, but not for checking which sets of routes the built app is using (e.g. someone could be using desktop with a narrow viewport, then everything will break). Fixed this by adding a context-based `useAppPlatform` hook – replaced all of the navigation-based switches on `useIsWindowNarrow` that I could see.

### DIFF
--- a/apps/tlon-mobile/src/hooks/useNotificationListener.ts
+++ b/apps/tlon-mobile/src/hooks/useNotificationListener.ts
@@ -16,6 +16,7 @@ import * as api from '@tloncorp/shared/api';
 import * as db from '@tloncorp/shared/db';
 import * as store from '@tloncorp/shared/store';
 import { whomIsDm, whomIsMultiDm } from '@tloncorp/shared/urbit';
+import { useIsWindowNarrow } from '@tloncorp/ui';
 import {
   Notification,
   addNotificationResponseReceivedListener,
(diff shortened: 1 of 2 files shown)


dnbrwstr:
Two things re: app platform:
- We should probably add the provider to the `tlon-web-new` app.tsx, right?
- Does this account for mobile web layout? That has a separate set of routes which matches the app routes.

davidisaaclee:
@dnbrwstr 🤦 will add to web-new.

I didn't account for mobile web (is this `tlon-web`?) – any thoughts for what value to provide for that? maybe switch to `useAppRoutingConfig(): 'single-pane' | 'double-pane'` or something?

dnbrwstr:
Mobile web is still `tlon-web-new` -- we just switch between routers depending on the screen layout. `useAppRoutingConfig(): 'single-pane' | 'double-pane'` would work for me!

davidisaaclee:
I'm splitting off the provider stuff into another PR; just reset head here to the single commit fixing the push bug

dnbrwstr (approved):
Lgtm!

PR #4292, opened 2024-12-16
invites: fix onboarding invite preview image size

latter-bolden:
Micro PR to reset onboarding invite images to normal size for now. A lot of times these Lure groups don't have images anyhow, in which case the first letter fallback is too dominating at 100px.

Fixes TLON-3405


dnbrwstr (approved):
Works for me!

PR #4293, opened 2024-12-16
privacy: make sure secret value is correct and make sure we close sheet

arthyn:
OTT, found today in QA. Going from secret to public/private was not working, just had it in the wrong spot on the conditional. And then sheet wasn't closing when clicking privacy option in menu.

PR Checklist


patosullivan (approved):
Looks good

PR #4294, opened 2024-12-16
channel-options: use channel volume in channel options

arthyn:
OTT, found this today while QA'ing, we weren't using the channel volume when we're in a channel

PR Checklist


jamesacklin (approved):
Good catch, thanks

PR #4297, opened 2024-12-17
PersonalInviteSheet cosmetics

jamesacklin:
Another totally extra and non-blocking cosmetic PR. :)

- Fixes an issue with hero Buttons where we weren't properly setting icon colors (was returning a hex from button context, not a Tamagui token)
- Does away with the size computation for Button.Icon for simplicity's sake. The Button.Icon component was only used in one other place and it doesn't make a functional difference to have it be its natural size.
- Makes the PersonalInviteButton match InviteFriendsToTlonButton in appearance, but with a hero treatment
- Better aligns the content of PersonalInviteSheet

### DIFF
--- a/packages/ui/src/components/PersonalInviteButton.tsx
+++ b/packages/ui/src/components/PersonalInviteButton.tsx
@@ -2,8 +2,9 @@ import { AnalyticsEvent, createDevLogger } from '@tloncorp/shared';
 import * as db from '@tloncorp/shared/db';
 import { useCallback, useMemo } from 'react';
 import { Share } from 'react-native';
-import { Text, View, XStack, isWeb } from 'tamagui';
+import { isWeb } from 'tamagui';
 
+import { TlonText } from '..';
 import { useContact, useCurrentUserId } from '../contexts';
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/PersonalInviteButton.tsx
```diff
@@ -61,22 +62,13 @@ export function PersonalInviteButton() {
...
-        <Text flex={1} color="$background" fontSize="$l">
-          Invite Friends to Tlon
-        </Text>
-        <Icon type="ChevronRight" size="$l" color="$background" />
-      </XStack>
+    <Button hero onPress={handleInviteButtonPress}>
+      <Button.Icon>
+        <Icon type="Link" />
+      </Button.Icon>
+      <TlonText.Text color="$background" size="$label/l">
```
jamesacklin, line 69: Used TlonText.Text here instead of Button.Text because Button.Text needs to be 100% wide—I didn't want to completely refactor Button to determine if its children contained both text and an icon to conditionally style that variant

latter-bolden (approved):

PR #4298, opened 2024-12-17
Personal Invite Link Fix

latter-bolden:
If something went wrong creating your link on the invite service/branch, but it went through on the provider, the initial attempt to verify your link would fail but subsequent attempts would succeed. We should never show the invite link unless we positively have it in both places.

This adds extra handling to enforce that invariant.


arthyn (approved):
lgtm!

PR #4300, opened 2024-12-19
chat input: fix issue with attachments not clearing and input not resetting immediately on send

patosullivan:
OTT, fixes tlon-3422


arthyn (approved):
lgtm!

PR #4301, opened 2024-12-19
load perf analytics

dnbrwstr:
This PR adds duration tracking via PostHog for two events: initial sync and channel initial page load.


patosullivan (approved):
Yes ✅

PR #4302, opened 2024-12-19
clean up warnings

dnbrwstr:
This pr cleans up a few existing warnings:

- QRCode prop warning: happened because cached invite link was null before being loaded, which the QR component doesn't like. Fixed by waiting to render until we have a valid invite link.
- ` useFindSuggestedContacts` warning: happened because this hook would sometimes run before the urbit client was ready. Fixed by moving client initialization logic to an outer component (also had to do some deps cleanup as part of this, including memoizing the `usePosthog` result
- Fixed a few of the easier circular import errors.


patosullivan (approved):
Looks good

PR #4308, opened 2025-01-03
Enable lure flow logging in %grouper

mikolajpp:
The Lure backend infrastructure comprises of three agents: %grouper, %reel and %bait. 
This PR augments %grouper with lure redemption tracing, as described in detail in TLON-3378.
(%logs agent functionality is also extended with a new event type, `%tell`, which carrier an `$echo` message with specified volume: `%info`, `%warn` or `%crit`.)

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -74,6 +74,7 @@
   |_  =bowl:gall
   +*  this  .
       def   ~(. (default-agent this %|) bowl)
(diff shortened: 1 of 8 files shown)


jamesacklin:
Thanks @mikolajpp!

@latter-bolden and @arthyn, if you have any early comments, would love to see them.

This is a good step forward.

arthyn (approved):
lgtm

PR #4310, opened 2025-01-06
desktop: bring back leap

patosullivan:
fixes tlon-3343

only works on desktop, only shows items that could appear in the chat list, looks like this:

https://github.com/user-attachments/assets/c9eb8a23-8a1e-45a5-a1d5-65b402c66855

I imagine in the future we'll expand it to include all channels.


packages/ui/src/components/GlobalSearch/index.tsx
```diff
@@ -0,0 +1,361 @@
...
+  });
+
+  const resolvedChats = useMemo(() => {
+    return {
+      pinned: chats?.pinned ?? [],
+      unpinned: chats?.unpinned ?? [],
+    };
+  }, [chats]);
+
+  const displayData = useFilteredChats({
```
latter-bolden, line 57: is the object literal syntax going to re-run the hook on every render? Should that input be memoized?

packages/ui/src/components/GlobalSearch/index.tsx
```diff
@@ -0,0 +1,361 @@
...
+  useEffect(() => {
+    const firstItemIndex = listItems.findIndex(
+      (item) => !isSectionHeader(item)
+    );
+    if (firstItemIndex >= 0) {
+      setSelectedIndex(firstItemIndex);
+    }
+  }, [searchQuery]); // Only run when search query changes
+
+  const contentContainerStyle = {
```
latter-bolden, line 87: Probably wanna memoize or pull into top level const

packages/ui/src/components/GlobalSearch/index.tsx
```diff
@@ -0,0 +1,361 @@
...
+  }, [isOpen]);
+
+  if (!isOpen) return null;
+
+  return (
+    <>
+      <View
+        // eslint-disable-next-line
+        onPress={() => {
+          console.log('pressed');
```
latter-bolden, line 235: stray log?

packages/ui/src/components/GlobalSearch/index.tsx
```diff
@@ -0,0 +1,361 @@
...
+      >
+        <TextInputWithIconAndButton
+          ref={inputRef}
+          placeholder={`Search for groups or DMs (${
+            navigator.platform.includes('Mac') ? '⌘K' : 'Ctrl+K'
+          })`}
+          icon="Search"
+          value={searchQuery}
+          onChangeText={setSearchQuery}
+          onKeyPress={(e: NativeSyntheticEvent<TextInputKeyPressEventData>) => {
```
latter-bolden, line 270: Worth memoizing?

latter-bolden (approved):
Code looks good! Just a few tiny nits

PR #4312, opened 2025-01-07
convert remaining keyval items to new format

dnbrwstr:
This PR converts the remaining `keyVal` items and hooks to use the new format. It doesn't change any stored keys, so shouldn't cause any changes to stored state.

Fixes TLON-3122

### DIFF
--- a/apps/tlon-mobile/index.js
+++ b/apps/tlon-mobile/index.js
@@ -26,7 +26,7 @@ setStorage(AsyncStorage);
 function Main(props) {
   useEffect(() => {
     async function init() {
-      const appInfo = await db.getAppInfoSettings();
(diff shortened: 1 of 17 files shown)


jamesacklin:
Very nice. Are we concerned with migration at all?

dnbrwstr:
> Very nice. Are we concerned with migration at all?

This doesn't modify any of the keys associated with the stored values, so all previously set values should be retained. You can verify locally by changing values on `develop`, switching  to this branch, and checking that settings haven't changed.

jamesacklin (approved):
All my settings persisted across branch changes.

latter-bolden (approved):

PR #4314, opened 2025-01-07
clean up bare chat input

patosullivan:
While looking into the message send/draft bugs (which I couldn't repro on develop) I got annoyed with the duplicate code and the overall busy-ness of the code in the component itself, so I decided to clean it up a bit.


dnbrwstr (approved):
Nice! Lgtm.

PR #4315, opened 2025-01-07
MobileAppToast: use settings state

jamesacklin:
- Replaced local storage usage with settings state to manage the visibility of the mobile app toast.
- Introduced a new property `seenMobileAppToast` in the settings state to track if the toast has been dismissed.
- Updated the toast component to conditionally render based on the loading state and the new settings property.
- Modified the dismiss button to update the settings state instead of local storage.


patosullivan (approved):
lgtm!

PR #4319, opened 2025-01-08
refine manage channels screen

patosullivan:
fixes tlon-3340

This refines the manage channels screen to:

- use arrows rather than drag and drop
- match (as closely as we can for now) the new ochre design
- look less weird on desktop
- work reliably

It also:
- hardens the associated `groupActions`
- hardens sync updates
- adds a missing db query
- fixes some deps issues on other queries
- and fixes some UI bugs having to do with actually rendering the channels in the right order (according to the index we have on the backend)

Note:

- For now we've disabled the section moving arrows (I added "Edit" and "Delete" buttons on the section title row, though)
- The "default" section (which we currently call "Sectionless") appears at the bottom of the list since this is actually where it appears in the GroupChannelsScreen. If we wanna change that in both places to match the ochre design showing it at the top, with the "Default section" title, that's fine by me.

Looks like this on desktop:
image

Looks like this on mobile:

https://github.com/user-attachments/assets/bc4479ce-20c0-4e33-b953-8794276fdfe8

### DIFF
--- a/packages/ui/src/components/ManageChannels/ManageChannelsScreenView.tsx
+++ b/packages/ui/src/components/ManageChannels/ManageChannelsScreenView.tsx
@@ -674,16 +674,15 @@ export function ManageChannelsScreenView({
         <YStack
           backgroundColor="$background"
           gap="$2xl"
-          // padding="$xl"
           alignItems="center"
           flex={1}
         >
           <ScrollView
             style={{ zIndex: 1, elevation: 1, width: '100%' }}
             gap="$2xl"
-            paddingBottom={bottom}
             contentContainerStyle={{
               alignItems: 'center',
+              paddingBottom: bottom,
             }}
           >
             {renderSectionsAndChannels}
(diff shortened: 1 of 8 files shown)


patosullivan:
Just noticed I need to clean up a bit around adding a new section, converting to draft for now.

patosullivan:
Ok, should be good to go now.

packages/ui/src/components/ManageChannels/ManageChannelsScreenView.tsx
```diff
@@ -393,129 +376,295 @@ export function ManageChannelsScreenView({
...
+      }
     },
     [sections, moveNavSection]
   );
 
+  const handleMoveChannelWithinNavSection = useCallback(
+    async (channelId: string, sectionId: string, newIndex: number) => {
+      const sectionIndex = sections.findIndex((s) => s.id === sectionId);
+      if (sectionIndex === -1) {
+        console.error('Invalid section ID:', sectionId);
```
arthyn, line 470: do we want to use logger error here?

arthyn (approved):
lgtm just some logging questions

packages/ui/src/components/ManageChannels/ManageChannelsScreenView.tsx
patosullivan: good point

jamesacklin (approved):
Not bad at all, looks really nice!

PR #4321, opened 2025-01-09
scroller: dynamically calculating item widths

arthyn:
Not sure if we want to take this approach but figured I'd put it up here. This dynamically calculates item width using a max width of 250px which I could see being a setting in the custom config. I wasn't sure how much to generalize here.

Fixes TLON-3335

PR Checklist


patosullivan (approved):
Looks great

PR #4311, opened 2025-01-06
ImageViewerScreenView: implement image download

jamesacklin:
Implements image downloads in ImageViewerScreenView, allowing users to save images to their library. Update expo-media-library to version 15.9.2 and add necessary permissions for photo library access. Update related Podfile and lock files to reflect changes.

The first time the user taps the down-arrow button, we pop a permission modal asking to grant Tlon permission to download files to the camera roll. We have a success/fail indicator for both scenarios.

Screen recording:

https://github.com/user-attachments/assets/a95d52d4-54cc-4076-99df-cd73cbf9c321

PR #4322, opened 2025-01-10
Revert "Merge pull request #4311 from tloncorp/ja/image-save"

jamesacklin:
This reverts #4311, hopefully eliminating the on-launch crash of nightly builds.


latter-bolden:
Closing since the pull this is reverting doesn't actually seem to be problematic.

PR #4325, opened 2025-01-10
invites: enhanced error and debug reporting during invite creation

latter-bolden:
Makes sure we're always getting debug context in PostHog while attempting to generate or show existing invite links. Also adds a warning during signup if the invite is missing relevant metadata.

Fixes TLON-3444


arthyn (approved):
lgtm

PR #4326, opened 2025-01-13
chat: fix issue with delay on sending followup messages

patosullivan:
We were awaiting `sync.syncChannelMessageDelivery` when we didn't need to, and `syncChannelMessageDelivery` is fairly heavy.


latter-bolden (approved):
Makes sense! Good find

PR #4327, opened 2025-01-13
manage channels: fix issue w/updating local state after adding/deleting channels

patosullivan:
fixes tlon-3448

### DIFF
--- a/packages/ui/src/components/ManageChannels/ManageChannelsScreenView.tsx
+++ b/packages/ui/src/components/ManageChannels/ManageChannelsScreenView.tsx
@@ -330,9 +330,40 @@ export function ManageChannelsScreenView({
         type: c.type,
       })),
     }));
+    const currentTotalChannels = sections.reduce(
+      (acc, section) => acc + section.channels.length,
+      0
+    );
+    const newTotalChannels = newNavSections.reduce(
+      (acc, section) => acc + section.channels.length,
+      0
+    );
 
-    // Only update if the total number of sections have changed
+    // Only update local state if the total number of sections have changed
+    // OR if a newly created channel has been added to the default section
+    // OR if a channel has been deleted
     if (newNavSections.length === sections.length) {
+      if (newTotalChannels !== currentTotalChannels) {
+        // Check if a new channel has been added to the default section
+        if (
+          newNavSections.some(
+            (s) =>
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/ManageChannels/ManageChannelsScreenView.tsx
```diff
@@ -330,9 +330,40 @@ export function ManageChannelsScreenView({
...
+    );
+    const newTotalChannels = newNavSections.reduce(
+      (acc, section) => acc + section.channels.length,
+      0
+    );
+
+    // Only update local state if the total number of sections have changed
+    // OR if a newly created channel has been added to the default section
+    // OR if a channel has been deleted
     if (newNavSections.length === sections.length) {
```
latter-bolden, line 345: Based on the comment, it seems like we'd wanna run those inner IF's even if this conditional is false (no new section added)?

packages/ui/src/components/ManageChannels/ManageChannelsScreenView.tsx
```diff
@@ -330,9 +330,40 @@ export function ManageChannelsScreenView({
...
+      if (newTotalChannels !== currentTotalChannels) {
+        // Check if a new channel has been added to the default section
+        if (
+          newNavSections.some(
+            (s) =>
+              s.id === 'default' &&
```
latter-bolden, line 351: Why do we only check if new channels were added to the _default_ section? Wouldn't we want to update if any section has a different set of channels?

patosullivan: As of now they can only be added to the default section

packages/ui/src/components/ManageChannels/ManageChannelsScreenView.tsx
patosullivan: Not sure what you mean, this condition would only be false if a new section was added. We're trying to catch cases where a new section wasn't added but we need to know if a channel was added or deleted.

latter-bolden: That makes sense, I understand the intention now.

latter-bolden (approved):

PR #4330, opened 2025-01-13
db: add db/triggers changes back in, fix issues with drizzle-orm package

patosullivan:
The issue we were seeing was caused by two problems with drizzle-orm:
1) lingering `executeAsync` and `executeAsyncRaw` uses.
2) a bug in drizzle-orm w/the way it attempts to create migration tables in sqlite: https://github.com/drizzle-team/drizzle-orm/pull/2958


packages/app/features/top/ChannelScreen.tsx
```diff
@@ -64,11 +64,6 @@ export default function ChannelScreen(props: Props) {
   const channelIsPending = !channel || channel.isPendingChannel;
   useFocusEffect(
     useCallback(() => {
-      if (!channelIsPending) {
```
arthyn, line 67: was this intentionally changed?

patches/drizzle-orm@0.36.1.patch
```diff
@@ -88,6 +92,15 @@ index f1ce597e8a6e0e325788cac540a8fde32d36bb42..e6e2e6a65efa6f725d07264cc7c3a2a7
      const params = (0, import_sql.fillPlaceholders)(query.params, placeholderValues ?? {});
      logger.logQuery(query.sql, params);
      if (!fields && !customResultMapper) {
+@@ -127,7 +138,7 @@ class OPSQLitePreparedQuery extends import_session.SQLitePreparedQuery {
+   values(placeholderValues) {
+     const params = (0, import_sql.fillPlaceholders)(this.query.params, placeholderValues ?? {});
+     this.logger.logQuery(this.query.sql, params);
+-    return this.client.executeRawAsync(this.query.sql, params);
++    return this.client.executeRaw(this.query.sql, params);
```
arthyn, line 100: these need to be synchronous?

arthyn (commented):
Two questions but mostly looks good to me, @dnbrwstr should probably review though

packages/app/features/top/ChannelScreen.tsx
patosullivan: Yes, I found that keeping this in the useFocusEffect caused unnecessary syncs that would then trigger unnecessary react query invalidations through the change listener.

patches/drizzle-orm@0.36.1.patch
patosullivan: Yes, the async methods don't actually exist

arthyn (approved):
lgtm would be good to have @dnbrwstr take a look too

dnbrwstr:
Looks good to me! Let's give it a try.

PR #4341, opened 2025-01-15
Desktop: save home tab nav state when navigating to another tab

patosullivan:
Saves the home tab nav state when a user navigates to another tab in the TopLevelDrawer and then restores it when the user returns back to home.

fixes TLON-3467


arthyn (approved):
lgtm

PR #4342, opened 2025-01-16
logs: improve Lure tracing

mikolajpp:
Arbitrary JSON data can now be supplied with a %logs report.
Taking advantage of this, Lure tracing in %grouper is improved.


desk/ted/posthog.hoon
```diff
@@ -1,9 +1,9 @@
-::  posthog: submit an log event to PostHog
+::  posthog: submit an log event to Posthog
 ::
 /-  spider
 /+  io=strandio, l=logs
 ::
-=+  posthog-key='phc_GyI5iD7kM6RRbb1hIU0fiGmTCh4ha44hthJYJ7a89td'
+=+  posthog-key='phx_5Cee2IQzFw5XapO8xiXZj9VvQEAcJY2NRaEGly9ZKXbb9sf'
```
latter-bolden, line 6: I don't recognize this new key?

desk/app/grouper.hoon
```diff
@@ -149,26 +166,34 @@
...
         [id %add memo [%notice ~] ~]
       =/  =cage  chat-dm-action+!>(`action:dm:c`action)
-      =*  dez  %^  tell:log  `token.bite  %info
+      =*  dez  %^  lure-log  %info  'DM Invite Sent'
                ~[leaf+"{<joiner.bite>} invited to DM"]
       (snoc [dez caz] [%pass wir %agent dock %poke cage])
     ?~  group=(~(get by fields.metadata.bite) 'group')
-      ~[(tell:log `token.bite %warn 'group field missing' ~)]
+      :_  ~
+      %^  lure-log  %warn  'Group Field Missing'
```
latter-bolden, line 174: how are you thinking about `%warn` vs `%crit` here?

desk/app/grouper.hoon
```diff
@@ -135,11 +135,28 @@
...
     =;  caz=(list card)
       ?~  inviter=(~(get by fields.metadata.bite) 'inviter')
-        ~[(tell:log `token.bite %crit 'inviter field missing' ~)]
-      ?.  =((slav %p u.inviter) our.bowl)  ~
+        :_  ~
+        %^  lure-log  %crit  'Inviter Missing'
+        ~['inviter field missing in lure bite']
+      ?.  =((slav %p u.inviter) our.bowl)
+        :_  ~
+        %^  lure-log  %crit  'Foreign Inviter'
```
latter-bolden, line 158: Can we change the `eventName`'s for all the failures to be _Sending DM Invite Failed_ or _Sending Group Invite_ failed?

That way it'll be super clear there's a problem + easier to aggregate and we can look to the message property for additional context

latter-bolden (changes requested):

desk/ted/posthog.hoon
latter-bolden: @arthyn any way to get this into a desk without storing in plaintext on GitHub?

arthyn: ugh I've wanted something like an env var store for a while now. we could abuse settings store maybe? Or just make this a piece of state in the logs agent that you have to poke in. The problem with any of those is that we have to find a way to trigger those pokes that doesn't reveal the key? Either way someone could trivially read the state so idk if that really helps us. I think what you really want is just an authenticated service that accepts requests.

arthyn (approved):
lgtm

latter-bolden (changes requested):
Let's double check that PostHog token, but otherwise looking good!

Sorry to expand scope, but could we add one last bit:

- From `%reel` log an _Invite Link Created_ event when it receives `%reel-confirmation` from the Provider (include token and flag)
- From `%reel` log an _Invite Link Removed_ event when it receives `%reel-undescribe` (include token and flag)

desk/ted/posthog.hoon
latter-bolden: cool, fine to leave in for now but let's make sure the key is for the correct project

mikolajpp: We are now using a personal key for this that only has write access to Posthog activities. If someone abuses it, it is trivial to revoke, unlike the previous key which is probably used in many other places (and which, unfortunately, is already out!)

desk/app/grouper.hoon
mikolajpp: The bite here is still going to be processed and a DM invite is going to be send regardless (unless some other error happens) – I would not think it is possible for the group field to be missing from metadata if the original Lure invite was for a group.

It just occurs to me that there used to be personal lure invites – do we still support those? If so, perhaps this should not trigger any warning, as a personal Lure invite would not contain `group` field in the metadata.

mikolajpp:
Re Posthog token. If it the key is on the ship, the user can just read it, no matter where we put it. If want to do anything about it we would have to setup a logging service somewhere.

mikolajpp:
@arthyn @latter-bolden Just in case you missed it, there is also a sibling PR in the landscape repo: https://github.com/tloncorp/landscape/pull/304

desk/app/grouper.hoon
latter-bolden: We use a stub group ID for personal invites because if the `group` field wasn't structured like a flag, the DM invite wasn't getting sent. Those will always be `~zod/personal-invite`

I think we'll end up getting a _Group Invite Failed_ log for those, but we can deal with that noise in PostHog for now. Ideally eventually we'd special case that `group` and not log a failure.

latter-bolden (approved):
Thanks for the clarification and fixes!

PR #4345, opened 2025-01-16
desktop: get anchor scroll lock working

patosullivan:
fixes tlon-3470

This splits useAnchorScrollLock out to its own file and simplifies the logic and state management quite a bit. There is now just a single place where `scrollToIndex` gets called. I think it was failing on web because of a race condition.

Probably best for @davidisaaclee to take a look at this one, but open to reviews from anyone else too.


davidisaaclee (approved):
Code generally looks good + appreciate the cleanup – putting some notes down since I need to log off for the night:
- Outside of this PR's scope, it looks like anchors are kind of broken now, I think due to [unconfirmed posts](https://github.com/tloncorp/tlon-apps/blob/54884b415b20dcdade364edac5c8023188385025/packages/shared/src/store/useChannelPosts.ts#L170-L172) (i.e. loading the most recent messages asap) – you can see this by doing a search in channel and jumping to a result :( commenting out this `useEffect` disables loading unconfirmeds and restores the anchor behavior (which lgtm with this PR's changes)
- During a `onScrollToIndexFailed`, `develop` scrolls to an estimated offset, then attempts to do another `scrollToIndex`. I can't get a `onScrollToIndexFailed` to trigger, but I remember that when `onScrollToIndexFailed` happened, the estimated offset was often wrong, leaving you at an irrelevant part of the scroll. I'm trying to get this to trigger; but if I can't hit it, I'm happy to see it simplified

PR #4349, opened 2025-01-17
DMs: add reply action

patosullivan:
Adds a simple "Reply" action to DMs, just inserts a `> [quoted text]` string (with a carriage return) into the chat input.

Also renames the "Quote" message action to "Reply".

I don't expect we'll keep it like this, this is just a stopgap until we get a *real* reply feature into DMs.

Fixes TLON-3515

### DIFF
--- a/packages/ui/src/components/BareChatInput/index.tsx
+++ b/packages/ui/src/components/BareChatInput/index.tsx
@@ -33,6 +33,7 @@ import { View } from 'tamagui';
 
 import {
   Attachment,
+  TextAttachment,
   UploadedImageAttachment,
   useAttachmentContext,
 } from '../../contexts';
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/BareChatInput/index.tsx
```diff
@@ -322,6 +327,22 @@ export default function BareChatInput({
...
+  // Handle text attachments by inserting them into the input
+  useEffect(() => {
+    const textAttachment = attachments.find(
+      (a): a is TextAttachment => a.type === 'text'
+    );
+    if (textAttachment) {
+      if (controlledText === '') {
+        handleTextChange(`${textAttachment.text}`);
+      } else {
+        handleTextChange(`${textAttachment.text}${controlledText}`);
```
arthyn, line 339: is this basically saying, if there is already content in the input, smash that with the quote as a prefix otherwise just add the quote?

arthyn (approved):
lgtm

packages/ui/src/components/BareChatInput/index.tsx
patosullivan: yep

PR #4353, opened 2025-01-21
logs: fix broken Posthog API key

mikolajpp:
It turns out a personal API key can't be used to capture events into Posthog. This PR sadly restores the previous API key, as it seems to be only way to capture events.


mikolajpp:
Merging, as this PR only changes the key.

PR #4354, opened 2025-01-21
lure: existing groups instantly generate, new groups 1-2s

arthyn:
Fixes TLON-3475 by cleaning up the lure code for the new apps. This removes any checking for "good" links, makes all groups enable links unequivocally, and scries first before doing a subscribe fetch. Additionally we remove the `checkForNewlyJoined` in the activity update handler since we now "join" all channels we hear activity for.

This speeds it up so that existing groups get a link instantly and new groups get it within 1-2s.


packages/shared/src/db/queries.ts
```diff
@@ -3295,9 +3297,20 @@ export const insertActivityEvents = createWriteQuery(
...
+            return txCtx.db
+              .update($channels)
+              .set({ currentUserIsMember: true })
+              .where(eq($channels.id, channelId));
+          })
+        );
+      }
     });
   },
-  ['activityEvents']
```
latter-bolden, line 3300: It probably should, yes. I wonder how that impacts performance

jamesacklin:
Awesome, excited to test this!

packages/shared/src/store/lure.ts
```diff
@@ -160,38 +70,36 @@ export const useLureState = create<LureState>((set, get) => ({
...
+    const localUrl = await scry<string>({
+      app: 'reel',
+      path: `/v1/id-url/${flag}`,
+    })
+      .then((u) => {
+        lureLogger.crumb(performance.now(), 'url fetched', u, flag);
+        return u;
+      })
+      .catch((e) => {
+        lureLogger.trackError(AnalyticsEvent.InviteDebug, {
```
latter-bolden, line 88: Nit, can we `trackEvent` if it's a debug?

packages/shared/src/store/lure.ts
```diff
@@ -160,38 +70,36 @@ export const useLureState = create<LureState>((set, get) => ({
...
-          },
-          LURE_REQUEST_TIMEOUT
-        )
-          .then((en) => {
-            lureLogger.crumb(performance.now(), 'enabled fetched', flag);
+    lureLogger.crumb(performance.now(), 'fetching url with scry', flag);
+    // url (includes the token as last element of the path)
+    const localUrl = await scry<string>({
+      app: 'reel',
+      path: `/v1/id-url/${flag}`,
```
latter-bolden, line 81: This scry is new?

packages/ui/src/components/InviteFriendsToTlonButton.tsx
```diff
@@ -73,25 +77,17 @@ export function InviteFriendsToTlonButton({
...
-        group: group?.id,
-        context: 'invite button: stale, describing',
-      });
-      describe();
-    }
-  }, [group, toggle, status, isGroupAdmin, describe]);
+
+    logger.trackEvent(AnalyticsEvent.InviteDebug, {
+      group: group?.id,
+      context: 'invite button: disabled and isAdmin, toggling',
```
latter-bolden, line 87: `enabling` not `toggling`

packages/shared/src/store/lure.ts
arthyn: no we just never used it

packages/shared/src/db/queries.ts
arthyn: it seems to work without it, so I think I'll leave like this for now. I'd rather not introduce perf issues.

packages/app/features/groups/GroupMetaScreen.tsx
```diff
@@ -33,21 +33,14 @@ export function GroupMetaScreen(props: Props) {
...
-  });
 
   const handleSubmit = useCallback(
     (data: db.ClientMeta) => {
       setGroupMetadata(data);
       props.navigation.goBack();
-      if (enabled) {
-        describe();
-      }
+      store.createGroupInviteLink(groupId);
```
latter-bolden, line 41: why don't we need the `enabled` check here? How will this behave if it's a remote group that has them disabled?

packages/shared/src/store/lure.ts
```diff
@@ -321,136 +220,64 @@ export function useLure({
...
+    if (!bait) {
       return 'unsupported';
     }
 
-    if (fetched && !enabled) {
-      return 'disabled';
-    }
-
     if ((url && checkOldLureToken(url)) || (fetched && !url)) {
       return 'stale';
```
latter-bolden, line 237: are we still handling stale links? (previous lure version)

latter-bolden (approved):
Had a few more questions, but otherwise looks good to me. 

Tested generating new links, showing existing ones from web, and signing up with one + generating a new personal link.

packages/app/features/groups/GroupMetaScreen.tsx
arthyn: it will still work since it's the inviter who will send the invite, as long as they have permissions, which they should because we check permissions upon showing the button

packages/shared/src/store/lure.ts
arthyn: let me make sure

arthyn: added something just to make sure

packages/app/features/groups/GroupMetaScreen.tsx
latter-bolden: ohh, so we're counting on that "violate hosts grouper state" loophole we found earlier?

PR #4357, opened 2025-01-22
Onboarding Fixes

latter-bolden:
Fixes TLON-3513, OTP display issue needed to be carried over to another component.
Fixes TLON-3522, Phone form was never being marked invalid
Fixes TLON-3521, prevent stopped ship sequence from locking up if navigated back while still mounted

Also fixes a recovery issue I noticed where `hostedNodeId` wasn't getting set if we found an existing ship attached to an account.


patosullivan (approved):
Looks good!

jamesacklin:
Looks good by my eye. Thanks for bringing in a library to handle phone numbers.

PR #4359, opened 2025-01-22
Onboarding: Custom message for incorrect OTP code

latter-bolden:
Instead of displaying raw JSON/hosting error, present a custom message for `401` responses (incorrect/expired code).

Fixes TLON-3524


patosullivan (approved):
Makes sense to me

PR #4361, opened 2025-01-22
chat details screen

dnbrwstr:
Enabled for groups only for now.

Fixes TLON-3024


packages/app/features/top/ChatDetailsScreen.tsx
```diff
@@ -0,0 +1,506 @@
...
+          <Pressable onPress={handlePressSeeAllMembers}>
+            <XStack
+              height="$4xl"
+              justifyContent="space-between"
+              gap="$l"
+              alignItems="center"
+              $group-press={{ backgroundColor: '$secondaryBackground' }}
+            >
+              <TlonText.Text size="$label/l">
+                {canInvite ? 'Manage members' : 'See all '}
```
patosullivan, line 447: the members screen doesn't currently allow inviting, just accepting/rejecting requests or kicking/banning current members, and all of those are admin actions, so maybe we switch this based on whether the current user is a group admin?

packages/app/hooks/useChatSettingsNavigation.ts
```diff
@@ -54,6 +58,7 @@ export const useChatSettingsNavigation = () => {
 
   const onPressManageChannels = useCallback(
     (groupId: string) => {
+      console.log('navigate to manage channels', groupId);
```
patosullivan, line 61:
```suggestion
```

patosullivan (changes requested):
Looks great for the most part. Noticed some navigation issues on desktop, pushed a commit fixing those. The last remaining issue I see is that navigating to the members screen from chat details and then pressing back lands the user on a blank chat details screen.

Other than that, just two tiny nits.

packages/app/features/top/ChatDetailsScreen.tsx
dnbrwstr: Makes sense, changed.

patosullivan (approved):
Looks good!

PR #4362, opened 2025-01-22
Onboarding: reset boot sequence after completing signup

latter-bolden:
Not resetting this was causing issues if you tried to signup multiple times in a row. Edge case in the wild, but is a confounding variable during testing.


patosullivan (approved):
code looks good, makes sense

PR #4363, opened 2025-01-22
group creation: fix issue with collapsed contact book

patosullivan:
OTT, fixes tlon-3531
Simulator Screenshot - iPhone 15 - 2025-01-22 at 14 04 39


jamesacklin (approved):
Double-nested Y-stack was the issue? Seems good

PR #4368, opened 2025-01-23
Mobile: prevent unintentional logouts

latter-bolden:
We noticed people were periodically getting logged out due to authentication failure. This was mainly due to the http client realizing it was no longer properly authenticated and failing to recover.

To avoid that, in this PR:
- new logins/signups will securely store the ship auth code and use it to refresh the session as needed
- hosted users who are already logged in will continue trying to recover the code via Hosting API if possible
- going forward, automatic logouts will only happen if the auth code can't be found and you're a) self hosted or b) hosted and hosting auth has expired

Note, if the reason we can't get your code is because your ship is offline, you'll be kicked to the _GettingNodeReady_ screen where we re-authenticate anyhow.

Fixes TLON-3429


apps/tlon-mobile/src/lib/hostingAuth.ts
```diff
@@ -1,9 +1,10 @@
+import NetInfo from '@react-native-community/netinfo';
 import CookieManager from '@react-native-cookies/cookies';
 import { createDevLogger } from '@tloncorp/shared';
 import { getHostingHeartBeat } from '@tloncorp/shared/api';
 import * as db from '@tloncorp/shared/db';
 
-const logger = createDevLogger('refreshHostingAuth', false);
+const logger = createDevLogger('refreshHostingAuth', true);
```
arthyn, line 7:
```suggestion
const logger = createDevLogger('refreshHostingAuth', false);
```

arthyn (approved):
lgtm

PR #4369, opened 2025-01-24
grouper: don't send a group invite for personal lure bites

mikolajpp:
Personal lure link generate an error, since presently %grouper attempts to send a group invite in all cases. This PR adds a check on the `inviteType` field in bite metadata. If it is present and set to `user`, we don't attempt to send group invite and don't generate a failure.


latter-bolden (approved):
Change makes sense and is cleaner than switching on our hard coded `groupId`.

Chatting out of band, sounds like redeeming one correctly suppresses the error in testing as expected.

Nice work!

PR #4371, opened 2025-01-27
native: use channel.meta to store custom channel config on API

davidisaaclee:
fixes TLON-3290

> Diff stinks – this branch necessarily was branched off of `develop` (for latest custom channels code), and has `release-channels-updates` merged into it; but `release-channels-updates` does not have latest `develop`. Comparing this branch to `develop` (which is the current state) includes all of `release-channels-updates`; comparing this branch to `release-channel-updates` includes all commits in `develop` which are not in `release-channel-updates`.

> [814aaa8](https://github.com/tloncorp/tlon-apps/pull/4371/commits/814aaa869cd3cb6780bf31790b39cb4f44c6a107) should be the start of this branch's changes.

---

Updates custom channel API code to use new `meta` field on channel instead of putting a structured string in `description`.

- Since `channel.meta` is only in `channels` app and not `groups` (nor DMs), I removed a number of places where we were populating content cfg from group's channel's descriptions or DM's description
- There are some changes to the API types (`ChannelMetadata`, `ContentRenderer`) – these are safe, since the backend treats all of the data in `channel.meta` as an opaque string (and there are no prod clients that have already stored data in one of these types)
- [Use groups-ui /v5/init](https://github.com/tloncorp/tlon-apps/commit/a2ef5504114b3a8b1f5ab425b11004053c153194) – `groups-ui` `/v5/init` was added to include `channel.meta` in the init payload; we use this to seed the channel content config when loading client data for the first time.
- Extras
  - [Fix mistyped parameter name showAuthor -> showAuthors](https://github.com/tloncorp/tlon-apps/commit/771c1a283367e5015fb745938c4359e67c8e3a2d)
  - [Make custom channels flag visible to dev clients](https://github.com/tloncorp/tlon-apps/commit/dbd627d06c99fa26ccca447ab999838c814c2a29) – useful to enable custom channels for dev moons
  - [Avoid forced unwrap (after seeing it cause a crash)](https://github.com/tloncorp/tlon-apps/commit/38f42eb2d2574135897375c893574c379faafcaf) (I took the time to fix this one layer deeper, since the called hook doesn't have an uninitialized state – this problem goes deeper into the hooks called by `useContactName`, but at least one layer is more explicit)


davidisaaclee:
This branch currently has this bug:
- i have a group hosted on ship A
- ship A sees the correct, populated meta for a custom channel in this group via /v5/init
- ship B is added to that group; /v5/init for ship B shows meta: null
so ship B does not get the customizations when they join (and I don't think they will resolve the customizations until someone makes a change to the meta)

davidisaaclee:
no point in keeping this in the PR list!

PR #4373, opened 2025-01-27
old web: use new lure generation logic + prevent toggling on/off

latter-bolden:
OTT

Fixes TLON-3541

### DIFF
--- a/apps/tlon-web/src/state/lure/lure.ts
+++ b/apps/tlon-web/src/state/lure/lure.ts
@@ -133,7 +133,6 @@ export const useLureState = create<LureState>(
         let url = localUrl;
         if (!url || checkOldLureToken(url)) {
           // start the process of creating the lure
-          // createGroupInviteLink(flag);
           createLure(flag, linkMetadata);
 
           // listen for the result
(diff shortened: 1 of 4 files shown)


apps/tlon-web/src/state/lure/lure.ts
```diff
@@ -211,29 +194,14 @@ export const useLureState = create<LureState>(
...
-    $og_description: description,
-    $og_image_url: image,
-    $twitter_title: title,
-    $twitter_description: description,
-    $twitter_image_url: image,
-    $twitter_card: meta.cover
-      ? 'summary_large_image'
-      : meta.image
-        ? 'summary'
-        : undefined,
```
arthyn, line 230: why don't we need these?

arthyn (approved):
lgtm, just one question

apps/tlon-web/src/state/lure/lure.ts
latter-bolden: So on `tlon.network` these aren't used. For Branch, my thinking is we wanna have the service set them itself so we can control it in a single place. Notably, it already does that for personal invites

PR #4376, opened 2025-01-29
Recover pre-migration non-serialized SecureStore items

latter-bolden:
OTT 

Also makes sure we clear session items when directing to `WelcomeScreen`, adds handling for `JSON.parse` errors on hosting requests, and adds a few logs.

Fixes TLON-3560


dnbrwstr (approved):
Lgtm!

PR #4379, opened 2025-01-30
activity: prevent base volume from being reset

arthyn:
Fixes TLON-3574, we were erroneously putting the old volume on the side that gets overridden if there already exists a key so we were just always set back to defaults any time activity reloads.


Fang- (approved):
As discussed out of band: yeah, classic. Approved.

PR #4380, opened 2025-01-30
Invite Fixes

latter-bolden:
Fixes TLON-3575  by biting the lure if we're not already a member or invited.
Fixes TLON-3577 by adding better lure metadata extraction fallbacks and being more careful about how the invite preview card displays info.


arthyn (approved):
lgtm, ship it

PR #4381, opened 2025-01-31
activity: make sure flagged post activity items route to the post when pressed

patosullivan:
OTT
fixes tlon-2602


latter-bolden (approved):
Nice catch

PR #4384, opened 2025-01-31
chat details: fix missing navigation on leaving a group

patosullivan:
OTT
fixes tlon-3582

we're expecting this method to be named `onLeaveGroup`, not `navigateOnLeave`.


patosullivan:
Note, the issue was that `chatOptions` was expecting `onLeaveGroup` (which we rename to `navigateOnleave` in there), an optional parameter, and it wasn't there because `useChatSettingsNavigation` had it named to `navigateOnLeave`. Kinda weird that we renamed it in one spot and then used the rename elsewhere. Anyway, just renaming it to the name that chatOptions was expecting fixes it.

dnbrwstr (approved):
Code lgtm!

PR #4385, opened 2025-02-03
%logs: handle fakeships

mikolajpp:
Tests for fake identities and adjust the id in the `-posthog` thread. 

Also adds `/lib/strandio.hoon` and `/lib/strand` which, surprisingly, are missing from this desk.


arthyn (approved):
lgtm, these files were included as part of our deploy step, but may be good to keep them here

PR #4386, opened 2025-02-03
debounce update hook

dnbrwstr:
Previously, when a large number of changes was made at once, the update hook would trigger for each row all at once, meaning that all of the updates would attempt to read changes before any of the changes were actually cleared, leading to an exponential change handler explosion. This update ensures that previous changes are cleared before the next change handler attempts to read them. 

Fixes TLON-3430.


patosullivan (approved):
Nice! Makes sense to me.

latter-bolden (approved):

PR #4387, opened 2025-02-04
Channel manager: reflect channel name changes immediately

patosullivan:
fixes TLON-3585

In the channel manager we use a `useEffect` to selectively update the `sections` local state var with the groupNavSectionsWithChannel data from the sqlite db in certain cases in order to enable an optimistic UI for the channel/section movement without unnecessary re-renders (we don't want the whole thing to re-render when a user is just moving channels/sections around).

Channel name changes weren't being reflected because we weren't checking for them in this useEffect. This just adds a check for them.


arthyn (approved):
lgtm

PR #4391, opened 2025-02-04
Notebook editor: fix text color on alternative themes

patosullivan:
This change ensures that we pass the theme value for the text color (and the background) into the webview so that the colors will match the theme (and removes the use of useIsDark in the editor react app). Also updates the web version of the editor to use the appropriate colors.

Fixes tlon-3536
Simulator Screenshot - iPhone 15 - 2025-02-04 at 15 38 17
Screenshot_1738705415
Apps Groups iPhone 14 Pro Max


arthyn (approved):
lgtm

PR #4392, opened 2025-02-05
prevent unnecessary full group syncs

dnbrwstr:
This PR ensures that we only perform a full group sync once per session -- if the session is intact, we should get all group updates through subscriptions.

Question: are there any changes that *don't* come through updates that we might miss?


patosullivan (approved):
This logic makes sense, but I do think we need to comb through the `handleGroupUpdate` function in sync to make sure everything is covered.

I just went through and checked to make sure we're accounting for all of the roles changes, and it looks like we are (...but we also have some duplicate db queries that need to be cleaned up). Roles was one area where I was worried something could fall through the cracks, but it seems fine.

It also occurs to me that it'd be nice if we could have some subscription tests in our new e2e tests.

PR #4393, opened 2025-02-06
initial sync tweaks

dnbrwstr:
Tweaks to initial sync to improve apparent+actual performance:

- Ensure high priority tasks finish before kicking off lower priority
- Ignore syncing status when deciding what header message to show
- Simplify syncing tracking a bit


patosullivan (approved):
Much simpler. Makes sense!

latter-bolden (approved):
Looks good. I was worried this would fail silently for the _Connecting..._ issue we've seen, but looking at the code that should still get surfaced.

PR #4396, opened 2025-02-06
maestro: break out flows, fix app IDs and config

arthyn:
This adds some instructions for running maestro locally and also makes sub-flows of the two logins that are easiest to test with. Additionally this corrects the package to be production instead of preview so we're closer to testing the real app.


jamesacklin (approved):
Love it!

patosullivan (approved):
Looks great

PR #4397, opened 2025-02-07
Mobile: fix gallery crash

latter-bolden:
The issue here is with a navigation helper setting `selectedPostId` when returning from the detail view. When you land back in the channel, the `scrollToIndex` fails. This isn't new behavior, but I confirmed the issue crashes on the last release as well.

I'm a bit confused about why `scrollToIndex` is failing though — the post is in the FlatList and the scroll is in response to the anchored item being laid out. It does work well in chat channels. Tweaking the FlatList render window/settings had no effect.

In any sense, the two changes I made here are:
- Catch the `scrollToIndex` error so we at least don't take the app down if it fails
- Avoid scrolling to the post when returning from galleries & notebooks (the scroll position persists by default without the anchor)

Also fixes the issue Bill reported with gallery detail views failing to show old images. This was a problem with missing dimensions when relying on aspect ratio. Now uses a default value if missing.

Confirmed neither change breaks on web.

Smuggled in a tiny onboarding copy change as well.

Fixes TLON-3615
Fixes TLON-3623

### DIFF
--- a/packages/app/navigation/utils.ts
+++ b/packages/app/navigation/utils.ts
@@ -193,7 +193,6 @@ export function useNavigateBackFromPost() {
 
   return useCallback(
     (channel: db.Channel, postId: string) => {
-      console.log(`using navigate back from post`, postId);
       if (lastScreenWasActivity) {
         navigation.navigate('Activity');
         return;
--- a/packages/app/navigation/utils.ts
+++ b/packages/app/navigation/utils.ts
@@ -198,22 +198,21 @@ export function useNavigateBackFromPost() {
         navigation.navigate('Activity');
         return;
       }
+      const isChatShaped = ['chat', 'dm', 'groupDM'].includes(channel.type);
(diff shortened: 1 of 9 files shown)


jamesacklin:
Good sleuthing. I wonder if it has to do with the column-based layout vs the vertical stack, traditional "list" layout. Either way, looking forward to seeing it in action.

packages/app/navigation/utils.ts
```diff
@@ -197,18 +197,21 @@ export function useNavigateBackFromPost() {
         navigation.navigate('Activity');
         return;
       }
+      const isChatShaped = ['chat', 'dm', 'groupDM'].includes(channel.type);
```
patosullivan, line 200: Good candidate for a hook (not necessarily in this PR)

patosullivan (approved):
Makes sense to me. I don't think anyone is expecting to be able to scrollToIndex in a gallery

PR #4402, opened 2025-02-10
mobile: schedule nudge notification during "GettingNodeReady"

davidisaaclee:
fixes TLON-3618

image
We'd be happy to change copy! Just let us know! 😀

### Summary
- Schedules 10 minute push when `GettingNodeReadyScreen` is shown (just a `useEffect`)
- Canceled via `cancelNodeResumeNudge` on any of the following:
  - Ship context gets a ship set (semantically, "login succeeded")
  - `useHandleLogout` is called (both native and web variants)

### How did I test?
Logged out, uninstalled, added the following code to check scheduled notifications (once added, you can run `__check()` in a console to check scheduled notifications):

<details>

<summary>Patch</summary>

```diff
diff --git a/apps/tlon-mobile/src/App.main.tsx b/apps/tlon-mobile/src/App.main.tsx
index 10d1985fd..b3e0d1539 100644
--- a/apps/tlon-mobile/src/App.main.tsx
+++ b/apps/tlon-mobile/src/App.main.tsx
@@ -27,6 +27,7 @@ import {
   View,
   usePreloadedEmojis,
 } from '@tloncorp/ui';
+import * as Notifications from 'expo-notifications';
 import { PostHogProvider } from 'posthog-react-native';
 import type { PropsWithChildren } from 'react';
 import { useEffect, useMemo, useState } from 'react';
@@ -68,6 +69,17 @@ const App = () => {
     };
   }, []);
 
+  useEffect(() => {
+    const check = async () =>
+      console.log(
+        'Scheduled notifs',
+        await Notifications.getAllScheduledNotificationsAsync()
+      );
+
+    check();
+    Object.assign(global, { __check: check });
+  }, []);
+
   const showAuthenticatedApp = useMemo(() => {
     const blockedOnSignup = currentlyOnboarding;
     const blockedOnLoginHosted =
```

</details>

- Signing into a stopped hosted node shows the "getting node ready" screen, and schedules a nudge. (Using a shorter delay, verified that the banner shows; tapping the banner dumbly opens the app.)
- After getting a nudge scheduled, I force quit app, then reopened to the same "getting node ready" screen. Checked using `__check()` that there was only 1 notification scheduled.
- After nudge got scheduled, I logged out – `__check()`d that there was only 1 notification scheduled
- After nudge got scheduled, I waited for node to start back up while app was open – app automatically resumed login and showed home screen; `__check()`d that the scheduled notification was cancelled


davidisaaclee:
I noticed a small bug while working on this:
1. Log into app
2. Uninstall app
3. Reinstall app
App will remember you, but you'll hit `GettingNodeReadyScreen` immediately; although you'll get let in after a few seconds. (The scheduled notification will get cleared as soon as login completes, so this PR shouldn't make this any worse.)

packages/app/lib/notifications.ts
```diff
@@ -130,3 +143,41 @@ export function useUpdatePresentedNotifications() {
...
+
+// Internal ID for this notification. We use a static ID so we can (1) check
+// for existing nudges, and (2) ensure we always overwrite an existing nudge if
+// one was not canceled: we don't want two concurrent nudges scheduled.
+const NODE_RESUME_NUDGE_ID = 'node-resume-nudge';
+
+const NUDGE_DELAY_SECONDS = 10 * 60; // 10 minutes
+
+export async function scheduleNodeResumeNudge(ship: string) {
+  await requestNotificationPermissionsIfNeeded();
```
latter-bolden, line 155: do we want to avoid scheduling here if the result is `false`?

latter-bolden (approved):
Code makes sense, worked in testing.

~We should settle on notif copy before merging.~
Copy has sign off

latter-bolden:
> Uninstall app, App will remember you

Good callout. This is a known issue/feature, we don't need to address here.

packages/app/lib/notifications.ts
davidisaaclee: done in [2c266dd](https://github.com/tloncorp/tlon-apps/pull/4402/commits/2c266dd2f1c4f516eea6c6873575e5a85d1f8b62) (which also has a little change to `requestNotificationPermissionsIfNeeded` comment intended to clarify that it can return true if permission was previously granted)

PR #4404, opened 2025-02-10
optimize latest posts fetch

dnbrwstr:
Specify last sync time when fetching latest posts so that we only get new items. 

Fixes TLON-3626


patosullivan (approved):
Interested to see how this feels

PR #4406, opened 2025-02-10
e2e: settings menu, profile editor

jamesacklin:
Introduces end-to-end tests for basic profile updates and settings menu navigation.

- Uses the hosted Tlon login sub-flow
- Taps on the lower-right sigil menu
- Taps on the gear icon
- Navigates through all Settings menus and verifies that navigation works
- Edits the profile with a test nickname, status, bio, and avatar (avatar is only on iOS; I'm not sure how the camera roll interaction works on Android and if they share identifiers for Maestro's purposes)
- Sets Tlon Local as a pinned group
- Verifies that the changed profile details display both on the self-view and the top of the Contacts list
- Resets the profile

Makes a few changes to verbiage in system dialogs to avoid two elements with the same text on the screen in cases where we can't use an ID. Also adds a few test ID hooks where string-matching wouldn't work (icon buttons, mostly).

Fixes TLON-3655

### DIFF
--- a/.maestro/00-start.yaml
+++ b/.maestro/00-start.yaml
@@ -6,3 +6,5 @@ appId: io.tlon.groups
     appId: io.tlon.groups
 - runFlow: 01-self-hosted-login.yaml
 - runFlow: 02-tlon-login.yaml
+- runFlow: 10-settings.yaml
+- runFlow: 20-profile.yaml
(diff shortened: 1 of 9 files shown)


arthyn (approved):
lgtm!

PR #4408, opened 2025-02-10
Noun enabled client and initial verifier integration

latter-bolden:
For our verifier integration, we've been working off a branch that diverged in November. I want to stop battling conflicts, so this branch represents the minimal amount of changes I think will be safe to merge into `develop`:

- Adds noun scries and pokes to the http-api client and our wrapper for it
- Pulls in `@urbit/nock-js` dependency and our new helper methods
- Removes all `/desk` changes from `m/verification-and-discovery`
- Removes all partial UI implementations for Verifier functionality
- Leaves the `lanyardApi`, DB model, queries, and DB hooks

The last item is still a work in progress, but should be safe to incorporate since there are no remaining codepaths that call into it.

With this in we should avoid needless repeated conflicts and can limit future changes to actual backend verifier code and working UI components.

### DIFF
--- a/apps/tlon-mobile/package.json
+++ b/apps/tlon-mobile/package.json
@@ -66,6 +66,7 @@
     "@tloncorp/ui": "workspace:*",
     "@urbit/aura": "^1.0.0",
     "@urbit/http-api": "3.2.0-dev",
+    "@urbit/nockjs": "^1.4.0",
     "classnames": "^2.3.2",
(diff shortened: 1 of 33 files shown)


latter-bolden:
We should wait to merge until after `5.2.0` goes out

Fang- (approved):
Good enough, let's get it merged. We'll kvetch about details later.

arthyn (approved):
lgtm

PR #4410, opened 2025-02-11
ops: set new web action env vars

latter-bolden:
I'm trying to configure the env vars that are rolled into the compiled glob in our deploy action. I want to avoid having to check in values explicitly.

I set them up as secrets in GH and am attempting to reference them in the action.

Going to merge without review, we can revert in the morning if needed.


latter-bolden:
Sitrep: this did not work

PR #4413, opened 2025-02-11
Desktop: show group channels screen content in drawer when navigating directly to a post (from activity)

patosullivan:
We weren't accounting for the nested params that we can have in our nested nav structure.

the nested params aren't in the type, torturing the nav types to get them in there is probably possible but I'm not sure it's totally necessary at this point.

fixes tlon-3617


arthyn (approved):
lgtm

PR #4419, opened 2025-02-11
Roles: don't allow admins to be removed from reader/writer roles

patosullivan:
OTT, fixes tlon-3589.

(it would look like Admin was still selected, but we'd actually remove it from the `readers` or `writers` form value).


latter-bolden (approved):

jamesacklin:
Ohhhhhh, interesting

PR #4420, opened 2025-02-11
Web: Load env vars

latter-bolden:
This revives the functionality we reverted in https://github.com/tloncorp/tlon-apps/pull/4417

As a refresher: we want env vars to be accessible in `shared` so we can avoid continuously threading them through manually. We added loading for these on mobile, but it's not yet in place on web.

This bifurcates the loading process based on platform. On mobile, it'll use `Expo`'s compiled constants, on web it'll use `Vite`s. Env vars themselves are split out into a separate file to avoid duplicating non-env constants by platform.

A secondary goal is to avoid needing to check in `.env` values to have them present in the compiled web glob. The gitignored `.env.local` should be preferred in development setups.

Should address two of the issues we saw with the last merge:
- env vars not getting picked up in the CI build process (they're now tacked onto the correct job)
- unsafe web access crashing during the EAS build (it'll now only load the appropriate platform file while building)

@patosullivan I'm still wondering about that cpu spike you saw that seemingly went away when this was reverted. I'm not sure what's up with that, does anything stick out to you?


patosullivan (commented):
The changes themselves make sense, and it builds on both web and iOS, but I can't test for Android because I'm seeing this build failure (on this branch and develop):

```
FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':expo-image:kaptGenerateStubsDebugKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
   > Internal compiler error. See log for more details

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.
==============================================================================

2: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':expo-updates:kaptGenerateStubsDebugKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
   > Internal compiler error. See log for more details
````

patosullivan (approved):
Lgtm since you're able to build it on android locally

PR #4428, opened 2025-02-12
desktop: show message actions context menu for gallery and notebook posts

patosullivan:
Uses the same pattern we use for chat messages. fixes tlon-3688


arthyn (approved):
lgtm

PR #4429, opened 2025-02-12
chat: prevent on fail loop

arthyn:
OTT, the new on-fail logging was getting triggered by blocked ships trying to DM the blocker. Because we didn't handle `/logs` in the `agent` arm, it would crash and produce another poke to `/logs` creating a loop. This just no-ops the ack handling of that poke. I checked all other agents in this desk and they are either handling that already or have a different `on-agent` setup that prevents this from happening.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -641,6 +641,7 @@
   ?+    pole  ~|(bad-agent-wire/pole !!)
       ~  cor
   ::
+      [%logs *]  cor
       [%epic ~]  cor
       [%hook *]  cor
   ::


latter-bolden:
Want to confirm we're safe on `%landscape` as well?

mikolajpp:
A terrible overlook on my part, sorry! Every agent that uses %logs is supposed to explicitly handle the `/logs` wire, I will make sure this happens throughout our agents.

mikolajpp:
I have went over the agents both in %landscape and %groups. All other agents either intercept on `/logs` or use the default handler `+on-agent:def` which is not going to crash in %poke-acks, thus preventing infinite loop from occurring.

PR #4430, opened 2025-02-13
desktop: fix issue with weird avatar render in profile editing

patosullivan:
fixes tlon-3377

it turns out the `useStyles` on web transforms numeric height/width values into strings (e.g., `56` becomes `'56px`). This breaks the calculation for the `size` prop that gets passed into UrbitSigil, which expects a number.

Tested on iOS and Android as well to make sure this doesn't break anything over there, it's fine.

### DIFF
--- a/packages/ui/src/components/Avatar.tsx
+++ b/packages/ui/src/components/Avatar.tsx
@@ -6,6 +6,7 @@ import {
   Text,
   View,
   getTokenValue,
+  isWeb,
   styled,
   useStyle,
 } from 'tamagui';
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Avatar.tsx
```diff
@@ -327,9 +328,21 @@ export const SigilAvatar = React.memo(function SigilAvatarComponent({
     if (size && size !== 'custom') {
       return getTokenValue(size);
     } else {
+      if (isWeb && (props.width || props.height)) {
+        // Sigil size must be a number
+        if (
+          typeof props.width === 'string' ||
+          typeof props.height === 'string'
+        ) {
+          return 20;
```
arthyn, line 337: do we always want 20 in this case? I don't see any usages of `props.width` or `props.height` being sent as strings

patosullivan: I'm not sure about 20, it's just what we were returning by default before (on line 343). Not sure what you mean on the second part of this.

patosullivan: Resolved out of band.

arthyn (approved):
lgtm

PR #4434, opened 2025-02-14
Onboarding: Design pass on Stopped Node screen

latter-bolden:
OTT

Also deprioritizes the _Waitlist_ in favor of emphasizing invite submission flow.

https://github.com/user-attachments/assets/08c79854-a72e-4b1f-b305-aafb62a85d7b


latter-bolden:
I'm going to cut a demo branch to try this in action. 

Want to verify after this change we continue to always prompt for notifications at some point across all flows before merging.

PR #4435, opened 2025-02-14
desktop: sidebar pt2

arthyn:
Fixes TLON-3704. This finishes the sidebar work by adding the messages tab similar to old desktop and adding the "leap" search as a top level navigation button.

To do the messages tab I had to start tracking which tab we were last in. Additionally I had to add two new "tabs" to use as filters, so that current mobile behavior could stay the same. Now on desktop cmd+k will show a mixture of channels, groups, DMs, and group DMs.


patosullivan (approved):
This looks great! Two things, but I don't think they need to block merge:

- If I'm in a chat in the messages tab and I switch to another tab and come back to messages, I'm back at the default state for that that tab rather than the chat I was in.
- Do we want some kind of filter mechanism in the messages tab (so we can toggle between DMs and DMs + chat channels)?

arthyn:
> * Do we want some kind of filter mechanism in the messages tab (so we can toggle between DMs and DMs + chat channels)?

Yes I just forgot to include this ty haha

arthyn:
will add filter in follow-up

PR #4436, opened 2025-02-18
zuse 410: import landscape agents

Fang-:
We import several agents from the landscape desk/repo, making them live in this desk/repo instead. These agents are very specific to features or behavior of the Tlon (groups) app and as such are better off here than in landscape, where more generically useful infrastructure belongs.

The agents are as follows:
contacts, bait, reel, bark, growl, genuine

This set is based on a doc from early last year. Discussion could be had about whether %settings (and to a lesser extent, %storage) should also be migrated. I have not (yet) included those here.

We copy their files from [commit `ad4d247` in the landscape repo](https://github.com/tloncorp/landscape/tree/ad4d247e1f6a2b841ac9117a6a8b24c1fdc1fb9c) (the HEAD of the develop branch at the time of writing). Some of those files already had an older version inside of this repo. We bring them up to latest. See also the sister PR, tloncorp/landscape#311.

We update the sys.kelvin to indicate we only support zuse 410K. Landscape will do the same when removing these agents from itself. This gives us a coordination point, a clear boundary across which "ownership" of the agents will be moved. This avoids "agent can't run in two desks at once" type problems one would hit when doing this through normal OTAs instead.

Closes TLON-3180.

### DIFF
--- a/desk/app/contacts.hoon
+++ b/desk/app/contacts.hoon
@@ -1,0 +1,804 @@
+/-  activity
+/+  default-agent, dbug, verb, neg=negotiate
+/+  *contacts
+::
+::  performance, keep warm
+/+  j0=contacts-json-0, j1=contacts-json-1, mark-warmer
+::
+|%
+::  conventions
+::
+::    .con: a contact
+::    .rof: our profile
+::    .rol: [legacy] our full rolodex
+::    .far: foreign peer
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm, all the sur files were the same I'm assuming?

PR #4438, opened 2025-02-18
optimize latest posts fetch

dnbrwstr:
Only fetch new latest posts since last sync. Fixes previous issue which caused desktop to be missing post previews on subsequent loads by clearing the stored `headsSyncedAt` value before initial sync for desktop only.


latter-bolden (approved):
Fix makes sense, lgtm

PR #4439, opened 2025-02-18
screen-header: revert change to wrap, just add pointer

arthyn:
OTT, this fixes the issue seen from merging first sidebar PR preventing login from advancing.


latter-bolden (approved):
Nice find

PR #4441, opened 2025-02-19
ChatMessage: send and failure marker refinements

jamesacklin:
Fixes TLON-3334 where the double-arrow send status badge was too spaced-out on desktop.

As it turns out, Tamagui doesn't do negative-gaps too well, so I decided to absolutely-position everything in the double-arrow send status marker in ChatMessage.

I did need to rearrange some things in ChatMessage so that all the absolutely-positioned transient status badges were all roughly in the same place and didn't fall under the chat message content in some cases.

Adds a transparent background and some padding to match the AuthorRow's avatar should we decide to do embellish it later (or use it as a badge on other post types).

Also fixes TLON-3333 by changing the copy of the "foo to retry send" badge depending on screen width.


patosullivan (approved):
Code looks good. Tested on all three platforms, looks great!

PR #4443, opened 2025-02-20
web-new: themed scrollbars, app loader

jamesacklin:
I use Safari daily and I was getting annoyed with the default non-matching scrollbar appearance (even in dark mode). Adds a "color-scheme" property so Safari 17+ knows how to style its own chrome. As a nice side-quest, I made the scrollbars respect the user's currently selected theme (looks quite nice in the more colorful themes).

Also makes a very basic visual change to the app's loading screen. I didn't do anything too fancy; just want to communicate that something (anything) is happening.


patosullivan (approved):
Looks good! Good idea on changing the loading screen.

PR #4444, opened 2025-02-20
e2e: use photo ID from iOS simulator for avatar selection

jamesacklin:
Selects a specific photo from the iOS Camera Roll grid to use in the profile and contacts end-to-end tests when the user chooses an overlay avatar. If this proves to be problematic, I can always switch to a coordinate-based tap action, but I'd rather not if we can get away with it.


arthyn (approved):
lgtm

PR #4445, opened 2025-02-20
client: reduce scope of ui/ package exports

davidisaaclee:
Implements the 1/2 steps described in https://gist.github.com/davidisaaclee/bc255c623e7529fc531deb7b1888faa9

- Commands are listed in commit descriptions where possible; a lot of file moves relied on VS Code's tsserver integration to automatically update import paths, and could not be scripted.
- `[wip]` is meant to indicate that the commit will not pass typecheck.
- 3737739762098b8bf4bd5fc6e8717fea25292d24 I ran Prettier over all code in `app/` and `ui/` to clean up after automated edits – it seems like we had some un-Prettier'd code that was affected by this
- Some specific code moves:
  - [Export OnboardingBenefitsSheet in barrel](https://github.com/tloncorp/tlon-apps/commit/ab91a8f7fe18cc3381a0939b2ec410a1e8d8550a) – we were reaching into the specific file here (which was skipping Tamagui and moreover didn't match the pattern)
  - [Add missing tsc check in app package](https://github.com/tloncorp/tlon-apps/commit/63e06974b49a418be0ca941f25fbf3973c50ed63) – we did not have a `tsc` script for this package, which means that any type errors were not caught in CI – fixed all existing type errors in [Fix existing type errors in app package](https://github.com/tloncorp/tlon-apps/commit/b0ac3ac40be0ff5eb0bc22b5babf09490cfeddd7)
  - [Remove re-export of app/ code from ui/](https://github.com/tloncorp/tlon-apps/commit/d21365fdc30fd466409847e354c03f9cffcf8e0e) fixes a case where we were re-exporting an `app/` module from `ui/`, only to import it again from `app/`. (It looks like this was caused by tsserver's automated import updating – I'm not sure how we could avoid this, apart from adding a lint rule forbidding importing `app` from `ui`.)

---

This PR loses fidelity of some import paths. This makes future code movement (something that would look like this PR) more difficult, as we'll need to rebuild the absolute paths (i.e. the same issue we have with importing from `../../ui` in this PR, which we'll need to manually expand to `../../ui/Button` or whatever).

To make up an illustrating example:
```diff
- // packages/ui/src/components/ChatList.tsx
+ // packages/app/ui/components/ChatList.tsx

- import { Button } from './Button';
+ import { Button } from '@tloncorp/ui'; // <- we lose the fidelity that we are importing from `components/Button`
```
- We could trivially change this PR to import from e.g. `@tloncorp/ui/src/components/Button` instead of `@tloncorp/ui`, but we'd be ignoring the effects of the Tamagui compiler (which outputs to `ui/dist/*`)
- We can theoretically use tsserver to automate updating (99% of) these imports to their absolute, resolved files: we'd feed each imported identifier from `@tloncorp/ui` in the `app/` package into tsserver's "go to definition", and use the reported definition's file path instead of `@tloncorp/ui`. I tried for a little while to do something like this, but couldn't get it to work – but it seems possible, so maybe it's okay to build this kind of debt.

---

I tested by building/running and tapping through the iOS client; and by running `pnpm run -r tsc`.


davidisaaclee:
If reviewers get to this Thursday morning, please feel free to merge for me

dnbrwstr (approved):
Code lgtm! Should get it in before things start conflicting.

PR #4446, opened 2025-02-20
e2e: add group creation + customization tests

jamesacklin:
Adds several tests for the creation and customization of groups.

- Tap on the "+" button
- Assert that the creation sheet pops
- Assert that the contacts sheet pops
- Create the group
- Navigate back to Home and assert that the group is created
- Open the Customize group screen from the empty "General" channel
- Change the group name and confirm on the Home screen
- Change the group icon/picture (iOS only)
- Change the group description
- Open the overflow menu from the channel header
- Pin and unpin the group, checking on the Home screen for position
- Copy group reference

Deletes the test group and cleans up after each create/modify/verify sequence.

Also switches choosing a photo to coordinate-based tap commands rather than choosing by ID, which was causing trouble in the cloud test runner.


jamesacklin:
Closing in favor of #4458

PR #4447, opened 2025-02-20
edit-profile: fix mobile image input

arthyn:
In my sidebar PR I tried fixing the avatar width on desktop, to do so web needed an explicit height set. Doing so however makes the image huge on mobile. So now we only set it on desktop.


jamesacklin:
Makes sense to me, verbally approved at standup.

latter-bolden (approved):

PR #4450, opened 2025-02-20
desktop: show proper browser titles

patosullivan:
fixes tlon-3329

This shows either a pretty route name or the name of the group/channel when navigated to a group/channel.

### DIFF
--- a/apps/tlon-web-new/src/app.tsx
+++ b/apps/tlon-web-new/src/app.tsx
@@ -1,4 +1,4 @@
-// Copyright 2024, Tlon Corporation
+// Copyright 2025, Tlon Corporation
 import {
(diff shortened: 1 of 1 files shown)


patosullivan:
> Makes sense to me. Another place to remember to add something—did we do this declaratively in the route definition in old web?

We used react-helmet to specify it on individual screens, i.e., DiaryChannel had it, HeapChannel, FindGroups, DiaryNote, etc. It was distributed throughout the app.

jamesacklin:
OK, no issue there then. For some reason I thought we had the pretty title as a route prop, but I'm fine with this either way.

latter-bolden (approved):
Lgtm

PR #4451, opened 2025-02-20
Mobile: Prevent groups getting stuck in "requested invite" state

latter-bolden:
The issue here was keeping `haveRequestedInvite` in sync on the group model. 

If that flag is set, we'll display the group as an unjoined preview at the top of your list (overrides `currentUserIsMember`). The groups serializer was never modifying that column.  If you 1) had requested an invite and 2) accepted that invite on another client, the field would get "stuck" and would continue to seem unjoined.

Updates the serializer to properly set `haveRequestedInvite` and the insert query to allow overwriting it.

Fixes TLON-3707


patosullivan (approved):
Makes sense to me!

PR #4454, opened 2025-02-21
http-api: fix order of disposes so we aren't aborting new valid requests

arthyn:
OTT, @pkova and I found this when trying to investigate why we were getting stuck in a "Connecting..." state. Basically we were aborting after initiating a bunch of new requests to resubscribe to things. Aborts trigger an error on the fetch which we don't handle, so we were never hitting the code path where the event source gets reinitiated.


latter-bolden (approved):
🔥 awesome find

PR #4458, opened 2025-02-21
maestro: add mobile UI test flows for group creation, customization

jamesacklin:
Adds several tests for the creation and customization of groups.

- Tap on the "+" button
- Assert that the creation sheet pops
- Assert that the contacts sheet pops
- Create the group
- Navigate back to Home and assert that the group is created
- Open the Customize group screen from the empty "General" channel
- Change the group name and confirm on the Home screen
- Change the group icon/picture (iOS only)
- Change the group description
- Open the overflow menu from the channel header
- Pin and unpin the group, checking on the Home screen for position
- Copy group reference
- Set group privacy
- Edit the Admin role
- Create a role
- Create a channel
- Move channels up and down
- Rename a channel
- Create a channel section (currently failing due to TLON-3714)
- Set the notification volume for the group
- Verify that the channel selection screen appears in the navigation stack for a 2-channel group
- Delete a channel
- Assign a role to a user
- Change a channel's permissions to use a role
- Attempt to delete a role that still has members/channels assigned to it
- Rename the role
- Remove a role from a user
- Remove a role from a channel's permissions
- Delete a role that has no members or channels assigned to it

Deletes the test group, cleans up, and logs out after each create/modify/verify sequence.

Also makes some changes to existing tests to ensure cross-platform compatibility in the cloud runner context:

- Adds a "Close" button to the favorite group selection sheet in the profile editor (swipe-down is unreliable on iOS)
- Asserts that the "Photo Library" button appears when changing an avatar or group icon; does not upload the image (no way to check it succeeded, and was fraught w/ selectors)
- Only runs the Contacts test on iOS because the Contacts screen fails to update in the Android emulator due to its lack of support for SSE.

Finally, changes the GitHub workflow to build both app editions in parallel, then test with Robin in series. This lets us avoid overlapping states between the two tests which could cause false negatives.

### DIFF
--- a/.github/workflows/mobile-testing.yml
+++ b/.github/workflows/mobile-testing.yml
@@ -6,7 +6,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    name: Build iOS and Android
+    name: Build apps
     strategy:
       matrix:
         platform: [ios, android]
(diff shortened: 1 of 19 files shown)


arthyn (approved):
lgtm

jamesacklin:
Made some changes to the GitHub workflow. You can see it in action here: https://github.com/tloncorp/tlon-apps/actions/runs/13524335100

arthyn (approved):
latest changes lgtm

PR #4459, opened 2025-02-21
fix issue with app crashing on long press of chat messages

patosullivan:
fixes tlon-3712

We just needed to add `expo-blur` to the package.json for the `app` package.


arthyn (approved):
lgtm

PR #4463, opened 2025-02-24
Better Groups v1 

mikolajpp:
This PR introduces several important improvements to %groups.

1. We transition to lib-negotiate, removing the old epic-based system.
2. Group metadata is preserved across invitation, group entry boundary
3. We maintain and expose set of active channels that we are subscribed to as part of the group state.
4. We expose group members count both in the group and in the group preview.


Fang- (commented):
As discussed synchronously. Some comments, stuff to pay attention to, etc. glhf!

desk/sur/groups.hoon
```diff
@@ -17,6 +12,8 @@
 +$  nest  (pair term flag)
 ::
 ::  $sect: ID for cabal, similar to a role
+::TODO rename sect -> role. It seems role is not in
```
arthyn, line 15: 🙏

desk/sur/groups.hoon
```diff
@@ -138,19 +136,29 @@
...
       secret=?
       meta=data:meta
       =flagged-content
   ==
-::
-+$  group-ui  [group saga=(unit saga:e)]
++$  group-ui  [group init=? count=@ud]
 ::  $cabal: metadata representing a $sect or role
 ::
+::TODO rename cabal -> role-meta?
```
arthyn, line 149: perhaps cabal -> role and sect -> role-id

arthyn (commented):
still working through agent changes but everything else looks good so far

desk/sur/groups.hoon
mikolajpp: Good suggestion! Do note that I haven't planned to do these name changes just yet. I will soon do another round on the %groups, this time to survey everything that's in the agent and document endpoints. Name changes could be done at that time.

desk/app/groups.hoon
```diff
@@ -631,35 +769,99 @@
...
+++  to-group-ui-2
   |=  [=net:g =group:g]
-  ^-  group-ui:g
-  :-  group
+  ^-  group-ui:v2:g
+  :-  (to-group-2 group)
   ?+  -.net  ~
-      %sub  `saga.net
+      ::XX make sure this plays well with
+      ::   the frontend
```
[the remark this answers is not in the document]
mikolajpp: I have filled this in, using information from lib-negotiate obtained by scrying. If we match, we give out a saga `%chi`; if we mismatch we default to `%lev`. Otherwise we leave saga null.

desk/app/groups.hoon
```diff
@@ -506,32 +584,58 @@
       meta             meta.gr
       flagged-content  ~
     ==
+  --
+::
+++  inflate-io
+  ^+  cor
+  ::XX initiate:neg is not needed here, since subscribing
+  ::   starts version negotiation. same argument applies
+  ::   to group subscriptions below
```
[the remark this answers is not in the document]
mikolajpp: After reviewing the pokes, there does not seem to be a way to handle poke errors differently than crashing, and, possibly logging the failures. A retry queue for pokes does not seem like a good idea. It is less surprising to fail now, than to store up actions to be executed at some unknown point in the future. 

The best we can do currently is to have the client refuse to perform operations on connections that have not been negotiated, if it does not do so already. Perhaps @latter-bolden can shed some light here.

An altogether different approach would be to introduce the action-response model, that I have seen discussed somewhere, and report action errors through that channel, but that would be a separate project.

desk/app/groups.hoon
```diff
@@ -387,6 +387,8 @@
     ::  force leave
     (emit [%pass wire %agent dock %leave ~])
   ::
+  =.  cor
+    (emit [%pass /load %arvo %b %wait now.bowl])
```
Fang-, line 391: Consider a slightly more specific wire, like `/load/active-channels`. We may need to defer from `+on-load` for other things in the future, we'll want to disambiguate them.

And, shouldn't this be conditional on the 4-to-5 migration? ie `=?  cor  ?=(%4 -.old)` and putting this in between the relevant state transforms above.

desk/app/groups.hoon
```diff
@@ -846,23 +834,26 @@
...
+  |=  [=flag:g =net:g =group:g]
   ^-  group-ui:v2:g
   :-  (to-group-2 group)
-  ?+  -.net  ~
-      ::XX make sure this plays well with
-      ::   the frontend
-      %sub  ~
+  ?.  ?=(%sub -.net)  ~
+  =/  =status:neg
+    (read-status:neg bowl [p.flag %groups])
```
Fang-, line 842: Note that `+read-status:neg` also does a `.^`... which I think will still crash on you (even though you're scrying into yourself) in scenarios where in the previous event, this agent was suspended, and now it's being brought back online again and runs this migration. (The scry goes into gall, who is trying to make the agent live, but hasn't committed the live agent into its own/arvo's state yet.)

So for filling this in, too, you'll want to defer to the next event.

Fang- (commented):
Some `.^`-during-load shenanigans remain.

desk/app/groups.hoon
mikolajpp: The scries here should be good, since these gates are not used during `+on-load`.

desk/app/groups.hoon
mikolajpp: Done.

desk/app/groups.hoon
Fang-: You're right, my bad.

desk/app/groups.hoon
Fang-: With your changes, it still sends the `%wait` unconditionally on every `+on-load` call, instead of only when we migrate from 4 to 5.

Fang- (approved):
Changes look good, thank you!

arthyn (approved):
lgtm

mikolajpp:
Closes TLON-3701, TLON-3563, TLON-3562, TLON-3565

mikolajpp:
Closes TLON-3701, TLON-3563, TLON-3562, TLON-3565

PR #4464, opened 2025-02-24
channels: don't strictly need to initiate:negotiate

Fang-:
As of 34014ad the library will let pokes through (without crashing) even if the target's version is unknown. We previously had to explicitly +initiate:negotiate to make sure that pokes wouldn't crash, but that is no longer necessary.


arthyn (approved):
lgtm

PR #4470, opened 2025-02-24
channel manager: fix issue with section creation

patosullivan:
This broke when we merged https://github.com/tloncorp/tlon-apps/pull/4408, because `formatUv` was not expecting the value produced by `BigInt`, apparently. ~Using the old `bigInt` and calling `.toString()` made it happy again.~ We needed to bump to the latest @urbit/aura in tlon-mobile and tlon-web-new.

I also noticed we weren't clearing the form on EditSectionNameSheet before closing it, so I added that in here, too.

fixes tlon-3714

### DIFF
--- a/packages/shared/src/urbit/utils.ts
+++ b/packages/shared/src/urbit/utils.ts
@@ -1,5 +1,4 @@
 import { formatUd, formatUv, isValidPatp, unixToDa } from '@urbit/aura';
-import bigInt from 'big-integer';
 import { useMemo } from 'react';
 
 import { ContentReference, PostContent } from '../api';
@@ -424,7 +423,7 @@ export function extractGroupPrivacy(
 }
 
 export function createSectionId() {
-  const idParts = formatUv(bigInt(Date.now()).toString()).split('.');
+  const idParts = formatUv(BigInt(Date.now())).split('.');
   const newSectionId = `z${idParts[idParts.length - 1]}`;
 
   return newSectionId;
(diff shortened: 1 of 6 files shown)


packages/shared/src/urbit/utils.ts
```diff
@@ -424,7 +424,7 @@ export function extractGroupPrivacy(
 }
 
 export function createSectionId() {
-  const idParts = formatUv(BigInt(Date.now())).split('.');
```
arthyn, line 427: this should work fine hmm. what did you see as the error?

patosullivan: [ManageChannelsScreenView] Failed to create section: TypeError: x.neq is not a function
    at formatUv (uv.ts:24:12)
    at createSectionId (utils.ts:427:19)
    at Module.addNavSection (groupActions.ts:336:24)
    at useGroupContext.ts:136:21
    at onSave (ManageChannelsScreenView.tsx:757:19)
    at EditSectionNameSheet.tsx:29:7
    at createFormControl.ts:1153:17

arthyn: https://github.com/urbit/aura-js/blob/master/src/uv.ts I think our module is out of date somehow because that's not even present anymore

arthyn: I think tlon-mobile and tlon-web-new have it listed as 1.0.0

patosullivan: Maybe we didn't release a new version of aura after the bigint changes?

patosullivan: Or we just didn't bump it in our apps

arthyn (approved):
lgtm

jamesacklin (approved):

PR #4472, opened 2025-02-24
groups: allow immediate cancel of private grp request

arthyn:
OTT we previously were forcing you to be able to communicate with host to be able to cancel a private group request. I think the behavior we want is to immediately clear the local state and send the command to the host, then not care whether command successfully goes through or not.


jamesacklin:
Phew, I love this. Thank you!

mikolajpp:
Seems like a good change, approved.

mikolajpp (approved):

PR #4475, opened 2025-02-25
desktop: make sure we only run syncStart once

patosullivan:
fixes tlon-3743

We were running `sync.syncStart` multiple times because we were using `useState` and re-running the useEffect after we'd already started a sync, which would then call `sync.syncStart` again with `alreadySubscribed` set to `true` for no reason.

Replacing the useState-based tracking of our sync with a ref ensures we only run it one time, and we should always run `sync.syncStart` with `false` here since we're not in a discontinuity situation.

I also turned off a few loggers that had been left on in other spots.


arthyn (approved):
lgtm

PR #4476, opened 2025-02-25
nav: fix nav'ing from activity on desktop

arthyn:
Fixes TLON-3744. We weren't always taking into account `lastOpenTab` and also we weren't handling navigations from Activity very well.


patosullivan (approved):
Looks good

PR #4477, opened 2025-02-26
nav-utils: better behavior if at root

arthyn:
I noticed leap was broken, the cause was because we were not handling the "root" navigator very well. This fixes that by checking for root and replaces the error with a return of `lastOpenTab` so we CYA in case some weird nav case isn't handled.


patosullivan (approved):
Makes sense

PR #4478, opened 2025-02-26
e2e: chat channels

jamesacklin:
Introduces a new test that:

- Creates a new group
- Inserts some text into the General channel and sends it
- Navigates back to the Home screen and asserts that the message preview is visible
- Navigates back into the group
- Long-presses the message and starts a thread
- Sends a message in the thread
- Navigates back to the channel root and asserts that the UI shows 1 reply
- Long-presses the message and reacts with a thumbs-up
- Asserts that a thumbs-up reaction appears below the message
- Taps on the reaction and asserts that it is removed
- Long-presses the message and starts a quote-reply with a reference
- Asserts that the quote-reply is sent
- Sends a message, long-presses on it, hides the message, and asserts that the message is hidden
- Sends a message, long-presses on it, reports the message, and asserts that the message is hidden
- Sends a message, long-presses on it, deletes the message, and asserts that the message is deleted
- Sends a message, long-presses on it, edits the message, sends the edit, and asserts that the edit appears in the chat log

Does the same for chat threads as well.

Also deletes the group and logs the user out as part of cleanup.

### DIFF
--- a/.maestro/51-threads.yaml
+++ b/.maestro/51-threads.yaml
@@ -1,0 +1,200 @@
+appId: io.tlon.groups
+---
+- runFlow: subflows/launch.yaml
+- runFlow: subflows/tlon-login.yaml
+- assertVisible: 'Home'
+# Tap on the "+" button
+- tapOn:
+    id: 'CreateGroupButton'
+# Assert that the sheet pops
+- assertVisible: 'Create customizable group chat'
+- tapOn: 'New group'
+# Assert that the contacts sheet pops
+- assertVisible: 'Select contacts to invite'
+# Create the group
+- tapOn: 'Create group'
+- runFlow:
+    when:
+      visible: 'General'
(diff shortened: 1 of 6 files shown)


.maestro/51-threads.yaml
```diff
@@ -0,0 +1,202 @@
+appId: io.tlon.groups
+---
+- runFlow: subflows/launch.yaml
+- runFlow: subflows/tlon-login.yaml
+- assertVisible: 'Home'
+# Tap on the "+" button
```
arthyn, line 6: might be worth sub-flowing the group creation

arthyn (approved):
lgtm

.maestro/51-threads.yaml
jamesacklin: Possibly, we do it enough

PR #4479, opened 2025-02-26
reply-summary: simpler more reliable notify tracking

arthyn:
We were using a combination of checking volume settings and the actual `notify` flag on the activity summary. However, that flag already takes settings into account. So we were getting some situations where the client thought thread was "muted" but in reality it was trying to notify.

Should see blue dots more often now in the correct places.


patosullivan (approved):
Looks good!

PR #4481, opened 2025-02-27
native: add periodic background sync

davidisaaclee:
Fixes TLON-3628

Adds a background task which periodically performs an initial sync (i.e. `startSync(true)`).

- When does this task run?
  - Android: Runs every 10 minutes; does not run if app has been force-quit or if the app hasn't been opened since last system boot
  - iOS: Runs as often as possible, subject to OS's fuzzy scheduling – supposedly every 10-15 minutes. Does not run if app has been force-quit.
- What gets synced?
  - Everything we sync on initial sync – which might be too much, but for me, this sync was very fast. On iOS, if sync takes more than 30 seconds, the background task will be killed (and will likely be invoked less often in the future).
- Is this hard on the battery?
  - This is hard to tell – I tried checking expo-battery's `Battery.getBatteryLevelAsync` before and after the task was run on a real iOS device, but there was not enough resolution to give a difference (i.e. diff = 0).
  - iOS seems pretty conservative about not running the task if it'll be a battery drain, based on docs.
- What effect does this have on returning to app?
  - iOS
    - When app is still active but backgrounded (i.e. user swiped out of app recently), the background sync makes newer messages appear immediately on return to app.
    - When app is inactive (i.e. user swiped out of app and has used many apps since): I don't know how to test this. I think we'll probably hit [this issue](https://linear.app/tlon/issue/TLON-3628/fetch-latest-conversations-from-background#comment-f92d6612) where we need to wait for a session to be established before showing posts.
  - Android: Hard to tell since I've only gotten the background task to work on a dev build, so returning to the app causes a JS reload; but it seems like the new messages are immediately shown in the channel, then the rest of the channel loads.

#### iOS

> `Debug > Simulate background fetch` only works on a release build, and has only worked on simulator for me.

Xcode has an option at `Debug > Simulate background fetch`, which backgrounds the app and triggers the `application:performFetchWithCompletionHandler:` app delegate method. I swapped out the `logger` in `backgroundSync.ts` with a special logger that prints its logs to local notification banners, and checked that (1) background sync was completing, (2) the channel contents before sync did not have message received while backgrounded, and (3) channel had new messages after performing background sync.

I also installed onto a real device with debug logs enabled, and saw that it did background sync as expected after waiting.

<details>

<summary>

`Debug > Simulate background fetch` appears to only work on release builds: I think makes sense because`expo-task-manager` probably needs to load the static JS without hitting the server.

</summary>

`expo-task-manager` executes background tasks by:
1. receive request to perform task
2. load JS app in headless mode (using [`UMAppLoader`](https://github.com/expo/expo/tree/main/packages/unimodules-app-loader))
3. JS app registers tasks synchronously on app launch (this is why `registerBackgroundSyncTask` is called at global scope)
4. after headless app execution, expo-task-manager collects all the registered tasks matching the request and executes them.
My thinking is that it can't do step 2 in the background via dev server – no real data to back this up. [relevant code](https://github.com/expo/expo/blob/a2fda19a5aba854cdee8c0c82240efc5c0b1a2e9/packages/expo-task-manager/ios/EXTaskManager/EXTaskService.m#L534)

</details>

#### Android
Set `minimumInterval: 60` when calling [`BackgroundFetch.registerTaskAsync`](https://github.com/tloncorp/tlon-apps/blob/f153338131bb229ff8a5401c15d9deaaffd89bda/packages/app/lib/backgroundSync.ts#L83); do `pnpm run --filter tlon-mobile android` onto an emulator; then do similar checks as listed above (waiting a minute after backgrounding app to have background fetch automatically triggered).

### Future work
**This PR does not run background sync when receiving a notification.** expo-notifications has a `registerTaskAsync` hook to register running a `expo-task-manager` task when receiving a data-only notification (`content-available: 1`), but that doesn't appear to work for our use case: our notifications (which have `content-available: 1` *and* `mutable-content: 1`) do not (reliably?) trigger the required app delegate callback (`application:didReceiveRemoteNotification:fetchCompletionHandler:`).

Some solutions to trigger background sync on our user-facing notifications:
- Update backend to send separate silent, data-only notifications to trigger background sync. (These would use the same mechanism as our old pushes, and Apple technically says to only send 3 / hour.)
- Find a way to call an `expo-task-manager` task from the notifications native code.
  - iOS: The notif extension is a separate bundle from the host app – we'd need to find a way to access the JS bundle. This seems doable (and opens more options for extensions in the future, e.g. lockscreen widgets), but off the beaten path.
  - Android: I don't know whether we can access the JS bundle from the notification intent, but I think it'd be similar to iOS once we have access to the JS bundle.
- Re-implement background sync in native code. (This would be a lot of work, and require maintenance to keep in lockstep with JS.)

### DIFF
--- a/apps/tlon-mobile/ios/Podfile.lock
+++ b/apps/tlon-mobile/ios/Podfile.lock
@@ -95,6 +95,8 @@ PODS:
     - glog
     - RCT-Folly (= 2022.05.16.00)
     - React-Core
+  - ExpoBackgroundFetch (11.8.1):
+    - ExpoModulesCore
   - ExpoBattery (7.7.2):
     - ExpoModulesCore
   - ExpoBlur (12.9.2):
@@ -1399,7 +1401,7 @@ PODS:
     - glog
     - RCT-Folly (= 2022.05.16.00)
     - React-Core
-  - UMAppLoader (4.5.0)
+  - UMAppLoader (4.5.1)
   - Yoga (1.14.0)
 
 DEPENDENCIES:
@@ -1420,6 +1422,7 @@ DEPENDENCIES:
   - expo-dev-launcher (from `../../../node_modules/expo-dev-launcher`)
   - expo-dev-menu (from `../../../node_modules/expo-dev-menu`)
   - expo-dev-menu-interface (from `../../../node_modules/expo-dev-menu-interface/ios`)
+  - ExpoBackgroundFetch (from `../../../node_modules/expo-background-fetch/ios`)
   - ExpoBattery (from `../../../node_modules/expo-battery/ios`)
   - ExpoBlur (from `../../../node_modules/expo-blur/ios`)
   - ExpoClipboard (from `../../../node_modules/expo-clipboard/ios`)
@@ -1582,6 +1585,8 @@ EXTERNAL SOURCES:
     :path: "../../../node_modules/expo-dev-menu"
   expo-dev-menu-interface:
     :path: "../../../node_modules/expo-dev-menu-interface/ios"
+  ExpoBackgroundFetch:
+    :path: "../../../node_modules/expo-background-fetch/ios"
   ExpoBattery:
     :path: "../../../node_modules/expo-battery/ios"
   ExpoBlur:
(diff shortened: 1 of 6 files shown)


arthyn (approved):
lgtm

PR #4482, opened 2025-02-27
groups: handle http requests, if we ever get bound

Fang-:
We may in some (hosting) circumstances want to bind the `/` catch-all binding to something we control, so that we can redirect root (and other unhandled pages) back into the groups web app.

Here we make the groups agent handle http requests by redirecting to `/apps/groups/`, setting us up for `|pass [%e %connect [~ /] %groups]` or some equivalent.

This goes into `release-410` because the same is true for tloncorp/landscape#313, which this soft-depends on. The kelvin is a nice boundary for solidifying that relation.

Someone sign off on the trailing `/` on the redirect path for me please.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -250,6 +250,20 @@
       [~ %| *]  ~&  [dap.bowl %overwriting-pending-import]
                 cor(pimp `|+egg-any)
     ==
+  ::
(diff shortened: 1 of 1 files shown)


jamesacklin (approved):

arthyn:
@Fang- trailing slash pls

PR #4483, opened 2025-02-28
native: omit param instead of passing `undefined`

davidisaaclee:
We want to use the default value for `minimumInterval`; and since this is such an important parameter for UX, I want to make sure our usage is explicit, so I set `minimumInterval: undefined` to call out that we're using the default value. 

Instead of omitting an `undefined` entry when converting to native code, expo-background-fetch converts this to `"<null>"`, which breaks the native module.

(This change was made in my previous local branch but not included in the final set of commits – double-checked my stash, I didn't see any other omissions.)

context: /1/chan/chat/~bitpyx-dildus/interface/msg/170141184507263636203902727312912678912


arthyn (approved):
lgtm

PR #4484, opened 2025-02-28
favorite-groups: dont leak secret groups

arthyn:
Fixes TLON-3753 by filtering the groups we insert into the selector to only those which are not secret. Additionally desktop was rendering this selector as a sheet so I went down the trail of doing that. Needed to move around the scroll and content elements to make it work. I tried opening as many modals as possible and they all seem to render fine. I'm not sure why `ScrollableContent` had `overflow: hidden` which seems wrong? so I removed.


thelifeandtimes:
image
I feel excluded. 

Of course, not unreasonable since what I am trying to push is a pendantic typo fix..
image


Thanks for fixing this stuff :)

packages/app/ui/components/ActionSheet.tsx
```diff
@@ -174,10 +178,11 @@ const ActionSheetComponent = ({
             maxWidth={800}
             minWidth={400}
             key="content"
-            maxHeight="100%"
+            // prevent the modal from going off screen
+            maxHeight={maxHeight}
```
patosullivan, line 182: Nice

patosullivan (approved):
Looks good!

jamesacklin (approved):

PR #4485, opened 2025-03-03
groups: make sure we can cancel an errored join

arthyn:
Fixes TLON-3179 for cases where the join in progress actually errors (stuck private requests were fixed with tloncorp/tlon-apps#4472). I was tempted to refactor the way we track status here because we're juggling a lot of booleans, but decided to push it off for a little longer, although I still think we should change it sooner than later.


patosullivan (approved):
Makes sense to me

PR #4490, opened 2025-03-05
Fix %groups scry mismatches 

mikolajpp:
This PR adjusts the %groups scry types to correspond to the actual result.

### DIFF
--- a/desk/app/diary.hoon
+++ b/desk/app/diary.hoon
@@ -202,11 +202,11 @@
     (toggle-post toggle)
   ::
       %leave-old-channels
-    =/  groups-path  /(scot %p our.bowl)/groups/(scot %da now.bowl)/groups/noun
-    =/  groups  .^(groups:g %gx groups-path)
+    =/  groups-path  /(scot %p our.bowl)/groups/(scot %da now.bowl)/groups/groups
+    =/  groups  .^(groups:v2:g %gx groups-path)
     =/  diary-flags-from-groups
       %+  turn  ~(tap by groups)
-      |=  [group-flag=flag:g group=group:g]
+      |=  [group-flag=flag:g group=group:v2:g]
       %+  turn
         %+  skim  ~(tap by channels.group)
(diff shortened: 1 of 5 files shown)


desk/app/groups.hoon
```diff
@@ -709,7 +709,7 @@
     ``groups+!>(groups-2)
   ::
       [%x %v1 %groups ~]
-    ``groups-1+!>((~(run by groups) tail))
+    ``groups-1+!>(`groups:v5:g`(~(run by groups) tail))
```
Fang-, line 712: Yes, very good to have clear types on all scry result vases.

desk/app/diary.hoon
```diff
@@ -202,11 +202,11 @@
     (toggle-post toggle)
   ::
       %leave-old-channels
-    =/  groups-path  /(scot %p our.bowl)/groups/(scot %da now.bowl)/groups/noun
-    =/  groups  .^(groups:g %gx groups-path)
+    =/  groups-path  /(scot %p our.bowl)/groups/(scot %da now.bowl)/groups/groups
+    =/  groups  .^(groups:v2:g %gx groups-path)
```
Fang-, line 206: `/noun` at the end _should_ still work, right? Arguably that's more robust against mark renaming shenanigans, making it preferable, but all of that _should_ be static in practice so it _shouldn't_ matter either way.

Fang- (approved):
Yep, these look like stragglers. I'll say it again: surely it can't be _that_ big of a research project to do code analysis to find all possible scries in a codebase... (^:

desk/app/diary.hoon
mikolajpp: It would work, but so far I have followed a convention of using the mark whenever it is available. (There are untyped endpoints, and for those `noun` is the proper choice). 

I can see one other reason to prefer `noun`: there is less typing when upgrading endpoints, since then the only change is bumping the version number in the path and type.

PR #4491, opened 2025-03-05
embeds: fix issue with rendering embeds inline

patosullivan:
The actual issue didn't end up being the height of the twitter embeds, the issue was that if you posted a link to some embeddable content *on the same line* as some other text, then the webview would render within the Text component and it would either a) break out of that container on iOS or b) push the height of that container tall enough to fit the embed webview on Android (which also looked bizarre).

To fix this, embeds are now extracted as standalone block elements, separate from text. Links from trusted providers are now converted to block-level embeds.

Also:

I hardened are height change detection and fixed an issue that was causing crashes on iOS (changing the key on Webview based on the webViewHeight was causing crashes in some instances, moving the key to the parent View ensures that we still force an adjustment when necessary and prevents the crash, the WebView will just expand to fill the parent).

I created a maxAllowedHeight value that will be calculated based on the viewport height. If a tweet is too big, we'll hide the media in the tweet (video, pictures, external links) to constrain the height a bit.

https://github.com/user-attachments/assets/df8f3f78-f589-4335-9964-7018f3a88483

### DIFF
--- a/packages/app/ui/components/PostContent/contentUtils.tsx
+++ b/packages/app/ui/components/PostContent/contentUtils.tsx
@@ -5,6 +5,8 @@ import * as ub from '@tloncorp/shared/urbit';
 import { useContext, useMemo } from 'react';
 import { createStyledContext } from 'tamagui';
 
+import { trustedProviders } from '../Embed/EmbedContent';
+
 export interface ContentContextProps {
   isNotice?: boolean;
   onPressImage?: (src: string) => void;
@@ -96,6 +98,8 @@ export type VideoBlockData = {
 
 export type EmbedBlockData = {
   type: 'embed';
+  url: string;
+  content?: string;
 };
 
 export type ReferenceBlockData = api.ContentReference;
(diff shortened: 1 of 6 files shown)


patosullivan:
Just realized that making embeds render as a block will effect the way we render embeddable content in gallery posts in gallery channels (not detail view). Looking into that now.

packages/app/ui/components/PostContent/contentUtils.tsx
```diff
@@ -277,6 +278,95 @@ function convertTopLevelInline(verse: ub.VerseInline): BlockData[] {
...
+    } else {
+      // Not a link, add to normal paragraph
+      currentSegment.push(inline);
+    }
+  }
+
+  // Flush any remaining inlines as a paragraph
+  flushSegment();
+
+  // If no blocks were created (only had non-link inlines), create a single paragraph if needed
```
arthyn, line 343: not sure I understand this because the previous line shoves a block in if we have pending segments and seems to do same logic

patosullivan:
> Just realized that making embeds render as a block will effect the way we render embeddable content in gallery posts in gallery channels (not detail view). Looking into that now.

fixed in latest commit

packages/app/ui/components/PostContent/contentUtils.tsx
patosullivan: Yep, you're right. I meant to remove this after adding flushSegment. Fixed in latest commit.

arthyn (approved):
lgtm!

PR #4492, opened 2025-03-06
desktop: prevent navigating to post on menu press in galleries

patosullivan:
fixes tlon-3764
fixes tlon-3767

We weren't disabling the onPress for the Pressable when the popover was open. Any selection in the menu would cause the user to be navigated to the post detail view.


arthyn (approved):
lgtm

PR #4493, opened 2025-03-06
chat details: fix navigation issues

patosullivan:
fixes tlon-3769

Now:

- If we navigate to `ChatDetails` (group settings) on desktop we'll also render the group channels list in the sidebar.

- If we press the back button (on desktop) while we're in ChatDetails, and we're in the details for a group, we'll just navigate to the group.

- If we edit the group meta and press save, we'll be navigated back to ChatDetails rather than back to the group.


arthyn (approved):
lgtm

PR #4495, opened 2025-03-06
groups: avoid %kick-ing empty list of paths

Fang-:
Kicking empty list of paths means "kick subscription that triggered this event". This does mean that when constructing a path list of unknown size, you have to catch the empty list case to avoid falling into that specialized behavior.

Here, we check for the empty list, and avoid emitting kicks (or facts) in that case.

(In practice, this resulted in the kick going into clay during an update reload, which failed without providing a clear trace or legible failure reason. Notably, clay's error handling only handles errors on the `/drip` wire, crashing in other cases. Potentially not the right approach...)

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -1025,10 +1025,13 @@
       ~
     `path
   ?+    -.sign  cor
-      %kick  (give %kick matching ~)
+      %kick
+    ?~  matching  cor
+    (give %kick matching ~)
   ::
       %watch-ack
     ?~  p.sign  cor
+    ?~  matching  cor
     (give %kick matching ~)
   ::
       %fact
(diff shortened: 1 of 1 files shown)


jamesacklin (approved):
Works on my machine!

arthyn (approved):
10/10 doctors recommend

mikolajpp:
Leaving a comment here for posterity: it is likely these seemingly unused subscriptions were kicked by the migration code in `lib-negotiate`, which kicks _every_ subscription during `+on-load`. Since the %groups agent probably never pruned its connections, some ships would carry over the unused subscriptions from distant past. 

Yet another reason to do a thorough cleanup of the %groups agent (on my todo list!)

PR #4496, opened 2025-03-06
desktop: show contact nicknames in emoji hover tooltip

patosullivan:
fixes tlon-3745

we were just using the bare contactIds before because we weren't including contact data with the reactions subquery. this updates the query def for `getPosts` to pull contact data with reactions, and updates the ReactionsDisplay and component and related types appropriately.


arthyn (approved):
lgtm

PR #4497, opened 2025-03-06
native: swipe through sibling posts in gallery

davidisaaclee:
Adds the ability to swipe through posts in a gallery `PostScreen`.
There are some other UI and behavior changes added by @dnbrwstr (see the first commit in this PR) – most visible is changing gallery text posts to have a square frame and `$secondaryBackground` background fill.

The view tree is roughly like this now:
```jsx
<PostScreen>
  <KeyboardAvoidingView>
    <Header />
  
    {isCarousel ? (
      <PostCarousel posts={siblingsOf(parentPost)}>
        {eachPost => <SinglePostView post={eachPost} />}
      </PostCarousel>
    ) : (
      <SinglePostView post={parentPost} />
    )}
    
    <ReplyInput />
  </KeyboardAvoidingView>
</PostScreen>
```

This is a big diff because
- `PostScreenView` had a lot of baked-in assumptions about displaying a single post (which, in a carousel setting, would affect basic operation and would incur more loads than needed)
- `PostScreenView` receives a big bundle of props, many of which also assumed that it was only operating on a single post

Highlights:
- I pulled much of the single-post logic into `SinglePostView`
- There were a lot of passed props that could've been pulled into `SinglePostView`, but would've made an even huger diff:
  - some props could just be pulled from contexts in the child, like navigation callbacks
  - there are a lot of props pulled from a single `useChannelContext` at `PostScreen`: I could move this into `SinglePostView`, but I see that there is state managed in `useChannelContext`, and I don't know how managing multiple instances of this hook would work. (I think we should add an explicit provider for this "context," maybe the existing `ChannelContext`, and share state within the subtree.) For these props, I just pulled them into a `channelContext` object so I could pass them around easily and hopefully aid clean up later.

## Screen recordings
`CarouselPostScreenContent` fixture (demonstrating that loading more posts should maintain scroll position)

https://github.com/user-attachments/assets/ec07365c-cb2a-4ca1-87cc-af35e53c09c7

Swiping in real gallery; sending replies; following refs from/to gallery
https://github.com/user-attachments/assets/8ca62b5b-32c4-4c19-8136-5b8a4310e12e

I noticed these bugs / weird behavior which are also present in the most recent release:
- Following a ref from gallery post to a reply on a gallery post loads just that reply by itself
- Sending a reply in a gallery thread does not autoscroll to the new message
- Following a ref from a gallery reply to another reply in that same thread fails to scroll to that reply (code existed for this but doesn't work)


packages/app/features/top/PostScreen.tsx
```diff
@@ -136,57 +119,44 @@ export default function PostScreen(props: Props) {
...
+      onPressRetry={handleRetrySend}
+      onGroupAction={performGroupAction}
+      goToDm={handleGoToDm}
+      negotiationMatch={negotiationStatus.matchedOrPending}
+      headerMode={headerMode}
+      // NB: If we're showing posts in a carousel, all carousel items share the
+      // same editingPost.
+      editingPost={editingPost}
+      setEditingPost={setEditingPost}
+      editPost={editPost}
```
davidisaaclee, line 159: emphasizing this comment – realizing now that I think this breaks saving drafts per thread. going to investigate more now!

jamesacklin: Makes sense to me, thank you

packages/app/ui/components/PostScreenView.tsx
```diff
@@ -169,28 +167,31 @@ export function PostScreenView({
...
-        flatListRef.current?.scrollToIndex({
-          index: anchorIndex,
-          animated: false,
-          viewPosition: 0.5,
-        });
-        return;
-      }
+  // TODO: lost in changes - but this is not working for me in latest
+  // release anyways (tried by following a ref to a post in the same
+  // thread, which has been loaded & is offscreen)
```
davidisaaclee, line 172: emphasizing (will delete before merge)

packages/app/ui/components/PostScreenView.tsx
```diff
@@ -205,6 +206,264 @@ export function PostScreenView({
...
+                  channel={channel}
+                  goBack={handleGoBack}
+                  mode={headerMode}
+                  showEditButton={showEdit}
+                  goToEdit={handleEditPress}
+
+                  // When adding the ability to swipe through posts, we lost
+                  // the ability to show a spinner based on loading posts. I
+                  // don't think we were ever really showing this though, as we
+                  // only spun while waiting for DB load, not network load.
```
davidisaaclee, line 250: emphasizing (will delete before merge)

packages/app/ui/components/PostScreenView.tsx
```diff
@@ -205,6 +206,264 @@ export function PostScreenView({
...
+  {
+    channel: db.Channel;
+  },
+  'channel' | 'group' | 'title' | 'showSearchButton' | 'post'
+>) {
+  const isChatChannel = getIsChatChannel(channel);
+
+  const { focusedPost: parentPost } = useContext(FocusedPostContext);
+
+  // TODO: Swap title (and other header properties) based on currently-focused post in carousel
```
davidisaaclee, line 324:
```suggestion
```

packages/app/features/top/PostScreen.tsx
davidisaaclee: Drafts not being saved in their respective thread was not caused by `editingPost`, but by the `get/store/clearDraft` callbacks – fixed in [ccc7c34](https://github.com/tloncorp/tlon-apps/pull/4497/commits/ccc7c343b45784eae248449b86b80ac75135fe71) and [4c4b211](https://github.com/tloncorp/tlon-apps/pull/4497/commits/4c4b21189da1df1c8334d87c6d26d4e616905ce8)

I think we can punt on `editingPost` because we don't allow editing in any carousel contexts (i.e. galleries).

packages/app/ui/components/PostScreenView.tsx
latter-bolden: Agree it's not working for refs in the same thread. Even with this commented out, ref clicks appear to behave the same as on `develop`. I think this can be removed.

packages/app/ui/components/PostScreenView.tsx
latter-bolden: I'm fine removing for now

latter-bolden (approved):
Pending passing CI and removing those noted comments, I think this is good to go.

The `FlatList` approach is interesting, wouldn't have guessed that's how it's wired up. Appreciate the featureful fixtures.

PR #4498, opened 2025-03-06
activity: only run fix once

arthyn:
We noticed this was making reloading activity take really long. This really doesn't need to be run on every load just between state 7 and 8.


mikolajpp:
great catch, PR looks good.

mikolajpp (approved):

PR #4500, opened 2025-03-06
heap/diary: removing unnecessary operations on deprecated agents

arthyn:
OTT, while were debugging found that these are doing a lot of things that simply aren't necessary anymore. This just removes those and no-ops.


mikolajpp (approved):
The changes here look correct. The ability to migrate old content to %channels is probably only thing we care about with these agents, and it remains unaffected.

PR #4502, opened 2025-03-07
chat/channels: fix new heads endpoints

arthyn:
I found this testing out `release-channels-updates` basically we had the wrong types for the v1 heads endpoint and needed to add a new v2 endpoint with the new types and new mark. Presented as never seeing any dates/times on the chat list until visiting a particular channel which makes sense.


desk/app/chat.hoon
```diff
@@ -855,15 +855,15 @@
       ?^  tim=(slaw %da i.t.t.path)  `u.tim
       `(slav %ud i.t.t.path)
     :^  ~  ~  %chat-heads
-    !>((old-heads-3 (heads since)))
+    !>(`chat-heads:old-3`(old-heads-3 (heads since)))
```
mikolajpp, line 858: Not that anything is wrong with this, but generally I have refrained from casting where the function is meant to enforce the type. Here, `+old-heads-3` is never going to change in its output type, that is `chat-heads:old-3`.

mikolajpp (approved):
Looks good, thanks for catching this.

PR #4503, opened 2025-03-07
chat-list: don't show options on invites

arthyn:
The triple dot icon was overlapping the invite bubble and really didn't make sense to have in the first place. This disables options on pending items.


latter-bolden (approved):

jamesacklin (approved):
LGTM

PR #4504, opened 2025-03-07
Telemetry: Fix model capture & log chatlist taps

latter-bolden:
OTT

Fixes a few limitations that cropped up while analyzing remote telemetry. Also fixes a bug where we were never calling an intended `ChannelScreen` focus effect.


jamesacklin (approved):
LGTM

PR #4505, opened 2025-03-07
embeds: remove key on parent view that was causing re-renders

patosullivan:
This wasn't causing an issue on every tweet, and it didn't seem to cause an issue at all on iOS, but it was causing an issue where many tweet embeds would reload continuously on Android. I initially added it in the first iteration of embeds to force the component to remount when we got a height update (and originally it was a key on the WebView itself), but in my testing today it looks like it's unnecessary and just giving us trouble.


jamesacklin (approved):
Seems fine to me. No regression on iOS or web?

patosullivan:
> Seems fine to me. No regression on iOS or web?

change only effects native (we have a separate component for web), and I tested on both iOS and Android, looks fine.

PR #4506, opened 2025-03-10
desktop: add electron app

patosullivan:
This includes everything necessary to get the desktop app up and running with electron.

This isn't an MVP yet, but it works. You can run it as dev, or build the app for macOS. You can log into the app and use it with any ship.

I'd still like to get it to where we can re-enable `webSecurity`.

I'm sure there's still more cleanup that can be done here, but I figured we could cut off this initial spike at this point.

Note: there's still some sqlite weirdness going on (FOREIGN_KEY constraint error when re-opening the app. This has got to be an issue in the renderer somewhere, fixing that is my next step after this gets in).

Also, you can't use the dev app and the prod app on the same machine at the moment (you'll run into issues with sqlite since it's using the same file), I'll look into fixing that when I fix the other sqlite bug.


patosullivan:
Looking into the test failure issue

jamesacklin:
I think we're good here for MVP, let's merge and proceed

PR #4507, opened 2025-03-10
profile-navigator: need to limit height for scrolling

arthyn:
OTT fixes TLON-3775

### DIFF
--- a/packages/shared/src/db/queries.ts
+++ b/packages/shared/src/db/queries.ts
@@ -119,7 +119,7 @@ const POST_RELATIONS_DEFAULT = {
   reactions: {
     with: {
       contact: true,
-    }
+    },
   },
   threadUnread: true,
   volumeSettings: true,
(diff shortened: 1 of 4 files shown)


packages/shared/src/db/queries.ts
```diff
@@ -1748,13 +1760,19 @@ export const updateChannel = createWriteQuery(
...
-    await ctx.db.delete($chatMembers).where(eq($chatMembers.chatId, channelId));
-    await ctx.db.delete($channels).where(eq($channels.id, channelId));
+export const deleteChannels = createWriteQuery(
+  'deleteChannels',
+  async (channels: string[], ctx: QueryCtx) => {
+    logger.log(`deleteChannels`, channels);
+    for (const channelId of channels) {
+      withTransactionCtx(ctx, async (txCtx) => {
+        await txCtx.db.delete($posts).where(eq($posts.channelId, channelId));
+        await txCtx.db
```
latter-bolden: I think cascading deletes like this can be specified in the schema relations

latter-bolden (approved):

PR #4508, opened 2025-03-10
negotiate: make note of simulated %kick danger

Fang-:
During the library's +on-load handling, it may simulate %kick events by injecting them directly into the inner agent. This means parts of the inner agent other than +on-load may be getting called during agent loading. This is risky, because scries are brittle during loading scenarios: other agents might not have been brought back up yet, etc.

Here we leave a warning so that we don't forget about this detail.


arthyn (approved):
lgtm

PR #4511, opened 2025-03-10
groups: ignore %kicks for deleted groups

mikolajpp:
A %kick on a subscription to a group that had been deleted would crash. This caused a problem in conjunction with the use of lib-negotiate: during first load, the library sends a %kick on every subscription in order to force protocol negotiation. If a subscription was not properly cleaned-up by the agent, the %kick will trigger a crash, interrupting the update.


arthyn:
changing the base branch to staging so that it goes to both develop and staging

arthyn (approved):
lgtm

PR #4513, opened 2025-03-10
embeds: webview rerender jihad

patosullivan:
fixes tlon-3778 by:

- replacing multiple useState hooks with one useReducer for atomic state updates
- eliminating the unnecessary debounced height setter, replaced with direct dispatch calls
- reducing handler dependencies by using refs
- making onLayoutHandler dependency free via refs
- memoizing everything

This should fix the cascading state updates that were causing the re-renders/visual flicker we were seeing in threads (and when loading a chat channel where a tweet was near the bottom of a chat, or the gallery details view).

### DIFF
--- a/packages/app/ui/components/Embed/EmbedWebView.tsx
+++ b/packages/app/ui/components/Embed/EmbedWebView.tsx
@@ -1,5 +1,12 @@
 import { LoadingSpinner } from '@tloncorp/ui';
-import { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import {
+  memo,
+  useCallback,
+  useEffect,
+  useMemo,
+  useReducer,
+  useRef,
+} from 'react';
 import {
   Dimensions,
   LayoutChangeEvent,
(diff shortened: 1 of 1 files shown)


packages/app/ui/components/Embed/EmbedWebView.tsx
```diff
@@ -32,15 +39,63 @@ interface WebViewMessageData {
...
+  action: WebViewAction
+): WebViewState {
+  switch (action.type) {
+    case 'HIDE_MEDIA':
+      return { ...state, hideTweetMedia: true };
+    case 'SET_HEIGHT':
+      return { ...state, webViewHeight: action.height };
+    case 'LOADING_COMPLETE':
+      return { ...state, isLoading: false };
+    case 'HIDE_MEDIA_AND_UPDATE_HEIGHT':
```
arthyn, line 66: this case seems unused

latter-bolden (approved):
Code LGTM!

arthyn (approved):
lgtm

PR #4515, opened 2025-03-10
Chat list: Make Groups Swipeable Again

patosullivan:
fixes tlon-3787


jamesacklin (approved):
You are most certainly going to need that for groups as well. Makes sense to me!

PR #4516, opened 2025-03-10
GalleryPostDetailView: image tap navigation to lightbox

jamesacklin:
Fixes TLON-3791 by adding tap-to-lightbox functionality in the Gallery detail view.

- First and foremost, adds navigation to ImageViewer when an image is pressed
- Memoizes image detection so we can be sure to have the src in hand at all times
- Add fallback image press handling if the navigation fails
- Plenty of logging

In action, with swiping:

https://github.com/user-attachments/assets/4d98e1b1-da3c-4f11-ba4d-4dc50c48b538


patosullivan (approved):
looks good, works well (fixed the desktop carousel issue with latest commit)

PR #4517, opened 2025-03-11
Better Groups: implement preview response status

mikolajpp:
When requesting a group preview, no information about possible cause of a failure is sent back to the requester. There are two main possibilities:
(1) the group is missing
(2) the requester does not have preview access permissions

This PR introduces a `$preview-response` type, an `(each preview access-error)`, which allows for a group preview error to be signaled back to the requester. 

Since this affects the state type, we also take opportunity to deprecate and remove the `cast` functionality.


desk/app/groups.hoon
```diff
@@ -1435,49 +1429,59 @@
...
+        (emit %give %fact ~ group-preview-1+!>(preview))
+      ::
+          %v2
+        ?.  allow
+          ?:  secret.group
+            ::  conceal secret private group
+            (emit %give %fact ~ group-r-preview-0+!>(|+%missing))
+          (emit %give %fact ~ group-r-preview-0+!>(|+%forbidden))
+        ::
+        (emit %give %fact ~ group-r-preview-0+!>(&+preview))
```
Fang-, line 1483: I think we generally don't do the `r-` prefix for mark names. Also, that prefix is for "response" types, that go from the subscriber agent to the frontend client.

That pattern/convention isn't really present in the groups types anyway, because we haven't rewritten them since inventing it. I wouldn't be opposed to starting to use it, but the usage here as-is would be incorrect/misleading I think.

(Also we're gonna want casts inside these `!>` for safety.)

Fang- (commented):
The change from #4511 shows up in the diff, but the commit from there doesn't. Probably staging vs develop branch shenanigans, no big deal, but worth noting that it slipped in.

A comment about mark naming, but seems good otherwise!

desk/app/groups.hoon
```diff
@@ -1435,49 +1429,63 @@
...
+          %v1
+        =/  =preview:v5:g  preview
+        (emit %give %fact ~ group-preview-1+!>(preview))
+      ::
+          %v2
+        ?.  allow
+          ?:  secret.group
+            ::  conceal secret private group
+            ::
+            =/  pev  `preview-response:v6:g`|+%missing
```
Fang-, line 1481: If you're gonna `=/` might as well make use of it. (;

```suggestion
            =/  pev=preview-response:v6:g  |+%missing
```

Fang- (approved):
Changes look good, thanks! Style nit below, but no big deal.

desk/app/groups.hoon
mikolajpp: oh, definitely!

arthyn (approved):
lgtm!

arthyn (changes requested):
I started looking at integrating with this and something I think is missing here. We need to update the scries that use gangs because the types won't be what they expect. And we'll need new versions of the scries here and in `%groups-ui` so that we can pull the new gang data on init.

mikolajpp:
I pushed a commit implementing missing scry endpoints, and also a watch endpoint for latest version of gangs.

arthyn (approved):
lgtm!

mikolajpp:
@arthyn should this be blocked by TLON-3800? I have posted a comment there, it seems any group update will trigger a group channels join.

arthyn:
@mikolajpp will write Mark and I's findings on 3800, but I guess to be safe lets hold off yes

mikolajpp:
I pushed a commit to fix wrong protocols declaration, and also fix a wrong subscription endpoint for %contacts.

Fang- (approved):
Changes look good. This does affect version negotiation (for real this time (^: ) so we'll want to keep an eye on things in pre-release just in case.

PR #4520, opened 2025-03-11
embeds: fix issue with opening embed links on web

patosullivan:
we need to add a directive that allows the iframe to open new tabs/windows outside of the iframe context (on web, obv).

https://issues.chromium.org/issues/40275209

h/t to jackson for finding this.


jamesacklin (approved):
That's a doozy. Good catch!

PR #4521, opened 2025-03-11
groups: Add support for blank channel group metadata navigation

jamesacklin:
Fixes TLON-3794. Updates the GroupSettingsStackParamList type to include an optional fromBlankChannel parameter in the GroupMeta route. GroupMetaScreen now checks for this parameter and handles the navigation differently.

Also updates the useChatSettingsNavigation and ChatOptionsContext to accept the fromBlankChannel parameter. Finally, changes EmptyChannelNotice to pass true when the user taps the Customize button.

Therefore, when a user taps the "Customize" button in a new, blank channel:

1. We'll pass fromBlankChannel=true through the navigation stack
2. When they press "Save", the GroupMetaScreen will detect that fromBlankChannel is true
3. It will go back to the previous screen (the blank channel) instead of navigating to the group settings screen

Desktop behavior is unchanged.

### DIFF
--- a/packages/app/features/groups/GroupMetaScreen.tsx
+++ b/packages/app/features/groups/GroupMetaScreen.tsx
@@ -23,7 +23,7 @@ type Props = NativeStackScreenProps<
 };
 
 export function GroupMetaScreen(props: Props) {
-  const { groupId } = props.route.params;
+  const { groupId, fromBlankChannel } = props.route.params;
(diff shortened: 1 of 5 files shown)


patosullivan (approved):
Makes sense!

jamesacklin:
> I was looking at this too. Let me try this before you merge, have a corner case in mind

Ack! Didn't see this. Feel free to revert or commit straight to develop.

PR #4522, opened 2025-03-11
embeds: use ContentRenderer to render fallback links in EmbedContent

patosullivan:
Our fallback links were rendering with tiny text. This didn't matter when twitter was working, but now we're seeing twitter return failing SSL certs (they seem to be having trouble), so we're seeing the fallback links a lot.


arthyn (approved):
lgtm

PR #4524, opened 2025-03-12
calm: disable avatars should not turn off group icons

jamesacklin:
Fixes TLON-3792 by explicitly always showing image avatars for group icons. This way, the user can turn off user avatars and still see group icons. Adds an optional prop so we can explicitly ignore calm settings for other icons in the future, should we need it.

Furthermore, I noticed that contact and profile avatars were not affected by calm settings, so I made sure the AppDataContextProvider had the user's calm settings injected.

Tested on iOS simulator and web-new development.


patosullivan (approved):
lgtm after the type def fixes

PR #4526, opened 2025-03-12
channel: fix mark read on desktop

arthyn:
We weren't accounting for the fact that we're rendering in a stack which means if we load a channel once and then navigate to a different one, that original channel render is still in memory. When we come back to it, the `hasLoaded` flag is still true, so we don't actually mark read. This adds a flag for `isFocused` so we know if this channel comes to focus and can mark read appropriately.


latter-bolden (approved):

jamesacklin (approved):
Good old useIsFocused() saves us again. Makes perfect sense to me

PR #4528, opened 2025-03-14
desktop: add apple signing, tweak gh action to only deploy with desktop tags

patosullivan:
and specify working directories


jamesacklin (approved):
Looks very good to me, thank you!

PR #4533, opened 2025-03-17
electron: fix activity foreign key constraint issue

patosullivan:
We were unable to insert rows into the `activity` table on desktop because better-sqlite3 enforces foreign key constraints by default (our other sqlite implementations don't) and we use a composite primary key in `activity` that we don't properly reference in `activity_events_contact_group_pins`.

Simplest solution here is just to match the other sqlite implementations (this is the default for sqlite in general) and disable the strict fk enforcement.

This PR also fixes an issue with changes not making it into the `handleChange` function on desktop (we were trying to parse a string that was actually already JSON into JSON) and removes some debug logging from the electron-store-wrapper that I had put in earlier.

More on the foreign keys: https://www.sqlite.org/pragma.html#pragma_foreign_keys

### DIFF
--- a/apps/tlon-desktop/electron-store-wrapper.cjs
+++ b/apps/tlon-desktop/electron-store-wrapper.cjs
@@ -83,13 +83,6 @@ module.exports = {
       const store = await initStore();
       if (!store) return null;
       const value = store.get(key);
-      if (key === 'encryptionKey' || key === 'encryptedAuthCookie') {
-        console.log(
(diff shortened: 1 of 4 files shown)


jamesacklin (approved):
Seems good, you're just adding loggers to the main shared lib?

patosullivan:
> Seems good, you're just adding loggers to the main shared lib?

That was the only change to changeListener, yeah

PR #4534, opened 2025-03-17
Activity: Sync bell across clients

latter-bolden:
OTT

Uses settings store to persist your Activity seen marker. Clients pull this down instead of storing a local one.

To test, try running new web and mobile simulator locally. Try navigating to activity on one and see the bell clear on the other.

Fixes TLON-3687


patosullivan (approved):
Code looks good, makes sense

PR #4535, opened 2025-03-17
desktop: add notifications (and icons)

patosullivan:
Uses an activity subscription to create desktop notifications.

Adds icons for all three platforms.

fixes tlon-3816


latter-bolden (approved):
Very cool! Code looks good

PR #4536, opened 2025-03-18
Fix missing latest posts

latter-bolden:
The prod build last night included a DB migration to save the new activity read marker setting. This uncovered a bug with our latest posts logic — when the migration wipes all SQL data, the keyval store keeps its existing `headsSyncedAt` timestamp. This means `syncLatestPost` thinks it already has messages that it doesn't.

To fix, we'll now clear out that key anytime we purge the DB. I tried upgrading from an older version to confirm it works, but wanted to double check with y'all if purge is always called in these migration scenarios?

Another thought I had: if we have more of these sync optimization flags, it might be worth considering an optional `persistAfterMigration` storage item flag (analogous to `persistAfterLogout`).

### DIFF
--- a/packages/app/lib/nativeDb.ts
+++ b/packages/app/lib/nativeDb.ts
@@ -1,5 +1,6 @@
 import { open } from '@op-engineering/op-sqlite';
 import { escapeLog } from '@tloncorp/shared';
+import * as kv from '@tloncorp/shared/db';
 import { schema, setClient } from '@tloncorp/shared/db';
 
 import {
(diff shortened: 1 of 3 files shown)


packages/app/lib/nativeDb.ts
```diff
@@ -73,6 +74,10 @@ export class NativeDb extends BaseDb {
     this.connection.delete();
     this.connection = null;
     this.client = null;
+
+    // reset values related to tracking db sync state
+    await kv.headsSyncedAt.resetValue();
```
patosullivan, line 79: we should probably add this to `purgeDb()` on in `electronDb`, too, since we actually use these values there

patosullivan (approved):
makes sense, let's just update electrondb as well before merging

packages/app/lib/nativeDb.ts
latter-bolden: good catch!

PR #4541, opened 2025-03-19
gallery: image posting UX

jamesacklin:
Fixes TLON-3761. This nearly completely overhauls GalleryInput to handle the UX changes we want for Gallery image uploads. I am preserving BigInput's current behavior until we make more progress on the Notebook design and back-end spec.

- Only prompts the user for camera or photo library access permission if they tap the corresponding option, rather than slamming them with permission requests for both.
- Does not use "cover" image fitting in the gallery image upload preview; uses "fit" method instead so that the longest edge occupies 100% of either the height or width of the preview canvas.
- Moves the "Post" button to the header of the upload preview screen and does away with the confusing up-arrow button. 
- Eliminates the "…" and other controls that are erroneously present in the header of the upload preview screen.
- Converts the image caption input to a native text input, since the caption is a single inline anyway.
- Correctly routes the user to a version of the upload preview screen when they edit an image post; currently they're navigated to BigInput where they can only edit the caption and cannot see the image.

Progress thus far:

https://github.com/user-attachments/assets/bfb6ac79-a5a4-46e2-a69b-383107baa357


patosullivan (approved):
Looks good. I did find an issue with the "Edit post" button in the popover/context menu on the post not working on desktop, but I don't think that's related to this PR.

PR #4542, opened 2025-03-19
Web: Fix hosted uploads

latter-bolden:
When checking hosted upload eligibility, we confirm the ship's URL is Tlon controlled. This check was failing on web because the http client's url property is not explicitly set. This updates the method to fall back to window location or env configuration during development.

Fixes TLON-3246


jamesacklin:
Looks good to me, seems pretty safe.

arthyn (approved):
lgtm

PR #4543, opened 2025-03-20
Web: Mention keyboard controls

latter-bolden:
Adds a selected mention index that can be controlled with arrow keys. Handles enter to select and escape to dismiss.

@jamesacklin flagging you for UX review.

Fixes TLON-3783

### DIFF
--- a/packages/app/ui/components/BareChatInput/index.tsx
+++ b/packages/app/ui/components/BareChatInput/index.tsx
@@ -230,6 +230,7 @@ export default function BareChatInput({
     mentions,
     setMentions,
     showMentionPopup,
+    handleMentionEscape,
   } = useMentions();
   const maxInputHeight = useKeyboardHeight(maxInputHeightBasic);
   const inputRef = useRef<TextInput>(null);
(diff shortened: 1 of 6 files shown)


packages/app/ui/components/BareChatInput/index.tsx
```diff
@@ -737,6 +758,8 @@ export default function BareChatInput({
             ...(isWeb ? placeholderTextColor : {}),
             ...(isWeb ? { outlineStyle: 'none' } : {}),
           }}
+          // Hack to prevent @p's getting squiggled on web
+          spellCheck={!mentions.length}
```
jamesacklin, line 762: Thanks

jamesacklin (approved):
Sold

packages/app/ui/components/BareChatInput/index.tsx
```diff
@@ -670,16 +673,33 @@ export default function BareChatInput({
         return;
       }
 
+      if (
+        keyEvent.key === 'ArrowUp' ||
+        (keyEvent.key === 'ArrowDown' && showMentionPopup)
```
patosullivan, line 678: do we want the conditional to work like this? seems like it should be (up or down) and showMentionPopup, not up or (down and showMentionPopup)

patosullivan (approved):
Overall, looks good, one question on a conditional

packages/app/ui/components/BareChatInput/index.tsx
latter-bolden: Yeah, that's definitely mistyped. Nice catch!

PR #4544, opened 2025-03-20
Settings Table Refactor

latter-bolden:
The release candidate is crashing on logout. The issue is a race condition with clearing the current user ID before all listeners for the settings table are unmounted.  Current user ID is the primary key for the settings table, but we don't actually store other user's settings (table only ever has the singleton row).

To fix this I changed the primary key for the settings table to a static string. I also updated the query methods to assume you always want to read/write that singleton without needing to manually pass in a row ID.

Fixes TLON-3832


patosullivan (approved):
Looks good, makes sense

PR #4547, opened 2025-03-20
Update Expo to 51, React Native to 0.74

dnbrwstr:
OTT, fixes SSE for Android in debug mode.

Tested locally + confirmed expo builds still work.


apps/tlon-mobile/package.json
```diff
@@ -69,38 +69,38 @@
...
-    "expo-device": "~5.9.4",
-    "expo-file-system": "~16.0.9",
-    "expo-haptics": "~12.8.1",
-    "expo-image": "^1.10.6",
-    "expo-image-manipulator": "~11.8.0",
-    "expo-image-picker": "~14.7.1",
-    "expo-linear-gradient": "^12.7.2",
-    "expo-localization": "~14.8.4",
-    "expo-media-library": "~15.9.2",
-    "expo-modules-core": "~1.11.14",
```
davidisaaclee, line 90: It's surprising to me that expo-modules-core went away (I'm guessing all of these were auto-updated by expo upgrade tool, so I'd assume it knows what it's doing) – calling it out here in case this causes issues e.g. on an empty pnpm cache

davidisaaclee:
built and ran iOS locally, wfm as well

apps/tlon-mobile/package.json
dnbrwstr: Ya, this was a suggestions from `expo-doctor` -- core is supposed to be a dependency of the expo package, not installed directly.

davidisaaclee (approved):

PR #4548, opened 2025-03-20
image-viewer: modal to avoid shrinking drawers

arthyn:
Fixes TLON-3801 by using a modal to pop the lightbox. This does hide the top level drawer but I figured that was ok. I also removed some cases where images weren't centered or were distorted.

### DIFF
--- a/packages/app/navigation/desktop/HomeNavigator.tsx
+++ b/packages/app/navigation/desktop/HomeNavigator.tsx
@@ -44,13 +44,12 @@ export const HomeNavigator = () => {
         const state = navigation.getState();
         const routes = state.routes[state.index].state?.routes;
         const currentScreen = routes?.[routes.length - 1];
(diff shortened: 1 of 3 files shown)


patosullivan:
Fixes the weird issues we see with shrinking drawers, but I'm noticing that if I view an image from the post in the backscroll and then exit back to the chat channel, I'm brought back to the bottom of the chat

arthyn:
I'm seeing same behavior on develop in a channel that has unreads. I think we're just doing this any time the channel is routed to

patosullivan:
> I'm seeing same behavior on develop in a channel that has unreads. I think we're just doing this any time the channel is routed to

right, but the issue is that, the way the current image viewer works, the user will be dropped back into the chat where they left off after they exit the image viewer.

patosullivan (approved):
LGTM! Ignore my earlier comments, didn't realize I tested against channels with unreads that have this same scroller behavior with the current image viewer.

PR #4554, opened 2025-03-24
desktop: better "empty" state

patosullivan:
OTT, looks like this:
Home App Groups


jamesacklin (approved):
Cool, this is just HomeSidebar? Probably good to scope just to this

latter-bolden (approved):

PR #4560, opened 2025-03-25
gallery: fix issue with shorter gallery posts in index

patosullivan:
fixes tlon-3779

we were inserting `<PostBlockSeparator/>` unnecessarily after gallery posts, which shoves an empty View of a defined height after a post or set of posts. Inserting that component makes sense in chat, but not in galleries or notebooks.

### DIFF
--- a/packages/app/ui/components/Channel/Scroller.tsx
+++ b/packages/app/ui/components/Channel/Scroller.tsx
@@ -261,6 +261,7 @@ const Scroller = forwardRef(
         );
         const isLastPostOfBlock =
           post.type !== 'notice' &&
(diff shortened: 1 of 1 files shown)


jamesacklin:
Makes sense. Does this affect the last row in a gallery where there might be a single block? I remember us having to shim everything over to the left, else the last block would be centered.

patosullivan:
> Makes sense. Does this affect the last row in a gallery where there might be a single block? I remember us having to shim everything over to the left, else the last block would be centered.

it's unrelated to that

jamesacklin (approved):

PR #4562, opened 2025-03-25
threads: use day and time if no title

arthyn:
Something that's bothered me since we did the new client is that if you're in a thread in a DM you just see "Thread:" which just looked broken. Trying just "Thread" with no colon also felt weird. Not sure if this is what we want to go with, but seemed reasonable:

image


jamesacklin (approved):
Perfect, no notes

PR #4571, opened 2025-03-26
chat options menu: remove animation prop from popover (again)

patosullivan:
fixes tlon-3850 (finally)

missed a spot when removing the animation props earlier


latter-bolden (approved):
🥾🪲

PR #4572, opened 2025-03-26
desktop: fix context menu issues in gallery and notebook

patosullivan:
fixes tlon-3836

Fixes the original issue by making sure we pass `onPressEdit` to the `onEditProp` in ChatMessageActions in GalleryPost.

Also does the same thing for notebooks.

While working on this, I noticed that pressing any menu item in the popover for a notebook post in a notebook channel would redirect the user to the detail view for the post. We weren't disabling the onPress on the Pressable in NotebookPost when the popover was open. This also fixes that.

Also cleans up the props on NotebookPost.


latter-bolden (approved):
Makes sense

PR #4573, opened 2025-03-26
ops: track deploy failures

latter-bolden:
We had an issue today where deploys seemed to succeed despite the remote commands not executing correctly (in this case `rsync` permission issues). 

Wanted to try and capture those failures and thread them through to the GitHub action. Would an approach like this make sense? (my bash skills are tragically lackluster)

To test, I was thinking we could try against a low-risk canary deploy.


.github/helpers/deploy.sh
```diff
@@ -54,4 +60,12 @@ gcloud compute \
...
-echo "OTA performed for $desk on $ship"
+DEPLOY_EXIT_CODE=$?
+
+if [ $DEPLOY_EXIT_CODE -ne 0 ]; then
+  echo "Remote commands failed with exit code $DEPLOY_EXIT_CODE"
+  exit $DEPLOY_EXIT_CODE
+else
+  echo "OTA performed for $desk on $ship"
+  exit 0
+fi
```
arthyn, line 71: we might need to add the `--command` flag to the gcloud compute command which would let it return a result

arthyn: https://stackoverflow.com/questions/39098341/gcloud-compute-execute-command-remotely

latter-bolden: My hunch is it was returning a result, but without that `set -e` it would continue past errors and the final command exits 0.

paging unix/gcloud wizard @yapishu

latter-bolden: yeah, will fix conflicts and give it a spin next week

latter-bolden:
Closing until we get around to it

PR #4576, opened 2025-03-27
Web: optimize startup time

latter-bolden:
It seems the main driver of laggy startup time is waiting for all of `syncStart` to complete. Limiting the UI blocking to high priority sync seemed to alleviate a decent portion of the wait.

The one low priority request we have right now that's required to display the root ChatList view is `syncContacts`. Bumping it up to high priority still nets most of the savings without initially flashing empty contact info in the list.

Going a step further, it also occurred to me that `syncContacts` results are more cacheable than `syncInit` and `syncLatestPosts` — generally you won't have new contacts or contacts with updated avatars/nicknames between webapp opens. I added a basic mechanism to store contacts outside the DB on sync and try loading them from cache during `syncStart`. On web, if we get a cache hit we'll defer syncing to low priority and paint the UI before refreshing them.

Kind of on the fence as to whether it's worthwhile to add that complexity. Let me know what y'all think, I'm also fine simply bumping it up to high priority unconditionally.

### DIFF
--- a/apps/tlon-web-new/src/app.tsx
+++ b/apps/tlon-web-new/src/app.tsx
@@ -352,7 +352,6 @@ function ConnectedWebApp() {
         hasSyncedRef.current = true;
       }
 
-      console.log(`bl: session`, session);
       if (!session?.startTime) {
         return;
       }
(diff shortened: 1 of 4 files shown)


arthyn (approved):
lgtm

patosullivan (approved):
This makes sense to me. Tested locally and it works well, startup is super fast now.

PR #4577, opened 2025-03-27
native: revert showing unconfirmed posts in channel (#4280)

davidisaaclee:
#4280 added a feature to "hand-off" posts from a push notification to the host app to be inserted as an "unconfirmed" post. We then aimed to show these unconfirmed posts in the channel while loading the rest of the channel.

This approach's benefits can be largely achieved through the in-progress background sync work[0], and has proven to have some bad trade-offs, including:
- A bad UX where entering a channel caused unnecessary scroll movement, which looks like the following sequence:
  1. Unconfirmed post(s) are loaded and shown to user immediately: these are likely the most recent posts (at the bottom of the scroll). **User sees 1-3 unconfirmed posts at the bottom of the screen, white space above.**
  2. Cached posts are loaded from DB after ~500ms (just due to existing slow query issue) and added above the unconfirmed posts. **User sees posts appear in the scroll space above the previously-rendered unconfirmed posts.**
  3. If the user had an unread marker, it is likely in the set of cached, confirmed messages; the app scrolls to that unread marker. **User sees channel scroll away from where they were looking, stopping at unread marker.**
  There shouldn't be any scroll movement when entering a channel – user should start at their unread (or a loading indicator)
- Complicated query logic (see diff in `packages/shared/src/db/queries.ts`)

This PR does not revert fetching the payload from the native side – that is currently an unnecessary fetch, but includes work we want included so we can move to the activity-event API in that code.

<details>

<summary>🙋 <i>Can we just avoid scrolling to the unconfirmed message?</i></summary>

This would mean hiding the unconfirmeds until we've loaded the unread marker; otherwise, we wouldn't know where to scroll to. At that point, the unconfirmeds are not helping at all.

In other words, having any kind of gap (i.e. unsynced posts between two cached posts) in the timeline is not helpful, since we don't have a way to display the gap (so we have to just leave off one of the cached "sides" when rendering). ("Heads" are still helpful in other contexts, like showing the latest post preview in the channel list.)

</details>

[0] The behavior is largely already covered by periodic background sync fetching the "heads" of each channel; and will be even more covered by TLON-3802

<details>

<summary>Reverts the following commits (all from #4280):</summary>

lost the hashes, sorry

```
  Add simple handoff when dmPost is provided on notification payload
  [wip] Hand off data from notif payload to store
  When filtering posts to current window, include unconfirmed posts
  Add queries + hook for working with unconfirmed posts
  Handoff notif data from presented notifs and newly from tapped notif
  Update getChannelPosts to be aware of unconfirmed posts
  Fetch and insert unconfirmed posts into useChannelPosts results
  HACK: Add manual conflict resolution when inserting posts
  Fix order of inserted unconfirmeds
  Remove unnecessary sort
  Fetch latest post in getUnconfirmedPosts query
  Fix dependencies for useChannelUnconfirmedPosts
  Fetch unconfirmed posts once per channel open
  Preserve cursor ID when using it as a window base
  Remove stray log
  Combine dependent useMemos
  Add comment explaining why unconfirmed posts are managed separately
  Add Drizzle issue reference
  Add option to disable loading unconfirmed posts
```
</details>


packages/shared/src/db/queries.ts
```diff
@@ -118,17 +118,6 @@ const GROUP_META_COLUMNS = {
   coverImageColor: true,
 };
 
-const POST_RELATIONS_DEFAULT = {
-  author: true,
-  reactions: {
-    with: {
-      contact: true,
-    },
```
davidisaaclee, line 126: i made sure to mirror this `reactions.with.contact` change into `getChannelPosts` below!

jamesacklin:
Can't comment on the code but this is a much better experience in the simulator. You have my approval

patosullivan (approved):
Changes look good. Much better UX on both mobile and web.

PR #4579, opened 2025-03-28
app: low-priority cosmetic omnibus

jamesacklin:
Fixes TLON-3633 by echoing the group name in the channel header in single-channel groups; eliminates the confusing "General" when you create a new group.

Fixes TLON-3788 by allowing you to also filter ChatList by patp.

~Fixes TLON-3871 by doing some additional text processing before we display it in the post preview in ListItem.~

Fixes TLON-3869 by rephrasing a few activity summary items.


jamesacklin:
Taking back to draft status to fix a build error.

patosullivan (changes requested):
I pulled this down to test locally on web and the app fails to initialize with:
```
index.tsx:7 Uncaught ReferenceError: Cannot access 'ListItem' before initialization
    at index.tsx:7:28
```

which points at `packages/app/ui/components/ActionList/index.tsx`

jamesacklin:
Ack, probably some circular dependency silliness with convertContent(). Will fix!

jamesacklin:
@patosullivan I pulled the inline pre-processing out, that's gonna be a circular dependency wormhole for sometime when we're all truly bored :)

patosullivan (approved):
Looks good!

PR #4583, opened 2025-03-28
verifier, lanyard: compat with 410 jael vile scry

Fang-:
Copies over the 410 $feed:jael type into our sur file, because we scry for it in (due to operational weakness) environments in either 411 _or_ 410. This types hould be compatible with both.

We pluck suss out the $seed from the result we receive and operate on it as normal.

We should do a proper 410-only patch here once the situation around that scry endpoint in 410 solidifies.


latter-bolden (approved):
🔥🚢

arthyn (approved):
lgtm

PR #4584, opened 2025-03-28
desktop: allow paste-to-post for images in galleries

patosullivan:
fixes tlon-3863

~At first, I attempted to implement this in `app/ui/components/Channel/index.tsx`, but I ended up implementing this in ChannelScreen using useUploadState directly and not using the attachment context because:~

~1) It avoids race conditions between attachment state updates and upload operations (calling `attachAssets` and `waitForAttachmentUploads` from the attachment context in succession was fraught).
2) It works within component hierarchy constraints (ChannelScreen is a parent of AttachmentProvider and it feels like this functionality should be in ChannelScreen).
3) It provides granular control for auto-posting each image as a separate message immediately after upload completion~

~Open to ideas if anyone can think of a better approach, though.~

I added a new method to our attachment context that handles this case.


dnbrwstr:
Would it be possible to extend or modify the attachment context to work better here? This seems like something it really *should* be able to handle.

patosullivan:
> Would it be possible to extend or modify the attachment context to work better here? This seems like something it really _should_ be able to handle.

maybe we could add a new method to AttachmentState? `uploadAndAutoPost` or similar.

patosullivan:
@dnbrwstr updated to use the context

arthyn (approved):
lgtm

PR #4587, opened 2025-03-30
fix: modal issues

latter-bolden:
Fixes TLON-3870 by calling the correct close method. Also avoids open state thrashing by waiting to pop the sheet until reading the real storage value.

Fixes TLON-3875 by only rendering the `Modal` on web. More carefully managing the `Modal` state also worked, but since it's only present to escape the web drawer navigators, I think it's better to avoid using it here on mobile whatsoever.

### DIFF
--- a/packages/app/ui/components/ImageViewerScreenView.tsx
+++ b/packages/app/ui/components/ImageViewerScreenView.tsx
@@ -3,7 +3,7 @@ import { Icon } from '@tloncorp/ui';
 import { Image } from '@tloncorp/ui';
 import * as FileSystem from 'expo-file-system';
 import * as MediaLibrary from 'expo-media-library';
-import { ElementRef, useRef, useState } from 'react';
+import { ElementRef, PropsWithChildren, useRef, useState } from 'react';
 import { Alert, Dimensions, Modal, TouchableOpacity } from 'react-native';
 import {
   Directions,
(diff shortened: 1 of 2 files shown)


packages/app/ui/components/ImageViewerScreenView.tsx
```diff
@@ -196,6 +196,15 @@ export function ImageViewerScreenView(props: {
...
       </GestureDetector>
-    </Modal>
+    </ImageViewerContainer>
   );
 }
+
+function ImageViewerContainer(props: PropsWithChildren) {
+  // on web, we wrap in a mobile to escape the drawer navigators
+  if (isWeb) {
+    return <Modal animationType="none">{props.children}</Modal>;
```
jamesacklin, line 206: 🫡

jamesacklin (approved):

packages/app/ui/components/ImageViewerScreenView.tsx
```diff
@@ -196,6 +196,15 @@ export function ImageViewerScreenView(props: {
           ) : null}
         </ZStack>
       </GestureDetector>
-    </Modal>
+    </ImageViewerContainer>
   );
 }
+
+function ImageViewerContainer(props: PropsWithChildren) {
+  // on web, we wrap in a mobile to escape the drawer navigators
```
patosullivan, line 204:
```suggestion
  // on web, we wrap in a modal to escape the drawer navigators
```

patosullivan (approved):

PR #4588, opened 2025-03-31
Mobile: bundle of fixes

latter-bolden:
- Fix the issue where phone number can not be submitted multiple times during signup
- Ensure client always sends `E.164` encoded phone numbers to lanyard (cc @Fang- ) and in turn, runs pretty display processing before displaying values read from lanyard
- Explicitly sets `eas` build image for iOS to fix the warning email. 
  - Note: this should get resolved automatically before the deadline by the Expo Team, but configuring manually for now should remove the warning

### DIFF
--- a/apps/tlon-mobile/eas.json
+++ b/apps/tlon-mobile/eas.json
@@ -23,7 +23,8 @@
         "gradleCommand": ":app:bundlePreviewRelease"
       },
       "ios": {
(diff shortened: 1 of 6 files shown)


arthyn (approved):
lgtm

jamesacklin (approved):

PR #4592, opened 2025-03-31
refs: render smaller images in refs

patosullivan:
fixes tlon-3872

This adds a new, simple context: `IsInsideReferenceContext`. We can use this context to conditionally change the way we render some content in refs. For now, we just show smaller images in refs and we return null for refs-in-refs.

Removes the need for the `renderReferences?` prop in PostContentRenderer.


arthyn (approved):
lgtm

PR #4593, opened 2025-03-31
telemetry: trackedPoke & subscribeOnce timeout logging

latter-bolden:
Looking into create group failures, I noticed we were not properly timing out tracked pokes which prevent these errors from ever bubbling up to PostHog. Added a configureable timeout and default it to 20 seconds. The new log will include the SSE connection state at the time of failure as well as details about the poke and sub paths.

I also added a named `tag` param to `trackedPoke` and `subscribeOnce`. Often times poke params and sub endpoints are too generic to know what operation was intended. We can now indicate the user action that the operation corresponds to which will make interpreting telemetry easier. Updated all existing operations to include a tag.

Layered in some explicit piecemeal debugging for `groupActions.createGroup` as well.


packages/shared/src/api/urbit.ts
```diff
@@ -14,7 +14,7 @@ import {
 import { desig, preSig } from '../urbit';
 import { getLandscapeAuthCookie } from './landscapeApi';
 
-const logger = createDevLogger('urbit', false);
+const logger = createDevLogger('urbit', true);
```
arthyn:
```suggestion
const logger = createDevLogger('urbit', false);
```

arthyn (approved):
lgtm

PR #4594, opened 2025-04-01
attachments: improve UX while image is being resized and uploaded

patosullivan:
fixes tlon-3664 by eagerly showing a loading indicator and closing the attachment sheet as soon as the image picker or camera are launched. The hang that users were experiencing had to do with:

1) the sheet staying open until the launcher method returned a promise
2) the attachment preview list not showing any loading indicator (user only saw a gray box when we had the actual attachment ready, and that was just until the image rendered in).

Also, cleans up our use of the `attachAssets` method throughout the app. Rather than passing it in as a prop to AttachmentSheet, we just pull it out of the context there. Components that need to process the attached images can use the `onAttach` prop.


arthyn (approved):
lgtm

PR #4596, opened 2025-04-01
web: working telemetry

latter-bolden:
We weren't capturing app events in PostHog from new web, only a few particular events were making it through. 

Updates `posthog-js`, initializes it with our dev loggers, and uses our shared telemetry library for identification and mandatory events.

Also rips off the bandaid with old style `trackEvent` calls.

### DIFF
--- a/apps/tlon-web-new/src/app.tsx
+++ b/apps/tlon-web-new/src/app.tsx
@@ -581,12 +581,6 @@ function RoutedApp() {
     }
   }, [isStandAlone, body]);
 
-  // useEffect(() => {
-  //   if (posthog && analyticsId !== '' && logActivity) {
-  //     posthog.identify(analyticsId, ANALYTICS_DEFAULT_PROPERTIES);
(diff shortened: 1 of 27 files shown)


apps/tlon-web-new/src/app.tsx
```diff
@@ -29,7 +30,7 @@ import { Helmet } from 'react-helmet';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 
 import EyrieMenu from '@/eyrie/EyrieMenu';
-import { ANALYTICS_DEFAULT_PROPERTIES } from '@/logic/analytics';
+// import { ANALYTICS_DEFAULT_PROPERTIES } from '@/logic/analytics';
```
patosullivan, line 33: do we need to keep this?

patosullivan (approved):
Looks good, app builds and runs fine locally.

patosullivan (approved):

PR #4598, opened 2025-04-01
group creation error handling

dnbrwstr:
- fixes a race condition where group wouldn't be created in time by creating optimistically
- fixes issue that caused failed poke errors to get swallowed


packages/shared/src/http-api/Urbit.ts
```diff
@@ -713,15 +713,20 @@ export class Urbit {
...
+        },
+        onError: (err) => {
+          onError(err);
+          reject(err);
+        },
+      });
+      this.sendJSONtoChannel(message).catch(reject);
     });
-    await this.sendJSONtoChannel(message);
     return message.id;
```
latter-bolden, line 730: is this return extraneous now?

latter-bolden (approved):

packages/shared/src/http-api/Urbit.ts
dnbrwstr: yup

PR #4600, opened 2025-04-01
desktop: adjust type size

arthyn:
Fixes TLON-3342, we change the tamagui config based on the platform (couldn't find a way to change tokens based on window size), and to deal with text we add variants based on the screen size. Also had to adjust the hardcoded desktop sidebar widths. I tried to get as close to the 90% zoom as possible. Most of this is just manual tweaking to get magic numbers. Probably best to pull down and just click around. There may be spots where we have some values that are hardcoded

New sizing on the left, 90% zoom on the right
Screenshot 2025-04-01 at 4 22 40 PM

Screenshot 2025-04-01 at 4 23 35 PM


packages/app/ui/components/ListItem/ListItem.tsx
```diff
@@ -47,7 +47,7 @@ export const ListItemFrame = styled(XStack, {
   alignItems: 'stretch',
   backgroundColor: '$transparent',
   userSelect: 'none',
-  height: '$6xl',
+  // height: '$6xl',
```
patosullivan, line 50:
```suggestion
```

packages/app/ui/tamagui.config.ts
```diff
@@ -411,18 +418,135 @@ export const media = createMedia({
...
+
+// const config = {
+//   ...baseConfig,
+//   tokens: {
+//     ...baseConfig.tokens,
+//     size: measures,
+//     space: measures,
+//     radius: measures,
+//   },
+// };
```
patosullivan, line 513:
```suggestion
```

patosullivan (approved):
This looks great imo

PR #4601, opened 2025-04-02
onboarding: wayfinding

latter-bolden:
Adds the new "wayfinding" story which features:
- an initial splash sequence explaining a bit about the app
- a new initial "beginner" group that includes tooltips and all three channel types

The new group is created client side (see `scaffoldPersonalGroup`), so we're responsible for detecting when it's needed. Most of the UI wayfinding elements are within that group itself, so displaying those is more straightforward (either you have the group or you don't).

## Mobile
We scaffold the new group during the signup `bootSequence` and store that you need to complete the splash sequence as part of your authentication details. For revival cohort users, we'll do the same thing at the end of `handleLogin`.

## Desktop
When initially blocking the UI, we now check for a new `?did-signup=true` query param that hosting will set when redirecting after web signup. Since that's more fickle than our mobile methodology, it uses a fallback heuristic of "do you have 4 or less groups?". If either of those conditions are true, we'll scaffold the new group and trigger the splash if needed.

## Backend Persistance
We need to make sure wayfinding progress is saved so you don't constantly get hit with annoying messages and tooltips. Two new settings store keys are added:
`completedWayfindingSplash` — whether or not you've already finished the splash sequence
`completedWayfindingTutorial` — whether or not you've completed the main wayfinding "tutorial". Right now, we flip this after you've visited all of the channels in your new beginner group.

Fixes TLON-3940
Fixes TLON-3990
Fixes TLON-3988
Fixes TLON-3991
Fixes TLON-4002
Fixes TLON-3989


jamesacklin:
He's cooking hard.

What's the best way to trigger this sequence on my own ship?

latter-bolden:
> What's the best way to trigger this sequence on my own ship?

Lemme cut a build, just doing a genuine signup will be easiest.

packages/app/ui/components/Wayfinding/Notices.tsx
```diff
@@ -0,0 +1,237 @@
...
+          You can customize the appearance of your group before you send it to
+          your friends.
+        </NoticeText>
+      </NoticeContainer>
+    </View>
+  );
+}
+
+export function ChatInputTooltip() {
+  return (
```
arthyn, line 160: we don't have a conditional here like gallery/notebook

packages/app/ui/components/MessageInput/MessageInputBase.tsx
```diff
@@ -201,6 +206,7 @@ export const MessageInputContainer = memo(
               </View>
             ) : (
               <View marginBottom="$xs">
+                {showWayfindingTooltip && <Notices.ChatInputTooltip />}
```
arthyn, line 209: are we putting it here for positioning reasons? seems like it could maybe be placed in ChatInput where we initially call the hook.

packages/app/ui/components/Wayfinding/SplashSequence.tsx
```diff
@@ -0,0 +1,492 @@
...
+    props.onCompleted();
+  }, [props, store]);
+
+  return (
+    <View
+      flex={1}
+      // marginTop={insets.top}
+      // marginBottom={insets.bottom}
+      // paddingTop="$2xl"
+      // paddingBottom="$xl"
```
arthyn, line 52: stray

packages/app/ui/components/Wayfinding/SplashSequence.tsx
```diff
@@ -0,0 +1,492 @@
...
+} from 'tamagui';
+
+import { useActiveTheme } from '../../../provider';
+import { useStore } from '../../contexts';
+import { ListItem } from '../ListItem';
+import { Squiggle } from '../Squiggle';
+import { PrivacyThumbprint } from './visuals/PrivacyThumbprint';
+
+enum SplashPane {
+  Welcome = 'welcome',
```
arthyn, line 27:
```suggestion
  Welcome = 'Welcome',
```

packages/app/ui/components/Wayfinding/SplashSequence.tsx
```diff
@@ -0,0 +1,492 @@
...
+        <ChannelsPane
+          onActionPress={() => setCurrentPane(SplashPane.Privacy)}
+        />
+      )}
+      {currentPane === 'Privacy' && (
+        <PrivacyPane onActionPress={() => setCurrentPane(SplashPane.Invite)} />
+      )}
+      {currentPane === 'Invite' && (
+        <InvitePane onActionPress={handleSplashCompleted} />
+      )}
```
arthyn, line 70: is it difficult to use react navigation here? just curious because I know we use this pattern elsewhere too

arthyn (approved):
lgtm, nothing blocking. you had a note about how we determine who gets a personal group. are we just going to do this for everyone?

packages/shared/src/logic/wayfinding.ts
```diff
@@ -0,0 +1,110 @@
...
+  const didSignup = url.searchParams.get('did-signup');
+  return didSignup === 'true';
+}
+
+export function personalGroupHasDefaultTitle(group?: db.Group | null) {
+  if (!group) {
+    return false;
+  }
+
+  return group.title?.toLowerCase().includes('group');
```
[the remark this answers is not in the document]
latter-bolden: I think it'll suffice for now, I'd rather err on the side of false positives since it'll only ever be run against the actual personal group's title.

packages/app/hooks/useBootSequence.ts
```diff
@@ -313,6 +343,28 @@ export function useBootSequence() {
       }
     };
 
+    // if we're stuck trying to scaffold wayfinding, bail
+    if (bootPhase === NodeBootPhase.SCAFFOLDING_WAYFINDING) {
+      if (!tryingWayfindingSince.current) {
+        tryingWayfindingSince.current = bootStepCounter;
+      } else if (
+        bootStepCounter - tryingWayfindingSince.current >
```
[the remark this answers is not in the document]
latter-bolden: we chatted out of band

packages/app/ui/components/MessageInput/MessageInputBase.tsx
latter-bolden: yeah, clearer position targeting. I can try to refactor if you think it's meaningfully detrimental?

packages/app/ui/components/Wayfinding/SplashSequence.tsx
latter-bolden: yeah, good question. I generally do it if a feature/story feels "self contained" and splitting into separate screens seems "too heavy". You do however lose the transition animations. Using react navigation would complicate the desktop modal, but I don't see other significant drawbacks.

Open to refactoring if you'd prefer?

jamesacklin:
> lgtm, nothing blocking. you had a note about how we determine who gets a personal group. are we just going to do this for everyone?

New hosted ships and existing hosted ships we flag for revival in Horizon. Self-hosters are exempt (at least that's my understanding).

PR #4602, opened 2025-04-02
desktop: fix desktop image attachment regression

patosullivan:
OTT, fixes tlon-3991

Browser doesn't like the fake image trick.


arthyn (approved):
lgtm

PR #4605, opened 2025-04-02
ops: swap to new desktop

arthyn:
This swaps the current %groups deployed glob to come from the new web codebase. We can rip out the old files after this sometime. Fixes TLON-3885


patosullivan (approved):
🫡

This should cover it

jamesacklin:
🫡

PR #4607, opened 2025-04-03
channels: fix unread scroller lock

latter-bolden:
cc @arthyn 

This resolves it, but we should find a better fix if possible.


arthyn (approved):
lgtm, lets do this for now maybe we can dig deeper when we remove post windows

PR #4612, opened 2025-04-03
sw: make old one unregister itself, change path of desired SW

arthyn:
OTT hopefully resolves our SW woes


patosullivan (approved):
Let's try it

PR #4614, opened 2025-04-03
native: predictive channel posts sync

davidisaaclee:
fixes TLON-3802

### Summary
1. During `syncStart`, we now invoke `syncRelevantChannelPosts` alongside other low-priority tasks https://github.com/tloncorp/tlon-apps/blob/c1d86509837ec1c51181cff4ec383a29fadc30cb/packages/shared/src/store/sync.ts#L1384-L1386
2. In `syncRelevantChannelPosts`, we first collect an ordered list of channels that should fetch new posts[0] (max 20 channels). https://github.com/tloncorp/tlon-apps/blob/18960c3fac822444ef09e8d12f31ee29dfb399bf/packages/shared/src/store/sync.ts#L203-L206 https://github.com/tloncorp/tlon-apps/blob/18960c3fac822444ef09e8d12f31ee29dfb399bf/packages/shared/src/db/queries.ts#L442-L471
3. Then, for each channel, we follow the logic of `useChannelPosts`, fetching a relevant set of posts based on the channel's unread. We do at most 2 fetches for each channel. https://github.com/tloncorp/tlon-apps/blob/18960c3fac822444ef09e8d12f31ee29dfb399bf/packages/shared/src/store/sync.ts#L209-L220 https://github.com/tloncorp/tlon-apps/blob/18960c3fac822444ef09e8d12f31ee29dfb399bf/packages/shared/src/store/sync.ts#L1489-L1594
5. If any failures happen in a channel's sync, we log the error and move onto the next channel.

[0] Channels that fail `hasChannelCachedNewestPosts` are considered candidates for predictive sync; these are then ordered by last viewed at, then by last post at. [query](https://github.com/tloncorp/tlon-apps/blob/c1d86509837ec1c51181cff4ec383a29fadc30cb/packages/shared/src/db/queries.ts#L442-L471)

### Code duplication
This PR adds a couple of unfortunate code duplications:
- `hasChannelCachedNewestPosts`'s logic is converted into a Drizzle query [here](https://github.com/tloncorp/tlon-apps/blob/c1d86509837ec1c51181cff4ec383a29fadc30cb/packages/shared/src/db/queries.ts#L454-L463) – I don't see a way around this, since using the JS `hasChannelCachedNewestPosts` would require loading all channels into memory, which could be a lot. (We could do something wacky with reusing the Drizzle predicate in `getChannelPosts` instead of doing that in JS, but it doesn't seem worth the confusion.)
- The [code to fetch a page of posts](https://github.com/tloncorp/tlon-apps/blob/c1d86509837ec1c51181cff4ec383a29fadc30cb/packages/shared/src/store/sync.ts#L1528) is very similar in logic to `useChannelPosts`'s [`queryFn`](https://github.com/tloncorp/tlon-apps/blob/c1d86509837ec1c51181cff4ec383a29fadc30cb/packages/shared/src/store/useChannelPosts.ts#L67), mixed with [`ChannelScreen`'s invocation of it](https://github.com/tloncorp/tlon-apps/blob/c1d86509837ec1c51181cff4ec383a29fadc30cb/packages/app/features/top/ChannelScreen.tsx#L192-L206). I think we could reuse here, but there are a lot of seemingly important special cases in the `useChannelPosts` code path that I'd want to be careful about / split into a new PR at least.

### Outstanding bugs
- DMs will show up in the predictive sync channel set on every launch, even if they are fully cached (although they will get deprioritized if another channel has a more recent post)
  - It looks like DMs' `syncedAt` field is getting nulled out before the app launches – i.e. a DM's `channel` has `syncedAt` set as expected in all sync situations, but quitting and relaunching app witnesses `syncedAt` evaluating to `null`
  - **this is an existing issue in `develop`** [filed](https://linear.app/tlon/issue/TLON-3923/bug-dms-reset-syncedat-value-on-app-launch)
- Sometimes, loading a channel will show a single post, and won't show the rest of the posts. 
  - I can't find a repro for this – going to continue to try to find it, but opening for review anyways
  - checking the database, the expected posts are there – I'm guessing they're just not making it to `useChannelPosts`. (Predictive sync finished before navigating to channel.)
  - tapping out and back into channel shows posts

### How did I test?
Main test was opening the app with my production account and watching logs to check that everything loaded reasonably. Then, I went into channels and checked if posts were being redundantly loaded from API. Posts were never loaded from API, except when:
- The channel was not part of the predictive sync channel list
- The channel was already fully-read (even if not fully-synced before predictive sync) – in this case, we always hit the API (see `canFetchNewerPosts` in [this case](https://github.com/tloncorp/tlon-apps/blob/c1d86509837ec1c51181cff4ec383a29fadc30cb/packages/shared/src/store/useChannelPosts.ts#L81))

I also did a similar test, explicitly putting my channels into the following states:
- channel was fully read, then received new posts while client was closed
- unread marker is outside of any cached post window (e.g. while client A is closed, send 2 posts to my user, set unread on client B, reopen client A)
- wiped client database

The 20 channel sync takes ~1 minute to complete for my production hosted account. This is long, but given that we await the fetches as low-priority sync queue tasks, I think this shouldn't interfere with UX.


dnbrwstr (approved):
Nice, looks good to me!

jamesacklin:
Very exciting stuff. Let's merge and test it out!

PR #4617, opened 2025-04-04
web: telemetry fix

latter-bolden:
On new web, we're seeing hosted users initialized with Telemetry disabled. I think this is because previous iterations of the web build were hardcoded to unconditionally disable telemetry (PostHog opt out) and we were cueing off that setting when identifying.

This adds a one time forced telemetry reset for new web. Continues to only enable it by default for hosted users.


arthyn (approved):
lgtm

PR #4618, opened 2025-04-04
web: hide Profile back button when in contacts tab

latter-bolden:
OTT, the action is incoherent there

To avoid passing another navigation prop,  I reshuffled the ScreenHeader which cleans param passing up a bit.

Fixes TLON-3928


patosullivan (approved):
looks good!

PR #4624, opened 2025-04-05
msginput: fix dark mode toolbar

arthyn:
Fixes TLON-3933, however I'm noticing that dynamically changing the system theme doesn't update to the correct one. This should hold us over though.


latter-bolden:
Looks like swapping it at the top level carries through some other unintended styles

Screenshot 2025-04-06 at 7 45 21 PM

PR #4625, opened 2025-04-07
web: notebook bugs

latter-bolden:
(hopefully) OTT for 

Fixes TLON-3934
Fixes TLON-3933


patosullivan (approved):
Looks good! That figures that it was the animations causing the issue.

PR #4628, opened 2025-04-07
activity: fix navigating to threads

latter-bolden:
It looks like the issue here is actually just that we don't have the activity posts already loaded into the DB. If the post does already exist, it works as you'd expect. 

What tipped me off was that it wasn't working on mobile either. My guess is this broke after we refactored screen params to only accept primitive types. Previously activity would scaffold a full post and pass that along.

The post-navigation URL having unserialized objects in the query param seems to have been a red herring. We probably do want to avoid those somehow? But it seems that's not actually what's breaking things here.

Fixes TLON-3927


arthyn (approved):
makes sense lgtm

PR #4629, opened 2025-04-07
AttachmentButton.native, others: use Icon component

jamesacklin:
Switches the bare "Add" SVG to use the Icon UI component; makes it reappear on mobile.

Fixes TLON-3938

Also makes the same change in ChatOptionsSheet and ChannelFooter to avoid regressions there.


arthyn (approved):
lgtm

PR #4630, opened 2025-04-07
desktop (mostly): react perf optimization

patosullivan:
fixes tlon-3892 (makes it much better anyway). 

We were seeing a lag when quickly switching between channels because of some render perf issues that were triggered by channel switching. HomeSidebar, MessagesSidebar, TopLevelDrawer, ChatList, GroupChannelsScreenView, ChannelOptionsSheetLoader, and some other components were re-rendering several times on every channel switch.

I went through and tried to find every place in the app that was unnecessarily re-rendering when we switched channels. If you run the profiler or the JS performance tool (most noticeable difference), you'll see that we cut the rendering time pretty dramatically by memoizing the crap out of everything.

There's still some noticeable hitching when we're waiting on multiple requests to finish when switching between several channels very quickly, but it's much better.

### DIFF
--- a/apps/tlon-web-new/src/app.tsx
+++ b/apps/tlon-web-new/src/app.tsx
@@ -3,6 +3,7 @@ import {
   DarkTheme,
   DefaultTheme,
   NavigationContainer,
+  Route,
 } from '@react-navigation/native';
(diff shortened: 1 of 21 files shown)


dnbrwstr (approved):
Looks good to me!

PR #4638, opened 2025-04-08
embeds: support all youtube videos/urls, eliminate dependency on noembed, some other cleanup

patosullivan:
OTT, fixes tlon-3821

try using a `youtu.be` url, a `https://youtube.com/shorts/` url, or a `/embed` url.

Moved `trustedProviders` into the embed module in shared, made the `is` checks for each provider use it, and added an `isYouTube` check so we don't have to rely on noembed at all anymore.


jamesacklin:
Cool, makes perfect sense to me.

Is there a way to include the URL below the embedded video just in case the uploader prohibits playing from embeds or something else doesn't work? It's nice to be able to click through even if the in-app embed doesn't work for whatever reason.

arthyn (approved):
lgtm

PR #4644, opened 2025-04-09
expose: update for latest channels content types

Fang-:
Minimal viable change, to make sure `/app/expose` doesn't crash and burn. It was failing to compile, which would've prevented the upgrade on ships that had it enabled.

Most significantly, `kind-data` was replaced by a `kind=path`, and `author` might have a "bot profile" included. The former we handle by reading the path instead of the tag, the latter by overlaying the bot profile on top of whatever the underlying ship's contact profile is.

### DIFF
--- a/desk/app/expose.hoon
+++ b/desk/app/expose.hoon
@@ -236,7 +236,7 @@
     ::
         %fact
       ?.  =(%channel-response-1 p.cage.sign)  [~ this]
-      =+  !<(r-channels:d q.cage.sign)
+      =+  !<(r-channels:v7:old:d q.cage.sign)
       ::REVIEW  should this handle %posts also?
       ?+  -.r-channel  [~ this]
           %post
--- a/desk/app/expose.hoon
+++ b/desk/app/expose.hoon
@@ -235,7 +235,7 @@
       [~ this]
     ::
         %fact
-      ?.  =(%channel-response-1 p.cage.sign)  [~ this]
+      ?.  =(%channel-response-2 p.cage.sign)  [~ this]
       =+  !<(r-channels:v7:old:d q.cage.sign)
       ::REVIEW  should this handle %posts also?
       ?+  -.r-channel  [~ this]
--- a/desk/app/expose.hoon
+++ b/desk/app/expose.hoon
@@ -224,6 +224,12 @@
   |=  [=wire =sign:agent:gall]
   ^-  (quip card _this)
   ?+  wire  ~&([dap.bowl strange-wire=wire] [~ this])
+      [%profile %widget *]
+    ?>  ?=(%poke-ack -.sign)
+    ~?  ?=(^ p.sign)
+      [dap.bowl %nacked-by-profile]
+    [~ this]
+  ::
       [%channels ~]
     ?-  -.sign
       %poke-ack  !!
--- a/desk/app/expose/page.hoon
+++ b/desk/app/expose/page.hoon
@@ -11,8 +11,7 @@
   |=  [=bowl:gall =nest:g:c msg=post:d]
   ^-  (unit manx)
   =/  author=@p
-    ?@  author.msg  author.msg
-    ship.author.msg
+    (get-author-ship:u author.msg)
   =/  aco=(unit contact-0:co)
     (get-author-contact:r bowl author.msg)
   ::
--- a/desk/app/expose/page.hoon
+++ b/desk/app/expose/page.hoon
@@ -10,12 +10,15 @@
 ++  render
   |=  [=bowl:gall =nest:g:c msg=post:d]
   ^-  (unit manx)
+  =/  author=@p
+    ?@  author.msg  author.msg
+    ship.author.msg
   =/  aco=(unit contact-0:co)
-    (get-contact:co bowl author.msg)
+    (get-author-contact:r bowl author.msg)
   ::
   ::TODO  if we render replies then we can "unroll" whole chat threads too (:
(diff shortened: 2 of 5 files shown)


desk/app/expose.hoon
```diff
@@ -236,7 +236,7 @@
     ::
         %fact
       ?.  =(%channel-response-1 p.cage.sign)  [~ this]
-      =+  !<(r-channels:d q.cage.sign)
+      =+  !<(r-channels:v7:old:d q.cage.sign)
```
Fang-, line 239: This is questionable, because this type doesn't match that of the `%channel-response-1` mark. Is that an oversight?

Fang-: And maybe this isn't even right, because `response-2` _has_ the `v7` type...

mikolajpp: Given that %channels doesn't actually emit `%response-1` facts, I think this is incorrect. (I can't recall the reason  for `%response-1` being unused); the marks that are in use:
```
%channel-response -> $r-channels-simple-post:v7:old:c
%channel-response-2 -> $r-channels:v7:old:c
%channel-response-3 -> $r-channels:c
```

On the path `/v1` to which %expose subscribes to %channel-response-2 facts are sent.

desk/app/expose.hoon
```diff
@@ -236,7 +236,7 @@
     ::
         %fact
       ?.  =(%channel-response-1 p.cage.sign)  [~ this]
```
mikolajpp, line 238:
```suggestion
      ?.  =(%channel-response-2 p.cage.sign)  [~ this]
```
see the comment below

desk/app/expose.hoon
mikolajpp: Btw, in connection with this, I have been thinking about making marks behave like type labels, which they in fact are, and let them carry a version number of the underlying type. So we would have
`%channel-response-7` instead of `%channel-response-2` for instance.

I don't think incrementing marks like we do currently brings much benefit.

desk/app/expose/page.hoon
```diff
@@ -10,12 +10,15 @@
 ++  render
   |=  [=bowl:gall =nest:g:c msg=post:d]
   ^-  (unit manx)
+  =/  author=@p
+    ?@  author.msg  author.msg
+    ship.author.msg
```
mikolajpp, line 15: just fyi, there is `+get-author-ship` in `/lib/channels-utils`

desk/app/expose/page.hoon
```diff
@@ -10,12 +10,15 @@
...
   =/  aco=(unit contact-0:co)
-    (get-contact:co bowl author.msg)
+    (get-author-contact:r bowl author.msg)
   ::
   ::TODO  if we render replies then we can "unroll" whole chat threads too (:
-  |^  ?-  -.kind-data.msg
-          %chat
+  |^  ?+  kind.msg  ~
+          [%chat ~]
```
mikolajpp, line 21: This is maybe getting ahead of ourselves, but the way kind is gonna be used at the client side is by extending the basic path. Those extended post types would not be rendered by %expose, which we should keep in mind.

@latter-bolden @davidisaaclee

desk/app/expose/render.hoon
```diff
@@ -1,15 +1,40 @@
...
+::  +author-contact: turn a post author into a contact profile
+::
+::    if the author is a bot, replaces nickname and avatar fields,
+::    when those are specified.
+::
+++  get-author-contact
+  |=  [=bowl:gall =author:c]
+  ^-  (unit contact-0:co)
+  =/  aco=(unit contact-0:co)
+    (get-contact:co bowl ?@(author author ship.author))
```
mikolajpp, line 16: same as above, `+get-author-ship` is available from `/lib/channels-utils`

desk/app/expose/render.hoon
```diff
@@ -1,15 +1,40 @@
...
+  ==
+::
 ++  author
   =/  link=(unit @t)  ~
-  |=  [=bowl:gall author=ship]
+  |=  [=bowl:gall post-author=author:c]
   ^-  manx
+  =/  author=@p
+    ?@  post-author  post-author
+    ship.post-author
```
mikolajpp, line 35: We can consider using `+get-author-ship` here.

desk/app/expose/widget.hoon
```diff
@@ -38,8 +38,8 @@
   =/  link=tape
     (spud (print:c ref))
   =,  post.u.pon
-  ?-  -.kind-data.post.u.pon
-      %chat
+  ?+  kind.post.u.pon  ;div.exposed.fail:"kind lost: {(spud kind.post.u.pon)}"
+      [%chat ~]
```
mikolajpp, line 42: Same as above, custom content extending basic path types would not render here.

mikolajpp (changes requested):
I suggested a fix for the wrong mark type. 

Otherwise, there is a minor suggestion about using `+get-author-ship`, and a 'keep in mind' comment about custom content with extended kind paths that is not going render in expose.

Beyond these, changes look good.

desk/app/expose/page.hoon
Fang-: Do any of you think it'd make sense to render the extended post types as their normal post types for now, given that the alternative is not rendering them at all?

(Not that it matters much, it's "just" expose, after all.)

latter-bolden: It's fine for expose to drop client customizations for now. We can address later when we're "productionizing" public profiles.

PR #4645, opened 2025-04-09
db: make cache ids unique, remove unsupported extra conflict resolver

arthyn:
What present as the web app breaking and the mobile client not showing contact data, was actually an issue with inserting latest posts. We had multiple `onConflictDoUpdate` clauses which is unsupported in the current drizzle version we're using. On top of this we had a restrictive constraint for `cacheId` which wasn't accurate. Additionally we had a unique constraint on `sentAt` time which doesn't map with the backend. Finally we were bailing if we ended up trying to insert the exact same post window, so now we just no-op since it's already there (the key is the row).

In other words, most of this is just resolving discrepancies between client and backend.
- Post IDs are only unique across a single channel (not addressed in this PR, but looming issue, although it should almost never happen)
- Sent At is not unique globally, you could very easily have two posts with the same sent time
- Posts are not unique simply by their author and sent time, only by channel + author + sent

Does not address the fact that if sync bails on web, we just spin forever. Will follow up with another PR.


latter-bolden (approved):
Makes sense!

PR #4650, opened 2025-04-10
remove old web, replace it with web-new

patosullivan:
removed `tlon-web`, `git mv`'d `tlon-web-new` to `tlon-web`, added some missing dependencies to (and removed some unnecessary dependencies from) the new web app. Updated all references from `tlon-web-new` to `tlon-web`.


arthyn (approved):
lgtm, I expected the package.json in `tlon-web` to have some differences but seems unchanged which is a little surprising

apps/tlon-web/package.json
```diff
@@ -127,43 +94,30 @@
     "prosemirror-state": "~1.3.4",
     "prosemirror-transform": "~1.4.2",
     "prosemirror-view": "~1.23.13",
+    "@radix-ui/react-accordion": "^1.1.2",
+    "@radix-ui/react-popover": "^1.0.2",
```
arthyn, line 98: are we actually using these?

patosullivan: we use them in Eyrie.tsx and Subscription.tsx

PR #4651, opened 2025-04-10
channels: add %link $block element

Fang-:
For links that should be rendered as a preview card block element. Contains arbitrary metadata for use in that rendering, but we specify well-known meta keys for clarity.

Haven't tested this end-to-end yet, @latter-bolden's client integration work with `m/link-preview-fetching` is still pending.

Would be good to sneak this content AST change in along with the other protocol changes. This PR has all the changes needed to make our agents and helper libraries compile and makes sure to include the new addition in the json parser. That _should_ cover things, shouldn't it?

Closes TLON-3910.

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -401,7 +401,7 @@
   ::
   ++  story-4-to-5
     |=  [=ship old=content:two]
-    ^-  story:d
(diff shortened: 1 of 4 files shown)


arthyn (approved):
lgtm

PR #4652, opened 2025-04-10
chat: show error state when s3 uploads fail

patosullivan:
fixes tlon-3981

looks like this:

https://github.com/user-attachments/assets/ee9b3827-443e-49ef-9b9d-6e2d392182f2


arthyn (approved):
lgtm

PR #4654, opened 2025-04-10
metagrab: link metadata fetching utility

Fang-:
Implement an endpoint at `/apps/groups/~/metagrab/[@uw-encoded-url-string]` that can be used to make the urbit fetch the specified URL, parse it for metadata (from the resulting document's `<head>` section), and return that in JSON format as the response.

Motivation is that browsers often apply CORS restrictions, making it difficult for browser clients to fetch arbitrary URLs for "link preview metadata". Luckily, this is a fine task for a personal server to do.

We unfortunately must implement `/lib/de-html`, a fork of the standard library's `+de-xml`, because real-world HTML often does not pass as XHTML or any other strict parsing schemes. The library is nowhere near powerful enough to parse full HTML documents, but has been parsing just the `<head>` sections of those documents with good results.

We get data from `<meta>` tags primarily, but also some `<link>` tags, and always look for `<title>` as a fallback. Results are stored in the order in which they appear, with the namespacing and "meta-meta attributes" information rolled in.

If the URL points to a file instead of a document, or if the URL can't be successfully retrieved, we report as such in the response.

Closes TLON-3774. See also the distant-cousin PR #4651.


desk/lib/metagrab.hoon
```diff
@@ -0,0 +1,271 @@
...
+  $:  ns=@t     ::  property namespace, if any
+      key=@t    ::  property key
+      val=veal  ::  property value
+  ==
+::
++$  veal
+  $@  @t                 ::  flat property, or:
+  $:  top=@t             ::  property value
+      met=(map @t veal)  ::  w/ metadata
+  ==
```
Fang-, line 21: These type names are _garbage_ and I desperately need to replace them with something better.

desk/app/metagrab.hoon
```diff
@@ -0,0 +1,461 @@
...
+      ?~  title  ['_' 'title' url]
+      ['_title' 'title' u.title]
+    =/  hay=tape        (trip ?^(head u.head q.data.u.dat))
+    =/  hit=(unit @ud)  (find "<title>" hay)
+    =/  tit=(unit @ud)  (find "</title>" hay)
+    =?  tit  ?=(~ tit)  (find "</ title>" hay)
+    ?.  &(?=(^ hit) ?=(^ tit))  ~
+    ?:  (gte u.hit u.tit)       ~
+    =.  u.hit  (add u.hit ^~((lent "<title>")))
+    `(crip (swag [u.hit (sub u.tit u.hit)] hay))
```
arthyn, line 211: probably should pull this out into lib as a "simple" retrieval or w/e

arthyn (commented):
Makes sense to me, tried to give some naming advice lol

desk/lib/metagrab.hoon
```diff
@@ -0,0 +1,271 @@
...
+    |=  [pax=path =_lok]
+    (~(put by lok) pax buc)
+  |=  topes=(list tope)
+  %+  roll  topes
+  |=  [=tope out=(jar @t tope)]
+  =/  kay=path  ~[ns key]:tope
+  |-  ^+  out
+  ?:  (~(has by lookup) kay)
+    (~(add ja out) (~(got by lookup) kay) tope)
+  (~(add ja out) '' tope)  ::TODO  traverse?
```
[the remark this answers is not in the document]
Fang-: Yes, empty string is just the "everything without an explicit bucket" bucket.

arthyn (approved):
lgtm!

PR #4657, opened 2025-04-11
invite buttons: use copy rather than share on web

patosullivan:
fixes tlon-3964


jamesacklin (approved):
This is just on Web, right? The share sheet is pretty ergonomic on mobile

PR #4659, opened 2025-04-11
chat: show author row with edited tag when a post has been edited

patosullivan:
fixes tlon-3972

This isn't really a desktop issue, it's just that nobody noticed it until we rolled out the desktop web app. We had no way of showing if a post had been edited in chat if the message didn't show the author row. This change just ensures that we always show the author row if the post has been edited (just so we can show the edited tag).

### DIFF
--- a/packages/app/ui/components/ChatMessage/ChatMessage.tsx
+++ b/packages/app/ui/components/ChatMessage/ChatMessage.tsx
@@ -175,7 +175,7 @@ const ChatMessage = ({
         backgroundColor={isHighlighted ? '$secondaryBackground' : undefined}
         key={post.id}
       >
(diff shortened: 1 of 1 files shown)


jamesacklin:
This seems like it might break the flow of sequential messages. Can we universally put the edited status where the emoji reactions typically live? Might be a nice place for failed to send as well

patosullivan:
> This seems like it might break the flow of sequential messages. Can we universally put the edited status where the emoji reactions typically live? Might be a nice place for failed to send as well

Updated so that it shows in the same spot where we show delivery indication.

image

arthyn (approved):
lgtm

PR #4663, opened 2025-04-11
desktop: use right click for sidebar context menus

patosullivan:
fixes tlon-3966 by switching from the three dot overflow menu to using right click to trigger the context menus on ChannelListItem and GroupListItem.

Also updated the sizeRefs on GroupChannelsScreenView so that FlashList doesn't jump around on first render.

Edit: I added the options button back in, but now it just appears on hover. Right click will still show the menu, too.


arthyn (approved):
works well and lgtm. I wonder though if we shouldn't also offer the old three-dot menu as a hover? just because right-click is not really discoverable and a lot of users are trained to not right click on web.

patosullivan:
> works well and lgtm. I wonder though if we shouldn't also offer the old three-dot menu as a hover? just because right-click is not really discoverable and a lot of users are trained to not right click on web.

good question. @jamesacklin, thoughts?

jamesacklin:
How impossible is it to target the hover event? I was under the impression we were bound to JS events rather than CSS selectors which would make hiding and showing a lot simpler.

patosullivan:
> How impossible is it to target the hover event? I was under the impression we were bound to JS events rather than CSS selectors which would make hiding and showing a lot simpler.

We show on hover in the chat messages now (and used to do this on the ChatListItems too, but somewhere along the line I think we decided this was "too busy"), we could just do the same thing here.

patosullivan:
> Cool, happy to be overridden here in favor of simplicity

could do either/or/both, up 2 u

jamesacklin:
Hover it is

patosullivan:
Ok, I added the options button back in, but now it just appears on hover. Right click will still show the menu, too.

jamesacklin (approved):

PR #4665, opened 2025-04-11
notifications: render activity-event notifications using shared JS code

davidisaaclee:
Removes use of hark-yarn for notifications, using activity-event endpoints instead.

### Shared JS bundle
We want to use `activity-event` to fetch notification content going forward; however, `activity-event` does not have a server-built preview of the notification content (e.g. text we can put into a client alert) – for example, it has a full `ub.Story` for a "received new message" activity.

We need to render a preview of the event in native code, since we don't have access to a running React Native JS app when handling remote notifications (i.e. the Firebase messaging service in Android, and the notification service extension in iOS).[^0]

[^0]: See https://linear.app/tlon/issue/TLON-3834/investigate-moving-native-app-extension-code-into-rn-js-codebase for details on the difficulties of using the RN JS app in these contexts.

To avoid maintaining separate implementations of activity event parsing in Kotlin, Swift, and TypeScript, I added a TypeScript package at `packages/scripts/` which can be bundled into a single `.js` file, which can then be executed in a JS context in both Android service and iOS notification extension.[^1] This TS package can import from our other packages as needed[^2], making implementation much easier.

[^1]: Android uses a headless `WebView` to execute JS (a hack, but one that people seem to use); and iOS uses `JSContext` / `JavaScriptCore`.
[^2]: Currently, importing from `shared` causes lots of dependencies to get pulled in (including React, React Native), creating a ~0.5mb bundle. I think this fixable by shuffling exports without breaking older code.

Both Android and iOS project have new automated build steps that run `pnpm run --filter scripts build` and pull the resulting JS bundle into their respective app bundles. (I bet these build steps don't follow best practices, but the approaches have been working for me so far...)

### `PreviewContentNode`

```ts
const x = concat(lit("You received a message from "), userNickname("~ravseg-nosduc"))

console.log(await render(x));
// "You received a message from David"
```

This is a tiny data object that the shared JS bundle passes back to the native side – used when we want to defer some work to be done on native (e.g. any API call, since only native has auth / cache). Can be rendered to a string.
https://github.com/tloncorp/tlon-apps/blob/68839d6a893dcf82578019acbb414b9de41e3597/packages/scripts/src/index.ts#L11-L20

Rendering to a string on Android / iOS:
https://github.com/tloncorp/tlon-apps/blob/68839d6a893dcf82578019acbb414b9de41e3597/apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/ActivityEvent.kt#L154-L161
https://github.com/tloncorp/tlon-apps/blob/68839d6a893dcf82578019acbb414b9de41e3597/apps/tlon-mobile/ios/Notifications/NotificationPreviewPayload.swift#L81-L94

Decoding these on Android is done manually since we didn't already have a JSON serialization library – we could add gson or similar to cut out some code.

Here's where all of the notifications get constructed, using `PreviewContentNode`s for most of their fields: https://github.com/tloncorp/tlon-apps/blob/68839d6a893dcf82578019acbb414b9de41e3597/packages/scripts/src/index.ts#L70-L247

### How did I test?
I checked the following cases:
- received new post in DM, channel - tapping through brought me to the correct place
- received new reply to a subscribed channel thread - tapping through brought me to thread[^3]
- got invited to existing channel

I did not get pushes for the following:
- added my user to channel during channel creation
- kicked my user

These aren't all of the activity-event cases, but they cover all of the nodes of `PreviewContentNode`, so theoretically only the branches of `renderActivityEventPreview` are missing coverage (everything else "below" should be covered).

[^3]: This didn't work when the parent post was not yet loaded, only showing a black screen. I'm assuming this is an existing issue.


apps/tlon-mobile/ios/Notifications/NotificationPreviewPayload.swift
```diff
@@ -0,0 +1,96 @@
+import Foundation
+
+struct NotificationPreviewPayload: Decodable {
+    indirect enum ContentNode: Decodable {
```
davidisaaclee, line 4: This `Decodable` adoption is verbose because of the TS union type :(

packages/scripts/src/index.ts
```diff
@@ -0,0 +1,247 @@
...
+            groupTitle(ev['group-ask'].group),
+          ]),
+        },
+      };
+
+    case is(ev, 'group-join'):
+      return {
+        notification: {
+          body: concat([
+            userNickname(ev['group-join'].ship), // TODO: is this correct?
```
davidisaaclee, line 186: I bet I got some of these messages wrong – tried to reference the Hoon to determine the intent of the activity events but too hard.

It also looks like we don't actually receive pushes for some of these (e.g. `group-kick`).

packages/shared/src/urbit/activity.ts
```diff
@@ -676,60 +676,66 @@ export function getTop(bundle: ActivityBundle): ActivityEvent {
 
 export function getSource(bundle: ActivityBundle): Source {
   const top = getTop(bundle);
+  return getSourceForEvent(top);
+}
 
-  if ('post' in top) {
-    return { channel: { nest: top.post.channel, group: top.post.group } };
+export function getSourceForEvent(event: ActivityEvent): Source {
```
davidisaaclee, line 682: this big diff is just pulling `getSourceForEvent` out as an alternative to the bundle-based `getSource`

packages/scripts/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "scripts",
```
davidisaaclee, line 2: a bad name, but didn't know what else to call it – thinking that this package contains application scripts that will get loaded dynamically + do not have the react app context 🤷

latter-bolden (approved):
Code looking good, the JS execution is an elegant solution for this.

I tested on iOS/Android, did not see any regressions. As you called out, certain notification categories aren't making it through, but that's unrelated to these changes.

packages/scripts/package.json
latter-bolden: naming is hard, seems fine for now

PR #4666, opened 2025-04-14
gallery: handle ref attachments

patosullivan:
fixes tlon-3962

Allows users to posts refs in gallery text posts (also shows the inline preview for the ref in the big input). Doesn't allow posting refs in notebooks.


arthyn (approved):
lgtm

PR #4668, opened 2025-04-14
desktop: fix issue with chat input not allowing enter to send messages containing email addresses

patosullivan:
fixes tlon-4018

We couldn't use "enter" to send because our logic assumes that if the user has entered an "@" symbol then we're in mention mode and we're potentially selecting a user to mention. We need to avoid hijacking the enter key functionality here if we know there are no valid mentions (and therefore the popup is not being shown).

Also renamed the misnamed `showMentionPopup` to `isMentionModeActive`. The `MentionPopup` component (rendered by `InputMentionPopup`) only renders if we have valid mention candidates to show, if we don't it returns null. So we could be in a state where `showMentionPopup` was set to `true` but we didn't actually *see* the popup, because there weren't any actual mentionable users.

Also, we never want to show the mention popup from the old `MessageInput/index.tsx` component anymore, since it's only rendered as a child of a BigInput these days.


packages/app/ui/components/MessageInput/InputMentionPopup.tsx
```diff
@@ -10,21 +9,32 @@ import MentionPopup, { MentionPopupRef } from '../MentionPopup';
...
     mentionText?: string;
     groupMembers: db.ChatMember[];
     onSelectMention: (contact: db.Contact) => void;
+    setHasMentionCandidates?: (has: boolean) => void;
   }>,
   ref: MentionPopupRef
 ) {
+  useEffect(() => {
+    if (mentionText) {
+      const filteredMembers = groupMembers.filter((member) =>
```
arthyn, line 29: don't we do this logic somewhere already with fuzzy search?

patosullivan: We do it in MentionPopup itself, but we don't use fuzzy, just the .filter() method with some regexes. I'll drill the `setHasMentionCandidates` prop into there to avoid the duplicate calculation.

patosullivan: fixed in latest commit.

arthyn (approved):
lgtm

PR #4671, opened 2025-04-15
notebook: fix some formatting issues

patosullivan:
fixes tlon-3985

We no longer need to override the `enter` or `shift-enter` keys in our tentap editor because `enter` will never send a message in notebooks or galleries (it would in chat, when we used it there), so we can remove the shortcuts bridge. The shortcuts bridge was causing the issue where the user was unable to add a new line within a code block (it would just split the editor, taking the user out of the code block and adding a new paragraph node).

Also, our draft logic was overly complex, which led to issues with restoring drafts. It's already stored in prosemirror AST, we don't need to convert it to a Story and then back to the prosemirror AST.

The last thing I noticed here is that `diaryMixedToJSON` didn't account for the fact that we treat `code` nodes within an `inline` in our AST as if they are code blocks (inline code is always `inline-code` after some changes we made last year), so when editing an existing post with code blocks those code blocks would get turned into inline code. By adding a special case for handling `code` inlines and returning them as `codeBlocks` in the prosemirror AST, we can ensure that the post is formatted correctly in the editor when editing an existing post.

### DIFF
--- a/packages/app/ui/components/MessageInput/index.tsx
+++ b/packages/app/ui/components/MessageInput/index.tsx
@@ -265,24 +265,6 @@ export const MessageInput = forwardRef<MessageInputHandle, MessageInputProps>(
                 'Not editing and we have draft content',
                 draft
               );
-              // const inlines = tiptap.JSONToInlines(draft);
-              // const newInlines = inlines
-              //   .map((inline) => {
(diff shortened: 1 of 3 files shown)


arthyn (approved):
lgtm

PR #4672, opened 2025-04-15
add ability to cancel requests using AbortController

dnbrwstr:
Fixes TLON-4032

Adds ability to cancel requests using an `AbortController` signal. We should look for more places to apply this, and may want to abstract some usage patterns as they become clearer.


packages/ui/src/components/ParentAgnosticKeyboardAvoidingView.tsx
```diff
@@ -77,9 +77,12 @@ export function ParentAgnosticKeyboardAvoidingView({
   }, [adjustmentPaddingBottom]);
 
   const combinedStyle = React.useMemo(() => {
-    return StyleSheet.compose(contentContainerStyle, {
-      paddingBottom: adjustmentPaddingBottom,
-    });
+    return [
```
dnbrwstr, line 80: Unrelated, but clears an annoying warning.

PR #4674, opened 2025-04-16
embeds: text formatted as a link shouldn't render an embed

patosullivan:
fixes tlon-4003

if the href and the content of the link don't match, we know the user has formatted some content as a link and we should just render the content that way.


latter-bolden (approved):
Makes sense

PR #4677, opened 2025-04-16
channels: properly handle mismatches for dms and newer channels

arthyn:
Fixes TLON-3777. There were a few issues here. One was that we weren't actually showing mismatches for newer channels which use the channel configuration mechanism. Another was that for older notebooks/galleries we just weren't showing a version mismatch. Finally DMs were not being compat checked so I added that as well. I left out group DMs since I'm not sure what the proper behavior would be.

### DIFF
--- a/packages/shared/src/store/useNegotiation.ts
+++ b/packages/shared/src/store/useNegotiation.ts
@@ -4,7 +4,10 @@ import { useCallback, useEffect, useMemo, useRef } from 'react';
 
 import * as api from '../api';
 import { queryClient } from '../api';
+import { createDevLogger } from '../debug';
 import { MatchingEvent, MatchingResponse } from '../urbit/negotiation';
+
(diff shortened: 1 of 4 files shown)


packages/shared/src/store/useNegotiation.ts
```diff
@@ -4,8 +4,11 @@ import { useCallback, useEffect, useMemo, useRef } from 'react';
 
 import * as api from '../api';
 import { queryClient } from '../api';
+import { createDevLogger } from '../debug';
 import { MatchingEvent, MatchingResponse } from '../urbit/negotiation';
 
+const logger = createDevLogger('useNegotiation', true);
```
patosullivan, line 10:
```suggestion
const logger = createDevLogger('useNegotiation', false);
```

patosullivan (approved):
looks good

jamesacklin (approved):

PR #4679, opened 2025-04-17
embeds: fix issue with embeds losing interactability on mobile

patosullivan:
fixes tlon-4038

This is difficult to repro, but my hypothesis is that this has to do with either a race condition in onNavigationChangeHandler or the fact that we don't cleanup the webview on dismount. This PR attempts to address both of those issues and also adds a catch with a logger.crumb call to get some more observability on why/how frequently we're failing to open embed links in the wild.


arthyn (approved):
lgtm

PR #4680, opened 2025-04-17
mobile: integrate system contact book

latter-bolden:
Adds support for using connecting the system contact book on mobile. Once imported, system contacts can be sent Tlon Messenger invites individually from the Contacts tab or in bulk via the Personal Invite sheet.

Also registers any phone number signups with the verifier. Phone number discoverability can be managed in a new _Privacy Settings_  page.

System contacts are not persisted remotely and no-op on web. Managing system contacts follows general app patterns — api method to read from device, sync method to import and store, integrated into `syncStart`.

Fixes TLON-4015
Fixes TLON-4058
Fixes TLON-4061
Fixes TLON-4063


packages/app/features/settings/PrivacyScreen.tsx
```diff
@@ -0,0 +1,117 @@
...
+        maxWidth={600}
+        marginHorizontal="auto"
+        paddingHorizontal="$xl"
+      >
+        <YStack paddingHorizontal="$l" paddingTop="$2xl" gap="$xl">
+          <XStack justifyContent="space-between" alignItems="center">
+            <SizableText flexShrink={1}>Share Usage Statistics</SizableText>
+            <Switch
+              style={{ flexShrink: 0 }}
+              value={!state.telemetryDisabled}
```
patosullivan, line 80: Seeing something weird with this locally (on Android). I install the app, login as self hosted, leave "share anonymous usage" disabled, then I come to this screen and it shows as enabled.

packages/app/navigation/desktop/SettingsNavigator.tsx
```diff
@@ -71,6 +75,7 @@ function DrawerContent(props: DrawerContentComponentProps) {
       onManageAccountPressed={onManageAccountPressed}
       onExperimentalFeaturesPressed={onExperimentalFeaturesPressed}
       onThemePressed={onThemePressed}
+      onPrivacyPressed={onPrivacyPressed}
```
patosullivan, line 78: Also need to add it as a Screen to the SettingsDrawer.Navigator

packages/shared/src/store/systemContactActions.ts
```diff
@@ -0,0 +1,34 @@
+import * as db from '../db';
+import { createDevLogger } from '../debug';
+import * as domain from '../domain';
+import { AnalyticsEvent, AnalyticsSeverity } from '../domain';
+
+const logger = createDevLogger('SystemContactActions', true);
```
patosullivan, line 6:
```suggestion
const logger = createDevLogger('SystemContactActions', false);
```

patosullivan (changes requested):
Overall, looks great. Tested on Android emualtor, worked well. I was able to import contacts there from the contacts screen just fine.

I did notice two issues that we should probably fix before getting it in, but after that, we should be good to go.

packages/app/features/settings/PrivacyScreen.tsx
latter-bolden: My emulator is having trouble accessing localhost, so I haven't been able to test this on Android yet. Seems to work as expected on iOS.

That flag is set based on PostHog's opted out state, which is supposed to be updated when you navigate past the _SetTelemetry_ screen.

@patosullivan are you able to connect to local ships from your emulator? or were working with a remote self hosted one?

patosullivan: this was an env var issue

patosullivan (approved):
good to go

PR #4682, opened 2025-04-17
Mark channel threads read in "Mark read" action

dnbrwstr:
Previously, marking a channel as read from the `ChatOptionsSheet` would not that channel's threads as read. Now it does.

Fixes TLON-3839


latter-bolden (approved):
Nice change! Looks good

PR #4686, opened 2025-04-18
fix: verify message delivery status after channel reaps

patosullivan:
fixes tlon-3827

As far as I can tell (by looking through the code path and our posthog data), the issue here is that we send the poke for a message and then the channel happens to get reaped immediately, so the request gets `Aborted` and we don't know whether or not the ship actually knows about the message.

This PR adds a new delivery status, `needs_verification`. If we detect that the channel connection has been lost when we're sending a post we'll mark the post's delivery status as `needs_verification`. It also adds a new `verifyPostDelivery` function in postActions that just makes a request for an individual post and updates the delivery status accordingly. We use this function in sync's `handleChannelStatusChange` function, where we grab every `needs_verification` post in the db and attempt to verify it when the status switches back to `active`.


packages/shared/src/store/postActions.ts
```diff
@@ -414,6 +440,54 @@ export async function addPostReaction(
...
+  if (post.deliveryStatus !== 'needs_verification') {
+    logger.crumb('post does not need verification', {
+      status: post.deliveryStatus,
+    });
+    return false;
+  }
+
+  try {
+    // Attempt to fetch the post from the server
+    await api.getPostWithReplies({
```
latter-bolden, line 462: will the `post.id` we have cached in the DB match what we need for this scry? 

If it got marked `needs_verification` won't that mean we haven't received the backend update with it's true canonical ID?

patosullivan: Good point, this won't work as-is. We'd need a way to query for the post based on author and sent time.

latter-bolden: nice, change looks good!

latter-bolden (approved):

PR #4689, opened 2025-04-21
fix "untitled group" issue

patosullivan:
fixes tlon-4022

There were a few issues causing this.

1) We never know the group members when we're invited to a group, and if a title wasn't set we would just show the user "Untitled Group".
2) We weren't getting a full fleet list from the groups-ui init api response because we were purposefully dropping the fleet (with `drop-fleet` in the groups agent) and replacing it with just our own ship to speed the request up. This meant that when the app was initialized (which happens on every refresh on web), every group without a title would just appear as "Untitled Group".
3) After entering the group and triggering `syncGroup`, we grab the specific data for that group from the api and insert it into the db. This *should* have updated the group member list permanently in the db, but it wasn't because of an `onConflictDoNothing()` call. 

Fixes:
For 1): I changed it so we now show "New group by {host}" when the group doesn't have a title.
For 2): I updated the `drop-fleet` arm in `groups.hoon` to actually return at least 3 members of the fleet if we can.
For 3): I replaced that `onConflictDoNothing()` with ~a hard overwrite of the group members in insertGroups~ `onConflictDoUpdate()`.

Also, for good measure, I updated useResolvedChats to make sure that it updates when a group/chat title changes.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -837,8 +837,8 @@
   =.  fleet.group
     =/  our-vessel=vessel:fleet:g  (~(gut by fleet.group) our.bowl *vessel:fleet:g)
     =/  fleet-size=@ud  ~(wyt by fleet.group)
-    ?:  (lte fleet-size 3)
-      fleet.group  :: keep all members if 3 or fewer
+    ?:  (lte fleet-size 15)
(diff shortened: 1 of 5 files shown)


Fang- (commented):
Some comments about hoon style, for your edutainment. Take them or leave them! (:

PR #4690, opened 2025-04-21
ContactsScreen, ProfileNavigator: immediately add suggested contact

jamesacklin:
Fixes TLON-3935, a common complaint with suggested contacts. This changes the click/tap handlers for suggested contact rows to immediately add the suggested contact to the user's contacts, rather than navigating them to their profile.

### DIFF
--- a/packages/app/features/top/ContactsScreen.tsx
+++ b/packages/app/features/top/ContactsScreen.tsx
@@ -34,14 +34,14 @@ export default function ContactsScreen(props: Props) {
 
   const onContactPress = useCallback(
     (contact: db.Contact) => {
-      if (contact.isContactSuggestion) {
-        store.addContact(contact.id);
-      } else {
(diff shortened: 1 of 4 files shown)


patosullivan:
The code looks fine, but might this also become an annoyance? What if someone pops up in your suggested contacts and you don't recognize them, so you want to view their profile? How would you do it?

Or what if you tap one on accident?

I feel like the real issue is that users expect "Add" to be a separate button on top of the contact list item.

jamesacklin:
> I feel like the real issue is that users expect "Add" to be a separate button on top of the contact list item.

@patosullivan Good point. I had to wrap `Badge` in `Pressable` if it has an onPress, then move the conditionals around in ContactsScreen and Profile Navigator. Take a look!

patosullivan (approved):
Looks good!

PR #4691, opened 2025-04-21
chat input: fix issue with drafts not always clearing on send

patosullivan:
fixes tlon-4055

The root cause was a race condition in `storageItem.ts`. The getValue function could read from storage before a pending setValue or resetValue operation (triggered by clearDraft) had completed writing, due to not awaiting the internal updateLock.

This change modifies getValue to await updateLock before reading, ensuring that reads only occur after any pending writes for the same key are finished. This should eliminate the race condition and ensure that drafts are reliably cleared.


arthyn (approved):
lgtm

jamesacklin (approved):

PR #4692, opened 2025-04-21
nav: specify screen to ensure proper nav

arthyn:
Fixes TLON-3982, there's some weird behavior in react navigation where any navigate call that doesn't specify `initial: false` will make itself the "initial" route for a particular stack. This caused problems because if you navigated from activity to a thread/post, the initial route for the channel stack would be "Post". So later when we tried to nav with `initial: true` we'd get "Post" instead of "ChannelRoot". I just tried to avoid this implicit behavior by just being specific about what screen we actually wanted.

Also added some more logging.


patosullivan (approved):
Tested locally, works well

PR #4693, opened 2025-04-21
BareChatInput: reliably autofocus on web

jamesacklin:
Fixes TLON-3959 where the chat input doesn't autofocus, especially when the window loses and then regains focus.

BareChatInput had a useEffect to trigger the autofocus based on shouldAutoFocus, which was passed down from ChatInput. However, hasAutoFocused was set to true the first time the input was focused and would therefore never actually autofocus again. When the user navigated away and came back, the hook saw that hasAutoFocused was true, then skipped the focus() call.

To fix, I removed hasAutoFocused from the useEffect dependencies because we don't ever want to _prevent_ autofocusing. This seems to reliably keep the cursor active in the chat input field between channel / group navigations, browser tab switches, virtual desktop changes, etc.

### DIFF
--- a/packages/app/ui/components/BareChatInput/index.tsx
+++ b/packages/app/ui/components/BareChatInput/index.tsx
@@ -225,7 +225,6 @@ export default function BareChatInput({
   const [sendError, setSendError] = useState(false);
   const [hasSetInitialContent, setHasSetInitialContent] = useState(false);
   const [editorIsEmpty, setEditorIsEmpty] = useState(attachments.length === 0);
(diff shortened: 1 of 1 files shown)


jamesacklin:
@patosullivan You're right. Changed this to just auto-focus when the channel context changes; Leap still works.

patosullivan:
Still prevents focusing on the leap input. When leap opens, you have to manually click into the leap input, otherwise you'll start typing into the message input.

jamesacklin:
OK, got it -- wasn't sure what you meant last time. Back to the lab...

PR #4694, opened 2025-04-21
profile: click-to-copy feedback

jamesacklin:
Fixes TLON-4049 by making the click-to-copy ID function in profiles very obvious.

I simplified the ContactName component to handle either the nickname or the patp as the "primary name" for the profile. The secondary line shows the patp normally when a nickname is present and is now responsible for showing the copied indicator. This secondary line is only shown when props.hasNickname is true or if didCopy flips to true.

Quick demo for nickname vs non:

copy
copy_nonick


patosullivan (approved):
looks good!

PR #4695, opened 2025-04-22
scroller: move AuthorRow when deleting first message in sequence

jamesacklin:
Fixes TLON-3132 where deleting the first message in a sequence wouldn't move the AuthorRow to the second one, making it appear as though the person above you sent your second message. showAuthor was mostly just checking to see if the previous message was sent by the same author and wasn't taking the live state of the previous post into account (i.e. deleting).

- Adjusts the showAuthor logic to account for deleted posts preceding the current one.
- Uses useLivePost inside BaseScrollerItem to get the current state of the previous post and looks for the isDeleted flag.
- Tracks if the live version of the previous post exists and if it's deleted. If so, forces showAuthor to true.
- If the previous message was simply edited and not deleted, we fall back to the original showAuthor logic.

image

I don't actually know if this affects performance in the FlatList at all, since I know we've done a lot of work there and things are fairly torqued-down. @dnbrwstr would love your opinion here.

### DIFF
--- a/packages/app/ui/components/Channel/Scroller.tsx
+++ b/packages/app/ui/components/Channel/Scroller.tsx
@@ -269,6 +269,7 @@ const Scroller = forwardRef(
           !previousItem ||
           previousItem?.authorId !== post.authorId ||
           previousItem?.type === 'notice' ||
+          previousItem?.isDeleted === true ||
           isFirstPostOfDay;
         const isSelected =
           anchor?.type === 'selected' && anchor.postId === post.id;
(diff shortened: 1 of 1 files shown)


patosullivan:
This works, but only if you leave the channel and come back. If a delete comes through while you're looking at the channel we'll still see the issue.

jamesacklin:
That's a bit surprising, since that was precisely what useLivePosts is supposed to do. Let me see...

patosullivan (approved):
Latest commit fixes the issues I was seeing, I think we're good to go.

PR #4700, opened 2025-04-23
expose: pinning to (in-app) profile

Fang-:
Makes exposé inject exposed references into your contacts profile, for display on the in-app profile.

That change by itself is trivial, but we also make additions to the reference resolution pipeline in channels agent to increase the speed and reliability with which we can render these "pinned" posts. There, two changes:

1. Add a scry endpoint for retrieving entries from the `.voc`, which we were already populating, but never reading from.

2. Add a new `/said` subscription endpoint (`/v3/said/[target]/etc`) which in addition to the existing path parameters, also takes a "target" ship, from which we will try to retrieve the reference contents if we don't have them locally.

The intended usage pattern here is to first try retrieving from the cache, and if we can't get it there (or unconditionally, if we want to guarantee freshness) opening _multiple_ new-style `/said` subscriptions, targeting the host _and the ship that pinned the reference_, and working with whatever result we get back first.

To that end, ships now resolve references not only if they know the requester has read permissions, but also if the message being requested was authored and pinned by the ship itself. (TODO: check for pin isn't in yet.)

A `/ted/contact-pins` thread is added to apply that usage pattern to all references in a given ship's profile, for easy client-side retrieval of "external" content on profiles.

Closes TLON-4044, also the back-end parts of TLON-4006.

Based on `release-channels-updates` to avoid clobbering other `/said` endpoint and type changes.

Draft because I haven't tested this with over-the-network requests yet, but the overall structure/approach here should be ready to review.


desk/app/channels.hoon
```diff
@@ -723,35 +754,52 @@
...
   ?+    -.sign  !!
       %watch-ack
-    %.  cor
-    ?~  p.sign  same
-    (slog leaf+"Preview failed" u.p.sign)
+    ?~  p.sign  cor
+    %-  (slog leaf+"Preview failed" u.p.sign)
+    ::  treat subscription failures as if we received a %channel-denied
+    ::
+    $(sign [%fact %channel-denied !>(~)])
```
Fang-, line 761: Note that this also changes the behavior for old-style reference resolution. But I think this is correct: previously it would just leave the original subscriber hanging without notifying them of the subscription failure in any way.

desk/app/channels.hoon
```diff
@@ -1343,17 +1343,24 @@
...
-        channel-denied+!>(~)
-      (said-2:utils nest plan posts.channel)
+      =;  share=?
+        ?.  share
+          channel-denied+!>(~)
+        (said-2:utils nest plan posts.channel)
+      ?:  (can-read:ca-perms src.bowl)  &
+      ?^  q.plan  |  ::NOTE  expose/+grab-post doesn't support replies
+      ::  we need to grab the post first before we can check whether it's
+      ::  pinned, because its kind appears in the reference path...
```
Fang-, line 1356: This suggests that:

1. The reference path format is garbage (but we already knew that).
2. Making expose agent track exposed/pinned messages by `$cite` reference was maybe a mistake. Given that we only do this for messages right now (not other kinds of references, like groups or apps) it should maybe just be storing `$plan`s or similar in state, simpler message pointers that are easier to construct/work with.

But, both of those are out of scope for here, I'd say.

desk/app/channels.hoon
```diff
@@ -723,35 +754,52 @@
...
+      [%v3 %said (scot %p src.bowl) (tail path)]
+    =.  cor
+      %^  give  %fact
+        ~[v2+path v3-path]
+      ?~  got  cage.sign
+      channel-said-1+!>(u.got)
+    ::  they all got their responses, so kick their subscriptions
+    ::
+    =.  cor  (give %kick ~[path v0+path v1+path v2+path v3-path] ~)
+    cor
```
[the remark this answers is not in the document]
arthyn: yeah we prob should kick

desk/app/expose.hoon
```diff
@@ -261,56 +328,107 @@
...
+        (refresh-pages:e bowl ~(tap in open))
       ::
-      =+  !<(=news-0:co q.cage.sign)
-      ?.  =(our.bowl who.news-0)  `this
-      :_  this
-      %+  weld
-        (refresh-widget:e bowl open)
-      (refresh-pages:e bowl ~(tap in open))
+          %peer
+        ::  someone else's contact details changed. if they have any pinned
```
arthyn, line 395: I worry about the performance here, this means anyone that pins something will have to answer N subscriptions times the number of peers. For people that have 5000+ contacts from the old days, this could immediately DDOS the ship no?

arthyn (approved):
lgtm, but concerned about the perf for people with large numbers of contacts peers

desk/app/expose.hoon
Fang-: At the very least, we should do the `TODO` from below, where we check for a cache entry locally, and don't go over the network if we already have one.

Fang-:
From discussion during standup: should also have a poke-toggleable flag for expose that en/disabled prefetching, in case it turns out that's causing problematic amounts of load.

desk/app/expose.hoon
```diff
@@ -190,7 +191,11 @@
   ^-  (quip card _this)
   ?+  mark  !!
       %noun
+    ?>  =(src our):bowl
```
arthyn, line 194: good catch

arthyn (approved):
lgtm

PR #4701, opened 2025-04-23
lure: continue refetching if lure is invalid

latter-bolden:
In our client side Lure logic, if `fetchLure` returns results that are invalid, it's designed to refetch on an interval. However, that refetching is gated on `canCheckForUpdate` being set. If the lure is marked `fetched`, this will make `canCheckForUpdate` falsy and short circuit the polling.

This just updates that conditional to continue refetching so long as the lure `status = 'ready'` conditions are not met. This should prevent the hook from ever locking up in a `stale` state.


arthyn (approved):
lgtm

PR #4702, opened 2025-04-23
Revert "chat input: fix issue with drafts not always clearing on send"

patosullivan:
This reverts commit 316be3a1700adcec1f44160ed32c9ca6fe9a933d.

was causing an issue with `setValue()` calls that passed an updater function.


latter-bolden (approved):
Thanks! This fixes signup flow

PR #4703, opened 2025-04-23
chat input: fix issue with drafts not always clearing on send (again)

patosullivan:
fixes tlon-4055

adds an optional `waitForLock` param to `getValue`, we'll pass `true` when we know we wanna wait for the updateLock to resolve before performing the operation.

### DIFF
--- a/packages/shared/src/db/storageItem.ts
+++ b/packages/shared/src/db/storageItem.ts
@@ -17,7 +17,7 @@ export type StorageItemConfig<T> = {
 };
 
 export type StorageItem<T> = {
-  getValue: () => Promise<T>;
+  getValue: (waitForLock?: boolean) => Promise<T>;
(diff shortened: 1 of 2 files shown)


packages/shared/src/db/storageItem.ts
```diff
@@ -41,7 +41,11 @@ export const createStorageItem = <T>(config: StorageItemConfig<T>) => {
   const storage = getStorageMethods(config.isSecure ?? false);
   let updateLock = Promise.resolve();
 
-  const getValue = async (): Promise<T> => {
+  const getValue = async (waitForLock = false): Promise<T> => {
+    if (waitForLock) {
+      await updateLock;
+    }
+    await updateLock;
```
latter-bolden, line 48:
```suggestion
```

PR #4704, opened 2025-04-24
batch subscription events

dnbrwstr:
This pr improves performance of chatlist + other screens by batching incoming subscription events, reducing the number of invalidations they produce. May fix TLON-4047, but needs testing.

### DIFF
--- a/packages/shared/src/store/bufferedSubscription.ts
+++ b/packages/shared/src/store/bufferedSubscription.ts
@@ -17,7 +17,11 @@ export function createHandler<T>(
 ) {
   return createBatchHandler<T>(async (updates, ctx) => {
     for (const update of updates) {
-      await handler(update, ctx);
+      try {
+        await handler(update, ctx);
+      } catch (e) {
+        logger.trackError('failed to process buffered event', e);
+      }
     }
   });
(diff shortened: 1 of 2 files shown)


packages/shared/src/store/bufferedSubscription.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+const bufferedSubscriptions: BufferedSubscription<any>[] = [];
+
+export function createHandler<T>(
+  handler: (update: T, ctx: QueryCtx) => Promise<void>
+) {
+  return createBatchHandler<T>(async (updates, ctx) => {
+    for (const update of updates) {
+      await handler(update, ctx);
```
latter-bolden, line 20: maybe worth wrapping in a try catch so one failure doesn't take down the whole batch

latter-bolden (approved):
Code looks good. Agree best we to test will be keeping an eye out for issues during release QA.

packages/shared/src/store/bufferedSubscription.ts
dnbrwstr: makes sense, amended

PR #4707, opened 2025-04-25
attachments: don't show upload error for fake image uri, no-op instead

patosullivan:
OTT, fixes tlon-4087.

You should now see the loading spinner on the attachment preview rather than the failure message.


arthyn (approved):
lgtm

PR #4709, opened 2025-04-28
fixes for image upload issues

patosullivan:
fixes tlon-4113 (I think? I can't repro what's in the description, I can repro what was mentioned in the meeting that I think was supposed to be in the description) by only clearing the placeholder attachment, not all attachments, upon adding a new attachment.

fixes tlon-4094 by disabling the send button while we're sending (!).

Also, adds a loading spinner in place of the up arrow so users know something is happening before the message appears in the scroller. Thoughts?

https://github.com/user-attachments/assets/7e1e16a7-01c5-4180-94f8-b17a92a64b2e


arthyn:
I like the spinner

latter-bolden:
Not sure how the spinner will feel, but down to try it out

latter-bolden (approved):

PR #4711, opened 2025-04-28
mobile: fix invite button in chat option sheet

latter-bolden:
In a recent refactor, we moved the rendering of the ChatOptionsSheet into the `chatOptions` provider itself when on mobile. This broke the invite button because which chat is active for the options sheet is now handled internally rather than at the ChatList level (where the previous invite sheet behavior lived). 

To solve this while matching that new pattern, I pulled the invite sheet into `chatOptions` as well. If the context is not passed an `onInvitePressed` handler, it will display its own internal invite sheet.

The chat options logic is getting quite complicated imo (state is too intermingled between the context hooks and actual components with expected behavior differing by use case and platform). We should try to pull these apart, but that would be outside the scope of this fix.

Fixes TLON-4086

### DIFF
--- a/packages/app/features/top/ChatListScreen.tsx
+++ b/packages/app/features/top/ChatListScreen.tsx
@@ -263,10 +263,6 @@ export function ChatListScreenView({
     activeTab,
   });
 
-  useEffect(() => {
-    console.log(`slected group change`, selectedGroupId);
-  }, [selectedGroupId]);
-
   return (
     <RequestsProvider
       usePostReference={store.usePostReference}
--- a/packages/app/features/top/ChatListScreen.tsx
+++ b/packages/app/features/top/ChatListScreen.tsx
@@ -24,7 +24,6 @@ import {
   ChatOptionsProvider,
   GroupPreviewAction,
   GroupPreviewSheet,
-  InviteUsersSheet,
   NavBarView,
   NavigationProvider,
   PersonalInviteSheet,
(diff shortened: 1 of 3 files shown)


latter-bolden:
Confirmed this does not regress the sidebar `chatOptions` invite action on web.

packages/app/features/top/ChatListScreen.tsx
```diff
@@ -213,15 +205,6 @@ export function ChatListScreenView({
...
 
-  const handleSectionChange = useCallback(
-    (title: string) => {
-      if (activeTab === 'home') {
-        setScreenTitle(title);
-      }
-    },
-    [activeTab]
-  );
-
```
arthyn, line 224: accidental?

packages/app/features/top/ChatListScreen.tsx
```diff
@@ -280,8 +263,8 @@ export function ChatListScreenView({
     activeTab,
   });
 
-  const handleInvitePress = useCallback(() => {
-    setInviteSheetGroup(selectedGroupId);
+  useEffect(() => {
+    console.log(`slected group change`, selectedGroupId);
```
arthyn, line 267: remove

arthyn (approved):

packages/app/features/top/ChatListScreen.tsx
latter-bolden: was unused dead code

PR #4712, opened 2025-04-29
wayfinding: harden personal group scaffold

latter-bolden:
Various fixes around the new group scaffolding flow.

### Always reset session and sync lock
Found a bug where session doesn't get nulled out during logout. Since it's not in the KV store, it auto reset like generic storage items. Since this is what we're cueing off of for group scaffolding, we could think the connection is set up when it's not.

We also realized certain executions of `startSync` could error out without resetting the sync lock. Not positive this was happening, but should fix for good measure.

### Always get latest group before scafolding
We're seeing failures scaffolding the new personal group on signup. They seem to be tracked poke related where the group create request will consistently timeout but upon signing in, the group does exist (in an incomplete state since we never get to creating the channels).

To try and get around this, we'll now always sync the latest copy of the group from the backend before beginning the scaffold routine. Updated our group sync function to take a `config.force` param that ensures it always refetches regardless of session state.

### Avoid getting stuck on "ReserveShip" screen for too long
We also want to avoid getting stuck on the scaffolding step for too long, so this adds a timeout at the 60 second mark at what point we'll advance you to the authenticated app regardless.


packages/shared/src/store/sync.ts
```diff
@@ -1425,145 +1434,139 @@ export const syncStart = async (alreadySubscribed?: boolean) => {
...
 
-  const ship = await db.hostedUserNodeId.getValue();
-  if (ship) {
-    await api.getShip(ship);
+    // post sync initialization work
+    await verifyUserInviteLink();
+    db.userHasCompletedFirstSync.setValue(true);
+  } finally {
+    isSyncing = false;
   }
```
arthyn, line 1563: did we mean to delete these?

arthyn (approved):
lgtm, just the one question

packages/shared/src/store/sync.ts
latter-bolden: yeah, I don't think they were actually needed

PR #4713, opened 2025-04-29
android: group notifications by conversation

davidisaaclee:
_currently pointed at #4665_

I'd recommend reviewing the first commit a8a16cbbb8cf0c0e57dd69bcfd5fb699656727ba separately – that commit just translates old Java code into Kotlin (using an automatic IDE-provided translation), and messes with the diff.

---

| Expanded | Collapsed |
| --- | --- |
| Screenshot 2025-04-29 at 11 11 30 AM | These notifications auto-collapse when new notifications come in: Screenshot 2025-04-29 at 11 11 49 AM |

Groups subsequent notifications from the same channel or thread into "conversations" in the Android notification drawer.

iOS accumulates notifications posted to the same `threadIdentifier` automatically, but Android is different: it requires the service to (1) build the entire conversation history for each notification, and (2) replace the existing notification with the updated one.

(1) means that we need to hold onto older messages so we can prepend them when receiving new message notifications. I did this with a simple in-memory store: https://github.com/tloncorp/tlon-apps/blob/f9a6147c61ed8912273be274b8e0ea26fea2293c/apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/NotificationManager.kt#L29 https://github.com/tloncorp/tlon-apps/blob/f9a6147c61ed8912273be274b8e0ea26fea2293c/apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/NotificationManager.kt#L106-L116

This is not a perfect solution: if the service is killed (or crashes), we lose this store. **Once the store is lost, a subsequent new-message notif in a channel will lose the conversation history (and overwrite any existing notification), effectively making the older notifications disappear.** I don't know how bad this is in practice – tested that this is the actual behavior by triggering a crash in the service.

<details>

<summary>Some better solutions:</summary>

1. Generate a salt for each service launch, and use it with the thread ID when replacing any previous notification. This means that during a service's lifetime, a conversation's messages will be grouped (e.g. `conversation1` is grouped into `msg1, msg2, msg3`); and after getting killed, the next service's launch will again group messages, but separately from the previous batch (so you'd have the older batch `msg1, msg2, msg3` and a separate Android conversation `msg4, msg5`).
2. Persist pending messages between service launches (either using main SQLite or separate persistence)

</details>

### DIFF
--- a/apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/ActivityEvent.kt
+++ b/apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/ActivityEvent.kt
@@ -22,6 +22,7 @@ data class ActivityEventPreviewMessage(
 data class ActivityEventPreview(
     val title: String?,
     val body: String?,
+    val groupingKey: String?,
 
     // present when event represents a user-to-user message
     val messagingMetadata: ActivityEventPreviewMessage?,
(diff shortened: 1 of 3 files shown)


dnbrwstr (approved):
This looks good to me!

PR #4715, opened 2025-04-30
sync: remove clientside reconnect heartbeat

latter-bolden:
410 is out and widely deployed. It automatically sends some initial bytes to warm the connection, so we no longer need a "heartbeat" request from the client.


patosullivan (approved):

latter-bolden:
Closing until we're ready for it. Want to avoid PR clutter

PR #4716, opened 2025-04-30
groups: implement -group-create thread

mikolajpp:
The `-group-create` thread aids in creating and initializing a group based on supplied presets. 

Closes TLON-4132

### DIFF
--- a/desk/ted/group/create.hoon
+++ b/desk/ted/group/create.hoon
@@ -1,0 +1,108 @@
+::  create and preset a group
+::
+::  run via web at
+::  https://ship/spider/group/groups/group-create-thread/group-create/group-create/group-update-4
+::
+/-  spider
+/-  g=groups, gt=groups-thread, c=channels, meta
+/+  io=strandio
+::  keep warm
+/=  in-mark  /mar/group/create-thread
+::
+=,  strand=strand:spider
+^-  thread:spider
+::
+|=  arg=vase
+=/  m  (strand ,vase)
+=*  n  (strand ,~)
+^-  form:m
+=+  !<(arg=(unit create-group:gt) arg)
+?>  ?=(^ arg)
+=*  create  u.arg
+=*  channels  channels.create
+;<  =bowl:spider  bind:m  get-bowl:io
+?>  =(p.group-id.create our.bowl)
+;<  exists=?  bind:m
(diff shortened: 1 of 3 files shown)


latter-bolden (approved):
We tested this locally and it's working really well. Good to go from a usability standpoint.

desk/ted/group/create.hoon
```diff
@@ -0,0 +1,110 @@
...
+  ?.  exists  (pure:n ~)
+  ;<  ~  bind:n
+    =/  =action:v5:g
+      group-id.create^[now.bowl [%meta meta.create]]
+    (poke:io [our.bowl %groups] group-action-4+!>(action))
+  =/  =action:v5:g
+    :+  group-id.create
+      now.bowl
+    [%cordon %shut %add-ships %pending guest-list.create]
+  (poke:io [our.bowl %groups] group-action-4+!>(action))
```
arthyn, line 58: I thought groups did this for you in the create poke, no?

desk/ted/group/create.hoon
```diff
@@ -0,0 +1,110 @@
...
+  =+  guest-list=~(tap in guest-list.create)
+  |-
+  ?~  guest-list  (pure:n ~)
+  =/  =invite:v5:g
+    :_  i.guest-list
+    [our.bowl q.group-id.create]
+  ~&  invite+i.guest-list
+  ;<  ~  bind:n
+    (poke:io [our.bowl %groups] group-invite+!>(invite))
+  $(guest-list t.guest-list)
```
arthyn, line 106: same here

arthyn (approved):
lgtm

desk/ted/group/create.hoon
mikolajpp: The code below `?.`  runs only if the group had already existed – we don't send the create poke in that case.

desk/ted/group/create.hoon
mikolajpp: This code covers both cases – regardless if the group was created or not. We could test if the group was created, and thus the invites were sent, but I think it does not hurt to explicitly send them here, rather than rely on the group create poke sending them – especially if its semantics are ever changed.

Deferring to @Fang- for judgement.

arthyn: ahhh ok I see now sorry I missed that

PR #4717, opened 2025-04-30
mentions: allow mentioning everyone in channel or specific roles

arthyn:
Fixes TLON-4080. This introduces a new inline type `sect` which allows for mentioning either the entire channel or a specific role. I took an opportunity to lightly refactor how mention popup works and sources options.

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -278,6 +278,11 @@
   %+  welp
   /(scot %p our.bowl)/[dude]/(scot %da now.bowl)
   path
+++  get-vessel
+  |=  [=flag:g =ship]
+  =/  =path
+    /groups/(scot %p p.flag)/[q.flag]/fleet/(scot %p ship)/vessel/noun
+  .^(vessel:fleet:g %gx (scry-path %groups path))
 ++  poke
   |=  [=mark =vase]
   ^+  cor
@@ -1255,8 +1260,9 @@
     =/  key=message-key:a
       :_  time
       [(get-author-ship:ch-utils author.u.post) time]
+    =/  =vessel:fleet:g  (get-vessel group our.bowl)
     =/  mention
-      (was-mentioned:ch-utils content.u.post our.bowl)
+      (was-mentioned:ch-utils content.u.post our.bowl `vessel)
     `[time %post key nest group content.u.post mention]
   =/  replies=(list [time incoming-event:a])
     %-  zing
@@ -1280,8 +1286,9 @@
     =/  parent=message-key:a
       :_  id-post
       [(get-author-ship:ch-utils author.u.u.post) id-post]
+    =/  =vessel:fleet:g  (get-vessel group our.bowl)
     =/  mention
-      (was-mentioned:ch-utils content.u.reply our.bowl)
+      (was-mentioned:ch-utils content.u.reply our.bowl `vessel)
     [time %reply key parent nest group content.u.reply mention]
   =/  init-time
     ?:  &(=(posts ~) =(replies ~))  recency.unread
@@ -1320,7 +1327,7 @@
     |=  [=time =writ:ch]
     =/  key=message-key:a  [id.writ time]
     =/  mention
-      (was-mentioned:ch-utils content.writ our.bowl)
+      (was-mentioned:ch-utils content.writ our.bowl ~)
     `[time %dm-post key whom content.writ mention]
   =/  replies=(list [time incoming-event:a])
     %-  zing
@@ -1335,7 +1342,7 @@
       (tab:on:replies:ch replies.u.writ `(sub time.key 1) count)
     |=  [=time =reply:ch]
     =/  mention
-      (was-mentioned:ch-utils content.reply our.bowl)
+      (was-mentioned:ch-utils content.reply our.bowl ~)
     [time %dm-reply key parent whom content.reply mention]
   =/  init-time
     ?:  &(=(writs ~) =(replies ~))  recency.unread
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -961,6 +961,12 @@
   |=  [agent=term =path]
   ^-  ^path
   (welp /(scot %p our.bowl)/[agent]/(scot %da now.bowl) path)
+++  get-vessel
+  |=  [=flag:g =ship]
+  ~&  ['scrying for vessel' flag ship]
+  =/  =path
+    /groups/(scot %p p.flag)/[q.flag]/fleet/(scot %p ship)/vessel/noun
+  .^(vessel:fleet:g %gx (scry-path %groups path))
 ++  ca-core
   |_  [=nest:c channel=v-channel:c gone=_|]
   ++  ca-core  .
@@ -1001,7 +1007,10 @@
       ?:  =(author-ship our.bowl)
         =/  =source  [%channel nest group.perm.perm.channel]
         (send [%read source [%all `now.bowl |]] ~)
-      =/  mention=?  (was-mentioned:utils content our.bowl)
+      =/  =vessel:fleet:g  (get-vessel group.perm.perm.channel our.bowl)
+      ~&  ['got vessel' vessel]
+      =/  mention=?  (was-mentioned:utils content our.bowl `vessel)
+      ~&  ['was mentioned' mention]
       =/  action
         [%add %post [[author-ship id] id] nest group.perm.perm.channel content mention]
       (send ~[action])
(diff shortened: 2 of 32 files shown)


desk/app/channels.hoon
```diff
@@ -961,6 +961,11 @@
   |=  [agent=term =path]
   ^-  ^path
   (welp /(scot %p our.bowl)/[agent]/(scot %da now.bowl) path)
+++  get-vessel
+  |=  [=flag:g =ship]
+  =/  =path
+    /groups/(scot %p p.flag)/[q.flag]/fleet/(scot %p ship)/vessel/noun
+  .^(vessel:fleet:g %gx (scry-path %groups path))
```
Fang-, line 968: This probably wants to do an existence check scry (`/exists/[p.flag]/[q.flag]`?) before doing this one. Even if you generally expect to be operating on groups that exist (because you're getting channel activity for them) you're still risking a full-on crash if that assumption ever gets broken.

Running additional `.^`s on processing every individual post will hurt performance, scry overhead remains non-trivial. You might consider wrapping a `~+` around `+get-vessel` so that we don't re-run the same scries within the same event. Alternatively, add a `/fleet` endpoint to groups agent so that we can scry and cache a single endpoint.

desk/app/channels.hoon
```diff
@@ -2602,6 +2613,10 @@
       ?+  -.inline  |
         ?(%bold %italics %strike %blockquote)  ^$(p.verse p.inline)
         ?(%code %inline-code)                  $(inline p.inline)
+      ::
+          %sect
+        ?~  p.inline  $(inline '@all@everyone@channel')
+        $(inline p.inline)
```
Fang-, line 2619: For consistency, shouldn't specific sects also get an `@` prefix?

The whole thing is slightly questionable anyway, since it ties you to client-side input/representation logic, but I don't think there's a better answer if you want to include these in search intuitively. But, `@all@everyone@channel` should probably be checked as separate strings.

```suggestion
          %sect
        ?.  =(~ p.inline)
          $(inline (cat 3 '@' p.inline))
        ?|  $(inline '@all')
            $(inline '@everyone')
            $(inline '@channel')
        ==
```

Fang-: Also am I blind or do we just ignore `%ship` nodes lol

desk/app/activity.hoon
```diff
@@ -278,6 +278,11 @@
   %+  welp
   /(scot %p our.bowl)/[dude]/(scot %da now.bowl)
   path
+++  get-vessel
+  |=  [=flag:g =ship]
+  =/  =path
+    /groups/(scot %p p.flag)/[q.flag]/fleet/(scot %p ship)/vessel/noun
+  .^(vessel:fleet:g %gx (scry-path %groups path))
```
Fang-, line 285: Same comment about scry safety here.

Fang- (commented):
Some comments. Biggest blocker is scry safety. I also have some nits about search.

desk/lib/channel-utils.hoon
```diff
@@ -739,8 +748,13 @@
         ?(%code %inline-code)  ''
         %ship                  (scot %p p.c)
         %task                  (flatten [%inline q.c]~)
+    ::
         ?(%italics %bold %strike %blockquote)
       (flatten [%inline p.c]~)
+    ::
+        %sect
+      ?~  p.c  '@all'
+      (cat 3 '@' (scot %tas p.c))
```
mikolajpp, line 757: This is not really regarding the backend code, but I was curious whether or not these are going to be rendered differently from nicknames at the client. If not, there is a collision risk when someone cunningly sets their nickname to a role or even 'all'

mikolajpp (approved):
Apart from issues raised by @Fang-, everything else looks good.
I left one question about the rendering of mentions.

desk/app/channels.hoon
mikolajpp: If we don't want to couple search with client representation the solution would be to perform structured search. This would require client to submit a search `inline`/`verse` rather than a string as it currently does, but it does not seem overly difficult.

arthyn: yeah we should look into that. I will just use `@all` for now and make sure sects are prefixed with `@` as well

desk/lib/channel-utils.hoon
arthyn: at this time, no they aren't rendered differently. which is also a problem currently as you can set a nickname to match someone else or impersonate a ship. as far as the actual mention mechanism itself, I think we're safe as long as someone doesn't make a role named "all". This would probably trip up the frontend but maybe I can make a special ID for the frontend to prevent this from being an issue. Someone would just have to be careful between choosing all the role vs all.

packages/app/ui/components/MentionPopup.tsx
```diff
@@ -10,63 +11,91 @@ import {
...
+  const size = '$4xl';
+  return (
+    <Pressable borderRadius="$xl" onPress={handlePress}>
+      <ListItem
+        alignItems="center"
+        justifyContent="flex-start"
+        // setting the width to the screen width - 40 so that we can use
+        // ellipsizeMode="tail" to truncate the text
+        // width={Dimensions.get('window').width - 40}
+        // this is a hack to make the text not overflow the container
```
patosullivan, line 47:
```suggestion
```

I'm guessing we don't need this anymore?

desk/app/activity.hoon
```diff
@@ -278,6 +278,22 @@
...
+  ?.  groups-running  ~
+  =/  group-exists
+    =>  [scry-path=scry-path flag=flag ..zuse]  ~+
+    .^(? %gx (scry-path %groups /exists/(scot %p p.flag)/[q.flag]/noun))
+  ?.  group-exists  ~
+  =/  =path
+    %+  scry-path  %groups
+    /groups/(scot %p p.flag)/[q.flag]/fleet/(scot %p ship)/vessel/noun
+  =>  [vessel=vessel:fleet:g path=path ..zuse]  ~+
+  (some .^(vessel %gx path))
```
Fang-, line 296: `scry-path` is gonna have all the core/arm's context inside of it, so you'll miss the cache if you even just change state here before calling it again.

And, for any cache hit, the base path's gonna be the same, so you can just key off that and do a single cache entry:

```suggestion
  ^-  (unit vessel:fleet:g)
  =/  base-path
    (scry-path %groups /)
  =>  [flag=flag ship=ship base-path=base-path ..zuse]  ~+
  =/  groups-running
    .^(? %gu (weld base-path /$))
  ?.  groups-running  ~
  =/  group-exists
    .^(? %gx (weld base-path /exists/(scot %p p.flag)/[q.flag]/noun))
  ?.  group-exists  ~
  %-  some
  .^  vessel  %gx
    %+  weld  base-path
    /groups/(scot %p p.flag)/[q.flag]/fleet/(scot %p ship)/vessel/noun
  ==
```

desk/app/channels.hoon
```diff
@@ -2624,6 +2624,7 @@
       ?+  -.inline  |
         ?(%bold %italics %strike %blockquote)  ^$(p.verse p.inline)
         ?(%code %inline-code)                  $(inline p.inline)
+        %ship                                  $(inline (scot %p p.inline))
```
Fang-, line 2627: Perfect, thank you.

packages/app/ui/components/MessageInput/index.tsx
```diff
@@ -426,31 +429,31 @@ export const MessageInput = forwardRef<MessageInputHandle, MessageInputProps>(
...
+      //   : null;
+      // if (mentionTextFromInline !== null) {
+      //   messageInputLogger.log('Mention text', mentionTextFromInline);
+      //   // if we have a mention text, we show the mention popup
+      //   setShowMentionPopup(true);
+      //   setMentionText(mentionTextFromInline);
+      // } else {
+      //   setShowMentionPopup(false);
+      //   setMentionText('');
+      // }
```
patosullivan, line 456:
```suggestion
```

packages/app/ui/components/MessageInput/index.tsx
```diff
@@ -624,8 +627,8 @@ export const MessageInput = forwardRef<MessageInputHandle, MessageInputProps>(
         // @ts-expect-error setContent does accept JSONContent
         editor.setContent(newJson);
         storeDraft(newJson, draftType);
-        setMentionText('');
-        setShowMentionPopup(false);
+        // setMentionText('');
+        // setShowMentionPopup(false);
```
patosullivan, line 631:
```suggestion
```

patosullivan (approved):
Looks good, just some errant console logs and commented out code.

Also need to fix some type issues. 

After that, it's good to go imo.

PR #4718, opened 2025-05-01
desktop: show user profile in ActivityNavigator

patosullivan:
fixes tlon-4074

There was no way to get to user profiles from the Activity tab on desktop, so clicking on a status update (or avatar change) activity item would do nothing.

EditProfileScreen comes along for the ride too so that users can edit nicknames/avatars for their contacts.

### DIFF
--- a/packages/app/features/settings/EditProfileScreen.tsx
+++ b/packages/app/features/settings/EditProfileScreen.tsx
@@ -22,20 +22,12 @@ export function EditProfileScreen({ route, navigation }: Props) {
     [navigation]
   );
 
-  const handleGoBack = useCallback(() => {
-    // We'll always go back to the profile screen of the user being edited
-    // navigation.goBack() was sometimes not working as expected (particularly in the Activity tab)
-    navigation.navigate('UserProfile', {
-      userId: route.params.userId,
-    });
-  }, [navigation, route.params.userId]);
-
   return (
     <GroupsProvider groups={groups ?? []}>
       <AttachmentProvider canUpload={canUpload} uploadAsset={store.uploadAsset}>
         <EditProfileScreenView
           userId={route.params.userId}
-          onGoBack={handleGoBack}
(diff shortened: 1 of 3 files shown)


packages/app/features/settings/EditProfileScreen.tsx
```diff
@@ -22,12 +22,20 @@ export function EditProfileScreen({ route, navigation }: Props) {
     [navigation]
   );
 
+  const handleGoBack = useCallback(() => {
+    // We'll always go back to the profile screen of the user being edited
```
latter-bolden, line 26: This appears to work and I don't think there's anywhere else that we link to _EditProfileScreen_ from.

Setting `backBehavior="history"` on the ActivityNavigator seems to achieve the same thing. I wonder if we might want to do that to avoid special casing here?

packages/app/features/top/UserProfileScreen.tsx
```diff
@@ -90,8 +90,11 @@ export function UserProfileScreen({ route, navigation }: Props) {
 
   const shouldShowBackButton = useMemo(() => {
     const isWebDesktop = isWeb && !isWindowNarrow;
-    const isContactsTabRoot = navigation.getState().history?.length === 1;
-    return !(isWebDesktop && isContactsTabRoot);
+    const navHistory = navigation.getState().history;
+    const isContactsTabRoot = navHistory?.length === 1;
+    // @ts-expect-error - key is a valid property
+    const isActivityTab = navHistory?.[0]?.key.includes('ActivityEmpty');
+    return !((isWebDesktop && isContactsTabRoot) || isActivityTab);
```
latter-bolden, line 97: we want the same `isWebDesktop` check for the `isActivityTab` conditional, right? On mobile, we need that back button

latter-bolden (approved):
Looking good locally.

Probably want to make sure we can still navigate back on from Activity > Profile on mobile before merging.

PR #4721, opened 2025-05-01
groups agent: remove sort w/alphabetical ordering from drop-fleet

patosullivan:
fixes tlon-4142

we just sort alphabetically on the frontend instead.


Fang- (approved):
Looks Fine™.

PR #4724, opened 2025-05-01
mention: oddities omnibus

arthyn:
Fixes TLON-4130 where trying to search a mention with regex chars would break the app, and a few other things I found when working with mentions:

- we didn't allow you to select a mention if you only typed the trigger character and didn't search
- we were triggering mentions if you put a mention trigger character in the middle of the word. I think its fine to make it require whitespace before it (or be first character) 
- when mentioning the mention in the input would only be the patp not the nickname
- we tend to not show the full ID which in most cases won't matter for normal users, but for power users who have moons this made it impossible to know which moon it was
- if someone had never modified their profile and was in the group with you and never posted, they would be missing from the mentions list because they wouldn't have a `contact` field, we just yield the empty contact instead


patosullivan (approved):
looks good!

PR #4726, opened 2025-05-01
groups: use new thread for group creation

latter-bolden:
Uses the new thread we have for handling all group creation steps. In the current app, those two entry points are:
- default create group with a single initial chat channel
- creating the wayfinding "personal group" that has a configured title, icon, and standard set of channels

Cannot be merged until #4716 is merged and deployed.


patosullivan (approved):
Code looks great

PR #4733, opened 2025-05-05
settings: set theme on back-end, update across clients

jamesacklin:
Fixes TLON-4118 by storing the theme on the back-end and syncing across clients.

- Adds proper settings bucket support for themes
- Syncs across devices by listening for changes to the theme setting
- Appropriately handles the "auto" (null) theme selection
- Adds logging for theme setting + errors

https://github.com/user-attachments/assets/47412726-be97-4db6-9e9c-5bcffc28d389

### DIFF
--- a/packages/app/features/settings/ThemeScreen.tsx
+++ b/packages/app/features/settings/ThemeScreen.tsx
@@ -1,7 +1,6 @@
 import { NativeStackScreenProps } from '@react-navigation/native-stack';
+import { useThemeSettings } from '@tloncorp/shared';
 import { subscribeToSettings } from '@tloncorp/shared/api';
-import { themeSettings } from '@tloncorp/shared/db';
-import { useSettings } from '@tloncorp/shared/store';
 import { useContext, useEffect, useState } from 'react';
 import { ScrollView, YStack } from 'tamagui';
 import type { ThemeName } from 'tamagui';
@@ -25,6 +24,7 @@ type Props = NativeStackScreenProps<RootStackParamList, 'Theme'>;
 
 export function ThemeScreen(props: Props) {
   const theme = useTheme();
+  const { data: storedTheme, isLoading } = useThemeSettings();
   const { setActiveTheme } = useContext(ThemeContext);
   const isDarkMode = useIsDarkMode();
   const [selectedTheme, setSelectedTheme] = useState<ThemeName | 'auto'>(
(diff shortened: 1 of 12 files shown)


packages/app/features/settings/ThemeScreen.tsx
```diff
@@ -48,29 +50,93 @@ export function ThemeScreen(props: Props) {
...
 
+  // Handle theme syncing and updates
   useEffect(() => {
-    const checkSelected = async () => {
-      const storedTheme = await themeSettings.getValue();
-      setSelectedTheme(storedTheme ?? 'auto');
+    const syncAndUpdateTheme = async () => {
+      try {
+        // First sync with backend to get latest settings
+        await syncSettings();
```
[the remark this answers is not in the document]
jamesacklin: Yes - addressed

jamesacklin:
@dnbrwstr Thanks for the feedback. I've:

- Moved all theme updating logic into the shared store via updateTheme
- Removed the redundant syncing in ThemeScreen.tsx that's already handled by subscription
- Centralized theme conversion logic with helper functions
- Got rid of all the stray comments and logs

dnbrwstr (approved):
Looking good, few more nits!

jamesacklin:
@dnbrwstr OK! I...

- Got rid of the duplicate helper declaration
- Moved the theme mutation out of state
- Set up a new DB hook for settings *in general*, figure this might be useful down the line
- Changed the app provider to consume that hook, then fall back to direct access

Appreciate the help here!

packages/app/provider/index.tsx
```diff
@@ -1,87 +1,129 @@
...
+    </QueryClientProvider>
+  );
+}
+
+function ThemeProviderContent({
+  children,
+  tamaguiProps,
+}: {
+  children: React.ReactNode;
+  tamaguiProps: Omit<TamaguiProviderProps, 'config'>;
```
jamesacklin, line 45: Was following the pattern above where we omit 'config' from TamaguiProviderProps because we're providing our own config import and passing it in directly.

packages/shared/src/urbit/settings.ts
```diff
@@ -1,6 +1,8 @@
+import { ThemeName } from 'tamagui';
+
 import { DisplayMode, SortMode } from './channel';
 
-export type Theme = 'light' | 'dark' | 'auto';
+export type AppTheme = ThemeName | 'auto';
```
jamesacklin, line 5: Not stoked on the duplicate declaration but gets us around the circular dependency issue

packages/shared/src/store/settingsActions.ts
```diff
@@ -6,6 +8,8 @@ import * as logic from '../logic';
 import { withRetry } from '../logic';
 import { TalkSidebarFilter } from '../urbit';
 
+export type AppTheme = ThemeName | 'auto';
```
jamesacklin, line 11: Same here

dnbrwstr (approved):
Nice! Lgtm.

PR #4737, opened 2025-05-05
mobile: add contact matching

patosullivan:
Fixes tlon-4014.

Adds contact matching, uses a forked version of nockjs for the moment while we wait for this PR to get merged: https://github.com/urbit/nockjs/pull/7

The meat of the changes here are in `linkSystemContacts` and the functions in `lanyardApi.ts`.

Copilot generated description:

### Contact Discovery and Synchronization Enhancements:
* **Added `syncContactDiscovery` method**: Implements a new flow to discover and link contacts by phone numbers, ensuring duplicate contacts are filtered out and new matches are processed. (`packages/shared/src/store/sync.ts`, [[1]](diffhunk://#diff-6b06da71b059f3a3eef1b65d12d0f78b1b38464c89984c0b0cb85ff6c186c518R272-R309) [[2]](diffhunk://#diff-6b06da71b059f3a3eef1b65d12d0f78b1b38464c89984c0b0cb85ff6c186c518R1595-R1597)
* **Enhanced `discoverContacts` API**: Refactored the function to handle rate-limiting, logging, and error tracking. It now uses a more robust payload structure and integrates with the database for salt management. (`packages/shared/src/api/lanyardApi.ts`, [[1]](diffhunk://#diff-ae1cd4b8aa48a0d1a1a492c5188f808705e5a15b3bd55fef73fbfb13ccd8f9e6L63-R83) [[2]](diffhunk://#diff-ae1cd4b8aa48a0d1a1a492c5188f808705e5a15b3bd55fef73fbfb13ccd8f9e6L85-R176)
* **Updated `ContactBookPrompt`**: Calls the new `syncContactDiscovery` method after syncing system contacts. (`packages/app/ui/components/SystemNotices.tsx`, [packages/app/ui/components/SystemNotices.tsxR53](diffhunk://#diff-41de21cc5b05b697e4231bca34cb93cbeb38df2e605900db61485e1a10efb6a8R53))

### Database Improvements:
* **Introduced new storage items**: Added `lastLanyardSalt` and `lastPhoneContactSetRequest` for managing contact discovery state. (`packages/shared/src/db/keyValue.ts`, [packages/shared/src/db/keyValue.tsR325-R336](diffhunk://#diff-e60050fa0b9ae6b31ef3b2930557bc12d2a2bd4e5f5f576ec94079745790fd9cR325-R336))
* **Refactored contact linking queries**: Enhanced the `linkSystemContacts` query to handle nickname updates, system contact linking, and direct message (DM) channel creation in batches. (`packages/shared/src/db/queries.ts`, [[1]](diffhunk://#diff-efcb4cf7d0bc50169d5f497604e75d495ff54a546aaee16e0bf48a7841bab03dL402-R629) [[2]](diffhunk://#diff-efcb4cf7d0bc50169d5f497604e75d495ff54a546aaee16e0bf48a7841bab03dR657-R661)

### Dependency Updates:
* **Updated `@urbit/nockjs`**: Replaced the version with a GitHub-hosted branch (`modern-mug`) to incorporate the latest changes. (`apps/tlon-mobile/package.json`, [[1]](diffhunk://#diff-8deea517552ee81bbb5b9943b366a4b19324af6c3ba8d60ee96aee46c8cc7fcbL70-R70); `packages/shared/package.json`, [[2]](diffhunk://#diff-3752b1e4e3e2032593e21e9268ec0379680a59e4e7b8517d0d5492a71530d3e8L57-R57); `pnpm-lock.yaml`, [[3]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL239-R240) [[4]](diffhunk://#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL1072-R1073)

### Logging and Error Tracking:
* **Improved logging**: Added detailed logs for debugging and tracking across various functions, including `discoverContacts`, `linkSystemContacts`, and `syncContactDiscovery`. (`packages/shared/src/api/lanyardApi.ts`, [[1]](diffhunk://#diff-ae1cd4b8aa48a0d1a1a492c5188f808705e5a15b3bd55fef73fbfb13ccd8f9e6L85-R176); `packages/shared/src/db/queries.ts`, [[2]](diffhunk://#diff-efcb4cf7d0bc50169d5f497604e75d495ff54a546aaee16e0bf48a7841bab03dL402-R629)
* **Enhanced error handling**: Integrated error tracking for contact discovery and database operations. (`packages/shared/src/store/contactActions.ts`, [packages/shared/src/store/contactActions.tsR42-R61](diffhunk://#diff-b09736755d648b6c63e6e8dcaa3928f085842785dcec5c9fe0af5468cf1d9dbbR42-R61))

### Miscellaneous:
* **Refactored `insertChannels` logic**: Extracted the core functionality into `insertChannelsInternal` to streamline code and improve maintainability. (`packages/shared/src/db/queries.ts`, [[1]](diffhunk://#diff-efcb4cf7d0bc50169d5f497604e75d495ff54a546aaee16e0bf48a7841bab03dL1858-R2095) [[2]](diffhunk://#diff-efcb4cf7d0bc50169d5f497604e75d495ff54a546aaee16e0bf48a7841bab03dL1887-R2137)

These changes collectively enhance the functionality, reliability, and maintainability of the contact discovery and synchronization system.


packages/shared/src/api/lanyardApi.ts
```diff
@@ -82,32 +92,91 @@ export async function discoverContacts(phoneNums: string[]) {
...
-      const nextSalt = queryResponse.result?.['next-salt'];
-      const matches = queryResponse.result?.result
-        ? Object.entries(queryResponse.result.result).filter(([_key, value]) =>
-            Boolean(value)
-          )
-        : [];
-
-      return matches;
+      logger.trackError('contact discovery poke error', e);
+      logger.trackEvent(AnalyticsEvent.DebugContactMatching, {
```
latter-bolden, line 96: Instead of double logging here, can we add a `ErrorContactMatching` event type and tack on a `context` property to signal where it failed? 

See `DebugSystemContacts` and `ErrorSystemContacts` as an example.

packages/shared/src/api/lanyardApi.ts
```diff
@@ -82,32 +92,91 @@ export async function discoverContacts(phoneNums: string[]) {
...
+        errorMessage: e.message,
+      });
     }
+    try {
+      const queryResponse = await queryResponseSub;
+
+      if (queryResponse && queryResponse.query.nonce === encodedNonce) {
+        if (queryResponse.query.result === 'rate limited') {
+          logger.log('rate limited');
+          logger.trackEvent(AnalyticsEvent.DebugContactMatching, {
```
latter-bolden, line 107: See other comment, lets differentiate error events from debug ones

packages/shared/src/api/lanyardApi.ts
```diff
@@ -82,32 +92,91 @@ export async function discoverContacts(phoneNums: string[]) {
...
+
+    // find new phone numbers (additions)
+    const diff = phoneNums.filter((num) => !lastSet.includes(num));
+    logger.log('diff', diff);
+
+    // find removed phone numbers (deletions)
+    const delSet = lastSet.filter((num: string) => !phoneNums.includes(num));
+    logger.log('delSet', delSet);
+
+    // always store the CURRENT complete set of phone numbers
```
latter-bolden, line 157: how does immediately updating the stored set work if the actual `%verifier` request fails? 

If we try again, will there be an issue with previous salt not matching passed `last/add/del`? Would we incur a big hit to our rate limit balance?

packages/shared/src/api/lanyardApi.ts
```diff
@@ -82,32 +92,91 @@ export async function discoverContacts(phoneNums: string[]) {
...
   }
 }
 
-function diffContactBook(phoneNums: string[]) {
-  // TODO: diff last request
+async function diffContactBook(phoneNums: string[]) {
+  const last = await db.lastPhoneContactSetRequest.getValue();
+  logger.log('last phone contact set', last);
+
+  if (last) {
```
latter-bolden, line 145: if we process a difference, can add a debug log for the magnitude of change?

packages/shared/src/store/contactActions.ts
```diff
@@ -38,10 +39,26 @@ export async function addContacts(contacts: string[]) {
     })
   );
   await Promise.all(optimisticUpdates);
+  logger.log('Optimistic updates complete', {
+    optimisticUpdates,
+    contacts,
+  });
+
+  // Backend will balk if we try to add the same contact twice, so filter out
```
latter-bolden, line 47: Nice! Good call

packages/shared/src/store/sync.ts
```diff
@@ -268,6 +269,44 @@ export const syncSystemContacts = async (ctx?: SyncCtx) => {
...
 
+export const syncContactDiscovery = async (ctx?: SyncCtx) => {
+  const systemContacts = await api.getSystemContacts();
+  const phoneNumbers = systemContacts
+    .map((contact) => contact.phoneNumber)
+    .filter((phoneNumber) => phoneNumber && phoneNumber.length > 0) as string[];
+
+  if (!phoneNumbers.length) {
+    logger.trackEvent(AnalyticsEvent.DebugSystemContacts, {
+      context: 'no phone numbers found, skipping discovery',
```
latter-bolden, line 280: 🙏🏻

packages/shared/src/store/sync.ts
```diff
@@ -268,6 +269,44 @@ export const syncSystemContacts = async (ctx?: SyncCtx) => {
...
+      await syncQueue.add('discoverContacts', ctx, () =>
+        api.discoverContacts(phoneNumbers)
+      )
+    ).filter((match) => match[1] !== api.getCurrentUserId());
+    logger.log('got contact discovery matches', matches);
+
+    await db.linkSystemContacts({ matches });
+    logger.log('inserted contact discovery matches', matches);
+    const contactIds = matches.map((m) => m[1]);
+    await addContacts(contactIds);
```
latter-bolden, line 298: can we add a separate error contexts for api call failure vs add contacts failure? Would be helpful to distinguish between the two if we have to debug

packages/shared/src/store/sync.ts
```diff
@@ -268,6 +269,44 @@ export const syncSystemContacts = async (ctx?: SyncCtx) => {
   }
 };
 
+export const syncContactDiscovery = async (ctx?: SyncCtx) => {
```
latter-bolden, line 272: do we want to confirm they have an attested phone number before trying discovery? If we know we're going to get no matches might make sense to short circuit + avoid rate limit hit

packages/shared/src/store/sync.ts
```diff
@@ -268,6 +269,44 @@ export const syncSystemContacts = async (ctx?: SyncCtx) => {
...
+
+  try {
+    const matches = (
+      await syncQueue.add('discoverContacts', ctx, () =>
+        api.discoverContacts(phoneNumbers)
+      )
+    ).filter((match) => match[1] !== api.getCurrentUserId());
+    logger.log('got contact discovery matches', matches);
+
+    await db.linkSystemContacts({ matches });
```
latter-bolden, line 295: I know we're updating the local DB contact nickname's to match the system contact, but are we ever propagating that change to the backend? When trying locally, my discovery matches fell back to their on-network nicknames.

packages/shared/src/db/queries.ts
```diff
@@ -1855,6 +2083,44 @@ export const getChannelWithRelations = createReadQuery(
   ]
 );
 
+async function insertChannelsInternal(channels: Channel[], ctx: QueryCtx) {
```
latter-bolden, line 2086: I like that we're initializing a DM for matches, but it feels weird to have it be empty in the ChatList. We probably want to initialize with some sort of notice message? If we don't, the other party won't actually have a DM on their end, right?

There's something a bit tricky here — if we (User A) match with User B, User B would probably expect to see the name of their native system contact for User A (if one exists) 🤔

packages/app/ui/components/SystemNotices.tsx
```diff
@@ -50,6 +50,7 @@ export function ContactBookPrompt(props: {
       const result = await perms.requestPermissions();
       if (result === 'granted') {
         await store.syncSystemContacts();
+        await store.syncContactDiscovery();
```
latter-bolden, line 53: can we make sure we always dismiss the notice if discovery fails? I'm guessing that'll have a higher error rate than `syncSystemContacts` which is all local.

packages/shared/src/api/lanyardApi.ts
```diff
@@ -60,15 +60,25 @@ export async function checkAttestedSignature(signData: string) {
...
+export async function discoverContacts(
+  phoneNums: string[]
+): Promise<[string, string][]> {
   try {
-    const nums = diffContactBook(phoneNums);
-    // const lastSalt = formatUw('0');
-    const payload = ['whose-bulk', 0, nums.last, nums.add, nums.del];
-    const noun = dwim(payload);
-
+    const nums = await diffContactBook(phoneNums);
```
latter-bolden, line 67: should we short circuit if there's no difference since the last request?

packages/shared/src/store/sync.ts
latter-bolden: we may also want to distinguish `db.linkSystemContacts` failures since that logic is complicated

latter-bolden (changes requested):
This looks awesome, was really slick when I tried it out as a new signup.

Mostly nits, feedback about the new DM's we're initializing is the biggest question in my mind.

Outside of the review comments, I had one other thing to address: we need to account for system contact matches whenever prompting to invite people from your contact book (bottom of the contacts tab and _InviteSystemContacts_ screen). If I already have you as a match, the app shouldn't ask me to invite you again. This'll also be important when we layer in the invite step at the end of the signup splash.

packages/shared/src/store/sync.ts
patosullivan: propagating the nickname change? no, good point

packages/shared/src/api/lanyardApi.ts
patosullivan: Good point, we could only write it out after we successfully hear back from the verifier.

packages/shared/src/db/queries.ts
patosullivan: Yeah, I wanted to see what others thought on this one

packages/shared/src/store/sync.ts
patosullivan: fixed in the next commit I'm going to push, but note that we're also going to no-op if we don't have a diff/del set, so you won't see your contact names get updated unless you comment out that early return (or uninstall/reinstall the app).

packages/shared/src/api/lanyardApi.ts
Fang-: Except new numbers might have gotten verified, so you'll want to just send the query anyway, no?

apps/tlon-mobile/package.json
```diff
@@ -67,7 +67,7 @@
     "@tloncorp/ui": "workspace:*",
     "@urbit/aura": "^2.0.1",
     "@urbit/http-api": "3.2.0-dev",
-    "@urbit/nockjs": "^1.5.0",
+    "@urbit/nockjs": "github:Fang-/nockjs#m/modern-mug",
```
Fang-, line 70: This can be master branch now. Or the specific commit, might be better. (Until we push a new minor version, of course.)

```suggestion
    "@urbit/nockjs": "github:urbit/nockjs#8ed20e5",
```

packages/shared/src/api/lanyardApi.ts
```diff
@@ -60,15 +60,35 @@ export async function checkAttestedSignature(signData: string) {
...
+      logger.log('discoverContacts: no changes, returning empty');
+      logger.trackEvent(AnalyticsEvent.DebugContactMatching, {
+        context: 'discoverContacts',
+        diffSetLength: 0,
+        delSetLength: 0,
+        message: 'No changes, no need to send request',
+      });
+      return [];
+    }
+    const lastSalt = (await db.lastLanyardSalt.getValue()) ?? formatUw('0');
```
Fang-, line 78: The `lastSalt` in the default case is sending the string `'0w0'` instead of a number value. Not that it matters in the default case, but you should be fine just sending `0` (or `0n`).

The json string that you get and inject into the database directly (below, L128-129) is going to be a `@ux`-encoded number. You'll want to pass _the number value_, not its string representation, into `dwim`/`nounify`. Probably:

```suggestion
    const lastSalt = BigInt('0x' + parseUx((await db.lastLanyardSalt.getValue()) ?? '0x0'), 16);
```

(Would be nice if json supported `bigint`s, then we wouldn't be forced to do song-and-dance here... ): )

packages/shared/src/api/lanyardApi.ts
patosullivan: Missed this earlier, will update.

apps/tlon-mobile/package.json
```diff
@@ -67,7 +67,7 @@
     "@tloncorp/ui": "workspace:*",
     "@urbit/aura": "^2.0.1",
     "@urbit/http-api": "3.2.0-dev",
-    "@urbit/nockjs": "^1.5.0",
+    "@urbit/nockjs": "github:urbit/nockjs#8ed20e5",
```
dnbrwstr, line 70: Does mobile need the nockjs dep? Doesn't seem to be required directly anywhere.

packages/shared/src/api/lanyardApi.ts
```diff
@@ -60,15 +60,28 @@ export async function checkAttestedSignature(signData: string) {
...
+  phoneNums: string[]
+): Promise<[string, string][]> {
   try {
-    const nums = diffContactBook(phoneNums);
-    // const lastSalt = formatUw('0');
-    const payload = ['whose-bulk', 0, nums.last, nums.add, nums.del];
-    const noun = dwim(payload);
-
+    const nums = await diffContactBook(phoneNums);
+    const storedLastSalt = await db.lastLanyardSalt.getValue();
```
dnbrwstr, line 68: Can we pass this in to the function? Or make part of an action? In general we shouldn't be accessing `db` from `api`

packages/shared/src/api/lanyardApi.ts
```diff
@@ -82,32 +95,97 @@ export async function discoverContacts(phoneNums: string[]) {
...
-    console.error('error discovering contacts', e);
+    logger.error('error discovering contacts', e);
     throw e;
   }
 }
 
-function diffContactBook(phoneNums: string[]) {
-  // TODO: diff last request
+async function diffContactBook(phoneNums: string[]) {
+  const last = await db.lastPhoneContactSetRequest.getValue();
```
dnbrwstr, line 149: Similar to above, would say this should be an action rather than an api method.

packages/shared/src/db/queries.ts
```diff
@@ -399,10 +399,237 @@ export const insertSystemContacts = createWriteQuery(
   ['systemContacts', 'systemContactSentInvites', 'contacts']
 );
 
-export const linkSystemContact = createWriteQuery(
-  'insertSystemContactSentInvites',
-  async (params: { sentInvites: SystemContactSentInvite[] }, ctx: QueryCtx) => {
-    // TODO
+export const linkSystemContacts = createWriteQuery(
```
dnbrwstr, line 402: Not blocking, but feels like this could be simplified a bit.

dnbrwstr (changes requested):
Looking good overall, a few nits, and I think organization could be improved a bit by moving some pieces to contactActions.

packages/shared/src/db/queries.ts
```diff
@@ -399,10 +399,237 @@ export const insertSystemContacts = createWriteQuery(
...
+  'linkSystemContacts',
+  async (params: { matches: [string, string][] }, ctx: QueryCtx) => {
+    if (!params.matches.length) return;
```
dnbrwstr, line 405: Looking at this again, I do think this function is pretty hard to parse through. Could you look at:
- Reducing amount of logging to what's most useful
- Making sure all try-catches are necessary (eg, there's a try-catch around the whole thing, but there's also a try-catch around the call site in `sync.ts`, making it redundant)
- Possibly refactoring parts into helper functions to make control flow easier to follow

dnbrwstr (changes requested):
Looking good! One more nit.

dnbrwstr (approved):
Looks good to me! Last comment is I'd skip the new file for contact helpers -- the channel creator could go in `modelBuilders`, and the query helpers could stay in `queries`, just to keep things consistent.

PR #4738, opened 2025-05-06
verifier: logging & maintenance details

Fang-:
Two small fixes.

We weren't including the `src.bowl` in the logs for user queries, leading to "rate limited" info logs without the affected ship being legible.

Shared pool rate limiting allowance was being set through `$_`'s implicit bunt, but this interferes with the `;;` we do for dbug safety purposes, so we switch to doing a proper `$~` instead. (Same change as 3429eda.)

### DIFF
--- a/desk/app/verifier.hoon
+++ b/desk/app/verifier.hoon
@@ -740,6 +740,7 @@
     ==
   ::
       %verifier-user-query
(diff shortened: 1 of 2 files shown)


mikolajpp (approved):
Looks good.

PR #4740, opened 2025-05-06
groups: dont loop on section moves

arthyn:
Fixes TLON-4187. We were accidentally calling the API here when hearing a fact to update our local DB, but sending this to the API created a new fact which then triggered the same flow over and over. This switches that handling to only modify local DB.

### DIFF
--- a/packages/shared/src/store/groupActions.ts
+++ b/packages/shared/src/store/groupActions.ts
@@ -716,33 +716,25 @@ export async function addChannelToNavSection({
   }
 }
 
-export async function moveChannel({
-  groupId,
-  channelId,
+export async function updateChannelSections({
+  group,
   navSectionId,
+  channelId,
   index,
 }: {
-  groupId: string;
-  channelId: string;
+  group: db.Group;
   navSectionId: string;
+  channelId: string;
   index: number;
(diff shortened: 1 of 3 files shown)


packages/shared/src/store/sync.ts
```diff
@@ -838,20 +837,26 @@ async function handleGroupUpdate(update: api.GroupUpdate, ctx: QueryCtx) {
       break;
     case 'moveChannel':
       logger.log('moving channel', update);
-      await moveChannel({
+      // eslint-disable-next-line
```
latter-bolden, line 840: why do we need eslint disable here? If it's the variable thing, should we fetch the group in the action instead?

packages/shared/src/store/groupActions.ts
```diff
@@ -716,33 +716,25 @@ export async function addChannelToNavSection({
   }
 }
 
-export async function moveChannel({
-  groupId,
-  channelId,
+export async function updateChannelSections({
```
latter-bolden, line 719: looks good, but could we add tracked events for success and error?

latter-bolden (approved):

PR #4743, opened 2025-05-07
verifier: log twitter api responses

Fang-:
For investigating supposed twitter api flakiness, we'll want to know the kinds of responses we get back. Include the status code and response body in the logs.

Already pushed this onto the service, but PRing for persistence.


arthyn (approved):
lgtm

PR #4744, opened 2025-05-07
GroupRolesScreen: generate db-safe IDs for non-alpha-numeric role titles

jamesacklin:
Fixes TLON-4202. 

When creating a role, we honor alpha-numeric roles by switching the inputted title to lower-case, replacing any spaces with a hyphen, and using that as the ID.

Unfortunately, if the user inputs a single emoji as the title, we don't know how to handle it, and the role fails to create.

This PR:

- regexes for an alpha-numeric role title
- if it passes, we employ the previous ID generation logic
- if not, we generate an ID prefixed with `role-` followed by a random string

Since the emoji lives in the role metadata, it shows up whenever we call the role's name and the rest of the app gets to use the valid yet non-visible ID.

cc @thelifeandtimes


patosullivan (approved):
Looks good, makes sense

packages/app/features/groups/GroupRolesScreen.tsx
```diff
@@ -96,8 +96,15 @@ function GroupRolesScreenView() {
 
   const handleAddRole = useCallback(
     ({ title, description }: { title: string; description: string }) => {
+      const generateSafeId = (text: string) => {
```
dnbrwstr, line 99: Might make sense to put this in `logic/utils`

packages/app/features/groups/GroupRolesScreen.tsx
```diff
@@ -149,9 +156,9 @@ function GroupRolesScreenView() {
                 borderRadius="$2xl"
               >
                 <ActionSheet.MainContent>
-                  <ActionSheet.ActionTitle>
+                  <Text color="$primaryText" fontSize="$l" numberOfLines={1}>
```
dnbrwstr, line 159: Nonstandard text

dnbrwstr (approved):
Lgtm, minor nits.

PR #4745, opened 2025-05-07
http-api: fix thread abort signal

latter-bolden:
OTT

Thread `POST`s will no longer fail due to SSE disruption. Also adds more telemetry for thread failures and channel reconnects.

### DIFF
--- a/packages/shared/src/http-api/types.ts
+++ b/packages/shared/src/http-api/types.ts
@@ -143,6 +143,8 @@ export interface Thread<Action> {
    * Data of the input vase
    */
   body: Action;
+
+  timeout?: number;
 }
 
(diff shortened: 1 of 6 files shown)


packages/shared/src/http-api/types.ts
```diff
@@ -236,3 +238,15 @@ export class FatalError extends Error {}
 export class ReapError extends Error {}
 
 export class AuthError extends Error {}
+
+export class SSETimeoutError extends Error {}
```
arthyn, line 242: we didn't end up using this guy in fetch.ts, but we should

arthyn (approved):
just the one change then g2g

PR #4746, opened 2025-05-07
android: navigate to notification subject on tap

davidisaaclee:
fixes TLON-3795
fixes TLON-4097

Makes Android use the same routing logic as iOS when tapping on a notification, and fixes a case where tapping on a notification launched to a blank screen on Android.

## Summary
- **TODO: useLastNotificationResponse commit link** – Use `useLastNotificationResponse` instead of `addNotificationResponseReceivedListener` to act on a notification which was tapped to launch the app. Since we are mounting `useNotificationListener` a few renders after app launch (e.g. blocking on `MigrationCheck`, authentication), this ensures that we don't drop the last-tapped notification. (I don't have hard evidence for this, but it seemed like this was causing issues on Android.)
- [Ensure "extras" from notif tap intent are piped to JS app](https://github.com/tloncorp/tlon-apps/commit/70f616f6e3fb37e9a6c047073fb049735812a37b) – `TalkReactActivityDelegate.java` was explicitly stripping out any extras on the intent which launched the main activity, which includes the data that the notification service was passing to the main activity to say which notification was tapped. https://github.com/tloncorp/tlon-apps/blob/976d61721dd03528f9e3c93f7bc7ae2cccf7adfb/apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/TalkReactActivityDelegate.java#L16-L21 It's not clear why this was included – git history is not helpful here, and I don't see anywhere else where the extras are accessed.
- [Avoid double-migrating on same nativeDb client](https://github.com/tloncorp/tlon-apps/commit/c7d99f1ab129cf7e8fd833b0e0cfde554fef69b3) – This avoids an issue (largely visible with dev clients) where migrations were attempted to be run when app was launched from notification, but on a reused `nativeDb` instance – this caused the migration to be run concurrent to other queries, which for some reason caused a deadlock – **I think this was the cause of the blank screen TLON-4097 bug**, although I don't understand why this was deadlocking, and I don't see what was causing the migrations to run a second time on release builds[^0].

[^0]: Reading code, the migrations would be re-run if this ["migrations succeeded" `useState`](https://github.com/tloncorp/tlon-apps/blob/57a93658da01bc1d5ba36f95df759b4b2cdd68bb/packages/app/lib/baseDb.ts#L20) is "reset" without resetting the `nativeDb` global state ([construction](https://github.com/tloncorp/tlon-apps/blob/57a93658da01bc1d5ba36f95df759b4b2cdd68bb/packages/app/lib/nativeDb.ts#L116), [setup](https://github.com/tloncorp/tlon-apps/blob/57a93658da01bc1d5ba36f95df759b4b2cdd68bb/apps/tlon-mobile/index.js#L27)). The React tree could conceivably be remounting in some normal lifecycle on a release build – I'm not aware of any such event.


dnbrwstr (approved):
Nice, all lgtm!

PR #4747, opened 2025-05-07
web: double press on home icon in nav to reset navigation state (return home)

patosullivan:
fixes tlon-4122

If you're already navigated to some screen in the HomeNavigator and you press the Home Icon, nothing happens because you're already in the "home" tab. With this, if you double press, you'll be taken back to the root home screen/chat list.


dnbrwstr:
Discussed out of band -- code looks good, but I think it'd make sense if home always took you to root if you're already on home tab, rather than requiring double tap.

patosullivan:
closed in favor of https://github.com/tloncorp/tlon-apps/pull/4749

PR #4748, opened 2025-05-07
group settings: navigate back to chat details screen from leaf screens

jamesacklin:
Fixes an issue where using the upper-left caret icon on groups settings screens such as members, roles, etc. would navigate you back to the last-used channel, rather than the group info screen from whence you came. This resulted in onerous setup of groups (inviting, managing roles, editing channels, etc).

Does not introduce navigation regressions on mobile.

cc @hanfel-dovned


dnbrwstr:
Life would be a lot simpler if `goBack` worked correctly -- any idea why it doesn't? Maybe pulling the wrong navigator reference?

dnbrwstr:
Not blocking, and I know we do this kinda thing in other places -- let's take a quick look after standup?

jamesacklin:
@dnbrwstr I added a route parameter that tells `handleGoBack` to either navigate back to the parent stack or to use good old-fashioned `goBack()` if we use the screen in another stack elsewhere. I didn't touch the root stack. I also compressed some functions so they're a little more compact.

dnbrwstr (approved):
Looks good to me overall. A bit verbose, I do wonder if any of this can be simplified down, but nonblocking.

PR #4750, opened 2025-05-08
references: update copy for unavailable references

patosullivan:
fixes tlon-4149

Both errored refs and "not found" refs now just say "Content not available".


jamesacklin:
Does this cover the `cannot read null of undefined` case?

patosullivan:
> Does this cover the `cannot read null of undefined` case?

yep

dnbrwstr (approved):
Lgtm!

PR #4751, opened 2025-05-08
web/privacy: prevent telemetry toggle from showing true when we don't know

patosullivan:
fixes tlon-4119.

I think these users have posthog blocked, which is causing `posthog.has_opted_out_capturing()` to be undefined, so they were getting the default value of `false` for `getIsOptedOut` when it should have been `true`.

What are the other implications for this change, though?

### DIFF
--- a/packages/app/hooks/usePosthog.ts
+++ b/packages/app/hooks/usePosthog.ts
@@ -7,7 +7,7 @@ export function usePosthog() {
   const posthog = useWebPosthog();
   return useMemo((): PosthogClient => {
     return {
-      getIsOptedOut: () => posthog?.has_opted_out_capturing() ?? false,
+      getIsOptedOut: () => posthog?.has_opted_out_capturing() ?? true,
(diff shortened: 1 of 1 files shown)


packages/app/hooks/usePosthog.ts
```diff
@@ -7,7 +7,7 @@ export function usePosthog() {
   const posthog = useWebPosthog();
   return useMemo((): PosthogClient => {
     return {
-      getIsOptedOut: () => posthog?.has_opted_out_capturing() ?? false,
+      getIsOptedOut: () => posthog?.has_opted_out_capturing() ?? true,
```
dnbrwstr, line 10: Could you add a comment here basically explaining what you did in the description? The the function returns undefined if blocked.

patosullivan: added in latest commit

arthyn (approved):
seems fine?

PR #4752, opened 2025-05-08
native: render post plaintext preview from intermediate PostContent

davidisaaclee:
fixes TLON-4168

> **_🧞 Hidden in this PR is a design proposition..._**
This PR promotes the app model `PostContent` (different from [`api.PostContent`](https://github.com/tloncorp/tlon-apps/blob/dd0972daaca2272341cc3696030f88f03543aa72/packages/shared/src/api/postsApi.ts#L47)) as the canonical intermediate representation for post content. https://github.com/tloncorp/tlon-apps/blob/dd0972daaca2272341cc3696030f88f03543aa72/packages/shared/src/logic/postContent.ts#L143
If accepted, we'll use this `PostContent` whenever interpreting a post's content; then, we'll move the API post schema to match `PostContent`.

## What changed?

The big change here is in how the notifications render text: Black icon notification is from this PR's changes, white icon is current Testflight without this PR.
iPhone/desktop in-app renderings are included for reference, since some formatting is not consistent (iPhone screenshots include this PR's change, but the screenshotted surfaces should not have been affected).

| | iPhone | Desktop |
| --- | --- | --- |
| IMG_2874 IMG_2875 | IMG_2878 | Screenshot 2025-05-08 at 10 05 04 AM |
| IMG_2873 | IMG_2877 | Screenshot 2025-05-08 at 10 05 30 AM |

The plaintext renderer is also used in the "latest post" summary:
| Before this change | After this change |
| --- | --- |
| IMG_8B6DF1BEEE30-1 | IMG_2880 |

The renderer is a little bit configurable (in case we want to e.g. use a double space for newlines): https://github.com/tloncorp/tlon-apps/blob/6b38d5abc05ce72d33054d7e2c3afdf24608bfee/packages/shared/src/logic/postContent.ts#L145-L164

### Why not update the existing `getTextContent`?
`ub.getTextContent` operated directly off of the API's `post.content` – part of this PR is moving to using the in-app `PostContent` as the canonical intermediate representation (which means we get to share logic like automatically turning list items into block-level items, and eventually having a conversion layer for different post content schemas).

## How did I test?
I added a toggle to the `ChatMessage.All` fixture to preview how posts are rendered into plaintext: 

https://github.com/user-attachments/assets/6f02c259-46b9-4ea6-9d3c-cd87233ca669

**other stuff for reviewers:**
- [[wip] Move post content utils to shared/](https://github.com/tloncorp/tlon-apps/pull/4752/commits/4b432a75a9d7032eb3bf8fccc02881a28b0e7f02) is marked `wip` because the project will not build without the subsequent commit, but combining the commits seemed like it'd be noisy
- [Delete unused code](https://github.com/tloncorp/tlon-apps/pull/4752/commits/8ce2f12ef0634c9edc220475643239942819b39a) was intended to only remove the dead `getTextContent` and its helpers, but turned out to be a more aggressive cleanup – I did this by deleting all unused functions, including those which were `export`ed and not imported somewhere else, which ended up removing a handful of other unused members.

### DIFF
--- a/packages/shared/src/logic/postContent.ts
+++ b/packages/shared/src/logic/postContent.ts
@@ -1,4 +1,4 @@
-import type * as api from '../api';
+import * as api from '../api';
 import * as ub from '../urbit';
 import { assertNever } from '../utils';
 import { trustedProviders } from './embed';


packages/app/hooks/useDesktopNotifications.ts
```diff
@@ -1,7 +1,7 @@
 import { createDevLogger } from '@tloncorp/shared';
 import * as api from '@tloncorp/shared/api';
 import * as db from '@tloncorp/shared/db';
-import { getTextContent } from '@tloncorp/shared/urbit';
+import { getTextContent } from '@tloncorp/shared/logic';
```
davidisaaclee, line 4: NB: in all other moves from the old `getTextContent`, I specified `PlaintextPreviewConfig.inlineConfig` to more closely match the existing behavior. Here, I think that including newlines in the desktop notification bodies makes sense. (Same with mobile client notification bodies.)

packages/shared/src/logic/postContent.ts
```diff
@@ -0,0 +1,630 @@
+import type * as api from '../api';
```
davidisaaclee, line 1: This file is largely split from `packages/app/ui/components/PostContent/contentUtils.tsx`, but also includes the core renderer code for the new `getTextContent` (see `plaintextPreviewOf`)

jamesacklin:
End results look great, will leave the code review to others. Nice work!

dnbrwstr (approved):
This makes sense to me + feels like an improvement to the design -- I'm in favor of `BlockData` everywhere.

davidisaaclee:
Merge [1ca53e5](https://github.com/tloncorp/tlon-apps/pull/4752/commits/1ca53e57988f6eeada5242d7f3b8577a4c73eec4) was a little hairy – I tested by sending `**bold** *italic*` as a post, and then checked the chat list "latest post" preview to ensure there was no double spacing.

PR #4754, opened 2025-05-08
gallery: render all text on text posts

patosullivan:
fixes tlon-4124

if a list appeared as the second item in the `content` array it never rendered because we were slicing to grab just the first item from the array (which makes sense on image posts, we want the image and we don't want the caption, doesn't make sense for text-only posts).

Also updates the gallery post fixture so that we render a `postWithList` that has both an inline *and* a list after the inline, which is the case that was failing here.

### DIFF
--- a/packages/app/ui/components/GalleryPost/GalleryPost.tsx
+++ b/packages/app/ui/components/GalleryPost/GalleryPost.tsx
@@ -393,14 +393,25 @@ function LargePreview({
   ComponentProps<typeof PreviewFrame>,
   'content'
 >) {
-  const containsImage = useMemo(() => {
-    return content.some((block) => block.type === 'image');
+  const containsPreviewableContent = useMemo(() => {
+    return (
+      (content.some(
+        (block) =>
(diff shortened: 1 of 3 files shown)


packages/app/ui/components/GalleryPost/GalleryPost.tsx
```diff
@@ -393,10 +393,14 @@ function LargePreview({
...
 >) {
+  const containsImage = useMemo(() => {
+    return content.some((block) => block.type === 'image');
+  }, [content]);
+
   return (
     <PreviewFrame {...props} previewType={content[0]?.type ?? 'unsupported'}>
       <LargeContentRenderer
-        content={content.slice(0, 1)}
+        content={containsImage ? content.slice(0, 1) : content}
```
dnbrwstr, line 403: Is it only image blocks that this applies to? I think it'd be basic any case where there's a preview: refs, code blocks, etc.

dnbrwstr (approved):
Nice, this looks good to me!

PR #4755, opened 2025-05-08
InlineRenderer: styling code blocks

jamesacklin:
Fixes TLON-4184 by adding a context to `InlineRenderer` to track the parent style, then pass it to `InlineText` and conditionally render code blocks differently.

React Native styles don't automatically cascade to children, so recent changes to `InlineText` blew away the delicate balance of prop inheritance we had in place. 

We ended up with something to the effect of:

```
<CodeText> (monospace font)
  <RawText> (default font - overriding parent)
    "text content"
  </RawText>
</CodeText>
```

The RawText component was created with `unstyled: true`, making it ignore parent styling context altogether.

I had Copilot check to see if this plays nicely with #4752 and it seems to be fine; tagging @davidisaaclee in for good measure.

### DIFF
--- a/packages/app/ui/components/PostContent/InlineRenderer.tsx
+++ b/packages/app/ui/components/PostContent/InlineRenderer.tsx
@@ -15,10 +15,6 @@ import {
   TextInlineData,
 } from './contentUtils';
 
-const StyleContext = React.createContext<{
-  parentStyle?: 'code' | 'bold' | 'italic' | 'strikethrough';
-}>({});
-
 export const CodeText = styled(Text, {
   name: 'CodeText',
   size: '$mono/m',
@@ -86,13 +82,11 @@ export function InlineStyle({
   }[inline.style];
 
   return (
-    <StyleContext.Provider value={{ parentStyle: inline.style }}>
-      <StyleComponent {...props}>
-        {inline.children.map((child, i) => (
-          <InlineRenderer inline={child} key={i} />
-        ))}
-      </StyleComponent>
-    </StyleContext.Provider>
+    <StyleComponent {...props}>
+      {inline.children.map((child, i) => (
+        <InlineRenderer inline={child} key={i} />
+      ))}
+    </StyleComponent>
   );
 }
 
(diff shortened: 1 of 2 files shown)


dnbrwstr:
What was the change to `InlineText` that broke this? The intent of `RawText` is to always inherit parent style, so if that doesn't work it's likely to cause issues elsewhere.

jamesacklin:
I'm 90% sure it's https://github.com/tloncorp/tlon-apps/commit/76cf6f7a2b59cd8edde75acc541264c17aa517e9:

```diff
- return color ? <RawText color={color}>{inline.text}</RawText> : inline.text;
+ return <RawText color={color}>{inline.text}</RawText>;
```

This always wraps the text in `RawText` regardless of the presence of a parent wrapper. When the text was rendered directly, it would inherit styling, which doesn't happen with `RawText`.

Let me take another look.

jamesacklin:
@dnbrwstr always with the good questions. Take a look again, this fixes the issue in a more holistic manner.

davidisaaclee:
I was curious about where the `font-family: unset` was coming from – afaict it's here: https://github.com/tamagui/tamagui/blob/9c3a65a8cf9f182db856734bec249aab3857b9eb/code/core/web/src/views/Text.tsx#L46

It seems like Tamagui really doesn't want inherited styles (I didn't check how our other formatted text is working) – with that in mind, I think [dc622a1](https://github.com/tloncorp/tlon-apps/pull/4755/commits/dc622a1e08ad06a668ab67e24a45ae8520a5b9ba) seems like a reasonable approach (although I think `parentStyle` should allow for multiple styles, e.g. bold and italic)

jamesacklin:
I think I'm back to the drawing board here and just going to revert the one-line change in https://github.com/tloncorp/tlon-apps/commit/76cf6f7a2b59cd8edde75acc541264c17aa517e9. The main mentions processing logic happens far before render so that line is likely unrelated to line-wrapping. 

A quick regression test to this approach looks OK to me: 

image

dnbrwstr (approved):
Great, lgtm!

PR #4756, opened 2025-05-08
discipline: enforce mark sanity

Fang-:
Implements a wrapper agent style library that checks mark sanity on the inner agent.

This means the developer must give the library:
1. a list of marks and their types,
2. a list of subscription paths (path prefixes) and their marks,
3. and a list of scry paths (path prefixes) and their marks.

The library will then:
1. on reload, enforce that mark types haven't changed,
2. for each fact, check if the mark is expected, and type-check the data against it,
3. and for each scry, check if the mark is expected, and type-check the data against it.

Subscription and scry paths support empty string for wildcards. As written, they will simply use the first found matching path. Mismatches are printed to the terminal as priority debug prints and logged to posthog as criticals where possible.

Draft because I've only tested this on a dummy agent. I want to throw this onto channels agents and see how it behaves during the protocol upgrade scenario. Should already be good for initial developer review.

### DIFF
--- a/desk/lib/discipline.hoon
+++ b/desk/lib/discipline.hoon
@@ -43,7 +43,7 @@
 ::
 /%  m-noun  %noun
 ::
-|=  $:  marks=(list [=mark =type])
+|=  $:  marks=(list [=mark strict=? =type])
         facts=(list [=path marks=(list mark)])  ::  first matching is used, ~ for any
         scries=(list [=path =mark])
     ==
@@ -51,7 +51,7 @@
 ::
 =.  marks
   =-  (weld - marks)
-  :~  [%noun -:!>(*vale:m-noun)]
+  :~  [%noun & -:!>(*vale:m-noun)]
   ==
 =.  facts
   %+  weld
@@ -77,6 +77,11 @@
   (aor i.a i.b)
 =.  facts   (sort facts por)
 =.  scries  (sort scries por)
+::
+=/  mark-map=(map mark [strict=? =type])
+  (~(gas by *(map mark [? type])) marks)
+=/  mark-types=(map mark type)
+  (~(run by mark-map) tail)
 ::
 |=  inner=agent:gall
 |^  agent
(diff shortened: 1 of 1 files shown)


arthyn (commented):
lgtm so far

desk/lib/discipline.hoon
```diff
@@ -0,0 +1,307 @@
...
+      ::      ignore it, since we can't test it
+      ~
+    ?:  (~(nest ut u.type) | p.q.cage.p.card)
+      ~
+    =+  n=[on-path=path with-mark=p.cage.p.card %nest-fail]
+    ~&  >>>  [%discipline dap=dap.bowl %bad-fact-data n]
+    `(tell:log %crit ~['discipline: bad fact data' >n<] ~)
+  ::  +check-scry: validate result against known marks
+  ::
+  ++  check-scry
```
mikolajpp, line 172: Do we not intend to log mismatches in this arm as well?

desk/lib/discipline.hoon
```diff
@@ -0,0 +1,307 @@
...
+    ?.  ?=([[%discipline *] *] q.ole)
+      =.  last-marks  mark-types
+      =^  cards  inner  (on-load:og ole)  !:
+      =.  cards  (check-cards:help ~ cards)
+      [cards this]
+    =+  !<([%discipline old=state-0] (slot 2 ole))
+    =.  state  old
+    ?^  bad=(check-marks:help last-marks mark-map)
+      ~|  [%discipline dap=dap.bowl %mark-types-changed marks=;;((list mark) bad)]
+      !!  ::REVIEW  possibly annoying during development?
```
mikolajpp, line 247: It would be annoying in heavy works (such as refactors), and would prompt us to just disable lib-discipline until types stabilize. If this proves to be annoying in other cases, perhaps a "development" flag would be welcome, but for now I would just leave it as is.

mikolajpp (approved):

desk/lib/discipline.hoon
Fang-: Handling scries is read-only, so we cannot possibly log anything.

desk/lib/discipline.hoon
Fang-: The strictness flags on marks sort of already take care of this, so maybe can just remove the comment.

PR #4757, opened 2025-05-08
onboarding: longer timeouts + grab bag of fixes

latter-bolden:


arthyn (approved):
lgtm

latter-bolden:
This is deployed in the latest TestFlight build. Before merging, I'd like to confirm it's effective once the meld bandaid has been applied to the new signup ship pool on Hosting.

packages/shared/src/http-api/Urbit.ts
```diff
@@ -855,10 +857,13 @@ export class Urbit {
...
    */
   async scry<T = any>(params: Scry): Promise<T> {
-    const { app, path } = params;
+    const { app, path, timeout } = params;
     const response = await this.fetchFn(
       `${this.url}/~/scry/${app}${path}.json`,
-      this.fetchOptions
+      {
+        ...this.fetchOptions,
+        signal: timeout ? AbortSignal.timeout(timeout) : undefined,
```
davidisaaclee, line 865: @latter-bolden did you test this timeout on-device? I'm having issues with the client right now (root cause might be unrelated), with the following error being raised during sync: `TypeError: AbortSignal.timeout is not a function (it is undefined)`

Searching that, I see this issue that says that `AbortSignal.timeout` isn't defined in Hermes: https://github.com/facebook/react-native/issues/42042

PR #4758, opened 2025-05-09
native: clear push alerts and badge whenever app is foregrounded

davidisaaclee:
fixes TLON-2220

Removes all displayed notifications and sets badge count to 0[^0] whenever app is foregrounded.

If you receive a new message alert and then open the app, **that new message alert will be gone forever**, which I guess could be confusing if you have a custom sort order / lots of unreads. (Also the case for any other notification, like DM invites etc.) But the sticky notifications stink!

This would be better implemented via [TLON-4216](https://linear.app/tlon/issue/TLON-4216/dismiss-new-post-notifications-when-user-reads-post-on-another-client).

<details>

<summary>Note that this does not clear the in-memory Android conversation history:</summary>

1. User A sends messages 1, 2, 3 to a DM with user B
2. User B sees a conversation on their Android device with messages 1, 2, 3
3. User B opens client; conversation notification is cleared from system tray
4. User B closes client
5. User A sends message 4 to same DM
6. User B sees conversation with messages 1, 2, 3, 4

iiuc, we'll need to send an intent to our notification service to manually clear our in-memory conversation cache. There are a couple ways we could try doing this – I think the best way to do this would be to subclass expo-notification's [`NotificationService`](https://github.com/expo/expo/blob/28c218b351f34081ca15824cce57efe19276ac0c/packages/expo-notifications/android/src/main/java/expo/modules/notifications/service/NotificationsService.kt) to handle notifications instead of subclassing Google's Firebase service, and hook into [the existing "dismiss all" intent](https://github.com/expo/expo/blob/28c218b351f34081ca15824cce57efe19276ac0c/packages/expo-notifications/android/src/main/java/expo/modules/notifications/service/NotificationsService.kt#L627) (or more cleanly, [these PresentationDelegate methods](https://github.com/expo/expo/blob/28c218b351f34081ca15824cce57efe19276ac0c/packages/expo-notifications/android/src/main/java/expo/modules/notifications/service/interfaces/PresentationDelegate.kt#L9-L10)).

</details>

[^0]: We shouldn't be setting badge count at all, but my client has a permanently stuck badge count from our troubled past – surely I'm not the only one...

### DIFF
--- a/apps/tlon-mobile/src/hooks/useNotificationListener.ts
+++ b/apps/tlon-mobile/src/hooks/useNotificationListener.ts
@@ -1,6 +1,7 @@
 import crashlytics from '@react-native-firebase/crashlytics';
 import type { NavigationProp } from '@react-navigation/native';
 import { useNavigation } from '@react-navigation/native';
+import { useAppStatusChange } from '@tloncorp/app/hooks/useAppStatusChange';
 import { useFeatureFlag } from '@tloncorp/app/lib/featureFlags';
 import { connectNotifications } from '@tloncorp/app/lib/notifications';
 import { RootStackParamList } from '@tloncorp/app/navigation/types';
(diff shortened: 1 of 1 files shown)


dnbrwstr (approved):
Code lgtm. Personally, I'd prefer sticky-until-dismiss to clearing potentially unread notifications (since I basically only have them on for important convos), but will defer to y'all Android users.

arthyn:
yeah I wouldn't want my notifications to clear when opening the app. Is that something iOS does? We should just send dismisses when we hear that things are read which would solve this imo.

patosullivan:
> Code lgtm. Personally, I'd prefer sticky-until-dismiss to clearing potentially unread notifications (since I basically only have them on for important convos), but will defer to y'all Android users.

I agree with Dan on this one. I use notifications like an inbox.

davidisaaclee:
sounds good, closing!

PR #4760, opened 2025-05-09
notifications: add group title to message alert title

davidisaaclee:
fixes TLON-3008

Screenshot 2025-05-09 at 9 17 21 AM
Screenshot 2025-05-09 at 9 17 26 AM

I'm not sure I understand "group" vs "gang" – [this says a gang is a "view of foreign group"](https://github.com/tloncorp/tlon-apps/blob/d8c2ca2659f5ce688bd91354f793dc295de84748/desk/doc/dev/groups.udon#L27). So I'm inferring:
- When you are not a member of a group, you can only access a preview of that group via the "gang"
- There is no affordance on the API to indicate whether something should be accessed via gang or group (e.g. `group-invite` activity event references a `group` ID). (Related: groups and gangs use the same IDs.)
- It seems like once you join a group, you can no longer access the gang? If that's true, I don't understand it (and changing this would simplify the code, as we could rely on gangs instead of groups whenever looking for simple metadata)


packages/scripts/src/index.ts
```diff
@@ -161,7 +171,7 @@ export function renderActivityEventPreview({
         notification: {
           body: concat([
             lit('You were invited to '),
-            groupTitle(ev['group-invite'].group),
+            gangTitle(ev['group-invite'].group),
```
davidisaaclee, line 174: afaict, this is the only notification where the recipient is not a member of the group, so only using `gangTitle` here

dnbrwstr (approved):
Looks good to me!

PR #4761, opened 2025-05-09
xcode: update indentWidth/tabWidth to match code

davidisaaclee:
The Xcode project settings didn't match the default `swiftformat` settings, which made editing code in Xcode always mismatch indents against existing code. With this change, edits should automatically match the `swiftformat` defaults.


latter-bolden (approved):
Makes sense

PR #4771, opened 2025-05-12
mobile: revert DB changes & disable background sync

latter-bolden:
OTT

Let's see if this stabilizes things.

### DIFF
--- a/packages/app/lib/backgroundSync.ts
+++ b/packages/app/lib/backgroundSync.ts
@@ -1,101 +1,101 @@
-// import { createDevLogger, syncStart } from '@tloncorp/shared';
-// import { storage } from '@tloncorp/shared/db';
-// import * as db from '@tloncorp/shared/db';
-// import * as BackgroundFetch from 'expo-background-fetch';
-// import * as Notifications from 'expo-notifications';
-// import * as TaskManager from 'expo-task-manager';
+import { createDevLogger, syncStart } from '@tloncorp/shared';
+import { storage } from '@tloncorp/shared/db';
+import * as db from '@tloncorp/shared/db';
+import * as BackgroundFetch from 'expo-background-fetch';
+import * as Notifications from 'expo-notifications';
+import * as TaskManager from 'expo-task-manager';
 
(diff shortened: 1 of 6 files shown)


packages/app/lib/backgroundSync.ts
```diff
@@ -52,6 +52,9 @@ async function performSync() {
 }
 
 const TASK_ID = 'backgroundSync';
+export async function unregisterBackgroundSyncTask() {
+  await TaskManager.unregisterTaskAsync(TASK_ID);
```
latter-bolden, line 56: I'm a little confused about how the `TaskManager` works here in relation to the individual registrations on the `Notifications` and `BackgroundFetch` modules. 

It sounds like it _should_ work, but I also saw [this line in the docs](https://docs.expo.dev/versions/latest/sdk/task-manager/#taskmanagerunregistertaskasynctaskname) about module specific methods.

Do we wanna run `unregisterTaskAsync` on those?

davidisaaclee: yes! doing now

davidisaaclee:
putting some notes here:

We're reverting these changes because we're seeing error logs that look like someone's client failed to migrate: https://eu.posthog.com/project/7233/persons/e5b20d75-5d6a-5736-bd8f-3e6d3e0554a6

- user is on iOS
- I'd be surprised if background sync sends logs to PostHog – we don't do any PostHog setup in the task body. So these logs would be coming from the main app JS context, which should always be doing migrations before running DB code.
- It's possible that opening a concurrent connection to the DB in the background task (i.e. while main app still has a connection) could be causing the `database disk image is malformed` error we've been seeing

I see an error case with background sync where the background sync is using a new DB schema before the DB is ready (which would not look like the PostHog log):
1. App is installed and has background sync task on user's phone
2. App is updated with new DB schema
3. Background task is triggered – this causes phone to boot up headless app and run the JS task, which attempts to reference the DB using its new schema, which fails
I'm not sure that this would be much of an issue, because the app should just migrate on next launch, and then fill in the data that couldn't be inserted (assuming we don't mark the stuff we background-synced as "don't need this anymore", which we do not yet do).

PR #4772, opened 2025-05-12
native: revert double migration gate possibly causing skipped migration

davidisaaclee:
We're seeing error logs that look like migrations were not run on a client. Reverting this change in case it caused an incorrect migration skip. https://eu.posthog.com/project/7233/persons/e5b20d75-5d6a-5736-bd8f-3e6d3e0554a6


davidisaaclee:
#4771

PR #4773, opened 2025-05-13
desktop: clean up console spam

patosullivan:
Cleans up the console by:

- no longer throwing bad fetch errors for embeds/avoiding attempts to fetch twitter emebeds on web
- fixing up Reference.tsx (we were passing context values as props in many places, or spreading props where we shouldn't) to get rid of a few errors about props being passed to a DOM element that shouldn't have been  (`actionIcon` and `contentSize`) .
- remove `isNotice` from the ContentContext initialization, removes the error log about `isNotice` being passed to a DOM element.
- Updates tamagui to the latest version to get rid of the forwardRef/bad setState() spam (https://github.com/tamagui/tamagui/issues/1828). Updating to the latest tamagui caused some uses of Text to appear too small (specifically, the section headers on the GroupChannelsScreenView, which should have been using our existing component anyway, and the "Starting Up..." text on desktop).

To get rid of the multiple messages from sqlocal about transactions we'll need to update sqlocal. Do we wanna do that now or wait? I feel like that message is pretty innocuous since it's short (and not an error).

fixes tlon-4228

### DIFF
--- a/packages/app/ui/components/ContentReference/Reference.tsx
+++ b/packages/app/ui/components/ContentReference/Reference.tsx
@@ -65,8 +65,10 @@ const ReferenceComponent = ReferenceFrame.styleable<{
   hasData?: boolean;
 }>(
   ({ children, isLoading, isError, hasData, errorMessage, ...props }, ref) => {
+    // We cannot pass `contentSize` to the styled component, so we need to extract it
+    const { contentSize, ...rest } = props;
     return (
-      <ReferenceFrame {...props} pressable={!!props.onPress} ref={ref}>
+      <ReferenceFrame {...rest} pressable={!!props.onPress} ref={ref}>
         {children}
         {isLoading ? (
           <ReferenceSkeleton
@@ -183,8 +185,9 @@ export function ReferenceSkeleton({
   message?: string;
   messageType?: 'loading' | 'error' | 'not-found';
 } & ComponentProps<typeof ReferenceFrame>) {
+  const { contentSize, ...rest } = props;
   return (
-    <ReferenceFrame {...props}>
+    <ReferenceFrame {...rest}>
       <ReferenceBody padding="$l" justifyContent="center" alignItems="center">
         <YStack gap="$l" alignItems="center">
           {messageType === 'error' ? (
@@ -195,7 +198,7 @@ export function ReferenceSkeleton({
             />
           ) : null}
           <Text
-            size={props.contentSize === '$s' ? '$label/s' : '$label/m'}
+            size={contentSize === '$s' ? '$label/s' : '$label/m'}
             color="$tertiaryText"
             flex={1}
           >
--- a/packages/app/ui/components/ContentReference/Reference.tsx
+++ b/packages/app/ui/components/ContentReference/Reference.tsx
@@ -36,7 +36,6 @@ export const useReferenceContext = () => {
 };
 
 const ReferenceFrame = styled(YStack, {
-  context: ReferenceContext,
   name: 'ReferenceFrame',
   borderRadius: '$s',
   padding: 0,
@@ -99,7 +98,6 @@ const ReferenceComponent = ReferenceFrame.styleable<{
 );
 
 const ReferenceHeader = styled(XStack, {
-  context: ReferenceContext,
   name: 'ReferenceHeader',
   paddingLeft: '$l',
   paddingRight: '$l',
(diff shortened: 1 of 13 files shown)


patosullivan:
There's an issue with sheets/popovers in the latest tamagui, looking into it.

patosullivan:
Ok, I fixed the type issues, another issue of a prop getting passed down to a DOM element, and I fixed the issue with popovers.

The popover issue was happening because now SheetScrollView expects to be rendered within a sheet and to have access to the sheet context, which would normally include `setHasScrollView`. But, we built our popovers around the idea that we could render the same thing inside of them that we do our sheets (without actually being wrapped in a Sheet), and several of them render content that ultimately uses SheetScrollView under the hood. For now, the fix I'm proposing is that we just add some checks to the SheetScrollView where `setHasScrollView` gets called.

patosullivan:
Seeing some more small text in components that use `<Button.Text/>`

patosullivan:
Something is wrong with font sizes generally.

patosullivan:
The issue with font sizes:

Tamagui is not generating the css variables for our font sizes (`--f-size-[the size]`), instances of tamagui `<Text />` are instead using our space token measures.

I've narrowed this down to some change made with the release of v1.115.0: https://github.com/tamagui/tamagui/tree/v1.115.0

Nothing jumps out at me in there, but the release immediately prior to that one doesn't have this issue.

packages/app/ui/components/GroupChannelsScreenView.tsx
```diff
@@ -188,15 +187,11 @@ export const GroupChannelsScreenView = React.memo(
...
-            <Text
-              paddingHorizontal="$l"
-              paddingVertical="$xl"
-              fontSize="$s"
-              color={listSectionTitleColor}
-              onLayout={handleHeaderLayout}
-            >
-              {item.title}
-            </Text>
+            <SectionListHeader onLayout={handleHeaderLayout}>
```
patosullivan, line 190: I changed this because I noticed we were using this component in ChatList but not here, and afaict we should be using it in both places.

packages/app/ui/tamagui.config.ts
```diff
@@ -543,6 +543,7 @@ export const config =
         themes,
         media,
         settings: {
+          defaultFont: 'body',
```
patosullivan, line 546: Including this, for some reason, ensures that the theme actually uses `body` as the defaultFont. We shouldn't *need* to include it, since the default for this is `body` as-is, but we weren't getting our font variables without it.

patosullivan:
Ok, this should be good to go. There might be some straggler error messages here and there (one I know of is related to images on web, but I think that will go away with the changes to web images on `release-channels-updates`), but the bulk of the extraneous errors have been removed.

I also went ahead and removed tailwind while I was in here because a) we don't use it anymore b) it just causes confusion/distraction when we need to dig into any issues with CSS.

patosullivan:
> Nice! Overall looks good to me, and glad to see tailwind cleaned up. Two pieces that concern me are:
> 
> * Removing spread props from `GalleryPost` means that the props type is no longer accurate, and also that it's a little less flexible to use
> * The contexts on reference and actionsheet components are there because those components have variants that change based on properties in that context. I don't see anything broken immediately, but it seems fairly likely something is, and if not we should remove those variants. In general, being able to use context like that is a useful pattern, so I'd almost rather err on the side of suppressing the extra props error if that's possible.

both of those issues should be fixed. I keep finding little things as I work on this, but I think we're at a good stopping point now.

packages/app/ui/components/ContentReference/Reference.tsx
```diff
@@ -99,7 +100,6 @@ const ReferenceComponent = ReferenceFrame.styleable<{
 );
 
 const ReferenceHeader = styled(XStack, {
-  context: ReferenceContext,
```
dnbrwstr, line 102: Should re-add `ReferenceContext`

packages/app/ui/components/Channel/ChannelHeader.tsx
```diff
@@ -158,7 +159,9 @@ export function ChannelHeader({
             {showSearchButton && (
               <ScreenHeader.IconButton type="Search" onPress={goToSearch} />
             )}
-            {contextItems}
+            {contextItems.map((item, index) => (
+              <Fragment key={index}>{item}</Fragment>
```
dnbrwstr, line 163: Curious why this is necessary -- maybe add a comment?

dnbrwstr (approved):
Lgtm! Only blocking thing is `ReferenceContext`

dnbrwstr (changes requested):
Oops meant to request changes.

PR #4775, opened 2025-05-13
mobile: add telemetry for native DB lifecycle

latter-bolden:
Adds debug logs and error capturing for mobile DB setup, migration, and purge.

We can get more granular once we re-introduce the `await`'able executions.


latter-bolden:
we may want to remove the non-tracked logs as well? As is, it's feeling pretty verbose.

davidisaaclee:
if we want to keep these around, can we move the error logs up a level so we log errors for all platforms (e.g. `webDb`)?

packages/app/lib/nativeDb.ts
```diff
@@ -65,43 +80,98 @@ export class NativeDb extends BaseDb {
...
 
-    // reset values related to tracking db sync state
-    await kv.headsSyncedAt.resetValue();
+      // reset values related to tracking db sync state
+      await kv.headsSyncedAt.resetValue();
 
-    logger.log('purged sqlite database, recreating');
-    await this.setupDb();
+      logger.trackEvent(AnalyticsEvent.NativeDbDebug, {
+        message: 'purbeDb: completed purge, recreating',
```
davidisaaclee, line 107:
```suggestion
        message: 'purgeDb: completed purge, recreating',
```

packages/app/lib/nativeDb.ts
```diff
@@ -65,43 +80,98 @@ export class NativeDb extends BaseDb {
...
+      });
+      await this.setupDb();
+      logger.trackEvent(AnalyticsEvent.NativeDbDebug, {
+        message: 'purbeDb: post-purge setup complete',
```
davidisaaclee, line 111:
```suggestion
        message: 'purgeDb: post-purge setup complete',
```

davidisaaclee (approved):
if you don't want to move the logs up, this otherwise lgtm

PR #4776, opened 2025-05-13
android: list all explicit dependents of JS copy task

davidisaaclee:


davidisaaclee:
It looks like this successfully passed Android test build, but failed iOS e2e test: https://github.com/tloncorp/tlon-apps/actions/runs/15002619017 – hoping it's just flake, re-running

PR #4778, opened 2025-05-13
native: use default journal_mode=DELETE instead of MEMORY

davidisaaclee:
fixes TLON-4232

`MEMORY` has a chance to corrupt the database file if the process crashes during a transaction. Not sure if `DELETE` is the best choice, but it's the default behavior and it appears to work:

TLON-4232's cause is unclear, but it can be triggered somewhat reliably by crashing the background fetch process during a transaction (triggering ~1 out of 3 times). Using `journal_mode=DELETE` and running through the repro 6 times did not corrupt the DB once.

(If this is the correct cause of the `database disk image is malformed` message, then crashes of host app during transactions would cause this as well, which could explain the corrupted databases reported before background fetch was introduced.)

This PR does not re-introduce background fetch – will do that separately.


latter-bolden:
It seems like this will be strictly safer so I'm not concerned about data integrity. If the main drawback is performance, I'm wondering how best to test for meaningful degradation?

dnbrwstr:
I think we just play with it -- that flag really shouldn't be necessary, and in my original testing it made a bit of a difference, but not huge.

PR #4779, opened 2025-05-14
ChatOptionsSheet: disable channel leave for hosts

jamesacklin:
Fixes TLON-4183 by adding a simple check to see if the user is the channel host; if not, then we display a message that they have to delete the channel and can't leave it on their own accord.

This also adds a pretty robust fixture for the ChatOptionsSheet, which:

- Splits the demonstration out into three cases: individual channel, group, and DM
- Adds options for unread counts to show and hide the "Mark all as read" option
- Adds options for group privacy to toggle invite availability for members
- Adds an option for host/member status to test TLON-4183 ;)

rec


dnbrwstr (approved):
Lgtm to me! Love the fixture.

PR #4789, opened 2025-05-16
lanyard: %revoke-all poke

Fang-:
For usage during manual testing and for hosting to do clean-up while recycling ships.


arthyn (approved):
lgtm

PR #4791, opened 2025-05-16
content-helpers: make sure we extract the text for code blocks

arthyn:
OTT reported by in support channel


latter-bolden (changes requested):
Change makes sense. 

Can we add a unit test so we'd catch this in the future?

latter-bolden (approved):
Looking good!

PR #4792, opened 2025-05-16
dumb-proxy: stupid http proxy

Fang-:
Serves an api endpoint at `/apps/groups/~/proxy`, to which you can make
requests at `/[@uw-encoded-url-string]`, which will trigger this agent
to pass the request on to that target url mostly unchanged (only adding
in a Forwarded header), and eventually respond with whatever response it
got back from that target.

Response has an `x-dumb-proxy` header that's either `finished` for successful requests, or `cancelled` if the request got dropped.

Hopefully useful for circumventing CORS shenanigans, among other things. Probably shouldn't merge until this has found a client-side use-case. cc client devs.


desk/app/dumb-proxy.hoon
```diff
@@ -0,0 +1,192 @@
+::  dumb-proxy: stupid stateless http proxy
+::
+::    serves an api endpoint at /apps/groups/~/metagrab,
```
arthyn, line 3:
```suggestion
::    serves an api endpoint at /apps/groups/~/proxy,
```

desk/app/dumb-proxy.hoon
```diff
@@ -0,0 +1,192 @@
...
+        [@ ~]
+      =|  msg=@t
+      =*  bad-req
+        %-  (tell:l %warn msg url.request ~)
+        [(spout:hutils id [400 ~] `(as-octs:mimes:html msg)) this]
+      ?~  target=(slaw %uw i.site)
+        =.  msg  'target not @uw'
+        bad-req
+      ?~  (de-purl:html u.target)
+        ::TODO  if parser fails, just dumb find <title> and extract
```
arthyn, line 107:
```suggestion
```

arthyn (approved):
lgtm

Fang-:
Before we merge this: actually didn't check whether this receives/passes on the auth cookie. I don't think eyre strips them when passing the request in, so, they may be there and subsequently get sent on to the target...

PR #4793, opened 2025-05-16
ops: bump nockjs version

latter-bolden:
OTT

We published the new version with the changes that required us to use the git tag.


patosullivan (approved):
Nice

PR #4798, opened 2025-05-19
theme: simpler theme setting + recall

jamesacklin:


apps/tlon-web/src/app.tsx
```diff
@@ -429,6 +414,8 @@ function ConnectedWebApp() {
       if (!hasSyncedRef.current) {
         // Web doesn't persist database, so headsSyncedAt is misleading
         await db.headsSyncedAt.resetValue();
+        // Explicitly sync settings early in the initialization process
+        await syncSettings();
```
dnbrwstr, line 418: We should add this to `syncStart` -- I also tend to think it's not really so high priority we need to sync it before everything else on first boot? If your theme change takes a second to appear, maybe that's ok?

packages/app/provider/index.tsx
```diff
@@ -1,122 +1,155 @@
...
+
+  const themeIsDark = useMemo(() => {
+    const displayTheme =
+      activeTheme === 'auto' ? (systemIsDark ? 'dark' : 'light') : activeTheme;
+    return DARK_THEME_NAMES.includes(displayTheme);
+  }, [activeTheme, systemIsDark]);
+
+  const { data: storedTheme, isLoading, refetch } = useThemeSettings();
+
+  const setActiveTheme = useCallback(
```
dnbrwstr, line 52: Seems like some or all of this should be an action?

packages/app/provider/index.tsx
```diff
@@ -1,122 +1,155 @@
...
+      } catch (error) {
+        console.warn('Failed to save theme preference:', error);
+      }
+    },
+    [refetch]
   );
 
-  const { data: storedTheme, isLoading } = useThemeSettings();
+  const displayTheme = useMemo(() => {
+    return activeTheme === 'auto'
```
dnbrwstr, line 72: This logic crops up in a few places -- might make sense to pull out.

dnbrwstr (changes requested):
Looking good, added a few nits.

PR #4799, opened 2025-05-20
ops: pull request template

jamesacklin:
Adds a pull request template for new PRs.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -4,7 +4,7 @@ Briefly describe what this pull request does in 1-3 sentences.
 
 ## Changes
 
-Outline specific changes made in this PR.
+<!-- Outline specific changes made in this PR. -->
 
 ## Testing
 
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -1,0 +1,33 @@
+## Summary
+
+Briefly describe what this pull request does in 1-3 sentences.
+
+## Changes
+
+Outline specific changes made in this PR.
+
+## Testing
+
+### Automated tests
+
+- [ ] Unit tests added/updated
+- [ ] Integration tests added/updated
+- [ ] E2E tests added/updated
+
+### Manual testing steps
+
+1.
+2.
+3.
+
+## Risks and impact
+
+- Risk level: low | medium | high
+- Areas of concern:
+- Dependencies:
+
+## Rollback plan
+
+Describe the steps to revert these changes if needed.
+
+## Screenshots / videos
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
+## Summary
+
+Briefly describe what this pull request does in 1-3 sentences.
```
davidisaaclee, line 3: We could put the template instructions in an HTML comment so that you don't need to manually delete them each time:
```suggestion
<!-- Briefly describe what this pull request does in 1-3 sentences. -->
```

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+
+Outline specific changes made in this PR.
+
+## Testing
+
+### Automated tests
+
+- [ ] Unit tests added/updated
+- [ ] Integration tests added/updated
+- [ ] E2E tests added/updated
```
davidisaaclee, line 15: I feel like these checkboxes will be easy to see in the code diff (so maybe we don't need them in the template)

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Manual testing steps
+
+1.
+2.
+3.
+
+## Risks and impact
+
+- Risk level: low | medium | high
```
davidisaaclee, line 25: I'm not sure I'd know what to put for these three fields. 

I'd like to either make the risk levels explicit, or just leave this section freeform (i.e. just a header and a blank space to fill in) to reduce noise. If explicit, maybe:
- Safe to rollback without consulting PR author?
- Affects important code area [^0]:
  - [ ] login
  - [ ] channel load
  - [ ] message sends

[^0]: maybe this is better as a tag, or we can gh action them into tags (thinking if we're trying to quickly figure out what caused a breakage, tags would be better to filter with)

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
davidisaaclee: (I think we just need a heading `## How did I test?`)

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,33 @@
+## Summary
+
+Briefly describe what this pull request does in 1-3 sentences.
+
+## Changes
+
+Outline specific changes made in this PR.
```
davidisaaclee, line 7:
```suggestion
<!-- Outline specific changes made in this PR. -->
```

davidisaaclee (commented):
My understanding is that the PR description is targeted towards (1) easier code review, and (2) reviewing what changed if we're trying to address an issue. With that in mind, I think the template can be smaller to reduce noise / work, so most of my comments are about cutting stuff.

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
jamesacklin: Cool! Happy to leverage PR labels, will set them up and then convert this section to an implicit requirement. Let's get everyone comfortable with this before we start tagging.

PR #4801, opened 2025-05-20
display error message when posts fail to load

dnbrwstr:
## Summary

Displays an error message with retry button when posts failed to load. 

## Changes

- Pass `postQuery` errors and refetch function down into `EmptyChannelNotice`
- Rework loading so `EmptyChannelNotice` can be displayed even if there are no posts loaded
- Modify the caching of `useChannelPosts` load status to prevent flickering erroneous error messages during initial retries.

## How did I test?

Tested following scenarios on mobile and desktop:
- Channel with and without unreads
- Channel with and without posts cached in DB
- With and without artificial errors inserted in API load method and `usePostsQuery`

## Risks and impact

Affects channel display / load sequence logic.

- Safe to rollback without consulting PR author? yes
  - Channel display

## Screenshots / videos

<video src="https://github.com/user-attachments/assets/3833168b-d201-424a-917f-9346bff0f97e" width="300" />


patosullivan (changes requested):
Changes make sense, but I'm seeing this flash of red each time I load a channel on web: https://i.gyazo.com/50a79aff0c7fd2d1a7084233c1c1aab2.mp4

patosullivan (approved):
Was missing the latest commit during my last review. New commit fixed the issue I was seeing!

PR #4802, opened 2025-05-20
tiptap: dont change from code block -> inline code

arthyn:
Fixes TLON-4085. Found an issue when trying to save a notebook with code blocks. Seems we were trying to convert a code block with no language into inline code, which just seems bad from a UX perspective, but was also making it fail when sent to the backend. Also threw in `clearAttachements` in the notebook send because saving the post would make that image show up in the reply box after you get navigated.


latter-bolden (approved):
Change makes sense. Working for me on mobile and web

arthyn:
I was wrong about the issue here. The issue was that in the new world our type discriminator functions weren't discriminate enough. I've restored the original behavior for now and updated the tests.

PR #4803, opened 2025-05-20
navigation: add param extractor functions to navigation utils, move param extraction out of components

patosullivan:
This gives us just one place to check for navigation param related logic, which is pretty tricky in our app since we make heavy use of nested navigators.

I'm pen to other ideas on how to get this logic out of our components and into one central location, btw.

also fixes tlon-4273

### DIFF
--- a/packages/app/navigation/utils.ts
+++ b/packages/app/navigation/utils.ts
@@ -475,3 +475,151 @@ export function screenNameFromChannelId(channelId: string) {
       ? 'GroupDM'
       : 'Channel';
 }
+
+/**
+ * Safely extracts a channel ID from navigation route parameters
+ * handling various parameter structures in the navigation system.
+ * 
+ * @param route Any navigation route object
+ * @returns The channel ID if found, undefined otherwise
+ */
+export function getChannelId(route: any): string | undefined {
+  return getParamValue<string>(route, 'channelId');
+}
+
+/**
+ * Safely extracts a group ID from navigation route parameters
+ * handling various parameter structures in the navigation system.
+ * 
+ * @param route Any navigation route object
+ * @returns The group ID if found, undefined otherwise
+ */
+export function getGroupId(route: any): string | undefined {
+  return getParamValue<string>(route, 'groupId');
+}
+
+/**
+ * Safely extracts a user ID from navigation route parameters
+ * handling various parameter structures in the navigation system.
+ * 
+ * @param route Any navigation route object
+ * @returns The user ID if found, undefined otherwise
+ */
+export function getUserId(route: any): string | undefined {
+  return getParamValue<string>(route, 'userId');
+}
+
+/**
+ * Safely extracts chat details (chatId and chatType) from route parameters
+ * 
+ * @param route Any navigation route object
+ * @returns Object with chatId and chatType if found, undefined otherwise
+ */
+export function getChatDetails(route: any): { chatId: string; chatType: 'group' | 'channel' } | undefined {
+  const chatId = getParamValue<string>(route, 'chatId');
+  const chatType = getParamValue<'group' | 'channel'>(route, 'chatType');
+  
+  if (chatId && chatType) {
+    return { chatId, chatType };
+  }
+  
+  return undefined;
+}
+
+/**
+ * Safely extracts a post ID from navigation route parameters
+ * 
+ * @param route Any navigation route object
+ * @returns The post ID if found, undefined otherwise
+ */
+export function getPostId(route: any): string | undefined {
+  return getParamValue<string>(route, 'postId');
+}
+
+/**
+ * Safely extracts a selected post ID from navigation route parameters
+ * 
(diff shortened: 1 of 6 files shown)


dnbrwstr:
Spent a little time working through what's happening in navigation, and have a few thoughts.
- As far as I can tell, we should never get nested params in actual screens, only in contexts like the the body of the `ChannelStack` where we're outside a component. Is that the case? If it is, I think we should leave the channel screens as-is.
- Can we improve the type safety of the helper methods? It would be nice for instance if we got a type error if we make a type when specifying the param name in `getParamValue`. 
- It seems like there's lots of duplication between `HomeStack` and `MessageStack` -- would be good to collapse that where we can.

davidisaaclee:
I'd love to use this code – I can start a review now but pinging @dnbrwstr to respond to the above

packages/app/navigation/utils.ts
```diff
@@ -475,3 +475,216 @@ export function screenNameFromChannelId(channelId: string) {
...
+  if (
+    route.params?.screen &&
+    route.params.params &&
+    typeof route.params.params[paramName] !== 'undefined'
+  ) {
+    return route.params.params[paramName] as T;
+  }
+
+  // Deep nesting common in desktop navigators
+  if (
```
davidisaaclee, line 681: when would we use this? I'd think that either (1) we should only get params from the _outer_ screen (like inheritance – what if we start nesting a preview of a channel to jump to when clicking on a ref?), or (2) we are not passing the correct route to the param getter.

davidisaaclee (commented):
Here's a little diff that would guard against passing a route that is missing `params.chatId` into `getChatId()` (I didn't change all of the getters).

```diff
diff --git a/packages/app/navigation/utils.ts b/packages/app/navigation/utils.ts
index 61cb927f6..a5dc24e8f 100644
--- a/packages/app/navigation/utils.ts
+++ b/packages/app/navigation/utils.ts
@@ -1,6 +1,7 @@
 import {
   CommonActions,
   NavigationProp,
+  RouteProp,
   useNavigation as useReactNavigation,
 } from '@react-navigation/native';
 import { createDevLogger } from '@tloncorp/shared';
@@ -548,13 +549,17 @@ export function getFromBlankChannel(route: any): boolean {
   return getParamValue<boolean>(route, 'fromBlankChannel') ?? false;
 }
 
+type RouteLike<Params extends Record<string, unknown>> = {
+  params: Params;
+};
+
 /**
  * Safely extracts a chat ID from navigation route parameters
  *
  * @param route Any navigation route object
  * @returns The chat ID if found, empty string otherwise
  */
-export function getChatId(route: any): string {
+export function getChatId(route: RouteLike<{ chatId: string }>): string {
   return getParamValue<string>(route, 'chatId') ?? '';
 }
 
@@ -564,7 +569,9 @@ export function getChatId(route: any): string {
  * @param route Any navigation route object
  * @returns The chat type if found, empty string otherwise
  */
-export function getChatType(route: any): 'group' | 'channel' {
+export function getChatType(
+  route: RouteLike<{ chatType: 'group' | 'channel' }>
+): 'group' | 'channel' {
   return getParamValue<'group' | 'channel'>(route, 'chatType') ?? 'channel';
 }
 
@@ -574,7 +581,9 @@ export function getChatType(route: any): 'group' | 'channel' {
  * @param route Any navigation route object
  * @returns The preview group ID if found, undefined otherwise
  */
-export function getPreviewGroupId(route: any): string | undefined {
+export function getPreviewGroupId(
+  route: RouteLike<{ previewGroupId: string }>
+): string | undefined {
   return getParamValue<string>(route, 'previewGroupId');
 }
 
@@ -647,9 +656,13 @@ export function getRouteName(route: any): string {
  * @param defaultValue Optional default value to return if parameter is not found
  * @returns The parameter value if found, defaultValue or undefined otherwise
  */
-export function getParamValue<T>(
-  route: any,
-  paramName: string,
+export function getParamValue<
+  ParamName extends string,
+  T,
+  Params extends Record<ParamName, T>,
+>(
+  route: RouteLike<Params>,
+  paramName: ParamName,
   defaultValue?: T
 ): T | undefined {
   // Handle null or undefined route
```

packages/app/navigation/utils.ts
davidisaaclee: I'm guessing that this is relevant to the comment about

> we should never get nested params in actual screens, only in contexts like the the body of the ChannelStack where we're outside a component.

Can we explicitly pass the nested route to the getter in the `ChannelStack` case? react-navigation has this helper - maybe not usable directly but we could do something very similar if we're just trying to get the focused route: https://github.com/react-navigation/react-navigation/blob/main/packages/core/src/findFocusedRoute.tsx (but i'm still not clear what exactly we're trying to "look at" when we use that getter)

latter-bolden:
Sounds like the approach needs to be discussed further before merging. May be impacted by #4922.

Going to close in the meantime since stale.

PR #4807, opened 2025-05-21
remove bauble header and channel switcher

dnbrwstr:
## Summary

Remove unused bauble header + channel switcher sheet.

## How did I test?

Tested channel and channel fixture on mobile + web

## Risks and impact


patosullivan (approved):
Looks good, makes sense, works well locally.

PR #4808, opened 2025-05-21
track ordered posts from backend; add API binding to set ordered posts

davidisaaclee:
part of TLON-4215

Adds plumbing to support pinned posts using existing "ordered posts" API feature.

## Changes
- Sync changes to a channel's `order` from backend to client DB (under `postOrder` – `order` is reserved by Drizzle for the SQL feature).[^0]
- Add `setOrder` API method to directly set a channel's `order` on backend
- [extra] [Disable usePostWithRelations by passing null options](https://github.com/tloncorp/tlon-apps/commit/886417ee7494557ed750f5aa2cb799463b0420c4) – I added this when adding UI code to test pins; slipping it in here because I think it's a good change (i.e. if a component doesn't yet have a post ID, it can now pass `null` instead of an empty string to the hook until it gets an ID)

## How did I test?

[`dil/pinned-posts`](../tree/dil/pinned-posts)[^1] has changes to UI to allow pinning a single post and showing it in the chat options sheet. I manually ran through that flow (pinning post in a channel; going to a channel without a pinned post; opening channel options in both).

## Risks and impact

- [x] Safe to rollback without consulting PR author
  - [ ] Onboarding
  - [x] ~State~ Database / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Screenshots / video
Recorded using the UI changes from [`dil/pinned-posts`](../tree/dil/pinned-posts):
https://github.com/user-attachments/assets/7c282e89-bf08-45da-9f65-99d77383c428

[^0]: This is a simple JSON array – for a true pinned posts feature, we should probably have a separate merge table; but a static array seems okay while the API can only directly set `order` (i.e. no append / remove specific item, so to make a mutation, we need to either know the previous pinned set or replace all items)
[^1]: Importing `PostView` in `ChatOptionsSheet` opened a can of circular dependencies – that branch will randomly crash when navigating to different screens, complaining that a React component is undefined. More info in [this ticket](https://linear.app/tlon/issue/TLON-4269/client-js-code-has-circular-imports-causing-react-breakage-when-adding).

### DIFF
--- a/packages/shared/src/store/dbHooks.ts
+++ b/packages/shared/src/store/dbHooks.ts
@@ -504,14 +504,15 @@ export const usePostWithThreadUnreads = (options: { id: string }) => {
 };
 
 export const usePostWithRelations = (
-  options: { id: string },
+  options: { id: string } | null,
   initialData?: db.Post
 ) => {
   return useQuery({
-    queryKey: ['post', options.id],
+    enabled: options != null,
+    queryKey: ['post', options?.id],
     staleTime: Infinity,
     ...(initialData ? { initialData } : {}),
-    queryFn: () => db.getPostWithRelations(options),
+    queryFn: () => (options == null ? null : db.getPostWithRelations(options)),
   });
 };
 
--- a/packages/shared/src/api/channelsApi.ts
+++ b/packages/shared/src/api/channelsApi.ts
@@ -168,6 +168,7 @@ export function toClientChannelsInit(
 
 export type ChannelInit = {
   channelId: string;
+  order: string[];
   writers: string[];
   readers: string[];
 };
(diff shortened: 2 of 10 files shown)


packages/shared/src/store/dbHooks.ts
```diff
@@ -504,14 +504,14 @@ export const usePostWithThreadUnreads = (options: { id: string }) => {
...
-  options: { id: string },
+  options: { id: string } | null,
   initialData?: db.Post
 ) => {
   return useQuery({
-    queryKey: ['post', options.id],
+    queryKey: ['post', options?.id],
     staleTime: Infinity,
     ...(initialData ? { initialData } : {}),
-    queryFn: () => db.getPostWithRelations(options),
```
dnbrwstr, line 514: Might improve the semantics of the returned query to also set `enabled` based on `options` being null or not.

dnbrwstr (approved):
Looks good / makes sense to me! Just one nit.

packages/shared/src/store/dbHooks.ts
davidisaaclee: fixed and rebased – [0fd77bf](https://github.com/tloncorp/tlon-apps/pull/4808/commits/0fd77bfbfe51d328b5e2ff798e3207ec516c8571)

PR #4809, opened 2025-05-22
protocol change: patch the frontend to handle rendering either unicode emojis or shortcodes

patosullivan:
The new backend sends either unicode emojis (in subscription updates) or shortcodes (in the version of the scry we currently use for outlines).

For now, rather than refactor the way the frontend deals with emojis entirely we can just patch the way we render them in SizableEmoji.

If `emojiInput` can't be found in our map of shortcodes -> emojis and if it matches a unicode emoji pattern, we just render it directly.

fixes tlon-4279.

Later on we should probably move away from shortcodes on the frontend.


arthyn (approved):
lgtm

latter-bolden (approved):
No version looks good! I see no issues running locally

PR #4810, opened 2025-05-22
images: prevent crash from bad image source

latter-bolden:
After switching to `expo-image` on web, we're seeing a crash if the passed image `source` is a hex code instead of a URL. There are some legacy agent states that have image meta set to colors.

To prevent this, the image component validates the source string is a URL before attempting to render. I also added an error boundary that uses the existing fallbacks in case we encounter other errors — a crashing image component should never take down the whole app like it did in this case.

Fixes TLON-4287

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -60,6 +60,7 @@
     "browser-or-node": "^3.0.0",
     "exponential-backoff": "^3.1.1",
     "sorted-btree": "^1.8.1",
+    "validator": "^13.7.0",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


packages/shared/package.json
```diff
@@ -60,6 +60,7 @@
     "browser-or-node": "^3.0.0",
     "exponential-backoff": "^3.1.1",
     "sorted-btree": "^1.8.1",
+    "validator": "^13.7.0",
```
patosullivan, line 63: this will come in handy

patosullivan (approved):
Looks good, changes make sense. Glad to have an ErrorBoundary!

PR #4812, opened 2025-05-22
link-input: allow links with no preview

arthyn:
OTT fixes TLON-4294


patosullivan (approved):
Makes sense, tested locally, works.

PR #4813, opened 2025-05-22
links: fix form deletion + block send while loading

latter-bolden:
OTT

- form deletion was just a logic bug with it re-inserting the default title if it detected the input was empty
- chat was not attempting to block send while loading, now it does via a state variable
- gallery was attempting to block, but it wasn't effective until the debounce triggered

Fixes TLON-4276
Fixes TLON-4297

### DIFF
--- a/packages/app/ui/components/draftInputs/LinkInput.tsx
+++ b/packages/app/ui/components/draftInputs/LinkInput.tsx
@@ -88,6 +88,16 @@ export function LinkInput({ editingPost, isPosting, onSave }: LinkInputProps) {
 
   const form = watch();
   const url = useDebouncedValue(form.url, 500);
+  // We need to track debounce dirty state to appropriately disable the send button. useLinkGrabber's
+  // loading state will not be set until the debounce fires.
+  const [isPendingDebounce, setIsPendingDebounce] = useState(false);
+  useEffect(() => {
+    if (form.url !== url) {
+      setIsPendingDebounce(true);
+    } else {
+      setIsPendingDebounce(false);
+    }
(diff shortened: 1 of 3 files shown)


packages/app/ui/components/draftInputs/LinkInput.tsx
```diff
@@ -83,6 +88,16 @@ export function LinkInput({ editingPost, isPosting, onSave }: LinkInputProps) {
...
   const url = useDebouncedValue(form.url, 500);
+  // We need to track debounce dirty state to appropriately disable the send button. useLinkGrabber's
+  // loading state will not be set until the debounce fires.
+  const [isPendingDebounce, setIsPendingDebounce] = useState(false);
+  useEffect(() => {
+    if (form.url !== url) {
+      setIsPendingDebounce(true);
+    } else {
+      setIsPendingDebounce(false);
+    }
```
arthyn, line 99: Seems like it'd be better if `useDebouncedValue` just supported triggering on the lead

arthyn: true

arthyn:
```suggestion
  // We need to track debounce dirty state to appropriately disable the send button. useLinkGrabber's
  // loading state will not be set until the debounce fires.
  const [isPendingDebounce, setIsPendingDebounce] = useState(false);
  useEffect(() => {
    setIsPendingDebounce(form.url !== url);
```

arthyn (approved):
lgtm

PR #4816, opened 2025-05-22
link display tweaks

dnbrwstr:
A number of visual fixes for link display:

- Fix double borders on link blocks in galleries
- Set max width on link blocks in chat channels
- Remove descriptions from link blocks in galleries and chat channels
- Better handling for links with missing titles, descriptions, or images
- Reconcile max widths of images + other blocks
- Fix extra space above references in galleries
- Improves image component types


arthyn (approved):
lgtm

PR #4818, opened 2025-05-22
metagrab: rejected/challenged request handling

Fang-:
Two things:

First, we update the user-agent logic to pull the user-agent string from the user's original request. Only if that's absent do we fall back to the hardcoded fake chrome string.

Second, we add a small set of heuristics for the experimental feature of detecting when our request gets rejected for bot/scraper protection reasons. In that case, we put a `possibly-blocked` string in the `result` field (for `status: 400` results). The client might be able to use this to message differently about the failure, or retry in some other way.

### DIFF
--- a/desk/app/metagrab.hoon
+++ b/desk/app/metagrab.hoon
@@ -133,7 +133,19 @@
     ::TODO  extract message from response body somehow?
     ?:  (gte cod 500)
       [%500 ?~(dat `(scot %ud cod) `q.data.u.dat)]
-    [%400 `(scot %ud cod)]
+    ::  experimental bot-protection detection, which the client may
+    ::  want to know about so it can retry in a more advanced way.
+    ::
+    :-  %400
+    ?:  ?|  ::  datadome header
+            ::
+            =(`'protected' (get-header:http 'x-datadome' headers))
+            ::  captcha service in response body
+            ::
+            &(?=(^ dat) ?=(^ (find "captcha-delivery.com" (trip q.data.u.dat))))
+        ==
+      `'possibly-blocked'
+    `(scot %ud cod)
   ::  miscellaneous
   ::
   ?:  |((lth cod 200) (gte cod 600))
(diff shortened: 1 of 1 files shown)


desk/app/metagrab.hoon
```diff
@@ -132,7 +133,19 @@
     ::TODO  extract message from response body somehow?
     ?:  (gte cod 500)
       [%500 ?~(dat `(scot %ud cod) `q.data.u.dat)]
-    [%400 `(scot %ud cod)]
+    ::  experimental bot-protection detection, which the client may
+    ::  want to know about so it can retry in a more advanced way.
+    ::
+    :-  %400
```
latter-bolden, line 139: I was seeing `401`'s as well. Could we make the heuristic to include this slightly more expansive?

Fang-: This branches handles all `4xx` responses from the target. We flatten those into "status: 400" responses from metagrab, in the same way that all 5xx responses get flattened to "status: 500" etc.

If you care about knowing the "full" original status code, that could in theory be arranged, but any heuristics you might want to do with that should probably be handled in-agent instead.

latter-bolden: ahh I see, sounds like we're covered here then. 

Having indication of the "true" original status code sounds nice, but I can't think of any immediate need that would necessitate a change. Maybe `404` as a signal to avoid attempting via non-metagrab means?

desk/app/metagrab.hoon
```diff
@@ -320,7 +333,14 @@
         ::  no valid cache entry for this target, start a new fetch
         ::
         =.  await  (~(put ju await) u.target id)
-        [[(fetch u.target) ~] this]
+        =/  uas=@t
```
arthyn, line 336: when I was testing this, it wasn't just the user agent string that made a difference

arthyn (approved):
lgtm lets try this, may want more headers though

desk/app/metagrab.hoon
Fang-: Also the `Accept` header, right? We still include that statically as `*/*`.

PR #4819, opened 2025-05-22
image: fix detection for file/data URIs

latter-bolden:
It looks like these weren't being properly detected by our validator. For data URIs, it has a built in method which we instead use. For file URIs, it doesn't like certain forms (_file:///_) so we detect manually.

Also noticed the placeholder asset behavior on mobile looked wonky with the fallback. Special cased that while I was in there.


arthyn (approved):
lgtm

PR #4820, opened 2025-05-23
links: use rich preview fallback

latter-bolden:
Adds the option to fallback to our cloud infra for link previews that are hard to get. The fallback serverless function is already deployed.

Adds a privacy option to disable the fallback option and rely only on `%metagrab` itself.

To test, try any of the links that we were having trouble with.

Fixes TLON-4308

### DIFF
--- a/packages/app/features/settings/PrivacyScreen.tsx
+++ b/packages/app/features/settings/PrivacyScreen.tsx
@@ -175,9 +175,7 @@ export function PrivacySettingsScreen(props: Props) {
 
         <YStack paddingHorizontal="$l" paddingTop="$2xl" gap="$xl">
           <XStack justifyContent="space-between" alignItems="center">
-            <SizableText flexShrink={1}>
-              Disable centralized features
-            </SizableText>
+            <SizableText flexShrink={1}>Disable Tlon helpers</SizableText>
             <Switch
               style={{ flexShrink: 0 }}
               value={state.disableTlonInfraEnhancement}
@@ -185,8 +183,9 @@ export function PrivacySettingsScreen(props: Props) {
             ></Switch>
           </XStack>
           <Text size="$label/s" color="$secondaryText">
-            If disabled, the app will avoid falling back to Tlon's
-            infrastructure for progressive enhancements (e.g. link previews).
+            Your ship will always attempt to generate rich link previews
+            locally. If disabled, the app will avoid making backup requests to
+            Tlon's service if local generation fails.
           </Text>
         </YStack>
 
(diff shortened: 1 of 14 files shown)


packages/app/features/settings/PrivacyScreen.tsx
```diff
@@ -149,6 +173,23 @@ export function PrivacySettingsScreen(props: Props) {
...
+              Disable centralized features
+            </SizableText>
+            <Switch
+              style={{ flexShrink: 0 }}
+              value={state.disableTlonInfraEnhancement}
+              onValueChange={toggleDisableTlonInfraEnhancement}
+            ></Switch>
+          </XStack>
+          <Text size="$label/s" color="$secondaryText">
+            If disabled, the app will avoid falling back to Tlon's
```
latter-bolden, line 188: @jamesacklin copy

jamesacklin: I think "centralized features" is a bit of a misnomer; I would expect this to also disable my image-uploading capabilities and maybe even my hosting account? 

Let's rephrase to:

```
Disable Tlon helpers
Your ship will always attempt to generate rich link previews locally. If disabled, the app will avoid making backup requests to Tlon's service if local generation fails.
```

packages/app/ui/components/PostContent/BlockRenderer.tsx
```diff
@@ -242,7 +242,11 @@ export function LinkBlock({
   }, [block.url]);
 
   return (
-    <Reference.Frame {...props} onPress={clickable ? onPress : undefined}>
+    <Reference.Frame
+      {...props}
+      maxWidth={400}
```
arthyn, line 247: seems like this will override the 600 dan set earlier

packages/shared/src/store/metagrabActions.ts
```diff
@@ -1,12 +1,41 @@
...
+import { createDevLogger } from '../debug';
 import { isValidUrl } from '../logic';
 
+const logger = createDevLogger('metagrabActions', false);
+
+export async function getLinkMetaWithFallback(url: string) {
+  logger.trackEvent('Attempting metagrab link preview');
+  const response = await getLinkMetadata(url);
+
+  const failedToGetMeta =
```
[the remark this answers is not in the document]
arthyn: I think this is probably fine

arthyn (approved):
lgtm

packages/app/ui/components/PostContent/BlockRenderer.tsx
latter-bolden: oop, good catch. I was just toying with it locally, did not mean to commit.

PR #4821, opened 2025-05-27
embeds: store as rich link blocks

latter-bolden:
Updates our handling for embeddable links. Previously we were storing them as inline links without metadata. Now, we'll fetch nice previews for them and have this data available in contexts where we don't want to render an embed.

Updates the Link block renderer to optionally display as an embed if one is available.

Notably, this means we'll get nice rich previews in galleries instead of the embed falling back to a raw URL string.

Fixes TLON-4309


dnbrwstr (approved):
Cool, change looks good to me!

PR #4822, opened 2025-05-27
content-helpers: prevent intermittent inline code termination

jamesacklin:
## Summary

This PR fixes TLON-4249 by making stronger determinations around when to terminate `inline code spans` in messages. 

## Changes

The inline parser splits text by spaces and processes each word independently. `isCodeEnd` only checked if the word ends with a backtick, not considering punctuation after it. Furthermore, it would incorrectly handle any non-space characters, stripping the last character (the backtick) with word.slice(0, -1), and leave any trailing characters as part of the code content.

Now, we:

- Find the exact position of the closing backtick (before any punctuation)
- Extract only the code span up to the backtick (excluding it)
- Extract any punctuation that comes after the backtick
- Add the code span with code formatting marks
- Continue adding the punctuation as regular text (not part of the code span)

## How did I test?

I tested with the following string, which users have commonly used variations of:

```
`some text that should be in an inline code span`, more text
```

I also wrote several tests, which:

- Test the above string and expects it to split the code span directly at the backticks
- Attempt multiple inline code spans with more punctuation in the same message
- Test a code span followed by a space (which works perfectly in `develop` now)

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

A simple revert.

## Screenshots / videos

Before:
image

After:
image


latter-bolden:
@jamesacklin I modified the regex slightly to handle multiple characters after the backtick. 

I think there's also an issue if it ends in a space, but that's trickier to fix.

latter-bolden (approved):

PR #4825, opened 2025-05-28
blocking: restore

jamesacklin:
## Summary

This PR fixes TLON-4247 by restoring the ability to block and unblock users on the web app.

## Changes

On the user's profile screen, I fixed the block/unblock responsiveness by (A) making the function asynchronous and awaiting the response from the store and (B) invalidating the blocked users query to refresh state immediately. I also put an explicit call to sync in `store/dmActions`.

While investigating this bug, I found that unblocking didn't work on the web because we were calling a React Native alert confirmation, which crashed early and didn't fire the unblock event. I conditionally switched confirmation methods based on platform and now clicking a blocked user to unblock them works as expected (and refreshes the list immediately thanks to additional invalidation on handle).

## How did I test?

- Navigate to a user's profile on the web
- Click "Block" in the row of profile action buttons
- Confirm the "Message" button disappears and the "Unblock" button appears
  - Click the "Unblock" button and confirm their profile has all of the default action buttons
  - Click the "Block" button again
- Navigate to Settings -> Blocked users
- Click "Unblock" to unblock the user
- Confirm in the `window.confirm` browser dialog
- See them disappear from the list
- Navigate back to their profile
- Confirm that their profile has all of the default action buttons

I also abstracted the blocking logic out to a separate file `store/blockingActions` and added several simple unit tests.

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Simply revert.

### DIFF
--- a/packages/shared/src/store/__tests__/blockingActions.test.ts
+++ b/packages/shared/src/store/__tests__/blockingActions.test.ts
@@ -28,12 +28,7 @@ describe('blockingLogic', () => {
 
       expect(unblockUser).toHaveBeenCalledWith(userId);
       expect(blockUser).not.toHaveBeenCalled();
-      expect(queryClient.invalidateQueries).toHaveBeenCalledWith({
-        queryKey: ['blockedContacts'],
-      });
-      expect(queryClient.invalidateQueries).toHaveBeenCalledWith({
-        queryKey: [['contact', userId]],
-      });
+      expect(queryClient.invalidateQueries).toHaveBeenCalledTimes(1);
     });
 
     it('should block user when currently unblocked', async () => {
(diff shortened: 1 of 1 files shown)


packages/shared/src/store/blockingActions.ts
```diff
@@ -0,0 +1,27 @@
...
+  isCurrentlyBlocked: boolean
+): Promise<void> {
+  if (isCurrentlyBlocked) {
+    await unblockUser(userId);
+  } else {
+    await blockUser(userId);
+  }
+
+  // Invalidate queries to refresh UI
+  queryClient.invalidateQueries({
```
dnbrwstr, line 15: Any reason we can't use query table dependencies to handle this invalidation? That'd be more consistent.

packages/shared/src/store/dmActions.ts
```diff
@@ -58,6 +59,7 @@ export async function blockUser(userId: string) {
 
   try {
     await api.blockUser(userId);
+    await sync.syncBlockedUsers();
```
dnbrwstr, line 62: Are these explicit calls to sync necessary? Don't seem like they should be

dnbrwstr (approved):
Looks good, a couple nits / questions.

PR #4830, opened 2025-05-29
fix toast context on android

dnbrwstr:
## Summary

Fixes missing `SafeAreaProvider` error on android.

## Changes

Moves the `ToastProvider` inside the primary `PortalProvider` so that toasts rendered by its portal have access to `SafeAreaContext` (since context doesn't propagate through `Portal` on android).

## How did I test?

Tested on web, ios, and mobile.

## Risks and impact

  - [ ] Onboarding
  - [X] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications


davidisaaclee (approved):
code lgtm

ftr, looking at your diff, I don't understand why this was an issue on Android and not iOS. my understanding is that nodes get "moved" in render from the `<Portal>` subtree to the children of `<PortalProvider>` https://github.com/tamagui/tamagui/blob/5adbeb7ee6b6e13325ec83338f3d16e93339ae54/code/ui/portal/src/GorhomPortal.tsx#L343 , with no apparent platform-switching behavior between iOS/Android. my best guess is that somehow directly moving the React node ends up with context getting evaluated differently in Android vs iOS.

I also saw this while looking around: https://github.com/pmndrs/its-fine/blob/12efde68f78ca9a9ca528e24c906b47737cfd9c7/src/index.tsx#L225 which could be used to enable the `ReactDOM.createPortal`-style context inheritance – more info on https://github.com/gorhom/react-native-portal/issues/45

dnbrwstr:
Yeah, it is odd, and maybe worth doing a deeper dive on at some point. Some more info here too: https://github.com/gorhom/react-native-portal/issues/34

PR #4831, opened 2025-05-29
embeds: better twitter behavior

latter-bolden:
## Summary

Better default behavior for Twitter links.

## Changes

- `%metagrab` is insufficient for Twitter links. Instead, always prefer the fallback option for populating those
- Avoid trying to render embeds on Web, instead show the default rich `LinkBlock` visual.

## How did I test?

Paste a tweet link in chat and gallery.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

Simple revert is fine.

## Screenshots / videos
Before
image

After
image


patosullivan:
We could also use the new dumb-proxy to grab twitter oembed data and then we'd be able to display the rich embeds on web

latter-bolden:
@patosullivan I agree, see TLON-4315. 

I didn't want to introduce a new agent dependency on the release branch, but we should try that soon after deploying the protocol change.

PR #4834, opened 2025-06-02
queries: make sure we don't drop blocks

arthyn:
## Summary

This simply makes sure we don't empty out `isBlocked` if a contact is already in the DB and we're inserting it (because we just inserted it from blocks).

## Changes

- changes the `onConflictDoUpdate` to ignore `isBlocked`
- adds a test case to make sure this doesn't change

## How did I test?

Run `pnpm test` and also navigate to Blocked Users screen

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

n/a

## Screenshots / videos

n/a

### DIFF
--- a/packages/shared/src/db/queries.test.ts
+++ b/packages/shared/src/db/queries.test.ts
@@ -1,8 +1,10 @@
 import { expect, test } from 'vitest';
 
+import { v0PeersToClientProfiles } from '../api';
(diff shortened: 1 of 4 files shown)


dnbrwstr (approved):
Lgtm!

PR #4835, opened 2025-06-02
web: fix imageviewer crash

latter-bolden:
## Summary

Our newly upgraded Tamagui pulls in a newer `react-native-web` runtime. The required version prevents `findNodeHandle` calls, which have been deprecated in React itself. Our web implementation of the `ImageViewer` lightbox has a gesture detector + touch enabled zoom that calls out to this now invalid `findNodeHandle` api, causing a crash.

I'm not totally clear on where the call is happening. Removing `GestureDetector` or `Zoomable` independently doesn't solve the issue, but ripping out both does. There's [some old discussion](https://github.com/software-mansion/react-native-gesture-handler/issues/1036#event-6342907979) in `react-native-gesture-handler` about avoiding these calls.

## Changes

- Only wrap in a `GestureDetector` on mobile
- Display a standard image in the lightbox instead of one from `react-native-image-zoom`

As far as I can tell, the zoom functionality wasn't working on web anyhow. So removing this should have no UX impact.

## How did I test?

Click an image in chat before and after the changes on Web.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

The PR can be reverted, but we'll need a replacement fix immediately.

## Screenshots / videos
N/A

Fixes TLON-4342

### DIFF
--- a/packages/app/ui/components/ImageViewerScreenView.tsx
+++ b/packages/app/ui/components/ImageViewerScreenView.tsx
@@ -15,6 +15,7 @@ import {
 } from 'react-native';
 import {
   Directions,
+  FlingGesture,
   Gesture,
   GestureDetector,
 } from 'react-native-gesture-handler';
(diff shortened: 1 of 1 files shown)


patosullivan (approved):
Changes make sense given the rnw update. We don't need drag/zoom on web anyway.

Tested on web, iOS, and Android. Works well.

PR #4837, opened 2025-06-03
embeds: avoid duplicate display

latter-bolden:
## Summary

We previously parsed embeds from inline links, but now we attach them directly as rich link blocks. We need to avoid displaying duplicates, so we should avoid doing legacy inline parsing if we already have an attached link block. 

## Changes

Update the inline parser to avoid extracting embeds if a top level link block already exists.

## How did I test?

Post a youtube link without removing the URL.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

Revert the PR and push an alternative fix.


arthyn (approved):
lgtm

PR #4842, opened 2025-06-04
chat: fix dm responses

arthyn:
## Summary

We somehow did not catch that we weren't using the new marks and subscriptions from the upgrade. Additionally after using those, we discovered an issue with the way that %chat was trying to give responses. This fixes both of those so that DMs now are able to send the new link blocks.

## Changes

- adjusted the way we were giving `writ-response-1` and `writ-response` so that the types were correct and checked
- swapped our %chat subscription to v1
- swapped our chat types to match new backend types
- swapped poke marks to match

## How did I test?

Manually live on screenshare

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Consult the author

## Screenshots / videos

n/a

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -1559,13 +1559,14 @@
     =/  =whom:c  [%club id]
     =/  response=(unit response:writs:c)  (diff-to-response diff pact.club)
     ?~  response  cu-core
-    =/  old-response-3  [whom (old-response-writs-3:pac u.response)]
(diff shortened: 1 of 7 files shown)


Fang- (approved):
Looking at this again I would lightly prefer types on the `=/` instead of inside the `!>` with how it's currently written, but it's fine either way.

Hoon side changes look good.

PR #4847, opened 2025-06-05
NotebookPost: correct title type size

jamesacklin:
## Summary

Fixes TLON-4373 by explicitly passing `size` in to `NotebookPostTitle` rather than relying on prop inheritance.

## Changes

`NotebookPostTitle` had improper prop handling which became an issue when we upgraded Tamagui to the latest version. We used a context-based system for styling but the size prop was never passed. The recent Tamagui update made this more visible. 

## How did I test?

- Navigate to a Notebook channel
- Confirm that the post title text is the right size
- Navigate in to a Notebook post detail view
- Confirm that the post title text is the right size
- Subscribe to a Notebook channel w/ full notifications; add a post from another ship
- Confirm the title text size appears correctly in the Activity feed

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

Revert if needed

## Screenshots / videos

Screenshot 2025-06-05 at 9 04 15 AM


jamesacklin:
@dnbrwstr can you take a look again? I think I misinterpreted how the context is supposed to work.

dnbrwstr (approved):
Nice, lgtm!

PR #4848, opened 2025-06-05
chat input: fix image paste

latter-bolden:
## Summary

We broke image paste on the protocol release branch. This restores that functionality (while still handling image link pastes as well)

## Changes

- use the existing `usePasteHandler` in `BareChatInput`
- permit `blob` source paths for images in our base `Image` component

## How did I test?

Pasted images on web. ~Waiting for mobile to finish building, will try there as well~ it appears we don't support pasting images on mobile, its absence there isn't a regression

## Risks and impact

- Safe to rollback without consulting PR author? yes
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Revert and push alternative fix.

Fixes TLON-4378


dnbrwstr (approved):
Lgtm!

PR #4850, opened 2025-06-06
broadcaster: migrate chat types

arthyn:
## Summary

We forgot to migrate broadcaster because it is not enabled by default so it never tried compiling. This just adds migration for the new types. 

## Changes

- moved helper functions for writ transitions to library
- changed %chat to use said helper functions
- added migrations for %broadcaster to upgrade correctly

## How did I test?

I temporarily added %broadcaster to desk.bill and it compiled

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/app/broadcaster.hoon
+++ b/desk/app/broadcaster.hoon
@@ -5,7 +5,7 @@
     dbug, verb
 ::
 |%
-+$  versioned-state  $%(current-state state:v0)
++$  versioned-state  $%(current-state state-0)
 +$  current-state
   $:  %1
       cohorts=(map @t cohort)
@@ -32,13 +32,13 @@
   $%  action
   ==
 ::
-++  v0
-  |%
-  +$  state
++$  state-0
     $:  %0
-        cohorts=(map @t cohort)
+        cohorts=(map @t cohort:v0)
     ==
   ::
+++  v0
+  |%
   +$  cohort
     $:  targets=(set ship)
         logging=(list relive)
@@ -72,7 +72,7 @@
   ?>  ?=(%1 -.old)
   [~ this(state old)]
   ++  state-0-to-1
-    |=  old=state:v0
+    |=  old=state-0
     ^-  current-state
     [%1 (cohorts-0-to-1 cohorts.old)]
   ++  cohorts-0-to-1


arthyn:
Pointed at staging for ease of review, develop still needs to have staging merged to it

arthyn:
@midlyx-hatrys try running:
`|pause %groups`
`|install ~sogryp-dister-dozzod-dozzod %groups`

Do you by chance have %broadcaster agent turned on for the %groups desk? This PR fixes that but hasn't been deployed yet.

desk/app/broadcaster.hoon
```diff
@@ -1,48 +1,59 @@
...
+        cohorts=(map @t cohort)
+    ==
+  ::
+  +$  cohort
+    $:  targets=(set ship)
+        logging=(list relive)
+        outward=(list writ:c3)  ::NOTE  invented-here fake seal?
+    ==
+  --
+::
```
mikolajpp, line 48: While we do use state version anchor for types, so this is much appreciated, the current practice is to keep the state types explicitly versioned with a tag. I think in general it makes it for easier reading, since all state types are usually aggregated in the same place. My suggestion here is to define `$state-0` outside of the `|v0` core.

desk/app/broadcaster.hoon
```diff
@@ -1,48 +1,59 @@
...
-/-  c=chat
-/+  cj=channel-json,
+/-  c=chat, ch=channels, c3=chat-3
+/+  cj=channel-json, dm,
     dbug, verb
 ::
 |%
-+$  state-0
-  $:  %0
++$  versioned-state  $%(current-state state:v0)
```
mikolajpp, line 8:
```suggestion
+$  versioned-state  $%(current-state state-0)
```
If we follow the convention and explicitly version the state type.

mikolajpp (changes requested):
This looks good. I left a suggestion to explicitly version the state type and leave it outside the version core, as we do elsewhere.

mikolajpp (approved):
Looks good to go.

PR #4852, opened 2025-06-09
queries: split perms and order to avoid transaction issues

arthyn:
## Summary

This was causing channels to not load and was a late addition to my role change PR so that's why it didn't get caught. I wasn't sure how to keep `insertChannelPerms` with order insertion.

## Changes

- this splits inserting channel perms and order insertion since they aren't really required to be together
- adds new order insertion to sync

## How did I test?

Manually attempting to load channels and ran `pnpm test`

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/packages/shared/src/db/queries.ts
+++ b/packages/shared/src/db/queries.ts
@@ -1243,7 +1243,7 @@ export const getFlaggedPosts = createReadQuery(
 
 export const insertChannelPerms = createWriteQuery(
   'insertChannelPerms',
-  async (channelsInit: ChannelInit[], ctx: QueryCtx) => {
+  async (channelsInit: Omit<ChannelInit, 'order'>[], ctx: QueryCtx) => {
     const writers = channelsInit.flatMap((chanInit) =>
       (chanInit.writers || []).map((writer) => ({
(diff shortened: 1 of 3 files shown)


latter-bolden (approved):
Can confirm this fixes the issue for me locally

packages/shared/src/db/queries.ts
```diff
@@ -1295,7 +1295,13 @@ export const insertChannelPerms = createWriteQuery(
           set: conflictUpdateSetAll($channelWriters),
         });
     }
+  },
+  ['channelWriters', 'channels']
+);
 
+export const insertChannelOrder = createWriteQuery(
+  'insertChannelOrder',
+  async (channelsInit: ChannelInit[], ctx: QueryCtx) => {
```
davidisaaclee, line 1304: narrowing the type here would be nice to avoid future conflation

```suggestion
  async (channelsInit: Array<{ order: string[] }>, ctx: QueryCtx) => {
```

davidisaaclee (approved):

PR #4859, opened 2025-06-12
metagrab: fetch HEAD before doing a full GET

Fang-:
## Summary

Makes metagrab fetch HEAD first, and only do a full GET if it's necessary (for getting & parsing an html page).

## Changes

We only care about the response body when it's an html page. In all
other cases, we discard the body entirely and only extract data from the
headers.

Some urls might point to very large resources. Past a certain point, the
response is too big for vere to handle properly, and it will die.

Vere-side fix for that is forthcoming, but it's also not necessary for
metagrab to request the full responses in the first place. Instead, a
HEAD request will suffice.

Here, instead of going straight to a GET request, we always make a HEAD
request first. If the response has a text/html content-type, we proceed
with the GET request as normal. Otherwise, we make do with the headers.

Additionally, we now include a `size` field for `type: "file"` responses, and we no longer retry fetches whose request was `%cancel`ed by iris/cttp.c, lest we accidentally get into a crash-retry loop.

## How did I test?

Tested the metagrab endpoint on a local fakezod with html page urls, file urls, and specifically the 1.2gb file that made us discover the problems with doing straight GET requests always.

## Risks and impact

- Safe to rollback without consulting PR author, yes.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

State type change in ffcee37 need some additional attention during rollback, but can follow the same pattern of just ditching the previous state entirely. It's all transient/cache, anyway.

## Screenshots / videos

If only.

### DIFF
--- a/desk/app/metagrab.hoon
+++ b/desk/app/metagrab.hoon
@@ -104,17 +104,17 @@
   ==
 ::
 ++  fetch
-  |=  [url=@t uas=@t]
+  |=  [met=?(%head %get) url=@t uas=@t]
   ^-  card
   =/  =header-list:http
     :~  ['user-agent' uas]
         ['accept' '*/*']
     ==
   =/  =request:http
-    [%'GET' url header-list ~]
+    [?-(met %head %'HEAD', %get %'GET') url header-list ~]
   ::NOTE  outbound-config is actually meaningless,
   ::      iris doesn't do anything with it at present...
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

PR #4861, opened 2025-06-16
e2e follow up

patosullivan:
## Summary

Refactored the rube script, `playwright.config.ts`, and `auth.setup.ts` to use the list of ships from `shipManifest.json`. When adding or removing ships we no longer need to update all of these files, just the `shipManifest.json` file.

Added stronger types throughout the rube script.

Improved the process cleanup when the script is killed (this could still use some work, I'll make an issue for it. It's low priority imo).

Added tests for DM protocol mismatch and group protocol mismatch (fixes tlon-4396)

Added a test for ensuring DMs are received on other ship (fixes tlon-4395)

Added a test for ensuring group changes appear across ships (fixes tlon-4394)

Updates fakeships to latest so that we speed up the e2e tests (fixes tlon-4419)

## How did I test?

These are tests.

## Risks and impact

- Affects important code area: e2e/ci.
## Rollback plan
Just revert the merge commit.


dnbrwstr (approved):
Looks good to me!

PR #4865, opened 2025-06-16
mentions: fix regex crashes

arthyn:
## Summary

Fixes TLON-4439

## Changes

- escaped matched text regex construction to prevent things like "mismatching groups" etc.

## How did I test?

Manually running pnpm dev

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

n/a

## Screenshots / videos

n/a

### DIFF
--- a/packages/app/ui/components/ContactName.tsx
+++ b/packages/app/ui/components/ContactName.tsx
@@ -1,3 +1,4 @@
+import { escapeRegExp } from '@tloncorp/shared';
 import { Text } from '@tloncorp/ui';
 import { ComponentProps, useMemo } from 'react';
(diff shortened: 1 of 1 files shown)


patosullivan:
Could we maybe add a step in the chat functionality e2e test to catch mention bugs in the future?

latter-bolden (approved):
Fix makes sense, mentions worked on web and mobile.

If you don't wanna bite off an e2e test for mentions right now, could you file a ticket so we don't forget?

arthyn:
yeah I'll make an issue

PR #4867, opened 2025-06-17
chat: include sequence numbers

Fang-:
## Summary

Adds sequence numbers to chat (DM) messages, like we did for channels.

Closes TLON-4405 and TLON-4406.

## Changes

Adds sequence numbers to chat `$writ`s. Includes all the relevant api bumps. Note that this only affects client-facing apis. Inter-ship communication is unaffected and does not get bumped.

In practice, from a json perspective, the new types are backwards compatible: the only change is an additional 'seq' field inside the `$seal` object.

Also sneaks in a change that prevent chat scries from crashing in cases where they come up with zero messages.

## How did I test?

Smoke-tested by loading a DM, sending messages into it, refreshing etc. using old/current web client. Tested for difference between old and new scry endpoints.

Implemented a sequence-number-range fetching endpoint and saw that numbers appear on pre-migration messages as expected. (That endpoint is sorta outside the immediate scope of this PR.)

Would be good to throw /lib/discipline against this, probably, but I haven't yet. Should sit down for a couple days and finish thorough integration of that thing.

## Risks and impact

- Can't trivially rollback without consulting PR author.
  - [ ] Onboarding
  - [x] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Would need to write a state migration that drops the version numbers. Should then be able to revert all api/mark code changes, given that these changes only impact client-facing apis.

## Screenshots / videos

[Video.](https://www.youtube.com/watch?v=dQw4w9WgXcQ)


desk/lib/dm.hoon
```diff
@@ -384,6 +406,32 @@
...
+      (bap:on:writs:c wit.pac)
+    =|  out=(list [time writ:c])
+    |-
+    ?~  wits  ~
+    ?:  (gth seq.p.i.wits end)
+      $(wits t.wits)
+    ?:  (lth seq.p.i.wits start)
+      ~  ::  done
+    :-  i.wits
+    $(wits t.wits)
```
mikolajpp, line 434: This is not tail-recursive, is it? We probably want to use `out` from above.

desk/mar/chat/heads-2.hoon
```diff
@@ -0,0 +1,14 @@
+/-  c=chat
+/+  j=chat-json
+|_  =chat-heads:c
```
mikolajpp, line 3: This obviously works, but since we decided to version chat with separate sur files, this should ideally import the versioned sur file. Then we don't need to adjust all these marks next time :-)

```suggestion
/-  c=chat-5
/+  j=chat-json
|_  =chat-heads:c
```

desk/mar/chat/paged-writs-2.hoon
```diff
@@ -0,0 +1,14 @@
+/-  c=chat
```
mikolajpp, line 1:
```suggestion
/-  c=chat-5
```

desk/mar/chat/heads-2.hoon
```diff
@@ -0,0 +1,14 @@
...
+++  grad  %noun
+++  grow
+  |%
+  ++  noun  chat-heads
+  ++  json  (chat-heads:enjs:j chat-heads)
```
mikolajpp, line 8: Same here - it would be better to version json encoders too, especially since they might be silently updated. (How long since anyone had looked at mark-warmer?) As they are all in the same file for now,  we would just need to put in a version anchor at top level.
```suggestion
  ++  json  (chat-heads:v9:enjs:j chat-heads)
```

desk/mar/chat/paged-writs-2.hoon
```diff
@@ -0,0 +1,14 @@
...
+/+  j=chat-json
+|_  =paged-writs:c
+++  grad  %noun
+++  grow
+  |%
+  ++  noun  paged-writs
+  ++  json  (paged-writs:enjs:j paged-writs)
```
mikolajpp, line 8:
```suggestion
  ++  json  (paged-writs:v9:enjs:j paged-writs)
```

desk/mar/chat/scam-2.hoon
```diff
@@ -0,0 +1,19 @@
+/-  c=chat
```
mikolajpp, line 1:
```suggestion
/-  c=chat-5
```

desk/mar/chat/scam-2.hoon
```diff
@@ -0,0 +1,19 @@
...
+|_  =scam:c
+++  grad  %noun
+++  grow
+  |%
+  ++  noun  scam
+  ++  json
+    =,  enjs:format
+    %-  pairs
+    :~  'last'^?~(last.scam ~ (time-id:enjs:j u.last.scam))
+        'scan'^a+(turn scan.scam reference:enjs:j)
```
mikolajpp, line 12:
```suggestion
        'scan'^a+(turn scan.scam reference:v9:enjs:j)
```

desk/mar/chat/scan-2.hoon
```diff
@@ -0,0 +1,16 @@
+/-  c=chat
```
mikolajpp, line 1:
```suggestion
/-  c=chat-5
```

desk/mar/chat/scan-2.hoon
```diff
@@ -0,0 +1,16 @@
...
+/+  j=chat-json
+|_  =scan:c
+++  grad  %noun
+++  grow
+  |%
+  ++  noun  scan
+  ++  json
+    =,  enjs:format
+    a+(turn scan reference:enjs:j)
```
mikolajpp, line 10:
```suggestion
    a+(turn scan reference:v9:enjs:j)
```

desk/mar/writ-response-2.hoon
```diff
@@ -0,0 +1,14 @@
+/-  c=chat
```
mikolajpp, line 1:
```suggestion
/-  c=chat-5
```

desk/mar/writ-response-2.hoon
```diff
@@ -0,0 +1,14 @@
...
+/+  j=chat-json
+|_  [=whom:c =response:writs:c]
+++  grad  %noun
+++  grow
+  |%
+  ++  noun  [whom response]
+  ++  json  (writs-response:enjs:j whom response)
```
mikolajpp, line 8:
```suggestion
  ++  json  (writs-response:v9:enjs:j whom response)
```

mikolajpp (changes requested):
There are two things that need to be addressed:
(1) The range scry endpoint does seem to use a tail-unoptimized loop to compute a post range - probably an omission since the variable is already there.
(2) The new marks should ideally use the latest version `v5` for types and `v9`[^1] for JSON encoders. Otherwise we will need to check and adjust all of them in next %chat update.

Apart from these, looks good!

desk/lib/dm.hoon
Fang-: What do you mean, not tail-recursive? We put the recursion in the tail of the noun we produce, it's the textbook case lol. You're correct that we don't use out (vestigial from an earlier implementation), we don't need to accumulate when we can produce the product noun "in place".

mikolajpp: I was thinking about the cons there and wasn't sure if the Hoon compiler knows how to optimize it. If it does that is good to know!

desk/app/chat.hoon
```diff
@@ -1094,9 +1175,11 @@
     `[p.diff delta]
   ?+  -.q.diff  q.diff
       %add
-    =/  time=(unit time)  (~(get by dex.pact) p.diff)
+    =/  time=(unit time)    (~(get by dex.pact) p.diff)
     ?~  time  ~
-    [%add essay.q.diff u.time]
+    =/  writ=(unit writ:c)  (~(get by wit.pact) u.time)
+    ?~  writ  ~&(%diff-to-response-miss ~)
```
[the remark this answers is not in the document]
Fang-: It should be impossible (we shouldn't have entries in `dex.pact` that don't have a corresponding entry in `wit.pact`), but I'm not sure how safe/reasonable it is to crash here.

Calling this arm is part of handling incoming pokes, so crashing would produce a poke-nack instead of calling our own `+on-fail`. The sender might retry, but the message would eventually end up dropped on the floor, which is Very Bad.

I'll go and add logging into the cases where `+diff-to-response` produces a `~` result though.

desk/app/chat.hoon
```diff
@@ -1094,9 +1175,11 @@
     `[p.diff delta]
   ?+  -.q.diff  q.diff
       %add
-    =/  time=(unit time)  (~(get by dex.pact) p.diff)
+    =/  time=(unit time)    (~(get by dex.pact) p.diff)
```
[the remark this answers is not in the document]
Fang-: This aligns it with the `=/  writ` below.

mikolajpp (approved):
Changes look good, approved!

PR #4868, opened 2025-06-18
web: use web-native list for chat channels

davidisaaclee:
fixes TLON-3320
## Summary

When displaying a single-column channel, the web client will now use a custom web-only component instead of `FlatList`. This enables better text selection behavior.

## Changes

- For this change, we need to switch between the FlatList and web-specific list implementations – `Scroller` now calls out to a `PostList` component; we do the switching between `PostListFlatList` and the web-based `PostList` in that component
- The web `PostList`'s render tree is very simple – we're simply rendering all of the posts that we've loaded in a flex container. Notably, **no virtualization** – see heading below
- [extra] 008fdc0 shows overflow `...` trigger for posts on web, regardless of viewport width. **This means people who are using the web app on mobile will be forced to use the hover-based overflow menu.** I could not trigger the long-press on web, so I think this is not a regression.
- Noted bugs that (more or less) existed before this change:
  - ~~on a narrow web client, mouseover-ing on the most recent post of a chat causes extra space to be added to the bottom of the scroll. I think this is our existing phantom post menu mounting on hover.~~ fixed in [008fdc0](https://github.com/tloncorp/tlon-apps/pull/4868/commits/008fdc03080dbfcc9c68b824e3a798e314784236)
  - inline notices like `____ joined the chat` are incorrectly colored / aligned
  - on native, navigating into a thread and then back to main timeline shows a blank screen briefly. This was previously coincidentally avoided by a incorrectly-written "is ready to display" check (i.e. incorrect for first load, correct for subsequent appearances where we shouldn't need to wait)

### No virtualization??
- We could use a virtualization library like react-virtuoso – but there are two big downsides that seem difficult to overcome:
  - our post heights can range widely – which means we'll (inevitably?) have scroll jumping due to the unrendered elements using incorrect estimated heights
  - virtualization by omitting occluded elements from the DOM breaks the browser's "find in document" feature, which stinks!
- If we want to add virtualization, [`content-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility) seems to be the best way to do so, and we shouldn't need to change our implementation much. (Dumbly throwing `content-visibility: auto` in causes some problems with our scroll-to-anchor code; and there is scroll jumping.)
- `FlatList`'s web implementation does not have virtualization, so we are not regressing in this way.

## How did I test?

- Ran client on macOS Firefox and Safari; and on iOS native client.
- Updated the `Channel.chatWithSimulatedLoad` fixture to simulate lots of different load scenarios.
- Opened big channels of the three main types on iOS native client and navigated around; opened threads
- Opened small channels (I don't know how to use dev client with real account, so I'm limited to my personal test channels) of the three main types on web client and navigated around; opened threads

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

git revert

### Pagination in both directions, search

https://github.com/user-attachments/assets/aa5003ee-a11d-4a9d-9b9d-ed9526f84e49

### Copy-paste, and "stick to bottom" behavior

https://github.com/user-attachments/assets/1721c199-5cf2-43bb-a9b0-95911ac914ce

### New channel, sending/receiving a couple of messages

https://github.com/user-attachments/assets/ac9528b5-9966-4e35-a79e-378c07aed68a

### Thread blank screen flash mentioned above

https://github.com/user-attachments/assets/ad3c750d-864d-43d6-ace7-c9231a05004b

### Feature flag (**visible on mobile** - filtering it out isn't trivial but not that hard - just didn't seem worth adding complexity to our flags code)
Screenshot 2025-06-18 at 10 31 04 AM

### Safari example

https://github.com/user-attachments/assets/db5227b5-81df-43e3-801a-b67f279b2fff

Note the blank parts of the screen directly after loading more content. afaict, the content is positioned correctly, but there's just something preventing the scroll rendering until a subsequent scroll. My guess is Safari doesn't like us calling `scrollBy` in the way that we are, and e.g. maybe we need to batch them and queue them in a `requestAnimationFrame`.


packages/app/ui/components/Channel/useAnchorScrollLock.ts
```diff
@@ -173,16 +171,6 @@ export function useAnchorScrollLock({
     }
   }, [anchor?.postId]);
 
-  // Reset states when explicitly scrolling to bottom
```
davidisaaclee, line 176: I don't see any bad effects from removing this – tested by loading into a channel at an anchor, with lots of messages to forward-fill, and pressing go-to-bottom – I got to bottom just fine. but maybe this is more of an issue when it takes time to load the anchor? in any case, I don't understand what resetting the state does – it feels like this would possibly cause us to try to jump to anchor again after scrolling to bottom, which doesn't seem like what we'd want

packages/app/ui/contexts/scroll.tsx
```diff
@@ -26,43 +26,48 @@ export const ScrollContext = createContext<ScrollContextTuple>(INITIAL_VALUE);
 export const useScrollContext = () => useContext(ScrollContext);
 
 export const useScrollDirectionTracker = ({
```
davidisaaclee, line 28: the effective change here is returning a render-incurring `isAtBottom` along with the already-returned scroll handler

packages/app/ui/contexts/scroll.tsx
```diff
@@ -26,43 +26,48 @@ export const ScrollContext = createContext<ScrollContextTuple>(INITIAL_VALUE);
...
 
-    previousScrollValue.value = y;
+      previousScrollValue.value = y;
+
+      const atBottom = y <= AT_BOTTOM_THRESHOLD;
 
-    const atBottom = y <= AT_BOTTOM_THRESHOLD;
-    if (isAtBottom.value !== atBottom) {
-      isAtBottom.value = atBottom;
       runOnJS(setIsAtBottom)(atBottom);
```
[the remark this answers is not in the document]
davidisaaclee: GREAT POINT

davidisaaclee:
bug: loading into a channel with a large frontscroll doesn't trigger `onStartReached` (until content size changes once)

davidisaaclee:
looks like this needs some more Safari testing, closing until that's stable!

davidisaaclee:
The main change since posting this PR (at the pre-rebased [5fbc0d7](https://github.com/tloncorp/tlon-apps/pull/4868/commits/5fbc0d7c933d2a727d62250b642b5e6bd4022ff1)) is [Handle list content changing size asynchronously without overflow-anchor](https://github.com/tloncorp/tlon-apps/pull/4868/commits/0ccee71e6a17b67cb86155a51b06ae9daf49badf)

Problem:
- As we load new content in at the top of the list, we rely on browser [scroll anchoring](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scroll_anchoring/Scroll_anchoring) to maintain the visible content position so that content doesn't move away from the user (e.g. as we backfill chat and add posts to the visual top of the list)
- Scroll anchoring is not available everywhere – notably not available in Safari
- Where we don't have native scroll anchoring, we manually scroll anchor, using the hook [`useManualScrollAnchoring`](https://github.com/tloncorp/tlon-apps/blob/f2b43d1f23cdce01898ca9cae0f14739fb1c6143/packages/app/ui/components/Channel/PostList/PostList.web.tsx#L468)
- Previously, this hook used the change in scroll height to determine how much to jump (i.e. if we added 1000px of height, we should jump 1000px down). This only works if the added content is _above_ the user's viewport into the scroll. So we previously checked if the list data changed above the viewport, and if so, performed a jump: https://github.com/tloncorp/tlon-apps/blob/ca1b55fdb294535f7cf699d5656ab69c03e990a4/packages/app/ui/components/Channel/PostList/PostList.web.tsx#L122-L125
- This generally worked for static content and is efficient – but **fails when a list item changes height asynchronously** (e.g. a post image loads in). This wasn't caught during dev because the fixture post images were all being cached :(

To fix this:
- Follow RN's example and determine content adjustment amount by measuring the position of a chosen "anchor item": when the anchor item changes position, see how much it moved and move the scroll position by that amount.
- This is better than using scroll height: if we use a _currently-visible item_ as the anchor item, we're ensuring that we move the scroll position only enough to keep that visible item in view – which is pretty much exactly the behavior we want.
- Each time the scroll height changes, we:
  1. Get the stored anchor item's position before scroll change
  2. Get the current position for that item
  3. Subtract them, and scroll the scroll view by that amount
  4. Find a new anchor item and store it for the next adjustment
    When the manual scroll anchor coordinator asks for an anchor item, we give an item at the top of the user's viewport: https://github.com/tloncorp/tlon-apps/blob/f2b43d1f23cdce01898ca9cae0f14739fb1c6143/packages/app/ui/components/Channel/PostList/PostList.web.tsx#L105-L124

_Why sample the next anchor item directly after jumping? Couldn't that anchor item no longer be in view by the time the user pages in more data?_
Yeah, that's probably right. I came to this solution because:
- the event that kicks off this process is a scroll height resize
- we need to know the anchor measurements _previous_ to the scroll resize
- the "best" anchor changes on every user scroll

I think a more robust solution would be to sample the anchor on every scroll event (`scroller.addEventListener('scroll', () => this.updateAnchorItem())`), but I think this would thrash layout since we need to access `offsetTop`.
This way of choosing an anchor seems to work in my testing 🤷 and I think reduces reflows to a quick burst when scroll height changes.

davidisaaclee:
Added Safari screen recording in main PR description, which also shows a new bug where the screen partially blanks out when loading new content. I think we can fix this in an additive way and shouldn't change the architecture of this PR, so I don't think it should block review.

davidisaaclee:
appreciate the reviews!

> always hit a hard stop when scrolling back before you start loading
I don't see this happening consistently, although this is definitely a shape of bug I've been seeing (although I'd hoped that [6587da6](https://github.com/tloncorp/tlon-apps/pull/4868/commits/6587da6a2387e2f1597dd6ead8a662f3af893318) had fixed it 😢). Curious if you have a channel / state where it's happening consistently for you.

I'm going to go ahead and merge despite that bug, since we have this being a flag and I'd like to avoid conflicts building up.

PR #4869, opened 2025-06-19
anchor-lock: dont scroll to anchor if we already have

arthyn:
## Summary

OTT, fixes TLON-4156

## Changes

- simply uses boolean check to make sure we haven't already scrolled to anchor

## How did I test?

This is one way to simulate it. I think in reality this happens more with images/embeds but I didn't want to chase that.

Have two ships belonging to the same channel, while one ship (A) is away, create enough messages from the other ship (B) that would scroll you up a screen. Then from ship A, visit the channel with the unreads. It should bring you to the unread anchor (may be slightly above top of viewport). After this happens scroll to the bottom of the chat. Then from ship B post a message with lots of text like lorem ipsum. Then on ship A resize the window. It should not scroll you back up to the anchor with this change (but it will in production). 

This makes a call to `handleItemLayout` which we give to the scroller that does the anchor scrolling. Its unclear why we use this function to scroll to the anchor, but it seems like there's probably a better place to call this.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [X] Channel display
  - [ ] Notifications

### DIFF
--- a/packages/app/ui/components/Channel/useAnchorScrollLock.ts
+++ b/packages/app/ui/components/Channel/useAnchorScrollLock.ts
@@ -92,6 +92,7 @@ export function useAnchorScrollLock({
 
       if (
         !userHasScrolled &&
+        !didScrollToAnchor &&
         post.id === anchor?.postId &&
         flatListRef.current &&
         anchor?.postId === currentAnchorId.current &&


davidisaaclee:
> After this happens scroll to the bottom of the chat.

are you using the "go to bottom" button? if so, this might be causing the jump to anchor https://github.com/tloncorp/tlon-apps/blob/0e72d87b61d6cbb33975bc23a86c40720f1c3796/packages/app/ui/components/Channel/useAnchorScrollLock.ts#L177-L184

> Its unclear why we use [`handleItemLayout`] to scroll to the anchor

Since list items are laid out asynchronously, attempting to scroll to an item before its vertical position is determined will always fail – so we use `onLayout` with a [check that the calling view is the list item corresponding to our scroll target](https://github.com/tloncorp/tlon-apps/blob/0e72d87b61d6cbb33975bc23a86c40720f1c3796/packages/app/ui/components/Channel/useAnchorScrollLock.ts#L95).

dnbrwstr:
One thing I'm not clear on here -- since this involves scrolling on ship A, shouldn't `userHasScrolled` be getting set to true already, thus short-circuiting the scrollTo? That's how it's supposed to work.

latter-bolden (approved):

PR #4870, opened 2025-06-19
queries: fix query keys and table deps

arthyn:
## Summary

Not associated with any particular issue founding while exploring tloncorp/tlon-apps#4866. One thing to point out here is that we don't enforce that our query keys follow a certain pattern, even though that's the only way they'll get invalidated from the DB.

## Changes

- `getThreadUnreadState` had the wrong table dependencies
- Many queries in `dbHooks.ts` did not have correct query keys

## How did I test?

I didn't really. I found issues with getting reliable updates in my other PR, and this helped alleviate some of that. 

## Risks and impact

  - [ ] Onboarding
  - [X] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications


packages/shared/src/store/dbHooks.ts
```diff
@@ -353,7 +353,10 @@ export const useJoinedGroupsCount = () => {
 
 export const useGroupByChannel = (channelId: string) => {
   return useQuery({
-    queryKey: [['group', channelId]],
+    queryKey: [
+      ['group', channelId],
+      useKeyFromQueryDeps(db.getGroupByChannel, channelId),
```
latter-bolden, line 358: side note: I know we do this elsewhere, but I'm surprised calling hooks from inside the parameters of another hook is allowed?

latter-bolden (approved):
Good finds!

packages/shared/src/store/dbHooks.ts
arthyn: yeah I know! its just about order of call. as long as you're not conditionally calling it, it's fine to call it however you wish.

PR #4873, opened 2025-06-20
web: remove some erroneous react-native-gesture-handler imports

patosullivan:
## Summary

There were two more places where we were accidentally importing components from `react-native-gesture-handler` where we should have been importing from `react-native`. This fixes those imports.

We can't import anything from `react-native-gesture-handler` on components that need to be rendered on web, because of the `Error: findNodeHandle is not supported on web. Use the ref property on the component instead.` issue.

As far as I can tell, after this PR, there will be no places on web where a use could run into a gesture handler component (and I don't think we plan on allowing web users to use gestures at all, for now).

## Changes

Changed some import statements.

## How did I test?

Ran locally and made sure the app didn't crash/show white screen.

## Risks and impact

- Safe to rollback without consulting PR author? No, the notebook input won't work.

## Rollback plan

You'd ned to find another workaround for this issue.

### DIFF
--- a/packages/app/fixtures/PostReference.fixture.tsx
+++ b/packages/app/fixtures/PostReference.fixture.tsx
@@ -1,6 +1,6 @@
 import * as db from '@tloncorp/shared/db';
 import { PropsWithChildren } from 'react';
-import { ScrollView } from 'react-native-gesture-handler';
+import { ScrollView } from 'react-native';
 
 import { View } from '../ui';
 import {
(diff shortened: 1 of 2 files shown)


davidisaaclee (approved):
changes looks good to me, but I think we need a more robust solution. if we're really saying that we won't use RNGH on web, maybe we could:
- add a rule to web lint or build that fails when importing from RNGH
- if we need to import from RNGH on web (so we can share components) but we just want it to no-op, maybe we can add a file to re-export RNGH stuff, and add a `.web.ts` version that gives no-op versions (then add lint rule to forbid importing directly from RNGH)

PR #4874, opened 2025-06-20
use Gemfile to pin CocoaPods CLI 

davidisaaclee:
> After this PR is merged, use `pnpm run deps:ios` instead of `pod install`. (Or 

## Summary
Pins the `pod` CLI version using a Gemfile. This should fix at least a subset of the `Podfile.lock` churn we've seen.

There might be other things causing churn – see https://github.com/facebook/react-native/issues/31121
Version differences in `pod` cause at least one line of churn in Podfile.lock, and this change seems positive.

## Changes
- Add a `Gemfile` pinning `cocoapods`
- Add `pnpm` scripts to call `pod` using `bundle exec`
  - `pnpm run deps:ios` can now be run instead of `pod install`
  - `pnpm run deps` installs `pnpm` deps _and_ CocoaPods – you can use this when you want to sync everything

## How did I test?
Ran `pnpm run deps` once, checked in changes; then ran `pnpm run deps` again, and saw that `Podfile.lock` didn't change.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

git revert


dnbrwstr (approved):
Lgtm!

davidisaaclee:
fixes TLON-4318

PR #4875, opened 2025-06-20
fix sheet tap on android

dnbrwstr:
## Summary

Fixes TLON-4400 by replacing `Sheet.ScrollView` with `ScrollView` on Android.

## How did I test?

Tested on Android device.

## Risks and impact


patosullivan (approved):
Looks good. Tested locally, works as expected.

PR #4876, opened 2025-06-20
Fix infinite render loop in BigInput

patosullivan:
## Summary

The `BigInput` component was misbehaving because of an infinite render loop. This was most apparent on Android where the `Post` button was never enabled. I traced the issue to useRegisterChannelHeaderItem, the handleSend callback had many dependencies that changed frequently, causing the header button to be recreated on every render, which updated the header context and triggered re-renders.

## Changes

I implemented a stable callback pattern:

- Moved dynamic values (title, imageUri, editingPost, attachments) to refs instead of callback dependencies
- Created stableHandleSend that never changes but accesses current values through refs
- Reduced handleSend dependencies to only stable values

This breaks the dependency chain that was causing the render loop while maintaining all functionality.

## How did I test?
Tested on web, iOS simulator, and Android emulator.

## Risks and impact

- Safe to rollback without consulting PR author? No, the component will be broken again.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] None of the above.

## Rollback plan

Revert the merge commit.

### DIFF
--- a/packages/app/ui/components/BigInput.tsx
+++ b/packages/app/ui/components/BigInput.tsx
@@ -63,15 +63,28 @@ export function BigInput({
   const [isEmpty, setIsEmpty] = useState(true);
   const { attachments, clearAttachments } = useAttachmentContext();
 
+  const editingPostContent = useMemo(
+    () => editingPost?.content,
+    [editingPost?.id, editingPost?.content]
+  );
+
+  const currentValuesRef = useRef({
+    title,
+    imageUri,
+    editingPost,
+    attachments,
+  });
+
   const handleEditorContentChanged = useCallback(
     (content?: object) => {
+      const { attachments } = currentValuesRef.current;
       const hasAttachmentsAndIsGallery =
         attachments.length > 0 && channelType === 'gallery';
       const nextIsEmpty =
         contentIsEmpty(content) && !hasAttachmentsAndIsGallery;
       logger.log('Content empty check:', nextIsEmpty);
-      if (content && editingPost?.content) {
-        const originalContent = editingPost.content as { story: any };
+      if (content && editingPostContent) {
+        const originalContent = editingPostContent as { story: any };
         const inlines = tiptap.JSONToInlines(content);
         const story = constructStory(inlines);
         const hasChanges =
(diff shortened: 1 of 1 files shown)


davidisaaclee:
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md might be worth adding – I know I'm guilty of passing anonymous functions to props because it's quick and easy, but I don't think there's any case where an anonymous function is better than a `useCallback`'d one

patosullivan:
> I imagine this is being caused by some function prop higher up being constructed on every render. If that's the case, then https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md might be worth adding – I know I'm guilty of passing anonymous functions to props because it's quick and easy, but I don't think there's any case where an anonymous function is better than a `useCallback`'d one

Definitely agree. Unfortunately we've got ~327 places where were doing this in the app package, so probably out of scope for this PR to go through and fix all of those after adding the new rule. We should definitely add this rule and do this soon, though.

packages/app/ui/components/BigInput.tsx
```diff
@@ -116,14 +139,21 @@ export function BigInput({
...
-  const handleSend = useCallback(async () => {
+  const handleSend = useMutableCallback(async () => {
     if (!editorRef.current?.editor) return;
 
+    const {
+      title: currentTitle,
+      imageUri: currentImageUri,
+      editingPost: currentEditingPost,
+      attachments: currentAttachments,
+    } = currentValuesRef.current;
```
davidisaaclee, line 150: We shouldn't need to do any of the ref stuff if we're using `useMutableCallback` – to reduce noise, I think we should remove the ref stuff and just access `title`/`imageUri` etc directly

packages/app/ui/components/BigInput.tsx
```diff
@@ -63,15 +63,28 @@ export function BigInput({
   const [isEmpty, setIsEmpty] = useState(true);
   const { attachments, clearAttachments } = useAttachmentContext();
 
+  const editingPostContent = useMemo(
+    () => editingPost?.content,
+    [editingPost?.id, editingPost?.content]
+  );
```
davidisaaclee, line 69: what's the point of this? was there a real issue where `id` would change but `content` wouldn't? (in that case, I think `editingPostContent` wouldn't change identity either, since it's just a substitution for `editingPost?.content`, which, as said before, didn't change)

patosullivan:
> Tracing this through with @dnbrwstr it seems like just wrapping the anonymous definition of `setShowBigInput` in `GalleryInput` seems to stop it from looping.
> 
> Agreed with @davidisaaclee we should probably make a tech debt issue to fix all those instances and setup the rule.

yep, that did it. updated with just that change. glad you found that!

PR #4886, opened 2025-06-26
bait: allow GET requests from any cross-origin

Fang-:
## Summary

Make bait agent serve appropriate CORS-related headers when responding to GET requests.

## Changes

Since we want to pull lure link metadata from browser contexts, and theres no problem with the bait/lure service allowing this kind of access, we must include an Access-Control-Allow-Origin header to appease browser security mechanism.

Here, we do so for every GET request handled by the bait agent.

I'm assuming we only care about [the simple requests case](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#simple_requests), where we _never_ send credentials. We shouldn't need to, so we shouldn't.

## How did I test?

It's on loshut-lonreg right now! You can see the header being returned correctly. @latter-bolden take her for a spin, see how she drives.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

`git revert afc64e5`

## Screenshots / videos

image

### DIFF
--- a/desk/app/bait.hoon
+++ b/desk/app/bait.hoon
@@ -114,6 +114,19 @@
     ?+    method.request  (give not-found:gen:server)
       %'GET'  (get-request line)
     ::
+        %'OPTIONS'
+      %-  give
+      =;  =header-list:http
+        [[204 header-list] ~]
+      :~  :-  'access-control-allow-methods'
+          =-  (fall - '*')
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

PR #4887, opened 2025-06-26
mobile: fix certain emojis

latter-bolden:
## Summary

Our regex for detecting native emojis wasn't comprehensive enough to cover all cases of mobile input. The result was displaying our fallback _stop sign_ emoji instead of the desired one.

We've tried custom regex's (including `\p{Emoji_Presentation}`) as well as libraries without finding a solution that covers all cases. The change in this PR is to instead default to displaying whatever's passed in, so long as it's under a character count threshold.

Since this biases towards showing whatever is received,  we may end up displaying non-emoji short strings if users manually insert them. However the UI will only ever permit actual emojis to be sent, so you'd have to issue pokes manually to abuse this.

## How did I test?

- tested with the violating emoji before and after

## Risks and impact

- Affects important code area: message display

### DIFF
--- a/apps/tlon-web/src/logic/utils.ts
+++ b/apps/tlon-web/src/logic/utils.ts
@@ -38,7 +38,6 @@ import anyAscii from 'any-ascii';
 import bigInt, { BigInteger } from 'big-integer';
 import { hsla, parseToHsla, parseToRgba } from 'color2k';
 import { differenceInDays, endOfToday, format } from 'date-fns';
-import emojiRegex from 'emoji-regex';
 import _ from 'lodash';
 import { useCallback, useMemo, useRef, useState } from 'react';
 import ob from 'urbit-ob';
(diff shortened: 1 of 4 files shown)


arthyn:
May want to check `packages/shared/src/logic/utils.ts` for `containsOnlyEmoji` believe this was using the same regex dep.

latter-bolden:
>  found this yesterday: https://www.unicode.org/reports/tr51/#def_rgi_set 

interesting, I think this is worth trying. I'd much rather have a regex with no false negatives if possible

arthyn:
> > May want to check packages/shared/src/logic/utils.ts for containsOnlyEmoji believe this was using the same regex dep.
> 
> Yeah I noticed this. Don't think it appears to be working as is?

it's used in `postContent.ts` to determine if we should show a big emoji if it's by itself

patosullivan (approved):
Lgtm, fixes the "stopsign" issue. Let's rip the band-aid off.

PR #4888, opened 2025-06-27
chat-message: fixing cursor

arthyn:
## Summary

Fixes TLON-3973 by making sure we have `cursor="default"` in places where we're wrapping a multi-faceted item like `ChatMessage` in a `Pressable` to make sure we can attach things like long presses. The pointer cursor should be reserved for actual buttons. 

## Changes

- made Pressable carry any cursor settings through
- replaced `none` with `default` because this was making the cursor disappear
- applied `cursor="default"` to `BlockRenderer` and `ReactionsDisplay` in places where it was incorrect to have the value be `"pointer"`

## How did I test?

Manually tested with pnpm dev hovering over chat messages to make sure we didn't see places with pointer where it should be default.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/packages/app/ui/components/ChatMessage/ChatMessage.tsx
+++ b/packages/app/ui/components/ChatMessage/ChatMessage.tsx
@@ -174,7 +174,7 @@ const ChatMessage = ({
       onHoverIn={handleHoverIn}
       onHoverOut={handleHoverOut}
       pressStyle="unset"
-      cursor="none"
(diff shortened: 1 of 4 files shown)


patosullivan (approved):
Looks good, makes sense

PR #4889, opened 2025-06-27
channels: sequence number recovery

Fang-:
## Summary

Recent channels protocol upgrade added sequence numbers. This had two problems:
1. The way the migration was implemented caused tombstones for deleted messages to be dropped from state.
2. Channels agent is not guaranteed to have the full message backlog, and so might come up with different sequence numbers than the host.

This PR attempts to post-hoc fix those problem.

## Changes

9da6784 makes the channels-server recover deleted messages from the log: it inserts the tombstones back into the `posts` map and sends out `%channel-update` facts to all subscribers. This solves (1).

a68a2f4 makes channels agent ask the host of each channel it knows about to send back the sequence number for each message. If the host hasn't updated yet, it will nack, and channels agent will retry later. If it understands the request, it will poke back with a list of `id-post`-sequence-number pairs. The channels agent will apply this to the messages it has in state. This solves (2).

1eaab50 attempts to make sure that subscribers always have the same `count` on a channel that the publisher does, and respects the highest number it's heard from the host (as opposed to blindly incrementing the number like we do now).

1442fab fixes the migration logic that caused (1). This way, the "host re-sends the tombstones" fix doesn't need to be re-run for subscribers that are late to the party. (REVIEW: Can we _guarantee_ subscribers retain their tombstones? Unclear. Certainly it'd require another "negotiation" mechanism, which is complexity that might be greater than the size of the gap it'd fill.)

## How did I test?

Ran the bad migration, observed dropped state. Ran the migration from this PR, observed deleted message tombstones returned in both agents.

Have not yet tested over the network, have not yet tested with partial backlog (for which the sequence number negotiation part is important).

## Risks and impact

The sequence number "negotiation" in a68a2f4 feels a tad risky, especially when imagining a ship that is in a ton of channels. The perceived risk is around processing load. To mitigate, we spread these out over 15 minutes, and wait at least an hour between retries. Individual negotiations shouldn't be horrible, but we can and should try to pull some numbers.

- Not safe to rollback without consulting PR author.
  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

We actually don't change the state type, so rolling back would be easy (but doesn't work out-the-box due to state version tag).

## Screenshots / videos

Nah.

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -522,13 +522,15 @@
         [~ %| *]  (run-import p.u.pimp)
       ==
     ::
-        [%sequence-numbers * *]
-      =+  !<([%sequence-numbers =nest:c seqs=(list [id=id-post:c seq=@ud])] vase)
+        [%sequence-numbers * @ *]
+      =+  !<([%sequence-numbers =nest:c count=@ud seqs=(list [id=id-post:c seq=@ud])] vase)
       ?>  =(src.bowl ship.nest)
       ?.  (~(has by v-channels) nest)  cor
       =.  v-channels
         %+  ~(jab by v-channels)  nest
         |=  channel=v-channel:c
+        =.  count.channel  count
+        |-
         ?~  seqs  channel
         =*  next  $(seqs t.seqs)
         ?~  p=(get:on-v-posts:c posts.channel id.i.seqs)  next
@@ -1612,7 +1614,7 @@
     =/  old  posts.channel
     =.  posts.channel
       ((uno:mo-v-posts:c posts.channel posts.chk) ca-apply-unit-post)
-    =.  count.channel  ~(wyt by posts.channel)
+    =.  count.channel  count.chk
     =?  ca-core  &(send !=(old posts.channel))
       %+  ca-response  %posts
       %+  gas:on-posts:c  *posts:c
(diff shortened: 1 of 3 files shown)


desk/app/channels.hoon
```diff
@@ -488,11 +515,30 @@
...
+        %+  ~(jab by v-channels)  nest
+        |=  channel=v-channel:c
+        =.  count.channel  count
+        |-
+        ?~  seqs  channel
+        =*  next  $(seqs t.seqs)
+        ?~  p=(get:on-v-posts:c posts.channel id.i.seqs)  next
+        ?~  u.p  next
+        =.  posts.channel
+          (put:on-v-posts:c posts.channel id.i.seqs u.p(seq.u seq.i.seqs))
```
mikolajpp, line 539: Perhaps it is just me, but I had to look three times before understanding the precedence in the `u.p(seq.u seq.i.seq)` that lends sense to `seq.u`.

desk/app/channels.hoon
```diff
@@ -995,6 +1053,13 @@
     =^  caz=(list card)  subs
       (~(handle-wakeup s [subs bowl]) pole)
     (emil caz)
+  ::
+      [%numbers kind=?(%chat %diary %heap) ship=@ name=@ ~]
+    =/  host=ship  (slav %p ship.pole)
+    =/  =nest:c    [kind.pole host name.pole]
+    %-  emit
+    =/  =cage  [%noun !>([%send-sequence-numbers nest])]
+    [%pass pole %agent [host %channels-server] %poke cage]
```
mikolajpp, line 1062: Reusing poles (no pun intended) to execute a sequence while avoiding possible wiring mistakes is really nice. Adding to my hoon toolbox.

mikolajpp (approved):
I am not sure if this is what PR description alluded to, when it said _"Can we guarantee subscribers retain their tombstones?"_ 
It looks like, while we make sure subscribers have correct sequence numbers and post count, the tombstones would only be filled in if the update send out when the host upgrades reaches them. If the subscriber loses the subscription and then resubscribes it would not receive tombstoned posts, since the subscription query will not reach beyond the point necessary to see these. Is that correct?

Everything else looks good, many thanks for fixing this!

PR #4890, opened 2025-06-27
android: adding posthog logging and fallback notification

arthyn:
## Summary

Fixes TLON-3614. This adds PostHog logging to our notification flow so that we record any errors we encounter as well as tracking successful deliveries.

## Changes

- added PostHog to native android as dep
- added PostHog initialization to notification service
- created a notification logger helper object
- modified `NotificationManager` to include better logging
- converted `TalkMessagingService` to Kotlin

## How did I test?

Hooked up my device to android studio debugger and sent arbitrary requests to our Twilio endpoint that I knew would fail. 

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [X] Notifications


apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/NotificationManager.kt
```diff
@@ -125,22 +188,75 @@ suspend fun processNotification(context: Context, uid: String) {
...
+
+    // Extract basic info from original payload or use defaults
+    val title = originalPayload?.getString("title") ?: "New message"
+    val body = originalPayload?.getString("body") ?: "You have a new message"
+
+    val extras = Bundle()
+    if (activityEvent != null) {
+        extras.putString("activityEventJsonString", activityEvent.toString())
+    } else if (originalPayload != null) {
+        extras.putAll(originalPayload)
```
davidisaaclee, line 221: I don't think putting `originalPayload` on `extras` is necessary (and since we've been historically ducktyping this payload, it'd be good to avoid adding more stuff here)

apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/NotificationManager.kt
```diff
@@ -125,22 +188,75 @@ suspend fun processNotification(context: Context, uid: String) {
...
+    }
+    extras.putString("fallbackReason", reason)
+
+    val tapIntent = Intent(context, MainActivity::class.java)
```
davidisaaclee, line 225: seems like you could DRY this with the corresponding code in showRichNotification

apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/TalkMessagingService.kt
```diff
@@ -0,0 +1,76 @@
...
+     */
+    private fun ensurePostHogInitialized() {
+        synchronized(initLock) {
+            if (!isPostHogInitialized) {
+                try {
+                    val cs = ConstantsService(this)
+                    val constants = JSONObject(cs.constants["manifest"] as String);
+                    Log.d(TALK_MESSAGING_SERVICE, constants.toString())
+                    val key = constants.getJSONObject("extra").getString("postHogApiKey");
+                    Log.d(TALK_MESSAGING_SERVICE, key)
```
davidisaaclee, line 35: I don't think we need this log if we're already logging all of the constants above
```suggestion
```

apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/NotificationManager.kt
```diff
@@ -29,23 +31,79 @@ import kotlin.coroutines.suspendCoroutine
...
+                override fun onComplete(response: JSONObject?) = cnt.resume(response)
+                override fun onError(error: VolleyError?) = cnt.resumeWithException(error ?: Error("Unknown error"))
+            })
+        }
+    } catch (e: Exception) {
+        val message = "Activity event fetch failed"
+        NotificationLogger.logError(getLogPayload(uid, message, e))
+        // Fall back to basic notification with original payload
+        showFallbackNotification(context, uid, originalPayload, message)
+        return
```
davidisaaclee, line 68: nit: I think it'd be a little tidier to move `showFallbackNotification` up a level (either to `TalkMessagingService#onMessageReceived` or by moving the current body of `processNotification` to a private helper and doing a small try/catch in `processNotification`) – e.g. wherever we're currently doing `showFallbackNotification`, throw a custom exception with any necessary info (e.g. `ActivityEventFetchFailed()` or `AcitivtyEventPreviewRenderFailed(activityEvent)`), and catch at the caller (where we'd also have access to the `originalPayload`, so you wouldn't need to pass it down here).
- it'd avoid passing `originalPayload` here unnecessarily
- we'd have the error handling all in one place
- it'd keep the main logic of `processNotification` cleaner (this is a bad place to have a bug)

apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/NotificationManager.kt
```diff
@@ -29,23 +31,79 @@ import kotlin.coroutines.suspendCoroutine
...
+        showRichNotification(context, uid, preview, activityEvent)
+        NotificationLogger.logDelivery(mapOf("uid" to uid, "message" to "Rich notification delivered"))
+    } catch (e: Exception) {
+        val message = "Rich notification display failed"
+        NotificationLogger.logError(getLogPayload(uid, message, e))
+        showFallbackNotification(context, uid, originalPayload, message, activityEvent)
+    }
+}
+
+@SuppressLint("MissingPermission")
```
davidisaaclee, line 105: iirc, this was an issue before this PR, but if you have the error message handy, could you add a comment about why the lint would be triggered / why we're suppressing it?

apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/NotificationManager.kt
davidisaaclee: I still think you could share code with above! a nice way of doing it would be extending `NotificationCompat.Builder`, something like

```kt
private suspend fun NotificationCompat.Builder.buildMessagingTappable(extras: Bundle): NotificationCompat.Builder {
  // ...
}
```

davidisaaclee (approved):
some optional nitpicks, otherwise lgtm

apps/tlon-mobile/android/app/src/main/java/io/tlon/landscape/notifications/NotificationManager.kt
arthyn: I can yeah, it seemed like it was having trouble seeing that we already checked for permissions since the check happened in a function above it

PR #4892, opened 2025-06-30
desk: update egg-any version

pkova:
urbit/urbit#7181 was recently merged to fix a long-standing Gall bug. In the PR we wrote a Gall migration that did not change the type of the state but modified the `blocked` queue somewhat. We decided to increment the `egg-any` version also to keep it in sync with the Gall state version number, even though in this case it isn't strictly necessary. The state of Gall %16 and %17 is exactly the same.

All `egg-any` version bumps are technically breaking changes so in an ideal world that Gall PR should go out with a kelvin release. In practice I think that the amount of desks using the `egg-any` backup-and-restore feature is so small that we can get away with releasing the Gall fix in a 410 bugfix release, synchronizing it in such a way that this groups PR has been released beforehand. This allows us to release a fix for Gall bug as soon as possible.

### DIFF
--- a/desk/app/activity.hoon
+++ b/desk/app/activity.hoon
@@ -334,7 +334,7 @@
       %egg-any
     =+  !<(=egg-any:gall vase)
     ?-  -.egg-any
-        ?(%15 %16)
+        ?(%15 %16 %17)
       ?.  ?=(%live +<.egg-any)
         ~&  [dap.bowl %egg-any-not-live]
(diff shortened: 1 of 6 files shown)


Fang- (changes requested):
As discussed during core blitz today, if they're exactly the same (shape-wise and semantics-wise) we just shouldn't bump the version label on this type.

PR #4893, opened 2025-06-30
settings: disable options menu on group/channel list items in push notification settings

patosullivan:
## Summary

We use GroupListItem and ChannelListItem to display groups or channels that follow a different push notification rule than the rule the user has set at the top level. Both of these components show an options menu by default unless we pass in `disableOptions`. Since that prop wasn't passed in, the options menu button was displayed on top of the "close" icon.

fixes tlon-4293

## Changes

We pass the `disableOptions` prop on both of those components.

## How did I test?

Tested on web, the only place this problem appears.

## Risks and impact

- Safe to rollback without consulting PR author? Yes, but the bug will come back.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Revert the merge commit.

## Screenshots / videos

image

### DIFF
--- a/packages/app/features/settings/PushNotificationSettingsScreen.tsx
+++ b/packages/app/features/settings/PushNotificationSettingsScreen.tsx
@@ -188,6 +188,7 @@ export function ExceptionsDisplay({
                 />
               }
               paddingVertical="$s"
+              disableOptions
             />
           );
         })}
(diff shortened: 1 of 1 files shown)


dnbrwstr (approved):
Lgtm!

PR #4897, opened 2025-06-30
UI: cosmetic omnibus

jamesacklin:
## Summary

Makes several cosmetic changes to the app.

## Changes

- Adds an `autoFocus` prop to `SearchBar`, then drills the value down from the create chat sheet, through `ContactBook`
- Bumps the max unread counter to 256 from 99
- Adds a background color to the otherwise conditionally invisible message actions trigger button on Notebook posts
- Positively sets the text input color of `BareChatInput` and the comment input in `GalleryInput`
- Replaces all instances of the term "TM"
- Adds a prefix to edit and delete message actions if you are a group administrator and you are operating on someone else's post
- In `BlockRenderer`, accesses the content context and passes in the `isNotice` prop to restore styling of system notices ("foo joined the chat")
- Removes the title-flipping on the Home screen when the user taps between tabs (duplicitous and results in a half-second delay when the text changes)

## How did I test?

I almost exclusively used existing fixtures.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

git revert


dnbrwstr (approved):
Lgtm! I think the `Admin:` thing is a little odd, but not blocking.

PR #4918, opened 2025-07-09
chat: rip out old pins which are no longer used

arthyn:
## Summary

We were checking for behavior we no longer support or use, and it was causing failures. This just rips it out instead.

## Changes

- removed /pins scry in %chat
- removed lib/discipline checks for %chat-pins mark

## How did I test?

Compiling the hoon

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -19,7 +19,6 @@
 /%  m-chat-heads           %chat-heads
 /%  m-chat-heads-1         %chat-heads-1
 /%  m-chat-paged-writs     %chat-paged-writs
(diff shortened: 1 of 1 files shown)


Fang- (approved):
My fakezods all somehow have `/mar/chat/pins.hoon`. I guess it's in the pill? Caused me to not get a build failure on this.

We may want to remove handling for the `%chat-pins` poke also, to match these, but good either way. Thank you!

PR #4920, opened 2025-07-10
queries, postUtils, ChatMessage: consistent nickname display in reaction tooltip

jamesacklin:
## Summary

Fixes TLON-4536 where we weren't querying for contact details along with reactions to posts, which caused an issue where nicknames occasionally would fail to display in the reaction tooltip on desktop. Furthermore, this adds an additional step to the nickname computation to show overridden nicknames.

## Changes

- Modifies the getThreadPosts, getChannelPostsAround (beforePosts, afterPosts), getChannelSearchResults, and getPostWithRelations queries to also include contact details (root cause)
- Moves the more complex logic for useReactionDetails inside its own function, which is still memoized and returned in the parent hook
- Computes the "correct" nickname to show in the reaction details following the override/user-set/bare-patp priority chain
- Defensively trims the list to 3 reactors as before

## How did I test?

- Navigate into a channel on desktop using either develop or production.
- Do not click on anyone's profile, do not load any other data, etc.
- Hover over an emoji reaction a few people have added to a message.
- Note that you will not see their nickname -- or, you might. Sometimes!
- Apply the fix
- Repeat the process above
- Note you will now see their nicknames displayed in the tooltip
- Add one of the reactors as a Contact and set a custom nickname for them
- Navigate back to the channel and hover back over the reaction
- Note you will now see their overridden nickname displayed in the tooltip

I added a unit test to ensure the logic for displaying the correct nickname is sound. 

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

git revert

### DIFF
--- a/packages/app/ui/components/ChatMessage/ReactionsDisplay.tsx
+++ b/packages/app/ui/components/ChatMessage/ReactionsDisplay.tsx
@@ -59,19 +59,23 @@ export function ReactionsDisplay({
     ]
   );
 
-  const firstThreeReactionUsers = useCallback(
-    (reaction: ReactionListItem) =>
-      reaction.users
-        ? reaction.users
-            .slice(0, 3)
-            .map((user) => user.name || user.id)
-            .join(', ') +
-          (reaction.users.length > 3
-            ? ` +${reaction.users.length - 3} more`
-            : '')
-        : '',
-    []
-  );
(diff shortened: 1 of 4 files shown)


patosullivan (approved):
Looks great, changes make sense. Tested locally, works as expected. Glad to see some new tests.

PR #4922, opened 2025-07-11
web: stabilize navigation when switching between narrow and wide navigators

davidisaaclee:
## Summary
fixes TLON-4227
Fixes issue where changing browser width would lose place in app (or, at least fixes this at key routes).

## Changes
We dynamically swap navigators based on the browser width: https://github.com/tloncorp/tlon-apps/blob/c621eff06e0e2b0ab10e2e138264f55ae9fc0df8/packages/app/navigation/BasePathNavigator.tsx#L39-L41

When the navigator is swapped, we lose the navigation state (current location + history). We can't simply move the navigation state to the new navigator, since the navigators have different structures that require different states.[^0] We can't even pick out the focused route's name + params and navigate to that in the new navigator, since we have some patterns where the focused routes are not isomorphic between the two navigators (e.g. `ChannelRoot` vs `Channel`).

To have an apparently stable state when swapping navigators, this PR takes the current navigator's state and derives a minimal `initialState` for the other navigator, so that when we swap navigators, that `initialState` makes the new navigator start where the other one left off.

To convert between state shapes, this PR adds a new type `NavigationIntent`, a succinct Tlon-specific value that describes the feature we're navigated to (e.g. `{feature: chat-list`}`). There are methods to convert a `NavigationState` for either mobile or desktop navigator to a `NavigationIntent`, and to convert a `NavigationIntent` back into a `NavigationState` for the specified navigator.

> A limitation of this approach is that when we swap navigators, **_we lose the history encoded in the previous `NavigationState`_**, meaning in-app back buttons will break, but browser history will not. It'd be difficult to retain this state, since we'd need to convert every route (as they are not necessarily compatible between routers). I think it's bad to have this behavior, but I don't think it's worth the effort to fix this before unifying the navigators.

[^0]: Towards the end of implementation, I learned that you can define a [`getRehydratedState()`](https://reactnavigation.org/docs/custom-routers/) in a custom navigator. If we wrote a custom navigator which wrapped the two inner navigators, we could use this to intelligently convert one navigator's state to another, and get rid of all of this external control code. (Not sure if this would be much better, but it'd at least be more compartmentalized.)

## How did I test?
I tested by opening the following routes and rapidly resizing browser to trigger layout change 4 times in succession:
- Chat list / home
- Channel
- Drilled down into a notebook post
- Settings (and child settings pages in desktop)
- Activity (does not maintain position)

I also added some unit tests of `NavigationIntent`, although I'm not sure how much they prove that usage will work, since there's nothing checking that the paths we're supplying in tests match what the real router will generate.

To enable these unit tests in `app`, I had to add a mock for react-native.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

git revert

## Screenshots / videos

https://github.com/user-attachments/assets/bad24d86-ed74-4af9-b78a-7d93affaa2ad


packages/app/navigation/linking.ts
```diff
@@ -26,6 +26,7 @@ export const getMobileLinkingConfig = (
             path: 'group/:groupId/channel/:channelId/:selectedPostId?',
             parse: parsePathParams('channelId', 'groupId', 'selectedPostId'),
           },
+          ChatList: 'ChatList',
```
davidisaaclee, line 29: Without explicit entries here, `getStateForPath` fails. I added entries where I needed them, and took the liberty of lowercasing some of them to look more URL-y.

davidisaaclee:
> it can't keep state when switching from desktop to mobile while looking at a thread or gallery/notebook post screen.

fixed via [Add post NavigationIntent for threads or gallery/notebook items](https://github.com/tloncorp/tlon-apps/pull/4922/commits/f9f772ea77cf8c63a001d407fd248298b4168f21)

> it can't load the gallery post screen at all

this was an existing bug due to react-native-gesture-handler being loaded via gallery carousel on web; fixed by never using Carousel when `Platform.OS === 'web'` [Avoid using carousel on web (avoids RNGH)](https://github.com/tloncorp/tlon-apps/pull/4922/commits/ea6f2f5b3ed101055d6ccc922cfa69d12f80f6e0)

packages/app/navigation/intent.ts
```diff
@@ -187,9 +214,15 @@ export function getStateFromNavigationIntent(
 }
 
 // stubs to be replaced by https://github.com/tloncorp/tlon-apps/pull/4803
+function getAuthorId(route: any): string {
```
davidisaaclee, line 217: @patosullivan think we should just leave these in until #4803 is merged? I would be just as happy to inline them, but I don't want to do that if we're just going to go back to `get___` later.

patosullivan: Works for me

patosullivan (approved):
Changes look good, earlier issues were fixed!

packages/app/navigation/intent.ts
```diff
@@ -0,0 +1,272 @@
...
+export function getStateFromNavigationIntent(
+  intent: NavigationIntent,
+  navigatorType: 'mobile' | 'desktop'
+): PartialState<NavigationState<CombinedParamList>> | null {
+  const url = pathFromNavigationIntent(intent, navigatorType);
+  if (url == null) {
+    return null;
+  }
+
+  // @ts-expect-error - `getStateFromPath` from react-navigation uses a
```
latter-bolden, line 209: My TS checker doesn't like this, I still get the warnings on the `getStateFromPath` calls. Looks fine on your end?

davidisaaclee: I'm getting no errors in my code editor, and `pnpm run -r tsc` doesn't show me any errors.

I'm seeing that we aren't consistently specifying the `typescript` dependency version (we specify it in `overrides` at workspace root, and as a `devDependency` in `shared` – funnily, these two versions don't match). I added typescript@5.4.5 to root `package.json` dependencies, and I still have no errors in editor nor tsc – maybe give that a shot? I'd like to figure out why the discrepancy or at least get a third data point.

```diff
diff --git a/package.json b/package.json
index 9abb7b708..a0dbde906 100644
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "lint:all": "pnpm -r lint"
   },
   "dependencies": {
-    "@tloncorp/ui": "workspace:*"
+    "@tloncorp/ui": "workspace:*",
+      "typescript": "5.4.5"
   },
   "devDependencies": {
     "@trivago/prettier-plugin-sort-imports": "^4.3.0",
```

PR #4923, opened 2025-07-11
threads: fix thread mute issue

patosullivan:
## Summary

fixes tlon-4472

Fixes an issue where the "mute thread" action on a message never worked.

## Changes

We had been passing the child post as the `source` for the volume setting adjustment, rather than the parent. The fix is just grabbing the parent post and using that as the `source` instead. If we don't have a parent post ID, then we are the parent, and we pass the post itself instead (we currently don't allow muting the root/parent post in the UI, but this enables us to allow that in the future if we want).

## How did I test?

Tested on web.

## Risks and impact

## Rollback plan

Revert


jamesacklin (approved):
Seems like we'd need the parent, yes

PR #4924, opened 2025-07-11
gallery: fix issue with editing gallery link blocks

patosullivan:
## Summary
fixes tlon-4443

We had an issue where if a user posted a link block in a gallery and went back to edit it the link would disappear after their edit. We have a `useEffect` in the `GalleryInput` component where we try to determine what the post type is and route the user to the correct input. We weren't accounting for links there.

## Changes

Updates that `useEffect` to account for link blocks, adds a `mockAttachment` so we can re-attach the link.

## How did I test?

Tested on web.

## Risks and impact

## Rollback plan

Revert

## Screenshots / videos

https://github.com/user-attachments/assets/d4ac8773-3110-465d-b77e-a9ec78b8f569


jamesacklin (approved):
Looks great, simple enough!

PR #4925, opened 2025-07-11
client: sequence number integration

latter-bolden:
Sequence numbers are in place and available for DMs and group channels. At the moment, only channels hosted by `wannec` ships will be assured to not have gaps.

Still fine tuning `useChannelPosts`, some reversed ordering of post pages and edge cases around beginnings of channels.

Fixes TLON-3572


latter-bolden:
Going to branch off this known good point for the deleted post workarounds. Closing this PR in the meantime

PR #4926, opened 2025-07-12
MessageActions: prevent users from hiding their own posts

jamesacklin:
## Summary

Fixes TLON-4554 where one could "hide" their own message without being able to unhide, resulting in an unrecoverable state.

## Changes

- Adds a conditional to our "which message actions to show when" chain to cover this
- Adds a unit test to ensure one can't hide their own posts
- Changes the chat functionality e2e tests to use ~zod and ~ten (no version mismatch)
- Introduces a new step to the e2e tests in which ~zod authors a message, switches contexts to ~ten, and hides the message on ~ten
- Fixes a race condition in the e2e tests where hovering over a chat message would fail because it would incorrectly target the sidebar preview text rather than the message itself in the channel
- Adds a navigateToChannel helper for e2e tests
- Fixes the single-test running issue for e2e

## How did I test?

1. Author a post
2. Long-press; observe "Hide post" is not visible
3. Long-press on another's post; observe "Hide post" is visible

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

git revert

### DIFF
--- a/apps/tlon-web/e2e/README.md
+++ b/apps/tlon-web/e2e/README.md
@@ -157,9 +157,6 @@ test('test description', async ({ page }) => {
     // Navigate to app
     await page.goto(`${shipManifest['~zod'].webUrl}/apps/groups/`);
 
-    // Handle welcome flow
-    await helpers.clickThroughWelcome(page);
-
     // Test-specific logic using helpers
     await helpers.createGroup(page);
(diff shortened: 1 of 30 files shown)


jamesacklin:
@patosullivan Beautiful, thank you

PR #4933, opened 2025-07-14
grouper: make sure we use the right mark and handle logging correctly

arthyn:
## Summary

This fixes an issue found where newly joining ships were not getting a DM invite from their inviter. This was caused by a mark/type mismatch. Additionally logging was lacking here so I fixed that too.
 
## Changes

- change `chat-dm-action` to `chat-dm-action-1` to match the type we were sending
- move 'DM Invite Sent' event to only send on successful poke to ourselves
- add 'DM Invite Fail' event when we get a poke-nack

## How did I test?

Ran changes locally, used personal invite lure and sent an HTTP request to simulate a bite.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/app/grouper.hoon
+++ b/desk/app/grouper.hoon
@@ -131,7 +131,6 @@
       ?~  p.sign  `this
       %-  (slog leaf/"Poke failed {<wire>}" u.p.sign)
       `this
-    ?>  ?=([%dm ship=@ token=@ ~] wire)
     =*  joiner  i.t.wire
     =*  token  i.t.t.wire
     :_  this
--- a/desk/app/grouper.hoon
+++ b/desk/app/grouper.hoon
@@ -126,8 +126,34 @@
       `this(outstanding-pokes (~(del in outstanding-pokes) [src.bowl i.t.t.wire]))
     ==
   ?-  -.sign
-      %poke-ack   `this
+      %poke-ack
+    ?.  ?=([%dm ship=@ token=@ ~] wire)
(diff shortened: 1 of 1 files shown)


desk/app/grouper.hoon
```diff
@@ -126,8 +126,34 @@
       `this(outstanding-pokes (~(del in outstanding-pokes) [src.bowl i.t.t.wire]))
     ==
   ?-  -.sign
-      %poke-ack   `this
+      %poke-ack
+    ?.  ?=([%dm ship=@ token=@ ~] wire)
+      ?~  p.sign  `this
+      %-  (slog leaf/"Poke failed {<wire>}" u.p.sign)
+      `this
+    ?>  ?=([%dm ship=@ token=@ ~] wire)
```
Fang-, line 134: The `?>` is superfluous, since the `?.` already checks for the same condition, no?

Fang- (approved):
Was going to point out that removing logging from the _sending_ of the poke means we might delay logging indefinitely until we hear back, but in this case we poke a local agent, so we should hear back within the same event. This is fine.

Was also going to point out maybe needing to handle old-style wires for that poke-ack, but that's not necessary for the same reason: we won't have any pokes outstanding.

One small inconsequential nit, but otherwise seems fine and good to me. Wondering if we would've caught this earlier if /lib/discipline also checked mark-data compat for outgoing pokes...

PR #4935, opened 2025-07-15
groups: change to correct type for take-contact

arthyn:
## Summary

OTT, we've had the wrong type since at least https://github.com/tloncorp/tlon-apps/commit/090064947280d4d2e342227783e63598c4356b8c 

## Changes

- changed update-0 to news-0

## How did I test?

Manually changing profile from one ship and watching terminal from listening ship

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -1157,10 +1157,10 @@
     cor
   ::
       %fact
-    =+  !<(=update-0:tac q.cage.sign)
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Looks like the correct type to me.

PR #4944, opened 2025-07-22
link input: fix URL validation issue

patosullivan:
## Summary

fixes tlon-4550.

Our `URL_REGEX` only checked if a string contained a valid URL *anywhere* in the string, not whether the string is *only* a valid URL. So a user could input any string into the URL input on gallery's LinkInput and it would be considered valid so long as that string had a URL within it. Adding the `^` and `$` anchors to the regex pattern fixes this.

The crash that Eleanor was seeing was caused by BlockRenderer attempting to render a link block where the URL was one of these invalid strings that contained URLs but weren't *just* URLs. We shouldn't crash in this case, we should just show a URL preview that tells the user the URL is invalid (and going forward, there shouldn't be anymore invalid URLs that we attempt to render in BlockRenderer).

## How did I test?

Tested on web.

## Risks and impact

## Rollback plan

Revert

## Screenshots / videos

Link blocks with invalid URLs will now render like this:
image


latter-bolden (approved):
Nice find

PR #4945, opened 2025-07-22
groups: fix foreign insertion

arthyn:
## Summary

We were crashing here because the foreigns didn't exist yet (jab crashes on key not existing)

## Changes

- added case for when we don't have the foreign in state

## How did I test?

Ship A creates a private group (that ship B has never seen). Ship B then clicks "Join group" from the plus menu, and pastes shortcode from ship A the clicks go. Should see a non-empty state.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications


mikolajpp (approved):
This works, thanks for fixing!

PR #4934, opened 2025-07-15
toast: improvements

jamesacklin:
## Summary

Fixes TLON-4531 where the toast message was undismissable and unreadable.

## Changes

- Properly cleans up useEffect to clear the 3-second toast timeout
- Adds a click-to-dismiss handler
- Restyles the toast to be readable and match the rest of the visual system
- Moves the toast well above the Chat input, rather than atop
- Adjusts the e2e "invite people" helper to account for the button on group settings + info as well as the long-press context menu
- Adds an e2e test to confirm that forwarding a chat message to a DM both (A) pops the toast notification and (B) successfully forwards the message.

## How did I test?

- Create a DM with another ship
- Create a group and invite the other ship to it
- Send a chat message
- Forward the chat message to the DM with the other ship
- Confirm that the toast message appears and dismisses itself
- Log in to the other ship and confirm you receive a DM with a reference to the forwarded message

Adds an e2e test to cover this and a helper function to assist with writing further forwarding tests.

## Risks and impact

- Safe to rollback without consulting PR author? Yes
- Affects important code area:
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

git revert

## Screenshots / videos

Before:
Screenshot 2025-07-14 at 4 13 30 PM

After:
Screenshot 2025-07-14 at 4 33 21 PM

PR #4949, opened 2025-07-23
toast: remove animation

jamesacklin:
## Summary

Removes the RN-reanimated animation from the Toast component, relying entirely on show-hide.

## Changes

The changes in #4934 resolved in dev mode for web and in device-built for mobile, but fail in production web (local and served from a ship). The toast notification fails to dismiss and the console shows an error (Uncaught TypeError: Cannot convert undefined or null to object) thrown from minified RN-reanimated.

I attempted to use worklet-safe handling with `useAnimatedStyle`, using `requestAnimationFrame`, and running RN-reanimated's timer on the JS thread. None of these appeared to resolve the issue. The safest thing I can think of is to remove the animation for now.

## How did I test?

- cd apps/tlon-web; npm build && npm serve
- Forward a post to a DM
- Watch the toast disappear

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

git revert

### DIFF
--- a/packages/ui/src/components/Toast.tsx
+++ b/packages/ui/src/components/Toast.tsx
@@ -6,12 +6,6 @@ import React, {
   useRef,
   useState,
 } from 'react';
-import Animated, {
-  Easing,
-  useAnimatedStyle,
-  useSharedValue,
-  withTiming,
-} from 'react-native-reanimated';
 import { Pressable } from 'react-native';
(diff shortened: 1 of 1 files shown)


dnbrwstr (approved):
Lgtm! Warrants further investigation at some point.

PR #4951, opened 2025-07-24
fix slow theme change

davidisaaclee:
fixes TLON-4109 (I think!)

## Summary
Changing between themes was very slow and would non-deterministically fail on some deep children, causing sticky wrong colors. This PR fixes the main issue, which was that we had two conflicting versions of Tamagui installed over each other; and also adds some smaller optimizations.

## Changes
- Bumps `@tamagui/react-native-media-driver` dependency to match other Tamagui versions. Having an older version of this dependency was pulling in a second Tamagui version, which was causing bad behavior on theme change.
- [Use color tokens directly instead of useTheme](https://github.com/tloncorp/tlon-apps/commit/9cafc3289b3bbfef96fa541ca74ad1c609d8154f) – when we use color tokens directly instead of a `useTheme,` it seems that Tamagui can be a little faster in theme updates (skipping a React render). I think it is a little easier to read as well.
- [Make all properties of ChatOptionsProvider optional](https://github.com/tloncorp/tlon-apps/commit/c9120469fe05c3db6cc138bbaaf4d84ba3428b82) – I was using this in a minimal repro and it was frustrating to need to supply callbacks that weren't strictly necessary.
- [Move app theme sync to hook](https://github.com/tloncorp/tlon-apps/commit/1624f668b57b55f0572329c2e0bf7a9af785e073) / [Simplify theme loading](https://github.com/tloncorp/tlon-apps/commit/56cf9e6abc1d975c95d4d9d3aa757750ec45b08d) – The big change here was that we were doing an unnecessary `syncSettings` every time dark mode or stored theme changed. I moved all the logic into a private hook in the first commit, and the second commit is the actual change that tries to remove unnecessary work on theme change.

## How did I test?
- Changed theme using settings and saw that it was faster
- Changed theme iOS simulator on same account, saw that the theme synced across network
- Set a custom theme locally, force quit app, re-opened – app opened with that selected theme
- Force quit iPhone app, set a custom theme on iOS simulator, re-opened iPhone app – app opened with old custom theme, but updated to new custom theme after a few seconds
- Resting on home screen (chat list), toggle dark mode on / off and saw that the change was fast and did not retain any straggling colors
- Added a debug floating button to toggle theme, toggled theme in a couple of places in the app, all fast(er) and no stragglers

<details>

<summary>Debug floating button diff</summary>

I forgot to stash the imports, sorry

```diff
diff --git a/packages/app/provider/index.tsx b/packages/app/provider/index.tsx
index f1bc93f38..eb81f2ad9 100644
--- a/packages/app/provider/index.tsx
+++ b/packages/app/provider/index.tsx
@@ -48,6 +48,17 @@ function ThemeProviderContent({
         defaultTheme={activeTheme}
       >
         {children}
+        <View style={{ position: 'absolute', left: 10, top: 100 }}>
+          <Button
+            onPress={() => {
+              setActiveTheme((prev) =>
+                prev === 'nord' ? 'solarized' : 'nord'
+              );
+            }}
+          >
+            <Text>Active Theme: {activeTheme}</Text>
+          </Button>
+        </View>
       </TamaguiProvider>
     </ThemeContext.Provider>
   );
```

</details>

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan
`git revert`

## Screenshots / videos

https://github.com/user-attachments/assets/9f9d59e3-66ef-4121-b019-b0fbb528b65a

### DIFF
--- a/packages/app/provider/index.tsx
+++ b/packages/app/provider/index.tsx
@@ -31,6 +31,50 @@ function ThemeProviderContent({
   children: React.ReactNode;
   tamaguiProps: Omit<TamaguiProviderProps, 'config'>;
 }) {
+  const [activeTheme, setActiveTheme] = useSyncedAppTheme();
+  return (
+    <ThemeContext.Provider value={{ setActiveTheme, activeTheme }}>
+      <TamaguiProvider
+        {...tamaguiProps}
+        config={config}
+        defaultTheme={activeTheme}
+      >
+        {children}
+      </TamaguiProvider>
+    </ThemeContext.Provider>
+  );
+}
+
+export const setTheme = async (
+  theme: AppTheme,
+  setActiveTheme: (theme: AppTheme) => void
+) => {
+  try {
+    setActiveTheme(theme);
+    await updateTheme(theme);
+  } catch (error) {
+    console.warn('Failed to save theme preference:', error);
+  }
+};
+
+export const clearTheme = async (
+  setActiveTheme: (theme: AppTheme) => void,
+  isDarkMode: boolean
+) => {
+  try {
+    setActiveTheme(isDarkMode ? 'dark' : 'light');
+    await updateTheme('auto');
+  } catch (error) {
+    console.warn('Failed to clear theme preference:', error);
+  }
+};
+
+export const useActiveTheme = () => {
+  const { activeTheme } = React.useContext(ThemeContext);
+  return activeTheme;
+};
+
+function useSyncedAppTheme() {
   const isDarkMode = useIsDarkMode();
   const [activeTheme, setActiveTheme] = useState<AppTheme>(
     isDarkMode ? 'dark' : 'light'
@@ -79,44 +123,5 @@ function ThemeProviderContent({
     }
   }, [isDarkMode, isLoading, storedTheme]);
 
-  return (
-    <ThemeContext.Provider value={{ setActiveTheme, activeTheme }}>
-      <TamaguiProvider
-        {...tamaguiProps}
-        config={config}
-        defaultTheme={activeTheme}
-      >
-        {children}
-      </TamaguiProvider>
-    </ThemeContext.Provider>
-  );
+  return [activeTheme, setActiveTheme] as const;
 }
-
-export const setTheme = async (
-  theme: AppTheme,
-  setActiveTheme: (theme: AppTheme) => void
-) => {
-  try {
-    setActiveTheme(theme);
-    await updateTheme(theme);
-  } catch (error) {
-    console.warn('Failed to save theme preference:', error);
-  }
-};
-
-export const clearTheme = async (
-  setActiveTheme: (theme: AppTheme) => void,
-  isDarkMode: boolean
-) => {
-  try {
-    setActiveTheme(isDarkMode ? 'dark' : 'light');
-    await updateTheme('auto');
-  } catch (error) {
-    console.warn('Failed to clear theme preference:', error);
-  }
-};
-
-export const useActiveTheme = () => {
-  const { activeTheme } = React.useContext(ThemeContext);
-  return activeTheme;
-};
--- a/packages/app/provider/index.tsx
+++ b/packages/app/provider/index.tsx
@@ -1,8 +1,7 @@
-import { syncSettings, updateTheme, useThemeSettings } from '@tloncorp/shared';
-import { subscribeToSettings } from '@tloncorp/shared/api';
-import { queryClient } from '@tloncorp/shared/api';
-import { themeSettings } from '@tloncorp/shared/db';
-import React, { useEffect, useState } from 'react';
+import * as store from '@tloncorp/shared';
+import * as api from '@tloncorp/shared/api';
+import * as db from '@tloncorp/shared/db';
+import React, { useEffect, useMemo, useState } from 'react';
 import { TamaguiProvider, TamaguiProviderProps } from 'tamagui';
 
 import { useIsDarkMode } from '../hooks/useIsDarkMode';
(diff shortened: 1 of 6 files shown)


pnpm-lock.yaml
```diff
@@ -14205,6 +14194,19 @@ snapshots:
       lru-cache: 5.1.1
       semver: 6.3.1
 
+  '@babel/helper-create-class-features-plugin@7.25.9(@babel/core@7.23.7)':
```
davidisaaclee, line 14197: I don't know why all of these `@babel/*` dependencies were added!

davidisaaclee:
I'm closing this while investigating why we now appear to have multiple versions of babel installed

davidisaaclee:
did some learning – afaict, we had two versions of `@babel/core` installed:
- 7.25.2 was the newer and more widely-used version across the workspace
- 7.23.7 was also being pulled in, and was causing lots of downstream dependencies to be duplicated as well (e.g. `react-native`) – we'd have one react-native with babel@7.23, and one with babel@7.25
- `pnpm why` showed that babel@7.23.7 was only being pulled in by `@vitejs/plugin-react` – however, the installed version of `@vitejs/plugin-react` lists `@babel/core` as requiring `^7.23.5`, which should allow minor changes, i.e. should accept `7.25.*`

So why didn't the `@babel/core@7.23.7` get "bumped up" to `7.25.*` in resolution? This explanation isn't proven but it appears to be how pnpm workspace resolution works:
- pnpm resolves each project's dependency tree independently
- then it merges the trees together into the lockfile
- pnpm avoids merging dependencies across dependency trees if it means changing the resolved versions – so even though `@vitejs/plugin-react` could accept the other dep tree's resolved version of `@babel/core`, it did not and just kept the one its tree (e.g. `editor` package) resolved
similar issue discussed here: https://github.com/pnpm/pnpm/issues/2443

To fix, I ran [`pnpm dedupe`](https://pnpm.io/cli/dedupe), which attempts to dedupe resolutions across workspace projects. This operation changes the installed versions of packages in a semver-compatible way – so this could theoretically cause breakages in our app if either (1) the packages we're pulling in are breaking semver, or (2) we are using unstable APIs from our dependencies.

davidisaaclee:
looks like that broke vitest – going to keep this PR open because the rest should be reviewable, and I think we could even return to having multiple versions of `@babel/core` if we can't find a better solution.

davidisaaclee:
- [Avoid importing from @react-navigation/native in test-covered code](https://github.com/tloncorp/tlon-apps/pull/4951/commits/d94bd091818a3f8c4b855e34c4c56b074f3e3f61) – minimal changes to import from `@react-navigation/core` instead of `@react-navigation/native` – unlike `/native`, `/core` exposes a commonjs `main`, so we can import it via Vitest with no issue
- [Remove unnecessary vitest.config.ts](https://github.com/tloncorp/tlon-apps/pull/4951/commits/8412df3ff70e0cc8b1bf5198f7c4a28c62328de5) – since we're no longer touching React Native in these tests, we can just get rid of this file

jamesacklin:
@davidisaaclee Yes! I updated the description of TLON-3432 and read your comments there.

We can use the appropriate light/dark splash screen based on the user's system theme preference; I don't think we need to dynamically compose the splash screen based on their Tlon.app theme preference.

packages/app/provider/index.tsx
```diff
@@ -98,7 +57,7 @@ export const setTheme = async (
 ) => {
   try {
     setActiveTheme(theme);
```
latter-bolden, line 59: is continuing to set this manually here (and in `clearTheme`) just a perf optimization? I'd expect the store action to trigger a change in `useSyncedAppTheme` itself via the `useThemeSettings` hook

latter-bolden (commented):
Good find on the dep conflict. The reasoning there and the lockfile cascade from `dedupe` make sense to me.

On the theme state changes I'm a little confused, added a few questions.

davidisaaclee:
@latter-bolden good pointers, made some changes:
- [Use direct theme writes / DB subs where possible](https://github.com/tloncorp/tlon-apps/pull/4951/commits/abb9e2b032b0d293f93774bbdce30e7c0e8a4ace) – replaces use of `ThemeContext` `setActiveTheme` with direct `store.setTheme` calls, relying on subscription to store via `useThemeSettings` to propagate to Tamagui
- [Remove redundant theme subscription / pull](https://github.com/tloncorp/tlon-apps/pull/4951/commits/0dc9e55187f99e57a2899c841c9088e0e83291b1)

I tested by:
- changing theme on one client, seeing that the theme updated automatically on another open client
- changing theme on one client while other client was also in theme manager screen; seeing that theme manager selection automatically updated on listening client
- changing theme on one client while other client was closed; opening second client flashed default `light` theme, then loaded the chosen theme from the network

davidisaaclee:
e2e tests failed CI, but running `pnpm e2e` locally passed, with 1 flaky test

latter-bolden (approved):
Edits look good! I dm'd you as well, but we may wanna fix the theme flash on web if there's a mismatch with system setting (with how this is working now, I think that's a regression)

But otherwise lgtm and works in testing!

davidisaaclee:
@latter-bolden 
> I think that's a regression

this is what I see on `develop`:

dark mode refresh
https://github.com/user-attachments/assets/f3f934cc-a6e3-467b-a889-e8e2738afd3d

light mode refresh
https://github.com/user-attachments/assets/70b386e4-e181-4b9c-b51b-d33ca7ba3868

is that the flash you're talking about?

davidisaaclee:
resolved offline, sounds like not a regression

PR #4954, opened 2025-07-25
ActionSheet: fix issue with unreliable button presses

patosullivan:
## Summary
Fixes tlon-4602.

~Fix intermittent button press issues in ActionSheet on iOS where buttons required multiple taps to trigger actions (or, in some cases, the buttons would never trigger the actions at all). Replaced unreliable onPress callbacks with onPressIn/onPressOut events and manual state tracking to ensure reliable touch handling in modal contexts.~

~I assume there is something happening here that is causing normal onPress to fail because we're in a modal context. When we were using onPress you could still see the press register visually (the press style would be applied), but the handler would never fire.~

Update: the issue was resolved by switching from `ActionSheet.ScrollableContent` to `ActionSheet.Content` in `ChatOptionsSheet`. The `onPressIn`/`onPressOut` solution was unnecessary, and I don't see any reason why we should need to scroll this content.

## Changes

- Use `ActionSheet.ScrollableContent` rather than `ActionSheet.Content` in `ChatOptionsSheet`.
- Move `handlePress` into a `useCallback`

## How did I test?

Tested on iOS, web, and Android.

## Risks and impact

## Rollback plan

revert

## Screenshots / videos

https://github.com/user-attachments/assets/64766f3b-f1b1-4f3c-8753-1f89a60f1bbb

### DIFF
--- a/packages/app/ui/components/ActionSheet.tsx
+++ b/packages/app/ui/components/ActionSheet.tsx
@@ -565,9 +565,18 @@ function ActionSheetAction({
     pressStarted.current = false;
   }, [action, accent]);
 
+  const handlePress = useCallback(() => {
+    if (accent !== 'disabled' && !action.disabled && action.action) {
+      action.action();
+    }
+  }, [action, accent]);
+
   if (action.render) {
     return action.render({ action });
   }
+
+  // Use regular onPress on desktop/web where popovers work fine
+  const useAlternativePress = Platform.OS !== 'web' && isWindowNarrow;
 
   return (
     <ActionSheetActionFrame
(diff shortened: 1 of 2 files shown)


patosullivan:
Converting to draft while I figure out why this broke some of the e2e tests.

patosullivan:
> Converting to draft while I figure out why this broke some of the e2e tests.

We just need to make sure we use the normal `onPress` handler on desktop web.

dnbrwstr (changes requested):
Could this be related to the issue with Pressables inside `ScrollView`s? I think this might warrant a little more investigation to determine how prevalent the issue is, or if it's only happening in this case why that as, as there are a bunch of places it could pop up.

PR #4955, opened 2025-07-25
Update to Expo 52

dnbrwstr:
## Summary

Updates project to expo 52, following the diff from this upgrade helper: https://docs.expo.dev/bare/upgrade/?fromSdk=51&toSdk=52

## How did I test?

Tested basic functionality on iOS, Android, and web.

## Risks and impact

This upgrades a number of packages, so warrants an extra-thorough QA pass.

- Affects important code area: Pretty much everything

## Rollback plan

Revert


pnpm-lock.yaml
```diff
@@ -1686,14 +1702,18 @@ packages:
     resolution: {integrity: sha512-zevQbhbau95nkoxSq3f/DC/SC+EEOUZd3DYqfSkMhY2/wfSeaHV1Ew4vk8e+x8lja31IbyuUa2uQ3JONqKbysw==}
     engines: {node: '>=6.9.0'}
 
-  '@babel/helper-annotate-as-pure@7.22.5':
-    resolution: {integrity: sha512-LvBTxu8bQSQkcyKOU+a1btnNFQ1dMAd0R6PyW3arXes06F6QLWLIrd681bxRPIXlrMGR3XYnW9JyML7dP3qgxg==}
+  '@babel/generator@7.28.0':
```
davidisaaclee, line 1705: icymi, it looks like this diff has the issue with multiple versions of `@babel/core` that I was talking about – it looks like this is even adding a third version of some of these plugins. I don't know how the diff will look against #4951 but you could try doing `pnpm dedupe` to see if it fixes it.

dnbrwstr: This should be resolved now

latter-bolden (approved):
Admittedly skimming since I'm not super familiar with all these config knobs, but changes look reasonable.

Was able to install on my devices without noticing regressions on either platform.

PR #4957, opened 2025-07-28
channel: represent "cannot read" state for a currently viewed channel

patosullivan:
## Summary

fixes tlon-4608. 

Users can get into a weird state if they're actively viewing a channel while the admin for the group changes the channel permissions in such a way to prevent that user from viewing it.

The UI wouldn't update to show this new state, so the user wouldn't know that they can't read it (or post to it).

Normally it would be impossible to get into this state since you can't navigate to channels you're not allowed to read, this can only happen if you're actively viewing the channel.

This PR adds a new type to `ReadOnlyNotice`, `no-longer-read`, that lets the user know they can no longer read the channel. It also adds a new hook, `useCanRead`, that we call in the channel component to determine if the user is in this state.

## How did I test?

Tested on web and iOS simulator.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

revert

## Screenshots / videos

Simulator Screenshot - iPhone 16 - 2025-07-28 at 11 27 37

### DIFF
--- a/packages/app/ui/components/Channel/ReadOnlyNotice.tsx
+++ b/packages/app/ui/components/Channel/ReadOnlyNotice.tsx
@@ -5,11 +5,18 @@ import { SizableText, View, YStack } from 'tamagui';
 export function ReadOnlyNotice({
   type,
 }: {
-  type: 'read-only' | 'dm-mismatch' | 'group-dm-mismatch' | 'channel-mismatch';
+  type:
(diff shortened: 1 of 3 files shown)


latter-bolden (approved):
Nice, good clean fix

PR #4962, opened 2025-07-29
InviteUsersWidget: disable users who are already members to prevent double-inviting

jamesacklin:
## Summary

Fixes TLON-4622 by preventing the user from inviting contacts and other ships who are already members to the group again.

## Changes

When users access the "Invite people" menu, the InviteUsersWidget now extracts all existing group member IDs. These member IDs are passed to the ContactBook component as the immutableIds prop.

The ContactBook component then uses these IDs to mark contacts as immutable, which shows a blue checkmark next to their names, prevents them from being selected for invitation, and provides clear visual feedback that they're already group members.

## How did I test?

- Select the "Invite people" group option for a group you have invite rights for (admin/host of a private/secret group or member of a public group)
- Note that if some of your contacts are in the group, they are unselectable
- Enter a portion of a valid ship name you know is in the group but not in your contacts
- Note that they are marked ineligible as well
- 
## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

git revert

## Screenshots / videos

Screenshot 2025-07-29 at 10 36 13 AM

Screenshot 2025-07-29 at 10 36 10 AM

### DIFF
--- a/packages/app/ui/components/InviteUsersWidget.tsx
+++ b/packages/app/ui/components/InviteUsersWidget.tsx
@@ -22,6 +22,11 @@ const InviteUsersWidgetComponent = ({
   const [loading, setLoading] = useState(false);
   const [invitees, setInvitees] = useState<string[]>([]);
 
+  // Extract existing group member IDs to mark them as immutable
+  const existingMemberIds = useMemo(() => {
+    return group.members?.map(member => member.contactId) ?? [];
+  }, [group.members]);
+
   const handleInviteGroupMembers = useCallback(async () => {
     setLoading(true);
     try {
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):
Good defense, I think this is fine to merge.

Worth noting though, I don't think we're guaranteed to always have the full member set available when this widget is shown.

jamesacklin:
> Worth noting though, I don't think we're guaranteed to always have the full member set available when this widget is shown.

Why not? Aren't we in the group?

latter-bolden:
> Why not? Aren't we in the group?

We expose it on ChatList long press. We also avoid re-syncing group members too frequently, so they can be out of sync if facts are missed.

In other words, I think this is good to have but we shouldn't rely on it for correctness (e.g. duplicate invite backend fix still important)

PR #4963, opened 2025-07-29
channels: tombstones with more information

arthyn:
## Summary

@Fang- started this PR and I finished it out, fixes TLON-4571. This solves an issue we found with sequence numbers in that deletes do not contain any data so we can't be sure what sequence number they would have. Since our logic depends on getting a complete picture of the sequence numbers, we needed to make this change. We will also have to do the same for %chat.

FYI I did not go through and change all helper functions to use new conversion library, but this would probably significantly reduce the amount of code there.

## Changes

- added tombstone type which contains id, sequence number, author, and deleted time
- modified all instances of posts/replies to be either the actual post or a tombstone instead of a unit
- modified the agent to work with new tombstone type
- added migration to restore tombstones from host data
- added new endpoints to work with and marks to go with them
- added conversion library to easily convert between types
- changed how we referenced versioned types
- added more to mop extensions

## How did I test?

Added tests for tombstone migration and mop extensions, waiting for frontend-integration to test endpoints

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [X] Message sync
  - [X] Channel display
  - [ ] Notifications

## Rollback plan

We'd have to create another migration to go back to the old types and roll the protocol version back (which I'm not sure works we've never done that).


mikolajpp (commented):
I have reviewed the bulk of these changes. The only remaining unreviewed piece is `/lib/channels-utils.hoon`. Everything looks good, found only one possible error regarding ship author comparison, and left some comments here and there.

desk/lib/channel-utils.hoon
```diff
@@ -274,76 +353,89 @@
...
+  ?:  ?=(%| -.v-reply)  ~
+  `[time (uv-reply-1 parent-id +.v-reply)]
 ::
 ++  uv-replies-1
   |=  [parent-id=id-post:c =v-replies:c]
-  ^-  replies:v7:old:c
-  %+  gas:on-replies:v7:old:c  *replies:v7:old:c
+  ^-  replies:v7:c
+  %+  gas:on-replies:v7:c  *replies:v7:c
   %+  murn  (tap:on-v-replies:c v-replies)
```
mikolajpp, line 390: Why do we use `+murn` here if we always return a full unit below?

desk/lib/channel-utils.hoon
```diff
@@ -274,76 +353,89 @@
...
+  ?:  ?=(%| -.v-reply)  `[time ~]
+  `[time `(uv-reply-1 parent-id +.v-reply)]
 ::
 ++  uv-replies-2
   |=  [parent-id=id-post:c =v-replies:c]
-  ^-  replies:c
-  %+  gas:on-replies:c  *replies:c
+  ^-  replies:v8:c
+  %+  gas:on-replies:v8:c  *replies:v8:c
   %+  murn  (tap:on-v-replies:c v-replies)
```
mikolajpp, line 400: Same question as above, `+murn` is not necessary here, `+turn` would be enough.

desk/lib/channel-utils.hoon
```diff
@@ -613,144 +706,207 @@
...
+  |=  ref=reference:c
+  ^-  reference:v7:c
+  ?-  -.ref
+      %post
+    =-  ref(post -)
+    ?:  ?=(%& -.post.ref)
+      (simple-post-1 +.post.ref)
+    %*  .  *simple-post:v7:c
+      content  `story:v7:c`[%inline ['[deleted message]']~]~
+      author  ~nul
```
mikolajpp, line 743: I think I saw it mentioned somewhere. We should definitely do away with the ~nul posting in a later round of channels changes.

desk/lib/channel-utils.hoon
```diff
@@ -613,144 +706,207 @@
...
 ++  to-said-1
   |=  =said:c
-  ^-  said:v7:old:c
-  =-  said(q -)
-  ?-  -.q.said
-     %post  q.said(post (simple-post-1 post.q.said))
-     %reply  q.said(reply (simple-reply-1 reply.q.said))
-  ==
+  ^-  said:v7:c
+  ::NOTE  the tombstone branches here _shouldn't_ get hit in practice,
```
mikolajpp, line 759: Hm, does this mean ~nul posts above will actually never resolve?

mikolajpp (commented):
Reviewed `/lib/channels-json.hoon` as well. The only interesting thing I found were some unnecessary +murns.

desk/lib/channel-conv.hoon
```diff
@@ -0,0 +1,840 @@
...
+  ++  reference
+    |%
+    ++  v8
+      |=  =reference:v9:c
+      ^-  reference:v8:c
+      ?:  ?=(%reply -.reference)
+        :-  %reply
+        :-  id-post.reference
+        ?:  ?=(%| -.reply.reference)
+          *simple-reply:v8:c
```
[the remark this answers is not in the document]
arthyn: yeah we would see ~zod, I can change this to match some of our other stuff that actually puts something in the content.

desk/app/channels.hoon
```diff
@@ -106,7 +131,10 @@
           [/v2 %channel-response-3 ~]
           [/v2/said %channel-said-1 ~]
         ::
+          [/v3 %channel-response-4 ~]
```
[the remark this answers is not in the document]
arthyn: for real, mvp!

desk/lib/channel-utils.hoon
arthyn: you mean like not returning anything if asked about a tombstoned post?

desk/lib/channel-utils.hoon
arthyn: I think that's the intention

desk/app/channels.hoon
```diff
@@ -2674,58 +2824,51 @@
...
+    ::
         [%newest count=@ ~]
       =/  count  (slav %ud count.pole)
-      =/  reply-map  (gas:on *v-replies:c (top:mo-v-replies:c replies count))
-      ?:  ?=(%v3 version)
-        ``channel-replies-3+!>((uv-replies-2:utils parent-id reply-map))
-      ?:  ?=(%v2 version)
-        ``channel-replies-2+!>((uv-replies-1:utils parent-id reply-map))
-      ``channel-replies+!>((uv-replies:utils parent-id reply-map))
+      `(gas:on *v-replies:c (top:mo-v-replies:c replies count))
```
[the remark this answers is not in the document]
arthyn: yep

desk/lib/channel-utils.hoon
mikolajpp: we could have a dedicated "missing"/"unknown" $reference variant, similar to how we do `/chat/deleted` kind nowadays.

mikolajpp: Ok, I see this is the old type which does not have a path-based kind yet.

mikolajpp (approved):
The changes here look good, approved!

PR #4969, opened 2025-07-30
add splash screen and app launch progress manager

davidisaaclee:
## Summary
fixes TLON-3432
Adds a splash screen to all platforms (web uses the `Starting up...` spinner as a splash screen), and adds a global "progress manager" that keeps the splash screen visible until all registered tasks (e.g. starting web SQLite, loading theme) have finished.

<details>

<summary>Note: I found out late into this task that Android's splash screen behavior is unique, and `expo-splash-screen` works on Android in a "fake" way – this means that we have two splash screens on Android (one that looks like a cropped app icon, and the newer splash screen added by this PR).</summary>

You can see this happen in the Android screen recording below. afaict, expo-splash-screen simply shows a view as soon as the app opens, while the RN host is loading – not an OS-level splash screen.

For lack of better terms, I'll call the first Android-required splash screen the "launch icon", and the second `expo-splash-screen` screen the "splash screen".

I see a few ways of fixing this:
- (S) Use a transparent image for the launch icon – I don't know what other effects this has in Android. It'll be a little weird – user will just have a blank screen before seeing the splash screen.
- (M) Use `expo-splash-screen`'s `NATIVE` resize mode – this still gives us two splash screen phases, but uses the same graphic for the launch icon. However, this method has an unavoidable visual jank, since the image will change position while transitioning from launch icon to splash screen.
- (XL) Do it "the right way," and define the Android splash screen in native terms (disabling expo-splash-screen for Android). In this way, we could even show a little animation from app icon (upright TM logo) to the splash screen logo (tilted, larger corner radius). We'd need to expose a hook from JS -> Kotlin to indicate "hide splash screen."

</details>

## Changes
- Set min iOS version in Podfile to 16.0 – this matches the min version set on the Xcode target, so this is not breaking. (Using iOS 15 as min, expo-splash-screen failed to install.) https://github.com/tloncorp/tlon-apps/blob/4f14cac29253aefa14f6e4c2da978c6c3635c4f0/apps/tlon-mobile/ios/Podfile#L32
- Upgrades `expo-splash-screen` and specifies a splash screen for Android / iOS based on newest designs
- Adds a generic `ProgressManager` that is used via a singleton `splashScreenProgress` to track tasks blocking splash screen hide
- Adds some type safety + `.off()` + utility hook to existing `EventEmitter` implementation
- [Store explicit 'auto' theme in DB](https://github.com/tloncorp/tlon-apps/pull/4969/commits/3ec4ae52ac25256edd3acabc87e263e753c84e00) – Previously, when saving the user's theme, if the selected them was `'auto'`, we'd store `null` instead of `'auto'`. I don't see the benefit of this, and I do see the benefit of storing an explicit selection of `'auto'` – so I removed the code that special-cases `'auto'`.
- [Remove unnecessary Android night splash screen spec](https://github.com/tloncorp/tlon-apps/pull/4969/commits/4fca363286fd34d8e84681c622077c87f8687a34) – with this file in place, the Android expo-splash-screen did not match the system appearance. It seems like this .xml is trying to spec a splash screen in an older method. (We switch the background color in night mode via `values/colors.xml` vs. `values-night/colors.xml`.)

## How did I test?
See screen recording below.
I also refreshed each mobile app while logged out (web app can't log out) to ensure the splash screen doesn't block. (Maybe a good follow-up task would be a way to kill the splash screen, either by timing out or on user tap.)

## Risks and impact
**A bug in the splash screen logic could get users stuck on the splash screen, unable to enter the app.**

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan
git revert

### iOS
Production release (no dev server)
Added debug patch that delays theme load from DB by 500ms
Opening fresh while light mode; while dark mode; then changing theme on another client while this client is killed; then opening client and seeing theme change (asynchronously)

https://github.com/user-attachments/assets/a0cf7c08-0442-4925-b48e-fbfd33736ae7

### Android
Preview release (no dev server)
Open fresh while light mode; while dark mode
(My computer was chugging a bit, app was smooth otherwise)

https://github.com/user-attachments/assets/0b209892-e172-413d-b6d3-1a16cc6df09d

### Web
Added debug patch that delays theme load from DB by 500ms
Run with persistence enabled on SQLite (so theme could be remembered)
Reload while auto theme / light mode; reload while auto theme / dark mode; reload while custom theme (light mode, dark mode)

https://github.com/user-attachments/assets/c69eeec7-8489-4b64-8565-297d400c40fa

### DIFF
--- a/apps/tlon-mobile/android/app/src/main/AndroidManifest.xml
+++ b/apps/tlon-mobile/android/app/src/main/AndroidManifest.xml
@@ -32,7 +32,7 @@
         <action android:name="MARK_AS_READ"/>
       </intent-filter>
     </receiver>
-    <activity android:name=".MainActivity" android:label="@string/app_name" android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode" android:launchMode="singleTask" android:windowSoftInputMode="adjustResize" android:theme="@style/Theme.App.SplashScreen" android:exported="true" android:screenOrientation="portrait">
+    <activity android:name=".MainActivity" android:label="@string/app_name" android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode" android:launchMode="singleTask" android:windowSoftInputMode="adjustResize" android:theme="@style/AppTheme" android:exported="true" android:screenOrientation="portrait">
       <intent-filter>
         <action android:name="android.intent.action.MAIN"/>
         <category android:name="android.intent.category.LAUNCHER"/>
--- a/apps/tlon-mobile/android/app/src/preview/AndroidManifest.xml
+++ b/apps/tlon-mobile/android/app/src/preview/AndroidManifest.xml
@@ -8,7 +8,7 @@
         tools:ignore="GoogleAppIndexingWarning"
     >
         <meta-data android:name="io.branch.sdk.TestMode" android:value="true" />
-        <activity android:name=".MainActivity" android:label="@string/app_name" android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode" android:launchMode="singleTask" android:windowSoftInputMode="adjustResize" android:theme="@style/Theme.App.SplashScreen" android:exported="true" android:screenOrientation="portrait">
+        <activity android:name=".MainActivity" android:label="@string/app_name" android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode" android:launchMode="singleTask" android:windowSoftInputMode="adjustResize" android:theme="@style/AppTheme" android:exported="true" android:screenOrientation="portrait">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN"/>
                 <category android:name="android.intent.category.LAUNCHER"/>


apps/tlon-mobile/src/App.main.tsx
```diff
@@ -35,6 +37,15 @@ import { OnboardingStack } from './OnboardingStack';
 import AuthenticatedApp from './components/AuthenticatedApp';
 import { SignupProvider, useSignupContext } from './lib/signupContext';
 
+SplashScreen.preventAutoHideAsync().catch((err) => {
+  console.warn('Failed to prevent auto hide splash screen', err);
+});
+splashScreenProgress.emitter.on('complete', () => {
+  SplashScreen.hideAsync().catch((err) => {
```
latter-bolden, line 44: I'm a bit paranoid about getting stuck here. Can we wrap this in a `withRetry` and make sure it triggers a tracked error?

packages/app/provider/index.tsx
```diff
@@ -56,17 +57,34 @@ function useSyncedAppTheme() {
...
+    if (storedTheme == null) {
       return;
     }
-    const normalizedTheme =
-      storedTheme == null ? 'auto' : normalizeTheme(storedTheme);
-    setActiveTheme(getDisplayTheme(normalizedTheme, isDarkMode));
-  }, [isLoading, storedTheme, isDarkMode]);
+    setActiveTheme(storedTheme.tamaguiTheme);
+
+    splashScreenProgress.complete(SplashScreenTask.loadTheme);
```
latter-bolden, line 86: Am I reading this correctly that as soon as the DB query loads, it'll always complete the progress task? Meaning even if the DB is wiped (app updated in background) and we can't connect to the ship, it'll still proceed past the splash?

packages/shared/src/http-api/Urbit.ts
```diff
@@ -5,6 +5,7 @@ import { isBrowser } from 'browser-or-node';
 import { TimeoutError } from '../api';
 import { desig } from '../urbit';
 import * as utils from '../utils';
+import { EventEmitter } from '../utils/EventEmitter';
```
latter-bolden, line 8: thanks for propagating the util down to this layer 🙏🏻

packages/app/provider/index.tsx
davidisaaclee: that's how I read it as well (it sounds like in that case, we'd just get a `storedThemeRaw` of `null`) – I can try to simulate that (if you have a good way to simulate it, lmk)

(it's a good call-out that the case where `storedTheme` is `null` is not 100% clear and could accumulate a bug with changes, so I'll try to fix that as well.)

apps/tlon-mobile/src/App.main.tsx
davidisaaclee: yeah, it's a fair worry – I'll do that, and I'm also going to put in a generous timeout (I think unconditionally hiding splash after 5 seconds is fair).

I also think extending `ProgressManager` to take a timeout per task would be nice, but a global timeout for splash hiding seems safest / smallest.

packages/app/provider/index.tsx
latter-bolden: I could give it a shot too. Easiest might be:

1. log in from a local node running off your machine
2. with the app open, shut down the node and use the devtools to wipe DB
3. force close and re-open, confirm proceed past splash

davidisaaclee: Following those steps on iOS, I get this screen:

simulator_screenshot_4545D364-E719-4F93-8D08-303FAF284AED

(For web, iiuc, this case isn't possible because "can't connect to the ship" means that we can't get a client running either.)

apps/tlon-mobile/src/App.main.tsx
davidisaaclee: [Wrap SplashScreen.hideAsync in withRetry; track failing error](https://github.com/tloncorp/tlon-apps/pull/4969/commits/ac5e369b8c5721c2e02f8df212cfbf48f6d404b3)

[Add timeouts per task in ProgressManager; add default timeouts to splash](https://github.com/tloncorp/tlon-apps/pull/4969/commits/e5b424549def16a791ffe23959c70cb985e3ef10) – doing it per-task ended up being more straight-forward than a global timeout.

I tested this by:
1. Simulating 3 errors in the body of `withRetry`, printing them as they ran; on the fourth try, the call succeeds, and the splash disappears
2. I added a 10 second sleep in `useThemeSettings` body – after 5 seconds, the splash screen progress timed out, and splash screen got hidden

I did not test that the error makes it to the cloud – let me know if you think this is the wrong way of doing it (`splashscreenLogger` is a `createDevLogger`): https://github.com/tloncorp/tlon-apps/blob/ac5e369b8c5721c2e02f8df212cfbf48f6d404b3/apps/tlon-mobile/src/App.main.tsx#L48-L52

packages/app/provider/index.tsx
davidisaaclee: [Use loaded: false instead of null for clarity](https://github.com/tloncorp/tlon-apps/pull/4969/commits/fa8716d8ffdb5a22a8098ae80ba5e6eae7b5b296)

latter-bolden: Yeah, it's not a great state. Imagine it's the same on develop.

Getting past splash means they have access to the logout button, that's what I wanted to confirm.

apps/tlon-mobile/src/App.main.tsx
latter-bolden: Yep, that's the correct way to track it. Will make it to cloud.

PR #4970, opened 2025-07-31
channel: show when group has been deleted

patosullivan:
## Summary

fixes tlon-4620

We use ReadOnlyNotice to show the user (who is currently viewing a channel) that the group has been deleted.

## Changes

- Detect group deletion by checking if we don't have `group`.
- Display specific message in ReadOnlyNotice for this case.

## How did I test?

Web

## Risks and impact

## Rollback plan

Revert

## Screenshots / videos

Home Groups Screenshot

### DIFF
--- a/packages/app/features/top/ChannelScreen.tsx
+++ b/packages/app/features/top/ChannelScreen.tsx
@@ -62,6 +62,8 @@ export default function ChannelScreen(props: Props) {
     editPost,
     channel,
     group,
+    groupIsLoading,
+    groupError,
   } = useChannelContext({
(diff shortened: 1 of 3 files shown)


dnbrwstr (commented):
Is not having the group a reliable sign that the group has been deleted? Vs not having been loaded from remote or retrieved from database yet.

dnbrwstr (approved):
This works for now -- at some point would be nice to have positive confirmation of a group being deleted + store it as a flag on the db. Or maybe we should just remove the group altogether on deletion?

PR #4973, opened 2025-07-31
staging -> develop

jamesacklin:
Seems like the checks failed due to a conflict in desk.docket-0


latter-bolden:
I think you need to use a temp branch here to resolve the docket conflict without committing the change to staging.

PR #4975, opened 2025-08-01
chat: fix edit functionality

patosullivan:
## Summary

fixes tlon-4643

We were always passing the post id as the parentId for the draft when `isEdit` was true. 

## Changes

Only pass the parentId to parentId.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications


davidisaaclee:
I don't see where we're specifying which post we're editing in the edit case – like, if a channel has 3 posts in it, and I send an edit draft like:

```ts
{
          channelId,
          content,
          channelType,
          isEdit: true,
          parentId: '', // we're editing a post in the main timeline
        }
```
... how do I know which of the 3 posts I'm trying to edit?

PR #4976, opened 2025-08-01
fix thread emoji reactions

patosullivan:
## Summary

We weren't sending a parentId (or the related pokes) for adding/remove reactions to thread reply messages.

## Changes

Pass in the parentId to `addReaction` and `removeReaction`, add the relevant pokes.

## How did I test?

Tested on web.

## Risks and impact

## Rollback plan

Revert


jamesacklin:
Confirmed working for me

latter-bolden (approved):
fix looks good, works locally!

PR #4979, opened 2025-08-04
channels/server: ;; instead of !< for sequence nr pokes

Fang-:
## Summary

In real-world conditions, the vases coming in to agents will have the types on them that match the marks on their cages. The tests were failing to account for this, and so didn't catch that this code was subtly broken. We patch both the tests and the code here.

## Changes

First we patch up the existing tests to account for the new discipline wrapper properly. Then we add a new test for the channels-server side of the poke-based sequence nr negotiation flow.

With that setup done, we patch the tests to more accurately simulate real-world invocations, and finally make one-line changes to the agent code to handle these invocations correctly.

## How did I test?

Updated the tests to be more accurate, which made them fail. Updated the code to make the new tests pass.

## Risks and impact

`;;` is notorious for being slow in some select cases. ~~I haven't yet tested this with real-world amounts of content. Please hold off on merge until I (or you, dear reviewer) do so and confirm that it's Probably Fine™.~~ I have confirmed this is not hilariously slow. For ~6k sequence nrs, the `;;` only takes 14ms.

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Just revert, I suppose.

## Screenshots / videos

image

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -402,7 +402,7 @@
       ==
     ::
         [%send-sequence-numbers *]
-      =+  !<([%send-sequence-numbers =nest:c] vase)
+      =+  ;;([%send-sequence-numbers =nest:c] q.vase)
       ?~  can=(~(get by v-channels) nest)  cor
       =;  =cage
         (emit [%pass /numbers %agent [src.bowl %channels] %poke cage])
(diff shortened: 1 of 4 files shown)


Fang-:
I have confirmed this is not hilariously slow. For ~6k sequence nrs, the `;;` only takes 14ms.

arthyn (approved):

PR #4983, opened 2025-08-05
chat: tombstones

arthyn:
## Summary

A follow-up to tloncorp/tlon-apps#4963 and part of fixing TLON-4571. This PR adds the same delete tombstoning to %chat so that we can have fully contiguous sequence numbers.

Also fixes TLON-4406

## Changes

- added tombstones to chat post/replies
- added new conversion library for chat
- brought more channel conversion functions into library
- added new marks/endpoints

## How did I test?

Still in progress

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [X] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/lib/dm.hoon
+++ b/desk/lib/dm.hoon
@@ -197,7 +197,7 @@
           author.u.reply
           now
       ==
-    :-  pac(dex (~(del by dex.pac) id))
+    :-  pac
     (put:on:replies:c replies u.tim %| tombstone)
   ::
       %add-react


desk/sur/chat.hoon
```diff
@@ -1,4 +1,5 @@
 /-  d=channels
+/-  c2=chat-2, c3=chat-3, c4=chat-4, c5=chat-5
```
Fang-, line 2: We've talked about this previously, but for the record: I don't like these dependencies just to re-export them. If you need one, import the relevant file.

desk/lib/dm.hoon
```diff
@@ -148,204 +171,70 @@
...
+    [pac (put:on:replies:c replies now %& reply-seal memo.delta)]
   ::
       %del
     =/  tim=(unit time)  (~(get by dex.pac) id)
     ?~  tim  [pac replies]
-    =/  =time  (need tim)
-    =^  reply=(unit reply:c)  replies
-      (del:on:replies:c replies time)
-    =.  dex.pac  (~(del by dex.pac) id)
-    [pac replies]
```
Fang-, line 160: What happened to the `=.  dex.pac` here? The new code just produces `pac` as-is, always, but it should delete the lookup for this reply right?

desk/lib/dm.hoon
```diff
@@ -436,11 +328,13 @@
...
+      ?:  ?=(%| -.writ)  [~ ~]
       ?-  ver
-        %v0  ``writ+!>((old-writ-3 writ))
-        %v1  ``chat-writ-1+!>((old-writ-4 writ))
-        %v2  ``chat-writ-2+!>(writ)
+          %v0  ``writ+!>((v3:writ:v5:cv (v5:writ:v6:cv +.writ)))
+          %v1  ``chat-writ-1+!>((v4:writ:v5:cv (v5:writ:v6:cv +.writ)))
+          %v2  ``chat-writ-2+!>((v5:writ:v6:cv +.writ))
       ==
     ``loob+!>(?~((get ship `@da`time) | &))
```
Fang-, line 339: For v0 through v2, this may need to check both whether it's present and whether it's not a tombstone. Otherwise we may give "true" for the `%u` check, but then still produce `[~ ~]` when it gets scried out.

desk/app/chat.hoon
```diff
@@ -671,14 +763,14 @@
...
+      ?(%chat-dm-diff-1 %chat-dm-diff-2)
     =+  !<(=diff:dm:c vase)
     di-abet:(di-take-counter:(di-abed-soft:di-core src.bowl) diff)
   ::
       %chat-club-create
     cu-abet:(cu-create:cu-core !<(=create:club:c vase))
   ::
-      %chat-club-action-1
+      ?(%chat-club-action-1 %chat-club-action-2)
     =+  !<(=action:club:c vase)
```
Fang-, line 774: Are you sure these are correct? Do these marks actually have the same type? Why?

desk/app/chat.hoon
```diff
@@ -1297,8 +1295,9 @@
       %add
     =/  time=(unit time)    (~(get by dex.pact) p.diff)
     ?~  time  ~
-    =/  writ=(unit writ:c)  (get:on:writs:c wit.pact u.time)
+    =/  writ=(unit (may:c writ:c))  (get:on:writs:c wit.pact u.time)
     ?~  writ  ~&(%diff-to-response-miss ~)
+    ?:  ?=(%| -.u.writ)  ~&(%diff-to-response-miss ~)
```
Fang-, line 1300:
```suggestion
    ?:  ?=(%| -.u.writ)  ~&(%diff-to-response-miss-on-tomb ~)
```

desk/app/chat.hoon
```diff
@@ -2240,25 +2252,28 @@
       (di-give-writs-diff diff)
     ::
         %del
-      =?  di-core  ?=(^ had)
-        =*  content  content.writ.u.had
+      =?  di-core  &(?=(^ had) ?=(%& -.writ.u.had))
+        =*  content  content.+.writ.u.had
```
Fang-, line 2256: Here and elsewhere, if there isn't a face on there, you don't need the lark. Name resolution will just walk the tree to the first instance of the face you ask for.

```suggestion
        =*  content  content.writ.u.had
```

Fang- (commented):
Some questions/comments, but overall looks good. Nothing surprising here!

In the future, let's save big renaming/restructuring efforts for after protocol work is done. Or do it before, but separate.

desk/app/chat.hoon
```diff
@@ -1735,12 +1728,16 @@
   ::
   ++  cu-give-action
     |=  =action:club:c
+    =/  v5  (v5:action-club:v6:cv action)
```
mikolajpp, line 1731: Same as I commented elsewhere, I think using a descriptive face with a version tag is better for readability.
```suggestion
    =/  action-5  (v5:action-club:v6:cv action)
```

mikolajpp (changes requested):
Looks good, I only have a few styling comments and other questions.

desk/app/chat.hoon
```diff
@@ -38,30 +54,47 @@
...
+/%  m-writ-response-3      %writ-response-3
 ::
 /*  desk-bill  %bill  /desk/bill  ::  keep warm
 ::
 %-  %-  discipline
     :+  ::  marks
         ::
         :~  :+  %chat-blocked-by      &  -:!>(*vale:m-chat-blocked-by)
-            :+  %chat-club-action     &  -:!>(*vale:m-chat-club-action)
+            :+  %chat-club-action     |  -:!>(*vale:m-chat-club-action)
```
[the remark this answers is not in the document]
mikolajpp: Discussed in person, the reason for this is that the type was wrong and we want to correct. We should put a comment here and flip it back in the next upgrade.

desk/app/chat.hoon
```diff
@@ -2290,25 +2302,28 @@
...
     ::
         %reply
       =*  delta  delta.q.diff
-      =/  entry=(unit [=time =writ:c])  (get:di-pact p.diff)
-      =?  meta.q.diff  !=(~ entry)  `reply-meta.writ:(need entry)
+      =/  entry=(unit [=time writ=(may:c writ:c)])  (get:di-pact p.diff)
+      ::  if we don't have the entry, we can't reply to it
+      ?~  entry  di-core
+      ?:  ?=(%| -.writ.u.entry)  di-core
+      =.  meta.q.diff  `reply-meta.+.writ:(need entry)
```
mikolajpp, line 2317:
```suggestion
      =.  meta.q.diff  `reply-meta.writ:(need entry)
```

desk/app/chat.hoon
```diff
@@ -2322,11 +2337,11 @@
           ?(%add-react %del-react)  (di-give-writs-diff diff)
       ::
           %del
-        =?  di-core  &(?=(^ entry) ?=(^ reply))
-          =*  content  content.reply.u.reply
+        =?  di-core  &(?=(^ reply) ?=(%& -.reply.u.reply))
+          =*  content  content.+.reply.u.reply
```
mikolajpp, line 2341:
```suggestion
          =*  content  content.reply.u.reply
```

desk/app/chat.hoon
```diff
@@ -2322,11 +2337,11 @@
...
           =/  mention  (was-mentioned:utils content our.bowl ~)
           =/  concern
-            [%delete-reply [id time]:reply.u.reply [id time]:writ.u.entry]
+            [%delete-reply [id time]:+.reply.u.reply [id time]:+.writ.u.entry]
```
mikolajpp, line 2344: I hope this works
```suggestion
            [%delete-reply [id time]:reply.u.reply [id time]:writ.u.entry]
```

desk/app/chat.hoon
```diff
@@ -2339,8 +2354,7 @@
         =.  recency.remark.dm  now.bowl
         =?  cor  &(!=(old-unread di-unread) !=(net.dm %invited))
           (give-unread ship/ship di-unread)
-        ?~  entry  (di-give-writs-diff diff)
-        =/  top-con  [id.writ.u.entry time.writ.u.entry]
+        =/  top-con  [id time]:+.writ.u.entry
```
mikolajpp, line 2357:
```suggestion
        =/  top-con  [id time]:writ.u.entry
```

desk/app/chat.hoon
```diff
@@ -2449,38 +2463,42 @@
...
+            =/  msg=(unit [=time writ=(may:c writ:c)])
               (get:di-pact our.bowl id)
             ?~  msg  [%del ~]
+            ?:  ?=(%| -.writ.u.msg)
+              [%del ~]
             :-  %add
-            ^-  memo:old-2
-            =,  writ.u.msg
+            ^-  memo:v2:c
+            =,  +.writ.u.msg
```
mikolajpp, line 2477: It seems like there is no need to expose just the tail, below should work fine.
```suggestion
            =,  writ.u.msg
```

mikolajpp (changes requested):
Marked some more places with redundant larks.

desk/lib/dm.hoon
```diff
@@ -44,7 +44,7 @@
     =/  parent   (get id)
     ?~  parent   [sum threads]
     ?:  ?=(%| -.writ.u.parent)  [sum threads]
-    =*  writ  +.writ.u.parent
```
Fang-: For these, where the lark is the very last thing being addressed, this is a meaningful change. In this context (and the `=,`s in /app/chat) it doesn't matter in practice, since all you do with the result is pull parts out of it, but worth keeping an eye out.

desk/lib/dm.hoon
```diff
@@ -187,7 +187,7 @@
           author.u.reply
           now
       ==
-    :-  pac
+    :-  pac(dex (~(del by dex.pac) id))
     (put:on:replies:c replies u.tim %| tombstone)
```
Fang-: It's not obvious to me this is correct. We're putting a tombstone in the replies of the post, but also deleting its index entry? The reply continues existing though. So, why delete it?

It's surprising to me anyway that we touch the `dex.pac` for replies at all (in the `%add` case too). I thought that only was for top-level keys. But maybe my mental model of reply handling for chat is off.

Fang- (commented):
One last question, then I think we should be good w/ recent changes.

desk/lib/dm.hoon
arthyn: I was just going off your previous comment about missing the `=.  dex.pac...` but yeah looking at it now what I originally had was correct. We add reply IDs to the dex so that they can be looked up by ID later.

PR #4986, opened 2025-08-06
emojimart, sigil: jam data to drastically reduce compile times

pkova:
It's getting increasingly annoying to publish any kind of OTA that changes Lull or Clay because these chonky libraries always get cache busted and the compile times are horrendous. For reference, running:

`-build-file /=groups=/lib/emojimart/hoon` 

on my lil digital ocean droplet takes **1 minute 48 seconds**, and my improved version from this PR takes around a second. I included instructions for updating the jammed data in both libraries in case it ever needs to change. The source files for the jammed data are also in the desk, they're just not imported anywhere.


Fang- (commented):
Left some comments on the emojimart lib, but the suggestions apply to both libs.

Also naming/structure-wise I'd do this slightly differently. We can inline the logic from `symbols.hoon` into `sigil.hoon` (as an arm there, probably), and then leave the data file only for the big data.

```
/desk/lib
  /emojimart.hoon
  /emojimart
    emoji.hoon
    emoji.jam
  /sigil.hoon
  /sigil
    symbols.hoon
    symbols.jam
```

pkova:
The commit event succeeds If the jamfiles contain data that doesn't micmic, but then it calls the `+on-fail` arm of `%profile`  which is the consumer here. That is certainly unexpected...

PR #4988, opened 2025-08-06
make sqlite exec async

dnbrwstr:
## Summary

Make sqlite connection `exec` method async and await calls. Fixes TLON-4234.

## How did I test?

Tested on web, iOS + Android.

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Revert


davidisaaclee (approved):
lgtm – I don't know how this was different from the reverted version, and I think we ran into some kind of issue last time, but I'm down to try it out from a fresh QA

PR #4989, opened 2025-08-06
channels: include latest seq nr in paged responses

Fang-:
## Summary

Per @latter-bolden's request, we now include the latest sequence nr in paged post responses.

## Changes

We modify the paged post types to include the sequence number of the newest message. Note that we don't bump any api versions here and change the mark type directly: we're working with the new version of the api introduced by #4963 which hasn't yet gone live.

As a small precaution for development ships, we temporarily set the `%paged-posts-4` mark to not be strictly checked by lib discipline.

## How did I test?

Manually did an http scry and saw the new value included in the result.

## Risks and impact

- Not quite safe to rollback without consulting PR author, though until this makes it to `develop` or onward it's pretty much free game.
  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Revert, but make sure client code isn't using affected endpoints.

## Screenshots / videos

???

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -79,7 +79,7 @@
             :+  %channel-posts           &  -:!>(*vale:m-channel-posts)
             :+  %channel-posts-2         &  -:!>(*vale:m-channel-posts-2)
             :+  %channel-posts-3         &  -:!>(*vale:m-channel-posts-3)
-            :+  %channel-posts-4         &  -:!>(*vale:m-channel-posts-4)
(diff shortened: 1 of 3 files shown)


latter-bolden (commented):
Very helpful, confirmed `newest` shows up as expected on the endpoint

latter-bolden (approved):

arthyn (approved):
lgtm

PR #4990, opened 2025-08-06
channels/server: ;; instead of !< for tombstone pokes

Fang-:
## Summary

Same as #4979, but for the tombstone engraving pokes.

## Changes

First we patch up the existing channels-server tests to account for new types.

With that setup done, we patch the channels agent test to more accurately reflect real-world invocations, and then make the one-line changes to handle those properly.

We also clean up some git merge artifacts that had made their way into a test file somehow.

## How did I test?

Saw the tests fail before, and succeed after the change here.

## Risks and impact

- Yes, safe to rollback without consulting PR author as long as the tombstone stuff hasn't gone live yet.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

`git revert`

## Screenshots / videos

No relevant media.

### DIFF
--- a/desk/tests/app/channels.hoon
+++ b/desk/tests/app/channels.hoon
@@ -212,7 +212,6 @@
       =.  posts.chan
         +:(del:on-v-posts:c posts.chan tombstone-key)
       (~(put by *v-channels:c) *nest:c chan)
-<<<<<<< HEAD
     ::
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

PR #4991, opened 2025-08-06
queue post edits

davidisaaclee:
## Summary
part of TLON-4650
Edits now queue similarly to sends – uploads do not block send; queue pauses if session disconnects; edits send in serial.

## Changes
- Draft previously had some confusion around the draft's "parent" (I was using it both as the `editingPost` and the thread root) – this caused tlon-4643 which was fixed by #4975; but there was some other places where this was being used incorrectly. I just removed all mentions of parentId, as we're not doing replies in this code path yet (and for edits, we can just grab the original post's `parentId` from database)
- Split `editPost` into variants similar to `sendPost` (facilitating sending from a raw draft, or from a "finalized" [i.e. uploaded] draft)

## How did I test?
- Did an edit on a text message in a chat
- Did an edit on a photo message in a chat
- Added multiple photos to a post via edit
- Did a bunch of edits + new messages while pausing network, then unpaused and made sure everything went through as expected

I went through the same process in a thread, and I did a quick edit of a notebook post.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan
git revert

## Screenshots / videos

<details>

<summary>This demo has the following patch applied, which pauses attachment uploads until I run `__resolveUploads()` in the console.</summary>

```diff
diff --git a/packages/shared/src/store/storage/storageActions.ts b/packages/shared/src/store/storage/storageActions.ts
index 6bceccda5..41273f16f 100644
--- a/packages/shared/src/store/storage/storageActions.ts
+++ b/packages/shared/src/store/storage/storageActions.ts
@@ -148,6 +148,18 @@ export const performUpload = async (
       isDigitalOcean ? headers : undefined,
       isWeb
     );
+
+    await new Promise((resolve) => {
+      console.log('pausing uploads');
+      const prev = (window.global as any).__resolveUpload;
+      Object.assign(window.global, {
+        __resolveUpload: () => {
+          prev?.();
+          resolve(null);
+        },
+      });
+    });
+
     return config.publicUrlBase
       ? new URL(fileKey, config.publicUrlBase).toString()
       : signedUrl.split('?')[0];
```

</details>

https://github.com/user-attachments/assets/a1003363-5925-4b41-980e-53cb6158b3d8


packages/shared/src/store/postActions.ts
```diff
@@ -406,52 +396,98 @@ export async function forwardGroup({
...
+  postBeforeEdit,
+  buildFinalizedPostData,
+  buildOptimisticPostData,
+}: {
+  postBeforeEdit: db.Post;
+  buildFinalizedPostData: () => Promise<domain.PostDataFinalizedEdit>;
+  buildOptimisticPostData: () => domain.PostDataFinalizedEdit;
+}) {
   logger.trackEvent(
     AnalyticsEvent.ActionStartedDM,
```
davidisaaclee, line 450: This seems wrong to me – anyone have context?

dnbrwstr (approved):
Code lgtm!

davidisaaclee:
I ran e2e locally (after merging in `develop`), and the only failure was `e2e/invite-service.spec.ts` – which I think is probably unrelated 😬

PR #4995, opened 2025-08-07
groups: add ban button for non-public group types, add e2e tests for bans

patosullivan:
## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications


packages/app/ui/components/ActionSheet.tsx
```diff
@@ -142,6 +143,18 @@ const ActionSheetComponent = ({
   // For popovers, use a more conservative max height to ensure it fits in viewport
   const popoverMaxHeight = Math.min(maxHeight, height * 0.5);
 
+  // listen for escape key to close the sheet
+  // this is helpful for e2e tests
+  useEffect(() => {
```
dnbrwstr, line 148: Nit: would probably make sense to only bind the listener if the sheet is currently open.

packages/shared/src/store/groupActions.ts
```diff
@@ -1061,10 +1061,21 @@ export async function banUserFromGroup({
     return;
   }
 
-  if (existingGroup.privacy !== 'public') {
-    console.error('Group is not public', groupId);
+  // Prevent banning the group owner
+  if (contactId === existingGroup.hostUserId) {
+    console.error('Cannot ban group owner', groupId, contactId);
```
dnbrwstr, line 1066: Nit: would be more consistent to use `logger.error` here and below.

dnbrwstr (approved):
Lgtm! Two small nits.

latter-bolden (approved):
Change makes sense, glad we have a test for it too.

PR #4996, opened 2025-08-07
channels: request join upon hearing new membership

arthyn:
## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -669,7 +669,7 @@
 ::
 ++  watch-groups  (safe-watch /groups [our.bowl %groups] /v1/groups)
 ::  +take-groups: process group update
-::  
+::
 ++  take-groups
   |=  =r-groups:v7:gv
   =/  affected=(list nest:c)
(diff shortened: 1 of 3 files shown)


Fang- (changes requested):
Literally impossible to get boolean conditionals right first-try.

Downside here is that channel host will send "join pushes" even for channels that the new members don't necessarily have access to. This will just make the channel join fail on their end, essentially status quo but with extra work.

Resolving that by doing the permissions check on-the-spot when sending out the "join pushes" runs the risk of hitting similar race/info conditions where the channel host has learned about the new members but doesn't have their new roles yet, or things along those lines. This feels like it'd be turtles all the way down, and is another one on the list of reasons why splitting permissions/checks across agents/ships is a bad idea.

All of that is out of scope here though. The change as written (minus flipping the conditional) solves the common case just fine, and I'm not too worried about the pointless extra work it creates in edge cases.

Fang-:
Also shouldn't this target `staging`? I guess we'll merge everything back together before release but this fix is specifically for symptoms caused by what's on `staging`, right?

Fang- (approved):
"Exercising authority."

PR #4998, opened 2025-08-08
FacePile: add

jamesacklin:
## Summary

Adds a new `FacePile` component for use.

## Changes

Introduces a new `FacePile` component, which expects several `db.Contact[]`s and optionally a maximum number to show (defaults to 4). Over the maximum number of contacts to be shown, a "+n" badge is shown.

The component does not make any decisions which order the contacts should be stacked. For example, if we want the most recently-joined members to appear atop others, we will need to handle this before passing the contacts in. 

I did have to hard-code the pixel values for the badge height/width because even though `$2xl` is nominally 24px, the Avatar component insists that it's 26px at the `$2xl` size. I thought this to be something less than a cardinal sin. 

## How did I test?

I developed this in isolation in Cosmos. (Our Cosmos config has drifted somewhat given our recent changes with providers, so I'll issue a separate PR to show off what I did to resolve.)

I added a few unit tests to ensure that the overflow count computes correctly.

## Risks and impact

- Affects important code area: none

## Rollback plan

Don't use it?

## Screenshots / videos

iOS:

Screenshot 2025-08-08 at 8 56 51 AM

Web:

Screenshot 2025-08-08 at 8 57 19 AM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,23 @@ jobs:
         run: pnpm exec playwright install chromium --with-deps
         working-directory: ./apps/tlon-web
 
+      - name: Run Production Build Smoke Test
+        id: smoke-test
+        run: |
+          pnpm e2e:prod:smoke:force || (echo "::warning::Production smoke test failed, retrying..." && pnpm e2e:prod:smoke:force)
+        working-directory: ./apps/tlon-web
+        continue-on-error: false
+
+      - name: Upload Production Test Artifacts
+        if: failure() && steps.smoke-test.outcome == 'failure'
+        uses: actions/upload-artifact@v4
+        with:
+          name: production-smoke-test-results
+          path: |
+            ./apps/tlon-web/test-results/
+            ./apps/tlon-web/playwright-report/
+          retention-days: 7
+
       # For some reason, caching is breaking with the new piers.
       # TODO: fix this.
       # - name: Cache E2E Piers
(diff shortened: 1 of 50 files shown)


packages/app/ui/components/FacePile/FacePile.tsx
```diff
@@ -0,0 +1,53 @@
...
+export interface FacePileProps {
+  contacts: db.Contact[];
+  maxVisible?: number;
+}
+
+export const FacePile = React.memo(function FacePileComponent({
+  contacts,
+  maxVisible = 4,
+}: FacePileProps) {
+  const visibleContacts = contacts.slice(0, maxVisible);
```
davidisaaclee, line 17: `.slice()` will allocate a new array on each render, which is small but worth a `useMemo` imo (especially if we'll use this on a big list of chats or something)

(this _might_ cause re-renders of user avatar SVGs as well? which would be very expensive in bulk. I'm like 70% sure that React is smarter than that.)

```suggestion
  const visibleContacts = useMemo(() => contacts.slice(0, maxVisible), [contacts, maxVisible]);
```

(you might need to fix prettier/lint if accepting that)

packages/app/ui/components/FacePile/FacePile.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+const mockContacts: db.Contact[] = [
+  createMockContact('~rilfun-lidlen', 'James'),
+  createMockContact('~latter-bolden', 'Brian'),
+  createMockContact('~fabled-faster', 'Ed'),
+  createMockContact('~nocsyx-lassul', 'Hunter'),
+  createMockContact('~solfer-magfed', 'Dan'),
+  createMockContact('~ravmel-ropdyl', 'Galen'),
+];
+
+// Helper function to test FacePile logic without rendering
```
davidisaaclee, line 32: this isn't actually sharing code with the FacePile implementation – probably good to share so that the test breaks if the component breaks!

davidisaaclee (changes requested):

PR #4999, opened 2025-08-08
Robustly handle inconsistent gangs state during the migration

mikolajpp:
## Summary

We prune bad gangs entries for groups that are already joined and initialized. Without this, a migration running on a ship which contains "join in progress" despite being already joined to the group would trigger a re-join and the ship would have left the group, deleting its local state.

## Changes

We introduce two changes to make the agent robust against this and similar problems in the future. 
1. Most immediately, we prune the gangs state removing entries for groups that are already joined  and had been initialized. 
2. We handle a watch-nack for a group robustly – if a group is already joined and initialized, we don't leave the group, but only log a critical error. 

By the way, we improve `+se-compat-send-invites` by initiating negotiation before issuing an invite poke. (Also fixing a `+roll` bug where `+se-core` `+emit` was not properly referenced)

## How did I test?

Two fake ships running old groups were setup to reproduce this bug.
~zod hosted four groups: two public groups A, B, and two private groups C D. ~dev joined all four groups.
Next, ~dev gangs state was poisoned. A gang was inserted for public group A and the progress set to %added. Private group C was poisoned in the same way.
The group host ~zod was then migrated. After ~zod migrated, ~dev was migrated.

It was observed that immediately after ~dev completed migration, poisoned groups A and C disappeared from the state, and an error message was shown in the client, thus reproducing the issue. The good groups B and D were migrated without problems.

The same scenario was then repeated with the fix deployed. This time all four groups migrated without problems, and the poisoned gangs state was cleared.

## Risks and impact

- Safe to rollback without consulting PR author? Rolling back risks triggering the bug at some unfortunate ships.
  - [ ] Onboarding
  - [x] State / providers
  - [x] Message sync
  - [x] Channel display
  - [ ] Notifications

Fixes TLON-4663

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -840,7 +840,13 @@
       ^-  (unit (pair flag:g gang:v6:gv))
       ?~  group=(~(get by groups) flag)
         `[flag gang]
-      ?:  ?=(%pub -<.u.group)  ~
+      =*  net  -.u.group
+      ?:  ?=(%pub -.net)  ~
+      ::  drop gang for groups that are already initialized,
+      ::  otherwise we risk losing user data during migration
+      ::  of group join flows. see /load/v7/foreigns case in +arvo.
+      ::
+      ?:  load.net  ~
       `[flag gang]
     :-  caz
     :*  %7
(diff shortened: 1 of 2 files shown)


arthyn (approved):
lgtm

Fang- (approved):
This change looks good. I tried writing a test for this but somehow couldn't get it to fail for the code prior to these changes, the group stayed in state just fine...

mikolajpp:
> This change looks good. I tried writing a test for this but somehow couldn't get it to fail for the code prior to these changes, the group stayed in state just fine...

Can I see the test? It does not fail during `+on-load`, but after the agent received the relevant timer event.

PR #5004, opened 2025-08-08
channels: /range endpoint tweaks

Fang-:
## Summary

Fixes two small issues with the `/range` endpoint identified by @latter-bolden.

## Changes

- Give `newest: 0` instead of `newest: 1` if a channel has no posts yet.
- Don't respect messages with `seq=0` while traversing the message list for building the `/range` response.

## How did I test?

Tested manually by running queries and observing the changes. 

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Could just revert if needed. This endpoint isn't used in production yet.

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -2745,7 +2745,7 @@
           `key:(head older)
         =/  count  (wyt:on-v-posts:c posts)
         =/  latest=@ud
-          ?~  latest=(ram:on-v-posts:c posts.channel)  1
+          ?~  latest=(ram:on-v-posts:c posts.channel)  0
           ?-  -.val.u.latest
             %&  seq.val.u.latest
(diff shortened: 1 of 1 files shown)


latter-bolden (changes requested):
After committing the desk, I still seem to be seeing `newest` at 1 on an empty channel:

```
{
"newest": 1,
"posts": {},
"total": 0,
"newer": null,
"older": null
}
```

Confirmed it bumped with only this file modified. Hash on my test ship matches the local node where the desk was installed.

@Fang- if you curl `/range` from a freshly created channel, you're seeing it at 0?

latter-bolden:
Thanks for the fix @Fang- , confirmed it works ✅

PR #5006, opened 2025-08-11
groups: checking from self higher up to avoid fact return src mixup

arthyn:
## Summary

Fixes issue found in onboarding testing with non-host admin created invites. 

## Changes

- moved `?>  from-self` check to be in more idiomatic place

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -144,8 +144,8 @@
       [~.groups^%1 ~ ~]
     %-  my
     :~  %groups^[~.groups^%1 ~ ~]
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Noting for the record that the implicit, actually-important change here is that the `from-self` check no longer fires when `+go-a-invite` gets called in response to a `%group-token` fact on an `/invite/[ship]/token` subscription.

As discussed out of band, this looks good and appropriate!

PR #5012, opened 2025-08-12
channels: add missing mark for /said paths

mikolajpp:
## Summary

Adds a missing mark `%channel-denied` that can be received on all `/said` endpoints.

## How did I test?

Compiled.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

PR can be reverted.

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -91,22 +91,22 @@
       ::  facts
       ::
       :~  [/ %channel-response %toggle-post ~]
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Good catches. We indeed just pass on what we get from the channels-server, which might be `%channel-denied` for all of these.

PR #5017, opened 2025-08-12
groups: resub instead of crash when inconsistent

Fang-:
## Summary

When we encounter a group update that confuses us (ie channel added to section we don't know about, unknown roles added as readers, etc.), resubscribe to the group to fetch all of its state anew. (Instead of crashing, resubscribing at last-known, and crashing again, continuously.)

## Changes

Logic for processing group updates sometimes checks whether the update makes sense for the state that we have. Previously, if an inconsistency was detected, it would crash.

Crashing results in a %kick, which we handle by resubscribing. But we resubscribe at the same timestamp, causing us to receive the same update, bringing us back where we started.

Here, we instead do a controlled unsubscribe and resubscribe in such a way that we fetch the full log for the group. Since this _should_ always start with a %create update, we will discard our (likely inconsistent) group state and rebuild from scratch. That should bring us back in line with the group host's state, resolving inconsistencies.

## How did I test?

Untested, putting up as draft for discussion.

I can probably contrive a case that would hit this in automated agent testing, but ideally we'd test this on a real-world affected ship also. See also risks below.

## Risks, impact, discussion

- Applying the full log will send a bunch of responses, right? Could result in confusing UX if the client is open and rendering those, essentially time-traveling/speed-running the group history.
- Applying updates (whether one-offs or the whole log) may have side-effects. As written, those will probably re-fire even though they're historical.
  - This includes not just calls to `+go-activity`, but also channels-server poking in `+go-u-role`, channel joining in `+go-u-channel`, and `+go-leave` in `+go-u-seat`, which would be particularly disastrous.
  - We might say, any subscription with `*@da` as the timestamp just gets a single artificial `%create` event... but you risk missing executing side-effects that you actually _do_ need to execute.
    - But you can't really know what you're missing, or what you need to do, because your state is inconsistent. From that angle, the situation is unsalvageable, isn't it? Doomed if you live-replay the whole log, doomed if you don't...
- If the group log itself is borked (which probably wouldn't happen organically but you can envision that happening through migration logic or w/e) then this will infinitely retry like before, except with bigger payloads...
  - Can/should group hosts run sanity checks on their logs periodically? Probably out of scope here even if we do want it.

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Revert.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -284,8 +284,7 @@
       =+  ;;(=flag:g +.q.vase)
       ?.  |(=(our src):bowl =(p.flag src.bowl))
         cor
-      ?~  g=(~(get by groups) flag)
-        cor
+      ?.  (~(has by groups) flag)  cor
       go-abet:(go-safe-sub:(go-abed:go-core flag) |)
     ::
         %pimp-ready
@@ -638,13 +637,7 @@
     ?.  ?=([%epic ~] wire)  cor
     =^  caz=(list card)  subs.cor
       (~(unsubscribe s [subs bowl]) wire dock)
-    =.  cor  (emil:cor caz)
-    ::  force leave
-    (emit:cor [%pass wire %agent dock %leave ~])
-  ::  v4 -> v5: initialize $group .active-channels
-  ::
-  =?  cor  ?=(%4 -.old)
-    (emit [%pass /load/active-channels %arvo %b %wait now.bowl])
+    (emil:cor caz)
   =?  old  ?=(%4 -.old)  (state-4-to-5 old)
   =?  old  ?=(%5 -.old)  (state-5-to-6 old)
   =^  caz-6-to-7=(list card)  old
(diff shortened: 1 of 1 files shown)


desk/app/groups.hoon
```diff
@@ -3060,7 +3102,7 @@
     |=  =update:g
     ^+  go-core
     =?  net  ?=(%sub -.net)
-      ?>  (gte time.update time.net)
+      ?>  (gte time.update time.net)  ::REVIEW  restart subscription instead?
```
arthyn, line 3105: I think I'd be ok with this

arthyn (approved):
lgtm @mikolajpp lets talk about permissions next week

mikolajpp (approved):
We have effectively reviewed as part of the other PR. Approved.

PR #5020, opened 2025-08-13
chat: /changes endpoint

Fang-:
## Summary

Add a cross-chats /changes endpoint to chat agent. To achieve this, we must store an update-timestamps ordered-map in state, just as channels agent already does.

## Changes

- 6220d8c tracks the change timestamps. We don't currently trim this, so this will grow indefinitely. I think channels agent might suffer from the same. The diff here is surprisingly small, kinda nice.
- 1bcb254 implements the endpoint, in the same way that the one for channels was implemented on the `m/changes-endpoints-redux` branch was.

## How did I test?

Manually. Made requests to `/changes`, did new posts, edits etc, re-requested.

## Risks and impact

- Safe to rollback without consulting PR author, but would require state adapting.
  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Would need to write a state reversion, but not complicated. Could even just `*` out the mop.


arthyn (approved):
lgtm

PR #5021, opened 2025-08-14
groups: normalize channel kind in active-channels sync

mikolajpp:
## Summary

Some third-party developers (notably %quorum) seem to be using groups to host their custom content unsupported by channels. They do this by modifying the group-side nest kind. This causes crashes on ships hosting such content.

## Changes

We normalize the nest kind and cast it to the channels $kind type, ensuring the scry
is going to succeed.

## How did I test?

I have setup two ships and created a group with multiple channels. I have confirmed that the `active-channels` set is correctly updated. I have then setup a new channel, which was correctly reflected in `active-channels` on the subscriber.

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

Revert.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -3424,9 +3424,8 @@
       =/  pre=path
         /(scot %p our.bowl)/channels/(scot %da now.bowl)
       =/  active
-        ?.  ?=(?(%diary %heap %chat) p.nest)  |
-        =/  =kind:d  p.nest
-        .^(? %gu (weld pre /v3/[kind]/(scot %p p.q.nest)/[q.q.nest]))
+        ?.  ?=(kind:d p.nest)  |
+        .^(? %gu (weld pre /v3/[p.nest]/(scot %p p.q.nest)/[q.q.nest]))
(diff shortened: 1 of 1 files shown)


desk/app/groups.hoon
```diff
@@ -3423,7 +3423,10 @@
       ::
       =/  pre=path
         /(scot %p our.bowl)/channels/(scot %da now.bowl)
-      =/  active  .^(? %gu (weld pre /v3/[p.nest]/(scot %p p.q.nest)/[q.q.nest]))
+      =/  active
+        ?.  ?=(?(%diary %heap %chat) p.nest)  |
+        =/  =kind:d  p.nest
```
Fang-, line 3428: Maybe a direct type-check would work too:

```suggestion
        ?.  ?=(kind:d p.nest)  |
```

Fang- (approved):
Sweet, this should resolve the crashes caused by this.

Small restructuring suggestion, but good to go either way. Thank you!

PR #5028, opened 2025-08-15
channels: content negotiation fixes

Fang-:
## Summary

Ensure we fetch sequence numbers and tombstones, even if the channel host doesn't update as fast as we do.

## Changes

In #4889, we had made channels explicitly fetch sequence numbers from the host, to make sure we were in sync. In #4963, we made it do the same for tombstone data.

The logic for those two flows was/is identical, but had a serious flaw: when sending out the poke, we don't check whether lib negotiate thinks we match.  
This isn't a problem for the logic itself, it's robust against poke-nacks. But negotiate will crash the event, because sending a poke to a mismatching peer is considered an implementation error.  
That crash would kill the timer event and call `+on-fail`, which contains no recovery logic for this scenario (and couldn't even if it wanted to). As a result, some channels would never get their sequence numbers fetched.

Here, we resolve this problem by simply checking before we poke, and re-setting the timer if we don't match. Closes TLON-4701.

Because #4889 has already gone out to livenet, we must resolve the cases where ships hit this in practice. We move the sequence nr flow starting from the 8-to-9 migration into the 9-to-10 migration. This may result in duplicate timers for any given channel, but that's not the end of the world. Closes TLON-4702.

We also add `%request-seqs`/`%request-tombs` pokes for manual restarting of these flows, just in case. Please do not call out to that indiscriminately lest you end up with a bunch of identical refetching flows.

## How did I test?

Manually on a fakenet. We'll want to hit the same setup that led us to discover this the other day, too. cc @latter-bolden

## Risks and impact

- Technically safe to rollback without consulting PR author, but if we do we may start dropping timers again.
  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Consider keeping 70eae57 when rolling this back, it's the critical piece here.

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -741,6 +741,24 @@
         [~ %| *]  (run-import p.u.pimp)
       ==
     ::
+        [?(%request-seqs %request-tombs) nest:c]
+      ::NOTE  if the poke we send from this gets nacked, we will set a timer
+      ::      to retry. if the ship/situation is healthy there will likely
+      ::      already be a timer for retrying this poke. having multiple timers
+      ::      isn't the end of the world, it will just do the same work twice,
+      ::      with any subsequent invocations being no-ops. nonetheless,
+      ::      probably best to not run this indiscriminately. check for
+      ::      presence of the relevant timers for this agent first! patience.
+      ::NOTE  we set the timer instead of sending the poke right away. timer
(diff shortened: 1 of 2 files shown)


arthyn (approved):
lgtm just one question

PR #5029, opened 2025-08-15
chat: resequence chats to restore missing numbers

arthyn:
## Summary

Fixes TLON-4703. We found this when testing the tombstones migration. Because chats were sequenced before the tombstone work, any deletes that have happened since will create gaps in the sequence. This runs the same sequencing logic as the previous update, but now that we have tombstones we should never drop a number.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -453,14 +453,15 @@
   ++  pact-9-to-10
     |=  pact:v5:cv
     ^-  pact:v6:cv
-    =;  nu-wit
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Oh right, I guess we need to do a list iteration regardless.

But yes, this looks like the change. Approved!

PR #5033, opened 2025-08-18
groups: properly handle group resubscription

mikolajpp:
## Summary

This PR adds some changes necessary to make group subscription restart work correctly. 
In particular, we account for the fact that channel permissions are stored in two separate places: read permissions in the groups agent, while write permissions are stored in the channels-server.
As such, when a group subscription is re-established, we need to make sure that the local channels-server of each group member does a deep permission recheck to account for changes in read permissions (should we kick subscribers from the channel?) and write permissions (should we remove roles no longer present in the group?)

## Changes

1. Added two more cases in which we restart the subscription.
3. Modified channels-server to do a full permissions recheck when a group create event is received. That this is going to be triggered on a first join too is an advantage – it might help fix permissions in cases where the groups agent had been nuked, while channels server had not.
4. Removed unnecessary %del-writes pokes in groups that were emitted when a role has been deleted. channels-server internally triggers the same pokes when it receives an update from the groups agent.

## How did I test?

Testing in progress.

## Risks and impact

- Safe to rollback without consulting PR author? (Yes | No) Yes.
  - [ ] Onboarding
  - [x] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Revert if needed.

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -680,7 +680,8 @@
     `nest
   =*  r-group  r-group.r-groups
   ?+    r-group  cor
-    [%create *]  (recheck-perms affected ~)
+      [%create *]
+    (full-recheck-perms affected ~(key by roles.group.r-group))
   ::
         [%seat * %add *]
       (request-join flag.r-groups affected ships.r-group)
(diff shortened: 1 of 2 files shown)


Fang- (commented):
Approved.

PR #5040, opened 2025-08-19
client sync: /changes integration

latter-bolden:
Adds the new `syncLatestChanges` endpoint to the client. Right now it covers posts (channels & DMs), groups, and contacts. Takes a timestamp as input and returns any of those models that were added or updated since then. When it completes successfully, we update a kv stamp locally and only grab changes since that stamp on subsequent calls.

Along with `syncUnreads`, it's bundled into a helper (`syncSince` ) that we call anytime the app is opened or returns to foreground.

There's a few dev features smuggled in for now:
- haptic feedback for whether `syncSince` completed < 2 seconds
- ChatList header loading state is pegged to `syncSince` rather than the full session (doesn't track subs)

In light of this API, we'll need to refine how we're calling `syncStart` and likely split out some of it's sub-components for independent use. Optimizing the request pipeline will come as a follow on PR.

Notably this branch also includes our WIP `syncSince` background task.

### DIFF
--- a/packages/app/lib/backgroundSync.ts
+++ b/packages/app/lib/backgroundSync.ts
@@ -78,13 +78,6 @@ async function performSync() {
     timings.changesDuration = Date.now() - changesStart;
     logger.trackEvent('Background sync: changes complete', { taskExecutionId });
 
-    const latestPostsStart = Date.now();
-    await syncLatestPosts();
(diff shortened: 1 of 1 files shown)


dnbrwstr (approved):
Lgtm!

PR #5052, opened 2025-08-22
logs: improved function for debugging

arthyn:
## Summary

This PR adds new capabilities to the %logs agent so that we can use it to actually log to dojo. It implements debug logs that I used to debug sequence number migration. This lets us leave debugging logs around and enable them remotely in a production environment. We can also send these to posthog if we set the level lower.

One thing here is that I wasn't sure if we wanted this to actually be done through pokes since there are probably performance implications. The other option would be to slog in the log library but we'd have to scry the current log levels to know if we should slog.

## Changes

- adds state to %logs to track dojo and posthog levels
- adds action to set those levels
- now logs to dojo on every %tell and %fail event depending on levels set
- both dojo and posthog events can now be turned off by passing an empty unit for each

## How did I test?

Using the migration testing ships I added this and set dojo level to %dbug. This outputs migration attempts automatically based upon setting the level.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -729,7 +729,7 @@
     (recheck-perms affected roles.r-group)
   ::
       [%seat * %del ~]
-    ~&  "%channel-server revoke perms for {<affected>}"
+    ~&  "%channel-server: revoke perms for {<affected>}"
     %+  roll  affected
     |=  [=nest:c =_cor]
     %-  ~(rep in ships.r-group)


arthyn:
sorry about sigpams, meant to convert all of those to the new logs

Fang- (approved):
Still approved!

PR #5053, opened 2025-08-22
fix splash flash on iOS

davidisaaclee:
## Summary
fixes TLON-4734
Fixes two white flashes during splash sequence on iOS.

## Changes
- [Downgrade expo-splash-screen to 0.28.5 to avoid splash screen shift](https://github.com/tloncorp/tlon-apps/commit/e940705618fec303b6ec67cf0543d61f7109ce89)
  - The previously-installed version has a visual shift on iOS, apparently when moving from the native iOS splash screen (i.e. SplashScreen.storyboard as specified in target settings) to expo-splash-screen's runtime splash (which is created after native splash is shown so that it can manually be kept visible).

- [Rename splash screen background color set to match Expo supported color](https://github.com/tloncorp/tlon-apps/commit/5ecc1d6fb75a5a5d81a18610279628a05c4272c9)
  - This flash was being caused by the iOS native root view always having a white background color: in the frame where RN is not yet loaded, we see that white view.
  - expo-updates recreates the root view after `[ExpoAppDelegate customizeRootView:]` and assigns its own background color: https://github.com/expo/expo/blame/63e051de05bc62bfad0c11dea0c64d2e750051e4/packages/expo-updates/ios/EXUpdates/ReactDelegateHandler/ExpoUpdatesReactDelegateHandler.swift#L48
  - This means we can't use customizeRootView: to reliably set the background color to what we want.
  - That `ExpoUpdatesReactDelegateHandler` code attempts to set the background color to the named color set `SplashScreenBackground` - this isn't really documented (there's reference to this name in a expo-splash-screen guide but seems more like a suggestion in a managed app context?), but we can use this named color set to pass our own colors to expo-updates, and make the native root view's background what we want.

## How did I test?
Built for release via Xcode on a fresh simulator; logged into app; launched from killed app in dark and light mode.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan
git revert

## Screenshots / videos

https://github.com/user-attachments/assets/b94605a9-15fd-4df1-8d3f-6d882c61374f


latter-bolden (approved):
Likewise tested via Xcode without seeing any flash in dark mode.

PR #5057, opened 2025-08-25
groups: fixes for groups resubscriptions

mikolajpp:
## Summary

During the testing of the resubscription mechanism a bug was discovered. Similar to the recently discovered issue, `+safe-watch` would not actually resubscribe, since the subscription has not yet been left when a call is made. We fix this by introducing a `force` parameter, that, when enabled, causes `+safe-watch` to subscribe regardless of current subscription status.

## Changes

1. Modify `+safe-watch` to accept `force` argument.
2. Fix a type issue caused by incorrect use of `+by` with the `.active-channels` set, but also reset `.active-channels` completely when sending out a create event.
3. Make `.active-channels` state more reliable (while also fixing the bad state caused by (2)) by initializing it on every agent reload, and also initialize it upon group initialization.

## How did I test?

The resubscription mechanism was tested with two ships. Updates breaking the group states were repeatedly inserted into the subscriber ship, to observe that the state is reset and subscription correctly re-established.

## Risks and impact

  - [x] Onboarding
  - [x] State / providers
  - [x] Message sync
  - [x] Channel display
  - [x] Notifications

## Rollback plan

Can be reverted to go back to the previous subscription mechanism.

Note that something is broken with Github diff renderer: it renders changes to channels-server that were already applied by a previous PR, and some changes shown to groups are not real either. It would probably be best to view it commit by commit.


desk/app/groups.hoon
```diff
@@ -3463,7 +3451,7 @@
         ?.  ?=(kind:d p.nest)  |
         .^(? %gu (weld pre /v3/[p.nest]/(scot %p p.q.nest)/[q.q.nest]))
       =?  active-channels.group  active
-        (~(put by active-channels.group) nest)
+        (~(put in active-channels.group) nest)
```
Fang-, line 3492: Classic. I wonder if we can come up with a way to catch this stuff automatically/earlier.

desk/app/groups.hoon
```diff
@@ -3527,13 +3520,17 @@
       =.  go-core  (go-response %channel nest [%del-readers roles.u-channel])
       ?:  go-our-host  go-core
       ::
+      ?.  (has:by-ch nest)
+        (go-restart-updates 'missing channel')
```
Fang-, line 3562: Should be `'missing channel deleted readers'` or similar to distinguish this branch from others.

desk/app/groups.hoon
```diff
@@ -3527,13 +3520,17 @@
...
       =.  go-core  (go-channel-del-roles nest roles.u-channel)
       go-core
     ::
         %section
       =.  go-core  (go-response %channel nest [%section section.u-channel])
       ?:  go-our-host  go-core
       ::
+      ?.  (has:by-ch nest)
+        (go-restart-updates 'missing channel')
```
Fang-, line 3571: Should be `'missing channel modified section'` or similar to distinguish this branch from others.

desk/app/groups.hoon
```diff
@@ -3070,6 +3070,15 @@
...
     =?  go-core  !was-init
+      ::  initialize active-channels on group init
+      ::
+      =/  nests
+        ~(tap in ~(key by channels.group))
+      =?  active-channels.group  !=(~ nests)
+        %-  silt
+        %+  skim  nests
+        |=  =nest:g
+        .^(? %gu (channels-scry nest))
```
Fang-, line 3110: This does make us more dependent on correctly removing channels from channels agent state when we leave a channel, which kind of invalidates the point of the `active-channels` set, no?

@arthyn can I get a second opinion from you on the behavior/implications here?

desk/app/groups.hoon
```diff
@@ -2270,6 +2270,9 @@
     ::
         %set-admin
       =.  admins.group  (~(uni in admins.group) roles)
+      ::XX when a user becomes an admin, the group host should sync
+      ::   admin-restricted updates since the last sync.
```
Fang-, line 2285: As discussed, yes should do this.

desk/app/groups.hoon
```diff
@@ -1302,13 +1302,24 @@
...
 ::
 ++  safe-watch
   |=  [=wire =dock =path]
-  |=  [delay=? force=?]
+  |=  delay=?
+  ^+  cor
+  ?:  (~(has by wex.bowl) wire dock)  cor
+  =^  caz=(list card)  subs
+    (~(subscribe s [subs bowl]) wire dock path delay)
+  (emil caz)
```
Fang-, line 1316: This can just be a `((safe-force-watch wire dock path) delay)` call I suppose.

`+safe-force-watch` is a funny name. Should probably be `+force-watch` instead.

Fang- (commented):
Comments below. And an alternative implementation path to consider that might be cleaner:

Consider modifying the `wex.bowl` proactively to remove the subscription whenever we construct/issue a `%leave` card, so that the state during execution reflects the cards we are going to issue. That's a little bit treacherous/potentially confusing, but given that we already funnel all `%leave`s through a single arm (or handful or arms, w/e), this shouldn't be very high risk I don't think.

desk/app/groups.hoon
arthyn: I don't think channels being the de facto source of truth necessarily invalidates active-channels. The point was to have the set of joined channels tracked in the group model for the client. This is more convenient especially with the current changes endpoints and the data they return.

Regardless yes we'll have to make sure channels leaves when appropriate

desk/app/groups.hoon
```diff
@@ -1314,14 +1312,18 @@
   =^  caz=(list card)  subs
     (~(subscribe s [subs bowl]) wire dock path delay)
   (emil caz)
-::  +safe-force-watch: safely force watch a subscription path
+::  +safe-unwatch: safely unsubscribe from a path
```
Fang-, line 1315: It's not really "safe" in the way that "safe-watch" is "safe", so the naming here is a little confusing. Maybe better to do "eager-leave" or "proactive-leave" or something in that direction.

desk/app/groups.hoon
```diff
@@ -2793,40 +2788,30 @@
...
   ::
   ::    call this when encountering inconsistent state that suggests we need
   ::    to get back in proper sync with the group host.
   ::
+  ::    when .why is not null, the restart is considered abnormal and
+  ::    logged as a critical error.
+  ::
   ++  go-restart-updates
-    |=  why=@t
+    |=  why=(unit @t)
```
Fang-, line 2800: If we're gonna use it like this, maybe `error=(unit @t)` might be more clear?

desk/app/groups.hoon
```diff
@@ -3360,13 +3345,16 @@
...
+      =+  was-admin=(go-is-admin our.bowl)
       =.  seats.group
         %-  ~(rep in ships)
         |=  [=ship =_seats.group]
         =+  seat=(~(got by seats) ship)
         =.  seat
           seat(roles (~(uni in roles.seat) roles.u-seat))
         (~(put by seats) ship seat)
+      ?:  !=(was-admin (go-is-admin our.bowl))
+        (go-restart-updates ~)
```
Fang-, line 3357: You technically only need to do this in the "became an admin" case, right? Not in the "got demoted" one. In the latter all you achieve is dropping admin-only state that you might have learned about, but you can do that locally without touching the subscription if that's necessary. Same applies to all similar instances below.

desk/app/groups.hoon
```diff
@@ -3449,20 +3449,29 @@
...
       go-core
     ::
         %set-admin
       =.  go-core  (go-response %role roles [%set-admin ~])
       ?:  go-our-host  go-core
       ::
+      =+  was-admin=(go-is-admin our.bowl)
       =.  admins.group  (~(uni in admins.group) roles)
+
```
Fang-, line 3462: Blank line made its way in here.

Fang- (commented):
Some smaller comments and a question about approach, but I think this is very close!

desk/app/groups.hoon
```diff
@@ -2811,7 +2811,18 @@
...
+  ::
+  ++  go-lost-admin
+    ^+  go-core
+    =.  group
+      %_  group
+        tokens.admissions    ~
+        pending.admissions   ~
+        requests.admissions  ~
+      ==
+    go-core
```
Fang-, line 2825: This could just be a `%_  go-core` with the same edits I think, no need to `=.` in the context and produce it separately. Alternatively, this could be smaller if you `=.  admissions.group`, maybe.

desk/app/groups.hoon
```diff
@@ -3377,7 +3388,7 @@
           seat(roles (~(dif in roles.seat) roles.u-seat))
         (~(put by seats) ship seat)
       ?:  !=(was-admin (go-is-admin our.bowl))
-        (go-restart-updates ~)
+        go-lost-admin
```
Fang-, line 3391: A little bit implicit here that this will always be the "stopped being admin" case. The branch logic would also produce true for "started being admin", but we're in the `%del-role` branch, which would only ever result in stopping being admin.

Could maybe stand to have a little comment clarifying as such, but fine either way, just something to be cognizant of.

PR #5059, opened 2025-08-25
perf monitoring: fix issue with race condition on init

patosullivan:
## Summary
Fixes race condition in Firebase Performance monitoring initialization that caused app crashes for users with the `instrumentationEnabled` feature flag enabled.

The crash was caused by a race condition where the feature flag system (loading from AsyncStorage) would complete before the firebase performance monitoring endpoint was initialized. When `setEnabled(true)` was called without an endpoint present, the app would crash with "Performance monitoring endpoint not set".

I *think* that this latent race condition was exposed by recent changes to package dependencies (converting native dependencies to peer dependencies in commit 1c66142e8), which altered the module initialization order. The timing issue always existed but was masked by the previous init sequence.

## Changes

- Updated `setEndpoint` to check if monitoring was already enabled and apply it to the new endpoint when set
- Modified `setEnabled` to gracefully handle cases where endpoint is null by deferring the enable operation with a warning instead of throwing an error
- Refactored `InstrumentationProvider` to set endpoint first, then enable collection only if endpoint exists
- Added cleanup on unmount to set endpoint to null

## How did I test?

Manually tested app start up with `instrumentationEnabled` already enabled or disabled on both Android and iOS.

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Revert

## Screenshots / videos
N/A


packages/shared/src/perf.ts
```diff
@@ -23,14 +23,34 @@ const usePerformanceMonitoringStore = create<PerformanceMonitoringStore>(
...
+        );
+        set((prev) => ({
+          ...prev,
+          enabled,
+        }));
+        return;
       }
+
+      // If endpoint exists, propagate the enabled state to it
       endpoint?.setEnabled(enabled);
```
davidisaaclee, line 54: It seems like we could skip a lot of these checks by simply setting `state.enabled` here unconditionally, then doing `endpoint.setEnabled()` _only_ in the `useEffect` in the provider. (I think moving the `console.warn` into the `useEffect` would be reasonable as well.)

```
    setEnabled: (enabled) => set(prev => ({ ...prev, enabled })),
```

```
useEffect(() => {
  setEndpoint(endpoint);

   if (endpoint) {
      endpoint.setEnabled(state.enabled);
  }

  if (state.enabled && endpoint == null) {
    console.warn('...')
  }
}, [...])
```

patosullivan: You're right, that's much simpler. Updated.

davidisaaclee (approved):

PR #5060, opened 2025-08-25
limit poke size in http client

dnbrwstr:
## Summary

Limits the size of pokes to 512kb in http client in order to prevent massive post / avatar / icon payloads being stored on the server.

## Questions

- Is this the right size limit?
- Does it make sense to do this in http client?


latter-bolden:
@dnbrwstr probably wanna close and cherry pick the commit onto staging in a separate PR to avoid dragging along `develop` changes

latter-bolden (approved):
Code looks good.

Suspicious of threshold being too high, but fine either way

PR #5064, opened 2025-08-25
fix image validation

dnbrwstr:
## Summary

The regex we were using to determine whether the image url was valid only worked on mobile. This PR updates it to work for desktop and web.

## Changes

Updates valid image regex in profile and group info editors.


jamesacklin:
@dnbrwstr Same here, can you cherry-pick and target staging?

PR #5065, opened 2025-08-25
chat/channels/groups: limit payload sizes

arthyn:
## Summary

We were inadvertently allowing very large payloads to be submitted as post content or group metadata. This adds 256kb limits to each of these which should be plenty generous. We may need to adjust later but this should be good for now.

## Changes

- adds assertion to all the places where post content or group metadata is coming in

## How did I test?

Manually attempted to post very large bodies of text, each of which errored when attempted

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications


mikolajpp (commented):
This looks good. 

I will only note that, while this protects against injections from the client, it does not completely safeguards against malicious actor deliberately conjuring oversize payloads. For example, in groups anyone can report a post, and that allows injecting arbitrary atoms. I think we had discussions about this in the post when similar restrictions were put in place in the contacts agent.

If we are seriously worried about this being exploited at some point, we should devise a general strategy to protect our agents from accepting unreasonably big payloads. One way to do that would be measuring the poke payloads at top level, rather than having to restrict those on a variant by variant basis.

Fang- (commented):
Question about the chat/dm behavior, but looks good otherwise.

Fang- (approved):
Change looks good, no-opping instead of crashing should at least prevent senders from retrying. UX concerns remain, but that's kind of unavoidable. Maybe you want logging when you hit this case, but that's horrible to thread through as-structured...

Good to go as far as I'm concerned!

PR #5074, opened 2025-08-26
ops: release sequence integration

arthyn:
## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications


latter-bolden (approved):
Let's do this

PR #5079, opened 2025-08-28
scroller performance optimization

dnbrwstr:
## Summary

Assorted performance optimizations aimed at improving scroller performance. Still more that could be improved here.

## Changes

- Memoize assorted inline props
- Convert callbacks that depend on navigation (the identity of which can frequently change) to instead depend on a mutable navigation ref.
- Try tweaking FlatList config to render slightly less offscreen content at a slightly larger batch size

## How did I test?

Tested on web + sim


latter-bolden (approved):
Memoizations and ref conversions all seem reasonable. Runs fine in an XCode release build.

PR #5080, opened 2025-08-28
native: patch background sync module

latter-bolden:
## Summary

Switches background sync back to using `expo-background-task`, which is powered by the newer `BGTaskScheduler` api on iOS. Includes a patch for the dependency to help resolve the issues we were seeing previously.

## Changes
- use `expo-background-task` instead of `expo-background-fetch`
- update it to use `BGAppRefreshTask` instead of `BGProcessingTask` for more frequent scheduling
- remove the default delay on iOS
- prevent the task from being unscheduled whenever the app is opened
- ensure we re-schedule the task if it hits the expiration handler
- add debug logging for use in Xcode
- on Android, attempt to run every 20 min instead of once a day

## How did I test?

Followed the instructions [here](https://developer.apple.com/documentation/backgroundtasks/starting-and-terminating-tasks-during-development) to simulate. Note, you'll need a `release` build or it'll hang on starting the JS metro bundle.

Cut a release to TestFlight, see it executing on a regular interval in the logs.

## Risks and impact

Fixes TLON-4721


dnbrwstr (approved):
Great that this works! Would be good to reduce changes we're making to a minimum when merging in to develop.

PR #5082, opened 2025-08-29
channels: check log and post integrity

arthyn:
## Summary

These threads iterate over all channels and checks the integrity of either their logs or posts. The post checker takes an arg, either %channels or %channels-server, which allows you to check all channels you know about or just the ones you host.

## Changes

- added scry to get channels-server channels out as nouns
- added thread to check log integrity from channels-server
- added thread to check post integrity from either channels or channels-server

## How did I test?

Ran each thread on local copy of ~dabben-larbet and presented results as expected

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -692,6 +692,8 @@
   =?  +.pole  !?=([%v0 *] +.pole)
     [%v0 +.pole]
   ?+  pole  [~ ~]
+      [%x %v0 %v-channels ~]
+    ``noun+!>(v-channels)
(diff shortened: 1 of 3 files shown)


Fang- (approved):
As discussed, the utility of log integrity checking is very limited for older channel hosts. It's probably fine for recent/modern channels though. And the posts integrity checker is good, been useful in testing #5086.

PR #5085, opened 2025-09-02
lure: implement metadata sync

mikolajpp:
## Summary

The lure service has currently no mechanism to keep issued invitations up to date when invite metadata change.
This PR addresses this issue by introducing a mechanism to keep links up to date when group metadata or when relevant fields in the user's profile are changed.

Closes TLON-4732, TLON-4733.

## Changes

1. The reel agent now subscribes to groups, contacts agents  for updates and request the bait provider to update affected invite links.
2. To correlate lure token with associated group, the bait provider keeps a stable-id index, similar to the existing index in reel.
3. The links are updated in the following manner: (1) when a group metadata change, all associated invites stored by the bait provider and local reel agents are updated. The reel agent only updates the most recent group invite stored in the stable-id index. (2) When a user profile changes, all issued invites, both locally and in the bait provider are updated to reflect the new nickname or avatar. 

## How did I test?

Tested in a two-ship setup with a fake bait provider: the invite links are updated when group metadata or user profile are changed.

## Risks and impact

  - [x] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Since this is a state upgrade, a special downgrade PR would have to be crafted.


desk/app/reel.hoon
```diff
@@ -94,15 +159,26 @@
...
+  ?>  ?=(%5 -.old)
+  =.  state  old
+  :_  this
+  %-  murn  :_  same
+  ^-  (list (unit card))
+  :~  ?:  (~(has by wex.bowl) /groups [our.bowl %groups])  ~
+      `[%pass /groups %agent [our.bowl %groups] %watch groups-path]
+    ::
+      ?:  (~(has by wex.bowl) /contacts [our.bowl %contacts])  ~
+      `[%pass /contacts %agent [our.bowl %contacts] %watch contacts-path]
```
Fang-, line 181: If we're gonna run these unconditionally, might be nicer inside an `+inflate-io` to make that more explicit.

Fang-: And, do we not want to call `subscribe:s` here too? We do so for `%kick` handling. I assume we want the subscriptions to end up in library state consistently.

desk/app/reel.hoon
```diff
@@ -221,7 +301,126 @@
...
+        %+  hand
+          (~(get cy:t our-profile) %avatar %look)
+        (~(get cy:t profile) %avatar %look)
+      =/  [color=(unit @ux) cax=?]
+        %+  hand
+          (~(get cy:t our-profile) %color %tint)
+        (~(get cy:t profile) %color %tint)
+      =.  our-profile  profile
+      ::  nothing relevant has changed, skip the update
+      ?:  &(axe bax cax)  `this
```
Fang-, line 359: We poke the bait service conditionally here, but not for group metadata changes, because our profile may contain a bunch of irrelevant/not-displayed info, but for groups all the metadata gets used always?

desk/app/reel.hoon
```diff
@@ -221,7 +301,126 @@
...
+            'invitedGroupDescription'^description.meta
+            'invitedGroupIconImageUrl'^image.meta
+        ==
+      ::  update our group invite link
+      ::
+      =.  our-metadata
+        ?~  our-meta=(~(get by our-metadata) u.token)
+          our-metadata
+        %+  ~(put by our-metadata)  u.token
+        u.our-meta(fields (~(uni by fields.u.our-meta) fields.update))
```
Fang-, line 416: I don't remember/understand what this gets used for on the reel side. Do we track this exclusively for the "changed link providers" case (`%set-service`)?

desk/app/bait.hoon
```diff
@@ -184,7 +231,7 @@
         [~ ~]
       ?>  =('2' u.type)
       :-  `[%bite-2 token joiner metadata]
-      ?~  inviter-field=(~(get by fields.metadata) 'inviter')
+      ?~  inviter-field=(~(get by fields.metadata) 'inviterUserId')
```
Fang-, line 234: Why this change? afaict reel still sets `'inviter'`.

The evils of arbitrary/untyped metadata field names...

desk/app/bait.hoon
```diff
@@ -230,13 +277,54 @@
...
   ::
       %bait-undescribe
     =+  !<(token=cord vase)
-    `this(token-metadata (~(del by token-metadata) token))
+    =+  metadata=(~(get by token-metadata) token)
+    =?  stable-id  ?=(^ metadata)
+      ?~  id=(~(get by fields.u.metadata) 'group')
+        stable-id
+      (~(del ju stable-id) u.id token)
+    `this
```
Fang-, line 299: I think this is forgetting to also `(~(del by token-metadata) token)`.

desk/app/reel.hoon
```diff
@@ -221,7 +301,126 @@
...
+      %-  (fail:log %watch-ack 'failed to subscribe to groups' u.p.sign)
+      `this
+    ::
+        %fact
+      =+  !<(=r-groups:v7:groups-ver q.cage.sign)
+      ?.  ?=(%meta -.r-group.r-groups)  `this
+      =*  flag  flag.r-groups
+      =*  meta  meta.r-group.r-groups
+      =+  id=(rap 3 (scot %p p.flag) '/' q.flag ~)
+      ?~  token=(~(get by stable-id) id)  `this
```
Fang-, line 401: So, we're not sending new group metadata to bait if we ourselves haven't generated an invite link for the group. Is it possible that we haven't, but other group members have? We want to make sure to send the new metadata in that case too.

Fang- (commented):
Coupling the broader group metadata updating to updating a specific invite links (in `%bait-update`) feels a bit odd. I could be convinced it's fine, but help me understand the reasoning for it.

Also, in theory, these agents support talking to a third-party bait. In practice, I think with these changes that starts to be a degraded experience: the group host will only send metadata updates to its own service provider. Given both that, and that the whole setup (iiuc) depends on additional hosting-side service provisioning, we may consider locking it down to just our dedicated bait provider. Doing so might let us simplify the reel agent too, letting it forego tracking link metadata itself.

In any case, see comments and questions below.

desk/app/bait.hoon
Fang-: Gross. Should've just erred on the side of not changing it.

But it is what it is, I suppose...

desk/app/bait.hoon
```diff
@@ -230,13 +285,73 @@
       %bait-describe
     =+  !<([=nonce:reel =metadata:reel] vase)
     =/  =token:reel  (scot %uv (end [3 16] eny.bowl))
-    :_  this(token-metadata (~(put by token-metadata) token metadata))
+    ::  record the token metadata and add the token to the stable-id set
+    ::  if the group field exists.
+    ::
+    =.  token-metadata
+      (~(put by token-metadata) token metadata)
+    =+  id=(~(get by fields.metadata) 'invitedGroupId')
```
Fang-, line 293: Wondering if we should just have a type for this field, that's like `$?(%'invitedGroupId' %'inviterUserId' %'etc')` so that we don't accidentally invent new ones...

desk/tests/app/reel.hoon
```diff
@@ -0,0 +1,323 @@
...
+    +<.res
+  ?~  peek
+    |+~['invalid scry path' (spat path)]
+  ?~  u.peek
+    |+~['unexpected empty result at scry path' (spat path)]
+  ::XX there is a compiler bug here if .q face is missing
+  ::   from the peek. it compiles, but fails at runtime when
+  ::   the result is dereferenced.
+  ::
+  &+[!<(mold q.u.u.peek) state]
```
Fang-, line 77: You mean, on the cage as produced by the agent? Weird either way, even if that doesn't cast, because the `=/  peek` above casts and as such should have the faces on there always...

Fang- (approved):
Changes look good. Small questions/comments, but nothing big. Thank you for adding the tests!

desk/app/bait.hoon
mikolajpp: I have talked with @latter-bolden about this yesterday. We have agreed that it would be best to type up some spec for what we want from lure invites, and then introduce proper types at the backend.

In the meantime, however, we could indeed type those fields as a bandaid.

desk/tests/app/reel.hoon
mikolajpp: I think the problem was with
`=/  peek  +<.res`
which does not seem to result in a face, even though the type should be there, no?
Above would give a compiler error though – unfortunately I did not preserve a version which seemed to fail at compile time.

`=/  peek  out.p.res`, on the other hand, will work properly.

PR #5086, opened 2025-09-02
channels-server: sequence number re-recovery

Fang-:
## Summary

Previous attempts at re-sequencing a channel's posts fell into traps laid by our own faulty mental models of the state. Due to historic migrations, the log isn't guaranteed to be complete or consistent, making it a poor basis for recovery.

Here, we re-sequence all known existing posts and tombstones, in post-id order, on the server side. We additionally walk the log and update log entries to include/match those numbers, dropping any logs for which we're lacking vital information.

## Changes

Channels-server applies the migration as described above, and described in more detail in the `/lib/channels-utils` comments. We additionally move channels agent's "resyncing" logic for sequence numbers and tombstones to trigger (either for the first time or again) on a new state migration.

## How did I test?

Tested extensively on a big and old channel host (bolbex-fogdys), by instrumenting the repair code, manual inspection of state and scry endpoints, and by using the integrity checkers from #5082. Post integrity comes up clean for local channels in both agents. Log integrity, as described above, cannot be guaranteed.

## Risks and impact

78b7a89 and 6811db0 are recent additions. Please inspect them closely and reason through those scenarios. We are dropping state on the floor there, but I believe we only do so for "lost causes". Please confirm/deny.

- Technically safe to roll back, though you may need to adjust state migration code. Important to be mindful of the need to trigger sequence nr/tombstone re-syncing which is affected by that.
  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Consult PR author.

### DIFF
--- a/desk/lib/channel-utils.hoon
+++ b/desk/lib/channel-utils.hoon
@@ -1205,7 +1205,7 @@
   ^-  [(unit u-channel:v9:c) stop=? ~]
   =*  info  [nest=n id=id.update log=time]
   ?+  update  [`update | ~]
-      [%post * %set %& *]
+      [%post * %set *]
     ~|  info
     ?.  (~(has by seqs) id.update)
       ?:  (~(has in dead) id.update)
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

PR #5087, opened 2025-09-02
channels: re-do the seq/tomb sync migration/kickoff

Fang-:
## Summary

Due to internal pre-release version churn, we got into another unique one-off bad spot: some clients got the latest patch before the publishers they care about. Unlike on livenet, there was no protocol mismatch to prevent them from asking for a sync. So they got a seq/tomb sync from their host, who hadn't yet cleaned up their data. End result: bad data on the client, even after the host finally cleaned up.

Here, we bump the channels agent version _again_, cleaning up bad tombstones _again_, and asking the hosts for their seqs and tombs _again_. Assumption here is that all hosts on pre-release have now gotten the latest patch, so should have good data, and we can do a sync from them again right away.

## Changes

That's it, that's the change. We move it out of the previous state migration and into the new one.

## How did I test?

Two-step upgrade by hand on 

## Risks and impact

- Safe but not practical to rollback without consulting author.
  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

lmk.

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -209,7 +209,7 @@
   |%
   +$  card  card:agent:gall
   +$  current-state
-    $:  %12
+    $:  %13
         =v-channels:c
         voc=(map [nest:c plan:c] (unit said:c))
         hidden-posts=(set id-post:c)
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm

PR #5093, opened 2025-09-04
trigger stick-to-scroll-start when viewport height changes

davidisaaclee:
## Summary
towards TLON-4706

## Changes
`useStickToScrollStart` uses the opaque `scrollerContentsKey` param to trigger a "scroll-to-bottom if needed" action. This was previously only triggered when the post list data changed.
This PR changes `scrollerContentsKey` to also key off the viewport / window height into the post list – i.e. when the scroll view resizes (due to getting squeezed by the draft input growing), `useStickToScrollStart` triggers and sticks the scroll to bottom if needed.

This still doesn't work great on Firefox – I took some time but couldn't figure out why. See video below – there's a mysterious scroll that happens on the first input _after_ growing the draft input. My best guess is that it's a bug in the browser's native scroll anchoring implementation (although enabling the `FORCE_MANUAL_SCROLL_ANCHORING` flag did not fix).

## How did I test?
on Chrome, Safari, Firefox (FF has some bugs, visible in videos below):
- draft some messages that grow the input height, while locked to scroll bottom (see screen recording below)
- draft some messages that grow input while scrolled away from bottom - scroll does not stick to bottom
- scroll away from bottom, then back to bottom, then grow input height – scroll sticks to bottom

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan
git revert

## Screenshots / videos
| Browser | Before | After |
| --- | --- | --- |
| Chrome | <video src="https://github.com/user-attachments/assets/eb7abc02-b2fb-4b08-82fa-0488de6768e4" /> | <video src="https://github.com/user-attachments/assets/2f709b36-50f6-4188-a97a-41edcfedc3c5" /> |
| Firefox (**reverted** - see below) | <video src="https://github.com/user-attachments/assets/497a4cf6-b7c4-427b-8e72-f1d290de05ab" /> | <video src="https://github.com/user-attachments/assets/923c890d-061a-445d-8e4f-331c85ae3d13" /> |

Safari not shown, but appears to act the same as Chrome

### DIFF
--- a/packages/app/ui/components/Channel/PostList/PostList.web.tsx
+++ b/packages/app/ui/components/Channel/PostList/PostList.web.tsx
@@ -152,8 +152,14 @@ const PostListSingleColumn: PostListComponent = React.forwardRef(
       insideScrolledToBottomBoundary,
     ]);
 
+    const viewportHeight =
+      useTrackContentRect(scrollerRef.current)?.height ?? 0;
+    const scrollerContentsKey = React.useMemo(
+      () => [orderedData, viewportHeight],
+      [orderedData, viewportHeight]
+    );
     useStickToScrollStart({
-      scrollerContentsKey: orderedData,
+      scrollerContentsKey,
       scrollerRef,
       inverted,
       hasNewerPosts,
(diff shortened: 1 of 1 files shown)


davidisaaclee:
Looking at the before/after on Firefox, I think we'd be better off without this change in that browser, since this change causes the scroll to become unstuck from the bottom after sending a multline message. It's unfortunate that I can't figure out exactly what feature is interfering with the implementation, since that means we can't do a precise capability-based check to disable this change – but we could `navigator.userAgent.includes('Firefox')` (which might be incorrect in some cases, but would be correct in most).

PR #5094, opened 2025-09-04
ImageInput: show errors on upload, don't let users without configs attempt to upload

patosullivan:
## Summary

1) show error messages from uploads in ImageInput as a toast to the user
2) don't allow users who can't upload to even attempt to upload

## How did I test?

Tested on web and iOS sim, with ships that do and do not have a storage config.

## Risks and impact

## Rollback plan

Revert


patosullivan:
Looking into the e2e test failure, looks legit.

patosullivan:
> Looking into the e2e test failure, looks legit.

fixed

PR #5095, opened 2025-09-04
groups: clear out >3kb image metadata

Fang-:
## Summary

Various bugs might lead to big data being injected into a group's metadata fields. Certainly, we don't do any data validation/sanitization right now. In practice, this means some groups have huge image metadata fields containing the raw image data.

Fixes TLON-4753

## Changes

Here, we modify agent load logic, so that we always do a quick pass over the groups we have in state. If a group we host is affected, run a command that clears it, ensuring updates get sent to subscribers informing them of the change. If it's a foreign group, just clear the data locally, but don't send any subscription updates.

## How did I test?

Tested on my fakezod with contrived group metadata setup.

## Risks and impact

Maybe the threshold is too low? I've seen links (in posts) as big as 2.7kb. That's kinda nuts honestly, imo if your groups image url is >3kb you had it coming. (;

Realizing now the big data probably ended up in the logs on group hosts. This data is mostly inert, all it really does it take up space in the loom, I don't think we ever send it out in practice. At least, not after the "flattened first update" change from #5017, which notably hasn't been merged yet.

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Can just `git revert`.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -668,7 +668,11 @@
   ::
   %+  roll  ~(tap by groups)
   |=  [[=flag:g =net:g =group:g] =_cor]
-  ?.  (gth (met 3 image.meta.group) 3.000)  cor
+  ?:  ?|  =('' image.meta.group)
+          =('http' (end 3^4 image.meta.group))
+          =('#' (end 3 image.meta.group))
+      ==
+    cor
   ?:  =(p.flag our.bowl)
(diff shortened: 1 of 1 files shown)


arthyn (approved):
lgtm, probably want to do the same for channel posts too

mikolajpp (approved):
Looks good.

PR #5096, opened 2025-09-04
client: fix group preview loading

arthyn:
## Summary

Fixes TLON-4795. We were seeing what looked like looping when loading a group preview. It turns out the first time you load a group preview and for as long as the `useGroupPreview` hook is active, the query will fetch and modify the DB for groups, which will then trigger an invalidation, which then triggers the query to fire again. This separates the DB insertion to happen only once if we don't have the group in the DB originally.

## Changes

- removed unused imports
- changed flow of `useGroupPreview` to prevent query loops

## How did I test?

On my main ship, navigated to a channel with a group preview and observed only one preview call instead of constant fetches

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

### DIFF
--- a/packages/shared/src/store/dbHooks.ts
+++ b/packages/shared/src/store/dbHooks.ts
@@ -3,7 +3,7 @@ import {
   UseQueryResult,
   useQuery,
 } from '@tanstack/react-query';
-import { useMemo } from 'react';
+import { useEffect, useMemo } from 'react';
 
 import * as api from '../api';
 import { getMessagesFilter } from '../api';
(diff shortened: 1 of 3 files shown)


patosullivan (approved):
Makes sense to me

packages/shared/src/store/dbHooks.ts
```diff
@@ -381,16 +377,27 @@ export const useMemberRoles = (chatId: string, userId: string) => {
...
-      const [preview] = await syncGroupPreviews([groupId]);
-      return preview;
-    },
+    enabled: !!groupId && !group,
+    queryFn: async () => await api.getGroupPreview(groupId),
+    placeholderData: group ?? undefined,
   });
+
+  useEffect(() => {
+    if (groupId && !group && groupPreview) {
```
latter-bolden, line 392: it seems like if we did have a pre-existing group preview and it's stale, we wouldn't update the DB here?

might be fine, we'll probably scoop up the fresh metadata eventually. But thinking it could also make sense to call `syncGroupPreview` but make sure we only do it once per hook mount

latter-bolden (approved):

PR #5100, opened 2025-09-05
changes: fallback to full sync if stamp is old & more telemetry

latter-bolden:
## Summary

- if our `/changes` stamp is more than 3 days old, fall back to a normal "full sync" on app open (i.e. `syncStart`)
- include breakdown of time to fetch vs time to write + a new check for whether your node was busy in `/changes` telemetry

## Changes
- modify `syncLatestChanges` accordingly
- new `http-api` method for using the `/~_~/healthz` endpoint (ty @Fang- )

## How did I test?

Faked an old stamp, confirmed it falls back.

Ran multiple times until I got both `busy` and `available` results from the busy check.

## Risks and impact

  - Message sync


arthyn (commented):
approved

PR #5102, opened 2025-09-08
lure: update lure invite when group is deleted

mikolajpp:
## Summary

We currently don't invalidate lure invite links when a group is deleted. This leads to a terrible UX when a new user onboards through a lure link that leads to a deleted group. We fix this at the backend by setting `invitedGroupDeleted` field to `true` when a group is deleted. This PR must be complemented by a corresponding UI changes at the on-boarding page and, possibly, in the client.

Closes TLON-4762.

## Changes

1. Listen to `%delete` events in reel and update the bait provider accordingly.
2. Fix the groups agent to generate a `%delete` response. The caveat here was that on the group host, the server core would delete the group, thus giving the client no chance to act on the update.

## How did I test?

Implemented unit tests and verified that they pass.
Run with a local bait provider and observed the group deletion update being generated from the reel client.

## Risks and impact

- Safe to rollback without consulting PR author? No, as this impacts deletion logic in groups.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Can be rolled back only if the client had not relied on %delete responses.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -1456,6 +1456,11 @@
     ?>  =(p.flag our.bowl)
     =?  se-core  gone
       (se-update [%delete ~])
+    ::  if the group is about to be deleted, this our only
+    ::  chance to trigger +go-core update.
+    ::
+    =?  cor  gone
+      go-abet:(go-u-group:(go-abed:go-core flag) now.bowl %delete ~)
     %_  cor  groups
       ?:  gone
         (~(del by groups) flag)


desk/app/groups.hoon
```diff
@@ -1456,6 +1456,11 @@
     ?>  =(p.flag our.bowl)
     =?  se-core  gone
       (se-update [%delete ~])
+    ::  if the group is about to be deleted, this our only
+    ::  chance to trigger +go-core update.
+    ::
+    =?  cor  gone
+      go-abet:(go-u-group:(go-abed:go-core flag) now.bowl %delete ~)
```
Fang-, line 1463: Oh interesting, because the host _would_ sent out the `%delete` update, but the local `go` core would never handle it, because the group would've already been deleted from state here, so the update would no-op. Right?

Fang- (approved):
I think this is good. Please confirm whether I understand the groups agent problem correctly. If I do, good to go!

PR #5103, opened 2025-09-08
AttachmentSheet: allow pasting image data from clipboard

jamesacklin:
## Summary

Adds uploading an image to all attachment fields using the image data the user has on their clipboard. This is a nice-to-have users have been asking about for some time.

## Changes

- Upgrades the `@react-native-clipboard/clipboard` dep, since support in iOS for image data on the clipboard was recently added.
- Adds a `getClipboardImageData` callback to `AttachmentSheet`, which checks if there's an image in the phone's clipboard when the attachment menu opens. It then tries different ways to grab the image data (PNG, then JPEG, then a generic method).
- Adds a "Paste from Clipboard" option to the sheet if there's compatible image data on the user's clipboard.
- Handles uploading into a generic 300x300 image, then resizes as the image upload completes.

React Native's generic `TextInput` only supports pasting text data from the clipboard on iOS, so I figured this was a good-enough cross-platform solution.

Doesn't touch the Web path at all.
 
## How did I test?

Using the iOS simulator: 

- Copy an image from Photos or from the Web to the device clipboard (long press -> "Copy")
- Press the "+" button in a Chat channel and a DM; tap "Paste from Clipboard," confirm the image uploads (and isn't base64)
- Press the "+" button in a Gallery channel, tap "Image," then "Paste from Clipboard." Confirm the image uploads (and isn't base64)
- Edit your profile, change your avatar image, select "Paste from Clipboard," confirm the image uploads (and isn't base64)
- Edit a group, change the group icon, select "Paste from Clipboard," confirm the image uploads (and isn't base64)

Copying a moving GIF to the clipboard and pasting will convert the first frame to JPG, which is consistent with the current approach.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: image attachments

## Rollback plan

Revert. We have enough safeguards in place now that I'm not concerned about this base64-ing the upload.

## Screenshots / videos

Excuse the chugging, running a lot at the moment on the host Mac.

https://github.com/user-attachments/assets/208623aa-5e24-4bb4-ab6e-27db0273a3e8

### DIFF
--- a/packages/app/ui/components/AttachmentSheet.tsx
+++ b/packages/app/ui/components/AttachmentSheet.tsx
@@ -1,4 +1,3 @@
-import Clipboard from '@react-native-clipboard/clipboard';
 import { PLACEHOLDER_ASSET_URI, createDevLogger } from '@tloncorp/shared';
 import { Button } from '@tloncorp/ui';
 import * as ImagePicker from 'expo-image-picker';
@@ -7,6 +6,7 @@ import { Platform } from 'react-native';
 import { isWeb } from 'tamagui';
 
 import { useAttachmentContext } from '../contexts';
+import { getClipboardImageWithFallbacks } from '../utils';
 import { ActionGroup, ActionSheet, createActionGroups } from './ActionSheet';
 import { ListItem } from './ListItem';
 
@@ -40,27 +40,7 @@ export default function AttachmentSheet({
     mimeType: string;
   } | null> => {
     try {
-      const hasImage = await Clipboard.hasImage();
-      if (!hasImage) return null;
-
-      // iOS supports getImagePNG/getImageJPG but not the generic getImage
-      try {
-        const imageData = await Clipboard.getImagePNG();
-        return { data: imageData, mimeType: 'image/png' };
-      } catch (pngError) {
-        try {
-          const imageData = await Clipboard.getImageJPG();
-          return { data: imageData, mimeType: 'image/jpeg' };
-        } catch (jpgError) {
-          // If both specific methods fail, try generic getImage for Android compatibility
-          try {
-            const imageData = await Clipboard.getImage();
-            return { data: imageData, mimeType: 'image/png' };
-          } catch (genericError) {
-            return null;
-          }
-        }
-      }
+      return await getClipboardImageWithFallbacks();
     } catch (error) {
       logger.trackError('Clipboard access failed', { error });
       return null;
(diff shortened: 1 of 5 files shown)


packages/app/ui/components/AttachmentSheet.tsx
```diff
@@ -32,6 +33,77 @@ export default function AttachmentSheet({
...
+          };
+
+          attachAssets([clipboardAsset]);
+          onAttach?.([clipboardAsset]);
+        }
+      } catch (error) {
+        console.error('Error pasting from clipboard:', error);
+        logger.trackError('Error pasting from clipboard', { error });
+      }
+    }, 50);
```
dnbrwstr, line 104: Maybe add a comment for why this timeout is necessary?

packages/app/ui/components/AttachmentSheet.tsx
```diff
@@ -32,6 +33,77 @@ export default function AttachmentSheet({
...
+            duration: undefined,
+            exif: undefined,
+            base64: undefined,
+          };
+
+          attachAssets([clipboardAsset]);
+          onAttach?.([clipboardAsset]);
+        }
+      } catch (error) {
+        console.error('Error pasting from clipboard:', error);
```
dnbrwstr, line 101: Two error logging methods here probably redudant.

packages/app/ui/components/AttachmentSheet.tsx
```diff
@@ -32,6 +33,77 @@ export default function AttachmentSheet({
   const { attachAssets, clearAttachments, removeAttachment } =
     useAttachmentContext();
 
+  const [hasClipboardImage, setHasClipboardImage] = useState(false);
+
+  const getClipboardImageData = useCallback(async (): Promise<{
```
dnbrwstr, line 38: Doesn't need to be in component, could be in utils.

packages/app/ui/components/AttachmentSheet.tsx
```diff
@@ -32,6 +33,77 @@ export default function AttachmentSheet({
...
+    };
+
+    checkClipboard();
+  }, [showAttachmentSheet, getClipboardImageData]);
+
+  const pasteFromClipboard = useCallback(async () => {
+    onOpenChange(false);
+    setTimeout(async () => {
+      try {
+        const clipboardData = await getClipboardImageData();
```
dnbrwstr, line 68: Most of this could be in utils as `createAssetFromClipboard` or something.

dnbrwstr (approved):
Looking good! Some nits. Feel free to address + merge without re-review.

PR #5110, opened 2025-09-09
lure: implement typed metadata fields

mikolajpp:
## Summary

@Fang- suggested we type known metadata fields in lure agents. This PR implements just that.
Note that defining new metadata fields will not require state migrations: the new type is going to be compatible with the old metadata type. Deprecating and removing fields, however, will require a state migration.

## Changes

1. Type the key of reel `$metadata` type.
2. Perform intelligent migration in reel and bait agents.

## How did I test?

Compiled and run unit tests.

## Risks and impact

  - [x] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Rollback as needed.


Fang- (approved):
This looks like the changes I'd expect here. Nice to have a little list of them in the type definition now, "source is documentation" and all that. (;

PR #5114, opened 2025-09-10
groups: let +fi-core only emit a single foreign update

mikolajpp:
## Summary

The bulk updating in +fi-core has been suspected of unnecessarily burdening the client. We change the updating logic to only emit a single update for the relevant foreign group.

## Changes

1. Adjust logic in `+fi-give-update` to only emit a single foreign group fact.

## How did I test?

Run the change on two ships. Verified that received group invitations show up.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Rollback as needed.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -3935,6 +3935,7 @@
 ::
 ++  fi-core
   |_  [=flag:g foreign:g]
+  +*  foreign  +<+
   ::
   ++  fi-core  .
   ::  +fi-abed: init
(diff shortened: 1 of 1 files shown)


desk/app/groups.hoon
```diff
@@ -3949,21 +3950,19 @@
...
-    =.  foreigns  (~(put by foreigns) flag +<+)
-    ::TODO figure out the foreign lifetime logic after designining
-    ::     new endpoints for the client
-    ::
+    =+  old-foreign=(~(get by foreigns) flag)
+    =.  foreigns  (~(put by foreigns) flag foreign)
     =?  foreigns  ?=([~ %done] progress)
       (~(del by foreigns) flag)
-    =.  fi-core  fi-give-update
+    =?  fi-core  |(?=(~ old-foreign) !=(u.old-foreign foreign))
```
Fang-, line 3957:
```suggestion
    =?  fi-core  !=(old-foreign `foreign)
```

Fang- (approved):
Tiny nit for slightly simpler conditional, but otherwise good. Nice and small pr!

PR #5119, opened 2025-09-11
channels/server: drop likely-bad (image) links

Fang-:
## Summary

We may have raw image data links in our state. These bog down response serialization and should never have existed.

Here we trawl through posts and logs in state, and drop any links in the content that look out of the ordinary.

## Changes

We opt not to issue any updates from the server, or even bump revision numbers on affected posts. Clients will be dropping these links on their side too, and formally syncing these changes complicates the logic and increases risk on rollout. As written, this is a simple state change.

We take a somewhat broad definition of what a "good link" is. Anything that starts with `http`, `#` or `/` is fair game. Note that this will drop `mailto:` links and other such constructions. We may or may not want to tighten this heuristic to only drop bad links beyond a certain size.

## How did I test?

Wrote a test for the core link cleaning logic, and smoke-tested running the state update on a fakezod.

## Risks and impact

We may want to have a slightly tighter heuristic here, where we also check link size/length. Also, checking `%link`s in the first place may or may not be overkill. rfc @arthyn @mikolajpp 

- Can't fully roll back due to state version bump, but can always rip out the logic.


arthyn (approved):
lgtm

PR #5122, opened 2025-09-11
queries: flip ordering of groupChats timestamp prop

jamesacklin:
## Summary

Fixes TLON-4591 where group events other than content changes (new posts, new replies) would change the sidebar / Home Screen order of groups opaquely. 

## Changes

We send various activity events from the back-end for group-level changes within a group. These are not mutually exclusive from actual content changes, which produce unread badges. At the moment, we calculate the timestamp for a group's position in the sidebar by first checking its "updated" time, _then_ the last time the content within changed.

Users expect that anything that creates a badge on the group list item would float it to the top based on recency; therefore it doesn't make sense to reorder based on group-level change (i.e. membership changes).

Therefore, we need to use a slightly different timestamp logic:

1. If we have an invite to the group (which is actionable), we use `unread.updatedAt` so the invite floats to the top of the sidebar / groups list.
2. Join/leave/role changes are ignored because they don't create posts or badge the item.
3. Otherwise, we use `lastPostAt` to determine sidebar order.

## How did I test?

- Open Ship A and leave it on the Home screen
- Join a group that Ship A is in on Ship B
- The sidebar position of the group should not change
- Post in the group with Ship B
- Verify on Ship A that the position changed
- Invite Ship A to a group Ship B hosts
- Verify that the invite floats to the top of the sidebar

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: queries

## Rollback plan

Revert

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -787,6 +787,7 @@
   =/  =cage  [%channel-request-join !>(request)]
   [%pass /request-join %agent [ship %channels] %poke cage]
 ::
+++  size-limit  256.000  :: 256KB
 ++  ca-core
   |_  [=nest:c channel=v-channel:c gone=_|]
   +*  ca-posts  ~(. not posts.channel)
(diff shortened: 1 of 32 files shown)


dnbrwstr (approved):
Makes sense to me!

PR #5123, opened 2025-09-12
GroupMembersScreenView: add filter input

jamesacklin:
## Summary

I ran into TLON-4639 while attempting to give ~wittyr-witbes admin control of Tlon Studio, which has some 6K members. Scrolling through the entire list was not fun. Therefore, this fixes TLON-4639 by adding a filter input to the group members screen view. 

## Changes

- Creates Fuse.js objects for members, banned users, invited users, and users asking for entry
- Adds an input box to the top of the members list screen to allow for filtering by name or @p
- Allows the user to clear the input and reset the filter

## How did I test?

- Open a group with lots of members (Tlon Studio)
- Search for a known @p far down in the alphabet (I used ~wittyr-witbes)
- Verify that the exact match shows at the top of the list
- Enter a known nickname
- Verify that the exact match(es) show at the the top of the list
- Enter a partial @p or partial nickname
- Verify that the results make sense

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: group settings

## Rollback plan

Revert

## Screenshots / videos

https://github.com/user-attachments/assets/fa40c75a-6585-4206-add6-36d164347cc7

### DIFF
--- a/packages/app/ui/components/GroupMembersScreenView.tsx
+++ b/packages/app/ui/components/GroupMembersScreenView.tsx
@@ -273,7 +273,7 @@ export function GroupMembersScreenView({
         <View paddingHorizontal="$l" paddingBottom="$s">
           <TextInput
             icon="Search"
-            placeholder="Search by name or @p handle"
+            placeholder="Search by name or ID"
             value={searchQuery}
             onChangeText={setSearchQuery}
             spellCheck={false}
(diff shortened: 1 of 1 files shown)


patosullivan (approved):
Code looks good, functionality works well locally.

PR #5128, opened 2025-09-15
Settings: hide "Log out" on desktop/web

jamesacklin:
## Summary

Fixes TLON-4548 where the "Log out" button does nothing on the desktop app.

## Changes

We don't have the hosting auth cookies in the app db on the web, so the log out button doesn't do anything even if the user is hosted.

Here we just hide the button on web.

## How did I test?

Web:
- Log in to a hosted OR self-hosted ship
- Navigate on the web to Settings
- Confirm that the Log Out option does not exist

Mobile:
- Connect the app to your hosted OR self-hosted ship
- Navigate in the app to Settings
- Confirm that the Log Out option exists and returns you to the app login screen

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: settings

## Rollback plan

Revert

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -787,6 +787,7 @@
   =/  =cage  [%channel-request-join !>(request)]
   [%pass /request-join %agent [ship %channels] %poke cage]
 ::
+++  size-limit  256.000  :: 256KB
 ++  ca-core
   |_  [=nest:c channel=v-channel:c gone=_|]
   +*  ca-posts  ~(. not posts.channel)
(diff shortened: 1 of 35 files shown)


patosullivan (changes requested):
It looks like this removes the log out option for self hosted users on mobile.

I pulled this down and ran it in the sim against a self-hosted ship, looks like this:
Simulator Screenshot - iPhone 15 - 2025-09-16 at 08 49 47

jamesacklin:
Updated with a much simpler hide. ;)

patosullivan (approved):

PR #5131, opened 2025-09-16
ios: better notification logging

arthyn:
## Summary

This mimics #4890 fixing TLON-4788. We add rich errors for every case where displaying the rich notification could fail. We also log for every successful delivery of either the rich notification or the fallback.

## Changes

- added new NotificationLogger class to facilitate logging to PostHog
- refactored NotificationService to use richer errors
- added a catch-all try/catch to consume all errors and log to PostHog
- unified event messages between Android and iOS

## How did I test?

Built the app/extension and sent myself bad notifications, then checked PostHog to make sure they appeared correctly.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [X] Notifications
  - [ ] Other:


apps/tlon-mobile/ios/Landscape/NotificationLogProcessor.swift
```diff
@@ -0,0 +1,71 @@
...
+
+import Foundation
+import PostHog
+import ExpoModulesCore
+
+@objc
+class NotificationLogProcessor: NSObject {
+    @objc static let `default` = NotificationLogProcessor()
+    
+    private let appGroupIdentifier = "group.tlon.Landscape"
```
dnbrwstr, line 16: Maybe it's not critical, but I don't think this app group is correct. In info.plist it's `group.io.tlon.groups` and in the Info-preview.plist it's `group.io.tlon.groups.preview`. Ideally we'd probably pull it directly from there?

apps/tlon-mobile/ios/Landscape/NotificationLogProcessor.swift
```diff
@@ -0,0 +1,71 @@
...
+        userDefaults.removeObject(forKey: "notificationLogs")
+        userDefaults.synchronize()
+
+        NSLog("Successfully processed \(logs.count) notification logs")
+    }
+
+    @objc
+    func startPeriodicProcessing() {
+        print("configuring posthog")
+        let config = PostHogConfig(apiKey: "phc_6BDPOnBfls3Axc5WAbmN8pQKk3YqhfWoc0tXj9d9kx0", host: "https://eu.i.posthog.com")
```
dnbrwstr, line 59: Is there a way to pass this in from an env variable? It's supposed to differ per-environment, which means that we might have unexpected stuff showing up in prod logs.

apps/tlon-mobile/ios/Landscape/NotificationLogProcessor.swift
```diff
@@ -0,0 +1,71 @@
...
+        PostHogSDK.shared.setup(config)
```
dnbrwstr, line 60: Is this able to associate the logs with a user? Will be a lot easier to work with the data if we have that set.

packages/app/utils/posthog.ts
```diff
@@ -2,6 +2,7 @@ import crashlytics from '@react-native-firebase/crashlytics';
 import { useDebugStore } from '@tloncorp/shared';
 import * as db from '@tloncorp/shared/db';
 import PostHog from 'posthog-react-native';
+import { NativeModules, Platform } from 'react-native';
```
dnbrwstr, line 5: Unused?

apps/tlon-mobile/ios/Shared/Notifications/NotificationLogger.swift
```diff
@@ -0,0 +1,266 @@
+//
+//  NotificationLogger.swift
+//  Landscape
+//
+//  Created by Claude Code
```
dnbrwstr, line 5: I feel like we can remove this lol

apps/tlon-mobile/ios/Shared/Notifications/NotificationLogger.swift
```diff
@@ -0,0 +1,266 @@
...
+//
+
+import Foundation
+
+private let NOTIFICATION_SERVICE_ERROR = "Notification Service Error"
+private let NOTIFICATION_SERVICE_DELIVERED = "Notification Service Delivery Successful"
+
+class NotificationLogger {
+    private static let postHogApiKey = "phc_6BDPOnBfls3Axc5WAbmN8pQKk3YqhfWoc0tXj9d9kx0"
```
dnbrwstr, line 14: Like other comment, would prefer not to have to hardcode this. Also, is there any setup shared between this and the other location that we could simplify?

dnbrwstr (commented):
Overall lgtm, added a few nits. Would be good to get @davidisaaclee 's approval on the native parts, since he's most familiar with that area.

apps/tlon-mobile/ios/Shared/Notifications/NotificationLogger.swift
```diff
@@ -0,0 +1,266 @@
...
+
+            let properties = switch $0 {
+            case .error(let error):
+                getLogPayload(error: error)
+            case .delivery(let properties):
+                properties
+            }
+
+            return [
+                "timestamp": Date().timeIntervalSince1970,
```
davidisaaclee, line 113: nit: I think all the existing codepaths are fine, but it seems weird to assign a timestamp here – feel like it should be in the construction of the `LogEvent`. (what if we start batching in this layer?)

apps/tlon-mobile/ios/Shared/Notifications/NotificationLogger.swift
```diff
@@ -0,0 +1,266 @@
...
+        existingLogs.append(contentsOf: logEntries)
+
+        // Keep only recent logs (last 100 entries to prevent from growing too large)
+        if existingLogs.count > 100 {
+            existingLogs = Array(existingLogs.suffix(100))
+        }
+
+        // Save back to UserDefaults
+        userDefaults.set(existingLogs, forKey: "notificationLogs")
+        userDefaults.synchronize()
```
davidisaaclee, line 132: `synchronize` unnecessary

apps/tlon-mobile/ios/Shared/Notifications/NotificationLogger.swift
```diff
@@ -0,0 +1,266 @@
...
+
+        NSLog("Events saved to UserDefaults for later processing")
+    }
+
+    private static func prepEvents(distinctId: String, events: [LogEvent]) -> [[String: Any]] {
```
davidisaaclee, line 137: not clear what `distinctId` is

davidisaaclee: optional: i think this method, constructPostHogEvent, `distinctId`, timestamp comment above all point to something like:

```swift
// renamed from LogEvent
enum LogEventData { ... }

struct LogEvent {
  let userId: String
  let createdAt: Date
  let data: LogEventData
}

extension LogEvent {
  var asPostHogEvent: [String: Any] {
    // `constructPostHogEvent` body here
  }
}

File: apps/tlon-mobile/ios/Shared/Notifications/NotificationLogger.swift [hunk 10]
```diff
@@ -0,0 +1,266 @@
...
+    return payload
+}
+
+enum LogEvent {
+    case error(NotificationError)
+    case delivery([String: Any])
+}
+
+// Custom error types for iOS notifications
+class NotificationError: NSError {
```
davidisaaclee, line 186: nit: in case you revisit this from my comments above, I think the swifty way to do this is to adopt the `Swift.Error` protocol. then you could avoid the class weight and move to struct/enum if you want. and if you want nicer error messages in dev, you could implement the `description` field (iirc)

```swift
struct ActivityEventFetchFailed: Swift.Error {
  let uid: String
  let underlyingError: Error? = nil

  var description: String { underlyingError?.description ?? "Activity event fetch failed" }
}

// or

enum NotificationError: Swift.Error {
  case activityEventFetchFailed(uid: String, underlyingError: Error?)
  // ...
}
```

apps/tlon-mobile/ios/Podfile
```diff
@@ -63,11 +63,13 @@ end
 target 'Landscape' do
   Pod::UI.puts 'Building Landscape'
   shared(podfile_properties)
+  pod 'PostHog', '~> 3.0'
 end
 
 target 'Landscape-preview' do
   Pod::UI.puts 'Building Landscape-preview'
   shared(podfile_properties)
+  pod 'PostHog', '~> 3.0'
```
davidisaaclee, line 72: should probably put this in `shared` below

davidisaaclee (changes requested):
mostly nits/optional (tagged appropriately). it's a little hard for me to fully grasp the iOS notification flow now.

apps/tlon-mobile/ios/Shared/Notifications/NotificationLogger.swift
```diff
@@ -0,0 +1,124 @@
...
+        }
+        
+        // Send directly to PostHog REST API
+        guard let url = URL(string: "\(postHogHost)/batch/") else {
+            print("Invalid PostHog URL")
+            fallbackToUserDefaults(userId: userId, events: events)
+            return
+        }
+
+        var request = URLRequest(url: url)
```
davidisaaclee, line 31: optional: This might be more clear if you pulled out the "send HTTP request" into a separate throwing method – then this body would look like:
1. collect data for request
2. try send http request
3. if failed, fallback to user defaults

apps/tlon-mobile/ios/Shared/Notifications/NotificationLogger.swift
```diff
@@ -0,0 +1,124 @@
...
+
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = .iso8601
+        
+        let logEntries: [Data] = events.compactMap({
+            do {
+                let event = LogEvent(userId: userId, data: $0)
+                return try encoder.encode(event)
+            } catch {
+                print("Unable to encode LogEvent")
```
davidisaaclee, line 97: optional: if we're going to log here, I think we should give more info (like the `event`)

otherwise, I think you could make this terser by just doing `events.compactMap { try? encoder.encode(LogEvent(userId: userId, data: $0)) }` (and maybe checking that `events.count == logEntries.count`)

apps/tlon-mobile/ios/Shared/Notifications/NotificationLogger.swift
```diff
@@ -0,0 +1,124 @@
...
+            print("Failed to serialize PostHog event data: \(error.localizedDescription)")
+            fallbackToUserDefaults(userId: userId, events: events)
+        }
+    }
+
+    // Fallback to UserDefaults if direct PostHog call fails
+    private static func fallbackToUserDefaults(userId: String, events: [LogEventData]) {
+        print("Using UserDefaults fallback for events")
+
+        guard let userDefaults = UserDefaults(suiteName: "group.tlon.Landscape") else {
```
davidisaaclee, line 84: shouldn't this be the app group UserDefaults?

davidisaaclee (changes requested):
- the `CodableValue` stuff is worth another look in case you can make it terser (which would be significant if you think about the effect multiplied across all log statements) – but fine if you can't make headway there, at least it's safe?
- `api_key` on `asPosthogEvent` – dropping this if unnecessary gives us one less dependency to maintain piping the api key down (I don't really love the way we're piping the api key [although I get the rationale and can't think of a good alternative], and it'd be nice to know if we can drop it).
- noted a lingering `UserDefaults.init(suite:)` call that is using a static suite name (so it'd be wrong on preview)

otherwise, I added a bunch of optional comments – this code is still hard for me to read for some reason, which might be just me having a stupid day, but figured I'd at least try to figure out what was making it hard for me to follow.

apps/tlon-mobile/ios/Shared/Models/NotificationLogging.swift
```diff
@@ -0,0 +1,175 @@
...
+    // Convert CodableValue to Any for PostHog
+    var anyValue: Any {
+        switch self {
+        case .string(let value): return value
+        case .int(let value): return value
+        case .double(let value): return value
+        case .bool(let value): return value
+        case .null: return NSNull()
+        }
+    }
```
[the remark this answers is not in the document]
arthyn: Seems like you have to do something like this https://www.fivestars.blog/articles/codable-swift-dictionaries/ or use https://github.com/Flight-School/AnyCodable

PR #5137, opened 2025-09-18
groups: robustly handle a missing channel section & other fixes

mikolajpp:
## Summary

Some group subscribers have a corrupted state where a channel is placed in a non-existent section. This prevent a channel section update from going through. We fix this by robustly handling a current missing channel section.

Another instance of invalid channel nests inserted by third-party developers was caught and fixed.

We also drive-by-fix the logs agent to pass the unit tests broken by recently merged changes: the default volume was not initialized.

## Changes

1. We avoid a `+jab` crash and replace it with a two step get, put to update a channel section.
2. Ignore channels with invalid nest kinds during active-channels loading.
3. Fix the logs agent to initialize its logging volume to `%info` by default.

## How did I test?

Compiled and run the unit tests.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Can be reverted at the cost of triggering desync on some ships.

### DIFF
--- a/desk/app/logs.hoon
+++ b/desk/app/logs.hoon
@@ -24,7 +24,9 @@
       def  ~(. (default-agent this %|) bowl)
       cor  ~(. +> [bowl ~])
   ::
-  ++  on-init  on-init:def
+  ++  on-init
+    ^-  (quip card _this)
+    `this(posthog `%info)
   ++  on-save  !>(state)
   ++  on-load
     |=  =vase
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -3684,9 +3684,10 @@
       =/  =channel:g  (got:by-ch nest)
       ?.  (~(has by sections.group) section.u-channel)
         (go-restart-updates `'missing channel updated section')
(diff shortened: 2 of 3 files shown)


arthyn (approved):
lgtm

desk/app/logs.hoon
```diff
@@ -24,7 +24,9 @@
       def  ~(. (default-agent this %|) bowl)
       cor  ~(. +> [bowl ~])
   ::
-  ++  on-init  on-init:def
+  ++  on-init
+    ^-  (quip card _this)
+    `this(posthog `%info)
```
Fang-, line 29: Oops, somehow I missed this. Thanks!

Fang- (approved):
All good changes!

desk/app/logs.hoon
mikolajpp: I would not find out, if I did not run the unit tests - they all now work. Just sayin' ;-)

PR #5138, opened 2025-09-18
groups: implement invite revocation

mikolajpp:
## Summary

After an invite has been issued, we don't track its whereabouts, and would not revoke it in any circumstances. However, when a group is deleted, the token expires or is revoked, the associated invite should be revoked for better UX.

## Changes

1. Anchor previous types at v7 in `/sur/groups-ver.hoon
2. Introduce a new v8 $invite that has an additional `valid` flag.
3. Introduce a new v8 $a-foreigns poke that has a new variant `%revoke` which can be used to revoke a token.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:


Fang- (commented):
Comments from our pair review below. glhf!

Doesn't implement handling of the `%revoke` poke lol.

Mostly nits, some real concerns around timers, and a weird scry path, etc.

desk/app/groups.hoon
```diff
@@ -1240,6 +1240,12 @@
   ::
       [%server %groups ship=@ name=@ rest=*]
     =+  ship=(slav %p ship.pole)
+    ::  ignore responses after group has been deleted 
+    ::
+    ?:  ?&  !(~(has by groups) ship name.pole)
+            ?=([%invite %revoke ship=@ ~] rest.pole)
+        ==
+      cor
```
Fang-: And we ignore only these because getting signs on other wires is unexpected when the group is no longer in state, right?

Fang- (approved):
Changes all look good. One question to confirm, but should be good to go afaic. Thank you!

PR #5139, opened 2025-09-18
logs: fix type issue

arthyn:
## Summary

In some cases this doesn't come in as a list? this fixes it

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

### DIFF
--- a/desk/app/logs.hoon
+++ b/desk/app/logs.hoon
@@ -13,7 +13,7 @@
         posthog=(unit volume:l)
     ==
   ::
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Wild. Still unclear why the import isn't always a proper `$wain`. But also, we don't use `/*` _that_ much, could be some sort of ford bug or behavioral oddity there.

PR #5146, opened 2025-09-19
tests: easier testing of negotiate-wrapped agents

Fang-:
## Summary

Introduces /lib/test-negotiate-agent, which wraps around /lib/test-agent to augment it with transparent handling and hiding of /lib/negotiate behavior.

Simplifies the groups agent tests by making them use this as a drop-in replacement of /lib/test-agent.

## Changes

/lib/test-negotiate-agent overloads the original test-agent library, changing the behavior of `+do` to process negotiate cards appropriately, and re-defining every arm that calls `+do` to make sure that gets used.

When we see a version negotiation subscription we out, we simulate an instant watch-ack and matching version fact. This will "release" the underlying subscriptions, if any, and from the test writer's perspective they happen as if the negotiate wrapper wasn't even there.

## How did I test?

I, uh, ran the tests.

## Risks and impact

I'm not very happy with the amount of code duplication. There's probably some way to make /lib/test-agent "pluggable" to reduce this, but I fear that would increase its implementation complexity non-trivially.

This also continues /lib/negotiate's pattern of "just pretend it's not there", which is fine most of the time but has turned out a bit annoying at times wrt inspectability, or odd surprised where it isn't quite 100% invisible. One known case of that here, is that if you inspect the bowl from the test execution state directly, you'll find all the negotiate subs and wires in there.

b593ce7 doesn't _seem_ to affect any of our existing tests, they continue passing, but it _is_ a subtle behavioral change. You have been warned!

## Rollback plan

Revert.

### DIFF
--- a/desk/lib/test-negotiate-agent.hoon
+++ b/desk/lib/test-negotiate-agent.hoon
@@ -1,0 +1,159 @@
+::  test-agent-negotiate: like /lib/test-agent, but negotiation transparency
+::
+::    use this as you would /lib/test-agent. this library will hide most
+::    /lib/negotiate behavior and side-effects, letting you write tests against
+::    your agent's logic without having to simulate version negotation.
+::
+::    caveats:
+::    - the bowl will still contain /~/negotiate wires
+::
+/+  *test-agent, negotiate
+::
+|%
++$  card  card:agent:gall
+::TODO  +get-bowl
+++  do
+  |=  call=$-(state [(list card) agent:gall])
+  =/  m  (mare ,(list card))
+  ::NOTE  ^do will play the cards with their negotiate wires
+  ;<  caz=(list card)  bind:m  (^do call)
+  ::  simulate immediate matching versions for everywhere we negotiate
+  ::
+  ;<  res=cage  bind:m  (got-peek /x/~/negotiate/config)
+  =+  !<(=config:negotiate q.res)
+  =|  cards=(list card)
+  |-  =*  loop  $
+  ?^  caz
+    ?.  ?=([%pass [%~.~ %negotiate %heed @ @ @ ~] *] i.caz)
+      loop(caz t.caz, cards (snoc cards i.caz))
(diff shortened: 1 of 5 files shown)


desk/lib/test-negotiate-agent.hoon
```diff
@@ -0,0 +1,159 @@
...
+  |-  =*  loop  $
+  ?^  caz
+    ?.  ?=([%pass [%~.~ %negotiate %heed @ @ @ ~] *] i.caz)
+      loop(caz t.caz, cards (snoc cards i.caz))
+    =*  who  (slav %p i.t.t.t.p.i.caz)
+    =*  dud         i.t.t.t.t.p.i.caz
+    =*  pro       i.t.t.t.t.t.p.i.caz
+    ::NOTE  at time of writing, /lib/negotiate no-ops on watch acks,
+    ::      so we can get away with not processing the cards here.
+    ;<  *  bind:m  (^do-agent p.i.caz [who dud] %watch-ack ~)
```
mikolajpp, line 34: As in, it could be a watch-nack? I am not sure I understand this comment.

desk/lib/test-negotiate-agent.hoon
```diff
@@ -0,0 +1,159 @@
...
+  ::  filter out /lib/negotiate internal effects,
+  ::  and clean up wrapped wires
+  ::
+  %-  pure:m
+  %+  murn  cards
+  |=  =card
+  ^-  (unit ^card)
+  ?+  card  `card
+      [%pass [%~.~ %negotiate %inner-watch @ @ *] *]
+    `card(p t.t.t.t.t.p.card)
```
mikolajpp, line 50: I mean, this works, but what about saying "skip 5 elements". Hopefully this typechecks :-)
```suggestion
    `card(p |5.p.card)
```

mikolajpp (approved):
One nit and one question. 

Also a general question: why do we duplicate the family of `+do-*` arms? Is it because a few arms are indeed different, so for completeness we include everything?

PR #5147, opened 2025-09-22
query benchmarks + optimization

dnbrwstr:
## Summary

- Add utilities for benchmarking queries
- Use indexes to optimize some queries.

## Changes

- Adds indexes on foreign keys to schema to improve relation join performance
- Adds `QueryBenchScreen` that allows running of benchmarks for most queries
- Caveats:
    - To avoid bloating the bundle with test data, `QueryBenchScreen` has to be inserted manually into the render tree (I suggest either `AuthenticatedApp` for mobile or `AppRoutes` for web.
    - Write benchmarks use fixture data from the `shared` package, read benchmarks read from local data, whatever exists, and so there will be variation depending on what's on your device.

## Risks and impact

## Screenshots / videos

simulator_screenshot_E0E8C553-A402-4D6B-9D3D-B69EE9EFE679

### DIFF
--- a/apps/tlon-web/playwright.config.ts
+++ b/apps/tlon-web/playwright.config.ts
@@ -9,8 +9,8 @@ const isProductionMode = process.env.USE_PRODUCTION_BUILD === 'true';
 const webServers = Object.entries(shipManifest).map(
   ([key, ship]: [string, any]) => ({
     command: isProductionMode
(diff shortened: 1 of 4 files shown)


patosullivan (approved):
Lgtm! Makes sense. QueryBenchScreen will be useful.

PR #5155, opened 2025-09-24
images: make sure files uploaded to memex or downloaded to device have a file extension

patosullivan:
fixes tlon-4593

## Summary

Users were seeing an alert saying "Failed to save image to photos. Please check your device storage and try again" when they attempted to download some images from the ImageViewerScreenView via the download button.

I was able to repro this by attempting to download any image with a filename that *didn't* include a filetype extension. 

That specific error language about checking device storage is just what we show anytime we see any errors with saving. The console showed this:

```
Save error: Error: This file type is not supported yet
    at construct (native)
    at apply (native)
    at _construct (http://192.168.0.46:8081/apps/tlon-mobile/index.bundle//&platform=ios&dev=true&hot=false&transform.engine=hermes&transform.bytecode=1&transform.routerRoot=app&unstable_transformProfile=hermes-stable:3623:67)
    at Wrapper (http://192.168.0.46:8081/apps/tlon-mobile/index.bundle//&platform=ios&dev=true&hot=false&transform.engine=hermes&transform.bytecode=1&transform.routerRoot=app&unstable_transformProfile=hermes-stable:3595:25)
    at construct (native)
    at _callSuper (http://192.168.0.46:8081/apps/tlon-mobile/index.bundle//&platform=ios&dev=true&hot=false&transform.engine=hermes&transform.bytecode=1&transform.routerRoot=app&unstable_transformProfile=hermes-stable:110001:170)
    at CodedError (http://192.168.0.46:8081/apps/tlon-mobile/index.bundle//&platform=ios&dev=true&hot=false&transform.engine=hermes&transform.bytecode=1&transform.routerRoot=app&unstable_transformProfile=hermes-stable:110012:25)
```

I added a function to take on a filetype extension to the URLs, and the images save just fine.

So, we actually had two problems:

1) Some images are uploaded to memex without a filtype extension.
2) Expo's MediaLibrary won't allow a user to attempt to download a file without an extension.

And a third problem if you count the fact that you *could* download these files on web, but without an extension your OS wouldn't know how to open them.
  

## Changes

- add `ensureFileExtension` to `ImageViewerScreenView` to catch any instance of a previously uploaded image without an extension (there are surely many of these floating around).
- add `getExtensionFromMimeType` to `storageActions` so we can add an appropriate filetype extension to any uploaded file before upload.

Both default to `.jpg` if we can't infer the appropriate extension from the MIME type.

## How did I test?

Tested on iOS, Android, and Web. Made sure image uploads/downloads still work where they should.

If you'd like to test yourself, a known-bad image is the latest image uploaded by Galen to the Tlon Studio groups's 2025 Studio gallery (it's a picture of a red flower). Try it on develop vs this branch.

Try uploading images via memex and make sure they have a file extension.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: image uploads.

## Rollback plan

Revert

## Screenshots / videos

N/A

### DIFF
--- a/packages/shared/src/store/storage/storageActions.ts
+++ b/packages/shared/src/store/storage/storageActions.ts
@@ -12,6 +12,7 @@ import { createDevLogger, escapeLog } from '../../debug';
 import { setUploadState } from './storageUploadState';
 import {
   fetchFileFromUri,
+  getExtensionFromMimeType,
   getMemexUpload,
   hasCustomS3Creds,
   hasHostingUploadCreds,
@@ -20,23 +21,6 @@ import {
 const logger = createDevLogger('storageActions', false);
 
 export const PLACEHOLDER_ASSET_URI = 'placeholder-asset-id';
-
-function getExtensionFromMimeType(mimeType?: string): string {
-  if (!mimeType) {
-    return '.jpg';
-  }
-
-  const mimeToExt: Record<string, string> = {
-    'image/jpeg': '.jpg',
-    'image/jpg': '.jpg',
-    'image/png': '.png',
-    'image/gif': '.gif',
-    'image/webp': '.webp',
-    'image/heic': '.heic',
-    'image/heif': '.heif',
-  };
-  return mimeToExt[mimeType.toLowerCase()] || '.jpg';
-}
 
 function getSaveFormat(mimeType?: string): SaveFormat {
   if (!mimeType) {
--- a/packages/app/ui/components/ImageViewerScreenView.tsx
+++ b/packages/app/ui/components/ImageViewerScreenView.tsx
@@ -1,5 +1,9 @@
 import { ImageZoom, Zoomable } from '@likashefqet/react-native-image-zoom';
-import { createDevLogger } from '@tloncorp/shared';
+import {
+  createDevLogger,
+  ensureFileExtension,
+  downloadImageForWeb
+} from '@tloncorp/shared';
 import { Icon } from '@tloncorp/ui';
 import { Image } from '@tloncorp/ui';
 import * as FileSystem from 'expo-file-system';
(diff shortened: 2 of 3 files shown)


packages/shared/src/store/storage/storageActions.ts
```diff
@@ -38,6 +55,7 @@ export const uploadAsset = async (asset: ImagePickerAsset, isWeb = false) => {
   try {
     logger.log('resizing asset', asset.uri);
     let resizedAsset = asset;
+    const originalMimeType = asset.mimeType;
     // avoid resizing gifs
     if (!asset.mimeType?.includes('gif')) {
       resizedAsset = await manipulateAsync(
```
arthyn, line 61: doesn't this automatically convert to JPEG? if I read this correctly I think that we'd convert something to JPEG and then later on try to tack on the original mime which would give it an incorrect extension.

jamesacklin:
Is there any risk/danger in allowing GIFs?

packages/shared/src/store/storage/storageActions.ts
```diff
@@ -21,6 +21,39 @@ const logger = createDevLogger('storageActions', false);
 
 export const PLACEHOLDER_ASSET_URI = 'placeholder-asset-id';
 
+function getExtensionFromMimeType(mimeType?: string): string {
```
dnbrwstr, line 24: Should also probably be in `storageUtils`

packages/shared/src/store/storage/storageActions.ts
```diff
@@ -21,6 +21,39 @@ const logger = createDevLogger('storageActions', false);
...
+  if (!mimeType) {
+    return '.jpg';
+  }
+
+  const mimeToExt: Record<string, string> = {
```
dnbrwstr, line 29: Should probably only be defining this once.

PR #5157, opened 2025-09-26
chat: fix DM sync across clients

patosullivan:
## Summary

fixes tlon-4834

Fixes DM sync issue where DM invites remained stuck on some clients after being accepted or declined on other clients.

Before this change, the frontend only updated DM invites during init or re-init (triggered by connection resets, newly joined channels, or stale data). There was no real-time sync for DM invite state changes. This PR adds real-time sync via the chat agent's `/v3` wire, ensuring all clients immediately reflect when DM invites are accepted, declined, or when regular DMs are left.

We *did* expect an `addDmInvites` event on `/v3` in the frontend previously, but no event was ever sent on that path, and no events were ever emitted when we declined an invite or left a DM, so frontend state could be stale on clients. On the clients with stale state, the user also couldn't leave an existing invite because we were using `di-abed` rather than `di-abed-soft` in `%chat-dm-rsvp`.

## Changes

**Backend:**
- Updated `di-rsvp` to emit updated DM invite lists when accepting or declining DMs
- Modified `give-invites` to correctly handle DM state transitions
- Changed from `di-abed` to `di-abed-soft` in `%chat-dm-rsvp` to gracefully handle missing DMs (i.e., if the client thinks we still have a DM and the user tries to leave it, we allow it)
- Updated discipline so that we accept the `%ships` mark on the versioned paths.

**Frontend API:**
- Changed event type from `addDmInvites` to `syncDmInvites` to reflect the new full sync
- Updated event handler to properly parse DM invite sync events from backend

**Frontend Sync:**
- Implemented `handleSyncDmInvites()` function with DM invite state reconciliation
- Added logic to distinguish between accepted invites (convert to regular DMs) and declined invites (remove from local state)
- Cross-references backend DM state to ensure local state accuracy

## How did I test?

- Tested DM invite synchronization across multiple clients
- Verified that accepting DM invites on one client properly syncs to other clients
- Confirmed declining DM invites removes them from all connected clients
- Tested leaving regular DMs to ensure proper cleanup across clients

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.

## Screenshots / videos

N/A

### DIFF
--- a/desk/app/chat.hoon
+++ b/desk/app/chat.hoon
@@ -115,17 +115,17 @@
           [/epic %epic ~]
           [/unreads %chat-unread-update ~]
         ::
-          [/v1 %chat-club-action-1 %writ-response-1 ~]
+          [/v1 %chat-club-action-1 %writ-response-1 %ships ~]
           [/v1/club/$ %writ-response-1 ~]
           [/v1/clubs %chat-club-action-1 ~]
           [/v1/dm/$ %writ-response-1 ~]
         ::
-          [/v2 %chat-club-action-1 %writ-response-2 ~]
+          [/v2 %chat-club-action-1 %writ-response-2 %ships ~]
           [/v2/club/$ %writ-response-2 ~]
           [/v2/clubs %chat-club-action-1 ~]
           [/v2/dm/$ %writ-response-2 ~]
         ::
-          [/v3 %chat-club-action-1 %writ-response-3 ~]
+          [/v3 %chat-club-action-1 %writ-response-3 %ships ~]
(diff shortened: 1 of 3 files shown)


patosullivan:
The tests that were failing pass locally, and now the full suite passes in CI. Trying to figure out what is causing the flakiness.

dnbrwstr (approved):
Cool, frontend code looks fine to me then. As long as it's happening infrequently should be ok.

PR #5169, opened 2025-10-02
onboarding: reduce signup time

latter-bolden:
## Summary

Once initial account setup is complete (OTP verified), we're taking 30-40 seconds to complete our boot sequence with the node. This PR makes some changes to get that number down (details below). In my testing, this cuts it to ~15 seconds.

Some of these changes remove checks that were in place to help guarantee our final signup state meets requirements. Loosening these may hurt reliability and is something we should keep an eye on. That said, I never saw a failure while testing locally. 
 
## Changes
- once authenticated, reduce delay before attempting to connect
- while connecting, quickly force an initial subscription down the channel (cuts connection time in half)
  - it's unclear to me why this has an effect? My understanding was the node should immediately send the initial SSE response these days. But it does seem to reliably provide a meaningful reduction...
- remove personal group consistency checks
   - the response for the group create thread is incomplete which was causing us to loop on that step a few times before detecting that the group had all its channels. The consistency check was largely vestigial from a time were the client was responsible for manually creating all of the channels separately
- don't wait for invite acceptance
  - we still initiate all the requisite group joins and dm accepts, but we avoid blocking until we can verify they processed correctly via re-sync
- reduce timeout for optional boot sequence steps
- consolidate Hosting api calls into one (inventory list/reservation, access code retrieval, check boot status)
- add boot step duration breakdown to the final signup telemetry report

## How did I test?
Tested via repeatedly running our maestro signup test locally.

## Risks and impact

  - [x] Onboarding

Fixes TLON-4911


jamesacklin:
> That's one of the things this PR does.

Right, I meant to ask: with this PR, if we have a slow join, does the user see something broken on the Home Screen?

apps/tlon-mobile/src/lib/signupContext.tsx
```diff
@@ -90,13 +90,11 @@ export const SignupProvider = ({ children }: { children: React.ReactNode }) => {
       };
       runPostSignupActions(postSignupParams);
       logger.trackEvent('hosted signup report', {
-        bootDuration: bootReport
-          ? bootReport.completedAt - bootReport.startedAt
-          : null,
+        ...bootReport,
         userSatWaitingFor: values.userWasReadyAt
-          ? Date.now() - values.userWasReadyAt
+          ? Number(((Date.now() - values.userWasReadyAt) / 1000).toFixed(1))
```
dnbrwstr, line 95: Super-nit, but the number cast shouldn't be necessary afaict. Also, `toFixed` is going to convert this to a string, which could be annoying for posthog queries, but maybe posthog handles conversion automatically?

packages/app/hooks/useBootSequence.ts
```diff
@@ -105,14 +114,16 @@ export function useBootSequence() {
...
+          reservedNode.code
+        );
         if (!shipInfo) {
           throw new Error('Could not authenticate with node');
         }
         setShip({ ...shipInfo, needsSplashSequence: true });
         telemetry?.identify(preSig(shipInfo.ship!), { isHostedUser: true });
 
-        await wait(2000);
+        await wait(200);
```
dnbrwstr, line 126: Would be nice to have a comment here as to why this wait is necessary. Any way it could be more deterministic?

dnbrwstr (approved):
Lgtm! Just nits.

packages/app/hooks/useBootSequence.ts
latter-bolden: It's race prone because deeper logic needs to read that `setShip` result, untangling is probably doable but out of scope. Will add a comment 👍🏻

packages/app/hooks/useBootSequence.ts
```diff
@@ -135,7 +146,8 @@ export function useBootSequence() {
     // CONNECTING: make sure the connection is established
     //
     if (bootPhase === NodeBootPhase.CONNECTING) {
-      await wait(1000);
+      // forcing a subscribe once down the channel reduces connection time significantly
```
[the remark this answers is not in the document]
dnbrwstr: I get that, but the phrasing is still unclear to me. `reduce subsequent channel connection time by force initializing subscription here`?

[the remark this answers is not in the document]
latter-bolden: `immediately subscribing on a path that generates a fact significantly reduces connection time`?

PR #5170, opened 2025-10-02
sync: message preload client integration

latter-bolden:
## Summary

Implements the client portion of our new `init-posts` endpoint for fetching initial message data after mobile login or web tab open. Previously on mobile we were not preloading messages and on web we were using a multi request "predictive sync" operation. This gives us a configurable single endpoint to fetch starting messages.

Uses adaptive loading on mobile where we first check the network characteristics before deciding how much data to sync. If you're on WiFi, it'll behave differently than if you're on 3G.

It's easy to switch the constants for num channels and num messages. We probably want to do a bit more perf testing to select the right constants there, but right now the base64 bug is complicating getting representative test data. I don't think that needs to block merge, but it should block deployment.

Fixes TLON-4901

## Changes
- add urbit types for `init-posts` result
- add `api` and `sync` methods for fetching the endpoint
- call it once on login for mobile
- call it on app open for web

## How did I test?

On simulator, I logged in and confirmed it synced messages successfully.

## Risks and impact

  - [x] Onboarding
    - sort of, affects initial UX when you're dumped into ChatList
  - [x] Message sync

### DIFF
--- a/apps/tlon-mobile/src/components/AuthenticatedApp.tsx
+++ b/apps/tlon-mobile/src/components/AuthenticatedApp.tsx
@@ -79,6 +79,8 @@ export default function ConnectedAuthenticatedApp() {
   useEffect(() => {
     async function setup() {
       configureClient();
+      // we store a flag to ensure this runs only once per login, not anytime
+      // the app is opened
       const didSyncInitialPosts = await db.didSyncInitialPosts.getValue();
       sync.syncStart().then(async () => {
         if (!didSyncInitialPosts) {
--- a/apps/tlon-mobile/src/components/AuthenticatedApp.tsx
+++ b/apps/tlon-mobile/src/components/AuthenticatedApp.tsx
@@ -1,3 +1,4 @@
+import NetInfo from '@react-native-community/netinfo';
 import {
   AppStatus,
   useAppStatusChange,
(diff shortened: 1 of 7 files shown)


apps/tlon-mobile/src/components/AuthenticatedApp.tsx
```diff
@@ -76,9 +77,25 @@ export default function ConnectedAuthenticatedApp() {
   const configureClient = useConfigureUrbitClient();
 
   useEffect(() => {
-    configureClient();
-    sync.syncStart();
-    setClientReady(true);
+    async function setup() {
+      configureClient();
+      const didSyncInitialPosts = await db.didSyncInitialPosts.getValue();
```
dnbrwstr, line 82: Why do we need this flag? Afaict, this effect should only ever run once per `configureClient` change, which should only change when user auth details change (in which case I'd image we do want to potentially rerun).

packages/shared/src/store/sync.ts
```diff
@@ -1496,6 +1456,39 @@ export async function syncSequencedPosts(
...
+    }
+
+    await db.didSyncInitialPosts.setValue(true);
+
+    logger.trackEvent('Synced Initial Posts', {
+      syncSize: config.syncSize,
+      numPostsSynced: posts.length,
+    });
+  } catch (err) {
+    console.log(`bl: it threw?`, err);
```
dnbrwstr, line 1485: Stray console log

dnbrwstr (approved):
Lgtm! Couple nits.

apps/tlon-mobile/src/components/AuthenticatedApp.tsx
latter-bolden: My thinking was we only want it to run once per login. Without checking a stored value, won't it run anytime the app is freshly opened? e.g. after force quit

PR #5171, opened 2025-10-02
contacts: new /changes that serves $pages

Fang-:
## Changes

Instead of looking at the change timestamp provided by the profile's publisher, track update timestamps locally and operate off those. We track them for edits of our own profile, edits to overlays on remote profiles, and foreign profile updates.

We serve an object containing pages (which serialize to an `[foreign-contact overlay-contact]` array) or `null` to indicate deletions. cc @latter-bolden for API design review.

Depends on the `/lib/kol` added in #5166.

## How did I test?

Hit the endpoint manually before and after making profile changes.

## Risks and impact

The state migration logic in /app/contacts was kinda awkwardly structured. This should be _slightly_ better, but still not ideal. Cursed core structuring etc.

- Yes, safe to rollback without consulting PR author, but you'd need to either undo the state migration or leave that as-is.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert accounting for state, see above.


mikolajpp (approved):
Looks good, approved!

desk/app/groups-ui.hoon
```diff
@@ -259,6 +259,21 @@
         'groups'^groups
         'contacts'^contacts
     ==
+  ::
+      [%x %v6 %changes since=@ ~]
```
mikolajpp, line 263: Converting to json already in groups-ui is the new way? A difference with the other approach of returning typed values and doing the final conversion in groups-ui is that this will not catch a type mistake in a scry endpoint, so we should make sure each of these has been disciplined in the respective agents.

If we decide to do that, we can consider making groups-ui consistent and converting the other endpoints to this style.

Fang-: Yeah this feels much nicer in that you don't have to construct a separate type for what is really intended to be just a collation of json scry requests.

_Every_ scry endpoint _should_ be disciplined, yes!

PR #5174, opened 2025-10-03
channels/server: drop bad links from .meta too

Fang-:
## Summary

In #5119 we had cleaned up "bad" links (usually: base64 blobs) from post and log contents. However, we neglected to check the meta.essay, which may have similar troubles in its image and cover fields.

## Changes

We lightly restructure the existing utilities, to be a core with multiple entry-points based on what type you want to drop bad links from.

We update the callsites as appropriate, and _move_ the cleanup logic into a new state transition to avoid duplicating effort for not-yet-up-to-date ships.

Lastly we update the tests to match.

## How did I test?

It compiles!

## Risks and impact

The core link cleaning/detection logic hasn't changed, so should be derisked by virtue of #5119 being out and fine.

- Yes, safe to rollback without consulting PR author, but runs up against state migration stuff.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert logic w/o reverting state transition.

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -60,7 +60,7 @@
   |%
   +$  card  card:agent:gall
   +$  current-state
-    $:  %12
+    $:  %13
         =v-channels:v9:c
         =hooks:h
         =pimp:imp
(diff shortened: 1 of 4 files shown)


jamesacklin:
Might want to target `staging` with this.

PR #5176, opened 2025-10-06
channels: ditch semi-redundant faces in $v-channel

Fang-:
## Summary

Proposal for dropping semi-redundant faces on revisioned parts of `$v-channel`.

## Changes

`$v-channel`'s `$global` type `+rev`isions most of its child nouns. As written, they get two faces: one on the outer [revision value] pair, and one on the value itself.

The way that's written makes it easier to skip past the revision and get just the value out, at the cost of making usage of the stuff _inside_ the value more verbose.

This commit makes the opposite tradeoff and removes the additional face. This way, skipping past the revision to the raw value requires raw tree addressing, but working with the stuff inside the value takes one less face.

Presented as RFC. Made a note of this a while ago and wanted to try it the other way, I'd be fine with either merging or closing. Argument in favor of this is that usage happens spread out all over the place, but stripping the revision off happens in more or less the same place for all these types, limiting the cognitive drawback of looking at the lark.

## How did I test?

Compiles, should be equivalent.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.

### DIFF
--- a/desk/app/channels-server.hoon
+++ b/desk/app/channels-server.hoon
@@ -745,7 +745,7 @@
   =/  affected=(list nest:c)
     %+  murn  ~(tap by v-channels)
     |=  [=nest:c channel=v-channel:c]
-    ?.  =(flag.r-groups group.perm.perm.channel)  ~
+    ?.  =(flag.r-groups group.perm.channel)  ~
     `nest
   =*  r-group  r-group.r-groups
   ?+    r-group  cor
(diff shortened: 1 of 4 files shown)


mikolajpp (commented):
I think this could go in, can't think of any drawbacks except that the `+` pattern is opaque when seen for the first time.

A while ago I was also thinking about changing the `+rev` mold to include a `val` face on the inner value, so that `perm.perm.channel` would become `val.perm.channel`, which looks less confusing that the doubled face.

Compared with this, however, it seems the benefit is largely cosmetic, given that using `+` can be considered an idiom as good as an explicit `.val` face. 

I would be in favor of merging, but let's sit on it one more day :-)

arthyn (approved):
lgtm

PR #5177, opened 2025-10-06
metagrab: include accept-language header

Fang-:
## Summary
If the request coming in to metagrab includes an `accept-language` header, pass that on to whatever resource we're trying to grab metadata from. This way, we are hopefully more likely to get results in the user's language.

## Changes

We restructure lightly to make request header tracking more generic and require smaller diffs if we want to include another header in the future. Instead of putting the `user-agent` header contents in the wire, we put a `%blob` `$coin` containing the whole `$header-list:http`. We unpack this carefully, making sure it fits the expected type, and falling back to not retrieving headers instead of crashing the processing entirely. (This is a stylistic/defensive choice. We will get out what we put in 100% of the time.)

## How did I test?

By hitting a local echo server and observing that the headers put in by my browser were passed on correctly.

## Risks and impact

- Technically safe to rollback without consulting PR author, worst you'll see is failing to process some in-flight-during-ota metagrab responses.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.

## Screenshots / videos
image

Fixes TLON-4878

### DIFF
--- a/desk/app/metagrab.hoon
+++ b/desk/app/metagrab.hoon
@@ -105,17 +105,14 @@
   ==
 ::
 ++  fetch
-  |=  [met=?(%head %get) url=@t uas=@t]
+  |=  [met=?(%head %get) url=@t hes=header-list:http]
   ^-  card
-  =/  =header-list:http
-    :~  ['user-agent' uas]
-        ['accept' '*/*']
-    ==
+  =.  hes  [['accept' '*/*'] hes]
   =/  =request:http
-    [?-(met %head %'HEAD', %get %'GET') url header-list ~]
+    [?-(met %head %'HEAD', %get %'GET') url hes ~]
   ::NOTE  outbound-config is actually meaningless,
   ::      iris doesn't do anything with it at present...
-  [%pass /fetch/(scot %t url)/(scot %t uas)/[met] %arvo %i %request request *outbound-config:iris]
+  [%pass /fetch/(scot %t url)/(crip ~(rend co %blob hes))/[met] %arvo %i %request request *outbound-config:iris]
 ::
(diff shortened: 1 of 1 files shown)


desk/app/metagrab.hoon
```diff
@@ -105,17 +105,14 @@
...
-        ['accept' '*/*']
-    ==
+  =.  hes  [['accept' '*/*'] hes]
   =/  =request:http
-    [?-(met %head %'HEAD', %get %'GET') url header-list ~]
+    [?-(met %head %'HEAD', %get %'GET') url hes ~]
   ::NOTE  outbound-config is actually meaningless,
   ::      iris doesn't do anything with it at present...
-  [%pass /fetch/(scot %t url)/(scot %t uas)/[met] %arvo %i %request request *outbound-config:iris]
+  [%pass /fetch/(scot %t url)/(crip ~(rend co %blob hes))/[met] %arvo %i %request request *outbound-config:iris]
```
mikolajpp, line 115: Yes, this works, and is quite neat in that it seamlessly handles the old wire style.
I do wonder, is the only motivation here avoiding having to track this explicitly in the state, and are there other benefits or drawbacks as well?

We might never, but if we wanted to track something else pertaining to the request, it seems like we would need to forever keep a bunch of `+soft` calls at the reception and see which type fits. (Or introduce a versioned wire, ha!)

mikolajpp (approved):
I have left a general question about _jam-typing_ the headers like we do here, but otherwise it looks good to go!

PR #5178, opened 2025-10-07
Enable backend unit tests in CI

mikolajpp:
## Summary

This PR implements a backend unit tests job. The test runner resorts to defensive timeouts set in click to work around deficiencies of click/conn interface, which presently has no way of reliably waiting for thread responses to be returned. We must ensure the timeout is long enough to allow the unit test thread to complete. 

Closes TLON-4862

## Changes

1. Implement backend tests job.

## How did I test?

Let it run.

## Risks and impact

- Safe to rollback without consulting PR author? (Yes | No) Yes

## Rollback plan

Simply revert.

### DIFF
--- a/desk/tests/app/channels-server.hoon
+++ b/desk/tests/app/channels-server.hoon
@@ -6,7 +6,7 @@
 ++  dap  %channels-server
 ::
 +$  current-state
-  [%11 =v-channels:v9:c =hooks:h =pimp:imp]
+  [%13 =v-channels:v9:c =hooks:h =pimp:imp]
 +$  state-8
   [%8 =v-channels:v8:c =hooks:h =pimp:imp]
 --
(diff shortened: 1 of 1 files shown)


backend/click
```diff
@@ -0,0 +1,285 @@
...
+#   Thin client for interacting with running Urbit ship via conn.c.
+#
+#   Compiles a Hoon command given as text against an optional list of
+#   dependencies and then executes it on a running ship.
+#
+#   The click-format helper script handles command formatting, whereas this
+#   script handles operational control flow. See the documentation in
+#   click-format for the details of command structure, dependency management,
+#   etc.
+#
```
Fang-, line 14: Should probably leave a comment describing the changes this has compared to the "original" click, and why. ie timeout option and `socat`.

backend/strandio.patch
```diff
@@ -0,0 +1,2 @@
+958d957
+<   ;<  ~      bind:m  (sleep ~s0)  ::  wait for thread to start
```
Fang-, line 2: As discussed, not clear why this sleep is here in the first place. Could be worth PRing to urbit/urbit, but also out of scope for here.

Fang- (approved):
Well CI passes, so surely this must be good. (;

Looks perfectly reasonable to me. As we discussed, maybe there's something to be done about the need for the timeouts here, but we should be very happy to take this as-is too!

patosullivan (approved):

PR #5179, opened 2025-10-07
various: remove hark interactions

Fang-:
## Summary

We stopped being dependent on landscape desk's %hark agent some time ago, but code calling out to it and using its types remains.

Here, we continue the un-harking started in #3995, and rip out the notify agent's hark-style backcompat from #3609.

## Changes

Notify agent now serves a 410 ("gone") when asked about hark-style notification details, and we rip out all the hark poking logic from channels and chat agents. Luckily, this was all fairly self-contained.

Note that we leave heap and diary agents untouched. Those really should be replaced with stubs instead, so we don't waste any time on them here.

## How did I test?

Smoke tested with two fakeships interacting. Unreads and notifications show up as expected.

## Risks and impact

This will break notification information fetching for very old clients. cc @jamesacklin 

- Should be safe to rollback without consulting PR author.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [x] Notifications
  - [ ] Other:

## Rollback plan

Should revert cleanly.


mikolajpp (approved):
Hark! A welcome change.

PR #5180, opened 2025-10-07
refresh roles ui

dnbrwstr:
## Summary

Updates UI for configuring channel roles.

## Changes

- Adds ability to set initial roles / privacy when creating a channel
- Updates edit channel roles ui to a table-based interface
- Adds link to channel settings in channel overflow menu

## How did I test?

Tested in sim + on desktop.

## Screenshots / videos

| channel creation | channel editing |
|-----------------|-----------------|
| <video src="https://github.com/user-attachments/assets/32f69c5d-1ef5-4742-84c3-5db935999df3" style="max-width: 350px;"/>|<video src="https://github.com/user-attachments/assets/1a3361ec-ce26-4873-8089-15dd0345acbd" style="max-width: 350px;">|

Fixes TLON-4696

### DIFF
--- a/.maestro/42-group-info-settings.yaml
+++ b/.maestro/42-group-info-settings.yaml
@@ -154,7 +154,7 @@ appId: io.tlon.groups
     id: 'EditChannelButton'
     index: 1
 - assertVisible:
-    text: 'Edit channel'
+    text: 'Channel settings'
 - longPressOn: 'Second chat channel'
 - runFlow:
     when:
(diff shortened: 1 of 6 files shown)


packages/app/features/groups/ManageChannelsScreen.tsx
```diff
@@ -41,10 +41,9 @@ export function ManageChannelsScreen(props: Props) {
       navigation.navigate('EditChannel', {
         groupId,
         channelId,
-        fromChatDetails,
```
dnbrwstr, line 44: This behavior was incorrect previously, since it passes on the value from the current route. Even if we arrived at `ManageChannelsScreen` from `ChatDetails`, we want to `EditChannel` screen to navigate back here on pressing `back`, not to `ChatDetails`.

PR #5181, opened 2025-10-07
groups: keep creation thread and its marks warm

Fang-:
## Changes

These weren't included anywhere, but they're part of the very first interactions that get made against a ship, and so need to be guaranteed snappy.

This makes sure both the threads and the marks that get used with it are in the build cache whenever groups agent gets built.

Hopefully closes TLON-4906.

## How did I test?

Compiles!

## Risks and impact

Super low risk, basically no-impact, except maybe responsiveness gains.

## Rollback plan

Revert.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -16,6 +16,9 @@
 /+  neg=negotiate, discipline
 ::  performance, keep warm
 /+  groups-json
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):

mikolajpp (approved):
Thanks, good to go!

PR #5182, opened 2025-10-07
settings: volume options

jamesacklin:
## Summary

Fixes TLON-4855 where it's possible to have granted permissions for push notifications, yet have groups, DMs, or channels muted such that it produces no effect.

Rather than slapping a loud alert when the user has selected "Nothing," I opted to redesign the screen slightly with better language.

I also reflected the "exception" groups/DMs/channels' %volume states into the ChatOptionsSheet to aid usability.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [x] Notifications
  - [ ] Other:

## Rollback plan

revert

## Screenshots / videos

Screenshot 2025-10-07 at 4 05 53 PM

### DIFF
--- a/packages/app/ui/components/ChatOptionsSheet.tsx
+++ b/packages/app/ui/components/ChatOptionsSheet.tsx
@@ -266,6 +266,7 @@ export function GroupOptionsSheetContent({
   const canInvite = currentUserIsAdmin || group.privacy === 'public';
   const isPinned = group?.pin;
   const isErrored = group?.joinStatus === 'errored';
+  const baseVolumeLevel = store.useBaseVolumeLevel();
 
   const wrappedAction = useCallback(
     (action: () => void, clearChat = true) => {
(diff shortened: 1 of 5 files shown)


packages/app/ui/components/ChatOptionsSheet.tsx
```diff
@@ -291,13 +292,26 @@ export function GroupOptionsSheetContent({
     store.leaveGroup(group.id);
   }, [group]);
 
+  const notificationTitle = useMemo(() => {
+    const hasCustomSetting = !!group.volumeSettings?.level;
```
dnbrwstr, line 296: Logic looks broadly similar for notificationTitle in both places, could it be pulled into a helper?

packages/shared/src/db/queries.ts
```diff
@@ -1920,12 +1920,11 @@ export const getChannelVolumeSetting = createReadQuery(
 export const getVolumeExceptions = createReadQuery(
   'getVolumeExceptions',
   async (ctx: QueryCtx) => {
-    const base = await ctx.db.query.volumeSettings.findFirst({
-      where: eq($volumeSettings.itemType, 'base'),
-    });
-
     const exceptions = await ctx.db.query.volumeSettings.findMany({
-      where: not(eq($volumeSettings.level, base?.level || 'default')),
+      where: or(
```
dnbrwstr, line 1924: What's the advantage of ignoring the base setting? I see a huge list of exceptions (that I don't remember making) when I look at my settings on this branch, but only two on develop.

PR #5468, opened 2026-01-31
Update to expo 54

janicduplessis:
## Summary

Upgrade to Expo 54, React Native 0.81, React 19, and Tamagui v2.0.0-rc.0 across the monorepo. This is a major dependency upgrade affecting both mobile and web platforms.

### Core Upgrades
- **Expo 52 to 54**: Updated Expo SDK, EAS config, and all Expo packages
- **React Native 0.73 to 0.81**: Updated React Native and related dependencies
- **React 18 to 19**: Updated React, React DOM, and adapted to new APIs
- **React Navigation v6 to v7**: Updated navigation libraries
- **Tamagui 1.x to 2.0.0-rc.0**: Updated Tamagui UI framework
- **Node.js 18 to 22**: Updated Node.js version requirement

### Mobile Platform Changes
- Migrated iOS AppDelegate from Objective-C to Swift
- Migrated Android MainActivity/MainApplication from Java to Kotlin
- Renamed index.js to index.tsx
- Moved Cosmos config to mobile app directory
- Updated build configuration and bundle generation for Expo 54
- Migrated from react-native-clipboard to expo-clipboard
- Updated PostHog integration for Expo 54

### Web Platform Fixes
- Fixed production build: added Vite preview proxy and expo polyfill
- Fixed envPrefix override by Tamagui vite-plugin (was hiding VITE_ env vars)
- Fixed Tamagui v2 dialog rendering issues:
  - Patched @tamagui/dialog to remove `render: 'dialog'` (native dialog element causes stacking context issues)
  - Added `disableRemoveScroll` to ActionSheet dialog mode
  - Fixed 0-height dialog content caused by `ScrollView flex={1}` changed to `flexShrink={1}`
  - Fixed text trimming margins causing click target overlap in GroupTypeCard

### Mobile Bug Fixes

- **Long-press message actions broken (TLON-5490)**: Removed `justifyContent="center" alignItems="center"` from the Modal ZStack that was interfering with ChatMessageActions positioning, blur overlay, and touch handling.

- **Chat input not shrinking after send (TLON-5492)**: Patched React Native's `BaseTextInputShadowNode.h` to fix uncontrolled multiline TextInput not resizing when children change. In Fabric, `measureContent()` is called by Yoga before `updateStateIfNeeded()` runs, so `attributedStringBoxToMeasure()` was returning stale native text for measurement. The patch detects when the React tree has diverged from state (for uncontrolled inputs only) and uses the React tree for measurement instead. Upstream PR: https://github.com/facebook/react-native/pull/56291

- **Swipe actions disappearing on chat list (TLON-5491)**: Removed `currentSwipeDirection` guards from `InteractableChatListItem`. In RNGH 2.28, `onSwipeableWillOpen` reports the gesture direction (`'left'` when swiping left to reveal right actions), not the side being opened. This caused `renderRightActions` to return an empty View when the right actions were being revealed. The guards are unnecessary since `ReanimatedSwipeable` already handles action visibility internally.

- **Keyboard padding accumulating on background/foreground (TLON-5493)**: When the app is backgrounded, iOS fires `keyboardWillChangeFrame` with `screenY=0` and `height=0`. The `ParentAgnosticKeyboardAvoidingView` interpreted this as "keyboard covers the entire screen" and applied massive paddingBottom, which accumulated on each background/foreground cycle. Fixed by ignoring these bogus keyboard events.

### Other
- Fixed React 19 and React Native compatibility issues across shared packages
- Added customizable hoverStyle prop to ChannelListItem and GroupListItem
- Updated rube Dockerfile Node.js setup to v22
- Fixed setTimeout type conflict in e2e test runner

## How did I test?

- Ran production smoke e2e test locally (passes)
- Ran create-group e2e test locally (dialog interactions work, groups create successfully)
- Verified VITE_ environment variables load correctly in dev and preview modes
- Verified dialog overlays no longer block pointer events on web
- Verified long-press message actions work (blur, positioning, dismissal)
- Verified chat input shrinks back to single line after sending multi-line messages
- Verified swipe actions (pin/mute) stay visible after swipe completes
- Verified keyboard padding stays stable after multiple background/foreground cycles

## Risks and impact

  - [x] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [x] Other: Build system, all UI components (Tamagui upgrade), mobile native code

## Rollback plan

Revert the entire branch. Due to the scope of the upgrade (Expo, React Native, React, Tamagui), partial rollbacks are not feasible.

## Screenshots / videos

N/A - infrastructure/dependency upgrade


apps/tlon-mobile/babel.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = function (api) {
   api.cache(true);
   return {
-    presets: [['babel-preset-expo', { jsxRuntime: 'automatic' }]],
+    presets: [['babel-preset-expo', { unstable_transformImportMeta: true }]],
```
janicduplessis: required for zustand to work

apps/tlon-mobile/metro.config.js
```diff
@@ -7,24 +7,17 @@ const { getSentryExpoConfig } = require('@sentry/react-native/metro');
...
+ *
+ * @type {import('@react-native/metro-config').MetroConfig}
+ */
+const config = {
   transformer: {
     babelTransformerPath: require.resolve('react-native-svg-transformer'),
-    getTransformOptions: async () => ({
-      transform: {
-        experimentalImportSupport: false,
-        inlineRequires: false,
```
janicduplessis: Was there any reason for disabling inline requires? Usually helps with performance.

package.json
```diff
@@ -56,20 +56,14 @@
...
-      "@react-navigation/drawer@6.7.2": "patches/@react-navigation__drawer@6.7.2.patch",
       "@likashefqet/react-native-image-zoom@4.2.0": "patches/@likashefqet__react-native-image-zoom@4.2.0.patch",
       "react-cosmos@6.1.1": "patches/react-cosmos@6.1.1.patch",
       "@10play/tentap-editor@0.5.21": "patches/@10play__tentap-editor@0.5.21.patch",
       "drizzle-orm@0.39.3": "patches/drizzle-orm@0.39.3.patch",
       "@tamagui/sheet@1.126.12": "patches/@tamagui__sheet@1.126.12.patch",
-      "react-native-country-codes-picker@2.3.5": "patches/react-native-country-codes-picker@2.3.5.patch",
-      "expo@52.0.47": "patches/expo@52.0.47.patch",
-      "expo-background-task@0.1.4": "patches/expo-background-task@0.1.4.patch",
-      "expo-localization@16.0.1": "patches/expo-localization@16.0.1.patch"
```
janicduplessis: Looks like expo added a default case, not same value as we had, but should be fine

https://github.com/expo/expo/blob/7d69e7580950831d6832a29db35379e9ad02b172/packages/expo-localization/ios/LocalizationModule.swift#L127-L129

apps/tlon-web/e2e/group-channel-management.spec.ts
```diff
@@ -48,36 +48,38 @@ test('should handle channel management operations', async ({ zodPage }) => {
     timeout: 5000,
   });
 
-  // Verify channel count is now 2 by checking for both channels
-  await expect(page.getByTestId('ChannelItem-General-0')).toBeVisible({
-    timeout: 10000,
-  });
+  // Verify channel count is now 2 by checking for both channels.
```
janicduplessis, line 51: To clarify why a lot of changes in tests were required - react navigation v7 drawer layout changed the order in the DOM of the main content and the drawer, so a lot of places that we have a selector that returns multiple instances assumed a certain order, but it is now reversed.

PR #5648, opened 2026-03-26
desk: 408 compatibility

Fang-:
## Summary

The zuse 408 update will contain hoon 135, which changes the type of `$type` from hoon 136. This affects everything that stores types, most commonly in vases. Groups desk is affected by this change in three ways:
- Some agent wrappers stored types, either directly (discipline) or due to accidental double-vasing of the inner agent's state (negotiate, vita-client).
- Most agents support importing from `$egg-any:gall`, logic for which now needs to properly upgrade old eggs to access "latest" state vases.
- Channels-server stored vases for its hooks functionality and will need to migrate those appropriately. Here, we drop them during `+on-load`.

We make all three of those changes. Details as follows.

## Changes

Discipline library stores types of marks so that, during load, it can nest-check new types against old and vice versa, to ensure mark types have remained unchanged. Migrating these is a simple `+next-type:h136` call.

Negotiate and vita-client libraries had poorly-written +on-save arms that stored the state vase of the inner agent inside a vase of their own. This double-vasing means that care must be taken when extracting inner agent state during `+on-load`: old versions of library state will need to upgrade those h136 vases into h135 ones before injecting them into the inner agent. We correct this double-vasing going forward.

(We could probably get rid of vita-client altogether, but that would require a state migration for groups-ui agent. We'll punt on it!)

Agents and libraries that deal with `%egg-any` import pokes are updated to call the `+latest:egg-aid:gall` helper before working with the egg. This dedents the code out of the egg version switch, but generally leaves the import code unchanged. The exceptions are the aforementioned libraries, which need to account for the double-vasing that might still be present in old exports.

For channels-server, it's possible to retain hook state during migration by upgrading the vases. However, this means migrating the agent proper, tying its versioning tightly into the release of 408. Since hooks are still experimental, we opt to drop their states instead. In order to do so safely across the kelvin upgrade, we "blank out" the vases in the hooks types that we extract from the state vase in `+on-load`, and add a step that "inflates" those blanked out types into the real thing. In practice, currently, that means ignoring what was in state, recompiling the hook from scratch, and dropping all its state on the floor.

cc @arthyn should we be sending facts/notifications about the changes here? Is dropping just state for awaiting-timer hooks even sane, or should we drop them wholesale?

With this set of changes, the agent _wrapper libraries_ are tied into the 408 release, but _agents themselves_ are free to gain additional migrations either before or after the 408 release without worry.

Also includes the one place I've see so far that gives TMI due to urbit/urbit#7298.

~~The lack of changes to verifier and lanyard assume that `+crub:crypto` will retain its original interface as a proxy for the new `+cric:crypto`. At the time of writing, that is not yet the case, we'll want to confirm before merge/release.~~ We might get suite C-style keys out of jael, so we ~~need to~~ use the new shared interface.

## How did I test?

- Brought a fakezod up to 408, revived the groups desk with these changes.
- Committed these changes onto a 409 fakezod, then applied the 408 base update.

In both cases, observed that obvious state and basic functionality was retained post-migration.

## Risks and impact

Definitely not safe to rollback without consulting PR author.

## Rollback plan

408 should be treated as a one-way migration.


mikolajpp (approved):
Great work, anxiously awaiting 408 release!

desk/app/channels-server.hoon
```diff
@@ -185,37 +236,42 @@
         state-1
         state-0
     ==
-  +$  state-14  current-state
+  +$  state-14
```
arthyn, line 239: kinda strange but I think I get why current-state is not in versioned-state

arthyn (approved):
wow look at this gall slop

PR #5668, opened 2026-04-01
guard: dynamically-ish typed userspace proof of concept

Fang-:
`/lib/guard` attempts to detect mark-mismatching data at compile-time. We do this by loading in our marks and replacing the `$cage`s in `%poke` and `%fact` cards with a `$rail` type, which is a `%$` over all the imported marks.

This approach has some advantages:
- Compile-time mark/type-validation of cages! Can easily do this for pokes, facts, also `+on-peek` results.
- It's just functions. Take them on/off as relevant. No agent wrapper needed.
- Keeps marks warm. (But does that keep conversions warm?)
  - Moot anyway once "build all /mar files" gets enabled...
- You get an untyped/unsafe "out" but you must explicitly write `%unsafe`.

~~Draft because this is incomplete (only chat agent uses it rn)~~ and wants some additional questions answered.
- Ergonomics around helper libraries (logging, /lib/dm) is a bit awkward. Maybe we can make nicer utilities.
- Is there some way to combine this with discipline-style API specification?
- ~~Do we just want to chuck every `/mar` file from our desk into `/lib/guardian`? You can also imagine per-agent lib files for this, and you may _need_ to do that if you also specify API details, but rn we could get away with a single (huge) file.~~
  - ~~Doing so would worsen our "changing one file recompiles _everything_" problem though. But maybe that's a price we can live with for what we gain.~~

Notably this already found a bad type in `+refs:migrate`. (;


arthyn (approved):
lgtm this is the beginning of ruby on rails for gall

desk/mar/foreigns-2.hoon
```diff
@@ -0,0 +1,14 @@
+/-  gv=groups-ver
+/+  j=groups-json
+|_  =foreigns:v10:gv
```
Fang-, line 3: Review: this was missing even though groups was using it for its latest foreigns scry endpoint. Correct to add this, yes?

desk/app/channels-server.hoon
```diff
@@ -1638,28 +1629,28 @@
...
     ::  originating post update has time to commit/propagate first.
-    =/  =cage  channel-action-2+!>(`a-channels:v10:cv`a-channels.effect)
-    (emit-late [%pass /hooks/effect %agent [our.bowl %channels] %poke cage])
+    =/  =rail  channel-action-2+`a-channels:v10:cv`a-channels.effect
+    (emit-late [%pass /hooks/effect %agent [our.bowl %channels] %poke rail])
   ::
       %groups
-    =/  =cage  group-action-4+!>(`a-groups:v7:gv`a-groups.effect)
-    (emit [%pass /hooks/effect %agent [our.bowl %groups] %poke cage])
+    =/  =rail  !! ::group-action-4+`a-groups:v7:gv`a-groups.effect  ::REVIEW
```
Fang-: This and the other nearby cases suggest hook implementation has drifted wrt the types. @arthyn confirmed. We'll probably want to clean these up somehow. Could be trivial, I haven't given it a serious try yet.

desk/app/channels.hoon
```diff
@@ -1341,7 +1232,8 @@
       ((slog tank u.p.sign) cor)
     ::
         %fact
-      (take-groups !<(=r-groups:v9:gv q.cage.sign))
+      ?>  ?=(%group-response-1 -.rail.sign)  ::REVIEW
+      (take-groups p.rail.sign)
```
Fang-, line 1275: Review: this one wasn't checking for the mark, just went straight to de-vasing. This is the correct mark, yes?

Fang- (commented):
Ready enough for review.

Some stuff highlighted for extra-care review. There was more here, but #5764 resolved most of it!

---

Note about the discipline stuff: I'm not a fan. Needing to explicate every mark that should be non-strict-ed (in the rail generator) is probably not the modality we want.

With some rejiggering we could probably move back to agents specifying the marks which they _do_ want strict-checked, whilst making `/lib/discipline` pull in the mark types from the `/lib/rail` instead of depending on agents to pass them.

afaic I'll consider that an outstanding TODO, but I'm open to hearing feedback/ideas.

---

Note about other libraries: lib-negotiate and lib-subscriber are quite low impact wrt this. lib-logs is a bit rougher, but I'm also hesitant to make that _always_ produce type-safe cards. afaic the current situation is fine, especially if we clean the local patches up just a tad. But open to feedback.

desk/mar/foreigns-2.hoon
mikolajpp: Yes, it should have been included! We should also remember to update the discipline while we are at it; looks like discipline for the singular `foreign-2` mark is also missing.

desk/app/channels-server.hoon
mikolajpp: The two main problems with hooks is that (1) they are not in real use and (2) we have never sat down to describe what should happen when we update hook type dependencies. I think they should probably use current types always and attempt to recompile when types change. The proper solution here is out of scope, but we should somehow make the types consistent so it is still possible to use hooks going forward.

desk/app/channels.hoon
mikolajpp: This is correct.

desk/lib/rail.hoon
```diff
@@ -0,0 +1,1878 @@
...
+::
+::  to update it, pulling in *all* marks in your current %groups desk, run:
+::  *=groups=/lib/rail/hoon +groups!rail
+::
+::  BE VERY CAREFUL, if there are mark files in your desk that are not in
+::  the "stock" distribution, the generated file will fail to built on
+::  other ships. CI *should* catch this, but remain mindful.
+::
+::  specifies a $rail type, a "typed $cage" containing all the desk's marks
+::  and a +de-rail helper for turning that back into a $cage.
```
mikolajpp: In what sense the `$rail` type is a "typed $cage"? It would rather seem be a typed `$page`. The `%unsafe` variant carries a vase, but all others just carry simple values, not vases.

desk/lib/guard.hoon
```diff
@@ -0,0 +1,72 @@
...
+++  on-agent
+  |*  f=$-([wire sign] *)
+  |=  [=wire =sign:agent:gall]
+  ?.  ?=(%fact -.sign)  (f wire sign)
+  (f wire sign(cage (en-rail cage.sign)))
+::
+++  exit
+  |=  cad=card:agent:gall
+  ^-  card
+  ?<  ?=(%slip -.cad)
```
mikolajpp, line 43: Interesting, never realized this was in the type. If this can never be passed to an agent, shouldn't gall exclude this variant?

desk/lib/guardian.hoon
```diff
@@ -0,0 +1,72 @@
...
+  |=  =cage
+  (f (en-rail cage))
+::
+++  on-agent
+  |*  f=$-([wire sign] *)
+  |=  [=wire =sign:agent:gall]
+  ?.  ?=(%fact -.sign)  (f wire sign)
+  (f wire sign(cage (en-rail cage.sign)))
+::
+++  exit
```
mikolajpp: The more Winnie the Pooh thought about this name, the more it seemed confusing :-)
The inverse of this arm is called `+card:unguard`. Here we say `exit`, so at the very least we should make it clear in the comment what place do we exit. It looks like we mean to say that we exit standard gall mode, and thus enter into the guardian mode, but if that's the case we would rather say `+enter:guardian`, or `+guard-card` in actual usage.

desk/gen/rail.hoon
```diff
@@ -0,0 +1,191 @@
+::  call as *=groups=/lib/rail/hoon +groups!rail
+::
+::TODO  investigate adding a "meta-build" for rune that evaluates the file
+::      to obtain a hoon ast, which it then compiles into the result
```
mikolajpp:
```suggestion
::TODO  investigate adding a "meta-build" ford rune that evaluates the file
::      to obtain a hoon ast, which it then compiles into the result
```

desk/gen/rail.hoon
```diff
@@ -0,0 +1,191 @@
+::  call as *=groups=/lib/rail/hoon +groups!rail
```
mikolajpp:
```suggestion
::  guardian rail api generator
::
::  call as *=groups=/lib/rail/hoon +groups!rail
::
```

desk/gen/rail.hoon
```diff
@@ -0,0 +1,191 @@
...
+      %channel-said-1
+      %channel-said-2
+      %channel-update
+      %hook-channel-preview
+      %hook-full
+      %hook-response-0
+      %hook-template
+    ::
+      %group-update
+      %group-log
```
mikolajpp: `%group-command` is missing.

desk/gen/rail.hoon
```diff
@@ -0,0 +1,191 @@
...
+    ::
+      %activity-feed
+      %activity-feed-init
+      %activity-full
+      %activity-full-1
+      %activity-summary
+      %activity-summary-1
+      %activity-update
+      %activity-update-1
+      %aqua-effect
```
mikolajpp: It is not ideal that we have to care about the marks that are not going to be used in any of our agents, but I understand this is because we scan all the marks in the desk, and thus need to specify all exceptions to allow for future changes.

This makes me think whether we should come up with a similar wrapper for threads, but this would be out of scope and rather low-priority, since historically we have not such problems with threads.

mikolajpp: One other alternative with perhaps better semantics would be to have a mark exclusion list for marks that are not of interest, which there are some.

desk/gen/rail.hoon
```diff
@@ -0,0 +1,191 @@
...
+::
+=/  [maz=(list mark) max=@ud]
+  %+  roll
+    .^((list path) %ct (en-beam bec(r da+now) /mar))
+  |=  [pax=path [maz=(list mark) max=@ud]]
+  ?.  =(%hoon (rear pax))  [maz max]
+  =/  mar=mark  (rap 3 (join '-' (snip (slag 1 pax))))
+  ?:  |(=('rail' mar) =('de-rail' mar) =('cage' mar) =('unsafe' mar))
+    ::TODO  be more clever?
+    %-  (slog (cat 3 'rail dropping "reserved" name: %' mar) ~)
```
mikolajpp, line 30: This conditional could also lookup an exclusion set, as I commented above, with these and other marks.

mikolajpp: That said, having read through later parts of this file I don't see why we would need this conditional which checks for marks we don't actually have in the desk: `rail`, `de-rail` and `unsafe`. Is that an omission?

desk/app/chat.hoon
```diff
@@ -325,11 +273,13 @@
     log   ~(. logs [our.bowl /logs])
     ol    (kol gte)
     log      ~(. logs [our.bowl /logs])
+    tell-log  (cork tell:log exit:guardian)
+    fail-log  (cork fail:log exit:guardian)
```
mikolajpp: I left a comment in `/lib/guardian.hoon` how this is confusing. We actually _enter_ into the guardian mode here, taking a standard gall card and transforming it into a rail, not the other way around.

mikolajpp (commented):
Reviewed all library files. Answered the questions.
The only substantial question I have thus far is regarding not having a dedicated mark exclusion list in the code generator.

We should also accompany, or follow this PR with a short documentation about this pattern (a few terse paragraphs, with a few code samples) , given that it substantially alters the usual gall agent api. This is not a blocker for the release obviously, but would be good to have to aid models in comprehending our codebase.

desk/lib/guard.hoon
Fang-: It _can_ be passed by agents! But nobody does it, and I'm not sure why it's supported. Indeed, in the gall 2026 work, we encountered this too and are considering dropping it from the userspace API.

desk/gen/rail.hoon
Fang-: We do this to avoid naming conflicts: the generated code contains types and arms `$rail` and `+de-rail`, expects to use the `$cage` type from the stdlib and puts mark names as the tag in a `$%` that also has an `%unsafe` case.

If we allowed this, we might end up with multiple `$rail` definitions or `%unsafe` cases.

desk/lib/guardian.hoon
Fang-: Yeah I'm open to bikeshedding the names here. Thinking was that you need this arm only if you're writing code in "guarded mode", and the result here is always `%unsafe`, so in that sense you're "existing" the guardrails implicitly.

But the other direction is also correct: you're turning the card _into_ a "guarded mode" card.

desk/lib/rail.hoon
Fang-: Perhaps it's more nuanced to say "statically typed $cage"?

The thinking is that the `$rail` gets put in all the places we would otherwise put `$cage`s (pokes, facts), and that they're now typed in such a way (dynamically) that the compiler can see the types. Vases always have a single (static) type. They _contain_ the information about the type of the data, but not in a way that's legible to the compiler.

desk/lib/guardian.hoon
mikolajpp: I see. I think we already talked about this previously but I forgot this line of thinking. I think a comment along the lines of `+exit: lift guardrails` will be informative enough for any passerby.

desk/app/channels-server.hoon
```diff
@@ -151,7 +140,7 @@
   =+  !<(old=versioned-state vase)
   =?  old  ?=(%0 -.old)  (state-0-to-1 old)
   =?  old  ?=(%1 -.old)  (state-1-to-2 old)
-  =?  cor  ?=(%2 -.old)  (emit %pass /trim %agent [our.bowl %chat] %poke %chat-trim !>(~))
+  =?  cor  ?=(%2 -.old)  (emit %pass /trim %agent [our.bowl %chat] %poke %unsafe %chat-trim !>(~))
```
mikolajpp, line 143: We use unsafe here because we don't have a corresponding mark?
Probably fine, given that this is historical. If gall will start to enforce marks on local pokes we will be able to easily find these cases with guardian.

mikolajpp (approved):
I have reviewed agent integrations. Looked more closely at `channels-server.hoon`, and skimmed over the other agents. Any mistakes or omissions would cause certain compile errors, so while changes are extensive they only amount to replacing cages with rails and consequences.

It seems like these changes should not cause any failures in any of the tests, but based on the logs the desk fails to upgrade. We should investigate it locally before merging.

desk/app/channels-server.hoon
Fang-: Correct.

desk/gen/rail.hoon
Fang-: > we have to care about the marks that are not going to be used in any of our agents

We could consider removing them from our desk. `%aqua-effect` for example is only included here (and in the generated output) because we have a `/mar/aqua/effect.hoon` in the groups desk distribution.

Having a mark exclusion list would let us clean those out of the library before even cleaning up desks. (We certainly don't want to do desk cleanup while the library still depends on those marks!)

arthyn (approved):
lgtm

desk/lib/discipline.hoon
```diff
@@ -238,16 +239,17 @@
...
+      =.  last-marks  (build-types-map strict-marks)
       =^  cards  inner  (on-load:og ole)  !:
       =.  cards  (check-cards:help ~ cards)
       [cards this]
     =+  !<([%discipline old=state-0] (slot 2 ole))
     =.  state  old
-    ?^  bad=(check-marks:help last-marks mark-map)
+    =/  next-marks  (build-types-map strict-marks)
+    ?^  bad=(check-marks:help last-marks next-marks)
       ~|  [%discipline dap=dap.bowl %mark-types-changed marks=;;((list mark) bad)]
```
mikolajpp: `;;((list mark) bad)`
Funny this is required. Is it because of the `?^` check?

desk/lib/discipline.hoon
```diff
@@ -119,6 +112,15 @@
     `mark.i.scries
   $(scries t.scries)
 ::
+++  build-types-map
+  |=  strict-marks=(set mark)
+  %+  roll  ~(tap in strict-marks)
+  |=  [=mark types=(map mark type)]
+  ?:  (~(has by types:rail) mark)
+    (~(put by types) mark (~(got by types:rail) mark))
```
mikolajpp: This runs on upgrades only, but typically one would capture a value with `+get:by` and dispatch on that, rather than search the map twice.

mikolajpp (approved):
The new guard interface looks good. Doing away with strictness flag in favor of inferring non-strict marks is particularly nice.

desk/lib/discipline.hoon
Fang-: I don't think it's required, I think I just put it in to pretty-print it like a normal list. Perhaps a normal cast could work, not sure why I `;;`ed here...

desk/lib/activity-json.hoon
```diff
@@ -261,6 +261,34 @@
...
+          time+(time time.te)
+          event+(event event.te)
+      ==
+    ::
+    ++  adjusted
+      |=  [s=source:v9:av v=(unit volume-map:v9:av)]
+      %-  pairs
+      :~  source+(source s)
+          volume+?~(v ~ (volume-map u.v))
+      ==
```
Fang-: For the record, had to add these to get `/mar/activity/update-5` to compile.

desk/gen/rail.hoon
mikolajpp: `%aqua-effect` is required for CI to run though, so it must stay. It was added in my first aqua tests PR.

Fang-:
Noting for the record that this isn't merged yet because:
1. tests/CI need patching,
2. without ford fusion (408) this has a pretty hefty impact on build/iteration times.

backend/run-aqua-tests.sh
```diff
@@ -43,7 +43,7 @@ echo "Running aqua tests"
 
 #download_url=`jq -r ".[\"$ship\"][\"downloadUrl\"]" < $ship_manifest`
 download_url="https://bootstrap.urbit.org/zod-aqua-tests-409k.xst"
-pill_download_url="https://bootstrap.urbit.org/aqua-tests-v10-1-0.pill"
+pill_download_url="https://bootstrap.urbit.org/groups-v11-2-2.pill"
```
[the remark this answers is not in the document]
mikolajpp: Nonsense, someone please bring a tape so we can gag this model.

desk/app/groups.hoon
```diff
@@ -4579,6 +4614,8 @@
     ::
     =/  r-groups-9=r-groups:v9:gv  [flag r-group]
     =/  v1-paths  ~[/v1/groups [%v1 go-area]]
+    =*  log  ~(. l `'group-join')
+    ::=.  cor  (tell:log %dbug leaf+"+go-response /v1/groups fact {<r-groups-9>}" ~)
```
mikolajpp: Not sure what happened here? I have intentionally removed debug prints until logs interface is adjusted to use traps.

desk/app/groups.hoon
```diff
@@ -4756,7 +4793,10 @@
       %-  gang:v2:foreign:v7:gc
       (v7:foreign:v10:gc foreign)
     =/  foreign-8  (v8:foreign:v10:gc foreign)
-    =.  cor  (give %fact ~[/v1/foreigns] foreigns-1+`foreigns:v8:gv`(my flag^foreign-8 ~))
+    =/  foreigns-8=foreigns:v8:gv  (my flag^foreign-8 ~)
+    =*  log  ~(. l `'group-join')
+    ::=.  cor  (tell:log %dbug leaf+"+fi-give-update /v1/foreigns fact {<foreigns-8>}" ~)
+    =.  cor  (give %fact ~[/v1/foreigns] foreigns-1+foreigns-8)
```
mikolajpp: Same as above.

mikolajpp (commented):
Some nits about bringing back commented dbug prints in groups. 

The important issue is the new egg-migration code in several agents. As far as I can tell this is pre-existing implementation by @Fang- simply merged back? If so, it should be fine to merge.

desk/lib/rail.hoon
```diff
@@ -7,771 +7,701 @@
 ::  the "stock" distribution, the generated file will fail to built on
 ::  other ships. CI *should* catch this, but remain mindful.
 ::
-::  specifies a $rail type, a "typed $cage" containing all the desk's marks
+::  specifies a $rail type, a "typed $cage" containing all the desk\'s marks
```
mikolajpp: There is a spurious `\` in comment.

desk/lib/rail.hoon
mikolajpp, line 1: I just recalled this is autogenerated. In that case we should only fix the generator not to insert the spurious escape in the comment there.

mikolajpp (commented):
There is something off with the rail generator: it inserts escaped quote verbatim into comments.

desk/lib/rail.hoon
arthyn: this is because the generator spits out a cord to dojo. I forgot to find and replace all `\` escapes in addition to replacing all the `\0a` 's

PR #5669, opened 2026-04-02
filteredchatlist: make sure we use flashlist correctly

arthyn:
## Summary

I think something changed with FlashList and this was required for it to actually re-render when `selectedIndex` changed, been bothering me because I couldn't actually see the selection when using the leap bar.

## Changes

- added extra data for FlashList to know to re-render

## How did I test?

Manually

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:


dnbrwstr (approved):
Lgtm

PR #5671, opened 2026-04-02
run prettier over all .{ts,tsx} files; add Github check for prettier conformance

davidisaaclee:
I updated all .ts / .tsx files by running:
```sh
pnpm prettier --write '**/*.{ts,tsx}' --ignore-path .gitignore
```

This is important because any change I was making in a non-passing file was creating diffs throughout the file.

I also added a step to the Github action that should block merge if code does not match Prettier.

**Complication:** We need to ignore e.g. `packages/api/dist/**/*` – prettier-ignored files map closely enough to .gitignore, but prettier does not support nested/multiple `.gitignore`s (e.g. `packages/api/dist/**/*` is ignored via `packages/api/.gitignore`).

[Add helper to build .prettierignore from nested .gitignores](https://github.com/tloncorp/tlon-apps/pull/5671/commits/37619d763f1ff888a7ab81e8498ac248a610ca5b) uses a third-party `prettierignore-monorepo` helper that compiles all nested `.gitignore`s into a single `.prettierignore`. To ensure that this is kept up-to-date, I also added a check to the CI gh action that `.prettierignore` matches exactly the result of running the `prettierignore-monorepo` helper.


dnbrwstr (approved):
Lgtm, I was thinking of doing the same thing. Let's merge in asap so we don't sprout conflicts.

davidisaaclee:
I was thinking that this will miss a certain class of offense: PR branch passes check; target branch passes check; but once merged, merged code check fails. This will end up with a failing `develop`.

PR #5676, opened 2026-04-03
feat: add gateway-status agent for tlon bot availability

patosullivan:
## Summary

Resolves TLON-5534

Adds a new `%gateway-status` Gall agent plus `@tloncorp/api` helpers to support lease-based gateway liveness for the Tlon bot. The agent tracks owner DM recency, sends offline auto-replies while the gateway is down, sends a back-online notice after restart when the owner was recently active, and uses a heartbeat lease so it can distinguish graceful shutdowns from hard crashes.

The final implementation also follows local Hoon conventions more closely: shared protocol types live in `sur/`, while agent state and app-local helper logic live in the app file.

## Changes

- Added `desk/app/gateway-status.hoon` with the new Gall agent, including the app-local `state-0`, liveness helper, `%activity /v4` subscription, lease/timer handling, and `%chat` DM notices
- Added `desk/sur/gateway-status.hoon` with the shared `status`, `action`, and `update` types plus `v1` indirection
- Added `desk/mar/gateway-status/action-1.hoon` for the inbound JSON action protocol (`%configure`, `%gateway-start`, `%gateway-heartbeat`, `%gateway-stop`)
- Added `desk/mar/gateway-status/update-1.hoon` for outbound `%noun` subscription updates (`%status`, `%owner-activity`, `%auto-reply`)
- Added `desk/tests/app/gateway-status.hoon` with 17 agent-level tests covering init, configure, lifecycle state transitions, owner DM auto-reply, cooldown suppression, deduplication, stale stop/heartbeat protection, and lease-expiry recovery
- Registered `%gateway-status` in `desk/desk.bill`
- Added `packages/api/src/urbit/gatewayStatus.ts` and `packages/api/src/client/gatewayStatusApi.ts` for the `%gateway-status` poke surface (`configureGatewayStatus`, `gatewayStart`, `gatewayHeartbeat`, `gatewayStop`)
- Added `packages/api/src/__tests__/gatewayStatusApi.test.ts` and re-exported the new API helpers from the package entrypoints
- Added `docs/gateway-status.md` as an implementation-facing spec for the new agent
- Updated `CLAUDE.md` with repo-specific Hoon guidance around agent structure, `sur/` responsibilities, type naming, mark conventions, and testing
- Bumped `packages/api/package.json` to `0.0.5`

## How did I test?

- Ran the 17 Hoon agent tests via the `test-agent` harness
- Ran the TypeScript unit tests for the new `@tloncorp/api` helpers
- Built the updated Hoon files directly on fake ships while iterating on the agent shape and mark surface
- Manually verified the full lifecycle with local fake ships (`~zod` bot, `~ten` owner) and the `openclaw-tlon` dev environment:
  - startup transitions `%gateway-status` to `%up`
  - heartbeat extends `lease-until`
  - graceful gateway shutdown transitions status to `%down`
  - owner DMs while down receive the offline auto-reply
  - cooldown suppresses repeated offline auto-replies within 5 minutes
  - restart after graceful shutdown sends the back-online notice and clears the pending restart flag
  - ungraceful gateway kill leaves status `%up` until the lease expires, then transitions to `%down`
  - owner DMs after crash-driven lease expiry also receive the offline auto-reply
  - restart after crash-driven lease expiry also sends the back-online notice

## Risks and impact

- Affects important code area: backend agent surface and `@tloncorp/api` integration

## Rollback plan

Revert.

### DIFF
--- a/desk/app/gateway-status.hoon
+++ b/desk/app/gateway-status.hoon
@@ -1,6 +1,6 @@
 ::  gateway-status: offline-reply and liveness agent for openclaw gateway
 ::
-/-  gs=gateway-status, a=activity, c=chat, s=story
+/-  gs=gateway-status, a=activity, cv=chat-ver, s=story
 /+  default-agent, verb, dbug
 |%
 +$  card  card:agent:gall
@@ -32,7 +32,7 @@
   ++  on-agent
     |=  [=wire =sign:agent:gall]
     ^-  (quip card _this)
-    =^  cards  state  abet:(agnt:cor wire sign)
+    =^  cards  state  abet:(agent:cor wire sign)
     [cards this]
   ++  on-arvo
     |=  [=wire sign=sign-arvo]
@@ -42,7 +42,7 @@
   ++  on-watch
     |=  =path
     ^-  (quip card _this)
-    ?>  =(src.bowl our.bowl)
+    ?>  =(src our):bowl
     ?+  path  (on-watch:def path)
         [%v1 ~]
       :_  this
@@ -73,20 +73,24 @@
   ^+  cor
   (emit %pass /activity %agent [our.bowl %activity] %watch /v4)
 ::
+++  poke
+  |=  [=mark =vase]
+  ^+  cor
+  ?>  =(src our):bowl
+  ?>  ?=(%gateway-status-action-1 mark)
+  =/  act  !<(action-1:gs vase)
+  ?-  -.act
+    %configure          (handle-configure +.act)
+    %gateway-start      (handle-gateway-start +.act)
+    %gateway-heartbeat  (handle-gateway-heartbeat +.act)
+    %gateway-stop       (handle-gateway-stop +.act)
+  ==
+::
 ++  cancel-lease-timer
   ^+  cor
   =/  lut  lease-until
   ?~  lut  cor
   (emit %pass /lease-check %arvo %b %rest u.lut)
-::
-++  send-dm
-  |=  text=@t
-  ^+  cor
-  =/  content=story:s  ~[[%inline ~[text]]]
-  =/  =essay:c  [[content our.bowl now.bowl] chat+/ ~ ~]
-  =/  =diff:dm:c  [[our.bowl now.bowl] %add essay `now.bowl]
-  =/  =action:dm:c  [(need owner) diff]
-  (emit %pass /dm/send %agent [our.bowl %chat] %poke %chat-dm-action-2 !>(action))
 ::
 ++  status-update
   |=  [sts=?(%unknown %up %down) lut=(unit @da)]
@@ -100,28 +104,14 @@
       (lth (sub now last-owner-message-at) active-window)
   ==
 ::
-++  should-auto-reply
-  |=  current-key=message-key:a
-  ^-  ?
-  ?:  (is-gateway-live:gs status lease-until now.bowl)  %.n
-  =/  lrt  last-offline-auto-reply-to
-  ?:  ?&(?=(^ lrt) =(u.lrt current-key))  %.n
-  =/  lra  last-offline-auto-reply-at
-  ?:  ?&(?=(^ lra) (lth (sub now.bowl u.lra) offline-reply-cooldown))  %.n
-  %.y
-::
-++  poke
-  |=  [=mark =vase]
+++  send-dm
+  |=  text=@t
   ^+  cor
-  ?>  =(src.bowl our.bowl)
-  ?>  ?=(%gateway-status-action-1 mark)
-  =/  act  !<(action-1:gs vase)
-  ?-  -.act
-    %configure          (handle-configure +.act)
-    %gateway-start      (handle-gateway-start +.act)
-    %gateway-heartbeat  (handle-gateway-heartbeat +.act)
-    %gateway-stop       (handle-gateway-stop +.act)
-  ==
+  =/  content=story:s  ~[[%inline ~[text]]]
+  =/  =essay:v7:cv  [[content our.bowl now.bowl] chat+/ ~ ~]
+  =/  =diff:dm:v7:cv  [[our.bowl now.bowl] %add essay `now.bowl]
+  =/  =action:dm:v7:cv  [(need owner) diff]
+  (emit %pass /dm/send %agent [our.bowl %chat] %poke %chat-dm-action-2 !>(action))
 ::
 ++  handle-configure
   |=  [who=ship aw=@dr orc=@dr]
@@ -175,7 +165,7 @@
     (send-dm 'Your Tlon bot is restarting. I should be back shortly. 🔧')
   (status-update status lease-until)
 ::
-++  agnt
+++  agent
   |=  [=wire =sign:agent:gall]
   ^+  cor
   ?+  wire  cor
@@ -201,6 +191,16 @@
       ((slog 'gateway-status: dm send failed' u.p.sign) cor)
     ==
   ==
+::
+++  should-auto-reply
+  |=  current-key=message-key:a
+  ^-  ?
+  ?:  (is-gateway-live:gs status lease-until now.bowl)  %.n
+  =/  lrt  last-offline-auto-reply-to
+  ?:  ?&(?=(^ lrt) =(u.lrt current-key))  %.n
+  =/  lra  last-offline-auto-reply-at
+  ?:  ?&(?=(^ lra) (lth (sub now.bowl u.lra) offline-reply-cooldown))  %.n
+  %.y
 ::
 ++  handle-activity-add
   |=  [who=ship =source:a =event:a]
@@ -231,9 +231,8 @@
       [%lease-check ~]
     ?>  ?=([%behn %wake *] sign)
     ?.  =(status %up)  cor
-    =/  lut  lease-until
-    ?~  lut  cor
-    ?.  (lte u.lut now.bowl)  cor
+    ?~  lease-until  cor
+    ?.  (lte u.lease-until now.bowl)  cor
     %-  (slog leaf+"gateway-status: lease expired, transitioning to down" ~)
     =.  status  %down
     =.  pending-restart-notice  %.y
--- a/desk/app/gateway-status.hoon
+++ b/desk/app/gateway-status.hoon
@@ -4,9 +4,42 @@
 /+  default-agent, verb, dbug
 |%
 +$  card  card:agent:gall
-+$  versioned-state  state-0:gs
+::  $state-0: agent state
+::
+::    .owner: ship whose DMs trigger offline replies (~ = unconfigured/inert)
+::    .last-owner-msg: timestamp of most recent owner DM
+::    .last-owner-msg-id: key of most recent owner DM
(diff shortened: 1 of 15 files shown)


desk/app/gateway-status.hoon
```diff
@@ -0,0 +1,242 @@
...
+++  arvo
+  |=  [=wire sign=sign-arvo]
+  ^+  cor
+  ?+  wire  cor
+      [%lease-check ~]
+    ?>  ?=([%behn %wake *] sign)
+    ?.  =(status %up)  cor
+    =/  lut  lease-until
+    ?~  lut  cor
+    ?.  (lte u.lut now.bowl)  cor
```
arthyn:
```suggestion
    ?~  lease-until  cor
    ?.  (lte u.lease-until now.bowl)  cor
```

desk/app/gateway-status.hoon
```diff
@@ -0,0 +1,242 @@
...
+  =.  cor  cancel-lease-timer
+  =.  status  %down
+  =.  boot-id  ~
+  =.  last-stop-at  `now.bowl
+  =.  pending-restart-notice  %.y
+  =?  cor  should-notify
+    (send-dm 'Your Tlon bot is restarting. I should be back shortly. 🔧')
+  (status-update status lease-until)
+::
+++  agnt
```
arthyn:
```suggestion
++  agent
```

desk/app/gateway-status.hoon
```diff
@@ -0,0 +1,242 @@
...
+  =/  lrt  last-offline-auto-reply-to
+  ?:  ?&(?=(^ lrt) =(u.lrt current-key))  %.n
+  =/  lra  last-offline-auto-reply-at
+  ?:  ?&(?=(^ lra) (lth (sub now.bowl u.lra) offline-reply-cooldown))  %.n
+  %.y
+::
+++  poke
+  |=  [=mark =vase]
+  ^+  cor
+  ?>  =(src.bowl our.bowl)
```
arthyn:
```suggestion
  ?>  =(src our):bowl
```

desk/app/gateway-status.hoon
```diff
@@ -0,0 +1,242 @@
...
+  ^+  cor
+  =/  lut  lease-until
+  ?~  lut  cor
+  (emit %pass /lease-check %arvo %b %rest u.lut)
+::
+++  send-dm
+  |=  text=@t
+  ^+  cor
+  =/  content=story:s  ~[[%inline ~[text]]]
+  =/  =essay:c  [[content our.bowl now.bowl] chat+/ ~ ~]
```
arthyn: would use versioned types here so these don't fall out of sync with the mark

desk/app/gateway-status.hoon
```diff
@@ -0,0 +1,242 @@
...
+++  cor   .
+++  abet  [(flop cards) state]
+++  emit  |=(=card cor(cards [card cards]))
+++  give  |=(=gift:agent:gall (emit %give gift))
+::
+++  init
+  ^+  cor
+  (emit %pass /activity %agent [our.bowl %activity] %watch /v4)
+::
+++  cancel-lease-timer
```
arthyn, line 122: I would move all these helper arms closer to where they are used

arthyn (approved):
lgtm! overall the names are a little more verbose than we normally do, but I'm ok with it. 

we may want an endpoint that we could hit from the client if we detect a hosted moon as the DM, for connection checker, but we'd probably need additional hosting support for that (or maybe just better to come from them idk).

desk/sur/gateway-status.hoon
```diff
@@ -0,0 +1,58 @@
...
+::    .status: gateway liveness as seen by the ship
+::    .boot-id: identifies the current gateway process (~ after graceful stop)
+::    .lease-until: when the current heartbeat lease expires
+::    .pending-restart-notice: whether to send a back-online DM on next start
+::    .last-owner-message-id: last owner DM key, used for reply deduplication
+::    .last-offline-auto-reply-to: key of the DM that last triggered an auto-reply
+::    .active-window: how recently owner must have messaged to get notices
+::    .offline-reply-cooldown: minimum interval between offline auto-replies
+::
++$  state-0
```
mikolajpp: The state definition should be moved to the main agent file. (See some other agents in the desk for placement.)

Also, the order of the comments does not match the order of fields, one of those should be reordered.

desk/sur/gateway-status.hoon
```diff
@@ -0,0 +1,58 @@
...
+  ==
+::  +is-gateway-live: check if the gateway is currently available
+::
+++  is-gateway-live
+  |=  [status=?(%unknown %up %down) lease-until=(unit @da) now=@da]
+  ^-  ?
+  ?&  ?=(%up status)
+      ?=(^ lease-until)
+      (gth u.lease-until now)
+  ==
```
mikolajpp: Only things that are shared between implementations should be defined in the structure file. These include types, but sometimes also helper functions, especially if these are helpful for working with some types.

However this is not the case here – this arm defines application logic and therefore should be moved to the agent file.

desk/sur/gateway-status.hoon
```diff
@@ -0,0 +1,58 @@
...
++$  action-1
+  $%  [%configure owner=ship active-window=@dr offline-reply-cooldown=@dr]
+      [%gateway-start boot-id=@t lease-until=@da]
+      [%gateway-heartbeat boot-id=@t lease-until=@da]
+      [%gateway-stop boot-id=@t reason=@t]
+  ==
+::  $update-1: outbound subscription facts for status observers
+::
++$  update-1
+  $%  [%status status=?(%unknown %up %down) lease-until=(unit @da)]
```
mikolajpp: `.status` here shares the type with the application state - it should be factored out as a separate type definition.

desk/sur/gateway-status.hoon
```diff
@@ -0,0 +1,58 @@
...
+::
++$  action-1
+  $%  [%configure owner=ship active-window=@dr offline-reply-cooldown=@dr]
+      [%gateway-start boot-id=@t lease-until=@da]
+      [%gateway-heartbeat boot-id=@t lease-until=@da]
+      [%gateway-stop boot-id=@t reason=@t]
+  ==
+::  $update-1: outbound subscription facts for status observers
+::
++$  update-1
```
mikolajpp: The naming convention from marks does not carry to types. For now, it should be ok to just leave these without versioning in a dedicated file, given the small surface area of this agent. A `v1` arm in the sur file referring to the current core should be enough. Then, all references to these types would become `update:v1:gs` and similar for `$action`

```suggestion
+$  update
```

desk/sur/gateway-status.hoon
```diff
@@ -0,0 +1,58 @@
...
+      last-start-at=(unit @da)
+      pending-restart-notice=?
+      last-offline-auto-reply-at=(unit @da)
+      last-offline-auto-reply-to=(unit message-key:a)
+      offline-reply-cooldown=@dr
+      active-window=@dr
+  ==
+::  $action-1: inbound poke protocol from openclaw-tlon
+::
++$  action-1
```
mikolajpp: We should not version these with tags, see comment below.
```suggestion
::  $action: inbound poke protocol from openclaw-tlon
::
+$  action
```

desk/sur/gateway-status.hoon
```diff
@@ -0,0 +1,58 @@
+::  gateway-status: types for the gateway liveness and offline-reply agent
+::
+/-  a=activity
+|%
+::  $state-0: agent state
+::
+::    .owner: ship whose DMs trigger offline replies (~ = unconfigured/inert)
```
mikolajpp:
```suggestion
::    .owner: ship whose DMs trigger offline replies (~ means unconfigured/inert)
```

desk/sur/gateway-status.hoon
```diff
@@ -0,0 +1,58 @@
...
+::    .status: gateway liveness as seen by the ship
+::    .boot-id: identifies the current gateway process (~ after graceful stop)
+::    .lease-until: when the current heartbeat lease expires
+::    .pending-restart-notice: whether to send a back-online DM on next start
+::    .last-owner-message-id: last owner DM key, used for reply deduplication
```
mikolajpp: `last-owner-message-id` is not actually used for anything in the agent...

If these comments were LLM-generated we *must* make sure they reflect reality. Also, there is a whole bunch of fields that do not have comments, which makes it hard to pair while reading. We should include them here too.

desk/sur/gateway-status.hoon
```diff
@@ -0,0 +1,58 @@
...
+      last-owner-message-id=(unit message-key:a)
+      status=$~(%unknown ?(%unknown %up %down))
+      boot-id=(unit @t)
+      lease-until=(unit @da)
+      last-heartbeat-at=(unit @da)
+      last-stop-at=(unit @da)
+      last-start-at=(unit @da)
+      pending-restart-notice=?
+      last-offline-auto-reply-at=(unit @da)
+      last-offline-auto-reply-to=(unit message-key:a)
```
mikolajpp: These names are just too long: see how we use them in the agent code. I am sure shorter yet legible names can be found.

desk/sur/gateway-status.hoon
```diff
@@ -0,0 +1,58 @@
...
+  $:  %0
+      owner=(unit ship)
+      last-owner-message-at=@da
+      last-owner-message-id=(unit message-key:a)
+      status=$~(%unknown ?(%unknown %up %down))
+      boot-id=(unit @t)
+      lease-until=(unit @da)
+      last-heartbeat-at=(unit @da)
+      last-stop-at=(unit @da)
+      last-start-at=(unit @da)
```
mikolajpp: We tend to learn the types, so it is just fine to name do away with suffixes like these. `last-heartbeat` is going to be clear from context, even for an LLM ;-)

mikolajpp (changes requested):
First batch of comments after skimming the spec and the sur file. I haven't yet done a deep dive into the actual logic of the application, but these issues should be fixed first before we carry on.

desk/sur/gateway-status.hoon
mikolajpp: Furthermore, there is no reason to pass arguments that are in state if this arm is moved to its proper place.

desk/app/gateway-status.hoon
```diff
@@ -0,0 +1,241 @@
...
+  |=  [bid=@t lut=@da]
+  ^+  cor
+  =/  owner-guard  (need owner)  ::  crash if owner not configured
+  =/  should-notify  ?&(pending-restart-notice (is-owner-recently-active now.bowl))
+  =.  cor  cancel-lease-timer
+  =.  status  %up
+  =.  boot-id  `bid
+  =.  lease-until  `lut
+  =.  last-start-at  `now.bowl
+  =.  pending-restart-notice  %.n
```
mikolajpp: The use of `%.y` and `%.n` is not idiomatic, should use `&` and `|` instead.
```suggestion
  =.  pending-restart-notice  |
```

mikolajpp (changes requested):

PR #5677, opened 2026-04-03
fix get-intrinsic error

davidisaaclee:
We had two conflicting versions of `get-intrinsic` installed, which appears to have been causing the error `TypeError: GetIntrinsic is not a function (it is Object)` on iOS launch.

Pinning this dependency to prevent double-install fixes the issue on iOS simulator.

I went with `1.2.x` because (before pinning):
```sh
[I] tlon-apps ) pnpm why get-intrinsic | grep "get-intrinsic 1.2" | wc -l
   18745
[N] tlon-apps ) pnpm why get-intrinsic | grep "get-intrinsic 1.3" | wc -l
    1129
```

>This obviously won't scale if we get similar issues in the future. I think the "best" solution would be to find which package has a strict requirement for the `get-instrinsic` version and patch it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
       "prosemirror-model": "1.19.3",
       "prosemirror-view": "1.33.4",
       "prosemirror-state": "1.4.3",
-      "@tiptap/pm": "2.6.6"
+      "@tiptap/pm": "2.6.6",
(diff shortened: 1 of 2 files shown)


dnbrwstr (approved):
Lgtm

PR #5679, opened 2026-04-03
channels: meet all authors even bots

arthyn:
## Summary

OTT

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:


Fang- (approved):
Sure, we should get contact deets for ships even if we've only ever seen automated posts by them.

PR #5680, opened 2026-04-06
improve forward sheet behavior

dnbrwstr:
## Summary

This builds on #5582 and makes the "Send to channel" sheet feel better.

Search updates right away, channel results are more direct, and tapping a result with the keyboard open is more reliable.

## Changes

- make search bar updates immediate when debounce is `0`
- refactor and simplify search sheets
- improve bottom sheet keyboard, footer, and tap behavior for the forward sheet

## How did I test?

- Tested on iOS, android, and web


dnbrwstr:
Superseded by #5681, which uses the rebased  branch.

PR #5683, opened 2026-04-06
Improve channel/group settings discoverability and header UX

jamesacklin:
## Summary

In #5289 we drastically overcorrected for header uniformity by using too many pictorial icons vs clear text labels. This PR rectifies by replacing confusing cog/pen icons with explicit labels, removing redundant navigation entry points, and improving the header UX across channel, group, and notebook views.

Fixes TLON-5439
Fixes TLON-5545
Fixes TLON-5544
Fixes TLON-5466
Fixes TLON-4529
Fixes TLON-5496

## Changes

- **Admin actions:** Any admin action within a group, channel, or post header uses the alternate positive action text color (blue in the default themes). Other run-of-the-mill actions use the primary text color. This makes it easy to see what actions you're taking as an administrator.
- **Channel header**: Removed redundant "Edit" button (available via channel info screen instead, via "Rename"). Moved connectivity indicator dot from right controls to immediately left of channel name.
- **Group sidebar header**: Replaced cog icon/popover with new `EditList` icon (list + pencil) for "Edit channels" action.
- **Notebook/gallery post detail header**: Changed cog icon to "Edit" text button for editing posts (TLON-5466).
  - Edit button uses `$primaryText` when editing your own post.
  - Edit button uses `$positiveActionText` (green) when an admin is editing another user's post.
- **Chat details screen**: Replaced hidden pen icon with "Rename" text button for editing group/channel metadata.
- **Notebook/gallery "+" button**: Changed `Add` icon button to "New" text button.
- **Home screen**: Changed magnifying glass to filter icon, updated placeholder from "Find by name" to "Filter by name".
- **New icon**: Added `EditList.svg` — a list-with-pencil icon for the group channel management action.

## How did I test?

Visual verification in browser and iOS simulator. TypeScript typecheck passes.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert the commit. All changes are UI-only (icons, labels, layout) with no data or state implications.

## Screenshots / videos

| Screen | Description | Image |
|--------|-------------|-------|
| Home | Changes the magnifying glass (implying 'search') to a filter, which is what the control actually does - filters the list by name. | home-filter |
| Group | Introduces the EditList icon, which is a middle-ground between using a text label ("Edit Channels," which takes up too much room) and the all-too-ambiguous "Draw" icon. Note the blue color, since this is an admin action. | group-channels |
| Chat channel | Removes the redundant and confusing cog/settings icon, which pointed to editing the channel name/description, in favor of the plainly obvious dropdown caret. | in-channel |
| Channel details | Once the user taps the down-caret / channel title, they are now presented with a smack-dab obvious "Rename" action. Note the blue color, since this is an admin action. | channel-info-rename |
| Non-chat channels (Gallery, Notebook) | Switches from the somewhat-ambiguous + button to an explicit "New" text button. We might consider the same on the Home screen, open to that discussion. | notebook-new gallery-new |
| Editing your own post | If the user is editing their own post, regardless if they're an admin or not, they will see an explicit "Edit" button, rather than the ambiguous "Draw" icon. Note the black color here, since any user can edit their own post and this is not unique to admins. | gallery-edit-yours |
| Editing others' posts | Similar to the above. Switches the "Draw" icon to an explicit "Edit" text action for others' posts; visible only to admins. Note the blue color as a result. | notebook-edit-others |

### DIFF
--- a/apps/tlon-mobile/ios/Podfile.lock
+++ b/apps/tlon-mobile/ios/Podfile.lock
@@ -401,8 +401,8 @@ PODS:
   - hermes-engine (0.76.9):
     - hermes-engine/Pre-built (= 0.76.9)
   - hermes-engine/Pre-built (0.76.9)
-  - libavif/core (0.11.1)
-  - libavif/libdav1d (0.11.1):
+  - libavif/core (1.0.0)
+  - libavif/libdav1d (1.0.0):
     - libavif/core
     - libdav1d (>= 0.6.0)
   - libdav1d (1.2.0)
@@ -446,7 +446,7 @@ PODS:
     - ReactCommon/turbomodule/bridging
     - ReactCommon/turbomodule/core
     - Yoga
-  - PostHog (3.31.0)
+  - PostHog (3.49.1)
   - PromisesObjC (2.4.0)
   - PromisesSwift (2.4.0):
     - PromisesObjC (= 2.4.0)
@@ -2453,12 +2453,12 @@ PODS:
   - SDWebImage (5.19.7):
     - SDWebImage/Core (= 5.19.7)
   - SDWebImage/Core (5.19.7)
-  - SDWebImageAVIFCoder (0.11.0):
+  - SDWebImageAVIFCoder (0.11.1):
     - libavif/core (>= 0.11.0)
     - SDWebImage (~> 5.10)
   - SDWebImageSVGCoder (1.7.0):
     - SDWebImage/Core (~> 5.6)
-  - SDWebImageWebPCoder (0.14.6):
+  - SDWebImageWebPCoder (0.15.0):
     - libwebp (~> 1.0)
     - SDWebImage/Core (~> 5.17)
   - Sentry/HybridSDK (8.57.2)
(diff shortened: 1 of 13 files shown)


patosullivan (approved):
Looks great!

dnbrwstr (approved):
Lgtm overall. Couple thoughts:

- Personally, I'm not sure about the switch from magnifying glass to filter on the home screen. Magnifying glass often is used for this function and imo is more discoverable than the filter. 
- The blue color for admin actions seems kind of odd / unnecessary differentiation to me.  I think it'd make more sense to just stick with a single treatment for header action buttons.

jamesacklin:
> Lgtm overall. Couple thoughts:
> 
> * Personally, I'm not sure about the switch from magnifying glass to filter on the home screen. Magnifying glass often is used for this function and imo is more discoverable than the filter.
> * The blue color for admin actions seems kind of odd / unnecessary differentiation to me.  I think it'd make more sense to just stick with a single treatment for header action buttons.

OK - fair enough, I think the switch to text labels is helpful enough as it is, so I can de-complicate the blue/black conds and switch back to a magnifying glass.

My concern is new users mistaking the magnifying glass for "search the network" i.e. contact/group discovery, but I have yet to see any complaints on that so no need to fix what ain't broke. Will make those changes then merge.

PR #5688, opened 2026-04-07
fix note paste crash

dnbrwstr:
## Summary

This fixes a note editor crash when pasting rich links.

It also restores the current note content if the editor webview reloads after a crash so the user doesn't lose data if the webview crashes.

Fixes TLON-5552

## Changes

- patch `@10play/tentap-editor@0.5.21` to remove the old HTML link paste fallback that was crashing
- restore content after editor crashes

## How did I test?

Tested on web.


patosullivan (approved):
Looks good!

PR #5690, opened 2026-04-07
rube: ensure optionals, allow fresh boot, update to 25

arthyn:
## Summary

OTT, I found that our archive-piers script was not updating mug. Additionally, when attempting to apply the latest desk changes ran into lib/discipline issues on compile, so I made an option to archive freshly booted fakezods which fixed the issue.

sidenote: @Fang- @mikolajpp we probably should find a way to have some notion of prod vs dev and disable lib/discipline from crashing compiles in a "prod" environment

## Changes

- allowed archiving fresh fakezods
- made sure archive piers script includes optional ships
- updated all fakezods to 25 which includes latest from channels/chat protocol upgrade

## How did I test?

Ran CI for openclaw plugin

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:


mikolajpp:
Do we still have the discipline crash? lib-discipline causing crashes is definitely not something we should workaround with fresh piers. We had some type inconsistencies in the past, which were fixed. It is possible that there is still some affected mark with a strict flag that should be flipped and commented as such. See examples of this problem in channels, where some marks are marked with a special comment.
```
::TODO make strict one day (affected by versioning mistake)
```

arthyn:
@mikolajpp it was only present on mug, maybe because it wasn't always getting update in lock step with the others it failed? you can check out the pier with  `curl -fSL https://bootstrap.urbit.org/rube-mug25.tgz -o rube-mug24.tgz` and try to sync latest desk changes to it

patosullivan (approved):

PR #5691, opened 2026-04-07
api: integrate with %presence agent

latter-bolden:
## Summary

Adds the `urbit`,`api`, and `store`  module layers for integrating with the new `%presence` agent. As is, this is split off from the UI changes to keep size manageable.

Targets `presence-release-1` branch, which we'll use to stage `%presence` code until feature complete.

## Changes

- `urbit/presence` for handling`%presence` agent types
- `api/client/presenceApi` for interacting with the agent itself
- `api/client/computingStatus` for the thinking indicator payload protocol (serialized via a blob-style container field)
- `store/sync` addition of the top level `%presence` firehose subscription
- `store/presence` for the in-memory store for powering UI components
- various generated test files

## How did I test?
Besides the included unit tests, here are the steps to try it locally:

1. Checkout `m/presence` and install it on 2 fake nodes
2. Start and accept a DM between the two nodes
3. Copy the updated `%groups` desk and `|commit` on both nodes
4. On Node A, run the web dev server
   - Mount the subscription in `sync`, same as existing patterns
   - Enable the `presenceApi` logger (or add your own debug prints)
6. Poke `tlon.computing-status.v1` shaped presence events from Node B dojo

## Risks and impact

- Affects important code area: no


packages/api/package.json
```diff
@@ -33,6 +33,7 @@
   "scripts": {
     "build": "tsup",
     "lint": "eslint src/",
+    "lint:format": "prettier ./ --write",
```
dnbrwstr, line 36: 🙏

packages/shared/src/store/sync/sync.ts
```diff
@@ -1885,6 +1886,10 @@ export const syncStart = async (alreadySubscribed?: boolean) => {
   isSyncing = true;
   updateSession({ phase: 'high' });
 
+  if (!alreadySubscribed) {
+    clearPresenceState();
```
dnbrwstr, line 1892: Would appreciate a comment here

packages/shared/src/store/presence.ts
```diff
@@ -0,0 +1,259 @@
...
+
+type PresenceStatusesByKey = Record<string, PresenceStatus>;
+export type PresenceStatusesByContext = Record<string, PresenceStatusesByKey>;
+
+const logger = createDevLogger('presenceStore', false);
+const EMPTY_CONTEXT_STATUSES = Object.freeze({}) as PresenceStatusesByKey;
+let statusesByContext: PresenceStatusesByContext = {};
+const presenceListeners: Array<() => void> = [];
+
+export const reducePresenceState = (
```
dnbrwstr, line 21: A more descriptive name would make this a little more legible, eg `groupPresencesByContext`

dnbrwstr (approved):
Overall looks good to me. I think it'd benefit from a quick pass adding explanatory comments in a few places to make it a little more scannable. In particular I'm thinking about the api types and some of the methods in `store/presence`.

PR #5692, opened 2026-04-08
Aqua tests: lure & others

mikolajpp:
## Summary

This PR implements aqua end-to-end tests for lure. It involves three virtual ships:
1. `~zod` is the group host and the inviter
2. `~bud` is the joiner
3. `~fen` is the bait provider

The only simulated part is the incoming "bite" request, normally issued by hosting. Here we simply inject it as an eyre event. All other interactions are between real agents running on real ships, exactly as happens on livenet.

We also include an unrelated test for a group join, that was ported over from lifetime subscriptions PR.

Lure test is a platform test that involves multiple agents and ships. As such, it is placed in a dedicated `for` (as in platform) directory, since there is no corresponding desk directory that is suitable.`sys` was considered, but it also used for real `sys` tests in the base desk. Suggestions for naming more than welcome.

## Changes

1. Implement test for group join.
2. Implement lure tests for group and personal invite redemption.
3. Fix a poke and mark type mismatch in %reel-describe. It had no real impact because the old type is a superset of the new one. However, given that the client has migrated to new-style lure metadata fields many months ago, we are safe to fix it.
4. Light instrumentation with %info logging on various involved agents. This is extremely useful when developing with aqua, since modifying and syncing desk takes quite some time. Instead of adding debug prints, we can simply enable verbose logging.

## How did I test?

Run the tests.

## Risks and impact

  - [x] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.


desk/tests/ph/for/lure.hoon
```diff
@@ -0,0 +1,267 @@
...
+  (pure:m p.json)
+::
+++  ph-test-lure-link-creation
+  =/  m  (strand ,~)
+  ^-  form:m
+  ::  host a group on ~zod and enable lure links
+  ::
+  ;<  ~  bind:m  create-test-group
+  ;<  ~  bind:m  (poke-app [~zod %grouper] grouper-enable+my-test-group-name)
+  ;<  token=@t  bind:m  (generate-lure-invite lure-group-metadata)
```
Fang-:
```suggestion
  ;<  invite-url=@t  bind:m  (generate-lure-invite lure-group-metadata)
```

desk/tests/ph/for/lure.hoon
```diff
@@ -0,0 +1,267 @@
...
+  ?>  (gth (met 3 token) 0)
```
Fang-:
```suggestion
  (ex-not-equal !>(token) !>(~))
```

Or similar. Even `?<  =('' token)` would be better, but we probably shouldn't be hard-crashing when we're testing/expecting values.

desk/tests/ph/for/lure.hoon
```diff
@@ -0,0 +1,267 @@
...
+::
+++  eyre-authenticate
+  |=  =ship
+  =/  m  (strand (unit @t))
+  ^-  form:m
+  ;<  =bowl:strand  bind:m  get-bowl
+  =/  aqua-pax
+    /i/(scot %p ship)/j/(scot %p ship)/code/(scot %da now.bowl)/(scot %p ship)/noun
+  =+  ;;  code=(unit @t)
+    (scry-aqua:util noun our.bowl now.bowl aqua-pax)
```
Fang-: Arguably the helper should handle the `/i/~ship` prefix. Weird for test code to have to deal with that. Maybe `+scry-aqua-ship:util` or some such is appropriate.

desk/tests/ph/for/lure.hoon
```diff
@@ -0,0 +1,267 @@
...
+  ^-  form:m
+  ::  host a group on ~zod and enable lure links
+  ::
+  ;<  ~  bind:m  create-test-group
+  ;<  ~  bind:m  (poke-app [~zod %grouper] grouper-enable+my-test-group-name)
+  ;<  token=@t  bind:m  (generate-lure-invite lure-group-metadata)
+  ?>  (gth (met 3 token) 0)
+  (pure:m ~)
+::
+++  eyre-authenticate
```
Fang-, line 135: The sheer size of this arm further reinforces my idea that eyre should expose a "session management api" as part of its tasks/gifts. Ideally you want to just tell it "create a session whose token is `0vwhatever`", and then use that. Two-liner.

desk/tests/ph/for/lure.hoon
```diff
@@ -0,0 +1,267 @@
...
+  ::
+  =/  =aqua-event
+    [%event ~fen /e/aqua/eyre/request task]
+  ;<  ~  bind:m  (send-events ~[aqua-event])
+  ::  ~bud receives group invites: one current and one backwards compatible
+  ::
+  ;<  kag=cage  bind:m  (wait-for-app-fact /~bud/groups/v1/foreigns [~bud %groups])
+  ?>  =(%foreigns-1 p.kag)
+  ;<  kag=cage  bind:m  (wait-for-app-fact /~bud/groups/v1/foreigns [~bud %groups])
+  ?>  =(%foreigns-1 p.kag)
```
Fang-, line 228: These look like two of the same marks, not "one current and one backwards compatible".

Fang-: Also here and below, we use `?>` to check instead of `+ex-equal` or w/e testing helper is appropriate. We should probably do the latter and consider, if at all reasonable, to inspect/test the full contents of the fact.

desk/tests/ph/for/lure.hoon
```diff
@@ -0,0 +1,267 @@
...
+    :*  %request
+        secure=|
+        ipv4+.127.0.0.1
+        request
+    ==
+  ::  ~fen the bait provider receives the onboarding request
+  ::
+  =/  =aqua-event
+    [%event ~fen /e/aqua/eyre/request task]
+  ;<  ~  bind:m  (send-events ~[aqua-event])
```
Fang-: You can imagine factoring this whole "request to a specific `lure-invite` url" logic out into a separate helper arm.

Fang- (commented):
Comments below!

desk/tests/ph/for/lure.hoon
mikolajpp: Fixed this by reimplementing `+scry-aqua` to take an additional aqua ship argument and making it monadic to match `+scry` from `lib-strandio`.

desk/tests/ph/for/lure.hoon
mikolajpp: Following our conversation: I think we can assert here, rather than expect, because the path already specifies the mark. At the same time, we should indeed inspect the facts here, because a foreign update is too generic to imply invite has been received.

mikolajpp: Update: this is in fact correct! The marks here are for facts, not pokes. There are two pokes behind the scenes, but here we watch for responses. A backward-compatible invite looks the same, since it is upgraded on ingress.

Fang- (approved):
Looks good overall, thank you for this work!

> add missing /lib/test file

There's probably many more files that we expect to be present thanks to coping in `urbit/urbit`'s `/pkg/base-dev` prior to copying in our `/desk`. I think we should either: 1) only hard-copy/vendor when we make meaningful changes, or 2) vendor _all_ files, so that we never copy in from base-dev desks.

The latter risks files going stale, confusion when merging in updates, and needing to explicate that in the first place. (It does avoid unexpected/uncaught file changes though, but, I think CI might already pull in latest files from those repos for testing anyway?)

Making a decision there is out of scope for this PR, but I would lean towards not adding a /lib/test file for now.

desk/lib/ph/io.hoon
```diff
@@ -432,4 +424,22 @@
...
+      /i/(scot %p ship)
+    pax
+  %-  pure:m
+  ;;  mold
+  .^  noun
+      (scot %p our.bowl)
+      %aqua
+      (scot %da now.bowl)
+      aqua-pax
+  ==
```
Fang-: Does this "just work" in practice? I think doing `.^  mold` might be functionally equivalent.

mikolajpp: I tested `.^  mold` seems to work just as good.

desk/lib/ph/io.hoon
```diff
@@ -432,4 +424,22 @@
...
+::
+++  scry-aqua
+  |*  [=mold =ship pax=path]
+  =/  m  (strand ,mold)
+  ^-  form:m
+  ;<  =bowl:spider  bind:m  get-bowl
+  =/  aqua-pax=path
+    %+  weld
+      /i/(scot %p ship)
+    (snoc `path`pax %noun)
```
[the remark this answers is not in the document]
mikolajpp: Large Lousy Model, ignore.

PR #5695, opened 2026-04-08
fix: improve sigil fallbacks for missing contact avatars

wca4a:
## Summary
- increase the default inner sigil size for the smallest avatar sizes
- use a stable per-ship fallback sigil color when a contact profile has not been synced yet
- reset `ImageAvatar` load state when the image URL changes so stale fallback state does not stick
- key reply-preview avatars by ship instead of array index so React does not reuse the wrong avatar instance

## Testing
- pnpm --filter "@tloncorp/app" tsc
- pnpm --filter "@tloncorp/app" exec vitest run ui/utils/colorUtils.test.ts


arthyn:
@wca4a made some slight adjustments so this would actually work you can see the effects here:

Screenshot 2026-04-16 at 3 37 15 PM
Screenshot 2026-04-16 at 3 33 54 PM

arthyn (approved):
also got manual approval in chat

PR #5696, opened 2026-04-08
fix: group details alignment and member profile actions

jamesacklin:
## Summary

Fixes alignment issues on the desktop group details page (TLON-5548) and makes member list items clickable with full admin actions. Extracts a shared `GroupMemberProfileSheet` component to deduplicate profile sheet logic between the group details preview and the full members screen.

Fixes TLON-5548

## Changes

- Fix connection status row padding to match settings items below it by replacing `ConnectionStatus` component with a custom `ListItem` using consistent `paddingHorizontal`
- Fix settings/notifications block alignment on desktop by adding `$gtSm={{ paddingHorizontal: 0 }}` override on `ActionSheet.ActionGroup` to prevent Tamagui responsive padding
- Fix members count label size from `$label/xl` to `$label/m` to match the description label
- Fix "Leave group" action block alignment with sections above it
- Make member list items clickable, opening a `ProfileSheet` with admin actions (kick, ban, role assignment) matching the full members screen
- Add "(X more)" count to the "Manage members" / "See all" link
- Extract `GroupMemberProfileSheet` component that encapsulates `useGroupContext` admin actions internally, replacing ~40 lines of duplicated prop wiring in both `MembersList` and `GroupMembersScreenView`
- Simplify `GroupMembersScreenView` and `GroupMembersScreen` by removing 6 admin callback props now handled internally by the shared component

## How did I test?

Verified visually in Chrome on desktop layout. Confirmed all alignment fixes, clickable members, and admin action sheet rendering.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [x] Other: Group details / member management UI

## Rollback plan

Revert the commit. No migrations or backend changes involved.

## Screenshots / videos

Public group, not an admin

Screenshot 2026-04-08 at 2 37 26 PM
 
Private group, admin

Screenshot 2026-04-08 at 2 37 10 PM

### DIFF
--- a/packages/app/features/top/chatDetails.tsx
+++ b/packages/app/features/top/chatDetails.tsx
@@ -5,7 +5,7 @@ import * as store from '@tloncorp/shared/store';
 import { capitalize } from 'lodash';
 import React, { useCallback, useMemo, useState } from 'react';
 import { Alert } from 'react-native';
-import { Spinner, isWeb } from 'tamagui';
+import { isWeb } from 'tamagui';
 
 import { useChatSettingsNavigation } from '../../hooks/useChatSettingsNavigation';
 import { useGroupContext } from '../../hooks/useGroupContext';
@@ -32,7 +32,7 @@ import {
   useGroupTitle,
   useIsAdmin,
 } from '../../ui';
-import { getStatusLabels } from '../../ui/components/ConnectionStatus';
+import { ConnectionIndicatorAction } from '../../ui/components/ConnectionStatus';
 import { useShipConnectionStatus } from './useShipConnectionStatus';
 
 // Utility functions
@@ -333,14 +333,6 @@ export function SettingsSection({
     enabled: entityType === 'group' && !!group,
   });
 
-  const connectionLabels = useMemo(
-    () =>
-      entityType === 'group' && group
-        ? getStatusLabels(connectionStatus.status)
-        : null,
-    [entityType, group, connectionStatus.status]
-  );
-
   const volumeSettings =
     entityType === 'group' ? group?.volumeSettings : channel?.volumeSettings;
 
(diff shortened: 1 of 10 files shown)


packages/app/features/top/chatDetails.tsx
```diff
@@ -422,14 +463,38 @@ export function SettingsSection({
...
         contentProps={{
           backgroundColor: '$background',
           borderRadius: '$2xl',
           borderWidth: 0,
         }}
       >
-        {entityType === 'group' && group && (
-          <ConnectionStatus contactId={group.hostUserId} type="action" />
+        {entityType === 'group' && group && connectionLabels && (
+          <ListItem
```
dnbrwstr: Curious if there's a way to do this with `ConnectionIndicatorAction` -- that looked like a much cleaner implementation and is now dead code.

dnbrwstr (changes requested):
One nit and one blocker. Nit is inline, blocker is that sheet is busted on mobile (probably just needs modal=true).

https://github.com/user-attachments/assets/06cf6157-cfe1-4766-90d2-e6db7f1a3005

packages/app/features/top/chatDetails.tsx
jamesacklin: Good call — replaced the manual ListItem/ActionSheet markup with `ConnectionIndicatorAction`, which encapsulates the same logic (getStatusLabels, icon-vs-spinner, title/subtitle). Also removed the now-unused `connectionLabels` useMemo and `Spinner` import. Done in 84220c0.

jamesacklin:
@dnbrwstr I have done as you requested. I had to add a BottomSheetProvider to the app root so we can safely use the user data context with `modal=true`, otherwise didn't touch the view.

image

dnbrwstr (approved):
Lgtm

PR #5697, opened 2026-04-08
presence: short-lived personal activity indication

Fang-:
## Summary

Introduces `/app/presence`, which tracks and publishes short-lived personal activity indicators across different contexts and topics. Think "typing..." or "thinking..." indicators.  
Originated from @arthyn's [specification](https://gist.github.com/arthyn/d790700b0da75721dca7934fd2db494f), though it's deviated a bit from that.

## Architecture

A `context` is a `path` indicating a "place" within groups, most commonly a `/channel/[kind]/[ship]/[name]` or a `/dm/[ship]`.  
A `topic` is a kind of thing you can be doing in a context, like `%typing` or `%computing`.  
A presence indication comes with `timing` and `display` attributes, affecting expiry and rendering respectively.

To indicate your presence, you send it to the context's host. The host gives a subscription update to subscribers. You can optionally tell the host to only disclose the update to a select set of ships.

Because they're symmetrical rather than centralized, we must take care around DMs. The client will describe the context as `/dm/[her]` to the local agent, but when talking to `her`, we must describe the context as `/dm/[our]`, because that's what the context is from `her` perspective.

We lean into "setup and inflate" patterns for subscription management, eagerly inflating during initialization, reload, and whenever we know _something_ changed but can't be sure exactly what. Only if we know a precise change do we manage the subscription in-place.

## Status

Draft because of unresolved review questions and some remaining TODOs. Already functional and integrated on 
@latter-bolden's branches.

## RFC

See below.


desk/app/presence.hoon
```diff
@@ -0,0 +1,475 @@
...
+    [%give %fact ~ %presence-response-1 !>(`response-1`[%init places])]~
+  ::
+      [%context @ *]
+    ::  context watch paths must be properly personalized
+    ::
+    ?>  =(src.bowl (slav %p i.t.path))
+    =.  subs  (~(put ju subs) t.t.path src.bowl)
+    ::NOTE  no initial fact, since all data is short-lived,        ::REVIEW
+    ::      and we don't want to hot-loop on mark incompatibility  ::REVIEW
+    [~ this]
```
Fang-: All data _is_ short-lived, right?

desk/app/presence.hoon
```diff
@@ -0,0 +1,475 @@
...
+    ::
+    %+  murn
+      ^-  (list [wire dude:gall path])
+      :~  [/chat/invited %chat /dm/invited]  ::REVIEW  not firing as often as expected..?
+          [/channels/all %channels /v3]  ::REVIEW  can we watch anything "smaller"?
+      ==
+    |=  [=wire =dude:gall =path]
+    ^-  (unit card)
+    ?:  (~(has by wex.bowl) wire our.bowl dude)  ~
+    `[%pass wire %agent [our.bowl dude] %watch path]
```
Fang-, line 223: I would love for chat to have a subscription that notifies whenever `+accepted-dms` changes. Right now this only listens to changes to `+pending-dms`, which is close but doesn't fire _quite_ as often as you'd want.

For channels, I would love to not listen to the firehose. As is, we process _every response_ that channels agent sends, which is wasteful of compute if nothing else.

desk/lib/presence-json.hoon
```diff
@@ -0,0 +1,86 @@
...
+++  dejs
+  =,  dejs:format
+  |%
+  ++  key      (ot 'context'^pa 'ship'^(se %p) 'topic'^topic ~)
+  ++  topic    (su (perk %typing %computing %other ~))
+  ++  display  (ot 'symbol'^(mu so) 'text'^(mu so) 'blob'^(mu so) ~)
+  ::
+  ++  action-1
+    ^-  $-(json ^action-1)
+    ::REVIEW  could flatten, could take 'type'
```
Fang-, line 15: @latter-bolden You've integrated with this fine it seems like. But just letting you know that we _could_ flatten the json objects down if you think that'd be more ergonomic.

desk/sur/presence.hoon
```diff
@@ -0,0 +1,35 @@
+::  presence: short-lived personal activity types
+::
+|%
++$  places    (map context topics)
++$  topics    (map topic people)
++$  people    (map ship [timing display])
+::
++$  context   path  ::  nest path or other
++$  topic     ?(%typing %computing %other)  ::TODO  @ ?
```
Fang-, line 9: On the one hand, we don't want to change the protocol for new topics. On the other hand, we don't want arbitrary strings to go in here...

desk/app/presence.hoon
```diff
@@ -0,0 +1,475 @@
...
+::    context hosts do not give initial facts to subscribers. timing out of
+::    presences does not get propagated to subscribers, but explicit %clear
+::    does. (the latter goes out to _all_ subscribers, regardless of the
+::    disclosure set of the prior presence.)
+::
+::    we do not solve for clock skew, instead respecting the received
+::    timestamps and timeouts as-is, and ignoring any that have expired
+::    by the time we receive them. if it's in the future, treat it as now.
+::
+::TODO  make chat, channels clear %typing whenever we receive a msg?
```
Fang-, line 31: This is somewhat questionable, I think it's a difficult trade-off: either you spend a bunch of compute always wiping presence for specific topics on-receive (and assume the sender will re-set it if they're composing a new msg or w/e), or you risk a typing indicator lingering briefly after someone has finished authoring & sending a message.

Or could we get away with doing such behavior client-side only?

desk/app/presence.hoon
```diff
@@ -0,0 +1,475 @@
...
+::      the dm case).
+::      wires only contain the context, because src.bowl indicates the ship.
++$  state-0
+  $:  %0
+      =places
+      :: ships=(jug ship context)
+    ::
+      want=(set [ship context])  ::  desired outgoing subs
+      subs=(jug context ship)    ::  real incoming subs
+      ::TODO  wack=(set ship)    ::  waiting for acks (load prevention)
```
Fang-, line 49: Once we start sending presence during normal usage in the client (ie as typing indicators) we'll want something like this to keep us from queueing up large streams of pending pokes for ships that have never acked any of them.

Though perhaps this case won't be that big of an issue in practice: you're not _that_ likely to keep returning to & typing into contexts that are offline.

desk/app/presence.hoon
```diff
@@ -0,0 +1,475 @@
...
+          %+  add  since.timing.upd
+          (fall timeout.timing.upd (default-timeout topic.key.upd))
+        ?:  (gth now.bowl end)
+          ::TODO  maybe delete existing one at key?
+          [~ this]
+        ::TODO  should we be adjusting the context.key.upd to the local
+        ::      perspective? or does sender take care of that?
+        ::      dm context paths hard to track...
+        :_  this(places (put-presence places +.upd))
+        :~  (give-response %here +.upd)
```
Fang-: I think this is fine given that @latter-bolden's integration produces working demos, but might be good to have a test or two for this as a sanity-check.

desk/sur/presence.hoon
```diff
@@ -0,0 +1,35 @@
...
++$  key       [=context =ship =topic]
```
arthyn, line 10: could see an argument for our author type from chat/channels here, but lets leave it ship for now

desk/lib/presence-json.hoon
```diff
@@ -0,0 +1,86 @@
...
+::
+/-  *presence
+::
+|%
+++  dejs
+  =,  dejs:format
+  |%
+  ++  key      (ot 'context'^pa 'ship'^(se %p) 'topic'^topic ~)
+  ++  topic    (su (perk %typing %computing %other ~))
+  ++  display  (ot 'symbol'^(mu so) 'text'^(mu so) 'blob'^(mu so) ~)
```
arthyn: probably should match the output

```suggestion
  ++  display  (ot 'icon'^(mu so) 'text'^(mu so) 'blob'^(mu so) ~)
```

desk/lib/presence-json.hoon
```diff
@@ -0,0 +1,86 @@
...
+      :-  %o
+      %-  ~(rep by people)
+      |=  [[ship=@p t=^timing d=^display] o=(map @t json)]
+      %+  ~(put by o)  (scot %p ship)
+      ::REVIEW  could flatten
+      (pairs 'timing'^(timing t) 'display'^(display d) ~)
+    ::
+        %here
+      %-  pairs
+      ::REVIEW  could flatten
```
arthyn, line 77: good as is imo

desk/lib/presence-json.hoon
```diff
@@ -0,0 +1,86 @@
...
+        :-  'blob'  ?~(blob ~ s+u.blob)
+    ==
+  ::
+  ++  response-1
+    |=  res=^response-1
+    %+  frond  -.res
+    ?-  -.res
+        %init
+      :-  %o
+      ::REVIEW  could flatten
```
arthyn, line 60: I could see the argument here for flattening, like an array of what the %here case below produces `{ key, timing, display }`

desk/app/presence.hoon
```diff
@@ -0,0 +1,475 @@
...
+  |=  [=mark =vase]
+  ^-  (quip card _this)
+  ~|  mark=mark
+  ?>  =(src our):bowl
+  ?+  mark  !!
+      %presence-action-1
+    =+  !<(act=action-1 vase)
+    ?:  ?=(%nuke -.act)
+      ::TODO  response?
+      [~ this(places (~(del by places) context))]
```
arthyn: is context coming from act here even though its faced, because of the =+?

desk/app/presence.hoon
```diff
@@ -0,0 +1,475 @@
...
+          [~ this]
+        ::  setup will scry out relevant dms
+        ::
+        [[(await-setup now.bowl ~)]~ this]
+      ::
+          [%channels %all ~]
+        ?.  ?=(%channel-response-4 p.cage.sign)
+          ::TODO  log formally
+          [~ this]
+        ::TODO  would love to do a smaller sub, so we don't end up doing these
```
arthyn, line 422: technically you could watch activity and catch both dms and channels but idk if that's much better.

arthyn (approved):
lgtm my only complaint is that contexts are not actually permissioned, but that's fine for now imo.

wca4a:
@codex Automated review: two high-confidence bugs.

1. `desk/app/presence.hoon`: the read/write paths are still effectively unpermissioned.

`+on-watch` accepts any personalized `[%context @ *]` watch as long as the caller encoded its own ship in the path, and then adds that ship to `subs`. `+on-poke` for `%presence-command-1` only enforces `ship.key == src` (with a DM special-case), but it never verifies that the sender is actually a participant in the channel/group context. In practice this means a ship that can guess a context path can subscribe to someone else's DM/channel presence or inject its own bogus presence into that context. This looks like a concrete privacy/integrity bug, not just a missing refinement.

2. `desk/app/presence.hoon`: incoming DM updates stay in the publisher's perspective instead of being translated back to the local one.

`+dm-contexts` intentionally subscribes to a peer using `/dm/[our]`, because that is the context from the peer's perspective. But in `+on-agent` when a `%presence-update-1` fact arrives, the code stores `+.upd` unchanged via `put-presence` and emits `%here`/`%gone` with that same key. The local client talks to this agent using `/dm/[peer]`, so remote DM presence winds up keyed under `/dm/[our]` instead of the local DM path. That means DM lookups / clears / expiry notifications won't line up for incoming presence.

desk/app/presence.hoon
Fang-: No, wth, how does this compile. Good catch, probably bugged, needs to be `context.act`.

desk/app/presence.hoon
```diff
@@ -390,6 +436,11 @@
         ::TODO  log formally
         [~ this]
       =+  !<(upd=update-1 q.cage.sign)
+      ::  translate dm context from peer's perspective to ours
+      ::
+      =/  =key  ?-(-.upd %set key.upd, %clear key.upd)
+      =?  context.key  ?=([%dm *] context.key)
+        /dm/(scot %p src.bowl)
```
Fang-: As discussed:
- we should check `context.key` matches `context` aka `t.wire`,
- in the dm case, here, we make an exception, but we should still test against `/dm/(scot %p our.bowl)` specifically, rather than any dm-shaped context.

Fang-: Also this whole thing could/should be an `=.` edit of `upd`, so that we don't have to remember to carry about a modified `key` and can keep using the `+.upd` wholesale, letting you get away with not doing the changes below.

desk/app/presence.hoon
```diff
@@ -0,0 +1,523 @@
...
+    [%channel kind=@ ship=@ name=@ ~]  (slav %p i.t.t.context)
+    [%group ship=@ term=@ ~]           (slav %p i.t.context)
+  ==
+::
+++  is-participant
+  |=  [=context =bowl:gall]
+  ^-  ?
+  ?+  context  &
+      [%dm @ ~]
+    =(src.bowl (slav %p i.t.context))
```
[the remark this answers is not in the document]
arthyn: I think this is fine as-is for now.

desk/app/presence.hoon
```diff
@@ -0,0 +1,523 @@
...
+      =/  =key  ?-(-.upd %set key.upd, %clear key.upd)
+      =?  context.key  ?=([%dm *] context.key)
+        /dm/(scot %p src.bowl)
+      ?-  -.upd
+          %set
+        =?  since.timing.upd  (gth since.timing.upd now.bowl)
+          now.bowl
+        =/  end=@da
+          %+  add  since.timing.upd
+          (fall timeout.timing.upd (default-timeout topic.key.upd))
```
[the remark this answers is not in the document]
arthyn: fixed

latter-bolden (approved):

PR #5700, opened 2026-04-09
gateway-status: refactor & follow-up review

mikolajpp:
## Summary

Refactor and follow-up review of the gateway-status agent.
There are few architectural questions and problems posted below in a comment separately.

## Changes

1. Conformance to established Hoon coding style.
2. Various comments about outdated/dangerous practices that require more work.

## How did I test?

Run the supplied unit tests. Unit tests were not reviewed. 

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.


mikolajpp:
Here are the questions that were not addressed in this PR, but should nonetheless be resolved.
1. Slog should not be used. We have a logging library.
2. There is a potential resubscription loop on the activity subscription. 
3. However, we should not actually use the activity endpoint in this case. This agent only cares about messages coming from a specific ship, but is getting activated on any event receiving by activity, which includes anything happening throughout entire system. We should rather subscribe on a per-owner dm subscription to %chat.
4. The owner parameter is never unset. Therefore, the `has-owner` guard used on gateway agent operations is only really meaningful on the first ever interaction with the gateway service; after that it is spurious and could potentially mask bugs. %configure should either be collapsed into %gateway-start if there is no good reason against. Alternatively, %gateway-start should carry the owner instead. Either of these approaches would enable owner to be cleared and the guard to be meaningful.
5. The application does not actually make use of the following variables: `last-heartbeat`, `last-stop`. Does it intend to? No mention of that in the "spec" for this tiny agent.
6. Why do we ignore the case when the owner is the bot itself in `+handle-activity-add`? We either want to support this, or else this is a configuration error which should crash and be reported.
7. I don't understand the supposed deduplication tied to `last-auto-reply-to`. Under what circumstances we could receive two events with the same key? Each DM message has a unique key. This is potentially another useless variable in the state.

arthyn (approved):
lgtm

PR #5703, opened 2026-04-09
Fix inline chat image appearing larger than expected

jamesacklin:
## Summary

Reverts the `ContentImage` styled default from `width: '100%'` back to `maxWidth: '100%'`. A prior change (127fc75f) removed the max-width cap, causing images on mobile to render at their full natural pixel dimensions and overflow the chat container.

Fixes TLON-5577

## Changes

- Changed `ContentImage` styled component default from `width: '100%'` to `maxWidth: '100%'` in `BlockRenderer.tsx`

## How did I test?

Code review and tracing the rendering path for both constrained (web chat) and unconstrained (mobile) cases. The constrained path sets explicit inline `width`/`height` props that override the styled default, so it is unaffected.

## Risks and impact

  - [x] Channel display
  - [x] Other: Chat message image rendering

## Rollback plan

Revert the single-line change in `ContentImage` styled component back to `width: '100%'`.

## Screenshots / videos

Screenshot 2026-04-10 at 11 22 46 AM

Screenshot 2026-04-10 at 11 27 21 AM

### DIFF
--- a/packages/app/ui/components/PostContent/BlockRenderer.tsx
+++ b/packages/app/ui/components/PostContent/BlockRenderer.tsx
@@ -585,7 +585,7 @@ export function ImageBlock({
 const ContentImage = styled(Image, {
   name: 'ContentImage',
   context: ContentContext,
-  width: '100%',
+  maxWidth: '100%',
 });
 
 export function RuleBlock({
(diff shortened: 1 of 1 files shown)


dnbrwstr (approved):
Lgtm

PR #5704, opened 2026-04-10
fix thread unreads while viewing threads

patosullivan:
## Summary

Fixes TLON-5336: thread unreads now clear in real time while the user is already viewing the thread, matching main-channel behavior. Previously, `useMarkThreadAsReadEffect` only re-fired on the initial `shouldMarkRead` transition, so replies that arrived while a user was inside a chat or DM thread left a stale unread.

## Changes

- Added `useLiveThreadUnreadByParentId` in `packages/shared/src/store/dbHooks.ts`, a `useQuery`-backed hook that reactively tracks a thread's unread state keyed by `parentPostId`, using the same table-level invalidation pattern as the adjacent `useLiveThreadUnread`.
- Updated `SinglePostView` in `packages/app/ui/components/PostScreenView.tsx` to consume the new hook and derive a `hasThreadUnreads` boolean that is passed into `useMarkThreadAsReadEffect`.
- Rewrote `useMarkThreadAsReadEffect` in the same file as a single guarded effect depending on `[shouldMarkRead, hasThreadUnreads, latestReplyId, store]`. The effect guards on `!shouldMarkRead || !hasThreadUnreads`, reads current opts through an `optsRef` for stable identity, and schedules `markThreadRead` after 150ms with a cleanup. This mirrors the channel unread-trigger pattern in `Channel/index.tsx`.
- Added `testID="ThreadUnreadDot"` to the `UnreadDot` rendered by `ThreadUnreadDot` in `packages/app/ui/components/ChatMessage/ChatMessageReplySummary.tsx` for E2E selector stability.
- Added regression coverage:
  - `apps/tlon-web/e2e/thread-unread-clearing.spec.ts` for the chat thread scenario.
  - `apps/tlon-web/e2e/dm-thread-unread-clearing.spec.ts` for the DM thread scenario.

  Both specs exercise the exact failing scenario: ~zod opens a thread, ~ten sends a reply while ~zod is viewing the thread, ~zod navigates back, and the parent message's `ThreadUnreadDot` is asserted absent.

## How did I test?

- Local E2E:
  - `thread-unread-clearing.spec.ts` passes.
  - `dm-thread-unread-clearing.spec.ts` passes.
  - `dm-thread-interactions.spec.ts` passes.
  - `thread-cross-ship.spec.ts` passes.
- Manually tested locally on web

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [  ] Message sync
  - [x] Channel display
  - [ ] Notifications

## Rollback plan

Revert.

### DIFF
--- a/packages/shared/src/store/dbHooks.ts
+++ b/packages/shared/src/store/dbHooks.ts
@@ -248,6 +248,30 @@ export const useLiveThreadUnread = (unread: db.ThreadUnreadState | null) => {
   });
 };
 
+/**
+ * Reactively tracks the unread state for a thread, keyed by the parent post's
+ * ID. Unlike `useLiveThreadUnread` (which requires a `ThreadUnreadState` seed
+ * and goes dormant when the seed is null), this hook always tracks reactively
+ * as long as `parentPostId` is non-null — even when the thread starts with no
+ * unread row.
+ */
+export const useLiveThreadUnreadByParentId = (parentPostId: string | null) => {
+  const depsKey = useMemo(
+    () => (parentPostId ? keyFromQueryDeps(db.getThreadUnreadState) : null),
+    [parentPostId]
+  );
+
+  return useQuery({
+    queryKey: ['liveUnreadCount', depsKey, 'thread', parentPostId],
+    queryFn: async () => {
+      if (parentPostId) {
+        return db.getThreadUnreadState({ parentId: parentPostId });
+      }
+      return null;
+    },
+  });
+};
+
 export const useLiveChannelUnread = (unread: db.ChannelUnread | null) => {
   const depsKey = useMemo(
     () => (unread ? keyFromQueryDeps(db.getChannelUnread) : null),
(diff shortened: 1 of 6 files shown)


packages/shared/src/store/dbHooks.ts
```diff
@@ -248,6 +248,30 @@ export const useLiveThreadUnread = (unread: db.ThreadUnreadState | null) => {
   });
 };
 
+/**
```
arthyn, line 251: I wonder if we should just combine them?

arthyn (approved):
🙏🙏🙏🙏

packages/shared/src/store/dbHooks.ts
patosullivan: I feel like we should probably just keep them separate for now. The old hook is built around already having an unread row and the new one needs to keep watching by parent post id even when we don't have an unread row yet.

PR #5713, opened 2026-04-13
negotiate: add reset poke, re-announce on reload, debug peeks

arthyn:
Addresses a class of stuck-version bugs where a subscriber's `heed` entry gets pinned at an old version with no recovery path, even when both sides' wex/sup agree the subscription is live. Observed with malmur/somfyl stuck at groups v1 after somfyl upgraded to v2, likely due to a dropped fact during a post-breach Ames flow reset window.

Three pieces:

1. `%negotiate-reset` poke (operator-invoked repair). Accepts `(unit gill:gall)`. For each matching heed key, emits `%leave` on the heed subscription wire, clears the entry, and calls `+negotiate` to re-subscribe. Forces the publisher to re-answer via `+on-watch` with its current version, healing stuck state immediately without waiting for a reload.

2. `+ours-changed` now always emits a `%fact` for every protocol in current `ours`, and the `+on-load` state-1 guard that gated it on `!=(ours our-versions)` has been removed. Every reload acts as a safety-net re-announce to all existing subscribers. `+heed-changed` short-circuits on same-version facts, so healthy subs pay nothing.

3. Debug peeks on the outer agent door: `/x/dbug/wex`, `/x/dbug/sup`, `/x/dbug/bowl`, `/x/dbug/orphans`. These bypass `+inner-boat` / `+inner-bitt` so operators can see library-internal wires that would otherwise be filtered from any `+dbug`-based scry.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

### DIFF
--- a/desk/lib/negotiate.hoon
+++ b/desk/lib/negotiate.hoon
@@ -345,23 +345,22 @@
     ++  reset-heed
       |=  target=(unit gill:gall)
       ^-  (quip card _state)
-      =/  keys=(list [gill:gall protocol])
+      =/  keys=(list [=gill:gall =protocol])
+        ?~  target  ~(tap in ~(key by heed))
         %+  skim  ~(tap in ~(key by heed))
-        |=  [=gill:gall *]
-        ?~  target  &
-        =(gill u.target)
+        |=([=gill:gall *] =(gill u.target))
       =|  out=(list card)
       |-
       ?~  keys  [out state]
-      =*  entry  i.keys
-      =/  =wire
-        /~/negotiate/heed/(scot %p p.entry)/[q.entry]/[+.entry]
-      =.  out  (snoc out [%pass wire %agent -.entry %leave ~])
       =.  out
-        %+  snoc  out
-        (tell:log %dbug ~[>[%reset-heed entry]<] ~)
-      =.  heed  (~(del by heed) entry)
-      =^  caz  state  (negotiate entry)
+        :+  (tell:log %dbug ~[>[%reset-heed i.keys]<] ~)
+          =,  i.keys
+          :+  %pass
+            /~/negotiate/heed/(scot %p p.gill)/[q.gill]/[protocol]
+          [%agent gill %leave ~]
+        out
+      =.  heed  (~(del by heed) i.keys)
+      =^  caz  state  (negotiate i.keys)
       $(out (weld out caz), keys t.keys)
     ::
     ++  ours-changed
@@ -749,28 +748,29 @@
         =/  dat  (on-peek:og path)
         ?:  ?=(?(~ [~ ~]) dat)  ~
         (fall ((soft (list mass)) q.q.u.u.dat) ~)
-      ?:  =(/x/dbug/state path)
-        ``noun+!>((slop on-save:og !>(negotiate=state)))
-      ?:  =(/x/dbug/wex path)
-        ``noun+!>(wex.bowl)
-      ?:  =(/x/dbug/sup path)
-        ``noun+!>(sup.bowl)
-      ?:  =(/x/dbug/bowl path)
-        ``noun+!>(bowl)
-      ::  orphan diff: heed keys whose /~/negotiate/heed/... subscription
-      ::  is missing from wex.bowl. non-empty result means +inflate's
-      ::  orphan detector should have fired but didn't (or hasn't yet).
-      ::
-      ?:  =(/x/dbug/orphans path)
-        :^  ~  ~  %noun
-        !>  ^-  (list [gill:gall protocol])
-        %+  murn  ~(tap by heed)
-        |=  [[=gill:gall =protocol] *]
-        =/  =wire
-          :+  %~.~  %negotiate
-          [%heed (scot %p p.gill) q.gill protocol ~]
-        ?:  (~(has by wex.bowl) [wire gill])  ~
-        `[gill protocol]
+      ?:  ?=([%x %dbug *] path)
+        ?+  t.t.path  [~ ~]
+            [%wex ~]   ``noun+!>(wex.bowl)
+            [%sup ~]   ``noun+!>(sup.bowl)
+            [%bowl ~]  ``noun+!>(bowl)
+            [%state ~]
+          ``noun+!>((slop on-save:og !>(negotiate=state)))
+        ::
+        ::  orphan diff: heed keys whose /~/negotiate/heed/... subscription
+        ::  is missing from wex.bowl. non-empty result means +inflate's
+        ::  orphan detector should have fired but didn't (or hasn't yet).
+        ::
+            [%orphans ~]
+          :^  ~  ~  %noun
+          !>  ^-  (list [gill:gall protocol])
+          %+  murn  ~(tap by heed)
+          |=  [[=gill:gall =protocol] *]
+          =/  =wire
+            :+  %~.~  %negotiate
+            [%heed (scot %p p.gill) q.gill protocol ~]
+          ?:  (~(has by wex.bowl) [wire gill])  ~
+          `[gill protocol]
+        ==
       ?.  ?=([@ %~.~ %negotiate *] path)
         (on-peek:og path)
       !:
--- a/desk/lib/negotiate.hoon
+++ b/desk/lib/negotiate.hoon
@@ -338,14 +338,10 @@
       ?<  (~(has by heed) for)
       :-  [(watch-version for)]~
       state(heed (~(put by heed) for ~))
-    ::  +reset-heed: operator-invoked repair for stuck heed entries
+    ::  +reset-heed: renegotiate targeted heed entries
     ::
     ::    for each matching heed key, emit %leave on the heed subscription
-    ::    wire, clear the entry, and call +negotiate to re-subscribe. this
-    ::    forces the publisher to re-answer via +on-watch with its current
-    ::    version, healing zombie/desynced subscription state that the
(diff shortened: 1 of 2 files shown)


desk/lib/negotiate.hoon
```diff
@@ -824,6 +885,14 @@
     ++  on-poke
       |=  [=mark =vase]
       ^-  (quip card _this)
+      ::  library-level repair poke: clear targeted heed entries and
+      ::  re-negotiate, healing stuck subscriptions without waiting
+      ::  for the next reload.
+      ::
```
mikolajpp: Code here is clear enough, comment at `+reset-heed` is enough.
```suggestion
```

desk/lib/negotiate.hoon
```diff
@@ -726,6 +763,30 @@
...
+      ?:  =(/x/dbug/orphans path)
+        :^  ~  ~  %noun
+        !>  ^-  (list [gill:gall protocol])
+        %+  murn  ~(tap by heed)
+        |=  [[=gill:gall =protocol] *]
+        =/  =wire
+          :+  %~.~  %negotiate
+          [%heed (scot %p p.gill) q.gill protocol ~]
+        ?:  (~(has by wex.bowl) [wire gill])  ~
+        `[gill protocol]
```
mikolajpp: Has this actually turn up any positives, or did it just confirm that there are no orphans? We can keep it for now just in case, but we should probably remove it in the future, as this problem seems so far to be only a theoretical possibility, where gall would just delete subscription without a corresponding kick.

CC @Fang-

desk/lib/negotiate.hoon
```diff
@@ -581,9 +614,13 @@
...
-          =/  caz1
-            ?:  =(ours our-versions)  ~
-            (ours-changed:up ours our-versions)
+          ::  always re-announce current versions on reload as a safety
+          ::  net. +ours-changed handles both the kick path (for removed
+          ::  protocols) and the fact emission (for current protocols);
+          ::  facts are no-ops on subscribers whose heed is already
+          ::  in sync via +heed-changed's early return.
+          ::
+          =/  caz1  (ours-changed:up ours our-versions)
```
mikolajpp: I would leave this change out while we investigate the problem. We don't want to mask deep system bugs by doing recovery preemptively. Overall there are very few ships suffering from this problem and these are our only chance at fixing such problems at the root.

desk/lib/negotiate.hoon
```diff
@@ -726,6 +763,30 @@
         (fall ((soft (list mass)) q.q.u.u.dat) ~)
       ?:  =(/x/dbug/state path)
         ``noun+!>((slop on-save:og !>(negotiate=state)))
+      ::  raw outer-bowl peeks: these bypass +inner-boat/+inner-bitt so
+      ::  operators can see library-internal wires that would otherwise
+      ::  be filtered out of any inner-agent (e.g. +dbug) scry.
+      ::
```
mikolajpp: Code here is clear enough.
```suggestion
```

desk/lib/negotiate.hoon
```diff
@@ -350,15 +380,18 @@
...
-      %+  murn  ~(tap by neu)
+      %+  turn  ~(tap by neu)
       |=  [=protocol =version]
-      ^-  (unit (list card))
-      ?:  =(`version (~(get by ole) protocol))  ~
-      %-  some
+      ^-  (list card)
       :~  (tell:log %dbug ~[>[%ours-changed 'giving version' version 'for protocol' protocol]<] ~)
           [%give %fact [/~/negotiate/version/[protocol]]~ %noun !>(version)]
       ==
```
mikolajpp, line 388: Likewise, I would leave this change out (see comment below). Let's focus this PR on manual recovery of known ships and instrumentation.

desk/lib/negotiate.hoon
```diff
@@ -338,6 +338,36 @@
...
       state(heed (~(put by heed) for ~))
+    ::  +reset-heed: operator-invoked repair for stuck heed entries
+    ::
+    ::    for each matching heed key, emit %leave on the heed subscription
+    ::    wire, clear the entry, and call +negotiate to re-subscribe. this
+    ::    forces the publisher to re-answer via +on-watch with its current
+    ::    version, healing zombie/desynced subscription state that the
+    ::    orphan detector cannot see (both sides' wex/sup agree, but the
+    ::    version we heard is stale).
+    ::
```
mikolajpp: The comment here is too verbose and the wording a bit uncanny, I assume this was at least partly llm-generated.

```suggestion
    ::  +reset-heed: renegotiate targeted heed entries
    ::
    ::    for each matching heed key, emit %leave on the heed subscription
    ::    wire, clear the entry, and call +negotiate to re-subscribe.
    ::
```

mikolajpp (changes requested):
Some styling nits. The only major item is a request to remove preventive measures while we investigate this problem. We should not mask the instances where this bug has occurred until we find the root cause, but should rather use manual recovery pathway this PR implements.

Also, something isn't right with tests. We should fix it before deploying.

desk/lib/negotiate.hoon
arthyn: no this didn't end up turning up any positives, this was a test to verify #5709 we can keep it or take it out w/e y'all think

arthyn:
@mikolajpp I made one final change to hoist the state type so that its easier to scry the state out

mikolajpp (approved):
Looks good to go!

desk/lib/negotiate.hoon
```diff
@@ -338,6 +337,32 @@
...
+      =*  entry  i.keys
+      =/  =wire
+        /~/negotiate/heed/(scot %p p.entry)/[q.entry]/[+.entry]
+      =.  out  (snoc out [%pass wire %agent -.entry %leave ~])
+      =.  out
+        %+  snoc  out
+        (tell:log %dbug ~[>[%reset-heed entry]<] ~)
+      =.  heed  (~(del by heed) entry)
+      =^  caz  state  (negotiate entry)
+      $(out (weld out caz), keys t.keys)
```
Fang-:
```suggestion
      =/  keys=(list [=gill:gall =protocol])
        ?~  target  ~(tap in ~(key by heed))
        %+  skim  ~(tap in ~(key by heed))
        |=([=gill:gall *] =(gill u.target))
      =|  out=(list card)
      |-
      ?~  keys  [out state]
      =.  out
        :+  (tell:log %dbug ~[>[%reset-heed i.keys]<] ~)
          :+  %pass
            =,  i.keys
            /~/negotiate/heed/(scot %p p.gill)/[q.gill]/[protocol]
          [%agent gill %leave ~]
        out
      =.  heed  (~(del by heed) i.keys)
      =^  caz  state  (negotiate i.keys)
      $(out (weld out caz), keys t.keys)
```

Changes:
- don't walk the whole list if we want all entries.
- `entry` is a horrible name and we don't want to access its components by lark, so we put faces, drop the meaningless alias and pull the "entry" into the subject only during wire construction for convenience/brevity.
- don't snoc, we don't care (that deeply) about move order here, we can just cons the cards together.

desk/lib/negotiate.hoon
```diff
@@ -726,6 +751,26 @@
         (fall ((soft (list mass)) q.q.u.u.dat) ~)
       ?:  =(/x/dbug/state path)
         ``noun+!>((slop on-save:og !>(negotiate=state)))
+      ?:  =(/x/dbug/wex path)
+        ``noun+!>(wex.bowl)
+      ?:  =(/x/dbug/sup path)
+        ``noun+!>(sup.bowl)
+      ?:  =(/x/dbug/bowl path)
+        ``noun+!>(bowl)
```
Fang-: We really should just be pattern matching the path at this point. If nothing else, these can all be one-liners. Arguably should be `/x/dbug/bowl/wex` and `/x/dbug/bowl/sup` because they're data inside/under/from the bowl.

(I think I wrote a style.md correction for this in the reference too.)

But: I think I have an idea for letting us get this information out of dbug at both levels (wrapper _and_ inner agent) without needing to explicate these endpoints in the wrapper. Will try and test that soon.

desk/lib/negotiate.hoon
Fang-: There's a theoretical _very long_ dojo one-liner that runs this check lol. I'm ambivalent but mostly leaning towards not keeping it in.

desk/mar/negotiate/reset.hoon
```diff
@@ -0,0 +1,17 @@
...
+++  grad  %noun
+++  grow
+  |%
+  ++  noun  target
+  --
+++  grab
+  |%
+  ++  noun  (unit gill:gall)
+  --
+--
```
Fang-, line 17: `:my-agent &some-mark data` works even if `/mar/some-mark` doesn't exist, so we don't strictly need this mark file. Toss-up on whether we keep it or not.

(Some argument here that having the mark is more future-proof wrt a version of gall that _does_ always consistently, even locally, validate data against the marks. ...And also future-proof for hypothetical aqua tests, /lib/guardian, etc.)

Fang- (commented):
So the only change here, in the end, is poke/scry debug tooling. This looks more or less fine. Some nits about code/algo structure, with the continuing note that claude writes rather unsatisfying hoon.

Fang- (approved):
This is fine.

PR #5718, opened 2026-04-15
activity: add reactions activity

mikolajpp:
## Summary

This PR implements backend support for react notifications in chat and channels. The logic changes are localized to specific portions in activity, chat and channels agents. We attempt to pattern the logic closely on corresponding code for posts and replies. Thus, for instance, as a reply to a post clears up unreads and notifications in a thread, so do reacts. 

The remaining changes mostly concern with further refactoring of activity (such as JSON activity conversions), and type/endpoint updates.

## Changes

1. Implement activity notifications on reactions in chat and channels. 
2. Introduce new activity endpoints: subscriptions and scries
3. Update groups-ui for new endpoints
4. Various type-versioning related refactoring in activity

## How did I test?

Given that the client portion of it is stale, testing integration with the client was not possible.
I have enabled debug mode in activity and observed relevant events being generated on reaction events.

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [x] Message sync
  - [ ] Channel display
  - [x] Notifications
  - [ ] Other:

## Rollback plan

This is a heavy touch up of all activity types. A clean revert is not possible, we should be prepared to fix any arising bugs related to type changes.


mikolajpp:
@claude review once

mikolajpp:
@claude review once

desk/app/activity.hoon
```diff
@@ -231,6 +238,11 @@
     ==
   +$  state-1
     [%1 =indices:v3:av =activity:v2:av =volume-settings:v3:av]
+  ::
+  ++  state-9-to-10
+    |=  old=state-9
+    ^-  state-10
+    [%10 allowed indices activity volume-settings]:old
```
mikolajpp: Types have widened so we don't need dedicated conversions here.

desk/app/activity.hoon
```diff
@@ -360,6 +372,10 @@
       %adjust   (adjust +.action)
       %allow-notifications  (allow +.action)
     ==
+  ::
+      %activity-action
+    =+  !<(=action:v8:av vase)
+    $(+< activity-action-1+!>(`action:v9:av`action))
```
mikolajpp, line 382: Likewise, types have widened, so a cast suffices.

desk/app/activity.hoon
```diff
@@ -509,44 +543,82 @@
...
   ::
-      [%x %v4 %event id=@ ~]
-    =/  =time-event:v8:av
+      [%x ver=?(%v4 %v5) %event id=@ ~]
+    =/  =time-event:a
       [id.pole (got:on-event:a stream:base (slav %da id.pole))]
-    ``activity-event+!>(time-event)
+    ?-    ver.pole
+        %v4
+      ?~  time-event-8=(v8:time-event:v9:ac time-event)  ~
```
mikolajpp, line 581: v9 introduced new variants which the previous version is unable to handle. If the conversion  returns empty, we don't have an update to send out.

desk/app/activity.hoon
```diff
@@ -645,7 +745,8 @@
   ::  after the start so we always get "new" sources when paging
   ?.  ?&  ?|(notified.event ?=(%contact -<.event))
           (lth latest.src-info start)
-          ?=  $?  %post  %reply  %dm-post  %dm-reply
+          ?=  $?  %post  %reply  %react
+                  %dm-post  %dm-reply  %dm-react
                   %flag-post  %flag-reply  %group-ask
                   %contact
```
mikolajpp, line 755: CC @arthyn for logic check.

desk/app/activity.hoon
```diff
@@ -711,7 +812,8 @@
   |=  [acc=out [=time =event:a]]
   ?:  =(limit.acc 0)  [~ & acc]
   ?:  child.event  [~ | acc]
-  ?.  ?=  $?  %post  %reply  %dm-post  %dm-reply
+  ?.  ?=  $?  %post  %reply  %react
+              %dm-post  %dm-reply  %dm-react
               %flag-post  %flag-reply  %group-ask
               %contact
```
mikolajpp, line 822: CC @arthyn for logic check.

desk/app/activity.hoon
```diff
@@ -757,7 +860,6 @@
       ==
   ~>  %spin.['give-update']
   ^+  cor
-  %-  (log |.("{<[update dist]>}"))
```
mikolajpp, line 760: I am not sure why we need those. They were probably useful for debugging at some point in the past. If we still want them, we can add them back using logs debugging mode. @arthyn

desk/app/activity.hoon
```diff
@@ -781,12 +894,17 @@
...
     (give-update update [%hose ~])
   =?  cor  &(!importing notify (is-allowed:evt allowed inc))
-    =/  =time-event:v8:av
-      [time-id event]
-    (give %fact ~[/v4/notifications] activity-event+!>(time-event))
+    =/  time-event=(unit time-event:v8:av)
+      (v8:time-event:v9:ac [time-id event])
+    =?  cor  ?=(^ time-event)
+      (give %fact ~[/v4/notifications] activity-event+!>(u.time-event))
+    (give %fact ~[/v5/notifications] activity-event-1+!>(`time-event:v9:av`[time-id event]))
```
mikolajpp, line 907: What the AI did here when asked to upgrade this portion of code for new version: it dispatched on the converted v8 time-event type, not returning anything if that was null. The code would work, except, of course, for actually returning what it was updated for! 

This, as far as I can tell, has nothing to do with Hoon/Urbit knowledge, but is a typical failure mode of approximating logic with a statistical language model.

An error easy to catch, but it highlights how using these things to code places enormous demands on providing *just* the right context, a massive amounts of tests etc. At which point you hear a buzz and a little voice whispering "just use the vim, Luke" :-)

That said, as I am reviewing this PR I am throwing bits and pieces refactoring for the model to finish, while carrying on with review, which works nice & easy.

The experiments shall continue while models improve.

desk/app/channels.hoon
```diff
@@ -1785,7 +1785,7 @@
       .^(? %gu p)
     ::
     ++  on-post
-      |=  v-post:c
+      |=  v-post:d
```
mikolajpp, line 1798: This is unfortunate. It could be fixed by unifying all standard imports across the codebase (we have a few conflicts, such as `c` for both chat and channels, which is something we should probably do with guardian in view, but we punt in this PR.

desk/app/chat.hoon
```diff
@@ -2184,8 +2186,18 @@
...
-      ?-  -.q.diff.delta
-          ?(%add-react %del-react)  (cu-give-writs-diff diff.delta)
+      ?-    -.q.diff.delta
+          %add-react
+        =?  cu-core  &(?=(^ had) ?=(%& -.writ.u.had))
+          =/  message-key  [p.diff.delta now.bowl]
+          =*  content  content.writ.u.had
+          =/  mention  (was-mentioned:utils content our.bowl ~)
+          =/  concern  [%react message-key ~ react.q.diff.delta]
+          (cu-activity concern content mention)
```
mikolajpp: It's likely we don't care about whether the message we react to mentioned us, correct?
I included it for completeness, since activity facts carry that field for all events.
@arthyn

desk/app/chat.hoon
```diff
@@ -2233,7 +2245,17 @@
             (emit (tell:log %crit message metadata))
           cor
         ?-  -.delt
-            ?(%add-react %del-react)  (cu-give-writs-diff diff.delta)
+            %add-react
+          =?  cu-core  ?&(?=(^ reply) ?=(%& -.reply.u.reply))
+            =/  top-key  [id time]:writ.u.entry
+            =/  message-key  [id time]:reply.u.reply
+            =*  content  content.reply.u.reply
+            =/  mention  (was-mentioned:utils content our.bowl ~)
```
mikolajpp: Same as before, we probably don't care about mentions here.

desk/app/chat.hoon
```diff
@@ -2707,8 +2739,18 @@
...
         cor
-      ?-  -.delta
-          ?(%add-react %del-react)  (di-give-writs-diff diff)
+      ?-    -.delta
+          %add-react
+        =?  di-core  ?&(?=(^ reply) ?=(%& -.reply.u.reply))
+          =/  top-key  [id time]:writ.u.entry
+          =/  message-key  [id time]:reply.u.reply
+          =*  content  content.reply.u.reply
+          =/  mention  (was-mentioned:utils content our.bowl ~)
```
mikolajpp, line 2799: Make sure that we can remove mention here.

desk/app/notify.hoon
```diff
@@ -203,14 +223,20 @@
...
+          (~(watch-our pass:io /reads) %activity /v5/reads)
+      ==
     =?  caz  !(~(has by wex.bowl) [/activity our.bowl %activity])
-      :-  (~(watch-our pass:io /activity) %activity /v4/notifications)
+      :-  (~(watch-our pass:io /activity) %activity /v5/notifications)
       ?.  =(~rivfur-livmet our.bowl)  caz
       :_  caz
       [%pass / %agent [our.bowl %notify] %poke %provider-state-message !>(0)]
     =?  caz  !(~(has by wex.bowl) [/reads our.bowl %activity])
       :_  caz
```
mikolajpp, line 239: Question for @arthyn. Should this also be guarded by the check for the provider, as above?

desk/app/notify.hoon
```diff
@@ -446,8 +472,9 @@
     ::
         [%x %note uid=@ %activity-event ~]
       =/  =uid  (need (mate (slaw %da uid.pole) (slaw %uv uid.pole)))
-      =/  =event  (~(got by notifications) `@`uid)
-      ``activity-event+!>(event.event)
+      =/  event=event-8  (~(got by notifications) `@`uid)
+      ?~  time-event=(v8:time-event:v9:ac [`@da`uid event.event])  [~ ~]
+      ``activity-event+!>(u.time-event)
```
mikolajpp, line 487: This is marked, so we preserve the old type. If this is actually used by the client we will soon know about it and update this endpoint.

CC @arthyn

desk/lib/activity-conv.hoon
```diff
@@ -0,0 +1,499 @@
+::  activity-conv: activity types conversions
```
mikolajpp, line 1: This is 90% machine generated. Required steering at times, but a dedicated hoon type/state migration agent seems to be within realm of possibility.

That said, if we had any good compiler tooling, this could actually be *machine generated* (as opposed to vibe-generated) with 100% precision, with some manual input, in a fraction of a second.

desk/lib/activity-conv.hoon
```diff
@@ -0,0 +1,499 @@
...
+  ++  activity-summary
+    |%
+    ++  v8
+      |=  =activity-summary:v9:av
+      ^-  activity-summary:v8:av
+      :*  newest.activity-summary
+          count.activity-summary
+          notify-count.activity-summary
+          notify.activity-summary
+          unread.activity-summary
```
mikolajpp, line 64: Question for @arthyn. The counts here do not include reactions, which are ignored (as implemented in `lib/activity.hoon`). Should they? It would require some work to make sure down-converted summaries (which do not carry react events) reflect the counts properly.

desk/lib/activity-conv.hoon
```diff
@@ -0,0 +1,499 @@
...
+      (v8:activity summaries.feed)
+    --
+  ++  activity-bundle
+    |%
+    ++  v8
+      |=  =activity-bundle:v9:av
+      ^-  (unit activity-bundle:v8:av)
+      %-  some
+      :*  (v8:source source.activity-bundle)
+          latest.activity-bundle
```
mikolajpp, line 101: Based on my reading of `+feed` in activity, we would include reaction events with `%all`. In that case, the timestamp here would be wrong if the top item be a reaction, and the whole bundle be down-converted. Could that manifest as client bugs @arthyn ?

I think this wasn't an issue when we released contact updates, which also did not bother with rectifying timestamps on downconversions.

desk/sur/activity.hoon
```diff
@@ -144,6 +150,13 @@
       content=story:s
       mention=?
   ==
+::
++$  dm-react-event
+  $:  key=message-key
+      parent=(unit message-key)  ::  either dm or dm-reply
+      =whom
+      =react:ch
+  ==
```
Fang-, line 160: This should probably include `.author`, like `$react-event` does. The ship in the `.key` is the author of the _message_, not of the react.

(I forgot we generally get the _message_ author from the key (right? or do we just not have it??). Feels kinda rough, semantically, but whatever...)

desk/app/activity.hoon
```diff
@@ -118,16 +118,13 @@
...
 ++  abet  [(flop cards) state]
 ++  cor   .
 ++  emit  |=(=card cor(cards [card cards]))
 ++  emil  |=(caz=(list card) cor(cards (welp (flop caz) cards)))
 ++  give  |=(=gift:agent:gall (emit %give gift))
 ++  from-self  =(our src):bowl
-++  log
-  |=  msg=(trap tape)
-  ?.  verbose  same
-  (slog leaf+"%{(trip dap.bowl)} {(msg)}" ~)
```
Fang-, line 130: I don't have warm feeling (or feeling at all) about this debug logging mechanism. We'll leave it deleted as-is unless @arthyn wants it back.

desk/app/activity.hoon
Fang-:
```suggestion
    old(- %10)
```

(For the record we had a whole discussion here about whether to even bump the tag. It really doesn't matter.)

desk/app/activity.hoon
```diff
@@ -395,101 +411,119 @@
...
+      ==
+    ?-    ver.pole
+        %v5
+      ``activity-feed-init-5+!>((v8:feed-init:v9:ac feed-init))
+    ::
+        %v6
+      ``activity-feed-init-6+!>(`feed-init:v9:av`feed-init)
     ==
   ::
-      [%x %v4 %feed type=?(%all %mentions %replies) count=@ start=?(~ [u=@ ~])]
```
Fang-, line 454: Where did this `/v4/feed` endpoint go? We should make sure it was unused.

desk/app/channels.hoon
Fang-: Consider throwing `^?` on the type cores. Either `+v9` or `+av` broadly. Not sure what works, or if it can work at all, but if `=,` is clobbering names you care about even though they're not defined inside the thing you're `=,`ing then that's a nasty problem for sure.

desk/app/channels.hoon
```diff
@@ -1802,8 +1802,30 @@
...
+        |=  [[=author:d =react:d] actions=(list action)]
+        =/  old-react=(unit react:d)  (~(get by old) author)
+        ?:  =(old-react `react)  actions
+        =/  author-ship  (get-author-ship:utils author)
+        ?:  =(author-ship our.bowl)
+          :_  actions
+          [%read source [%all `now.bowl |]]
+        ?:  (blocked author-ship)  actions
+        :_  actions
+        [%add %react [[author-ship id-post] id-post] ~ nest group.perm.channel author react]
```
Fang-: `author-ship` here is in the place of the message author, but the value will be that of the react author...

```suggestion
        [%add %react [[xx id-post] id-post] ~ nest group.perm.channel author react]
```

desk/app/channels.hoon
```diff
@@ -1864,8 +1886,34 @@
...
+        |=  [[=author:d =react:d] actions=(list action)]
+        =/  old-react=(unit react:d)  (~(get by old) author)
+        ?:  =(old-react `react)  actions
+        =/  author-ship  (get-author-ship:utils author)
+        ?:  =(author-ship our.bowl)
+          :_  actions
+          [%read source [%all `now.bowl |]]
+        ?:  (blocked author-ship)  actions
+        :_  actions
+        [%add %react [[author-ship id.reply] id.reply] `parent-key nest group.perm.channel author react]
```
Fang-: Same thing here!

Fang- (commented):
Comments below per our discussion!

desk/app/activity.hoon
mikolajpp: Checking client's git history it looks like it was at `%v5` 2 years ago. CC @arthyn to double check the client does not use `/v4/feed` endpoint anywhere.

desk/app/notify.hoon
arthyn: oh yeah probably

desk/app/notify.hoon
arthyn: not used by the client so I think we're ok

desk/app/gateway-status.hoon
```diff
@@ -38,8 +38,47 @@
       reply-cooldown=@dr
       active-window=@dr
   ==
+::  $state-1: activity subscription version upgrade
```
arthyn, line 41: did we actually need to migrate state here? `message-key` shouldn't have changed type, but maybe this is just so we have a marker to trigger on?

desk/app/chat.hoon
arthyn: yeah I don't think we care about mentions when it comes to reacts, so feel free to just pass false or leave in

desk/app/chat.hoon
arthyn: same comment here

desk/app/chat.hoon
arthyn: yep

desk/lib/activity-conv.hoon
arthyn: for older versions I think its fine to not include them

desk/lib/activity-conv.hoon
arthyn: I think the worst it'll do is put an activity bundle at the top when it should be lower, not a big deal, fine as-is.

desk/app/activity.hoon
arthyn: nope fine to use the logging library imo, there were some logs which were too large which is why these were trapped, but if those are gone its probably fine.

desk/app/activity.hoon
arthyn: I believe we immediately jumped to 5 because the original implementation didn't work like we wanted it to? I think its ok to delete

desk/app/activity.hoon
arthyn: lgtm

desk/app/activity.hoon
arthyn: 👍

desk/app/activity.hoon
arthyn: fine to remove

arthyn (approved):
lgtm!

Fang- (approved):
Thanks for the fixes!

desk/app/gateway-status.hoon
mikolajpp: Yes, that's what this comment alludes to. This is a migration to update the subscription, so we need a version number bump to trigger.

mikolajpp:
@codex review

mikolajpp:
@claude review once

desk/app/channels.hoon
mikolajpp: Didn't work. Putting the cast at top level `|av` core does not work. The context from inner cores is still exposed. Putting a cast on an inner core such as `|v9` also does not work, because it inherits names from the previous version `|v8` which are now rendered inaccessible.

mikolajpp:
@codex review

mikolajpp:
@codex review, and @claude review once

mikolajpp:
@codex review once.

desk/app/channels.hoon
```diff
@@ -1802,8 +1802,41 @@
...
+      =/  =message-key
+        [[post-author id.post] id.post]
+      =/  =source  [%channel nest group.perm.channel]
+      =/  del-actions=(list action)
+        %+  roll  ~(tap by old)
+        |=  [[=author:d =react:d] actions=(list action)]
+        =/  new-react=(unit react:d)  (~(get by new) author)
+        ?:  =(new-react `react)  actions
+        :_  actions
+        [%del-event source [%react message-key ~ nest group.perm.channel author react]]
```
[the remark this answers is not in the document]
mikolajpp: Is that a problem @arthyn ? I don't see us removing a post from parent sources, so I think the robot is missing some context on filtering events in activity.

arthyn (approved):

PR #5720, opened 2026-04-15
fix: pop to existing tab screens on NavBar taps

patosullivan:
## Summary

Fixes TLON-5598 by passing `{ pop: true }` to every `navigate()` call that targets a top-level tab route (`ChatList`, `Activity`, `Contacts`, `Settings`).

Also fixes TLON-5608.

React Navigation 7 silently changed the semantics of `navigate(name)`: in RN 6, calling `navigation.navigate('ChatList')` popped back to an existing `ChatList` instance already in the stack. In RN 7, the same call **pushes a new screen** with a fresh key, so `react-native-screens` mounts a brand-new instance with all its queries, subscriptions, and effects. Users saw the press ripple immediately but the new screen took 100–500 ms to mount and paint on Android, which felt like a missed tap. Passing `{ pop: true }` as the third argument restores the RN 6 behavior — RN 7's router re-uses the existing route in the stack instead of pushing a new one.

The `{ pop: true }` convention was already established on the Expo 54 branch in earlier cleanup commits (`378ea8196`, `b80321e60`); this PR covers the sites those passes missed. It also adds an ESLint rule so the convention stays enforced going forward.

## Changes

**Runtime fixes — direct response to the bug report**

Three mobile NavBar tap handlers, one set per top-level screen:

-   `packages/app/features/top/ChatListScreen.tsx` — 3× `NavBarView` handlers (`navigateToContacts`, `navigateToHome`, `navigateToNotifications`)
-   `packages/app/features/top/ActivityScreen.tsx` — 3× `NavBarView` handlers
-   `packages/app/features/top/ContactsScreen.tsx` — 3× `NavBarView` handlers

**Additional sites surfaced by the new lint rule**

Same bug class, same fix, different entry points. The lint rule caught these on its first run across `packages/app`:

-   `packages/app/features/top/ActivityScreen.tsx` — `handleNavigateToContacts`
-   `packages/app/features/top/ContactsScreen.tsx` — Settings header icon (`onPress={() => navigate('Settings')}`)
-   `packages/app/features/top/GroupChannelsScreen.tsx` — `handleGoBackPressed` (narrow-window branch)
-   `packages/app/features/groups/GroupMetaScreen.tsx` — `navigateToHome` after group delete
-   `packages/app/hooks/useChatSettingsNavigation.ts` — `onLeaveGroup` (narrow-window branch)
-   `packages/app/hooks/useGroupNavigation.ts` — `goToHome`
-   `packages/app/navigation/GroupSettingsStack.tsx` — `navigateToHome`, surfaced after tightening the selector to handle `TSAsExpression` wrappers in follow-up review. Rewritten to use `getParent()?.navigate('ChatList', undefined, { pop: true })` in place of the triple-`as never` cast — behaviorally equivalent, since when the inner stack can't handle a `NAVIGATE` action to `ChatList`, RN's router bubbles it to the parent anyway (verified in `@react-navigation/core`'s `useOnAction`).

**Lint rule — `.eslintrc.cjs` (29 additions, no modifications to existing rules)**

Two `no-restricted-syntax` entries flag any `navigate()` call whose first argument is the string literal `'ChatList' | 'Activity' | 'Contacts' | 'Settings'` and whose subtree does not contain `pop: true`. Notes:

-   The selector uses `:matches([arguments.0.value=…], [arguments.0.expression.value=…])` so TS-cast forms like `navigate('ChatList' as never)` are also caught (the `TSAsExpression` wrapper otherwise hides the literal from the `arguments.0.value` path).
-   The escape hatch is `[value.value=true]`, not just `:has(Property[key.name="pop"])`, so `{ pop: false }` does not bypass the rule.
-   Two variants: one for `someObj.navigate(...)` (`callee.property.name`) and one for destructured `const { navigate } = ...; navigate(...)` (`callee.name`).
-   Known non-blocking false negatives: `navigate('ChatList', { pop: true })` with `pop:true` in the wrong position (already a TS error in typed call sites), deeply nested `pop: true` buried under an unrelated property, and the dynamic-variable case `const r = 'ChatList' as const; navigate(r)` (general `no-restricted-syntax` limitation). A `useTabNavigate()` helper or a custom ESLint plugin would close these gaps if we want stronger guarantees later.

**Tradeoffs to be aware of**

`{ pop: true }` tells RN 7's stack router to pop back to an existing instance of the target route if one exists — which is what we want for tab-like behavior on a stack navigator — but it is not free:

-   It drops any intermediate stack entries between the current screen and the target. If a user is three screens deep in `ChatList → Channel → Post` and taps the Home tab, they land on `ChatList` with the intermediate screens unmounted. This matches the RN 6 behavior the app was built around and matches user expectations for a bottom tab bar, but it is not identical to a true `BottomTabs` navigator, which would preserve per-tab sub-stacks.
-   Per-tab state (scroll position, transient UI state in sub-screens) is not preserved across tab switches the way it would be with `BottomTabs`. In practice this is already how the mobile app behaves on the current `develop` branch — we are restoring that behavior, not regressing from it.
-   A proper longer-term fix would be to migrate the mobile top-level navigator to `BottomTabs`, which would preserve per-tab sub-stacks natively. That's out of scope here — this PR is the smallest change that gets Expo 54 to feel right on Android.

## How did I test?

-   `pnpm -r tsc --noEmit` clean in `packages/app`
-   `pnpm lint:all` passes with zero TLON-5598 errors across the monorepo, and re-running the new rule against the working tree flags the adjacent sites it was designed to catch (confirmed by reverting each fix locally)
-   Manually tested the bottom nav bar on a physical Android device on the Expo 54 branch before and after the fix. Taps feel responsive again — no more duplicate-mount pause.
-   Spot-checked each of the non-NavBar call sites (`useChatSettingsNavigation`, `useGroupNavigation`, `GroupMetaScreen`, `GroupChannelsScreen`, `GroupSettingsStack`) on a dev build to confirm the return-to-ChatList flows land on the existing screen instead of pushing a fresh one.
-   Audited desktop navigation: `packages/app/navigation/desktop/TopLevelDrawer.tsx` uses `reset(...)` for section switching, not `navigate()`, so there is no desktop equivalent of this bug and no desktop-side change is needed.

## Risks and impact

-   Safe to rollback without consulting PR author? **Yes**
-   Affects important code area: Navigation.

## Rollback plan

Revert.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -163,6 +163,35 @@ module.exports = {
         message:
           'Please use the useTypedReset() hook instead of importing reset from @react-navigation/native for type safety.',
       },
+      {
+        // `navigation.navigate('ChatList' | 'Activity' | 'Contacts' | 'Settings', …)`.
+        // Also matches TS casts like `navigate('ChatList' as never)`, which
+        // wrap the literal in a TSAsExpression.
+        // The escape hatch is a `pop: true` property anywhere in the call's
+        // subtree (accepts both the top-level `{ pop: true }` options arg
+        // and nested-param forms like `getDesktopChannelRoute`).
+        selector:
+          'CallExpression[callee.property.name="navigate"]' +
+          ':matches(' +
+          '[arguments.0.value=/^(ChatList|Activity|Contacts|Settings)$/],' +
+          '[arguments.0.expression.value=/^(ChatList|Activity|Contacts|Settings)$/]' +
+          ')' +
+          ':not(:has(Property[key.name="pop"][value.value=true]))',
+        message:
+          "navigate() to a top-level tab route must pass { pop: true } as the third argument. React Navigation 7's navigate() pushes a new screen by default — without pop:true this causes duplicate screen mounts and perceived input delay on Android. See TLON-5598.",
+      },
+      {
+        // Destructured `navigate('ChatList' | ...)` (e.g. `const { navigate } = props.navigation`).
+        selector:
+          'CallExpression[callee.name="navigate"]' +
+          ':matches(' +
+          '[arguments.0.value=/^(ChatList|Activity|Contacts|Settings)$/],' +
+          '[arguments.0.expression.value=/^(ChatList|Activity|Contacts|Settings)$/]' +
+          ')' +
+          ':not(:has(Property[key.name="pop"][value.value=true]))',
+        message:
+          "navigate() to a top-level tab route must pass { pop: true } as the third argument. React Navigation 7's navigate() pushes a new screen by default — without pop:true this causes duplicate screen mounts and perceived input delay on Android. See TLON-5598.",
+      },
     ],
   },
 };
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -167,28 +167,35 @@ module.exports = {
         // `navigation.navigate('ChatList' | 'Activity' | 'Contacts' | 'Settings', …)`.
         // Also matches TS casts like `navigate('ChatList' as never)`, which
         // wrap the literal in a TSAsExpression.
-        // The escape hatch is a `pop: true` property anywhere in the call's
-        // subtree (accepts both the top-level `{ pop: true }` options arg
-        // and nested-param forms like `getDesktopChannelRoute`).
+        // The escape hatch requires an ObjectExpression third argument
+        // containing `pop: true` (including quoted `"pop"` keys).
         selector:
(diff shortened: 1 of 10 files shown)


patosullivan:
Confirmed that this also fixes TLON-5608

janicduplessis (approved):

janicduplessis:
Nice! This new behaviour is quite annoying and was the main source of issues when fixing e2e tests.

PR #5730, opened 2026-04-17
fix share issues

dnbrwstr:
## Summary

- Fixes text shared into chat drafts and keeps quote insertion on the same draft path.
- Fixes iOS 26 screenshot-modal image shares. Those shares can arrive as PNG data instead of a file URL, so the share extension now writes the bytes to a temporary image before continuing through the existing media-share flow.
- Fixes occasional zombie share sheet popping up

Fixes TLON-5620
Fixes TLON-5623
Fixes TLON-5627

## Changes

- add draft reload support to the chat input
- use stored drafts for shared text and DM quote text
- remove the old text attachment path
- handle raw image data in the iOS share extension
- update the expo-share-intent patch hash in the lockfile
- stop rendering share sheet once it's closd

## How did I test?

- Tested on iOS and Android


packages/app/ui/components/ChatMessage/ChatMessageActions/MessageActions.tsx
```diff
@@ -324,6 +331,33 @@ export async function handleAction({
...
+  // Missing context means the provider wiring is wrong, not that we should fall back to refs.
+  if (!ctx) {
+    throw new Error('Cannot quote DM text without a draft input context');
+  }
+
+  const draft = await ctx.getDraft();
+  const { text: existing, mentions } = draft
+    ? logic.contentToTextAndMentions(draft)
+    : { text: '', mentions: [] };
+  await ctx.storeDraft(
```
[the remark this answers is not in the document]
dnbrwstr: This simplifies and the race window is small enough that I'm not worried.

packages/app/ui/components/ChatMessage/ChatMessageActions/MessageActions.tsx
```diff
@@ -324,6 +331,33 @@ export async function handleAction({
...
 }
 
+async function prependTextToDraft(
+  ctx: DraftInputContext | null | undefined,
+  text: string
+) {
+  // DM quote actions should be rendered under the current draft input context.
+  // Missing context means the provider wiring is wrong, not that we should fall back to refs.
+  if (!ctx) {
+    throw new Error('Cannot quote DM text without a draft input context');
```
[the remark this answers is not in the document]
dnbrwstr: Fixed by avoiding rendering quote action when draft context is unavailable in 3438675.

packages/app/ui/components/BareChatInput/index.tsx
```diff
@@ -689,31 +704,48 @@ export default function BareChatInput({
...
+    } catch (e) {
+      bareChatInputLogger.error('Error loading draft', e);
+    }
+  }, [editingPost, getDraft, setInputFromDraft]);
+
+  useImperativeHandle(
+    ref,
+    () => ({
+      exitFullscreen: () => {},
+      startDraft: () => void reloadDraft(),
```
davidisaaclee, line 718: future-looking: I wonder if making this `startDraft(draft: JSONContent | null): void` (across all input components - could be stubbed for now for other inputs) would be an easy/good step towards a stronger draft model. only worth doing now if you think someone might write a similar but different method on another draft input in the future.

packages/app/ui/components/Channel/index.tsx
```diff
@@ -761,41 +765,45 @@ export const Channel = forwardRef<ChannelMethods, ChannelProps>(
...
-                                    onScrollEndReached,
-                                    onScrollStartReached,
-                                    posts: posts ?? undefined,
-                                    scrollToBottom: onPressScrollToBottom,
-                                    selectedPostId,
-                                    setEditingPost,
-                                    LegacyPostView: PostView,
-                                    PostView: ConnectedPostView,
-                                  }}
+                                <DraftInputContextProvider
```
davidisaaclee: nit: semantically, I think this `DraftInputContextProvider` should be right at the ChannelProvider level – e.g. starting a draft from the `ChannelHeader` would not get this context (but it should)

packages/app/ui/components/ChatMessage/ChatMessageActions/MessageActions.tsx
```diff
@@ -250,6 +257,7 @@ export async function handleAction({
   onForward?: (post: db.Post) => void;
   onViewReactions?: (post: db.Post) => void;
   addAttachment: (attachment: Attachment) => void;
+  draftInputContext?: DraftInputContext | null;
```
davidisaaclee: nit: it feels like an abstraction leak to pass the full `DraftInputContext` here – I think we should at least limit it by doing `Pick<DraftInputContext, 'getDraft' | 'storeDraft' | 'draftInputRef'>` (ideally, would pull out the `startDraft` to not leak the `draftInputRef`)

packages/app/ui/components/draftInputs/shared.tsx
```diff
@@ -45,7 +45,7 @@ export interface DraftInputContext {
   channel: db.Channel;
   clearDraft: (draftType?: GalleryDraftType) => Promise<void>;
   configuration?: Record<string, JSONValue>;
-  draftInputRef?: React.Ref<DraftInputHandle>;
+  draftInputRef?: React.RefObject<DraftInputHandle>;
```
davidisaaclee: nit: no big deal but I think this shouldn't be necessary – I'm guessing this is because we're trying to both forward the ref and use it in the ref provider.

this is what I use for this use case:

```ts
import * as React from "react";

import { useMutableRef } from "./useMutableRef";

/** A hook that allows you to use a ref that is optionally passed in as a prop.
 *
 * ```tsx
 * function MyComponent({ ref }) {
 *   const [innerRef, update] = useInnerRef(ref); // <- `ref` is a React.Ref<T> from props
 *
 *   // Now you can use the ref
 *   useLayoutEffect(() => {
 *     innerRef.current?.focus();
 *   }, [innerRef]);
 *
 *   return <div ref={update} />;
 * }
 * ```
 */
export function useInnerRef<T>(
  innerRef: React.Ref<T> | null | undefined,
  willSet?: (value: T | null) => void,
): [ref: React.RefObject<T | null>, update: (x: T | null) => void] {
  const valueRef = React.useRef<T>(null);
  const willSetRef = useMutableRef(willSet);

  const update = React.useCallback(() => {
    const value = valueRef.current;
    if (innerRef == null) {
      return;
    }

    if (typeof innerRef === "function") {
      innerRef(value);
    } else {
      // @ts-ignore - need to write to readonly property `current`
      innerRef.current = value;
    }
  }, [innerRef, valueRef]);

  React.useEffect(update, [update]);

  return [
    valueRef,
    React.useCallback(
      (value: T | null) => {
        if (valueRef.current === value) {
          return;
        }
        willSetRef.current?.(value);
        valueRef.current = value;
        update();
      },
      [update, valueRef, willSetRef],
    ),
  ];
}
```

packages/app/ui/components/PostScreenView.tsx
```diff
@@ -768,36 +773,38 @@ function SinglePostView({
...
-        channel &&
-        canWrite &&
-        !(
-          isEditingParent &&
-          (channel.type === 'notebook' || channel.type === 'gallery')
-        ) && (
-          <View id="reply-container" {...containingProperties}>
-            <DraftInputContextProvider value={replyDraftInputContext}>
+      {/* Keep reply context off parent editing UI; attachment actions use whichever draft context they inherit. */}
+      <DraftInputContextProvider value={replyDraftInputContext}>
```
davidisaaclee: 🎈 Won't this break editing the parent post, since the "edit post" draft input will inherit the reply context?

patches/expo-share-intent@3.2.3.patch
```diff
@@ -32,6 +32,44 @@ index 1234567..89abcde 100644
...
++            return
+           }
+ 
+           var pixelWidth: Int? = nil
+@@ -235,5 +256,20 @@ class ShareViewController: UIViewController {
+     }
+     return screenshotURL
+   }
+ 
++  private func saveImageData(_ data: Data) -> URL? {
```
davidisaaclee: optional: I was curious if we could detect if we're passed an invalid/corrupt `Data` – we _could_ read the type from the header to check that the data is definitely an image: https://stackoverflow.com/a/32765708

This would also allow us to use the definitely-correct file extension (although I bet that the OS will always be passing PNGs as screenshots).

patches/expo-share-intent@3.2.3.patch
```diff
@@ -32,6 +32,44 @@ index 1234567..89abcde 100644
...
+           if let dataURL = item as? URL {
+             url = dataURL
+           } else if let imageData = item as? UIImage {
+             url = self.saveScreenshot(imageData)
++          } else if let imageData = item as? Data {
++            url = self.saveImageData(imageData)
++          }
++
++          guard url != nil else {
++            NSLog("[ERROR] Cannot load image URL content !\(String(describing: item))")
```
davidisaaclee, line 60: Error message is inaccurate:

> image URL

not necessarily an image

packages/app/ui/components/ForwardToChannelSheet.tsx
```diff
@@ -31,6 +31,12 @@ export function ForwardToChannelSheet({
 }: ForwardToChannelSheetProps) {
   const showSelector = useDelayedClose(open);
 
+  // Unmount after the close window; otherwise the empty sheet shell can
+  // visually resurface during later navigation.
```
davidisaaclee, line 35: any idea why the sheet was resurfacing?

packages/app/ui/components/ChatMessage/ChatMessageActions/MessageActions.tsx
```diff
@@ -135,6 +135,13 @@ const ConnectedAction = memo(function ConnectedAction({
       case 'unpinPost':
         // only show for admins on the currently pinned post
         return currentUserIsAdmin && pinnedPostId === post.id;
+      case 'quote':
+        // DM quotes write into the composer. Search/detail surfaces can render
```
davidisaaclee: 🎈 Why doesn't this apply to `chat` quotes?

davidisaaclee (changes requested):
marked blocking comments with 🎈

packages/app/ui/components/ForwardToChannelSheet.tsx
dnbrwstr: I chased this pretty far down, but wasn't able to get to a root issue in a reasonable amount of time. When the zombie sheet appeared, `open` was still set to false, which is why the sheet content was apparently missing. My best guess is that some kind of recalc, possibly keyboard related, was being applied inaccurately and thus positioning the sheet somewhere it was actually visible, even when it was supposed to be hidden.

packages/app/ui/components/PostScreenView.tsx
dnbrwstr: `editingPost` was getting passed through, so this didn't actually break, but this was kind of confusing in the context of chat posts which do reuse this same input for editing in the detail view. I've updated the naming (now `threadComposerContext`) and send function to make the split more explicit in 96b9f61

packages/app/ui/components/ChatMessage/ChatMessageActions/MessageActions.tsx
dnbrwstr: There's an odd behavior here that I left alone -- DM quotes are inserted into the draft as text, which requires the `DraftInputContext`, while chat quotes are inserted as refs, which requires `AttachmentContext`. 

That said, I did modify this check to apply to all channel types in 21fe4db. Even though `AttachmentContext` may theoretically be available in situations where `DraftInputContext` isn't, right now that's not the case for quoting. Whenever we're quoting, we expect to have an input available for the quote to go into, so I think it's fine to simplify.

patches/expo-share-intent@3.2.3.patch
dnbrwstr: Updated to handle this by detecting image extension using `CGImageSourceGetType` and alert on failure to detect type correctly in 7377a56. There's some churn in that commit because I also added a deterministic way to update the patch which pulled in some new changes but should prevent future churn. Although at this point I'm thinking we should probably consider `ShareViewController` to be owned by us and drop the patch...

patches/expo-share-intent@3.2.3.patch
dnbrwstr: I think this is accurate? This is inside `handleImages`, which is called if `attachment.hasItemConformingToTypeIdentifier(imageContentType)` is true, so the system is telling us that the thing we're handling should be an image.

packages/app/ui/components/ChatMessage/ChatMessageActions/MessageActions.tsx
dnbrwstr: Addressed in 887594a9b0

patches/expo-share-intent@3.2.3.patch
davidisaaclee: true, my mistake

packages/app/ui/components/Channel/index.tsx
dnbrwstr: Addressed in 57ed4f0.

packages/app/ui/components/draftInputs/shared.tsx
dnbrwstr: I ended up handling this by adding a `startDraft` prop to the context in 91d2fbf41b. This mirrors the other context props (`storeDraft` etc.) and eliminates the need for us to access the ref object directly in consumers.

packages/app/ui/components/BareChatInput/index.tsx
dnbrwstr: I think this makes sense, but am going to punt for now.

packages/app/ui/components/ChatMessage/ChatMessageActions/MessageActions.tsx
```diff
@@ -130,6 +135,10 @@ const ConnectedAction = memo(function ConnectedAction({
       case 'unpinPost':
         // only show for admins on the currently pinned post
         return currentUserIsAdmin && pinnedPostId === post.id;
+      case 'quote':
+        // Quote needs the active composer surface. Search/detail surfaces can
+        // render actions without one, so hide Quote there.
+        return !!draftInputContext;
```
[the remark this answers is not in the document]
dnbrwstr: Addressed in [3aa98f9](https://github.com/tloncorp/tlon-apps/pull/5730/commits/3aa98f93b9ab8b3e23710554a013e0746de00131)

packages/app/ui/components/PostScreenView.tsx
```diff
@@ -768,40 +778,42 @@ function SinglePostView({
...
-        channel &&
-        canWrite &&
-        !(
-          isEditingParent &&
-          (channel.type === 'notebook' || channel.type === 'gallery')
-        ) && (
-          <View id="reply-container" {...containingProperties}>
-            <DraftInputContextProvider value={replyDraftInputContext}>
+      {/* Thread composer context sends new drafts as replies; edits preserve their original target. */}
+      <DraftInputContextProvider value={threadComposerContext}>
```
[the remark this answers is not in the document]
dnbrwstr: Addressed in [3aa98f9](https://github.com/tloncorp/tlon-apps/pull/5730/commits/3aa98f93b9ab8b3e23710554a013e0746de00131)

packages/app/ui/components/BareChatInput/index.tsx
```diff
@@ -689,31 +687,48 @@ export default function BareChatInput({
...
+      setHasSetInitialContent(true);
+      setNeedsHeightAdjustmentAfterLoad(true);
+    },
+    [setMentions]
+  );
+
+  const reloadDraft = useCallback(async () => {
+    try {
+      const draft = await getDraft();
+      if (!editingPost) setInputFromDraft(draft);
```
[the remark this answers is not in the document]
dnbrwstr: Addressed in [1c959f2](https://github.com/tloncorp/tlon-apps/pull/5730/commits/1c959f232037a64162a71f67e148cdf684500682)

scripts/regenerate-expo-share-intent-patch.mjs
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env node
```
davidisaaclee, line 1: not really clear when I'd run this – adding a short comment would help (e.g. "run this if you edit ___")

davidisaaclee (approved):
code all lgtm

scripts/regenerate-expo-share-intent-patch.mjs
dnbrwstr: Added a comment in [2e91f20](https://github.com/tloncorp/tlon-apps/pull/5730/commits/2e91f200af6e81da51809eeadc0b96bafa030470)

PR #5731, opened 2026-04-17
chore(mobile): bump @shopify/flash-list to 2.3.1

janicduplessis:
## Summary

Bumps `@shopify/flash-list` from 2.0.2 to 2.3.1 on mobile to fix chat list layout regressions introduced with the Expo 54 upgrade.

## Changes

- `apps/tlon-mobile/package.json`: `@shopify/flash-list` `2.0.2` → `2.3.1`
- `pnpm-lock.yaml`: regenerated

## How did I test?

Reproduced [TLON-5567](https://linear.app/tlon/issue/TLON-5567/expo-54-incorrect-layout-when-filtering-chats) on iOS by typing into the home/chatlist filter bar — filtered results would land at unpredictable vertical offsets with large empty gaps above them (see issue screenshots). After bumping FlashList to 2.3.1, the filtered list renders correctly with no gaps as characters are typed or deleted.

Also smoke-tested the unfiltered chat list on both tabs (All / Messages) to confirm scrolling, pinned/unpinned ordering, and long-press options still work.

Expected to also fix [TLON-5568](https://linear.app/tlon/issue/TLON-5568/expo-54-chat-list-jumps-when-a-new-conversation-is-added) (chat list jumps when a new conversation arrives), which is the same FlashList layout class of bug in the same component.

## Risks and impact

  - [x] Channel display (chat list rendering)

FlashList 2.x has had a stream of layout bug fixes between 2.0.2 and 2.3.1; the upgrade is within the same major so no API changes. It is a pure-JS package (no native code) so no pod / prebuild changes are needed.

## Rollback plan

Revert the commit and run `pnpm install`.

## Screenshots / videos

See [TLON-5567](https://linear.app/tlon/issue/TLON-5567/expo-54-incorrect-layout-when-filtering-chats) and [TLON-5568](https://linear.app/tlon/issue/TLON-5568/expo-54-chat-list-jumps-when-a-new-conversation-is-added) for the buggy behavior being fixed.

### DIFF
--- a/apps/tlon-mobile/package.json
+++ b/apps/tlon-mobile/package.json
@@ -63,7 +63,7 @@
     "@react-navigation/native-stack": "^7.11.0",
     "@react-native-picker/picker": "^2.11.4",
     "@sentry/react-native": "^7.6.0",
-    "@shopify/flash-list": "2.0.2",
+    "@shopify/flash-list": "2.3.1",
     "@shopify/react-native-skia": "2.2.12",
     "@simform_solutions/react-native-audio-waveform": "^2.1.6",
     "@tamagui/react-native-media-driver": "~2.0.0-rc.22",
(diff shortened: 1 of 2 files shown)


janicduplessis:
For TLON-5568 I am not sure exactly how to repro so I can say it is 100% fixed, would @dnbrwstr or @jamesacklin be able to confirm?

dnbrwstr (approved):
Change makes sense to me. Verified the fix by logging out + back in several times.

PR #5736, opened 2026-04-20
ops: fix expo-share-intent patch

latter-bolden:
I think most of us were seeing issues when attempting to re-install deps since `expo-share-intent` was added. The error was: 

```
Could not apply patch /tlon-apps/patches/expo-share-intent@3.2.3.patch
```

It seemed like the patch itself was malformed, subsequent applications would not apply cleanly. I think it was related to the `@@ -221,8 +221,7 @@` chunk appearing out of order?

I had a model regenerate the patch while maintaining the desired modifications for `ShareExtensionViewController.swift`. With this one, I was able to run `pnpm install` multiple times without encountering the error.


dnbrwstr (approved):
Lgtm. I added a script to regenerate the patch programmatically in my share fixes branch which will hopefully prevent any future issues. I'm also thinking maybe we should just take ownership of the generated file and skip patching.

PR #5737, opened 2026-04-20
chatlist: add tooltip and styles for "Add Chat" button

latter-bolden:
## Summary

Adds heavier styling to emphasize the *Add chat* button and a coach mark for new signups.

## Changes
- new field in `wayfindingProgress` and corresponding DB hook
- new `/Wayfinding/Notices` component
- wire up the tooltip and adjust style for *Add chat* on the `ChatList`

## Risks and impact

  - [x] Onboarding

## Screenshots / videos

Simulator Screenshot - iPhone 16 Plus - 2026-04-20 at 17 27 44

### DIFF
--- a/packages/api/src/types/wayfinding.ts
+++ b/packages/api/src/types/wayfinding.ts
@@ -17,6 +17,7 @@ export interface WayfindingProgress {
   viewedChatChannel: boolean;
   viewedCollectionChannel: boolean;
   viewedNotebookChannel: boolean;
+  tappedHomeAdd: boolean;
   tappedAddNote: boolean;
   tappedAddCollection: boolean;
   tappedChatInput: boolean;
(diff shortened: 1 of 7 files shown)


jamesacklin:
Very heavy. Add a '.' to the end and if the user has a tlawnEnabled flag, say "and invite your Tlonbot to it." So:

`Tap here to create a group and invite your Tlonbot to it.`
or
`Tap here to create a group.`

dnbrwstr:
Code looks fine to me. The new plus button is quite loud and kind of out of line with the rest of our design system. Do you think the tooltip on its own is enough? Or if not, maybe make it blue, but keep our standard icon layout and size?

jamesacklin:
> Code looks fine to me. The new plus button is quite loud and kind of out of line with the rest of our design system. Do you think the tooltip on its own is enough? Or if not, maybe make it blue, but keep our standard icon layout and size?

In favor of making the icon itself blue.

latter-bolden:
> Do you think the tooltip on its own is enough?

Was asked to make the button itself heavier as well

> make it blue, but keep our standard icon layout and size?

Size right now is slightly smaller than was previously to account for padded background. 

I'm okay doing blue icon/no bg, but unless we change how the *Invite your Friends* left icon is styled, they'll initially have blue icons on both sides of the header?

wca4a:
i like what's in the screenshot

PR #5738, opened 2026-04-21
various: tree deletion correctness

Fang-:
## Summary

408 makes compiler changes that catch incorrectly typed keys during map and set deletions. That surfaced these mistakes.

These changes are also included in #5648, but can and should be reviewed and released separately from 408.

## Changes

In channels, when processing post and reply updates, we now make sure we construct a proper `$client-id`, by "unpacking" the author as necessary.

In groups, when handling a `$c-seat` membership command, we now always iterate over all the ships the command applies to.

In presence, when dropping from `.want` due to watch-nack, use the set operation to delete from the set, not the map operation.

## How did I test?

Compiles!

## Risks and impact

## Rollback plan

Revert while we're on 409. Once we're on 408, we won't be able to revert this; it won't compile.


desk/app/groups.hoon
```diff
@@ -2660,9 +2660,15 @@
           ==
       ::  clean up ask and pending record
       ::
-      =.  pending.ad   (~(del by pending.ad) ship)
+      =.  pending.ad
+        %-  ~(rep in ships)
+        |=  [=ship =_pending.ad]
+        (~(del by pending) ship)
```
mikolajpp, line 2666: At this point I want to do something horrible to the Hoon typesystem. The "thin macro layer over nock" language.

desk/app/presence.hoon
```diff
@@ -470,7 +470,7 @@
         [(tell:log %dbug ~['context sub ack ok' >src.bowl< >context<] ~)]~
       ::  nacked, can't do anything, drop desire
       ::
-      :_  this(want (~(del by want) src.bowl context))
+      :_  this(want (~(del in want) src.bowl context))
```
mikolajpp, line 473: Et tu, Brute?

mikolajpp (approved):
This is amazing work! Thank for making the original fix, and catching these really nasty mistakes.

arthyn:
@Fang- can you check out the conflict here so we can merge?

Fang-:
Accidentally superseded by #5796.

PR #5739, opened 2026-04-21
presence: include nack tangs in logs as-is

Fang-:
## Changes

Instead of tanking the tangs, include them as-is in the log tang.

## How did I test?

Compiles!

## Risks and impact

## Rollback plan

Revert.


arthyn (approved):
🫡

arthyn:
target staging!

PR #5740, opened 2026-04-21
fix(mobile): tap outside to dismiss mention popup

wca4a:
## Summary

On mobile there's no Escape key, so an open `@`-mention popup can feel stuck until the user types enough to clear all candidates or picks one. Tapping elsewhere in the chat should dismiss the popup — this is standard mobile dropdown UX.

## Fix

Pass `handleMentionEscape` through `MessageInputContainer` → `InputMentionPopup`. On `Platform.OS !== 'web'` (and only while candidates are visible), render a transparent `Pressable` backdrop positioned behind the popup card that extends above/around the popup but **not** over the message input, and dismisses on tap.

Web is unchanged and still uses the Escape key.

## Test plan

- [x] Verified on iOS simulator (iPhone 17 Pro): type `~` or `@` in a DM or channel input → popup appears → tap on chat messages above → popup dismisses
- [ ] Android — behavior should match iOS (same `Platform.OS !== 'web'` guard)
- [ ] Web — no change, Escape still works


wca4a:
@codex review

wca4a:
@claude review once

wca4a:
@claude review once

wca4a:
@codex review

wca4a:
@codex review

wca4a:
Addressed the review feedback across several amends — current commit is `21bb86240`. Rundown:

**Claude bot — soft vs. hard dismiss.** Backdrop tap was calling `handleMentionEscape`, which also sets `wasDismissedByEscape` and "poisons" the current trigger index so the popup stays hidden for subsequent keystrokes at the same `@`. Added a separate `handleMentionSoftDismiss` in `useMentions` that only toggles `isMentionModeActive` — the next keystroke re-evaluates the trigger normally.

**Claude bot — Android clipping.** `top: -9999` on a `Pressable` inside a bounded `YStack` doesn't work on Android (`clipChildren` default). Moved the backdrop + popup into a tamagui `Portal` so ancestor `View` bounds don't constrain hit-testing.

**Codex P1 — double-counting keyboard height on Android.** Android uses `adjustResize`, so the activity already shrinks above the keyboard and a `Portal`'s `bottom: 0` is the keyboard top. Only add `keyboardHeight` manually on iOS.

**Claude bot — keyboard-already-visible on mount.** Event-listener-only hook stayed at 0 if the component mounted while the keyboard was already up (e.g., channel-to-channel navigation). Initialized with `Keyboard.metrics()?.height ?? 0` so initial state reflects reality.

**Claude bot — scroll gestures swallowed by backdrop.** Decided to accept this as standard mobile UX. Every comparable app (Slack, Discord, iMessage autocompletes) treats the mention dropdown as modal-lite: dismiss or pick first, then scroll. The pre-PR state was "no way to dismiss at all," so trading free scroll for tap-dismiss is the right direction.

**Codex P2 — backdrop over tall multi-line composer.** `containerHeight` is hardcoded to 48 but the real composer grows with multi-line drafts, so the backdrop extended into the actual input area and ate cursor-placement taps. `MessageInputContainer` now measures the input bar's layout height and passes it to `InputMentionPopup` as a separate `inputBarHeight` prop used only for the backdrop's lower edge. The popup itself still anchors to the static `containerHeight` — intentionally, so a huge draft doesn't push the popup off-screen and out of reach.

Tested on iOS simulator: tap-outside dismisses with soft-dismiss behavior (continued typing reopens), keyboard-open and keyboard-closed positioning both correct, tall multi-line composer accepts cursor taps.

arthyn (approved):

PR #5748, opened 2026-04-21
web/db: fix hangs upon receiving a message

arthyn:
## Summary

Fixes TLON-5621, significant receive-path hangs on the web client (multi-second pauses when receiving messages, the DB getting wiped and re-synced on every reload). Addresses the underlying cache-stampede and transaction-queue race uncovered while diagnosing the reports. Also adds a `processChanges` coalescing layer that's necessary to keep the newly-added `volume_settings` trigger from pathologically interacting with SQLocal's postMessage callback model on bulk inserts. Changes are scoped to the web startup/receive paths plus shared DB layer; mobile and desktop share the DB-layer improvements but are unaffected by the web-specific fixes.

## Changes

Two logical commits on the branch.

### Commit 1 — `perf: fix web receive-path hangs`

1. **Opt-in perf instrumentation** — new `packages/shared/src/perfLog.ts` (enable via `localStorage.setItem('tlon:perf', '1')` or `globalThis.__TLON_PERF__ = true`); perf marks in `sync.ts`, `queries.ts`, `query.ts` covering the insert / invalidation hot paths. Zero cost when disabled.
2. **Devtools cache inspector** — `__tlonQueryClient` and `__tlonInspectQueries({group?, top?})` on `globalThis` in `reactQuery.ts`.
3. **`gcTime: 30_000` on per-post queries** — `usePostWithRelations` / `usePostWithThreadUnreads` in `dbHooks.ts`. Default was 5 min; zero-observer entries were bloating the invalidation-predicate scan set.
4. **`deleteReplacedCachedPosts` gate + rewrite** — skip the scan when the batch has no seq>0 non-reply posts (optimistic-send and reply-only batches). When run, use a batch-scoped `DELETE` keyed by incoming `(channelId, sentAt, authorId)` tuples instead of a full-table CTE self-join.
5. **Three partial indexes on `posts`** — `cached_posts_index` (backs `deleteReplacedCachedPosts`), `posts_channel_last_preview` + `posts_channel_last_seq` (back `setLastPosts` subqueries + `getLatestChannelSequenceNum`).

#### Option C — per-post `changeListener` routing

6. **Flatten `['post', id]` queryKey** in `dbHooks.ts`. Old shape `[['post', id], deps]` meant `changeListener`'s `invalidateQueries({ queryKey: ['post', id] })` never matched (partial-key matching is positional). Per-post events were silently no-oping.
7. **Empty `getPostWithRelations.tableDependencies`** in `queries.ts`. Takes per-post queries off the broad table-invalidation path now that the targeted path works.
8. **`changeListener` updates** — switch `post_reactions` and `thread_unreads` from `refetchQueries` to `invalidateQueries`; add a thread-scoped `volume_settings` handler.
9. **`volume_settings` triggers (web + desktop `TRIGGER_SETUP`)** — add INSERT/UPDATE/DELETE triggers gated on `WHEN item_type = 'thread'`, with `DROP TRIGGER IF EXISTS` prefix so existing installs replace any earlier ungated version.

#### Transaction serialization race fix

10. **Move `rootTransaction` from `QueryMeta` to `QueryCtx`** in `query.ts`. `QueryMeta` is shared across every invocation of the same wrapped query via closure, so concurrent calls were falsely detecting each other as "nested" and bypassing the `pendingTransactions` queue. Writes were racing inside a single BEGIN/COMMIT on SQLocal's single worker.

#### Cleanup-query narrowing

11. **`setLastPosts` single-channel monotonic fast path** — single-channel `insertPostsBatch` (the receive hot path) replaces three correlated `ORDER BY … DESC LIMIT 1` subqueries with a forward-only `UPDATE … CASE WHEN`. Multi-channel keeps the existing batched form.
12. **Narrow `cached_posts_index` to `parent_id IS NULL`** — replies also have `sequence_number = 0` per `toPostReplyData`, so the original predicate wasn't actually tiny. Now covers only outstanding optimistic top-level rows.

#### Send-path amplification

13. **`syncChannelMessageDelivery` dedupe** in `sync.ts`. The in-progress check only logged; now returns the in-flight promise. Stops rapid sends from stacking parallel 3s-delayed `syncPosts({count:20})` polls that compound under load.

#### Startup on web

14. **`webMigrator` idempotency** — split migration SQL on `--> statement-breakpoint`, tolerate "already exists" errors per-statement, record the migration as applied once all statements complete (including idempotent no-ops). Breaks the every-reload retry loop caused by `reset-migrations` rotating the tag on every schema change.
15. **Drop dry-run wipe** in `webDb.setupDb`. The "outdated DB → wipe and rebuild" gate was obsolete once the migrator became idempotent, and it was what was wiping the DB on every reload.
16. **`processChanges` debounce (web only)** — coalesce SQLocal callback invocations via `setTimeout(0)`. Without SharedArrayBuffer / Atomics (no COOP/COEP), each `SELECT processChanges()` trigger becomes a separate main-thread postMessage round-trip (~20 ms). Needed to unblock item 9: without coalescing, the new `volume_settings` trigger turned a 682-row bulk insert at startup into ~14s of main-thread blocking. Also benefits any future or existing bulk-insert path on a triggered table by reducing N per-row callbacks to one per event-loop tick. Mobile unaffected (uses `NativeDb` with its own `isProcessingChanges` flag and op-sqlite update notifications, not SQLocal's postMessage callback model).
17. **Stop resetting `headsSyncedAt`** on web load in `app.tsx`. Stale "web doesn't persist" comment from before `sqliteContent` was added — reset forced `syncLatestPosts` to refetch one post per channel across all channels on every reload, producing a multi-hundred-channel `insertLatestPosts` batch.

#### Additional instrumentation

18. **`__tlonDbState()`** — lists applied migrations, `posts` indexes, `posts` row count. **`__tlonRawSql(...)`** — runs arbitrary SQL, **gated behind `__DEV__`** so it doesn't ship in production bundles.
19. **Startup-path perf marks** — `processChanges`, `resetActivity.{fetch,write}`, `syncVolumeSettings.{fetch,setVolumes}`.
20. **`./perfLog` subpath export** in `packages/shared/package.json` so `packages/app/lib/baseDb.ts` can import the helper.

## How did I test?

1. **Reactivity (manual)** — walked through the behaviors that previously depended on the broad `tableDependencies` on `getPostWithRelations`:
   - Edit a post → updated content appears in the scroller without reload.
   - Add / remove a reaction → `ReactionsDisplay` updates immediately.
   - Thread reply arrives → parent-post reply count updates.
   - Mute / unmute a thread → action-menu label reflects the new state on next open.
   - Mark thread read → thread-unread badge clears.

2. **Receive-path perf (instrumented)** — captured `[perf]` traces across the branch. Relevant before → after numbers on the same ship / data:
   - `handleAddPost.total` typical: 200–7000 ms → 4–30 ms.
   - `setLastPosts` (single channel): 400–1278 ms → 4–42 ms.
   - `deleteReplacedCachedPosts` (single post): 100–1000 ms → 0.3–1 ms.
   - `invalidateQueries` peak: 415 ms → 20–110 ms.

   Startup-path `processChanges` also verified: with the coalesce in place, `processChanges ms=…` emits once per batch rather than once per row, and the `setVolumes` bulk insert no longer blocks the main thread. (Not called out as a before/after number — the amplification only existed under item 9's new `volume_settings` trigger, which the same PR adds.)

3. **DB-wipe-on-reload loop** — confirmed (via `__tlonDbState()`) that `__drizzle_migrations` now persists `0000_fearless_naoko` across reloads and the migration stops running.

4. **Mobile / desktop semantics** — `pnpm -r tsc` clean across `@tloncorp/shared`, `@tloncorp/app`, `tlon-web`. Mobile's `NativeDb` has its own coalescing (`isProcessingChanges` flag) and uses op-sqlite update notifications, so the web `processChanges` debounce is a no-op for it. The `volume_settings` `WHEN item_type = 'thread'` gate is a correctness refinement on all platforms. Not executed against a mobile ship in this PR — reviewers running native builds should smoke-test.

## Risks and impact

- Safe to rollback without consulting PR author? **Yes.** Every item is a diff unit that can be reverted independently. Rollback restores prior behavior verbatim (slow, but functional). No data migrations that can't be reapplied.
  - [x] State / providers — React Query cache routing for per-post queries
  - [x] Message sync — `handleAddPost`, `syncChannelMessageDelivery`, `syncLatestChanges`, `setLastPosts`, `deleteReplacedCachedPosts`
  - [x] Channel display — knock-on: `usePostWithRelations` queryKey flatten affects `useLivePost` which is called by every visible post
  - [x] Other: web startup (migrations, `processChanges` callback model), DB schema (3 new partial indexes, new trigger gate)

### Specific things to watch

- **Reactivity regressions on per-post reads.** If any consumer of `usePostWithRelations` / `usePostWithThreadUnreads` depended on the old broad table-deps invalidation for a reactivity surface I didn't enumerate, that surface would stop updating live. Migration into `changeListener`-driven invalidation covers the obvious call sites, but there may be ones I missed.
- **Mobile/desktop migration path.** Both share `packages/app/lib/triggers.ts` and the partial indexes. On first launch the new indexes build against existing `posts` tables — one-time cost proportional to the table size. Should be modest since the indexes are partial.
- **Volume-setting reactivity for non-thread scopes.** Channel/group volume changes no longer fire the changeLog trigger. Consumers of those reactivity paths use channel/group hooks directly (not `['post', id]`) so this should be a non-issue, but worth a sanity check if a reviewer is skeptical.

## Rollback plan

Revert the two commits on the branch: `git revert 62b4d6ff0 315b99744`. Push. All behavior returns to the prior state including the web startup wipe loop (pre-existing) and the receive-path hangs.

If only part of the work regresses something, individual commits can be reverted separately, or specific items within a commit can be backed out by file.

## Screenshots / videos

Not applicable — no UI changes. Perf numbers captured above in "How did I test?" stand in for visual proof.


apps/tlon-desktop/src/main/triggers.ts
```diff
@@ -119,4 +119,57 @@ BEGIN
...
+-- by channel and group entries) don't fire hundreds of __change_log events
+-- and block the main thread via processChanges callbacks. Channel/group
+-- volume reactivity is handled through the existing table-deps invalidation
+-- path for channel/group queries.
+--
+-- DROP first so upgrading from an earlier version of this branch (which had
+-- ungated triggers) replaces them.
+DROP TRIGGER IF EXISTS after_volume_settings_insert;
+DROP TRIGGER IF EXISTS after_volume_settings_update;
+DROP TRIGGER IF EXISTS after_volume_settings_delete;
```
davidisaaclee: nit: we're going to re-add this trigger on every connection. I think we only have one connection per session, so maybe not an issue, but we could consider adding these in a migration instead.

apps/tlon-web/src/app.tsx
```diff
@@ -455,8 +456,13 @@ function ConnectedWebApp() {
...
       if (!hasSyncedRef.current) {
-        // Web doesn't persist database, so headsSyncedAt is misleading
-        await db.headsSyncedAt.resetValue();
+        // NOTE: we previously reset headsSyncedAt here on the assumption that
+        // web doesn't persist the database. With ENABLE_DB_FILE_LOAD (see
+        // webDb.ts) the DB is actually persisted via sqliteContent, so the
+        // reset is harmful — it forces syncLatestPosts to refetch one head
+        // per channel on every load, producing a 300-channel
+        // insertLatestPosts batch whose setLastPosts spends tens of seconds
+        // recomputing lastPost* columns for channels that haven't changed.
```
davidisaaclee: O_O wow, good catch

davidisaaclee: should we reset `headsSyncedAt` on webDb setup? we'd know if we are loading a persisted db from there.

packages/app/lib/triggers.ts
```diff
@@ -119,4 +119,57 @@ BEGIN
         json_object('channel_id', OLD.channel_id, 'thread_id', OLD.thread_id)
     );
 END;
+
+-- Create triggers for volume_settings table. Used to drive reactive updates
```
davidisaaclee, line 123: why are there two definitions of these triggers? here and tlon-desktop/src/main/triggers.ts

packages/app/lib/webDb.ts
```diff
@@ -94,6 +93,47 @@ export class WebDb extends BaseDb {
...
       logger.log('sqlocal instance created', { sqlocal: this.sqlocal });
+
+      // Expose devtools helpers for diagnosing migration/schema state.
+      try {
+        const sqlocal = this.sqlocal;
+        const g: any = globalThis;
+        // __tlonRawSql runs arbitrary SQL against the user's local DB —
+        // gate behind __DEV__ so it only exists in dev builds.
+        if (__DEV__) {
+          g.__tlonRawSql = async (query: string) => {
```
davidisaaclee: nit: we should remove https://github.com/tloncorp/tlon-apps/blob/3ef1db08a4a196a1779fd32ebda14066d672d729/packages/shared/src/db/client.ts#L34-L38 so we only have one sql dev util

packages/app/lib/webDb.ts
```diff
@@ -192,12 +232,24 @@ export class WebDb extends BaseDb {
...
+  // insert of N rows would otherwise queue N separate callbacks, each
+  // paying a ~20 ms round-trip even after the first one drains the log.
+  // Bundling to a single trailing invocation per event-loop tick turns N
+  // callbacks into one.
+  private pendingProcessChanges: ReturnType<typeof setTimeout> | null = null;
 
-    if (ENABLE_DB_FILE_SAVE) {
-      this.saveToFile();
-    }
+  override async processChanges() {
```
davidisaaclee: This method now lies about completion – someone doing `await processChanges()` will proceed to the next statement without the processing actually completing.

If there's nothing relying on blocking for this, we should just change it to a non-async method, possibly named something like `enqueueProcessChanges`.

packages/shared/src/db/schema.ts
```diff
@@ -1156,6 +1156,24 @@ export const posts = sqliteTable(
     groupId: index('posts_group_id').on(table.groupId, table.id),
     authorIdIndex: index('posts_author_id_index').on(table.authorId),
     parentIdIndex: index('posts_parent_id_index').on(table.parentId),
+    // Partial index over outstanding optimistic top-level writes.
+    // Replies also carry sequence_number = 0 (see toPostReplyData), so the
+    // parent_id IS NULL clause keeps the index tiny — only rows that could
+    // actually be replaced by an incoming top-level post.
+    cachedPostsIndex: index('cached_posts_index')
```
davidisaaclee: nit: why shouldn't this match the index naming convention? `posts_*`

davidisaaclee:
from the PR description:

> Breaks the every-reload retry loop caused by reset-migrations rotating the tag on every schema change.

I don't get what caused this "every-reload retry loop." Can you explain?

arthyn:
@davidisaaclee I think it was speculating on something there. The app wouldn't get past the DB load sometime in the middle of working on this and it proposed this as fix. I can rip it out.

davidisaaclee (changes requested):
Here's a partial review – sorry for the slow review, this is a dense PR, especially since it's sometimes hard to read your agent's intention in the code.

Heads up on the failing check – I can't tell if it's a real failure.

packages/app/lib/webDb.ts
arthyn: bah missed this

apps/tlon-desktop/src/main/triggers.ts
arthyn: this is in tlon-desktop so I don't think we need to worry about it rn, I don't think anyone is using this. could just remove potentially

davidisaaclee: we're also re-adding the ones in [packages/app/lib/triggers.ts](https://github.com/tloncorp/tlon-apps/pull/5748/files/056c4b9b8e15d1394dddd24739b64a7819496f56#diff-c1bc4b9f1248a20b1d2762e6fc240b5acbad9a710494e118d34bdc6143ea1018) on each connection

apps/tlon-web/src/app.tsx
```diff
@@ -454,15 +454,10 @@ function ConnectedWebApp() {
     });
 
     const syncStart = async () => {
-      // Only call sync.syncStart once during the app's lifecycle
+      // Only call sync.syncStart once during the app's lifecycle.
+      // headsSyncedAt is reset by webDb.setupDb when the persisted DB doesn't
+      // load (fresh start), so we don't reset it here on every reload.
```
davidisaaclee, line 458: nit: I think this would just be confusing to someone reading this code fresh. would someone read this code and think "why aren't we resetting `headsSyncedAt`?"

packages/app/lib/webDb.ts
```diff
@@ -272,6 +294,8 @@ export class WebDb extends BaseDb {
     logger.log('purging sqlite database');
     this.saveToFile.cancel();
     await sqliteContent.resetValue();
+    await headsSyncedAt.resetValue();
+    await changesSyncedAt.resetValue();
```
davidisaaclee, line 290: nit: feels leaky, why aren't we resetting the entire kv store? (I'm guessing the answer is that there are entries we want to retain across DB sessions – in which case we should probably group these "purge-able" entries and reset all on DB reset, which would be out-of-scope.)

packages/shared/src/db/queries.ts
```diff
@@ -3198,18 +3199,43 @@ export const insertChanges = createWriteQuery(
...
+
+        const stopContacts = perfMark('insertChanges.contacts');
         await insertContacts(input.contacts, txCtx);
+        stopContacts({ count: input.contacts.length });
+
+        const stopDelChan = perfMark('insertChanges.deleteChannels');
         await deleteChannels(input.deletedChannelIds, txCtx);
+        stopDelChan({ count: input.deletedChannelIds.length });
+
+        const stopGroupUnreads = perfMark('insertChanges.groupUnreads');
```
davidisaaclee: nit (that has a large effect across codebase): as a human reader, this code is hard to read – `stopDelChan` reads like "stop deleting a channel", not stop the `perfMark` controller. I think these will also all leak (never stopping the `perfMark`) if any of the `await`s throw.

Recommendations:
- easier to read: either use a prefix on the identifiers for `perfMark` (`pm_stopTxn` - ugly but would be clear), or change the `perfMark` API to a callback (`await perfMark('insertChanges.groupUnreads', async () => { await insertGroupUnreads(...) })`)
- preventing leak: using the callback API I mentioned gives a way to `finally` stop the timers or mark them as errored. otherwise, I guess you could (1) create all the perfMarks up front, then cancel them in `catch`; or (2) change `perfMark` API to be able to spawn and cancel all children (i.e. `perfMark.fork: (label: string) => typeof perfMark` + `perfMark.cancelChildren: () => void`)

packages/shared/src/db/queries.ts
```diff
@@ -3298,61 +3328,78 @@ async function insertPostsBatch(posts: Post[], ctx: QueryCtx) {
...
-    inArray(
-      $posts.id,
-      receivedPosts.map((p) => p.id)
+async function deleteReplacedCachedPosts(replacers: Post[], ctx: QueryCtx) {
+  if (!replacers.length) return;
+  // Delete outstanding optimistic top-level rows whose (channelId, sentAt,
+  // authorId) tuple matches an incoming real post.
+  //
+  // Replies also have sequence_number = 0 (see toPostReplyData), so we also
+  // require parent_id IS NULL — keeps the lookup scoped to the cached_posts
```
davidisaaclee: query's no longer being bound to `cached_posts`, so this comment is out-of-date

packages/shared/src/db/queries.ts
```diff
@@ -3784,7 +3910,11 @@ export const getPostWithRelations = createReadQuery(
       })
       .then(returnNullIfUndefined);
   },
-  ['posts', 'postReactions', 'threadUnreads', 'volumeSettings']
+  // Invalidation for ['post', id] queries is driven by changeListener's
+  // per-post events (see db/changeListener.ts). The broad table-level
+  // invalidation path would re-stale every cached per-post entry on any
+  // posts-table mutation; the targeted path only touches the specific post.
+  []
```
davidisaaclee, line 4192: I don't feel good about this change, but I can't think of any specific stale data bugs it'd cause, apart from the AI code review comment on this line about `insertChannelPosts`

apps/tlon-desktop/src/main/triggers.ts
arthyn: sorry yeah this was unnecessary since it was trying to protect from things done in this branch. I committed some claude.md language to hopefully help with this in the future.

packages/shared/src/db/queries.ts
```diff
@@ -3493,6 +3551,74 @@ async function setLastPosts(newPosts: Post[] | null, ctx: QueryCtx) {
     );
 }
 
+async function setLastPostsMonotonic(newPosts: Post[], ctx: QueryCtx) {
```
davidisaaclee, line 3572: nit: I bet we could extend `setLastPostsMonotonic` to handle multiple channels and use it instead of the subquery-based implementation, which would turn 2 chunky procedures into 1.

davidisaaclee (approved):

PR #5750, opened 2026-04-21
chat: TLON-5606 split UI-pending and delivery-pending message handling (3/3)

patosullivan:
## Summary

Builds on the TLON-5606 core fix (PR #5746) and the PR #5747 cleanup by separating UI-pending semantics from delivery-pending semantics.

This PR is the query / merge / sync follow-up:

- chat-style views should stop resurfacing a cleared local-only failed row as a ghost
- delivery polling should still continue for rows that are deleted locally but are still in flight
- catch-up-window rows that are already server-backed should remain representable as tombstones until reconciliation finishes

This PR is intended to be reviewed stacked on PR #5747. The main review focus here is the pending-query, merge, and sync behavior. It does **not** introduce additional reply-summary repair or channel/group preview repair logic beyond the already-stacked PR #5747 base.

## Root cause

A single `getPendingPosts()` contract was trying to serve two different consumers:

- the UI merge layer, which decides what can still be rendered as a pending/tombstone row
- the delivery backoff loop, which decides whether the client still needs to poll for reconciliation

Those consumers do not want the same rows.

- a deleted `failed + sequenceNum === 0` row is a local-only terminal state and should disappear from the UI pending layer
- a deleted `enqueued` / `pending` row is still in flight and should keep delivery polling alive
- a deleted `sent + sequenceNum === 0` row is already server-backed but still waiting for the sequenced `addPost`, so it should remain representable as a tombstone during the catch-up window
- `needs_verification` rows may also be server-backed and cannot be treated like a purely local failed send

Without splitting those contracts, fixing one side risks regressing the other.

## Changes

All changes are in `packages/shared`:

- Narrowed `getPendingPosts()` in [packages/shared/src/db/queries.ts](/Users/patrick/workspace/homestead/packages/shared/src/db/queries.ts) so only truly local-only deleted failed sends drop out of the UI pending layer.
- Added `getDeliveryPendingPosts()` in [packages/shared/src/db/queries.ts](/Users/patrick/workspace/homestead/packages/shared/src/db/queries.ts) for `syncChannelWithBackoff()`.
  - includes `enqueued` / `pending` rows regardless of `isDeleted`
  - includes the `sent + sequenceNum === 0` catch-up window
  - excludes terminal-failed rows
- Updated `mergePendingPosts()` in [packages/shared/src/store/useMergePendingPosts.ts](/Users/patrick/workspace/homestead/packages/shared/src/store/useMergePendingPosts.ts) so:
  - locally-cleared failed optimistic rows are still dropped
  - surviving rows flagged in `deletedPosts` get an `isDeleted` overlay for immediate tombstone rendering
  - confirmed tombstones and catch-up-window rows remain visible
- Updated `syncChannelWithBackoff()` in [packages/shared/src/store/sync/sync.ts](/Users/patrick/workspace/homestead/packages/shared/src/store/sync/sync.ts) to use `getDeliveryPendingPosts()` instead of `getPendingPosts()`
- Added supporting regression coverage in [packages/shared/src/store/postActions/postActions.test.ts](/Users/patrick/workspace/homestead/packages/shared/src/store/postActions/postActions.test.ts) for how the new pending/query semantics interact with `deletePost()` / `deleteFailedPost()`

## Explicitly Not In Scope

- hard-delete classification for truly unsent failed optimistic rows
- reply parent metadata repair
- channel / group preview recomputation
- changing the current `needs_verification` delete posture beyond pending/query behavior

## How did I test?

Ran:

- `pnpm --filter @tloncorp/shared test run src/db/queries.test.ts src/store/mergePendingPosts.test.ts src/store/sync/sync.test.ts src/store/postActions/postActions.test.ts`

Coverage for this area lives in:

- [packages/shared/src/db/queries.test.ts](/Users/patrick/workspace/homestead/packages/shared/src/db/queries.test.ts)
- [packages/shared/src/store/mergePendingPosts.test.ts](/Users/patrick/workspace/homestead/packages/shared/src/store/mergePendingPosts.test.ts)
- [packages/shared/src/store/sync/sync.test.ts](/Users/patrick/workspace/homestead/packages/shared/src/store/sync/sync.test.ts)
- [packages/shared/src/store/postActions/postActions.test.ts](/Users/patrick/workspace/homestead/packages/shared/src/store/postActions/postActions.test.ts)

The current test coverage focuses on:

- deleted failed rows stay out of both the UI-pending and delivery-pending paths
- deleted in-flight rows still keep delivery polling alive
- catch-up-window rows remain visible until reconciliation
- overlayed deleted rows render as tombstones without mutating the input objects
- delete entrypoints still behave correctly around `needs_verification`, failed-edit, and failed-delete rows under the narrowed query contract

## Risks and impact

  - [x] State / providers
  - [x] Message sync
  - [x] Channel display

Main caveat:

- This PR still preserves the current `needs_verification` posture. It improves the pending/query behavior around those rows, but it does not add verify-before-delete or a server delete handoff.

## Rollback plan

Revert.

### DIFF
--- a/packages/shared/src/db/queries.ts
+++ b/packages/shared/src/db/queries.ts
@@ -3750,12 +3750,36 @@ export const addReplyToPost = createWriteQuery(
 );
 
 /**
- * Undo one optimistic reply-summary bump after a failed reply row is removed.
- * If the parent row has already been replaced by server-authoritative
- * `replyMeta`, the optimistic bump count is 0 and this becomes a no-op.
- * Otherwise, recompute from local replies when the cache is complete; if the
- * cache is partial, only decrement `replyCount` and preserve server-sourced
- * summary fields.
+ * Undo a single optimistic reply bump on a parent post after the optimistic
+ * reply row has been hard-deleted.
+ *
+ * This is the inverse of the optimistic side of `addReplyToPost` (which
+ * increments `replyCount`, sets `replyTime`, and appends to
+ * `replyContactIds` when the caller does not provide server-sourced
+ * `replyMeta`).
+ *
+ * Server-authoritative `replyMeta` updates reset `optimisticReplyBumpCount`
+ * to 0. When that has already happened, this helper becomes a no-op: the
+ * parent's reply summary no longer includes the failed local reply.
+ *
+ * Otherwise there are two regimes:
+ *
+ * 1. **Local cache is complete** — the parent's `replyCount` equals the
+ *    number of local undeleted replies + 1 (the one we just deleted). In
+ *    that case the local rows are authoritative and we can recompute all
+ *    three fields from them, using `appendContactIdToReplies` so ordering
+ *    matches the incremental path.
+ *
+ * 2. **Local cache is partial** — the parent's `replyCount` is larger than
+ *    (local undeleted replies + 1). That means the parent carries
+ *    server-sourced `replyMeta` for replies we have not cached locally.
+ *    We decrement `replyCount` by 1 (the optimistic +1 we applied at send
+ *    time) and leave `replyTime` / `replyContactIds` untouched so we do
+ *    not collapse server-known thread summary to a partial local view.
+ *
+ * Caller must have already removed the deleted reply row from the `posts`
+ * table before invoking this helper; the "alive replies" count is taken
+ * from the current DB state.
  */
 export const undoOptimisticReplyBump = createWriteQuery(
   'undoOptimisticReplyBump',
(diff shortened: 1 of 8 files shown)


patosullivan:
@claude review once

packages/shared/src/store/useMergePendingPosts.ts
```diff
@@ -27,10 +27,58 @@ export const mergePendingPosts = ({
...
+      keys.push(`${post.channelId}:${post.authorId}:${post.sentAt}`);
+    }
+    return keys;
+  };
+  const deletedPostKeys = new Set<string>();
+  [...newPosts, ...pendingPosts, ...existingPosts].forEach((post) => {
+    if (!post.isDeleted && !deletedPosts[post.id]) {
+      return;
+    }
+    postMergeKeys(post).forEach((key) => deletedPostKeys.add(key));
```
arthyn:
```suggestion
    if (post.isDeleted || deletedPosts[post.id]) {
      postMergeKeys(post).forEach((key) => deletedPostKeys.add(key));
    }
```

packages/shared/src/store/useMergePendingPosts.ts
```diff
@@ -27,10 +27,58 @@ export const mergePendingPosts = ({
   hasNewest: boolean;
   filterDeleted?: boolean;
 }): db.Post[] => {
+  const postMergeKeys = (post: db.Post) => {
```
arthyn, line 30: do we really have instances of posts without authors?

packages/shared/src/db/queries.ts
```diff
@@ -3750,12 +3750,36 @@ export const addReplyToPost = createWriteQuery(
 );
 
 /**
- * Undo one optimistic reply-summary bump after a failed reply row is removed.
- * If the parent row has already been replaced by server-authoritative
- * `replyMeta`, the optimistic bump count is 0 and this becomes a no-op.
- * Otherwise, recompute from local replies when the cache is complete; if the
- * cache is partial, only decrement `replyCount` and preserve server-sourced
- * summary fields.
+ * Undo a single optimistic reply bump on a parent post after the optimistic
```
arthyn, line 3753: comments are getting wild

packages/shared/src/db/queries.ts
```diff
@@ -3926,7 +3950,59 @@ export const getPendingPosts = createReadQuery(
...
+        not(eq($posts.type, 'reply')),
+        // Exclude only truly local-only deleted sends. Deleted rows whose
+        // `deliveryStatus` is `'sent'` or `'needs_verification'` may still
+        // be server-backed but not yet sequenced; keep them as a DB-backed
+        // tombstone source so the channel remount path still renders them
+        // before the sequenced `addPost` arrives. Enqueued / pending rows
+        // are still in-flight and also remain visible — they either reach
+        // `'sent'` (stay as tombstone) or `'failed'` (vanish on next tick).
+        //
+        // SQL NULL handling: `isDeleted` is a nullable boolean. A bare
```
arthyn, line 3962: maybe we should consider making `isDeleted` not nullable

arthyn (approved):
lgtm, we should consider https://linear.app/tlon/issue/TLON-5718/channels-optimistic-overlay-schema-to-reduce-management-overhead-on

PR #5759, opened 2026-04-22
SplashSequence: retitle and pin home group with bot nickname

jamesacklin:
## Summary

Once the user sets a bot nickname in the splash sequence, the home group provisioned by hosting is retitled to `{botName}'s Group` and pinned client-side. This makes the bot nickname flow through to the group the same way it already does for the bot DM, and lets the bot-related group sit alongside the bot DM in the pinned section of the chat list.

## Changes

- `packages/api/src/client/wayfinding.ts`: added `HOME_GROUP_SLUG`, `getHomeGroupId`, and `generateHomeGroupTitle` helpers next to the existing `generatePersonalGroupTitle` convention.
- `packages/shared/src/db/keyValue.ts`: added `homeGroupId` storage item for caching the resolved home group ID.
- `packages/app/hooks/useBootSequence.ts`: during the `RESERVING` phase, after caching the invite link, resolve the lure token metadata via `getMetadataFromInviteToken` and cache `invitedGroupId` as `db.homeGroupId` so downstream wayfinding code doesn't depend on the hardcoded slug. Failure is logged but non-fatal.
- `packages/app/ui/components/Wayfinding/SplashSequence.tsx`: after `setTlawnNickname` (and the optional avatar sync) in `persistBotIdentityInBackground`, call a new `applyHomeGroupBotIdentity` helper that loads the home group from the local db and pokes `%groups` via `api.updateGroupMeta` to update the title and `api.pinItem` to pin it. Uses a new `resolveHomeGroupId` helper that reads `db.homeGroupId` (with a lazy re-resolve from the cached invite link, falling back to `getHomeGroupId`). Goes through `api` directly rather than `store.updateGroupMeta` / `store.pinGroup` because those swallow poke failures and roll back optimistically; we want errors to propagate so `withRetry` actually retries transient failures and the Wayfinding analytics reflect real server state. Wrapped in `withRetry` to cover the window where the home group hasn't synced into the local db yet, and logs `Wayfinding Home Group Sync Succeeded/Failed` analytics events.
- `packages/app/ui/components/Wayfinding/useHomeGroupInviteLink.ts`: use the cached `db.homeGroupId` via the hook variant with fallback to `getHomeGroupId`; dropped the locally-duplicated `HOME_GROUP_SLUG` constant and `getHomeGroupId` function in favor of the canonical import from `@tloncorp/shared/logic`.
- `apps/tlon-mobile/src/hooks/analytics.ts`: read the cached `db.homeGroupId` with fallback to `getHomeGroupId(currentUserId)` when looking up the home group for the analytics digest; dropped the locally-duplicated `HOME_GROUP_SLUG` constant in favor of the canonical import.

## How did I test?

- `pnpm --filter @tloncorp/app tsc --noEmit`, `pnpm --filter @tloncorp/api tsc --noEmit`, `pnpm --filter @tloncorp/shared tsc --noEmit`, and `pnpm --filter tlon-mobile tsc --noEmit` all pass.
- Have not exercised the splash sequence end-to-end against a freshly provisioned hosted ship.

## Risks and impact

  - [x] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert this commit. The splash sequence will go back to leaving the home group with its hosting-provisioned title and unpinned.

### DIFF
--- a/apps/tlon-mobile/src/hooks/analytics.ts
+++ b/apps/tlon-mobile/src/hooks/analytics.ts
@@ -2,6 +2,7 @@ import * as api from '@tloncorp/api';
 import { useLureMetadata } from '@tloncorp/app/contexts/branch';
 import { AnalyticsEvent, createDevLogger } from '@tloncorp/shared';
 import * as db from '@tloncorp/shared/db';
+import { getHomeGroupId } from '@tloncorp/shared/logic';
 import * as Contacts from 'expo-contacts';
 import * as Notifications from 'expo-notifications';
 import { useEffect } from 'react';
@@ -9,7 +10,6 @@ import { useEffect } from 'react';
 import { checkLatestVersion } from '../lib/lifecycleEvents';
 
 const logger = createDevLogger('analytics hooks', true);
-const HOME_GROUP_SLUG = 'home-group';
 
 export function useCheckAppInstalled() {
   const lureMeta = useLureMetadata();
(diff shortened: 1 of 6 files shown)


jamesacklin:
@copilot review

jamesacklin:
@claude review once

PR #5760, opened 2026-04-22
Onboarding: pin bot home group & set nickname derived fields during Splash

latter-bolden:
OTT


jamesacklin (approved):
Much cleaner than mine!

packages/api/src/types/wayfinding.ts
```diff
@@ -5,6 +5,11 @@ export const PersonalGroupSlugs = {
   notebookSlug: 'tm-wayfinding-group-notebook',
 };
 
+export const BotHomeGroupSlugs = {
```
jamesacklin, line 8: Should probably make this dynamic later down the line

PR #5764, opened 2026-04-22
marks: patch up

Fang-:
## Summary

Many marks didn't compile into proper mark cores. Here, we make a bunch of changes to correct this.

These changes are necessary for #5668 to be able to incorporate all marks properly.

## Changes

Mostly straightforward changes: making sure we're in type parsing mode, specifying a `+grad`, making sure we can `+noun:grab`, etc.

Special review attention request for the following:
- [x] Do `/mar/chat/writs.hoon` and `/mar/group/invites.hoon` pull in the right type versions?
- [x] What's up with `/mar/group/create.hoon`? The json conversion util it tried using doesn't seem to exist at all...
- [x] `/mar/writ-diff.hoon` aka `%writ-diff` seems to be completely unused. I think it's a very old mark. Should we just remove it?

## How did I test?

Compiled all the marks for #5668.

## Risks and impact

- Technically safe to rollback without consulting PR author.
  - Message sync
  - Channel display

## Rollback plan

Revert.


desk/mar/group/create.hoon
```diff
@@ -9,6 +9,6 @@
 ++  grab
```
mikolajpp: This mark is a leftover from a time when there was a separate poke for group creation. Safe to remove!

desk/mar/group/invites.hoon
```diff
@@ -1,15 +1,15 @@
 /-  gv=groups-ver
```
mikolajpp: This appears unused. A grep for use of this mark does not turn up anything.

desk/mar/group/invites-1.hoon
```diff
@@ -1,14 +1,14 @@
 /-  gv=groups-ver
```
mikolajpp: This appears unused. A grep for this mark turns up nothing.

desk/mar/group/preview-update.hoon
```diff
@@ -1,12 +1,12 @@
 /-  gv=groups-ver
```
mikolajpp: Another vestigal mark.

mikolajpp (approved):
Good cleanup! A few group marks are remnants of old interfaces and could be removed altogether.

Edit: something's up with e2e tests though.

desk/mar/chat/writs.hoon
```diff
@@ -5,7 +5,7 @@
 ++  grow
```
mikolajpp: Based on the code, this mark is only ever use in a very old code from old groups. See `+summarize-activity-old-groups` in `/lib/summarize.hoon`. As such, the version is almost certainly wrong: the types here were current a long time ago.

We should remove this mark, while also removing the code path for old groups from `/lib/summarize.hoon`. Chat has since moved on to returning `chat-paged-writs` data.

desk/mar/writ-diff.hoon
```diff
@@ -1,14 +1,14 @@
-/-  c=chat
```
mikolajpp, line 1: Grep turns up nothing, should be safe to remove.

mikolajpp:
The e2e test fail here seems to be a client bug, but since we touch a whole bunch of marks let's wait until that gets fixed and merge develop to have a clean pass.

PR #5768, opened 2026-04-23
onboarding: handle auto navigation when system permissions resolve

latter-bolden:
## Summary

Handles navigation automatically on `SplashSequence` contacts pane and `AllowNotificationsScreen` once the system permission is confirmed granted/denied.

It will also auto-navigate if you land on the screen with your permission already set, whereas previously you'd still have to tap a button before we'd detect that perms were configured. We could pull that logic a screen forward and prevent ever showing the permission prompt screen/pane, but I think introducing that at this stage would be riskier and no new user should ever encounter it during an actual signup.

## How did I test?

Onboarded with/without perms pre-configured in iOS Simulator

## Risks and impact

  - [x] Onboarding

Fixes TLON-5670

### DIFF
--- a/packages/app/ui/components/Wayfinding/SplashSequence.tsx
+++ b/packages/app/ui/components/Wayfinding/SplashSequence.tsx
@@ -1503,8 +1503,8 @@ export function GroupsPane(props: {
           testID="got-it"
           onPress={props.onActionPress}
           label="Got it"
-          preset="secondary"
-          fill="text"
+          preset={props.hostingBotEnabled ? 'secondary' : 'hero'}
+          fill={props.hostingBotEnabled ? 'text' : undefined}
         />
       </YStack>
     </View>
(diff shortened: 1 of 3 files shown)


latter-bolden:
@claude review once

packages/app/ui/components/Wayfinding/SplashSequence.tsx
```diff
@@ -1503,8 +1503,8 @@ export function GroupsPane(props: {
           testID="got-it"
           onPress={props.onActionPress}
           label="Got it"
-          preset="secondary"
-          fill="text"
+          preset={props.hostingBotEnabled ? 'secondary' : 'hero'}
+          fill={props.hostingBotEnabled ? 'text' : undefined}
```
jamesacklin:
```suggestion
          backgroundColor={props.hostingBotEnabled ? '$transparent' : undefined}
```

jamesacklin: `fill='text'` is the enemy of all generous tap-targets, will have to throw a note in CLAUDE.md to use the secondary/transparent treatment rather than fill

jamesacklin (approved):
Approved with edits

PR #5769, opened 2026-04-23
reel: fix mark conversion

arthyn:
## Summary

This PR https://github.com/tloncorp/tlon-apps/pull/5692 introduced a bug with converting reel describes from JSON. It likely passed in aqua because it was being poked as a noun which is not affected.

## Changes

- limit field type conversion so it actually nests

## How did I test?

Create a new group see invite link generate immediately

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

### DIFF
--- a/desk/lib/reel.hoon
+++ b/desk/lib/reel.hoon
@@ -13,11 +13,36 @@
   :~  ['tag' s+tag.metadata]
       ['fields' (pairs:enjs:format fields)]
   ==
-++  dejs-metadata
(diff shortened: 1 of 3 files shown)


Fang- (approved):
This should give a narrowing (`$field`-nesting) product type on the json parser, yes.

PR #5773, opened 2026-04-23
 fix chat list swipe actions disappearing after release

patosullivan:
## Summary

Fixes TLON-5672: on native (iOS and Android), swipe actions on the chat list rows become unusable once the row settles open. Removes an app-level swipe-direction guard in `InteractableChatListItem` that was hiding the action panel the user had just revealed.

Same defect as TLON-5491 on the Expo-54 branch; this mirrors the fix in `adba45ea3` on `origin/janic/expo-54`.

## Changes

Single file: `packages/app/ui/components/listItems/InteractableChatListItem.tsx` (+3 / −20).

- Removed `currentSwipeDirection` state and its setters.
- Removed `onSwipeableWillOpen` / `onSwipeableClose` callbacks and their props on `<Swipeable>`.
- `renderLeftActions`: dropped the `currentSwipeDirection === 'right'` branch; kept the `!hasUnread` branch (unrelated).
- `renderRightActions`: dropped the `currentSwipeDirection === 'left'` branch.

### Why

`InteractableChatListItem` stored `onSwipeableWillOpen(direction)` in state and used it to suppress one side's action renderer. In `react-native-gesture-handler` 2.24.0 (`ReanimatedSwipeable.tsx:342-345`), that `direction` is the **gesture direction**, not the side being opened. So a right swipe opens `LeftActions` and simultaneously sets the flag that blanks `LeftActions` — the component was removing the panel it had just revealed. Symmetric on the other side.

The guard was added in `22c5d3d27` against an older RNGH contract where the callback reported which side was opening. It became user-visible when RNGH was pinned to 2.24.0 in `95eb41014` (mobile lightbox work). Removing it is the minimum change that restores correct behavior.

## How did I test?

Tested on the iOS Simulator and Android Emulator.

## Risks and impact

- Affects important code area: ChatList

## Rollback plan

Revert

## Screenshots / videos

[android emulator 4-23-2026 .webm](https://github.com/user-attachments/assets/42224aca-a799-474f-b60b-9ae812b6833c)

https://github.com/user-attachments/assets/68503481-72bb-429c-8f8a-f7f58eb983a2

### DIFF
--- a/packages/app/ui/components/listItems/InteractableChatListItem.tsx
+++ b/packages/app/ui/components/listItems/InteractableChatListItem.tsx
@@ -33,9 +33,6 @@ function BaseInteractableChatRow({
   ...props
 }: ListItemProps<db.Chat> & { onLayout?: (e: any) => void }) {
   const swipeableRef = useRef<SwipeableMethods>(null);
-  const [currentSwipeDirection, setCurrentSwipeDirection] = useState<
-    'left' | 'right' | null
-  >(null);
 
   const isMuted = useMemo(() => {
     return logic.isMuted(model.volumeSettings?.level, model.type);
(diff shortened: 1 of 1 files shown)


patosullivan:
@claude review once.

jamesacklin (approved):
Code / root cause makes sense to me, wait for another reviewer before merging.

latter-bolden (approved):

PR #5779, opened 2026-04-23
onboarding: fix nickname handling

latter-bolden:
- fix the bug where retries would fire immediately instead of delaying as expected
- set home group title whenever user nickname is updated if still has default title (same as old style personal group)
- check signup context for nickname in addition to pulling from local db contact

### DIFF
--- a/packages/app/ui/components/Wayfinding/SplashSequence.tsx
+++ b/packages/app/ui/components/Wayfinding/SplashSequence.tsx
@@ -90,6 +90,15 @@ enum SplashPane {
 
 type CustomBotSetupStatus = 'idle' | 'pending' | 'ready' | 'failed';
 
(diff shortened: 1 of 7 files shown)


packages/app/ui/components/Wayfinding/SplashSequence.tsx
```diff
@@ -893,8 +908,8 @@ export function BotNamePane(props: {
                 autoFocus
                 placeholder={
                   props.userNickname
-                    ? `${props.userNickname}'s Tlonbot`
-                    : 'Tlonbot'
+                    ? getPreviewBotName(props.userNickname)
+                    : 'TlonBot'
```
jamesacklin:
```suggestion
                    : 'Tlonbot'
```

jamesacklin (changes requested):

PR #5782, opened 2026-04-24
web: fix reanimated prod errors

arthyn:
## Summary

On web we're getting hammered with reanimated errors in the console, easily a few hundred just navigating a channel. I've seen the count 10k+ over long sessions

  This PR fixes production-only web crashes and warning floods coming from
  Reanimated when opening channels or switching sidebars. It adds a durable patch
  for react-native-reanimated@3.16.7, updates the lockfile to use it, and makes the
  web build resolve Reanimated through the patched top-level package so nested
  dependency copies do not reintroduce the old web updater path.

  ## Changes

  - Added a pnpm patch for react-native-reanimated@3.16.7
  - Updated package.json and pnpm-lock.yaml to apply the patch
  - Patched Reanimated’s web JS updater to:
      - guard missing refs and missing props
      - unwrap wrapper refs more defensively
      - update DOM style targets directly on web
      - fall back safely when transform serialization helpers are unavailable
  - Updated apps/tlon-web/vite.config.mts to:
      - share the Urbit proxy config with vite preview
      - alias react-native-reanimated to the top-level package so drawer/sidebar
        bundles do not pull a stale nested copy

  ## How did I test?

  - Built production web successfully with:
      - pnpm --filter tlon-web exec vite build
  - Verified the emitted production bundles no longer contained the old crashing
    updater path
  - Restored the nested @react-navigation/drawer copy of Reanimated to stock
    locally, rebuilt, and confirmed the app still bundled against the patched top-
    level package
  - Confirmed the drawer/sidebar bundle stopped inlining its own old Reanimated
    updater path

  ## Risks and impact

      - [ ] Onboarding
      - [ ] State / providers
      - [ ] Message sync
      - [x] Channel display
      - [ ] Notifications
      - [x] Other: Web build/runtime dependency resolution and prod preview proxying

  ## Rollback plan

  Revert this PR. That removes the Reanimated patch registration, lockfile
  references, and the Vite alias/proxy changes, returning web production behavior to
  the previous state.

  ## Screenshots / videos

  None.


wca4a:
@claude review once

wca4a:
@codex review

wca4a:
@codex review

apps/tlon-web/vite.config.mts
```diff
@@ -266,6 +281,9 @@ export default ({ mode }: { mode: string }) => {
       dedupe: ['@tanstack/react-query'],
       alias: {
         '@': fileURLToPath(new URL('./src', import.meta.url)),
+        'react-native-reanimated': fileURLToPath(
```
dnbrwstr, line 290: Would it make sense to do this as a top level version override in `package.json` for consistency?

dnbrwstr (changes requested):
Overall looks good to me. Bot comments seem valid. Only blocker is I'd add an entry in the patches `README.md` file -- trying to track these better.

arthyn:
@dnbrwstr I think this should be ready again

dnbrwstr (approved):
Changes all look good to me, just need to bump to apply to new reanimated version.

PR #5785, opened 2026-04-24
chore(deps): bump tamagui rc.22 -> rc.41

janicduplessis:
## Summary

Bumps tamagui from `2.0.0-rc.22` to `2.0.0-rc.41` across all workspace packages and drops the two tamagui patches that applied to rc.22. Targets the same `janic/expo-54` branch as #5784.

## Changes

Version spec bumped in `apps/tlon-mobile`, `apps/tlon-web`, `packages/ui`, and `packages/app`. Both tamagui patches removed:

- `@tamagui/sheet` — the patch supplied no-op defaults for `setHasScrollView` / `scrollBridge.onParentDragging` so the `SheetHandle` effect didn't crash when `useSheetContext` returned the default object. rc.41 added `if (!context.scrollBridge) return` in `createSheet.mjs`, so the patch is no longer needed.
- `@tamagui/web` — the patch injected `unset.fontFamily` from the default font inside `createTamagui`. The `configIn.unset` concept has been removed from `@tamagui/web` in rc.41 (only `value === "unset"` CSS-value check remains), so the patch targets code that no longer exists.

Lockfile regenerated; only rc.41 remains for both direct and transitive dependencies (109 `@tamagui/*` packages + root `tamagui`, all at rc.41).

## How did I test?

- `pnpm install` clean (0 `rc.22` references remaining in `pnpm-lock.yaml`).
- Confirmed no duplicate tamagui versions in `node_modules` (`find node_modules -name package.json -path "*@tamagui/*"` all report rc.41; no nested installs).
- Visually sanity-check needed — please exercise sheets (New group, DMs, attachment sheet, action sheets), dialogs, and any flow with `defaultFont` theming to confirm both removed patches are truly obsolete.

## Risks and impact

- Safe to rollback without consulting PR author? Yes — revert the commit.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: Tamagui core (sheet / dialog / font / theming behavior across the app)

If either removed patch turns out to still be needed, regenerate it against rc.41 rather than re-adding the rc.22 file.

## Rollback plan

`git revert` this commit and `pnpm install` to restore rc.22 + the two patches.

## Screenshots / videos

n/a


dnbrwstr (approved):
Lgtm assuming the e2e failure is unrelated.

janicduplessis:
Wanted to see if it fixed another layout bug I was working on, although it didn’t though would be good to keep since we can remove some patches.

will double check ci before merging

PR #5788, opened 2026-04-27
podfile: patch for xcode 26.4

latter-bolden:
OTT. I updated to Xcode 26.4 and there's an RN bug that breaks local builds. See [this issue](https://github.com/facebook/react-native/issues/55601) for context.

Applies the necessary patch to the Podfile if you're running that xcode version. Confirmed it fixes my setup locally.

### DIFF
--- a/apps/tlon-mobile/ios/Podfile
+++ b/apps/tlon-mobile/ios/Podfile
@@ -118,6 +118,8 @@ post_install do |installer|
     :ccache_enabled => podfile_properties['apple.ccacheEnabled'] == 'true',
   )
 
+  # Work around Xcode 26.4 fmt consteval build failures until React Native updates fmt.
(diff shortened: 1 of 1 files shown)


apps/tlon-mobile/ios/Podfile
```diff
@@ -83,6 +118,8 @@ post_install do |installer|
     :ccache_enabled => podfile_properties['apple.ccacheEnabled'] == 'true',
   )
 
+  patch_fmt_consteval_for_xcode_26_4(installer)
```
patosullivan, line 123: This seems fine, but maybe we should add a comment explaining why we need it

patosullivan (approved):
Looks good, just one nit about adding an explanatory comment

PR #5790, opened 2026-04-27
fix(rn): enable Yoga WebFlexBasis to recompute flex basis across layouts

janicduplessis:
## Summary

Enables Yoga's experimental `WebFlexBasis` flag in `YogaLayoutableShadowNode::initializeYogaConfig`. Fixes a long-standing Yoga bug where flex children stay sized to their initial intrinsic content size after the parent's main-axis size becomes definite.

## Root cause

In Yoga's `computeFlexBasisForChild` (`ReactCommon/yoga/yoga/algorithm/CalculateLayout.cpp:97`), `computedFlexBasis` is set once and never invalidated unless the `WebFlexBasis` experimental flag is on. When a parent transitions from auto-height to definite height, the cached basis from the unconstrained pass is used and flex distribution allocates the wrong size.

`WebFlexBasis` adds a `computedFlexBasisGeneration != generationCount` invalidation, restoring CSS-spec behavior. It exists in upstream Yoga but is gated behind an experimental flag because rolling it out broke Meta-internal apps (open issue: facebook/yoga#1552).

## Changes

- `patches/react-native@0.81.5.patch` — 2-line addition that calls `YGConfigSetExperimentalFeatureEnabled(&config, YGExperimentalFeatureWebFlexBasis, true)`.
- `patches/README.md` — new section documenting why and when to remove.
- `pnpm-lock.yaml` — large but unavoidable: 2028/2032 changed lines are `patch_hash=...` rotations because the patch content changed.

## How did I test?

- Built a minimal Expo dev build with RN 0.81.5 from source. Verified the bug repros with both `flex:1 ScrollView` and plain `flex:1 View` variants.
- Yoga `os_log` instrumentation confirmed `computedFlexBasis` carrying over from frame 1 to frame 2 as the cause.
- With the patch applied: layout numbers match CSS-spec flex distribution, footer visible.
- Patch also resolves the original `@gorhom/bottom-sheet` symptom without any gorhom-side workaround.

## Risks and impact

The flag's effect is narrow: it only changes one branch of `computeFlexBasisForChild`, gating the case where `resolvedFlexBasis` is defined AND the parent's main-axis size is defined. The `singleFlexChild` short-circuit (the common `flex: 1`-only-child case) bypasses my changed path entirely. Other branches (auto-basis, explicit width/height) are unaffected.

Known limitation: the flag's dirty test uses generation-count equality, so a node measured twice within a single layout generation under different constraints would still keep the first basis. Our cross-generation bug is not affected, but a same-generation regression is theoretically possible. QA areas worth a focused look: animated sheets, keyboard-resize behavior, screen rotation, navigation transitions where a screen container's size changes.

- Safe to rollback without consulting PR author? Yes — revert the patch file diff and `pnpm install`.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: low-level RN layout engine, affects all views

## Rollback plan

`git revert <this-commit>` then `pnpm install`.


janicduplessis:
Closing — turning on Yoga's `WebFlexBasis` flag from this patch requires building React Native from source (`ios.buildReactNativeFromSource: "true"` in `Podfile.properties.json`), since the C++ change has to land in the compiled framework. We currently consume the prebuilt React-Core framework, so this patch sits in `node_modules` but never makes it into the actual binary.

There's no userland (JS or native-module) API to flip this flag — the only call site of `YGConfigSetExperimentalFeatureEnabled` in RN is the one this PR patches.

Replacing this with a narrower `@gorhom/bottom-sheet` patch that fixes the visible symptom in our app without changing the underlying RN binary. Will link the new PR here.

Keeping the branch and the writeup so we can revisit if/when we move to RN-from-source builds, or once the Yoga flag is flipped on by default upstream (facebook/yoga#1552).

janicduplessis:
Replacement gorhom-side patch: #5794

PR #5793, opened 2026-04-28
Fix iOS channel sorting with duplicate nav entries

patosullivan:
## Summary

Fixes channel-sort drag and drop on iOS in Expo 54. The reported symptom (drag does not activate) traces to two iOS-only integration bugs in `react-native-sortables` consumers, plus an underlying sync-handler bug that persists the data shape under which the drag bug manifests. All three are addressed. Closes TLON-5656.

In my testing I realized that the only groups in which drag and drop did not work at all on iOS (and the drag gestures would only cause the ScrollView to scroll) were ones in which *duplicate channels in multiple sections existed*.

## Changes

**Drag activation (iOS).** `react-native-sortables` only flips its internal `usesAbsoluteLayout` flag — the gate that enables drag — after every item in `Sortable.Grid`'s data array fires `onLayout`. When the data array contains duplicate `channelId`s across nav sections, React mounts only one component per duplicate key and the gate is never satisfied. iOS hits this; Android tolerates the same data.

The fix is read-side dedupe + a save gate + a deterministic repair trigger:

-   `packages/app/hooks/useSortableChannelNav.ts`: dedupe `sortableNavItems` to first-occurrence-by-id so `Sortable.Grid` always sees unique keys. Also expose a `duplicateChannelIds` set, and refuse to persist a reorder via `handleActiveItemDropped` while it's non-empty. Without that gate, the deduped first-occurrence ordering would silently normalize the user's data toward an arbitrary section pick — the frontend has no way to know which duplicate section is canonical.
-   `packages/app/hooks/useSortableChannelNav.helpers.ts` (new): pure helpers `findDuplicateChannelIds` and `shouldFireDuplicatesCallback`. Kept in a separate file so vitest can unit-test them without dragging the hook's transitive runtime imports through vite's transform.
-   `packages/app/hooks/useDuplicateMembershipRepair.ts` (new): shared hook returning an `onDuplicatesDetected` callback. When the gate trips, this fires `store.syncGroup(groupId, undefined, { force: true })` once per distinct duplicate set (fingerprint-deduplicated inside `useChannelOrdering`'s effect; `syncGroup`'s `groupSyncsInProgress` guard coalesces concurrent calls). The forced sync runs through the full-payload write path in `db.insertGroups` (see "Underlying data corruption" below) and reconciles local memberships against the backend's view, so the gate clears and reorders save normally on the next render.
-   `ManageChannelsScreenView.tsx` (native) and `ManageChannelsScreenView.web.tsx` (web/desktop): both call `useDuplicateMembershipRepair(group?.id)` and pass it as `onDuplicatesDetected` to `useChannelOrdering`. Same behavior on every platform that uses the shared hook.

**Auto-scroll while dragging (iOS).** The library's `AutoScrollProvider` only mounts when consumers pass an `AnimatedRef<Animated.ScrollView>` as `scrollableRef` to `Sortable.Grid`. The previous integration didn't pass one, so `AutoScrollProvider` was never mounted and dragging near the top/bottom of a long channel list did not auto-scroll on iOS.

-   New platform-split scroll wrapper in `packages/app/ui/components/ManageChannels/`:
    -   `ManageChannelsScrollContainer.native.tsx`: `Animated.ScrollView` + `useAnimatedRef`, exposes the ref via context.
    -   `ManageChannelsScrollContainer.web.tsx`: Tamagui `ScrollView` with the same `style` and `contentContainerStyle` props as the prior inline version.
    -   `ManageChannelsScrollContainer.tsx` / `.types.ts`: neutral entry and shared types.
-   `ManageChannelsShared.tsx`: swap inline `ScrollView` for `<ManageChannelsScrollContainer>`.
-   `ManageChannelsScreenView.tsx`: read the ref via the context hook and pass it as `scrollableRef` to `Sortable.Grid`.

**Underlying data corruption.** `addChannelToNavSection` in `packages/shared/src/store/sync/sync.ts` was non-defensive on two axes: it didn't remove a channel's prior section memberships before insert, and it called `db.addChannelToNavSection` with `update.sectionId` (the bare backend zone id) instead of `update.navSectionId` (the prefixed local DB primary key, `${groupId}-${sectionId}`, which is also the FK target on `group_nav_section_channels`). Combined with a schema that allows multiple nav-section membership rows for the same `channelId` as long as the `groupNavSectionId` differs, normal create/move flows could leave duplicate persisted memberships. The `db.addChannelToNavSection` query helper itself was also not idempotent: it shifted target-section `channelIndex` values via `UPDATE` before the `INSERT ... ON CONFLICT DO NOTHING`, so replayed events corrupted ordering metadata even when no new row was inserted. And the full group sync write path was additive only
(`onConflictDoNothing`), so stale duplicate memberships persisted forever unless an unrelated event happened to touch the channel.

-   `packages/shared/src/store/sync/sync.ts`: in the `addChannelToNavSection` case, call `removeChannelFromOtherNavSections` before `addChannelToNavSection`, both using `update.navSectionId`, both threaded through the existing `ctx`. Each incoming event is treated as authoritative for the channel's section. `handleGroupUpdate` is exported for testability.
-   `packages/shared/src/db/queries.ts`:
    -   New `removeChannelFromOtherNavSections` write (`DELETE ... WHERE channelId = ? AND groupNavSectionId != ?`).
    -   `addChannelToNavSection` made idempotent: queries for an existing `(groupNavSectionId, channelId)` row at the top of the transaction and returns early if found, so replays don't shift sibling indexes.
    -   `insertGroups`: in the navSections write path, delete existing `group_nav_section_channels` rows for this group's nav sections before inserting the incoming backend payload. Makes a full sync authoritative for nav-section memberships, so any stale local duplicates are reconciled deterministically. Paired with the forced-sync trigger above, a user opening manage-channels on a duplicate-channels group gets repair-on-first-render.
    
### Tests

-   `packages/shared/src/store/sync/handleGroupUpdate.test.ts` (new) — handler writes the prefixed nav-section id, removes prior memberships, never writes a row under the bare backend zone id; idempotent on replay (no new row, no `channelIndex` drift).
-   `packages/shared/src/db/queries.test.ts` — new regression for full-payload reconciliation. Seeds a corrupt duplicate state via the standard insert path, processes a clean payload where the channel is in one section at a non-zero `channelIndex`, asserts the stale row is removed and the remaining row matches the backend's section + index. Idempotent on replay.
-   `packages/app/hooks/useSortableChannelNav.test.ts` (new) — 12 cases covering `findDuplicateChannelIds` and `shouldFireDuplicatesCallback` (set-equal regardless of order, fires once per distinct set, re-arms on resolve).

Each regression test was verified to fail when its targeted bug is reintroduced.

## How did I test?

**Automated:**

-   `pnpm --filter @tloncorp/app tsc`, `pnpm --filter tlon-mobile tsc`, `pnpm --filter tlon-web tsc` — clean.
-   `pnpm exec vitest run src/store/sync/handleGroupUpdate.test.ts` from `packages/shared` — both regression tests pass. Each was verified to fail when its targeted bug is reintroduced (the first via `SQLITE_CONSTRAINT_FOREIGNKEY` from writing the bare zone id, the second via observed `channelIndex` drift on replay).

**Manual (iOS simulator):**

-   Pre-fix baseline: drag does not activate on the duplicate-channels group (Bug A); no edge auto-scroll on the clean long list (Bug B). Establishes the broken baseline.
-   With H1 plumbing + read-side dedupe applied: drag activates on both the clean group and the duplicate-channels group; reorder persists.

**Manual (Android emulator):**

-   Pre-fix baseline: drag works on both groups, sort persists — confirms the iOS-only nature of Bug A despite identical persisted data.

## Risks and impact

    -   [ ] Onboarding
    -   [x] State / providers
    -   [ ] Message sync
    -   [ ] Channel display
    -   [ ] Notifications

Notes for reviewers:

-   `ManageChannelsShared.tsx` is the entry web also reaches. The web variant of the new scroll container renders the same Tamagui `ScrollView` with the same `style` and `contentContainerStyle` props as the prior inline `ScrollView`, so web behavior is unchanged.
-   The sync-handler change is in a hot path for every group subscription update. The change is additive (one extra `DELETE` before the existing insert), gated on `update.navSectionId`, which is non-empty by construction in the parser.
-   The schema prevents duplicate exact `(groupNavSectionId, channelId)` rows, but still permits the same `channelId` to be associated with multiple nav sections. The "one nav section per channel" invariant is now enforced by writers, not by a schema-level unique constraint on `channelId`. A schema-level hardening could follow.
-   Read-side dedupe is a behavior change for users with pre-existing duplicate persisted nav data: a channel previously visible in two sections in manage-channels will now appear in only one. This is the correct behavior — the duplicated state was the bug — but worth flagging.
-   The sync-handler defense prevents new duplicates. Existing duplicates in user databases get cleaned up opportunistically the next time the backend emits an `addChannelToNavSection` event involving any of the affected channels. No one-off cleanup migration is included.

## Rollback plan

Revert.

## Screenshots / videos

https://github.com/user-attachments/assets/0b7d5e4e-b75c-4f38-a544-df8f76586e1a

### DIFF
--- a/apps/tlon-web/e2e/group-channel-management.spec.ts
+++ b/apps/tlon-web/e2e/group-channel-management.spec.ts
@@ -48,36 +48,38 @@ test('should handle channel management operations', async ({ zodPage }) => {
     timeout: 5000,
   });
 
-  // Verify channel count is now 2 by checking for both channels
-  await expect(page.getByTestId('ChannelItem-General-0')).toBeVisible({
+  // Verify channel count is now 2 by checking for both channels.
+  // The backend prepends newly-created channels to a section's `zone.order`,
+  // so the new channel lands at index 0 and General is shifted to index 1.
+  await expect(
+    page.getByTestId('ChannelItem-Second chat channel-0')
+  ).toBeVisible({ timeout: 10000 });
+  await expect(page.getByTestId('ChannelItem-General-1')).toBeVisible({
     timeout: 10000,
   });
-  await expect(
-    page.getByTestId('ChannelItem-Second chat channel-1')
-  ).toBeVisible({ timeout: 10000 });
 
   // Enter sort/edit mode
   await page.getByText('Sort', { exact: true }).click();
   await page.waitForTimeout(500);
 
-  // Drag "General" (at position 0) below "Second chat channel" (at position 1)
+  // Drag "Second chat channel" (at position 0) below "General" (at position 1)
   // Use manual mouse operations on the drag handle for @dnd-kit compatibility
-  const generalDragHandle = page.getByTestId('ChannelDragHandle-General-0');
   const secondDragHandle = page.getByTestId(
-    'ChannelDragHandle-Second chat channel-1'
+    'ChannelDragHandle-Second chat channel-0'
   );
+  const generalDragHandle = page.getByTestId('ChannelDragHandle-General-1');
 
-  const generalHandleBox = await generalDragHandle.boundingBox();
   const secondHandleBox = await secondDragHandle.boundingBox();
+  const generalHandleBox = await generalDragHandle.boundingBox();
 
-  if (generalHandleBox && secondHandleBox) {
-    // Start drag from center of General's drag handle
-    const startX = generalHandleBox.x + generalHandleBox.width / 2;
-    const startY = generalHandleBox.y + generalHandleBox.height / 2;
+  if (secondHandleBox && generalHandleBox) {
+    // Start drag from center of Second chat channel's drag handle
+    const startX = secondHandleBox.x + secondHandleBox.width / 2;
+    const startY = secondHandleBox.y + secondHandleBox.height / 2;
 
-    // End drag below Second chat channel's drag handle
-    const endX = secondHandleBox.x + secondHandleBox.width / 2;
-    const endY = secondHandleBox.y + secondHandleBox.height + 20;
+    // End drag below General's drag handle
+    const endX = generalHandleBox.x + generalHandleBox.width / 2;
+    const endY = generalHandleBox.y + generalHandleBox.height + 20;
 
     await page.mouse.move(startX, startY);
     await page.mouse.down();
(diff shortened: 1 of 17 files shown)


patosullivan:
@claude review once

patosullivan:
@claude review once

janicduplessis (approved):
sortable fixes looks good to me, for the data corruption I am not super familiar with it so if someone else can stamp that part it would be great.

wca4a:
I did not find any discrete, actionable bugs in the changed code. The new duplicate-detection, repair, scroll-ref wiring, and sync/db reconciliation logic appear consistent with the stated behavior and covered scenarios.

PR #5794, opened 2026-04-28
fix(bottom-sheet): avoid flex overflow on first open

janicduplessis:
## Summary

Patches `@gorhom/bottom-sheet@5.2.6` so the very first frame of the sheet's content style returns `{ height: 0 }` instead of `{}`, then snaps the height directly (without `withTiming`) on the first real layout pass. Fixes the visible symptom where a `flex:1` ScrollView (or View) with content larger than the eventual sheet viewport causes the sheet's footer / submit button to be pushed off-screen on first open.

Fixes TLON-5660

## Root cause

Underlying Yoga bug: `computedFlexBasis` is set on the first measure pass and never invalidated, so a `flex:1` child measured at intrinsic content size while its parent is unconstrained keeps that stale basis when the parent's height arrives a frame later. The flex distribution then allocates the wrong size to the child and siblings (the footer) end up positioned past the parent's bounds.

The proper fix is enabling Yoga's experimental `WebFlexBasis` flag in React Native (facebook/yoga#1552), but that requires building React Native from source, which we don't currently do. See #5790 (closed) for the writeup of that path.

## How the patch works

1. **First frame** (`containerHeight === INITIAL_LAYOUT_VALUE`): return `{ height: 0 }` instead of `{}`. Children can't lay out at intrinsic content size, so Yoga doesn't cache an over-large flex basis. A shared-value flag `isInitialLayout` is set to `true`.
2. **First real layout pass** (containerHeight is finally measured, flag is `true`): return `{ paddingBottom, height }` directly — no `withTiming` wrapping. The height shared value snaps from `0` to the target value rather than animating up. Flag is set to `false`.
3. **Subsequent frames**: normal animated path with `animate()` for height and paddingBottom, so keyboard transitions, snap-point changes, etc. animate smoothly as before.

`useSharedValue` is added to track the first-frame state across worklet runs.

## Changes

- `patches/@gorhom__bottom-sheet@5.2.6.patch` — adds the snap-on-first-real-frame path described above.
- `patches/README.md` — section explaining why and when to remove.
- `package.json`, `pnpm-lock.yaml` — register the patch.

## How did I test?

- Reproduced the bug in a minimal Expo dev build with the unpatched `@gorhom/bottom-sheet@5.2.6` and a sheet containing `flex:1 ScrollView` + 1000pt content + footer. Footer is hidden on first open.
- Bisected the gorhom source to identify the early-`return {}` as the trigger.
- Earlier draft of this patch (`{ height: 0 }` without the snap-on-first-real-frame branch) caused the sheet's content to *animate* its height from 0 to the target on first open — visible on Android, subtle on iOS — because `withTiming(target)` interpolated from the prior 0. The current variant snaps the height in directly on the first real frame, then resumes animated transitions.

image

image

## Risks and impact

The patched code path runs only while `containerHeight === INITIAL_LAYOUT_VALUE` (the sentinel for "not yet measured") and the single first real frame after it. Subsequent transitions are unchanged.

Behavioral difference vs upstream: a sheet that previously rendered at intrinsic content size during its single initial frame now renders at zero size for that one frame. In practice this isn't visible because the sheet is still off-screen at that moment.

The patch only affects gorhom-based sheets. Tamagui Sheet / RN Modal / other components with the same shape would still be affected by the underlying Yoga bug.

- Safe to rollback without consulting PR author? Yes — revert the patch file diff and `pnpm install`.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display (any sheets used in chat: CreateChatSheet, group/channel detail sheets, etc.)
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

`git revert <this-commit>` then `pnpm install`.


patosullivan (approved):
Nice, looks good.

patosullivan:
This is for TLON-5660, right?

janicduplessis:
Yes! forgot to add in PR desc

PR #5796, opened 2026-04-28
various: tree deletion correctness

Fang-:
## Summary

Starting in 408, the compiler will catch these easy-to-make mistakes. This applies the cases found by that early. Twin commit to the one in #5648.

## Changes

Correct map-vs-set and deletion key in the following:
- Post and reply deletion.
- Group membership deletion.
- Presence interest dropping.

## How did I test?

Compiles in 408!

## Risks and impact

## Rollback plan

Revert.


mikolajpp (approved):
Great work! Approved.

PR #5797, opened 2026-04-28
notify: support generic notifications

latter-bolden:
## Summary

Adds a notification type for sending generic messages. For use in dispatching timely messages from Hosting.

Model generated, feel free to discard if unfit or insufficiently idiomatic.

## Changes
- Added `%send-message` to `%notify` client actions, with JSON mark support
- Added `%message` notify update action carrying a generic text payload
- Sends generic notification messages to connected providers without changing activity or badge counts
- Added URL encoding for notify provider form payloads before posting to Twilio
- Included message form field only for generic `%message` notifications

## How did I test?

`|commit`'ed locally

## Risks and impact

Fixes TLON-5713

### DIFF
--- a/desk/app/notify.hoon
+++ b/desk/app/notify.hoon
@@ -771,9 +771,8 @@
         id+(scot %ud uid.update)
         notify-count+(scot %ud notify-count.update)
         dismiss-source+?.(?=(%dismiss -.action.update) '' source.action.update)
+        message+?.(?=(%message -.action.update) '' message.action.update)
     ==
-  =?  params  ?=(%message -.action.update)
-    [message+message.action.update params]
   %:  post-form
       /send-notification/(scot %uv (sham eny.bowl))
       notify-endpoint.entry
--- a/desk/app/notify.hoon
+++ b/desk/app/notify.hoon
@@ -756,9 +756,8 @@
   |=  [[p=@t q=@t] out=@t]
   =/  key  (crip (en-urlt:html (trip p)))
   =/  val  (crip (en-urlt:html (trip q)))
-  ?:  =(out '')
-    (rap 3 key '=' val ~)
-  (rap 3 out '&' key '=' val ~)
+  =/  pam  ?:(=(out '') '' '&')
+  (rap 3 out pam key '=' val ~)
 ::
 ++  send-notification
   |=  [entry=provider-entry who=@p =update]
--- a/desk/app/notify.hoon
+++ b/desk/app/notify.hoon
@@ -346,6 +346,22 @@
         =/  pact=provider-action  [%client-leave service.act]
         :_  state
         [(poke:pass [who.act %notify] %notify-provider-action !>(pact))]~
+      ::
+          %send-message
+        =+  .^(=activity:v8:av %gx (scry:io %activity /v4/activity/activity-summary-4))
+        =/  notify-count=@ud
+          notify-count:(~(gut by activity) [%base ~] *activity-summary:a)
+        =/  v1-paths
+          %+  murn  ~(tap by sup.bowl)
+          |=  [duct ship =path]
+          ?:  ?=([%notify *] path)  ~
+          `path
+        ?~  v1-paths
+          [~ state]
+        =/  =update:v1
+          [notify-count `@`(sham eny.bowl) %message message.act]
+        :_  state
+        ~[(fact:io notify-update-1+!>(update) v1-paths)]
       ==
     ::
     ++  handle-http-request
@@ -738,9 +754,11 @@
   ^-  @t
   %+  roll  data
   |=  [[p=@t q=@t] out=@t]
+  =/  key  (crip (en-urlt:html (trip p)))
+  =/  val  (crip (en-urlt:html (trip q)))
   ?:  =(out '')
-    (rap 3 p '=' q ~)
-  (rap 3 out '&' p '=' q ~)
+    (rap 3 key '=' val ~)
+  (rap 3 out '&' key '=' val ~)
 ::
 ++  send-notification
   |=  [entry=provider-entry who=@p =update]
@@ -755,6 +773,8 @@
         notify-count+(scot %ud notify-count.update)
         dismiss-source+?.(?=(%dismiss -.action.update) '' source.action.update)
     ==
+  =?  params  ?=(%message -.action.update)
+    [message+message.action.update params]
   %:  post-form
       /send-notification/(scot %uv (sham eny.bowl))
       notify-endpoint.entry
--- a/desk/mar/notify/client-action.hoon
+++ b/desk/mar/notify/client-action.hoon
@@ -18,6 +18,7 @@
     :~  connect-provider+connect-provider
         remove-provider+remove-provider
         connect-provider-with-binding+connect-provider-with-binding
+        send-message+send-message
     ==
     ++  connect-provider
       %-  ot
@@ -36,6 +37,10 @@
       %-  ot
       :~  who+(su fed:ag)
           service+so
+      ==
+    ++  send-message
+      %-  ot
+      :~  message+so
       ==
     --
   --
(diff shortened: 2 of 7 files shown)


desk/app/notify.hoon
```diff
@@ -755,6 +773,8 @@
         notify-count+(scot %ud notify-count.update)
         dismiss-source+?.(?=(%dismiss -.action.update) '' source.action.update)
     ==
+  =?  params  ?=(%message -.action.update)
+    [message+message.action.update params]
```
mikolajpp: This looks good, but conflicts with the pattern above where we default to empty string. This should be resolved either way, unless there is a difference here.

desk/app/notify.hoon
```diff
@@ -346,6 +346,22 @@
...
+          %+  murn  ~(tap by sup.bowl)
+          |=  [duct ship =path]
+          ?:  ?=([%notify *] path)  ~
+          `path
+        ?~  v1-paths
+          [~ state]
+        =/  =update:v1
+          [notify-count `@`(sham eny.bowl) %message message.act]
+        :_  state
+        ~[(fact:io notify-update-1+!>(update) v1-paths)]
```
mikolajpp, line 364: I am not too familiar with notify. This probably needs to be improved, and checked for correctness, even if it does what it was supposed to do.

mikolajpp (changes requested):
This looks good enough. That said I left one question to answer, and I need to check some portions for correctness before we merge this.

desk/sur/notify.hoon
```diff
@@ -19,6 +19,7 @@
   $%  [%connect-provider who=@p service=term address=@t]
       [%connect-provider-with-binding who=@p service=term address=@t binding=@t]
       [%remove-provider who=@p service=term]
+      [%send-message message=@t]
   ==
```
Fang-, line 23: Important to note for ops that changes here require deployment to the provider _first_, otherwise interactions to it will fail.

desk/app/notify.hoon
Fang-: Don't we need client-side changes to read & use this additional field? What are the intended semantics here?

The existing pattern is that the push contains a notification ID, which the client then uses to fetch notification body/details from the backend. Here, we don't want to fetch the details from the backend, but want to show just the push body/provided message instead. Right?

desk/app/notify.hoon
Fang-, line 364: Important that we indeed don't put the notification into `.notifications` in state, because that will track when/whether the notification gets fetched by the client, used in notif responsiveness statistics. If we don't add it (because it won't get fetched) then this won't dirty our stats. (But also not show up in them.)

Fang-: Also kinda weird how we put `.notify-count` in every update, but that was already the case and is out of scope for this PR.

That does force us to do scry-and-construct for the update in all these different branches, which could be deduplicated logic-wise, but not strictly necessary for this PR. Would rather have a human do a pass over the agent in general at some point.

desk/app/notify.hoon
```diff
@@ -738,9 +754,11 @@
   ^-  @t
   %+  roll  data
   |=  [[p=@t q=@t] out=@t]
+  =/  key  (crip (en-urlt:html (trip p)))
+  =/  val  (crip (en-urlt:html (trip q)))
   ?:  =(out '')
-    (rap 3 p '=' q ~)
-  (rap 3 out '&' p '=' q ~)
+    (rap 3 key '=' val ~)
+  (rap 3 out '&' key '=' val ~)
```
Fang-: Good catch on the form data safety here. Probably necessary now that we're doing arbitrary strings instead of just already-url-safe atom encodings.

Fang-:
```suggestion
  =/  key  (crip (en-urlt:html (trip p)))
  =/  val  (crip (en-urlt:html (trip q)))
  =/  pam  ?:(=(out '') '' '&')
  (rap 3 out pam key '=' val ~)
```

Or:

```suggestion
  %+  rap  3
  :~  out
      ?:(=(out '') '' '&')
      (crip (en-urlt:html (trip p)))
      '='
      (crip (en-urlt:html (trip q)))
  ==
```

Fang- (commented):
Looks fair overall, but I have questions about how the client will deal with this. Not obvious from this PR alone, I would've expected to see a client diff too.

wca4a:
[P2] Handle `message` actions in push clients

The new `%send-message` path posts `action=message` to the provider from `desk/app/notify.hoon`, but the iOS notification service parser (`PushNotificationManager.parseNotificationUserInfo`) currently only accepts `notify` and `dismiss`. Generic message pushes will be classified as invalid/unrecognized unless the client payload/parser contract is updated alongside this server-side emission.

desk/app/notify.hoon
mikolajpp: I have taken a second look at this. This portion does look a bit odd, but apparently it is closely following the coding style found in notify. Improving that is out of scope. I don't have any more comments, but do address the one architectural question left below by @Fang- before merging.

desk/sur/notify.hoon
latter-bolden: Good call out, I wouldn't have remembered that

desk/app/notify.hoon
latter-bolden: Just added the client side changes

desk/app/notify.hoon
latter-bolden: just to confirm, it returns unmodified state so no change needed?

latter-bolden:
I tested this via manual desk install on a hosted node and custom `%notify` provider. Also updated the *Preview debug* service in Twilio.

On iOS and Android local builds, both existing notifs and these custom generic ones came through without issue.

desk/app/notify.hoon
```diff
@@ -755,6 +772,8 @@
         notify-count+(scot %ud notify-count.update)
         dismiss-source+?.(?=(%dismiss -.action.update) '' source.action.update)
     ==
+  =?  params  ?=(%message -.action.update)
+    [message+message.action.update params]
```
Fang-: If there are fields here that the client doesn't look at when processing a `%message` push (and I think it only looks at `action`, `uid` and `message`) then we could omit those from the push data entirely. This would also let us skip calculating the `notify-count`, which (due to `.^` scry overhead) is probably slowing us down a bit.

You'd do something like the following instead:

```hoon
=/  params=(list [@t @t])
  :*  identity+(rsh [3 1] (scot %p who))
      action+`@t`-.action.update
      uid+(scot %uv uid.update)
  ?-  -.action.update
    %notify  :~  ::  we use @ud here for string comparison on client dismissals
                 id+(scot %ud uid.update)
                 notify-count+(scot %ud notify-count.update)
              ==
    %dismiss  ~[dismiss-source+source.action.update]
    %message  ~[message+message.action.update]
  ==  ==
```

(and removing the  `=? params` you added. I'm not doing a `suggestion` here because gh is borking.)

If you do this or similar change, you'd replace `notify-count` in the code above with `0` and put a `::TODO  restructure $update`.

desk/mar/notify/client-action.hoon
```diff
@@ -37,6 +38,10 @@
       :~  who+(su fed:ag)
           service+so
       ==
+    ++  send-message
+      %-  ot
+      :~  message+so
+      ==
```
Fang-, line 44: You could consider just `so` if you want to send `{ 'send-message': 'my text' }` rather than `{ 'send-message': { 'message': 'my text' } }`!

(In that case, I'd remove this arm entirely and write `so` inline on line 21 above.)

desk/app/notify.hoon
Fang-: Indeed, no state change needed/desired.

Fang- (approved):
I understand the client behavior now, thanks for pushing & explaining @latter-bolden!

There's two places we could simplify: input json could be flatter, and we can probably get away with sending less data in the push, at least/especially for `%message` pushes.

But those simplifications aren't blocking for this. Arguably the push data simplification slightly increases the risk profile here, but maybe if you understand the client code well enough that risk falls away. Up to you @latter-bolden, if not now it could make for a good follow-up PR!

arthyn (approved):
lgtm

PR #5798, opened 2026-04-29
telemetry: log instances of protocol mismatch notice seen

jamesacklin:
## Summary

Adds a PostHog event for when a user sees the protocol mismatch message. Also logs their connection state to disambiguate between real clashes vs. trying to interact with an offline / left-behind ship, which both have interesting stories.

## Changes

- Adds a ProtocolMismatchNoticeSeen type to analytics
- Creates a logger in the ReadOnlyNotice component
- When the ReadOnlyNotice component mounts, create a ref to track its "seen" status. If there's a valid protocol mismatch, we log it and grab the connection status. We don't log for any of the ReadOnlyNotice conditions.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: analytics

## Rollback plan

git revert

## Screenshots / videos

n/a

### DIFF
--- a/packages/api/src/types/analytics.ts
+++ b/packages/api/src/types/analytics.ts
@@ -111,6 +111,7 @@ export enum AnalyticsEvent {
   ActionPinChat = 'Pinned Chat',
   ActionUnpinChat = 'Unpinned Chat',
   ActionVisitedChannel = 'Viewed Channel',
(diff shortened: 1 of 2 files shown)


latter-bolden (approved):
Very nice

PR #5800, opened 2026-04-29
desk: 408 compatibility (part 2, da boogaloo)

Fang-:
## Summary

Zuse 408K brings along Hoon 135K. With that, the behavior of `@da` rendering changes to pad month and day segments to be two digits consistently. 135K parsing accepts both the new and old format, but older ships (136K and older) will still only accept the old format.

Here, we update subscription path construction to always use old-style `@da` rendering. This way ships on 408 will be able to (re)open their usual subscriptions to ships running 409.

## Changes

22676b7 updates the tests to pass on 408, accounting for the breaking changes and the ones from #5648.

784f3d3 updates all the places where we serialize timestamps into wires. Found these by looking over all instances of `%da`. In practice, it's always a `+scot` call.

We probably don't want to do this forever. We don't even strictly need to do it right now! But changing the subscription paths constitutes a protocol change, and backcompat here is trivial, so we'll save cleanup here for another day.

We notably don't touch anything that's local only (like wires) or treated opaquely once serialized (like reel nonces).

## How did I test?

Ran a 409 migration and ran the tests. ~Presently awaiting on-the-ground report from mopfel-winrux, who encountered problems caused by this through running pre-release.~ Fix confirmed on an affected ship!

## Risks and impact

- Yes, safe to rollback without consulting PR author, but this is required for 408 compatibility.
  - Message sync

## Rollback plan

Revert.

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -2138,7 +2138,7 @@
     ?.  =(our.bowl ship.nest)
       =/  count  ?:(=(%diary kind.nest) '20' '100')
       /[kind.nest]/[name.nest]/checkpoint/before/[count]
-    /[kind.nest]/[name.nest]/checkpoint/time-range/(scot %da *@da)
+    /[kind.nest]/[name.nest]/checkpoint/time-range/(scot:h136 %da *@da)
   ::
   ++  ca-start-updates
     |=  delay=?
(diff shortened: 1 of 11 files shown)


mikolajpp (approved):
This looks good, approved! 

As we talked off-band, the CI will continue to fail until we prepare a new set of 408K ships for tests to use.

PR #5801, opened 2026-04-30
contacts: expose unified /v1/directory scry

mikolajpp:
## Summary

The new directory scry exposes all known contacts and peers, including user's own profile. This is the API desired by the client, which otherwise would have assemble data obtained from other scries.

 The `isContact` flag indicates whether the entry was derived from contact or peer data. User's profile is considered a contact.

Closes TLON-3879

This is an improved version of the PR originally targeted at landscape: https://github.com/tloncorp/landscape/pull/314

## Changes

1. Remove `/v1/all` scry (unused by the client), and repurpose it into `/v1/directory` which returns a map of `$marked-page`s

## How did I test?

Queried the endpoint.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.

### DIFF
--- a/desk/sur/contacts.hoon
+++ b/desk/sur/contacts.hoon
@@ -82,9 +82,12 @@
 ::  $book: contact book
 ::
 +$  book  (map kip page)
-::  $directory: merged contacts
+::  $marked-page: standalone contact page
 ::
-+$  directory  (map ship contact)
++$  marked-page  [contact=? page]
+::  $directory: peers and contacts directory
+::
++$  directory  (map ship marked-page)
 ::  $peers: network peers
 ::
 +$  peers  (map ship foreign)
--- a/desk/sur/contacts.hoon
+++ b/desk/sur/contacts.hoon
@@ -82,12 +82,12 @@
 ::  $book: contact book
 ::
 +$  book  (map kip page)
-::  $marked-page: standalone contact page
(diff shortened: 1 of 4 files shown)


desk/sur/contacts.hoon
```diff
@@ -82,9 +82,12 @@
 ::  $book: contact book
 ::
 +$  book  (map kip page)
-::  $directory: merged contacts
+::  $marked-page: standalone contact page
 ::
-+$  directory  (map ship contact)
++$  marked-page  [contact=? page]
```
Fang-: Bit of a strange name, given that we sometimes say "marked data" to refer to cages, casks etc. I don't have any better ideas though.

Fang- (approved):
Not clear to me it was entirely necessary to remove the `/v1/all` scry endpoint, but it slots in nicely and we don't use it ourselves. I thought some paldev stuff might, but apparently that's not the case; it still uses the old `/all` endpoint. (^:

Endpoint logic looks good.

desk/sur/contacts.hoon
mikolajpp: I went with `$leaf`, a good old 4-letter.

PR #5802, opened 2026-04-30
fix disabled posthog

davidisaaclee:
## Summary
Avoids PostHog error when running in dev mode without a PostHog key.

## Changes

- `envVars.native.ts` did not have a binding for `POST_HOG_IN_DEV` - added that and use it instead of directly accessing the string `process.env.POST_HOG_IN_DEV`
- `Boolean(process.env.POST_HOG_IN_DEV)` would evaluate to true with `POST_HOG_IN_DEV=false` - changed to check `=== 'true'`
- Optionally disable PostHog client when constructing (instead of further down in provider)

## How did I test?
On iOS:
- Ran development build without a PostHog API key – got past launch.
- Run release non-preview build with a real PostHog key, poked around in app and saw that PostHog events were being received in real-time in online dashboard

## Risks and impact
**This could break sending all analytics** so I'd appreciate a close review.

  - [x] Analytics

## Rollback plan
git revert

### DIFF
--- a/packages/app/utils/posthog.ts
+++ b/packages/app/utils/posthog.ts
@@ -5,7 +5,7 @@ import * as db from '@tloncorp/shared/db';
 import PostHog from 'posthog-react-native';
 import { Platform, TurboModuleRegistry } from 'react-native';
 
-import { GIT_HASH, POST_HOG_API_KEY } from '../constants';
+import { GIT_HASH, POST_HOG_API_KEY, POST_HOG_IN_DEV } from '../constants';
 import { createSentryErrorLogger } from './sentry';
 import { UrbitModuleSpec } from './urbitModule';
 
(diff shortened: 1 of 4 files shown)


davidisaaclee:
@claude review

packages/app/utils/posthog.ts
```diff
@@ -24,12 +24,28 @@ export type OnboardingProperties = {
   inviteType?: 'user' | 'group';
 };
 
-export const posthog: PostHog | undefined =
-  process.env.NODE_ENV === 'test' && !process.env.POST_HOG_IN_DEV
```
davidisaaclee, line 28: My diff changes the behavior here, but I _think_ this was a bug previously. Here's the commit that originally added this check: https://github.com/tloncorp/tlon-apps/commit/575f12cf4a4957e4a45b01ff74a46a448941d882 – I think it makes sense to do:
- If `POST_HOG_IN_DEV` is omitted from config, default to _not_ sending events to PostHog when `NODE_ENV === 'development'`
- If `POST_HOG_IN_DEV`, send events to PostHog in dev
- `POST_HOG_IN_DEV` should not affect behavior of test environments

tagging @latter-bolden in case you remember the intent here – do we want `POST_HOG_IN_DEV` to enable PostHog in tests? (I suppose the CI results will be the judge, but I think I'm going to log off before they complete.)

janicduplessis (approved):
Also noticed this today, looks good!

packages/app/utils/posthog.ts
davidisaaclee: @latter-bolden I'm going to merge this into the expo-54 branch but happy to follow up if you disagree about this change

PR #5803, opened 2026-04-30
fix gallery video layout (#5741)

davidisaaclee:
fixes TLON-5703

Cherry-picks dc85615 to apply video preview layout fix. (Heads up that this cherry-pick is not clean, as `GalleryContentRenderer` was moved – I think I applied the changes correctly.)

Tested on iOS by opening a video from gallery channel -> detail view: video was visible and could be tapped to play.

### DIFF
--- a/packages/app/ui/components/DetailView.tsx
+++ b/packages/app/ui/components/DetailView.tsx
@@ -1,7 +1,7 @@
 import * as db from '@tloncorp/shared/db';
 import { Text } from '@tloncorp/ui';
 import { ReactNode, useMemo } from 'react';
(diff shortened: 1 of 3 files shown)


davidisaaclee:
@claude review

janicduplessis:
Should I just merge develop in the expo branch instead of cherry picking this? I noticed some conflicts so will need to fix those eventually anyway

davidisaaclee:
> Should I just merge develop

yes, that's probably best (maybe after noting the manual conflict resolution I mentioned in PR description so you can apply it during the merge) -- although that seems like it'll be a hairy merge!

PR #5807, opened 2026-04-30
contacts: match notifications and onboarding display

arthyn:
## Summary

Surfaces lanyard contact-discovery matches to the user via three coordinated paths: an in-app "New" pill on the Contacts screen, an inline match list during the onboarding InvitePane, and a local notification when a backgrounded bg-task discovers new matches. All notification logic stays on-device — no payload data crosses to the notify provider.

### Discovery + state
- `syncContactDiscovery` now returns `{ newMatches }` and tracks the diff against existing-and-already-known contacts.
- New `matchedAt: timestamp` column on the `contacts` table; populated for genuinely new matches, cleared on Contacts-screen unmount.
- `addContacts`, `markContactsAsMatched`, and `updateContactMetadata` are scoped to genuinely new matches only — won't clobber an existing `customNickname`, `isContactSuggestion`, or peer profile data on contacts the user has already seen.
- "Already known" filter is broad: `isContact`, `isContactSuggestion`, any peer profile field set, or any user-set custom override.

### UI
- **Contacts screen**: "New" pill on rows where `matchedAt` is set (mobile + desktop). Pill clears on screen unmount.
- **InvitePane** (onboarding): runs lanyard discovery in the background after `syncSystemContacts`. Matches render inline above the invite list with a "Finding your contacts on Tlon…" spinner. If the user advances early, the in-flight promise is tail-fired so the notification still surfaces.
- **Cosmos `Invite Pane` fixture**: stub-store driven controls for match count, latency, and fail toggle. Logging match handler so the advance-early notification path is observable.

### Notifications
- `presentContactMatchNotification` (single) and `presentContactsMatchedNotification` (batched count) helpers using expo-notifications.
- `useNotificationListener` registers a handler that fires those, and routes `contactMatched` / `contactsMatched` taps to UserProfile / Contacts.
- Foreground handler configured via `setNotificationHandler` so banners surface when the app is open.
- Background-sync task fires the same notifications directly when discovery surfaces new matches (no React tree, so the registered handler isn't reachable). Skips first-install flood via `userHasCompletedFirstSync`.

### Infrastructure
- Dev-only mock: `EXPO_PUBLIC_MOCK_LANYARD_DISCOVERY=N` returns synthetic matches, also accepts `phone:ship,phone:ship` pairs for deterministic testing.
- Dev-only debug button: floating "Run bg sync" button gated on `__DEV__ && EXPO_PUBLIC_DEBUG_BG_SYNC_BUTTON=true`.
- Cosmos config pinned to ports 5555 (web) / 5556 (native) so the playground doesn't fight macOS AirPlay on 5000.
- New `clean:all` script for nuking workspace caches.

## How did I test?

- **Cosmos** (`pnpm cosmos:web`) — `Invite Contacts` and `Invite Pane` fixtures: tested empty / discovering / single-match / many-matches / discovery-fails. Verified the advance-early tail path fires the logged "would notify" handler.
- **iOS simulator + real device** with `EXPO_PUBLIC_MOCK_LANYARD_DISCOVERY=3` and `FORCE_SPLASH_SEQUENCE=true`:
  - Onboarding flow lands on InvitePane, discovery runs, matches render inline, advance-early tail fires the notification.
  - Contacts screen shows "New" pill on matched rows; pill clears on unmount.
  - Background sync via the dev-only debug button: discovery completes, notification scheduled, banner displays.
- **Discovery filtering**: confirmed already-existing contacts (with `isContact: true`, `isContactSuggestion: true`, or peer profile data) no longer trigger as new matches and don't get their `customNickname` overwritten.
- **Typecheck**: `pnpm -r tsc` clean across all packages.

Note: BGTaskScheduler's organic "iOS schedules the task on its own" path is hard to verify reproducibly — the debug button exercises the same code path. Worth verifying in TestFlight before relying on production cadence.

## Risks and impact

  - [x] Onboarding (InvitePane discovery surface)
  - [x] State / providers (contacts schema migration, sync)
  - [x] Notifications (new local-notification path on mobile)
  - [ ] Message sync
  - [ ] Channel display

## Rollback plan

Revert this branch — the changes are scoped to:
- `packages/shared/src/db/schema.ts` (and a regenerated migration adding `matchedAt`)
- `packages/shared/src/store/sync/sync.ts` and `lanyardActions.ts`
- `packages/shared/src/db/queries.ts`
- `packages/api/src/client/lanyardApi.ts` (env-gated dev mock; inactive in prod)
- `packages/app/lib/notifications.ts` and `ui/components/Wayfinding/SplashSequence.tsx` and `ui/components/ContactsScreenView.tsx`
- `apps/tlon-mobile/src/lib/backgroundSync.ts` and `src/hooks/useNotificationListener.ts` and `src/App.main.tsx`

The schema migration adds a nullable column; rolling back leaves it in place but unused.


packages/shared/src/store/sync/sync.ts
```diff
@@ -499,8 +502,18 @@ export const syncSystemContacts = async (_ctx?: SyncCtx) => {
...
+};
+
+export const syncContactDiscovery = async (
+  ctx?: SyncCtx,
+  opts?: { invokeHandler?: boolean }
+): Promise<ContactDiscoveryResult> => {
   logger.log('syncContactDiscovery: starting');
+  const invokeHandler = opts?.invokeHandler !== false;
+  const empty: ContactDiscoveryResult = { newMatches: [] };
+  const isMocked = !!process.env.EXPO_PUBLIC_MOCK_LANYARD_DISCOVERY;
```
patosullivan: This won't work on web as-is (there's no `process`)


```suggestion
  const isMocked =
  typeof process !== 'undefined' &&
  !!process.env?.EXPO_PUBLIC_MOCK_LANYARD_DISCOVERY;
```

packages/shared/src/store/sync/sync.ts
```diff
@@ -554,40 +600,72 @@ export const syncContactDiscovery = async (ctx?: SyncCtx) => {
...
-        severity: AnalyticsSeverity.Critical,
-        error: e,
+
+    if (newMatchIds.length > 0) {
+      await addContacts(newMatchIds).catch((e) => {
+        logger.trackEvent(AnalyticsEvent.ErrorContactMatching, {
+          context: 'failed to add contacts',
+          severity: AnalyticsSeverity.Critical,
+          error: e,
+        });
```
patosullivan, line 625: `addContacts` only calls `db.updateContact`, it doesn't call `db.upsertContact`, so if the row doesn't already exist it won't get created. Then, `markContactsAsMatched` also only updates existing rows. 

It looks like rows were only getting created when we heard an event on the subscription and `upsertContact` gets called, so it does eventually get added but we don't have a way to mark it as "New" because `matchedAt` is always null/undefined.

patosullivan (changes requested):
Couple issues, otherwise looks good to go

patosullivan (approved):

packages/app/lib/notifications.ts
```diff
@@ -289,3 +300,53 @@ export async function scheduleNodeResumeNudge(ship: string) {
...
+
+  const notificationId = await Notifications.scheduleNotificationAsync({
+    content: {
+      title: `${name} is on Tlon`,
+      body: 'Tap to say hi',
+      data: { type: 'contactMatched', contactId },
+    },
+    trigger: null,
+  });
+  console.log('[notifications] scheduled contact match notification', {
```
dnbrwstr: Any reason not to use devLogger here?

packages/app/navigation/desktop/ProfileNavigator.tsx
```diff
@@ -32,6 +32,14 @@ function DrawerContent(props: DrawerContentComponentProps) {
   const { data: userContacts } = store.useUserContacts();
   const { data: suggestions } = store.useSuggestedContacts();
 
+  // Clear "new match" pills when the contacts pane unmounts — the user
+  // has had a chance to see them.
+  useEffect(() => {
+    return () => {
+      db.clearContactsMatchedAt().catch(() => {});
```
dnbrwstr: Would probably make sense to log the error in the catch handler

packages/app/features/top/ContactsScreen.tsx
```diff
@@ -47,6 +47,14 @@ export default function ContactsScreen(props: Props) {
     [systemContacts]
   );
 
+  // Clear "new match" pills when leaving the screen — the user has had
+  // a chance to see them.
+  useEffect(() => {
+    return () => {
+      db.clearContactsMatchedAt().catch(() => {});
```
dnbrwstr: Log failure

apps/tlon-mobile/src/components/DebugBackgroundSyncButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+// Dev-only floating button that triggers a single background-sync run
+// against the live JS bundle. Useful for verifying the bg-sync code
+// path (lanyard discovery, match notifications, etc.) without waiting
+// on iOS BGTaskScheduler heuristics or jumping into the Xcode debugger
+// to fire `_simulateLaunchForTaskWithIdentifier:`. Mounted only when
+// __DEV__ is true and EXPO_PUBLIC_DEBUG_BG_SYNC_BUTTON=true is set in
+// .env.local — otherwise this returns null and tree-shakes out.
+const ENABLED =
+  __DEV__ && process.env.EXPO_PUBLIC_DEBUG_BG_SYNC_BUTTON === 'true';
```
dnbrwstr: I'm curious why we have to namespace this variable with `EXPO_PUBLIC` here -- in general it isn't something we need to do.

apps/tlon-mobile/src/components/DebugBackgroundSyncButton.tsx
```diff
@@ -0,0 +1,79 @@
+import { Text } from '@tloncorp/app/ui';
+import { useCallback, useState } from 'react';
+import { Pressable } from 'react-native';
+
+import { runBackgroundSyncFromDebugButton } from '../lib/backgroundSync';
+
+// Dev-only floating button that triggers a single background-sync run
```
dnbrwstr: Seems like it'd be simpler to just make this a dev menu item.

apps/tlon-mobile/src/lib/backgroundSync.ts
```diff
@@ -44,7 +48,6 @@ async function performSync() {
     return;
   }
 
-  logger.trackEvent('Configuring urbit client...');
```
dnbrwstr, line 47: Curious why this is removed.

apps/tlon-mobile/src/lib/backgroundSync.ts
```diff
@@ -70,6 +73,40 @@ async function performSync() {
       },
     });
     timings.changesDuration = Date.now() - changesStart;
+
+    // Run lanyard contact discovery as part of the bg cycle so new
+    // matches surface even if the user hasn't opened the app recently.
+    // We disable the registered-handler path because in a bg task there's
+    // no React tree (and so no useNotificationListener-registered
+    // handler) — instead we schedule a local notification directly.
+    const discoveryStart = Date.now();
```
dnbrwstr, line 75: Can we push the logic for actually running the lanyard sync elsewhere? Maybe in store or sync? In general this file should be a thin wrapper and the meat of tasks should live elsewhere.

packages/app/ui/components/Wayfinding/SplashSequence.tsx
```diff
@@ -1874,12 +1954,50 @@ export function InvitePane(props: {
     setShowInviteContacts(true);
   }, []);
 
+  const runDiscovery = useCallback(
+    async (contacts: db.SystemContact[]) => {
```
dnbrwstr: Might make sense to push this logic into a hook elsewhere.

packages/shared/src/store/sync/sync.ts
```diff
@@ -581,6 +596,39 @@ export const syncContactDiscovery = async (ctx?: SyncCtx) => {
     ).filter((match) => match[1] !== currentUserId);
     logger.log('syncContactDiscovery: got contact discovery matches', matches);
 
+    const contactIds = matches.map((m) => m[1]);
```
dnbrwstr: Another piece it might make sense to push elsewhere.

dnbrwstr (changes requested):
Looks good overall. A few nits. My two main questions are:
- Can we centralize more of the logic around syncing / matching / working with contact sets? Rather than doing it in leaf components, which it starts to bloat a bit
- Is there any way to better separate out the mocking / fixture code? In general we try not to interleave mocks with other code.

apps/tlon-mobile/src/lib/backgroundSync.ts
arthyn: I think the agent went too hard when I told it to cleanup logs that we had added.

packages/app/features/top/ContactsScreen.tsx
```diff
@@ -47,6 +50,18 @@ export default function ContactsScreen(props: Props) {
     [systemContacts]
   );
 
+  // Clear "new match" pills when leaving the screen — the user has had
+  // a chance to see them.
+  useEffect(() => {
```
dnbrwstr: Might be nice to extract a `useMarkMatchesSeen` helper or something like that since we do this a couple times.

packages/app/hooks/useContactDiscovery.ts
```diff
@@ -0,0 +1,91 @@
...
+      } finally {
+        if (pendingDiscoveryRef.current === promise) {
+          setIsDiscovering(false);
+        }
+      }
+    },
+    [storeContext]
+  );
+
+  const tailFireHandlerOnAdvance = useCallback(() => {
```
dnbrwstr: This name is kind of hard to parse -- maybe `notifyPendingMatches`? Something like that.

packages/shared/src/store/sync/sync.ts
```diff
@@ -592,40 +616,72 @@ export const syncContactDiscovery = async (ctx?: SyncCtx) => {
...
+            context: 'failed to mark contacts as matched',
+            error: e,
+          });
+        });
+
+      // Apply the address-book name as customNickname only for new
+      // matches that don't already have one. Avoids clobbering a name
+      // the user explicitly set in a prior session, and avoids racing
+      // with peer profile fetches that may arrive after our write.
+      const newSystemContacts =
```
dnbrwstr: I wonder if there's a nice way to do this in a query rather than manually munging everything

dnbrwstr (approved):
Lgtm, a few nits. Might benefit from a "could this be terser" prompt in a few places, but nothing blocking.

PR #5812, opened 2026-05-01
notes: integrate into channels for those running the desk

arthyn:
## Summary

This PR adds experimental integration with [%notes](https://github.com/arthyn/notes) as a replacement for current day notebooks. Old notebooks become "Bulletins" name TBD. The integration is done through an iframe of the web app served from your ship in an embedded mode which strips some of the chrome.

## Changes

- Widen available channel types to include notes if available
- Allow creation/deletion of notes notebooks
- Support rendering notes "channels"

## How did I test?

Manually with dev:web, on a ship with the %notes agent you see the option to use the new notebooks. On a ship without the agent, you see the standard UI as it exists today.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [X] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Possibly want to run a script to remove these channels?


dnbrwstr (approved):
What's the plan here? Looks fine for experimentation but would want to clean up divergence for a wider release.

arthyn:
@dnbrwstr plan here is just to experiment for now. if we decide this is the right direction we'd integrate this more fully with the backend and also make a UI here instead of a webview so that there's less divergence.

dnbrwstr:
> @dnbrwstr plan here is just to experiment for now. if we decide this is the right direction we'd integrate this more fully with the backend and also make a UI here instead of a webview so that there's less divergence.

Cool, makes sense to me.

PR #5819, opened 2026-05-04
fix: duplicate new section sheet on Android

patosullivan:
## Summary

Fixes TLON-5664: on Android, the `Add section` sheet (group → Channels → New → New section) stayed visible after tapping `Save` with the keyboard open. Adds an opt-in `unmountOnClose` prop to `BottomSheetWrapper` and uses it from `EditSectionNameSheet` so the desynced subtree is removed from the view tree on close.

I confirmed that the problem exists on both `develop` and the expo-54 branch.

## Background

My first attempt to fix this issue worked through five variants of "ask the sheet to close differently": reordering `reset()` around `setTimeout(0)` and `onOpenChange(false)`, firing the close before awaiting the save, opting into `keyboardBlurBehavior='none'`, calling `forceClose()` directly, and inserting an explicit `Keyboard.dismiss()` plus delay. None of them moved the needle. Adding logs to the wrapper around the close path explains why.

The diagnostic trace from that attempt shows `forceClose()` reaching Gorhom on the Add-section wrapper instance, Gorhom acknowledging the close with `onChange index=-1`, no preceding `onAnimate from=0 to=-1` (the open path produces a normal `from=-1 to=0` pair, so the asymmetry is real), no second `BottomSheetWrapper` mounted in the React tree, and the sheet, backdrop, handle, header, and `Save` button all still rendered for at least 5 seconds after that `onChange index=-1`. In other words, Gorhom's reported snap state on this instance is `-1` while its native/Reanimated subtree is still at the `index=0`-equivalent — the desync sits below Gorhom's controlled `index` and its imperative `close`/`forceClose` API. Every layer the prior variants acted on is a layer Gorhom is already (internally) acknowledging, which is why none of them changed the visible outcome. (Diagnostics indicate the visible duplicate-like state is one stuck sheet, not two mounted wrappers.)

The fix has to act below that API surface. Removing the desynced subtree from the view tree (unmounting and remounting `BottomSheet` around the close) replaces the stale native subtree before the next open, so the desync state is not carried forward by the same Gorhom instance. That is the layer the diagnostic trace identifies as desynced and the layer the new opt-in acts at.

I looked at and rejected a few smaller alternatives:

-   Bumping `@gorhom/bottom-sheet` from 5.2.6 to a newer 5.x. Gorhom 5.x is sensitive to its Reanimated and gesture-handler peers, and a bump touches every sheet in the app, seems too broad to bundle with this fix. Left as an optional post-merge follow-up.
-   Bumping a `key` on `<EditSectionNameSheet>` from `ManageChannelsShared.tsx`. Same mechanism, but applied at the wrong layer: every other input-bearing always-mounted Gorhom sheet that hits this Android desync would need its own parent-side glue. The wrapper-level opt-in generalises.
-   Conditionally mounting `<EditSectionNameSheet>` from `ManageChannelsShared.tsx` (mirroring `CreateChannelSheet`). Smaller diff, but it skips the close animation entirely on iOS where the animation is fine today. The wrapper opt-in keeps iOS animation via the timed grace window.

## Changes

-   `packages/app/ui/components/BottomSheetWrapper.types.ts` — added `unmountOnClose?: boolean` prop with JSDoc.
-   `packages/app/ui/components/BottomSheetWrapper.native.tsx` — added opt-in lifecycle (mounted state, `mountKey`, `prevOpenRef`, close timer using `ANIMATION_CONFIGS[animation].duration + 100ms` grace), refactored `useImperativeHandle` to read refs at call time so the methods stay correct across `mountKey` remounts, added `key={mountKey}` to both `<BottomSheet>` and `<BottomSheetModal>`, gated the render branches on `mounted` while preserving hook order, and extended the modal/non-modal control effects to depend on `mountKey` so a key bump while `open=true` re-applies `present()`/`expand()` to the new instance.
-   `packages/app/ui/components/ActionSheet.tsx` — added `unmountOnClose` to the `Pick<BottomSheetWrapperProps, ...>` union, destructured it out of the rest spread (so the prop does not leak as a DOM attribute through Tamagui's web `Sheet`), and forwarded it only to the native `<BottomSheetWrapper>`.
-   `packages/app/ui/components/ManageChannels/EditSectionNameSheet.tsx` — gated the form reset on `open === true` (covers add-empty, edit-current-title, and cancel/backdrop reopen), removed the save-path `reset()` to avoid a visible mid-flight blank state during the close-grace window, and opted into `unmountOnClose` on the `<ActionSheet>`. Applies to both `mode="add"` and `mode="edit"`.
-   `apps/tlon-mobile/ios/Podfile.lock` — checksum-only update from local pod setup; no dependency versions changed.

This changes shared native sheet infrastructure, but the new unmount/remount lifecycle is opt-in: existing consumers stay on the old always-mounted path unless they pass `unmountOnClose`. `unmountOnClose` defaults to `false`, the wrapper renders identically to before, the imperative handle keeps its API, `mounted` starts as `open || !unmountOnClose` so default callers stay mounted regardless of `open`, and `mountKey` only bumps on a real `false → true` transition so first-mount-with-`open=true` does not flicker.

## How did I test?

-   Manual Android (`Settings → Channels → New → New section`): typed a name, tapped `Save` with the keyboard open, confirmed the sheet is gone shortly after the spinner finishes with no blank/reset/duplicate residue.
-   Manual Android edit flow: edited an existing section name, saved, confirmed clean close.
-   Manual Android: cancel/backdrop dismiss, reopened, confirmed the form is empty.
-   Reopen checks (both add and edit): watched the open animation for stale first-frame form state, the title field was correct from the first visible frame, with no flash of the previous value.
-   Manual iOS: same flows; the close animation plays cleanly inside the ~350 ms grace window on the keyboard-open save path.
-   Manual web: same flows; no `unmountonclose` DOM attribute leak and no React unknown-prop warning in the console.
-   Regression check on Android default-path sheets (`New channel` and other text-input sheets reachable elsewhere): unchanged behavior.

## Risks and impact

-   Affects important code area: N/A

Additional notes for reviewers:

-   The render gate is opt-in; it never returns `null` for default-path callers.
-   `mountKey` bumps only on real `false → true` transitions, so first-mount-with-`open=true` does not cause a flicker.
-   On Android, in the bug case where Gorhom's close animation desyncs from the rendered native subtree, the sheet may visibly hard-cut to gone at the end of the ~350 ms grace window instead of animating off-screen. This is a deliberate trade-off; the alternative is the current 5+ second stick.
-   The approved plan's iOS fallback ladder (extend grace, `Platform.select` grace, Android-only opt-in, etc.) was not needed, iOS preserved its close animation inside the grace window and web showed no prop leak or warning.

## Rollback plan

Revert.


patosullivan:
@claude review once

packages/app/ui/components/BottomSheetWrapper.native.tsx
```diff
@@ -223,31 +239,92 @@ export const BottomSheetWrapper = forwardRef<
...
-        return {
-          present: () => (modal ? modalRef?.present() : sheetRef?.expand()),
-          dismiss: () => (modal ? modalRef?.dismiss() : sheetRef?.close()),
-          close: () => (modal ? modalRef?.dismiss() : sheetRef?.close()),
-          expand: () => (modal ? modalRef?.present() : sheetRef?.expand()),
-          collapse: () => (modal ? modalRef?.dismiss() : sheetRef?.collapse()),
+      () =>
+        ({
+          present: () =>
+            modal
```
dnbrwstr: Could this be terser? Seems a little verbose.

dnbrwstr (approved):
Makes sense to me, but it's a shame we have to do this. There have been several sheet sync related fixes on `@gorhom/bottom-sheet` since the version we're using -- could you verify that none of them help? 

If not, I think this is ok for now, but we should keep an eye on the upstream.

patosullivan:
> Makes sense to me, but it's a shame we have to do this. There have been several sheet sync related fixes on `@gorhom/bottom-sheet` since the version we're using -- could you verify that none of them help?
> 
> If not, I think this is ok for now, but we should keep an eye on the upstream.

I looked through 5.2.7 -> 5.2.13, all of the Android/index-render fixes are scoped to BottomSheetModal, not BottomSheet :(

packages/app/ui/components/BottomSheetWrapper.native.tsx
patosullivan: Fixed in latest commit

PR #5820, opened 2026-05-05
Aqua tests: agentic support

mikolajpp:
## Summary

This PR significantly reworks the aqua tests tooling to accommodate model-assisted generation of aqua tests.
We implemented a few tools to assist with setting up virtual aqua environment, wrote a human-oriented aqua documentation, which at the same time provides enough context for models.

We tested this setup by having an agent generate two basic aqua tests for the contacts agent.

## Changes

1. Implement two new tools `-ph-fleet` and `-ph-test-ls`. `-ph-fleet` aids in management of aqua snapshots, while `-ph-test-ls` compiles and lists available aqua tests.
2. Adjust the aqua test runner `-ph-test` to work with snapshots generated by `-ph-fleet`
3. Introduce aqua documentation.
4. Adjust the backend test runner to use new tooling.
5. Change the ci testing fleet to include real notify and bait providers, together with parent galaxies.
6. Add two model-generated aqua tests for contacts.

## How did I test?
I run the tests.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.


backend/run-aqua-tests.sh
```diff
@@ -224,15 +224,52 @@ ${run_click} $pier "/lib/pill/hoon"<<EOF
 (pure:m !>(%ok))  
 EOF
 
-# Run the unit tests
+echo "Preparing aqua snapshot..."
+result=$( $run_click -t 1200 $pier <<EOF
+=/  m  (strand ,vase)  
+;<  =bowl  bind:m  get-bowl  
+=/  tid  (scot %ta (cat 3 'strand_' (scot %uv (sham %ph-fleet eny.bowl))))
```
Fang-: Given we only run a single one of these threads for a CI run, we could get away with a hardcoded thread id, right?

(I know thread id construction gets copied around but they're not particularly useful/helpful. I'd much rather have the name/description of what the thread does in there...)

desk/lib/ph/test.hoon
```diff
@@ -1,6 +1,7 @@
 /-  spider
 /+  *strandio, *ph-io, test
 =,  strand=strand:spider
+=+  timeout=~s15
```
Fang-: You add this, but then don't use it anywhere?

desk/ted/ph/test.hoon
```diff
@@ -99,57 +103,22 @@
...
-::
 ^-  thread:spider
 |=  args=vase
 =/  m  (strand ,vase)
 ^-  form:m
-=+  !<(pax=(unit path) args)
+=+  !<(args=(unit [pax=(unit path) snap=@t]) args)
+=/  [pax=(unit path) snap=@]
+  ?~  args  ~|(%no-args-found !!)
+  [pax.u.args snap.u.args]
```
Fang-:
```suggestion
=+  ~|  %no-args-found
    (need !<(args=(unit [pax=(unit path) snap=@t]) args))
```

desk/ted/ph/fleet.hoon
```diff
@@ -0,0 +1,97 @@
...
+  ?~  args  ~|(%no-args-found !!)
+  [snap-id fleet sync]:u.args
+=.  fleet
+  ^-  (list ship)
+  :*  ~loshut-lonreg  ::  bait provider
+      ~rivfur-livmet  ::  notify provider
+      ~dem            ::  bait provider galaxy
+      ~fen            ::  notify provider galaxy
+      fleet
+  ==
```
Fang-: We always put these live regardless of the `.fleet` specification? Isn't it ok for some testing scenarios to not have a reachable default bait/notify provider? What if `.fleet` already contains one of these ships?

desk/ted/ph/fleet.hoon
```diff
@@ -0,0 +1,97 @@
...
+;<  ~  bind:m
+  =/  n  (strand ,~)
+  ?.  sync  (pure:n ~)
+  ~>  %slog.1^(crip "Syncing %groups desk to ships...")
+  |-
+  ?~  fleet  (pure:n ~)
+  ;<  ~  bind:n  (sync-desk i.fleet %groups)
+  $(fleet t.fleet)
+::  allow agents time to cool down. it takes about a minute
+::  for ames connections to be established.
```
Fang-, line 78: Is this the mesa jank you described? Kinda crazy.

I wonder if, given we're virtualizing these ships, we could run their timers in virtual time and just eat through them at 100x speed. (^:

desk/tests/ph/app/contacts.hoon
```diff
@@ -0,0 +1,92 @@
...
+  ;<  ~  bind:m  (watch-app /~bud/contacts/v1/news [~bud %contacts] /v1/news)
+  ::  ~bud meets ~zod and receives ~zod's initial empty profile.
+  ::
+  ;<  ~  bind:m  (poke-app [~bud %contacts] contact-action-1+[%meet ~[~zod]])
+  ;<  pay=cage  bind:m  (wait-for-app-fact /~bud/contacts/v1/news [~bud %contacts])
+  ?>  =(%contact-response-0 p.pay)
+  =+  !<(res=response:c q.pay)
+  ?>  ?=(%peer -.res)
+  ;<  ~  bind:m  (ex-equal !>(who.res) !>(~zod))
+  ::  ~zod updates his public profile, then ~bud receives the broadcast.
```
Fang-, line 48: Should also test con.res here for completeness, right?

I assume we don't do an ex-equal on the whole res all at once, and instead dig into it like this (here and elsewhere), to get smaller/more precise failure prints?

(Oh wait, this is generated code, so maybe there is no reason. (; )

Fang- (commented):
Very cool. Fair improvements to the setup/structure overall.

Haven't reviewed the .md files closely yet, but output looks good. Tests are particularly well-suited for generation because they sort of naturally gravitate towards the more-verbose, linear style of hoon that the models find easy to write.

Some comments and questions below!

desk/ted/ph/fleet.hoon
mikolajpp: Given how virtual ames driver behaves, you must always include `~fen` and `~dem`, otherwise aqua tests will crash after some time passes when `%notify` and `%reel` generate some poke which can happen regardless if you test them or not due to internal agent logic. So at the very least you must top up the fleet with these two, otherwise it will cause unexpected crashes if you forget to add them to the fleet.

I can think whether we can decouple this thread from its use in aqua tests, but I think for now the UX of that is not great, especially with agentic workflows: you would have to always remember to include four additional ships (providers and their galaxies), and in typical tests this would manifest as hard to decipher virtual vane crashes.

The real fix here is to fix aqua ames driver to stop crashing, but no one bothered to dig into that yet.

Double booting ships because of duplicates in this list is a good point. I am not sure if aqua will just happily oblige, but in any case I will turn the fleet into a set to avoid that.

desk/ted/ph/fleet.hoon
mikolajpp: @yosoyubik gave me the explanation, which sounds reasonable: we can't respond to ames packets in the same event(?) on first contact, so we drop them first, because we don't have the keys, and the keys must be injected after we retrieve them from azimuth/jael. I forgot why exactly it takes that long, rather than being bounded by the speed with which we can retrieve the keys.

@yosoyubik has been working on typed aqua snapshots which allow to tweak vane's subject so we will be able to modify ames timer soon.

desk/tests/ph/app/contacts.hoon
mikolajpp: The robot did actually follow the convention in other (manually written) tests. I have since made better test tooling, we now have `+ex-app-fact`, that does take the whole expected fact, together with wire and dock, and has better error messages while avoiding crashing assertions altogether. 

I think the agent did what is described in the test description, which is to care only whether `~zod` was successfully met. When using `+ex-app-fact`, it would be forced to account for the whole fact at once.

backend/run-aqua-tests.sh
mikolajpp: Yes, we could. I will transition these scripts to do that.

desk/ted/ph/fleet.hoon
mikolajpp: After giving it more thought, I have decided to stop hardcoding this ships. If this causes problems it will just put pressure to fix aqua, which we should do eventually.

mikolajpp:
Closes TLON-5746

Fang-:
Changes look good, thank you. CI isn't happy, but the trace (`/lib/strandio/hoon:<[740 13].[740 27]>`) doesn't make sense for our `/desk/lib/strandio.hoon`. Looks like it could be the one from urbit/urbit's master branch though. Are we copying these in, in the wrong order?

That may not be the root cause though, we may just be trying to eval bad code. Given it's complaining about `@tatid`, could be related to the thread id tweaking.

Fang- (approved):
All green!

Image

PR #5821, opened 2026-05-05
Fix duplicate shortcode reaction chips

wca4a:
## Summary
- Normalize legacy shortcode reactions (for example `:heart:` and `:+1:`) before aggregating reaction chips.
- Reuse normalized reaction values for the reactions detail pane so shortcode/native variants are grouped together.
- Add unit coverage for shortcode/native aggregation and unknown shortcode fallback.

## Tests
- `pnpm --filter @tloncorp/app exec eslint ui/utils/postUtils.tsx ui/utils/postUtils.test.ts`
- `pnpm --filter @tloncorp/app exec vitest run ui/utils/postUtils.test.ts`
- `pnpm --filter @tloncorp/editor build` (needed to generate editorHtml before app typecheck)
- `pnpm --filter @tloncorp/app tsc`

### DIFF
--- a/packages/app/ui/utils/postUtils.test.ts
+++ b/packages/app/ui/utils/postUtils.test.ts
@@ -1,7 +1,11 @@
 import * as db from '@tloncorp/shared/db';
 import { describe, expect, test } from 'vitest';
 
-import { computeReactionDetails } from './postUtils';
+import {
+  computeReactionDetails,
+  groupReactionsByValue,
+  normalizeReactionValue,
+} from './postUtils';
 
 describe('computeReactionDetails', () => {
   describe('contact name resolution', () => {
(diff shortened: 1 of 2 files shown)


wca4a:
@codex review

wca4a:
@claude review once

Fang-:
Important context: backend previously did a migration of legacy shortcodes, and should now canonically always be storing the unicode strings. (See #3949, #4105.) If we're seeing shortcode texts arrive from the backend, that's a smell.

I thought backend was actively sanitizing these on-ingress if detected, but apparently we only log with `Shortcode reaction detected`. I do see some entries for that in posthog...

If current client code always sends raw unicode, I suppose it's possible very old clients might still be sending shortcode reactions?

All that said, a client fix like this should be a temporary band-aid, and from that angle we might not even want it at all: we'd much rather:
1) Do on-ingress sanitization on the backend side so we don't ever store new shortcode reactions, and
2) Run a migration that cleans up stored shortcode reactions in backend state.

Of course, "3) find out who/where/what's still sending shortcode reacts" would be nice-to-have too.

wca4a:
Sure - we've spun countless hours on this bug and someone should just root cause it immediately

jamesacklin:
@Fang- and @arthyn, do we still need this?

arthyn:
@jamesacklin I don't think so

PR #5822, opened 2026-05-05
TlonBotSetupPaneView: use arvos.mp4

jamesacklin:
Traps arvos.mp4 in a circle; works with light mode and dark mode.

https://github.com/user-attachments/assets/cde70270-185f-4500-b44a-b621e22a9f96


dnbrwstr:
I like the animation, layout feels a little odd:
- Log out button jumps around in video
- Content isn't centered vertically (I think it's being pushed down by the logout button, but feels off visually)
- Circle bg feels off brand, not visual language we use elsewhere. Maybe make both the video and the bg bigger and trap in a rounded rect?
- Maybe consider $secondaryText for the text color? Or maybe a different text weight? There's something about the text that feels inconsistent to me

PR #5824, opened 2026-05-06
chat, channels: kill shortcodes again

Fang-:
## Summary

#3949/#4105 moved us off shortcodes into raw unicode representation. Client shenanigans might have caused us to still send shortcodes, so #4967 rejected shortcodes in the client and added logging to it and the backend.

Present-day: people are developing clients without understanding why the data is what it is, and as a result their data IS GARBAGE. But backend only logs, never sanitizes the data it gets. As a result, shortcode reacts are propagating once again.

Here, we sanitize the data we have in state, and on-ingress, and block actions/pokes that make this same mistake.

For client devs: we might want a sister PR that doubles down on the suggested change from TLON-4483 and remove _all_ handling of shortcodes. At most, we should drop them on the floor and not respect them.

> Let's remove all our workarounds to render the incorrectly-sent shortcodes so we can (A) simplify our own logic and (B) track down the source of the problem.

Supersedes #5821.

## Changes

Three main changes:

1. Migrate state to turn shortcodes into emoji, _again_. This is a band-aid.
2. Translate shortcodes into emoji _on-ingress_, on both publisher and subscriber. This, too, is arguably a band-aid.
3. On client agents, _nack_ react actions that look like they might contain some shortcode string. (That is, any string that's >1 chars and starts with `:`.) This lets clients know we don't like the input they're sending, and actually plugs the hole here.

We also remove the backend logging for shortcode detection, since we now transform those readily.

## How did I test?

Simulated requests from the client that set reacts using shortcodes. Prior to rejecting those on the backend, observed that the publisher and subscriber both sanitize shortcodes. After, observed that the action was getting rejected wholesale.

Also ran migration logic on palfun. Observed ~130 such reacts in channels, zero in chat. Time-cost of the migration no different from what it was originally, still on the order of about a second for my old/big ship.

## Risks and impact

  - Channel display

## Rollback plan

Revert.


Fang-:
@codex review, is this better?

Fang-:
@codex review

desk/app/chat.hoon
```diff
@@ -352,14 +352,21 @@
...
   =?  old  ?=(%10 -.old)  (state-10-to-11 old)
   =?  old  ?=(%11 -.old)  (state-11-to-12 old)
-  ?>  ?=(%12 -.old)
+  =?  old  ?=(%12 -.old)  (state-12-to-13 old)
+  ?>  ?=(%13 -.old)
+  =/  x
+    ~>  %bout.[0 'chat migr']
+    =+  (~(run by dms.old) dm:recover-emoji)
+    =+  (~(run by clubs.old) club:recover-emoji)
+    ~
```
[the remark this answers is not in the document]
Fang-: ftr the measurement here was less than half a second, so, idk about "significantly slow" (;

Fang-:
@codex review

desk/app/channels.hoon
```diff
@@ -521,6 +524,13 @@
         =pimp:imp
     ==
   ::
+  ++  state-17-to-18
+    |=  =state-17
+    ^-  state-18
+    %=  state-17  -  %18
+      v-channels  (~(run by v-channels.state-17) channel:recover-emoji:utils)
```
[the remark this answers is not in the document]
Fang-: `voc` could be purged instead, and references don't render references anyway. Cutting this from scope.

desk/app/channels-server.hoon
```diff
@@ -1102,24 +1113,15 @@
         ?~  react.p.result
           [%del-react id.c-post ship.p.result]
         [%add-react id.c-post [ship u.react]:p.result]
-      ::  log shortcode reactions for posts
+      ::  if we get reacts trying to pass shortcodes off as unicode
+      ::  (due to stupid dumb clients putting garbage in the field)
```
mikolajpp:
```suggestion
      ::  (due to dumb clients putting garbage in the field)
```

desk/app/channels-server.hoon
```diff
@@ -1135,24 +1137,15 @@
       =/  post  (get:on-v-posts:c posts.channel id.c-post)
       ?~  post  no-op
       ?:  ?=(%| -.u.post)  no-op
-      ::  log shortcode reactions for replies
+      ::  if we get reacts trying to pass shortcodes off as unicode
+      ::  (due to stupid dumb clients putting garbage in the field)
```
mikolajpp:
```suggestion
      ::  (due to dumb clients putting garbage in the field)
```

desk/app/chat.hoon
```diff
@@ -2126,6 +2174,19 @@
     =^  uid  cu-core
       ?:  |(from-self (lte uid club-eq))  cu-uid
       [uid cu-core]
+    ::  if we get reacts trying to pass shortcodes off as unicode
+    ::  (due to stupid dumb clients putting garbage in the field)
```
mikolajpp:
```suggestion
    ::  (due to dumb clients putting garbage in the field)
```

desk/app/chat.hoon
```diff
@@ -2627,6 +2654,20 @@
     ~>  %spin.['di-ingest-diff']
     ^+  di-core
     =.  last-updated  (~(put ol last-updated) [%ship ship] now.bowl)
+    ::  if we get reacts trying to pass shortcodes off as unicode
+    ::  (due to stupid dumb clients putting garbage in the field)
```
mikolajpp:
```suggestion
    ::  (due to dumb clients putting garbage in the field)
```

mikolajpp (approved):
This looks good. A few nits as a nod to third-party clients. They may be _dumb_, but needn't be stupid :-)

The error spot by the model seems to be partially correct: an `=*` alias is not a _local binding_, but indeed we should update the fields before they get used. This seems to have been addressed in one of the commits.

jamesacklin:
Merge at will

PR #5828, opened 2026-05-07
channels: fix post refs path

arthyn:
## Summary

OTT this was failing on the frontend because we were prefixing it with `v0` since it wasn't in the union

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:


Fang- (approved):
Very classic, nice catch.

PR #5831, opened 2026-05-07
Fix desktop reaction tooltip sizing and contact name resolution

patosullivan:
## Summary

Fixes TLON-5730: the desktop web reaction-hover tooltip rendering as a narrow vertical column with hyphenated ship IDs split across lines, and aligns reaction-tooltip name resolution with the canonical `ContactNameV2` path so nicknames and Calm settings behave consistently. Both fixes are UI-layer only.

## Changes

-   Added `reactionTooltipUtils.ts` with a pure display-plan helper that picks the first three reactor IDs and computes `+N more`. The comma punctuation is rendered in `ReactionsDisplay.tsx`, not in the helper.
-   Updated `ReactionsDisplay.tsx` so the hover tooltip uses an `XStack` with `flexWrap="wrap"`, applies `whiteSpace="nowrap"` + `numberOfLines={1}` + per-fragment `maxWidth` to each name, and caps the tooltip container with `maxWidth` while still allowing short names to shrink to content width. Long ship IDs now wrap between fragments instead of breaking inside a name.
-   Added a web-only viewport clamp gated by `Platform.OS === 'web'` that caps the tooltip width with a CSS `min(...)` expression. The CSS-function string is cast at the boundary so it never leaks onto native.
-   Extracted `resolveContactNameProps()` into a pure `contactNameResolver.ts` and routed both `useContactNameProps()` and rendered `ContactName` through it, so the hook path and the rendered component always produce identical strings.
-   `ReactionsDisplay` now resolves tooltip labels with `useContactName({ contactId, expandLongIds: true })` instead of `users[].name`. Reaction tooltips therefore follow the same Calm `disableNicknames` behavior as `ContactNameV2`. `postUtils.computeReactionDetails()` and the legacy `users[].name` field are intentionally preserved for existing consumers.
-   The shared resolver reads `contact.nickname` (the COALESCE field), so tooltip labels can render nicknames for contacts where only the generated `nickname` is populated — including the `initialContacts` fixtures — when the matching contact is available in app context.
-   Added unit tests for the tooltip helper (`reactionTooltipUtils.test.ts`) and the resolver (`ContactNameV2.test.ts`).
-   Added a `ReactionsDisplay.fixture.tsx` Cosmos fixture covering nine variants for desktop-web visual review, including a `singleShortName` case to catch regressions where a short name like `bill` gets an overly wide tooltip.
-   Added a dedicated `apps/tlon-web/vite.cosmos.config.mts` and pointed React Cosmos at it. This keeps only the Vite plugins needed for fixtures and omits the app `base`, proxy, PWA, and Sentry setup; `react-cosmos-plugin-vite@7.2.0` was not handling the app `base` correctly and was booting the real app inside the renderer iframe.
-   Removed the hardcoded Cosmos `rendererUrl` so the renderer uses the standard Cosmos/Vite entrypoint.

## How did I test?

Automated checks completed:

-   `pnpm --filter @tloncorp/app test reactionTooltipUtils` — 6/6 passing
-   `pnpm --filter @tloncorp/app test ContactNameV2` — 5/5 passing
-   `pnpm --filter @tloncorp/app test postUtils` — 14/14 passing (regression, no expected change)
-   `pnpm -r tsc` — all 9 workspaces clean
-   Manual web check, cosmos check

## Risks and impact

    -   [ ] Onboarding
    -   [ ] State / providers
    -   [ ] Message sync
    -   [x] Channel display
    -   [ ] Notifications

`ContactNameV2` is a canonical shared display component, so the resolver refactor has a broader blast radius than reaction chips even though it is intended to preserve behavior; both the hook (`useContactNameProps`) and the rendered `ContactName` paths now delegate to the same `resolveContactNameProps()`, and `ContactNameV2.test.ts` covers the resolver across nickname, Calm, planet, and comet cases.

`getSequencedChannelPosts()` and other data paths are intentionally untouched. The `noContactRelation` Cosmos variant demonstrates that tooltip names resolve without an eager `reaction.contact` relation when contacts are available through `AppDataContextProvider`; `noContactsAtAll` covers the full-ID fallback.

The Cosmos Vite config is dev-only and does not affect production web builds. The main risk is future drift from `apps/tlon-web/vite.config.mts` if shared plugin requirements change, so it intentionally mirrors only the minimum plugin/alias setup needed for package fixtures.

Implementation note: the pure resolver lives in a sibling `contactNameResolver.ts` rather than inside `ContactNameV2.tsx`, because importing the JSX/Tamagui surface directly in a pure Vitest test pulls in unrelated UI dependencies. `ContactNameV2.tsx` re-exports the resolver so consumers are unaffected.

## Rollback plan

Revert.


patosullivan:
@claude review once

patosullivan:
@claude review once

patosullivan:
@claude review once

packages/app/ui/components/ChatMessage/reactionTooltipUtils.ts
```diff
@@ -0,0 +1,14 @@
+export const TOOLTIP_USER_DISPLAY_COUNT = 3;
```
dnbrwstr: Separate file feels like overkill for this

packages/app/ui/components/ChatMessage/ReactionsDisplay.tsx
```diff
@@ -4,14 +4,90 @@ import { Icon, SizableEmoji, getNativeEmoji } from '@tloncorp/ui';
...
+  getReactionTooltipDisplayPlan,
+} from './reactionTooltipUtils';
+
+// Web-only style; native receives `undefined` and falls back to the numeric maxWidth prop.
+// Tamagui's stricter style typing rejects unknown CSS function strings, so cast at the
+// boundary to keep the clamp web-only without leaking onto native.
+const tooltipWebViewportClamp: ViewStyle | undefined =
+  Platform.OS === 'web'
+    ? ({
+        maxWidth: `min(${TOOLTIP_MAX_WIDTH_PX}px, calc(100vw - ${TOOLTIP_VIEWPORT_GUTTER_PX}px))`,
```
dnbrwstr: Do we really need the `calc`? In what situation will 100vw-32px be less than 280? At that point the layout mostly flips to mobile style, and I'd say just kill the tooltip. Would simplify this a bit to just have a numeric max width.

dnbrwstr (approved):
Lgtm, couple nits.

packages/app/ui/components/ChatMessage/ReactionsDisplay.tsx
patosullivan: Fixed in latest commit.

packages/app/ui/components/ChatMessage/reactionTooltipUtils.ts
patosullivan: Fixed in latest commit.

PR #5836, opened 2026-05-11
fix: close mention popup when sending without selecting a suggestion

patosullivan:
## Summary

Fixes TLON-5709: on desktop/web, typing `hello @all` and then clicking the send button (instead of pressing Enter or selecting a suggestion) sent the message but left `InputMentionPopup` visible above the cleared composer.

Root cause: `useMentions` only updates its search state from `handleTextChange`. The send-button path clears the composer programmatically, bypassing the text-change handler and leaving `isMentionModeActive`, `mentionStartIndex`, `mentionSearchText`, `wasDismissedByEscape`, and `lastDismissedTriggerIndex` stale. The Enter path is unaffected because it routes through `handleSelectMention`, which already resets state.

Adds a `resetMentionMode` helper in `useMentions` and calls it from `sendMessage` (before the await, so selected rich mentions still serialize) and from `handleCancelEditing` (which also now clears `mentions`, fixing a separate case where canceling an edit left stale rich mentions on the next draft).

## Changes

- `packages/app/ui/components/BareChatInput/useMentions.tsx`: added `resetMentionMode` that clears the five mention-search setters (deliberately does not touch the selected `mentions` array, callers decide).
- `packages/app/ui/components/BareChatInput/index.tsx`: call `resetMentionMode()` in `sendMessage` after the composer-clear block and before the network await; call `resetMentionMode()` and `setMentions([])` in `handleCancelEditing`.
- `apps/tlon-web/e2e/helpers.ts`: new exported helper `sendUnselectedMentionMessage` with an optional `containerSelector` for thread scoping. Asserts popup is visible pre-send, the outgoing channel-action payload contains the literal typed text with no `ship`/`sect` inline keys, the rendered Post is visible, the popup is dismissed, and the input is empty.
- `apps/tlon-web/e2e/chat-core-functionality.spec.ts`: two new cases at the end of `should test comprehensive chat functionality` covering main-chat unselected `@all` send and thread-reply unselected `@all` send (bookended with `startThread('Hello, world!')` and `navigateBack`).

Other things to be aware of:

- Clicking send while a mention is active and a suggestion is highlighted still sends the literal text rather than auto-selecting the highlighted suggestion.
- `handleBlur` does not dismiss the popup either.

## How did I test?

- Manual testing on desktop web.
- Confirmed Enter-to-send and clicking a suggestion still work, and that rich mentions selected before sending still serialize into the outgoing post.
- Confirmed cancel-edit no longer leaves stale mentions on the next draft.
- Ran the updated chat-core-functionality.spec.ts e2e test.

## Risks and impact

- Affects important code area: Message input.

## Rollback plan

Revert.

### DIFF
--- a/apps/tlon-web/e2e/chat-core-functionality.spec.ts
+++ b/apps/tlon-web/e2e/chat-core-functionality.spec.ts
@@ -125,6 +125,24 @@ test('should test comprehensive chat functionality', async ({
     expectedMentionInline: { sect: 'admin' },
     postMentionTestId: 'PostMention-admin',
   });
+
+  // Send with the mention popup open but no suggestion selected (TLON-5709):
+  // outgoing payload should be the literal typed text and the popup should dismiss.
+  await helpers.sendUnselectedMentionMessage(zodPage, {
+    inputText: 'TLON-5709 unselected @all',
+    suggestionTestId: '-all--group',
+    expectedLiteralText: /TLON-5709 unselected @all/,
+  });
+
+  // Same case in a thread reply composer.
+  await helpers.startThread(zodPage, 'Hello, world!');
+  await helpers.sendUnselectedMentionMessage(zodPage, {
+    inputText: 'TLON-5709 thread unselected @all',
+    suggestionTestId: '-all--group',
+    expectedLiteralText: /TLON-5709 thread unselected @all/,
+    containerSelector: '#reply-container',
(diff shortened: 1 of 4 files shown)


patosullivan:
@claude review once

dnbrwstr (approved):
Lgtm

PR #5838, opened 2026-05-12
add a2ui message rendering

dnbrwstr:
## Summary

Adds a first pass at A2UI rendering in messages. This keeps the client presentation-only for now, with validation around the small supported subset before we render it.

## Changes

- Add A2UI blob parsing and validation for a v0.9-style component payload.
- Render Text, Row, Column, Card, Divider, and Button blocks in post content.
- Wire supported button events to send visible message text from the draft context.
- Add Cosmos fixtures for weather, approval cards, basic components, and stacked examples.
- Release `@tloncorp/api@0.0.8` with the A2UI types and validation helpers.

## How did I test?

Tested on web + device


packages/api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tloncorp/api",
-  "version": "0.0.6",
+  "version": "0.0.8",
```
dnbrwstr, line 3: 0.0.7 was published but the version bump wasn't committed

dnbrwstr:
@claude review

patosullivan:
Two things stand out:

1) What happens if a2ui cards render in some place where the user shouldn't be able to send anything? Like search results or read-only channels. It looks like in those cases the actions still appear as enabled.

2) How can the user be sure that an a2ui button does what it's supposed to? It seems like someone could label a button anything they want and then assign any action they want, so a bad actor could send a DM with a card that says "Refresh" (or whatever) but it actually maps to some destructive action. I'm thinking we could surface the actual text of the action somewhere (if we want to fix this right now).

dnbrwstr:
> Two things stand out:
> 
> 1. What happens if a2ui cards render in some place where the user shouldn't be able to send anything? Like search results or read-only channels. It looks like in those cases the actions still appear as enabled.

Addressed this by making rendering conditional on ability to start a draft in context.

> 2. How can the user be sure that an a2ui button does what it's supposed to? It seems like someone could label a button anything they want and then assign any action they want, so a bad actor could send a DM with a card that says "Refresh" (or whatever) but it actually maps to some destructive action. I'm thinking we could surface the actual text of the action somewhere (if we want to fix this right now).

Addressed this by only rendering the blocks in dms. Right now, the only thing buttons can do is send a message with given text in the same context, which is innocuous in a dm. We can consider a more elaborate solution later.

patosullivan (approved):
Lgtm!

packages/app/ui/components/PostContent/A2UIBlock.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+  if (hasButton) {
+    return '$m';
+  }
+
+  if (isTextOnly) {
+    return component.component === 'Row' ? '$s' : '$xs';
+  }
+
+  return '$m';
```
arthyn: unless we want some other specific value for buttons maybe just:

```suggestion
  const isTextOnly = component.children.every(
    (child) => components.get(child)?.component === 'Text'
  );

  if (isTextOnly) {
    return component.component === 'Row' ? '$s' : '$xs';
  }

  return '$m';
```

arthyn (approved):
lgtm, do we want to update docs/post-blobs.md? Is the "update surface" message intended to update previous surfaces in the message history or is it just on a per message basis? would love to see more on how that's supposed to work

dnbrwstr:
> lgtm, do we want to update docs/post-blobs.md? Is the "update surface" message intended to update previous surfaces in the message history or is it just on a per message basis? would love to see more on how that's supposed to work

Updated docs -- right now it's just per-message, but we'd like to figure out a way to do live updates in the future.

PR #5839, opened 2026-05-13
BlockRenderer, postContent: parse plain-text markdown tables in messages

jamesacklin:
## Summary

Render GFM-style Markdown tables in messages as real tables instead of pipe-text. Detection runs client-side at receive time — wire format unchanged, old clients still see plaintext.

## Why is this PR so big?

About 1,300 lines net but **most of it is moved code, not new logic**. Here's how to read it:

**~870 lines of genuine new code** — start here:
- `extractTables.ts` (~415 lines) — the splitter
- `extractTables.test.ts` (~290 lines) — 9 unit tests for the splitter
- `groupMentionPlugin.ts` (~85 lines) — mirrors `shipMentionPlugin.ts`
- `postContentInlines.ts` (~80 lines) — `convertInlineContent` extracted out of `postContent.ts` to break a cycle between `convertContent` and `extractTablesFromContent`

**~160 lines of additions to existing files**:
- `BlockRenderer.tsx` — the `TableBlock` component (column-grouped layout, three-phase `onLayout` measurement)
- `postContent.ts` — `TableBlockData` type, `BlockData` union extension, `plaintextPreviewOf` case, single call to `extractTablesFromContent` at the end of `convertContent`
- `mdastToStory.ts` — `isGroupMention` type guard and handler in `phrasingToInlines`

**The rest is a directory move** (`packages/shared/src/logic/markdown/` → `packages/api/src/client/markdown/`).

## Changes

- New `TableBlockData` block type + `TableBlock` renderer (column-grouped layout, two-pass `onLayout` measurement, capped at 280px wide, header-row dim styling).
- Splitter at `packages/api/src/client/markdown/extractTables.ts` finds GFM tables in paragraph content, parses with remark-gfm, emits `TableBlockData`. Uses `mdast-util-to-markdown` for serialization so bold/links/mentions in cells survive intact.
- New `remarkGroupMentions` plugin (parallel to `remarkShipMentions`) so `@all` / `@admin` round-trip through table cells.
- **Moved the entire markdown module from `packages/shared/src/logic/markdown/` → `packages/api/src/client/markdown/`**. api now owns content conversion end-to-end; `convertContent` calls `extractTablesFromContent` directly.

## How did I test?

- 9 new unit tests in `extractTables.test.ts` covering simple extraction, soft-wrap continuation merging, no-outer-pipe tables, bold/link/group-mention preservation in cells, and pre/post-paragraph splitting.
- All 50 existing markdown tests + 51 storyToMarkdown tests + 58 roundTrip tests still pass.
- Typecheck clean for files in this branch.
- `pnpm lint` passes with zero errors (117 warnings, all pre-existing on `develop`).
- Verified visually with a bot post whose cells contained `bold` + `link` inlines.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

No wire-format or backend changes. New code path only fires on paragraphs whose text matches a GFM table signature.

## Rollback plan

Revert the branch. To narrowly disable detection without removing the type, remove the `extractTablesFromContent` call at the end of api's `convertContent`.

## Screenshots / videos

Appearance in chat:
image

Appearance in notebook:
image


jamesacklin:
@claude review once

packages/api/src/client/markdown/extractTables.ts
```diff
@@ -0,0 +1,415 @@
...
+  if (inline.type === 'text') return inline.text;
+  if (inline.type === 'lineBreak') return '\n';
+  const paragraph: Paragraph = {
+    type: 'paragraph',
+    children: [inlineDataToPhrasing(inline)],
+  };
+  const md = toMarkdown(paragraph, {
+    extensions: [gfmToMarkdown(), mentionHandlers],
+  });
+  return md.replace(/\n+$/, '');
```
patosullivan: Kind of an edge case, but it looks like the table could get corrupted if the text happens have a pipe `|` in it (like some valid urbit code or commands). GFM will treat that as a delimiter for the column. Maybe we just need to escape pipes within structured inlines like inline code (or similar).

packages/api/src/client/markdown/extractTables.ts
```diff
@@ -0,0 +1,415 @@
...
+
+    if (!isSeparator) {
+      i++;
+      continue;
+    }
+
+    const headerLineIdx = i - 1;
+    let endLineIdx = i;
+    let j = i + 1;
+    while (j < lines.length && !SEPARATOR_LINE.test(lines[j])) {
```
patosullivan: If a data row happens to look like a separator row after the header delimiter I think this could split a table too early. Maybe we could avoid looking for separator rows after we already found the first one.

packages/api/src/client/markdown/groupMentionPlugin.ts
```diff
@@ -0,0 +1,85 @@
...
+ */
+export interface GroupMention extends Literal {
+  type: 'groupMention';
+  value: string; // group/role name without the @
+}
+
+// `@` followed by an identifier-like token. Lookbehind ensures we don't match
+// inside emails or other adjacent contexts — the @ must be at the start of the
+// string or follow whitespace.
+const GROUP_MENTION_PATTERN = /(?<=^|\s)@([a-z][a-z0-9-]*)/g;
```
patosullivan: This won't catch `(@all)`, `@all,`, or `|@all|`. Last one is bad because that's valid table syntax.

packages/api/src/client/markdown/extractTables.ts
```diff
@@ -0,0 +1,415 @@
...
+        }
+      }
+    }
+    // eslint-disable-next-line no-fallthrough
+    case 'link':
+      return {
+        type: 'link',
+        url: inline.href,
+        children: [{ type: 'text', value: inline.text }],
+      };
```
patosullivan, line 87: Why are we using `// eslint-disable-next-line no-fallthrough` here? It seems like we should just have an explicit default that returns `assertNever(inline.style)` in the inner switch.

packages/api/src/client/markdown/extractTables.ts
```diff
@@ -0,0 +1,415 @@
...
+  if (tree.children.length !== 1) return null;
+  const first = tree.children[0];
+  if (first.type !== 'table') return null;
+  return first as MdastTable;
+}
+
+function extractTablesFromParagraph(block: ParagraphBlockData): BlockData[] {
+  const mapping = inlinesToText(block.content);
+  const regions = findTableRegions(mapping.text);
+  if (regions.length === 0) return [block];
```
patosullivan, line 384: Since this runs on every paragraph on every received message we should probably short-circuit here if there's no chance that the paragraph contains a markdown table (no pipe).


```suggestion
function extractTablesFromParagraph(block: ParagraphBlockData): BlockData[] {
  const hasPipe = block.content.some(
    (i) => i.type === 'text && i.text.includes('|')
  );
  
  if (!hasPipe) return [block];

  const mapping = inlinesToText(block.content);
  const regions = findTableRegions(mapping.text);
  if (regions.length === 0) return [block];
```

packages/api/src/client/markdown/extractTables.ts
```diff
@@ -0,0 +1,415 @@
...
+      // placeholders are single-character).
+      const sliceStart = Math.max(0, offsetStart - pos.start);
+      const sliceEnd = Math.min(pos.inline.text.length, offsetEnd - pos.start);
+      const slicedText = pos.inline.text.slice(sliceStart, sliceEnd);
+      if (slicedText) {
+        result.push({ type: 'text', text: slicedText });
+      }
+    }
+    // Non-text inlines that aren't fully contained are skipped — but this
+    // shouldn't happen since they're single-character placeholders.
```
patosullivan: I don't think these comments are right.

packages/app/ui/components/PostContent/BlockRenderer.tsx
```diff
@@ -678,6 +685,157 @@ export const HeaderText = styled(Text, {
...
 HeaderText.displayName = 'HeaderText';
 
+function alignToTextAlign(
+  align: cn.TableAlignment | null | undefined
+): 'left' | 'center' | 'right' | 'auto' {
+  if (align === 'left' || align === 'center' || align === 'right') return align;
+  return 'auto';
+}
+
+const TABLE_MAX_COLUMN_WIDTH = 280;
```
patosullivan, line 695: Will this width look good on desktop web?

packages/api/src/client/postContent.ts
```diff
@@ -263,6 +283,8 @@ export function plaintextPreviewOf(
           return plaintextPreviewOfListData(block.list, config);
         case 'bigEmoji':
           return block.emoji;
+        case 'table':
+          return '(Table)';
```
patosullivan: Maybe we return just the first cell of the header or something here

patosullivan (commented):
Overall looks good, but there are a few issues I think should probably be addressed before we merge. Most are pretty minor.

packages/app/ui/components/PostContent/BlockRenderer.tsx
jamesacklin: Check the screenshot, that's so we wrap long cell contents at some reasonable width

packages/api/src/client/markdown/extractTables.ts
jamesacklin: [aa1db9e](https://github.com/tloncorp/tlon-apps/pull/5839/commits/aa1db9eb0ecb00f6e03643bcd758692c230d6844)

packages/api/src/client/markdown/extractTables.ts
jamesacklin: [7051479](https://github.com/tloncorp/tlon-apps/pull/5839/commits/705147962d400821189404da4e0cc887f1861f11)

packages/api/src/client/markdown/extractTables.ts
jamesacklin: [7051479](https://github.com/tloncorp/tlon-apps/pull/5839/commits/705147962d400821189404da4e0cc887f1861f11)

packages/api/src/client/markdown/groupMentionPlugin.ts
jamesacklin: [a216e30](https://github.com/tloncorp/tlon-apps/pull/5839/commits/a216e30af25920ca36297ee8c2be8d4263383314)

packages/api/src/client/markdown/extractTables.ts
jamesacklin: [537bf87](https://github.com/tloncorp/tlon-apps/pull/5839/commits/537bf87f13d4da15915a2423e6b8b37fa1fe8403)

packages/api/src/client/markdown/extractTables.ts
jamesacklin: [817fbb0](https://github.com/tloncorp/tlon-apps/pull/5839/commits/817fbb065073820471e5eba43b49f4d14d498393)

packages/api/src/client/postContent.ts
jamesacklin: [57c00ae](https://github.com/tloncorp/tlon-apps/pull/5839/commits/57c00ae64c293f29a390696fc379abd8385491ed)

packages/app/ui/components/PostContent/BlockRenderer.tsx
patosullivan: Totally missed the screenshot!

patosullivan (approved):
lgtm

PR #5841, opened 2026-05-13
lure: ensure metadata gaps are filled when generating links

arthyn:
## Summary

%reel had a bug whereby empty nicknames wouldn't fill with any existing group title, minor but still good to fix. The other piece here is making sure we always fill in any gaps in the metadata. %reel has drifted into being more heavily integrated into groups/contacts and this just furthers that by making sure any kind of reel describe calls actually fill in gaps in the metadata from what we have on the backend. This was present in https://github.com/tloncorp/ylem/blob/8af770c82083f1031c84b72a877a7b8ad226c56f/pkg/runtime/pioneer/src/Pioneer/Command/Click.hs#L1537 where we're only giving `userId`, `type`, and `groupId`.

Fixes TLON-5790

## Changes

- **`%reel-describe` gap-filling** (`desk/app/reel.hoon`): when a caller's describe is sparse, backfill missing open-graph fields from server state — `invitedGroupTitle` / `invitedGroupDescription` / `invitedGroupIconImageUrl` scried from `%groups`, and `inviterNickname` / `inviterAvatarImage` read from reel's cached `our-profile`. Caller-provided values always win; the group scry is wrapped in `mole` so a missing group doesn't crash the poke.
- **"a Groupchat" fallback fix**: the no-nickname branch of the OG-title formula was hard-coding "You're Invited to a Groupchat" even when the group had a real title. It now uses `group-title` (which itself still falls back to "a Groupchat" when the title is empty).
- **`group-og-title` helper**: extracted the formula shared by the `/contacts` and `/groups` subscription handlers into a single helper in the upper `|%` core. No behavior change.
- **Tests** (`desk/tests/app/reel.hoon`): added a scry mock returning an empty `group:v9:gv` for the gap-fill path, wired it into the three tests that exercise `%reel-describe` (`test-reel-describe`, `test-groups-update`, `test-contacts-update`). Updated the `test-groups-update` `$og_title` expectation to reflect the corrected fallback (`"You're Invited to Early Sunrise"` instead of `"…a Groupchat"`).

## How did I test?

Ran `./backend/run-tests.sh` locally against a moon— all three reel tests pass (`test-reel-describe`, `test-groups-update`, `test-contacts-update`). No e2e or client changes

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: lure / invite-link metadata (`%reel` agent)

## Rollback plan

Revert the merge commit. The agent state-version is unchanged, so no migration concerns. Any lures registered while this is live keep their backfilled metadata fields, which is forward-compatible with the prior code path (extra fields are just ignored downstream if reverted).

## Screenshots / videos

N/A — backend / Hoon only. The user-visible effect is open-graph previews on shared invite links: titles like *"Tlon Messenger: You're Invited to &lt;Group Title&gt;"* instead of *"…to a Groupchat"* when the inviter has no nickname yet.

### DIFF
--- a/desk/app/reel.hoon
+++ b/desk/app/reel.hoon
@@ -117,6 +117,19 @@
     |*  [caz=(list card) etc=*]
     [[cad caz] etc]
   --
+::  +group-og-title: render the open-graph title for a group invite.
+::  empty group title falls back to "a Groupchat".
+::
+++  group-og-title
+  |=  [nickname=(unit @t) group-title=@t]
+  ^-  @t
+  =/  gt=@t
+    ?:  =('' group-title)  'a Groupchat'
+    group-title
+  %-  crip
+  ?:  |(?=(~ nickname) =('' u.nickname))
+    "Tlon Messenger: You're Invited to {(trip gt)}"
+  "Tlon Messenger: {(trip u.nickname)} invited you to {(trip gt)}"
 --
 |_  =bowl:gall
 +*  this  .
@@ -445,16 +458,9 @@
         ::
         =+  type=(~(get by fields.meta) %'inviteType')
         ?:  |(?=(~ type) =('group' u.type))
-          =/  group-title=@t
-            =+  til=(~(get by fields.meta) %'invitedGroupTitle')
-            ?:  |(?=(~ til) =('' u.til))
-              'a Groupchat'
-            u.til
           =/  title=@t
-            %-  crip
-            ?:  |(?=(~ nickname) =('' u.nickname))
-              "Tlon Messenger: You're Invited to {(trip group-title)}"
-            "Tlon Messenger: {(trip u.nickname)} invited you to {(trip group-title)}"
+            %+  group-og-title  nickname
+            (fall (~(get by fields.meta) %'invitedGroupTitle') '')
           =.  fields.update
             (~(put by fields.update) %'$og_title' title)
           =.  fields.update
@@ -499,15 +505,9 @@
         ?+    -.r-group.r-groups  ~
             %meta
           =*  meta  meta.r-group.r-groups
-          =+  nickname=(~(get cy:t our-profile) %nickname %text)
-          =/  group-title=@t
-            ?:  =('' title.meta)  'a Groupchat'
-            title.meta
           =/  title=@t
-            %-  crip
-            ?:  |(?=(~ nickname) =('' u.nickname))
-              "Tlon Messenger: You're Invited to {(trip group-title)}"
-            "Tlon Messenger: {(trip u.nickname)} invited you to {(trip group-title)}"
+            %-  group-og-title
+            [(~(get cy:t our-profile) %nickname %text) title.meta]
           %-  ~(gas by *(map field:reel cord))
           :~  %'invitedGroupTitle'^title.meta
               %'invitedGroupDescription'^description.meta
--- a/desk/app/reel.hoon
+++ b/desk/app/reel.hoon
@@ -230,6 +230,57 @@
           [%'inviterUserId' (scot %p src.bowl)]
           [%'invitedGroupId' id]
       ==
+    ::  gap-fill open-graph metadata from server state when the caller
+    ::  didn't provide it: title / description / image from %groups,
+    ::  and nickname / avatar from our cached profile. Caller-provided
+    ::  values always win; scry failures are tolerated.
+    ::
+    =/  type=(unit @t)  (~(get by fields.metadata) %'inviteType')
+    =?  fields.metadata  |(?=(~ type) =('group' u.type))
+      =/  fld  fields.metadata
+      ::  inviter fields from our-profile
+      ::
+      =/  nickname=(unit @t)  (~(get cy:t our-profile) %nickname %text)
+      =/  avatar=(unit @t)    (~(get cy:t our-profile) %avatar %look)
+      =?  fld
+          ?&  ?=(^ nickname)
+              !=('' u.nickname)
+              =(~ (~(get by fld) %'inviterNickname'))
+          ==
+        (~(put by fld) %'inviterNickname' u.nickname)
+      =?  fld
+          ?&  ?=(^ avatar)
+              !=('' u.avatar)
+              =(~ (~(get by fld) %'inviterAvatarImage'))
+          ==
+        (~(put by fld) %'inviterAvatarImage' u.avatar)
+      ::  group meta from %groups; only attempt when id parses as a flag
+      ::
+      =/  parsed=(unit flag:v0:groups-ver)  (rush id flag)
+      ?~  parsed  fld
+      =/  grp=(unit group:v9:groups-ver)
+        %-  mole  |.
+        .^  group:v9:groups-ver  %gx
+          /(scot %p our.bowl)/groups/(scot %da now.bowl)/v2/groups/(scot %p p.u.parsed)/[q.u.parsed]/group-2
+        ==
+      ?~  grp  fld
+      =*  gmeta  meta.u.grp
+      =?  fld
+          ?&  !=('' title.gmeta)
+              =(~ (~(get by fld) %'invitedGroupTitle'))
+          ==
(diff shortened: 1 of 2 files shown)


desk/app/reel.hoon
```diff
@@ -230,6 +243,57 @@
...
+        (~(put by fld) %'inviterAvatarImage' u.avatar)
+      ::  group meta from %groups; only attempt when id parses as a flag
+      ::
+      =/  parsed=(unit flag:v0:groups-ver)  (rush id flag)
+      ?~  parsed  fld
+      =/  grp=(unit group:v9:groups-ver)
+        %-  mole  |.
+        .^  group:v9:groups-ver  %gx
+          /(scot %p our.bowl)/groups/(scot %da now.bowl)/v2/groups/(scot %p p.u.parsed)/[q.u.parsed]/group-2
+        ==
```
Fang-: Did you test the failure case for these? Pretty sure you still can't catch/virtualize `.^` crashes in userspace, so this will blow up. If you want to harden, you must use (maybe even implement) the relevant `%u` scry endpoints in groups agent instead.

(Good one for the gall-2026 nice-to-haves list, gall auto-filling the `%u` response based on whether the `%x` would succeed.)

desk/app/reel.hoon
```diff
@@ -230,6 +243,57 @@
...
+    =?  fields.metadata  |(?=(~ type) =('group' u.type))
+      =/  fld  fields.metadata
+      ::  inviter fields from our-profile
+      ::
+      =/  nickname=(unit @t)  (~(get cy:t our-profile) %nickname %text)
+      =/  avatar=(unit @t)    (~(get cy:t our-profile) %avatar %look)
+      =?  fld
+          ?&  ?=(^ nickname)
+              !=('' u.nickname)
+              =(~ (~(get by fld) %'inviterNickname'))
```
Fang-:
```suggestion
              !(~(has by fld) %'inviterNickname')
```

Here and elsewhere. Alternatively, maybe you want to check these fields for the `''` empty string value too, in case that snuck in?

Fang- (commented):
Some safety concerns for the stuff you tried making safe. Looks good otherwise!

desk/app/reel.hoon
```diff
@@ -117,6 +117,19 @@
...
+++  group-og-title
+  |=  [nickname=(unit @t) group-title=@t]
+  ^-  @t
+  =/  gt=@t
+    ?:  =('' group-title)  'a Groupchat'
+    group-title
+  %-  crip
+  ?:  |(?=(~ nickname) =('' u.nickname))
+    "Tlon Messenger: You're Invited to {(trip gt)}"
+  "Tlon Messenger: {(trip u.nickname)} invited you to {(trip gt)}"
```
mikolajpp:
```suggestion
  =/  title=@t
    ?:  =('' group-title)  'a Groupchat'
    group-title
  %-  crip
  ?:  |(?=(~ nickname) =('' u.nickname))
    "Tlon Messenger: You're Invited to {(trip title)}"
  "Tlon Messenger: {(trip u.nickname)} invited you to {(trip title)}"
```

desk/app/reel.hoon
```diff
@@ -230,6 +243,66 @@
           [%'inviterUserId' (scot %p src.bowl)]
           [%'invitedGroupId' id]
       ==
+    ::  gap-fill open-graph metadata from server state when the caller
+    ::  didn't provide it: title / description / image from %groups,
+    ::  and nickname / avatar from our cached profile. Caller-provided
+    ::  values always win.
+    ::
```
mikolajpp, line 250:
```suggestion
    ::  gap-fill open-graph metadata from server state when the caller
    ::  didn't provide it: title / description / image from %groups,
    ::  and nickname / avatar from our cached profile. caller-provided
    ::  values take priority.
    ::
```

desk/app/reel.hoon
```diff
@@ -230,6 +243,66 @@
...
+    =/  type=(unit @t)  (~(get by fields.metadata) %'inviteType')
+    =?  fields.metadata  |(?=(~ type) =('group' u.type))
+      =/  fld  fields.metadata
```
mikolajpp: This does not feel idiomatic, one would just an alias here
```suggestion
      =*  fields  fields.metadata
```

desk/app/reel.hoon
```diff
@@ -230,6 +243,66 @@
...
+      ::  treat absent or empty-string fields as gap-fillable.
```
mikolajpp:
```suggestion
      ::  treat absent or empty-string fields as gap-fillable
```

desk/app/reel.hoon
```diff
@@ -230,6 +243,66 @@
...
+      ::
+      =/  blank
+        |=  [m=(map cord cord) k=cord]
+        ^-  ?
+        =/  v  (~(get by m) k)
+        ?|(?=(~ v) =('' u.v))
```
mikolajpp: This should be called `is-blank` for clarity. We should also tighten the arguments. The key of this map is not just any cord, but the specific set of keys from reel metadata.

Not a big fan of single letter variable names. This used to be reserved for standard library functions. CC @Fang- for second opinion.

desk/app/reel.hoon
```diff
@@ -230,6 +243,66 @@
...
+          ?&  ?=(^ avatar)
+              !=('' u.avatar)
+              (blank fld %'inviterAvatarImage')
+          ==
+        (~(put by fld) %'inviterAvatarImage' u.avatar)
+      ::  group meta from %groups; only attempt when id parses as a flag
+      ::  and the group exists locally.
+      ::
+      =/  parsed=(unit flag:v0:groups-ver)  (rush id flag)
+      ?~  parsed  fld
```
mikolajpp:
```suggestion
      ?~  parsed=(rush id flag)  fld
```

desk/app/reel.hoon
```diff
@@ -230,6 +243,66 @@
...
+      =/  base  /(scot %p our.bowl)/groups/(scot %da now.bowl)
+      ?.  .^(? %gu (weld base /groups/(scot %p p.u.parsed)/[q.u.parsed]))
+        fld
+      =/  grp=group:v9:groups-ver
+        .^  group:v9:groups-ver  %gx
+          (weld base /v2/groups/(scot %p p.u.parsed)/[q.u.parsed]/group-2)
+        ==
+      =*  gmeta  meta.grp
```
mikolajpp:
```suggestion
      =*  meta  meta.grp
```

desk/tests/app/reel.hoon
```diff
@@ -46,6 +46,20 @@
   %-  ~(gas by *contact:t)
   :~  %nickname^text+'Sampel Palnet'
   ==
+::  scry mock for the gap-fill path in %reel-describe. the %gu check
+::  reports the group as present so the %gx fetch proceeds; the %gx
+::  returns an empty group so the agent's gap-fill checks
+::  (!=('' title.gmeta) etc.) all fall through and no fields are added.
+::
```
mikolajpp: I think these scry paths are generally usable, it is not just for mocking, though we only use it for that now. I would remove this comment and think about specializing the group scry with a special case for a specific group that would indeed have an empty title.

mikolajpp (approved):
The changes look good, the logic is sound.
 A few coding nits, but that's about it!

desk/app/reel.hoon
Fang-: Eh, it's a super narrow function with straightforward logic, ultra-lapidary is perfectly fine here imo. (I agree `is-blank` is slightly clearer.)

We always pass `fld` as the map though. Especially in this context, might be nicer to use that directly instead of passing it as an argument.

(Maybe that gets you into trouble if you make edits and then later check the value again? But we don't do that. `=*  blank` wouldn't suffer from it either way.)

Fang- (approved):
Changes look good, thank you. Concur with @mikolajpp's comments.

PR #5845, opened 2026-05-14
SplashSequence: avoid layout pushing on Tlonbot nickname entry

jamesacklin:
## Summary

Fixes an issue @Fang- found in Tlonbot onboarding (in `SplashSequence`) where if one unfocused the nickname field on Android, the layout pushed up and above the device screen's top boundary.

## Changes

- Adds `behavior="height"` on the `KeyboardAvoidingView` for the nickname pane. The shared wrapper defaulted to `behavior="position"` on Android, which translated all content upward by the full keyboard height (offset is 0), pushing the heading off-screen.

- Adds a ref + `onBlur` that immediately re-focuses and the keyboard's "done" key doesn't trigger a blur. The input now stays focused whenever the pane is mounted, so the buggy focus/blur transition can't be triggered at all. Navigation away still works — the unmount runs before any refocus has a target.

## How did I test?

Built to iOS simulator and confirmed no regressions; confirmed blurring the field is impossible.

## Risks and impact

  - [x] Onboarding

## Rollback plan

Revert

### DIFF
--- a/packages/app/ui/components/Wayfinding/SplashSequence.tsx
+++ b/packages/app/ui/components/Wayfinding/SplashSequence.tsx
@@ -75,7 +75,7 @@ import {
 import { useStore } from '../../contexts/storeContext';
 import { useSystemContactSearch } from '../../hooks/systemContactSorters';
 import AttachmentSheet from '../AttachmentSheet';
-import { Field, TextInput } from '../Form';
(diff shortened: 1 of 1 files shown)


dnbrwstr (approved):
Lgtm

PR #5846, opened 2026-05-14
[codex] Add browser notifications

wca4a:
## Summary

Adds true browser Notification API support for Tlon web while the tab is open or backgrounded.

- subscribes the web app to `%activity` updates and emits native browser notifications for `shouldNotify` activity events
- adds a browser notification permission hook with secure-context/support guards
- exposes a browser notification permission row in notification settings using the current `master` settings UI

## Notes

This does not implement closed-tab Web Push. That would require service worker `push` handling plus PushManager/VAPID/provider support.

The remote does not have a branch named `stable`, so this branch is based on `origin/master`, which appears to be the stable/release branch.

## Validation

- `git diff --check` passed
- On the original working branch before rebasing to `master`: `pnpm --filter 'tlon-web' tsc` passed and `pnpm --filter 'tlon-web' build` passed
- After rebasing to `origin/master`, `pnpm --filter 'tlon-web' tsc --pretty false` is blocked by existing stable-branch/local dependency issues unrelated to this patch, including `@urbit/aura` export mismatches and missing native/web dependency typings
- After reinstalling `master` dependencies with `pnpm install --frozen-lockfile --force`, `pnpm --filter 'tlon-web' build` starts and transforms the app but is blocked by an existing `expo-audio` web export mismatch (`PLAYBACK_STATUS_UPDATE` is not exported by `ExpoAudio.web.js`)

## Manual Test

Verified locally against `https://malmur-halmex.tlon.network` through the web dev server. Browser permission state appears under Settings -> Notification settings -> Browser notifications, and blocked origins report `Blocked in browser`.

### DIFF
--- a/apps/tlon-web/src/app.tsx
+++ b/apps/tlon-web/src/app.tsx
@@ -8,6 +8,7 @@ import {
   Route,
 } from '@react-navigation/native';
 import { ENABLED_LOGGERS } from '@tloncorp/app/constants';
+import useBrowserNotifications from '@tloncorp/app/hooks/useBrowserNotifications';
 import { useConfigureUrbitClient } from '@tloncorp/app/hooks/useConfigureUrbitClient';
 import { useCurrentUserId } from '@tloncorp/app/hooks/useCurrentUser';
 import useDesktopNotifications from '@tloncorp/app/hooks/useDesktopNotifications';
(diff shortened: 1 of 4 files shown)


jamesacklin:
Will want to target `develop` with this.

patosullivan:
This avoids a big issue with the way we did it in landscape before: we'd check/ask for permission on each notification, then store that preference in settings store. If settings store told us that the user had already granted permission, we wouldn't ask again. This was a problem if you wanted to use the notifications on multiple browsers/machines.

This PR just checks the browser permission, that's it. Seems a lot better.

The biggest things missing, imo:

1) disable notifications while the app is foregrounded
2) need an unsubscribe/lifecycle cleanup in the hook where we call `api.subscribeToActivity`.
3) (maybe just nice to have): clicking on a notification should route you to whichever conversation produced that notification

And yeah, this should target develop, not master.

jamesacklin:
Taking this over

PR #5848, opened 2026-05-14
devex: route script-shell via TLON_SHELL so windows can use git bash

jamesacklin:
## Summary

Several package.json scripts use Unix shell tools (`rm -rf`, `cp -R`, `mkdir -p`) that fail on Windows' default `cmd.exe`. Pointing pnpm's script-shell at `${TLON_SHELL-/bin/bash}` lets Windows contributors set `TLON_SHELL` to Git Bash without rewriting any existing scripts; Unix users keep `/bin/bash` automatically via the default-value expansion. Documents the Windows setup in `CLAUDE.md`, including the WSL-bash trap on `PATH` and the Node 22 pin needed for prebuilt `better-sqlite3` binaries.

## Changes

- Tells pnpm to run all package.json scripts through whatever $TLON_SHELL points to, falling back to /bin/bash when the env var is unset
- Updates CLAUDE.md to tell the robots what to do when initializing on Windows

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: dev environment

## Rollback plan

git revert

## Screenshots

image

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 engine-strict=true
 node-linker=hoisted
+script-shell=${TLON_SHELL-/bin/bash}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -222,6 +222,19 @@ To identify which component to modify:
 -   Builds on top of web application
 -   Uses Electron for native desktop features
 
+### Windows Development
+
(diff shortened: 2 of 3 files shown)


dnbrwstr (approved):
Lgtm

PR #5849, opened 2026-05-15
limit large video uploads

dnbrwstr:
## Summary

Add a 150 MB video limit and apply it across the in-app picker, share intents, and the iOS share extension. This also tightens video detection for sparse media picker assets, surfaces attach failures to the user, and keeps share-extension uploads from failing just because thumbnail generation fails.

Fixes TLON-5793

## Changes

- Add shared video validation copy for too-large, unreadable, and unsupported videos
- Treat sparse media picker video assets as videos before validating them
- Show share intent attach errors in the channel UI
- Stop the iOS share extension from copying oversized videos and keep sharing usable when thumbnail generation fails
- Regenerate the expo-share-intent patch so clean native generation keeps the ShareViewController changes

## How did I test?

Tested on device


patches/expo-share-intent@5.1.1.patch
dnbrwstr, line 1: Large diff in this fie is because it wasn't regenerated after the update to `5.1.1`. This should have no effect in practice -- it just gets us to a generated `ShareViewController` which matches the one we've actually been using.

patosullivan (approved):
Lgtm

arthyn (commented):
makes sense to limit for share extension and memex, but I really think we shouldn't limit uploads where we can, like in web generally or in file picker on mobile. we should probably let self-hosters upload w/e size up to the S3 single request cap ~5gb.

dnbrwstr:
> makes sense to limit for share extension and memex, but I really think we shouldn't limit uploads where we can, like in web generally or in file picker on mobile. we should probably let self-hosters upload w/e size up to the S3 single request cap ~5gb.

I removed the limitation on web. Agree it would be nicer not to cap, but on mobile right now the experience is not great + pretty prone to failure.

PR #5850, opened 2026-05-18
gallery: @mention-pattern reply action for comments, blue dot

jamesacklin:
## Summary

Adds a "reply" action to Gallery comments that pre-fills the comment author's @p into the draft, allowing the user to mention the comment author directly. Fixes TLON-4695 where @HoonMasterWalt felt strange replying to the entire comment thread rather than a single comment.

Also adds an unread or notification marker for qualifying posts in Galleries, something we've been missing for some time.

## Changes

- Adds a `prependMentionToDraft` method to `MessageActions` which accepts an input @p and pre-fills a mention into the draft input context. 
- Makes use of this method in `replyToComment` for Gallery comment threads. I could foresee this being useful elsewhere, but is limited to this application for now.
- Adds a notification-aware `UnreadDot` to Gallery blocks. Appears gray if new comments are present, is blue if there's a notification-qualifying reply in there.

## How did I test?

Interactively, on desktop/web.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

git revert

## Screenshots / videos

https://github.com/user-attachments/assets/d3160561-0be1-4e65-830d-c42bb96e9117

### DIFF
--- a/packages/app/ui/components/GalleryPost/GalleryPost.tsx
+++ b/packages/app/ui/components/GalleryPost/GalleryPost.tsx
@@ -303,7 +303,14 @@ export function GalleryPostFooter({
     }
   }, [isWindowNarrow]);
 
-  const { data: threadUnread } = store.useLiveThreadUnreadByParentId(post.id);
+  // Ride the ['post', id] per-post invalidation channel instead of the
+  // table-wide one — the gallery grid mounts one footer per post, so a
+  // table-level invalidation would refetch every footer on any thread_unreads
+  // write. See db/changeListener.ts.
+  const { data: postWithUnread } = store.usePostWithThreadUnreads({
+    id: post.id,
+  });
+  const threadUnread = postWithUnread?.threadUnread;
   const unreadCount = threadUnread?.count ?? 0;
   const isNotify = threadUnread?.notify ?? false;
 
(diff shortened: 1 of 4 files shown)


dnbrwstr:
@claude review once

jamesacklin:
Updated with e2e coverage.

packages/app/ui/components/GalleryPost/GalleryPost.tsx
```diff
@@ -300,6 +303,10 @@ export function GalleryPostFooter({
     }
   }, [isWindowNarrow]);
 
+  const { data: threadUnread } = store.useLiveThreadUnreadByParentId(post.id);
```
patosullivan: Afaik this would be the only place in the app where we use this hook on the list itself. Could be some perf hit here. Not a blocker imo since galleries are pretty low traffic anyway, but if we run into issues later we might wanna check this.

patosullivan (approved):
lgtm

packages/app/ui/components/GalleryPost/GalleryPost.tsx
jamesacklin: Thanks -  I just used `usePostWithThreadUnreads({ id: post.id })` and read `data?.threadUnread` off it. Can you take another look?

patosullivan (approved):
Lgtm!

PR #5852, opened 2026-05-19
BlockRenderer: fix table rendering on mobile

jamesacklin:
## Summary

Fixes TLON-5840 by making several changes to the Markdown table renderer in `BlockRenderer`.

## Changes

1. Implements a simpler cell measurement pass by (1) swapping from a column-major layout to a row-major layout and (2) setting a default width for cells and then measuring/adjusting rather than using a state machine. This cuts down on measurement time significantly and therefore...
2. Eliminates the opacity switch when finished measuring. We were seeing the table not initially appear in TLON-5840 because React Native was not properly retargeting the element ref and would miss the `opacity=1` switch.
3. Uses `react-native-gesture-handler` for a horizontal ScrollView around the table element within the Chat scroller. The Tamagui JS events in `ScrollView` were getting eaten by the parent element; this uses native gestures and works.

## How did I test?

Reproduced the issues in the iOS simulator using an existing Markdown table which rendered fine on the web. Applied fixes, re-tested in sim, confirmed working.

## Risks and impact

  - [x] Channel display
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

git revert

## Screenshots / videos

Forthcoming.

### DIFF
--- a/packages/app/ui/components/PostContent/BlockRenderer.tsx
+++ b/packages/app/ui/components/PostContent/BlockRenderer.tsx
@@ -30,6 +30,7 @@ import {
   Linking,
   Platform,
 } from 'react-native';
+import { ScrollView as GHScrollView } from 'react-native-gesture-handler';
 import { ScrollView, View, ViewStyle, XStack, YStack, styled } from 'tamagui';
 
 import { useNowPlayingController } from '../../contexts/nowPlaying';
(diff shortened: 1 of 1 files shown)


jamesacklin:
@claude review once

dnbrwstr (approved):
Code looks good to me, but I'd verify that the interaction is smooth and doesn't mess with other gestures on both iOS and Android -- definitely potential for weirdness there.

patosullivan (approved):

PR #5856, opened 2026-05-19
fix stuck sends on missing poke ack

dnbrwstr:
## Summary

This makes plain pokes time out instead of keeping the send queue stuck forever. When a send times out waiting for the poke ack, we mark the optimistic post as `needs_verification` so sync can decide whether the ship accepted it.

Fixes TLON-5844

## Changes

- Add a poke ack timeout for plain `Urbit.poke()`.
- Treat poke ack timeout as a connection-related send error, moving the post to `needs_verification`.
- Add focused send/queue logging for blocked offline, queue action started, API send started, and queue failures.

## How did I test?

Tested on device

### DIFF
--- a/packages/api/src/http-api/Urbit.ts
+++ b/packages/api/src/http-api/Urbit.ts
@@ -559,7 +559,7 @@ export class Urbit {
     });
 
     this.outstandingPokes.forEach((poke) => {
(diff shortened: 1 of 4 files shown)


patosullivan (approved):

dnbrwstr:
@claude review once

PR #5857, opened 2026-05-20
mobile: fix push notif rendering

latter-bolden:
The problem here was #5839 adding a new markdown parsing dependency that crashes when run from the `scripts` bundle — our minimalist JS target that gets run natively on iOS via [JavaScriptCore](https://developer.apple.com/documentation/javascriptcore).

The runtime path we use from `scripts` is `getTextContent`. That helper is contained in a module that requires `extractTables`, which expects the DOM to be available at runtime and crashes if it's not.

I think https://github.com/tloncorp/tlon-apps/pull/5857/commits/930b2fa8c11d566e8167608ec5ebbc15a11e7474 alone fixes the immediate problem by changing the `esbuild` conditions to target RN. However, respecting that is opt-in at the module level so simply configuring the build command doesn't prevent similar issues in the future.

The other commits add a new `postNotificationText` helper that's intentionally limited in what it imports and is placed within the `scripts` module itself. I don't love having another "turn raw message into text string" helper, but I think the alternative is a thorny disentangle of `postContent.ts` that might regress down the line.

Implementation wise, the new helper produced identical output against post fixtures and has seemed fine anecdotally in the Preview App. However, if we're concerned about the added helper, we can leave it out and exclusively merge the `esbuild` fix.

Fixes TLON-5843


jamesacklin:
@claude review once

jamesacklin (approved):

dnbrwstr (approved):
Lgtm

PR #5859, opened 2026-05-20
BlockRenderer: stabilize table column measurement on mobile

jamesacklin:
## Summary

Markdown tables on mobile were rendering each row's cells at their natural width instead of snapping to the column's max width. Two compounding bugs: 

1. The measurement reset was triggering on every `block` reference change (which happens constantly from upstream data churn, even when the table content is unchanged), wiping the in-progress measurement cache before all cells had reported
2. Measurement itself relied on `onLayout` events that could be missed during busy renders.

## Changes

1. Reset only when a content fingerprint actually changes
2. Measure cells imperatively via `View.measure()` after first paint instead of waiting for events

## How did I test?

1. Built to device (preview app), see bug
2. Apply fix, build to device again, see bug gone

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

git revert

## Screenshots / videos

| Before | After |
|--------|-------|
| IMG_7520 | Screenshot 2026-05-20 at 8 21 49 AM |

### DIFF
--- a/packages/app/ui/components/PostContent/BlockRenderer.tsx
+++ b/packages/app/ui/components/PostContent/BlockRenderer.tsx
@@ -19,16 +19,16 @@ import React, {
   memo,
   useCallback,
   useContext,
-  useEffect,
+  useLayoutEffect,
   useMemo,
   useRef,
   useState,
 } from 'react';
 import {
   ActivityIndicator,
-  type LayoutChangeEvent,
   Linking,
   Platform,
+  View as RNView,
 } from 'react-native';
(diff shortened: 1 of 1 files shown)


jamesacklin:
@claude review once

jamesacklin:
@dnbrwstr Addressed all of the Claude PR feedback, feel free to take a look

dnbrwstr (approved):
Lgtm

PR #5860, opened 2026-05-20
Add A2UI chat action confirmation

wca4a:
## Summary
- render A2UI blobs in chat-like contexts beyond DMs
- add A2UI button handling for `tlon.sendMessage` actions
- scope the initial action UX to post-ingest lore compile-request cards
- require a confirmation dialog before sending the compile confirmation text
- disable A2UI actions when the current context cannot draft/send
- add fixtures, docs, and unit coverage for action helpers

## UX
For this slice, the button belongs inside the agent's post-ingest result card. When a user drops new material into the `%lore` Inbox and ingestion succeeds, the reply can include an A2UI compile-request card with a **Compile now** button at the bottom. Pressing it opens a confirmation prompt showing the action, button label, exact text (`lore confirm compile`), and destination. The compile only runs after the user confirms and that message is sent through the normal draft/send path.

## Tests
- `pnpm --filter @tloncorp/app exec vitest run ui/components/ChatMessage/a2uiActions.test.ts`
- `pnpm --filter @tloncorp/app tsc`

### DIFF
--- a/docs/post-blobs.md
+++ b/docs/post-blobs.md
@@ -86,9 +86,9 @@ The current renderer expects one `createSurface` message and one `updateComponen
 The supported v1 client subset is intentionally small:
 
 - components: `Card`, `Column`, `Row`, `Text`, `Divider`, and `Button`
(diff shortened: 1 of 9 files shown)


arthyn:
done elsewhere

PR #5862, opened 2026-05-20
fix iOS clipboard image paste crash

dnbrwstr:
## Summary

Fixes issues pasting images as post attachments -- we were passing around a massive base64 string as the image uri, which could cause issues in various places. We now write the bytes to cache and pass around a local uri before upload.

Fixes TLON-5848

## Changes

- Write clipboard image base64 data into the app cache before creating the image asset.
- Preserve clipboard image MIME type, dimensions, and file size when possible.
- Add focused coverage for plain base64 clipboard images and data URI clipboard images.

## How did I test?

Tested on device

### DIFF
--- a/packages/app/ui/components/AttachmentSheet.tsx
+++ b/packages/app/ui/components/AttachmentSheet.tsx
@@ -116,11 +116,8 @@ export default function AttachmentSheet({
           throw new Error('No image data available in clipboard');
         }
 
-        // TODO: we're doing two layers of conversion here:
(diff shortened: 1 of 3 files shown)


patosullivan (approved):
Lgtm, one question though: why use `expo-file-system/legacy`, we use `File`/`Directory` elsewhere

dnbrwstr:
The main reason to use the older api was that it was async, so potentially better for perf, but I agree nicer to use what's current, so I bumped it and we'll see if the synchronous write is noticeable.

PR #5863, opened 2026-05-21
feature: Tlonbot MCP

latter-bolden:
Adds native support for connecting and disconnecting select Tlonbot MCP servers via OAuth. Other Tlonbot settings remain webview based for now.

Like other Hosting API surfaces, we don't track this info in the DB. Provider set and status is supplied dynamically by querying and refreshing as needed.

Mobile only until we sort out CORS issues on Web. Has been tested in simulator and on Preview build.


latter-bolden:
@claude review once

packages/app/features/settings/botMcpSettingsHelpers.ts
```diff
@@ -0,0 +1,128 @@
...
+import { createDevLogger } from '@tloncorp/shared';
+
+import { APP_SCHEME } from '../../constants';
+import type { BotSettingsProviderRow } from '../../ui';
+
+export const COMPLETION_PATH = 'mcp-oauth/complete';
+export const GENERIC_ERROR_MESSAGE = 'Something went wrong. Please try again.';
+export const TOAST_BOTTOM_OFFSET = 24;
+export const MCP_TELEMETRY_EVENTS = {
+  initiatedOAuth: 'Tlonbot MCP: Initiated OAuth',
```
dnbrwstr: Tiny nit: Would it be more consistent to add these to `AnalyticsEvent` enum?

packages/app/features/settings/BotMcpSettingsScreen.tsx
```diff
@@ -0,0 +1,284 @@
...
+  const [startingProviderId, setStartingProviderId] = useState<string | null>(
+    null
+  );
+  const [disconnectingProviderId, setDisconnectingProviderId] = useState<
+    string | null
+  >(null);
+  const handledCompletionUrls = useRef(new Set<string>());
+  const isMounted = useRef(true);
+
+  const refreshStatus = useCallback(async () => {
```
dnbrwstr, line 53: Would it make sense to extract some of the logic into this component into hooks or helpers?

packages/app/features/settings/BotOtherSettingsScreen.tsx
```diff
@@ -0,0 +1,102 @@
...
+      } else {
+        throw new Error(
+          'Cannot access bot settings, failed to get hosting token or user ID.'
+        );
+      }
+    }
+    initialize();
+  }, []);
+
+  useEffect(() => {
```
dnbrwstr, line 69: Maybe add a helper for this so it's not duplicated

packages/app/features/top/UserProfileScreen.tsx
```diff
@@ -87,6 +88,26 @@ export function UserProfileScreen({ route, navigation }: Props) {
...
+    return api.isBotUserIdForUser(userId, currentUserId);
+  }, [currentUserId, userId]);
+
+  const handlePressBotSettings = useCallback(() => {
+    if (isWindowNarrow) {
+      navigation.navigate('BotSettings');
+      return;
+    }
+
+    rootNavigation.dispatch(
```
dnbrwstr: Seems like the common pattern is a navigateToBotSettings helper on useRootNavigation?

Also, I don't think we're using this dispatch/CommonActions pattern elsewhere, why do we need it here?

packages/app/navigation/desktop/SettingsNavigator.tsx
```diff
@@ -34,7 +34,7 @@ function DrawerContent(props: DrawerContentComponentProps) {
   const hasHostedAuth = useHasHostedAuth();
   const hostingBotEnabled = db.hostingBotEnabled.useValue();
   const isHostedUser = getCurrentUserIsHosted();
-  const botEnabled = isHostedUser && hostingBotEnabled;
+  const botEnabled = Platform.OS !== 'web' && isHostedUser && hostingBotEnabled;
```
dnbrwstr, line 40: We're doing this same check in a number of places -- maybe add a helper.

packages/app/ui/components/BotSettingsScreenView.tsx
```diff
@@ -0,0 +1,263 @@
...
+      {initialLoading ? (
+        <YStack flex={1} alignItems="center" justifyContent="center">
+          <LoadingSpinner />
+        </YStack>
+      ) : (
+        <ScrollView
+          style={{
+            flex: 1,
+            width: '100%',
+            maxWidth: 680,
```
dnbrwstr: Assuming the intent is to create a reasonable column size for desktop, max width probably belongs on the content container as this will create an awkward floating scrollbar.

packages/app/ui/components/BotSettingsHomeScreenView.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  return (
+    <View flex={1} backgroundColor="$background">
+      <ScreenHeader
+        borderBottom
+        backAction={isWindowNarrow ? onBackPressed : undefined}
+        title="Bot settings"
+      />
+      <ScrollView
+        style={{
```
dnbrwstr: Would consider creating a wrapper component as we're duplicating this in a couple places. Also, seems like a need that'd be common to all settings screens? How are we handling it elsewhere?

packages/app/ui/components/BotSettingsScreenView.tsx
```diff
@@ -0,0 +1,263 @@
...
+  disabled: boolean;
+  loadingProviderId: string | null;
+  onConnect: (providerId: string) => void;
+  onDisconnect: (providerId: string) => void;
+  providers: BotSettingsProviderRow[];
+  title: string;
+}) {
+  return (
+    <YStack gap="$xs">
+      <Text color="$tertiaryText" size="$label/s">
```
dnbrwstr: This should align with the list items and have the same spacing / padding as section headers on the chat list screen -- right now it looks a little wonky.

dnbrwstr (approved):
Looks good! These are all nits.

packages/app/navigation/desktop/SettingsNavigator.tsx
latter-bolden: I don't think this warrants a helper, the web check is intended to be removed subsequently

packages/app/features/settings/BotOtherSettingsScreen.tsx
latter-bolden: I don't think it's worth adding reusable helper for this since we plan on replacing this screen with native controls soon

packages/app/features/settings/BotMcpSettingsScreen.tsx
latter-bolden: I don't think so. We don't use this pattern elsewhere, I think pulling it into a helper file would add unnecessary indirection

PR #5872, opened 2026-05-21
pioneer: hosting threads

Fang-:
Early draft to track remaining work. I did a pass over all the threads @arthyn generated and cross-referenced them against ylem's `Pioneer/Server/API.hs`.

For now, I've only done so for overall usage. Once we have these sufficiently trimmed down we should do a pass to check them for parity with the existing Click threads and either make them match or leave warnings/instructions for hosting devs about the differences.

----

Here's the state of threads in this PR, broken down by status:

### 👍 Removed, already self-serve integrated by pioneer
- add-to-cordon [here](https://github.com/tloncorp/ylem/blob/ca8982bb9a8bd8a7be3d8a2124bfa874cb6eb7a1/pkg/runtime/pioneer/src/Pioneer/Server/API.hs#L1753-L1756)
- cancel-group-join [here](https://github.com/tloncorp/ylem/blob/ca8982bb9a8bd8a7be3d8a2124bfa874cb6eb7a1/pkg/runtime/pioneer/src/Pioneer/Server/API.hs#L1680)
- is-member [here](https://github.com/tloncorp/ylem/blob/ca8982bb9a8bd8a7be3d8a2124bfa874cb6eb7a1/pkg/runtime/pioneer/src/Pioneer/Server/API.hs#L1877)
- make-admin [here](https://github.com/tloncorp/ylem/blob/ca8982bb9a8bd8a7be3d8a2124bfa874cb6eb7a1/pkg/runtime/pioneer/src/Pioneer/Server/API.hs#L1805)
- force-join [here](https://github.com/tloncorp/ylem/blob/ca8982bb9a8bd8a7be3d8a2124bfa874cb6eb7a1/pkg/runtime/pioneer/src/Pioneer/Server/API.hs#L1636)
- get-contact [here](https://github.com/tloncorp/ylem/blob/ca8982bb9a8bd8a7be3d8a2124bfa874cb6eb7a1/pkg/runtime/pioneer/src/Pioneer/Server/API.hs#L343-L347)
- set-contact-avatar [here](https://github.com/tloncorp/ylem/blob/ca8982bb9a8bd8a7be3d8a2124bfa874cb6eb7a1/pkg/runtime/pioneer/src/Pioneer/Server/API.hs#L1017)
- set-contact-nickname [here](https://github.com/tloncorp/ylem/blob/ca8982bb9a8bd8a7be3d8a2124bfa874cb6eb7a1/pkg/runtime/pioneer/src/Pioneer/Server/API.hs#L502)

### 🔗 Can be made redundant by simple pioneer-side integration 

- [ ] ~~create-group-invite: poke with `%group-action-4`~~
  - ~~challenge: need to generate `@uv`-formatted entropy for the token~~
  - pioneer-side usage wasn't real
- [x] find-channel-string: scry at `/~/scry/channels/v5/[kind]/[ship]/[name]/search/text/0/1/[needle].json` and count resulting array size (thread too general, too)
- [x] find-dm-string: scry at `/~/scry/chat/v4/dm/[ship]/search/text/0/1/[needle].json` and count resulting array size (thread too general, too)
- [ ] ~~get-channels: scry at `/~/scry/channels/v5/channels.json`~~
  - ~~challenge: thread doesn't produce the same as the `getChatChannels` thread did.~~
- [x] post-chat-message: poke host with `%channel-command` (or local ship with `%chat-action`)
- [x] send-dm: poke with `%chat-dm-action-2`
- [x] force-join-with-token: poke with `%group-foreign-2`
  - [x] force-join could be made to do the same!

### ☑️ Tested, but needs pioneer integration

- [x] create-chat-channel
- [ ] create-group-lure-invite
  - thread here slightly different, doesn't extract the token
- [ ] create-lure-invite
  - same difference
- [x] create-private-group
- [x] delete-oldest-dm
- [x] ensure-home-group
- [x] get-dm-counts
- [ ] ~~hosted-groups / is-group-host~~ want pioneer-side flexibility
- [x] is-agent-running
- [x] set-group-title

### 🔁 Needs a rewrite

- [x] is-host: does a check which is trivially validated by its inputs. The pioneer-side equivalent instead wants to know if we host any group with >2 members.
- [x] get-channels: pioneer-side thread produces different output

### 🤔 Up for consideration

- [ ] create-chat-channel: poke already nacks if channel exists, could we observe that nack over SSE?
- [ ] create-private-group poke already nacks if group exists, could we observe that nack over SSE?


mikolajpp:
> create-group-invite: poke with %group-action-4
> challenge: need to generate @uv-formatted entropy for the token

What we mean here is to obtain a valid group token first, right? The corresponding thread in this PR is pretty close, it is just a single poke with empty token (that will trigger generation of a new one if no token is found).

desk/tests/ph/for/pioneer.hoon
```diff
@@ -0,0 +1,168 @@
+::  tests for the pioneer/* threads under /ted/pioneer/
+::
+::  shared aqua boot from /ted/ph/test brings up ~zod, ~bud, ~nec, ~fen, ~dem
+::  with the %groups desk synced and ~fen configured as lure provider.
+::
+::  to run all of these:
+::    ./backend/run-aqua-tests.sh
+::
+::  to run a specific subset, pass an arm-name pattern as the trailing
+::  path component to %ph-test (the test runner uses it as a prefix filter).
```
mikolajpp: I assume these instructions are for agents? It's not really the place to keep them here, we have more than enough information in `docs/aqua` in our repository.

desk/tests/ph/for/pioneer.hoon
```diff
@@ -0,0 +1,168 @@
...
+  ^-  form:m
+  ;<  =bowl:strand  bind:m  get-bowl
+  =/  pax=path
+    %+  weld
+      /gu/(scot %p who)/groups/(scot %da now.bowl)
+    /groups/(scot %p p.flag)/[q.flag]
+  ;<  out=(unit ?)  bind:m  (scry-aqua (unit ?) who pax)
+  (pure:m (fall out |))
+::
+::  ─── tests ───────────────────────────────────────────────────────────
```
mikolajpp: I would avoid such embellishments on principle to keep style consistent in aqua tests and avoid confusing agents. We know where tests start by simply searching for `++  ph`.

Fang-:
> What we mean here is to obtain a valid group token first, right? 

Yeah, the thread as-is is no good, and you can't trivially do it using plain API calls, because hosting cares about learning the invite token at the end.

desk/ted/pioneer/set-group-title.hoon
```diff
@@ -0,0 +1,31 @@
...
+;<  our=@p  bind:m  get-our
+=+  !<(=json arg)
+=/  args=[=flag:g title=@t]
+  ((ot flag+flag:dejs:gj title+so ~) json)
+=*  flag   flag.args
+::  fetch current group to preserve other metadata fields
+::
+;<  =group:v9:gv  bind:m
+  %+  scry  group:v9:gv
+  /gx/groups/v2/groups/(scot %p p.flag)/[q.flag]/group-2
```
arthyn, line 27: should we %gu first?

desk/ted/pioneer/get-channels.hoon
```diff
@@ -0,0 +1,55 @@
...
+=/  hosts=(jug ship flag:v9:gv)
+  %+  roll  ~(tap by groups)
+  |=  [[=flag:v9:gv =group:v9:gv] hosts=(jug ship flag:v9:gv)]
+  %+  roll  ~(tap by channels.group)
+  |=  [[=nest:v9:gv *] =_hosts]
+  ?.  ?=(%chat p.nest)  hosts
+  (~(put ju hosts) p.q.nest flag)
+::
+%-  pure:m
+!>  ^-  json
```
arthyn, line 28: probably should break this one up kinda hard to read

desk/ted/pioneer/ensure-home-group.hoon
```diff
@@ -0,0 +1,78 @@
+::  pioneer/ensure-home-group: idempotently create the home group + default chat channel
+::
+::    arg (json):
+::      { "title":       "Home",
+::        "description": "...",          // optional
+::        "image":       "...",          // optional
+::        "cover":       "..."           // optional
+::      }
+::
+::    creates a group at ~our/home with a "general" chat channel.
```
arthyn: may want to update the comment to reflect the actual values we use

desk/ted/pioneer/ensure-home-group.hoon
```diff
@@ -0,0 +1,78 @@
...
+=/  channel-name=@tas  %home-group-chat
+=/  =flag:g  [our.bowl group-name]
+=/  meta=data:mt
+  :*  title.args
+      (fall description.args '')
+      (fall image.args '')
+      (fall cover.args '')
+  ==
+::
+;<  group-exists=?  bind:m
```
arthyn: does the create group thread do this stuff also? if so we could just call that instead

desk/ted/pioneer/create-private-group.hoon
```diff
@@ -0,0 +1,66 @@
...
+  %-  ot
+  :~  name+(se %tas)
+      title+so
+      description+(mu so)
+      image+(mu so)
+      cover+(mu so)
+  ==
+=/  =flag:g  [our.bowl name.args]
+::
+;<  exists=?  bind:m
```
arthyn: same as the other, should this just use the create group thread? also I wonder what the original intention was, if it was to actually make the group `private` or `secret`? I'm guessing it was just mis-named

desk/ted/pioneer/create-lure-invite.hoon
```diff
@@ -0,0 +1,41 @@
+::  pioneer/create-lure-invite: create a shareable personal invite link
+::
+::    arg (json):
+::      { "id":  "lure-id",       // stable id; reuse to refresh a token
+::        "tag": "personal"       // optional, defaults to "personal"
```
arthyn: we don't vary the tag its always `%groups-0`

desk/ted/pioneer/create-lure-invite.hoon
```diff
@@ -0,0 +1,41 @@
...
+^-  form:m
+;<  =bowl:spider  bind:m  get-bowl:io
+=+  !<(=json arg)
+=/  parse-args
+  =,  dejs:format
+  (ot id+so tag+(mu (se %tas)) ~)
+=/  args=[id=@t tag=(unit @tas)]  (parse-args json)
+=/  id=cord  id.args
+=/  tag=@tas  (fall tag.args %personal)
+=/  =metadata:v1:reel  [tag ~]
```
arthyn: this needs to include both `['inviteType' 'user']` and `['inviterUserId' (scot %t our.bowl)]`

desk/ted/pioneer/create-lure-invite.hoon
```diff
@@ -0,0 +1,41 @@
+::  pioneer/create-lure-invite: create a shareable personal invite link
+::
+::    arg (json):
+::      { "id":  "lure-id",       // stable id; reuse to refresh a token
```
arthyn: since this is the personal invite link this should always be `~zod/personal-invite-link` but I guess it can't hurt to be generic

desk/ted/pioneer/create-group-lure-invite.hoon
```diff
@@ -0,0 +1,57 @@
...
+  %-  ot
+  :~  flag+flag:dejs:gj
+      title+(mu so)
+      image+(mu so)
+  ==
+=/  args=[=flag:g title=(unit @t) image=(unit @t)]  (parse-args json)
+=*  flag  flag.args
+=/  flag-str=cord  (crip "{<p.flag>}/{(trip q.flag)}")
+=/  fields=(map field:reel cord)
+  =/  acc=(map field:reel cord)  ~
```
arthyn:
```suggestion
  =/  acc=(map field:reel cord)
    :~  ['inviteType' 'group']
         ['inviterUserId' (scot %p our.bowl)]
         ['invitedGroupId' flag-str]
    ==
```

desk/ted/pioneer/create-group-lure-invite.hoon
```diff
@@ -0,0 +1,57 @@
...
+  =?  acc  ?=(^ title.args)
+    (~(put by acc) %'invitedGroupTitle' u.title.args)
+  =?  acc  ?=(^ image.args)
+    (~(put by acc) %'invitedGroupIconImageUrl' u.image.args)
+  acc
+=/  =metadata:v1:reel  [%group fields]
```
arthyn:
```suggestion
=/  =metadata:v1:reel  [%groups-0 fields]
```

arthyn (changes requested):

desk/lib/groups-json.hoon
```diff
@@ -1306,6 +1306,27 @@
...
     ==
   ::
+  ++  v9
+    =,  v8
+    |%
+    ++  a-foreigns
+      ^-  $-(json a-foreigns:v9:gv)
+      %-  of
+      :~  foreign+(ot flag+flag a-foreign+a-foreign ~)
+          ::TODO invite+invite
```
mikolajpp: This comment can be removed, we in fact don't want to serialize received invites in this poke.
They are only transferred between agents. (It was a poor choice to not have separate mark for them, but out of scope here)

desk/ted/pioneer/create-chat-channel.hoon
```diff
@@ -0,0 +1,63 @@
...
+  %-  ot
+  :~  flag+flag:dejs:gj
+      name+(se %tas)
+      title+so
+      description+(mu so)
+      image+(mu so)
+  ==
+=/  =nest:c  [%chat our.bowl name.args]
+;<  exists=?  bind:m
+  (scry ? /gu/channels/v1/chat/(scot %p our.bowl)/[name.args])
```
mikolajpp, line 37: Given that scry failures are pretty opaque, we might want to test if `channels` (and other agents) are running and crash with a laconic error message.

desk/ted/pioneer/create-chat-channel.hoon
```diff
@@ -0,0 +1,63 @@
...
+      flag.args
+      title.args
+      (fall description.args '')
+      image.args
+      ~                ::  readers
+      ~                ::  writers
+  ==
+=/  =a-channels:v9:cv  [%create create-channel]
+;<  ~  bind:m
+  (poke [our.bowl %channels] channel-action-1+!>(a-channels))
```
mikolajpp, line 59: Poke failure (here or in any other thread), will crash the thread. I think that will return a HTTP 5xx with the error stacktrace. It would be good to make sure that hosting appropriately captures and logs these.

desk/ted/pioneer/create-group-lure-invite.hoon
mikolajpp: Yes, we should set `inviteType`, but why do we set the other two fields? `%reel` injects them when poked with `%reel-describe`.

desk/ted/pioneer/create-group-lure-invite.hoon
```diff
@@ -0,0 +1,57 @@
...
+  (poke:io [our.bowl %reel] reel-describe+!>([flag-str metadata]))
+;<  =cage  bind:m  (take-fact:io wire)
+?.  ?=(%json p.cage)
+  ~|  unexpected-mark+p.cage  !!
+=/  url-json=^json  !<(^json q.cage)
+?>  ?=([%s *] url-json)
+=/  out=^json
+  %-  pairs:enjs:format
+  :~  url+s+p.url-json
+      flag+s+flag-str
```
mikolajpp: Flags, nests and other identifiers generally come first. (Yes I know JSON does not have order and this was generated :-))
```suggestion
  :~  flag+s+flag-str
      url+s+p.url-json
      
```

desk/ted/pioneer/create-lure-invite.hoon
mikolajpp: Yes for `inviteType`, but `inviterUserId` is always injected by `reel` in `%reel-describe`.

desk/ted/pioneer/create-lure-invite.hoon
mikolajpp: The robot which generated these tests can be pointed at our reel/bait unit tests, which contain examples of proper metadata for different kinds of invites.

desk/ted/pioneer/create-private-group.hoon
mikolajpp: We should use the group creation thread. If hosting needs their own interface, `/ted/group/create-1.hoon` can be even imported as a hoon and wrapped differently here.

mikolajpp: Also, it looks like this creates an empty group with no channels - I don't think we ever want users to land into that?

desk/ted/pioneer/delete-oldest-dm.hoon
```diff
@@ -0,0 +1,41 @@
...
+=/  target=ship  ((ot ship+(se %p) ~) json)
+=/  empty-out=^json  (frond:enjs:format deleted+b+|)
+=/  done-out=^json   (frond:enjs:format deleted+b+&)
+::  scry the oldest writ via the %newer paging mode with start=0
+::
+=/  =path
+  :*  %gx  %chat  %v4  %dm  (scot %p target)
+      %writs  %newer  (scot %ud 0)  (scot %ud 1)  %light
+      %chat-paged-writs-4  ~
+  ==
```
mikolajpp: This is some unorthodox formatting, making it into standard path w/ or w/o welds would be better.

desk/ted/pioneer/delete-oldest-dm.hoon
```diff
@@ -0,0 +1,41 @@
...
+;<  =paged-writs:v7:cv  bind:m  (scry paged-writs:v7:cv path)
+=/  writs-list  (tap:on:writs:v7:cv writs.paged-writs)
+?~  writs-list  (pure:m !>(empty-out))
+=/  oldest=(may:v7:cv writ:v7:cv)  +.i.writs-list
+?.  ?=(%& -.oldest)  (pure:m !>(empty-out))
```
mikolajpp: Confused about the desired logic here: if the oldest was deleted, should not this thread move on to the next-oldest message? It is hard to say without knowing the intention behind this thread, it is interesting that hosting wants to delete DM messages from ships in the first place.

desk/ted/pioneer/ensure-home-group.hoon
mikolajpp: Create group thread does all these and even more. It also returns a whole group state after creation. Despite that, we could simply wrap the group creation thread by importing it here and running it with our desired interface.

desk/ted/pioneer/set-group-title.hoon
```diff
@@ -0,0 +1,31 @@
...
+=/  meta=data:mt
+  [title.args description.meta.group image.meta.group cover.meta.group]
```
mikolajpp:
```suggestion
  :-  title.args
  [description image cover]:meta.group
```

desk/tests/ph/for/pioneer.hoon
```diff
@@ -0,0 +1,139 @@
+::  tests for the pioneer/* threads under /ted/pioneer/
```
mikolajpp: Ideally we should have a test for each of these threads. That said, I am ok with releasing without them for now, as long as someone confirmed they all work. It is going to be a good test of aqua harness to generate the missing tests.

desk/ted/pioneer/get-dm-counts.hoon
```diff
@@ -0,0 +1,57 @@
...
+;<  =bowl:spider  bind:m  get-bowl
+=+  !<(=json arg)
+=/  args=[target=ship limit=(unit @ud)]
+  ((ot ship+(se %p) limit+(mu ni) ~) json)
+=/  limit=@ud  (fall limit.args 1.000)
+=/  =path
+  :*  %gx  %chat  %v4  %dm  (scot %p target.args)
+      %writs  %newest  (scot %ud limit)  %light
+      %chat-paged-writs-4  ~
+  ==
```
mikolajpp: Same comment as above: this is hard to read. Especially the care should be visible at the scry callsite.

desk/ted/pioneer/get-dm-counts.hoon
```diff
@@ -0,0 +1,57 @@
...
+  |=  w=(may:v7:cv writ:v7:cv)
+  ^-  ship
+  =/  a=author:v7:cv  ?:(?=(%| -.w) author.w author.w)
+  ?@(a a ship.a)
+=+  ^=  counts
+  %+  roll  writs-list
+  |=  [[=time w=(may:v7:cv writ:v7:cv)] sent=@ud recvd=@ud]
+  =/  who  (author-of w)
+  ?:  =(who our.bowl)  [+(sent) recvd]
+  [sent +(recvd)]
```
mikolajpp:
```suggestion
  |=  [[=time w=(may:v7:cv writ:v7:cv)] sent=@ud recv=@ud]
  =/  who  (author-of w)
  ?:  =(who our.bowl)  [+(sent) recv]
  [sent +(recv)]
```

desk/ted/pioneer/get-dm-counts.hoon
```diff
@@ -0,0 +1,57 @@
...
+=/  out=^json
+  %-  pairs:enjs:format
+  :~  sent+(numb:enjs:format -.counts)
+      received+(numb:enjs:format +.counts)
+      total+(numb:enjs:format total.paged-writs)
+      sampled+(numb:enjs:format (lent writs-list))
+  ==
```
mikolajpp, line 51:
```suggestion
=*  sent  -.counts
=*  recv  +.counts
=/  out=^json
  %-  pairs:enjs:format
  :~  sent+(numb:enjs:format sent)
      received+(numb:enjs:format recv)
      total+(numb:enjs:format total.paged-writs)
      sampled+(numb:enjs:format (lent writs-list))
  ==
```

mikolajpp: Even better if there is a way to just directly put `sent` and `recv` into the subject. I am pretty sure I have seen code that was able to expose faces from the sample of some accumulator.

desk/ted/pioneer/get-channels.hoon
```diff
@@ -0,0 +1,55 @@
...
+::
+/-  spider, gv=groups-ver
+/+  *strandio
+=,  strand=strand:spider
+^-  thread:spider
+|=  arg=vase
+=/  m  (strand ,vase)
+^-  form:m
+;<  =groups:v9:gv  bind:m
+  (scry groups:v9:gv /gx/groups/v2/groups/groups-2)
```
mikolajpp, line 16:
```suggestion
  (scry groups:v9:gv %gx /groups/v2/groups/groups-2)
```

desk/ted/pioneer/get-channels.hoon
```diff
@@ -0,0 +1,55 @@
...
+::  split by channel host
+::
+=/  hosts=(jug ship flag:v9:gv)
+  %+  roll  ~(tap by groups)
+  |=  [[=flag:v9:gv =group:v9:gv] hosts=(jug ship flag:v9:gv)]
+  %+  roll  ~(tap by channels.group)
+  |=  [[=nest:v9:gv *] =_hosts]
+  ?.  ?=(%chat p.nest)  hosts
+  (~(put ju hosts) p.q.nest flag)
```
mikolajpp, line 25: Iterating with these two loops we will have already seen every group, host and channel. If we record all this information accordingly in something like `hosts=(map ship (jar flag @ta))` then this would correspond directly to the serialized result and we would not need to filter below.

mikolajpp (changes requested):
Overall this looks quite good. 

There is a question of how hosting handles thread crashes. If these are monitored and acted upon, I think we are good. If not, we can think about making these threads slightly more resilient, especially in places where we scry without checking if agent is running. 

There are also various styling nits to be addressed, but nothing major.

The only blocker seems to be lure issues already flagged by @arthyn

desk/ted/pioneer/get-channels.hoon
Fang-: I considered this originally but couldn't make it work in a way that didn't feel complex. Maybe I'll revisit one more time.

desk/ted/pioneer/set-group-title.hoon
Fang-: If the group doesn't exist, should the thread still run to completion? I'd expect failure here, right?

desk/ted/pioneer/create-lure-invite.hoon
Fang-: Well, do we want to make hosting always send `~zod/personal-invite-link`? What happens if some other value gets used?

desk/ted/pioneer/delete-oldest-dm.hoon
Fang-: tloncorp/ylem#2866, looks like intent is overwriting/re-sending the welcome DM. For that purpose, perhaps you're right?

desk/ted/pioneer/create-group-lure-invite.hoon
```diff
@@ -0,0 +1,57 @@
...
+  %-  ot
+  :~  flag+flag:dejs:gj
+      title+(mu so)
+      image+(mu so)
+  ==
+=/  args=[=flag:g title=(unit @t) image=(unit @t)]  (parse-args json)
+=*  flag  flag.args
+=/  flag-str=cord  (crip "{<p.flag>}/{(trip q.flag)}")
+=/  fields=(map field:reel cord)
+  =/  acc=(map field:reel cord)  [[%'inviteType' 'group'] ~ ~]
```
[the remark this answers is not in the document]
mikolajpp: We don't actually use bait landing page.

mikolajpp (approved):
This looks good. Thanks for the styling adjustements! I assume these are going to be, or were already tested off-band, since we have only a scant test coverage here. It is good to go as is though; we can generate remaining tests later under 408k.

yapishu (commented):
i had to make some changes to this to get it to work with pioneer; mainly, spider passes json thread input as `(unit json)`, not raw json, so i changed the thread args. (also note im just pasting the diff into a codeblock here, i haven't pushed to this branch)

also:

- some scries needed mark suffixes
- `create-group-lure-invite` now parses flag first and reads title/image only if present to avoid crashes

```diff
diff --git a/desk/ted/pioneer/create-chat-channel.hoon b/desk/ted/pioneer/create-chat-channel.hoon
index 0699adcc2..72467cd4f 100644
--- a/desk/ted/pioneer/create-chat-channel.hoon
+++ b/desk/ted/pioneer/create-chat-channel.hoon
@@ -20,7 +20,9 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  =bowl:spider  bind:m  get-bowl
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  args=[=flag:g name=@tas title=@t description=(unit @t) image=(unit @t)]
   %.  json
   %-  ot
diff --git a/desk/ted/pioneer/create-group-lure-invite.hoon b/desk/ted/pioneer/create-group-lure-invite.hoon
index 8077ad455..13243fe45 100644
--- a/desk/ted/pioneer/create-group-lure-invite.hoon
+++ b/desk/ted/pioneer/create-group-lure-invite.hoon
@@ -20,23 +20,31 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  =bowl:spider  bind:m  get-bowl:io
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  parse-args
   =,  dejs:format
-  %-  ot
-  :~  flag+flag:dejs:gj
-      title+(mu so)
-      image+(mu so)
-  ==
-=/  args=[=flag:g title=(unit @t) image=(unit @t)]  (parse-args json)
+  (ot ~[flag+flag:dejs:gj])
+=/  args=[=flag:g]  (parse-args json)
 =*  flag  flag.args
+=/  maybe-field
+  |=  key=@t
+  ^-  (unit @t)
+  ?.  ?=([%o *] json)  ~
+  =/  value=(unit ^json)  (~(get by p.json) key)
+  ?~  value  ~
+  ?.  ?=([%s *] u.value)  ~
+  [~ p.u.value]
+=/  title=(unit @t)  (maybe-field 'title')
+=/  image=(unit @t)  (maybe-field 'image')
 =/  flag-str=cord  (crip "{<p.flag>}/{(trip q.flag)}")
 =/  fields=(map field:reel cord)
   =/  acc=(map field:reel cord)  [[%'inviteType' 'group'] ~ ~]
-  =?  acc  ?=(^ title.args)
-    (~(put by acc) %'invitedGroupTitle' u.title.args)
-  =?  acc  ?=(^ image.args)
-    (~(put by acc) %'invitedGroupIconImageUrl' u.image.args)
+  =?  acc  ?=(^ title)
+    (~(put by acc) %'invitedGroupTitle' u.title)
+  =?  acc  ?=(^ image)
+    (~(put by acc) %'invitedGroupIconImageUrl' u.image)
   acc
 =/  =metadata:v1:reel  [%groups-0 fields]
 =/  =wire  /group-lure-invite/(scot %da now.bowl)
diff --git a/desk/ted/pioneer/create-lure-invite.hoon b/desk/ted/pioneer/create-lure-invite.hoon
index b31c6dfd5..3d73ac2d6 100644
--- a/desk/ted/pioneer/create-lure-invite.hoon
+++ b/desk/ted/pioneer/create-lure-invite.hoon
@@ -17,7 +17,9 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  =bowl:spider  bind:m  get-bowl:io
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  id=cord  (so:dejs:format json)
 =/  =wire  /lure-invite/(scot %da now.bowl)
 ;<  ~  bind:m
diff --git a/desk/ted/pioneer/get-dm-counts.hoon b/desk/ted/pioneer/get-dm-counts.hoon
index e6b86447b..c3ba78176 100644
--- a/desk/ted/pioneer/get-dm-counts.hoon
+++ b/desk/ted/pioneer/get-dm-counts.hoon
@@ -23,7 +23,9 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  =bowl:spider  bind:m  get-bowl
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  args=[target=ship limit=(unit @ud)]
   ((ot ship+(se %p) limit+(mu ni) ~) json)
 =/  limit=@ud  (fall limit.args 1.000)
@@ -31,16 +33,13 @@
   /v4/dm/(scot %p target.args)/writs/newest/(scot %ud limit)/light/chat-paged-writs-4
 ;<  =paged-writs:v7:cv  bind:m  (scry paged-writs:v7:cv %gx %chat path)
 =/  writs-list  (tap:on:writs:v7:cv writs.paged-writs)
-::  extract author ship from each writ; treats bot-author entries as their ship
-::
-=/  author-of
-  |=  w=(may:v7:cv writ:v7:cv)
-  ^-  ship
-  =/  a=author:v7:cv  ?:(?=(%| -.w) author.w author.w)
-  ?@(a a ship.a)
 =+  %+  roll  writs-list
     |=  [[=time w=(may:v7:cv writ:v7:cv)] sent=@ud recv=@ud]
-    =/  who  (author-of w)
+    ?:  ?=(%| -.w)
+      [sent recv]
+    =/  who=ship
+      =/  a=author:v7:cv  author.w
+      ?@(a a ship.a)
     ?:  =(who our.bowl)  [+(sent) recv]
     [sent +(recv)]
 =/  out=^json
diff --git a/desk/ted/pioneer/is-agent-running.hoon b/desk/ted/pioneer/is-agent-running.hoon
index fe29ac1fa..3f02181d6 100644
--- a/desk/ted/pioneer/is-agent-running.hoon
+++ b/desk/ted/pioneer/is-agent-running.hoon
@@ -14,7 +14,9 @@
 |=  arg=vase
 =/  m  (strand ,vase)
 ^-  form:m
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  app=@tas  ((ot agent+(se %tas) ~) json)
 ;<  running=?  bind:m  (scry ? /gu/[app]/$)
-(pure:m !>(`json`b+running))
+(pure:m !>(b+running))
diff --git a/desk/ted/pioneer/set-group-title.hoon b/desk/ted/pioneer/set-group-title.hoon
index 5908ff721..9cfa4b7f3 100644
--- a/desk/ted/pioneer/set-group-title.hoon
+++ b/desk/ted/pioneer/set-group-title.hoon
@@ -14,7 +14,9 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  our=@p  bind:m  get-our
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  args=[=flag:g title=@t]
   ((ot flag+flag:dejs:gj title+so ~) json)
 =*  flag   flag.args
diff --git a/desk/tests/ph/for/pioneer.hoon b/desk/tests/ph/for/pioneer.hoon
index 5dbada067..58a6d3ac0 100644
--- a/desk/tests/ph/for/pioneer.hoon
+++ b/desk/tests/ph/for/pioneer.hoon
@@ -22,7 +22,7 @@
     (cat 3 name (cat 3 '-' (scot %uv (sham now.bowl name))))
   =/  =beak  [who %groups da+now.bowl]
   =/  file=term  (cat 3 'pioneer-' name)
-  =/  start=start-args:spider  [~ `tid beak file !>(arg)]
+  =/  start=start-args:spider  [~ `tid beak file !>(`(unit json)`[~ arg])]
   =/  watch-path=path  /(scot %p who)/spider/thread-result/[tid]
   ;<  ~  bind:m  (watch-app watch-path [who %spider] /thread-result/[tid])
   ;<  ~  bind:m  (poke-app [who %spider] spider-start+start)
➜  tlon-apps git:(hm/hosting-threads) ✗ git diff
➜  tlon-apps git:(hm/hosting-threads) ✗ git diff > pr.diff
➜  tlon-apps git:(hm/hosting-threads) ✗ cat pr.diff
diff --git a/desk/ted/pioneer/create-chat-channel.hoon b/desk/ted/pioneer/create-chat-channel.hoon
index 0699adcc2..72467cd4f 100644
--- a/desk/ted/pioneer/create-chat-channel.hoon
+++ b/desk/ted/pioneer/create-chat-channel.hoon
@@ -20,7 +20,9 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  =bowl:spider  bind:m  get-bowl
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  args=[=flag:g name=@tas title=@t description=(unit @t) image=(unit @t)]
   %.  json
   %-  ot
diff --git a/desk/ted/pioneer/create-group-lure-invite.hoon b/desk/ted/pioneer/create-group-lure-invite.hoon
index 8077ad455..13243fe45 100644
--- a/desk/ted/pioneer/create-group-lure-invite.hoon
+++ b/desk/ted/pioneer/create-group-lure-invite.hoon
@@ -20,23 +20,31 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  =bowl:spider  bind:m  get-bowl:io
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  parse-args
   =,  dejs:format
-  %-  ot
-  :~  flag+flag:dejs:gj
-      title+(mu so)
-      image+(mu so)
-  ==
-=/  args=[=flag:g title=(unit @t) image=(unit @t)]  (parse-args json)
+  (ot ~[flag+flag:dejs:gj])
+=/  args=[=flag:g]  (parse-args json)
 =*  flag  flag.args
+=/  maybe-field
+  |=  key=@t
+  ^-  (unit @t)
+  ?.  ?=([%o *] json)  ~
+  =/  value=(unit ^json)  (~(get by p.json) key)
+  ?~  value  ~
+  ?.  ?=([%s *] u.value)  ~
+  [~ p.u.value]
+=/  title=(unit @t)  (maybe-field 'title')
+=/  image=(unit @t)  (maybe-field 'image')
 =/  flag-str=cord  (crip "{<p.flag>}/{(trip q.flag)}")
 =/  fields=(map field:reel cord)
   =/  acc=(map field:reel cord)  [[%'inviteType' 'group'] ~ ~]
-  =?  acc  ?=(^ title.args)
-    (~(put by acc) %'invitedGroupTitle' u.title.args)
-  =?  acc  ?=(^ image.args)
-    (~(put by acc) %'invitedGroupIconImageUrl' u.image.args)
+  =?  acc  ?=(^ title)
+    (~(put by acc) %'invitedGroupTitle' u.title)
+  =?  acc  ?=(^ image)
+    (~(put by acc) %'invitedGroupIconImageUrl' u.image)
   acc
 =/  =metadata:v1:reel  [%groups-0 fields]
 =/  =wire  /group-lure-invite/(scot %da now.bowl)
diff --git a/desk/ted/pioneer/create-lure-invite.hoon b/desk/ted/pioneer/create-lure-invite.hoon
index b31c6dfd5..3d73ac2d6 100644
--- a/desk/ted/pioneer/create-lure-invite.hoon
+++ b/desk/ted/pioneer/create-lure-invite.hoon
@@ -17,7 +17,9 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  =bowl:spider  bind:m  get-bowl:io
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  id=cord  (so:dejs:format json)
 =/  =wire  /lure-invite/(scot %da now.bowl)
 ;<  ~  bind:m
diff --git a/desk/ted/pioneer/get-dm-counts.hoon b/desk/ted/pioneer/get-dm-counts.hoon
index e6b86447b..b5b6f4ecb 100644
--- a/desk/ted/pioneer/get-dm-counts.hoon
+++ b/desk/ted/pioneer/get-dm-counts.hoon
@@ -23,7 +23,9 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  =bowl:spider  bind:m  get-bowl
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  args=[target=ship limit=(unit @ud)]
   ((ot ship+(se %p) limit+(mu ni) ~) json)
 =/  limit=@ud  (fall limit.args 1.000)
diff --git a/desk/ted/pioneer/is-agent-running.hoon b/desk/ted/pioneer/is-agent-running.hoon
index fe29ac1fa..3f02181d6 100644
--- a/desk/ted/pioneer/is-agent-running.hoon
+++ b/desk/ted/pioneer/is-agent-running.hoon
@@ -14,7 +14,9 @@
 |=  arg=vase
 =/  m  (strand ,vase)
 ^-  form:m
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  app=@tas  ((ot agent+(se %tas) ~) json)
 ;<  running=?  bind:m  (scry ? /gu/[app]/$)
-(pure:m !>(`json`b+running))
+(pure:m !>(b+running))
diff --git a/desk/ted/pioneer/set-group-title.hoon b/desk/ted/pioneer/set-group-title.hoon
index 5908ff721..9cfa4b7f3 100644
--- a/desk/ted/pioneer/set-group-title.hoon
+++ b/desk/ted/pioneer/set-group-title.hoon
@@ -14,7 +14,9 @@
 =/  m  (strand ,vase)
 ^-  form:m
 ;<  our=@p  bind:m  get-our
-=+  !<(=json arg)
+=+  !<(arg=(unit json) arg)
+?>  ?=(^ arg)
+=*  json  u.arg
 =/  args=[=flag:g title=@t]
   ((ot flag+flag:dejs:gj title+so ~) json)
 =*  flag   flag.args
diff --git a/desk/tests/ph/for/pioneer.hoon b/desk/tests/ph/for/pioneer.hoon
index 5dbada067..58a6d3ac0 100644
--- a/desk/tests/ph/for/pioneer.hoon
+++ b/desk/tests/ph/for/pioneer.hoon
@@ -22,7 +22,7 @@
     (cat 3 name (cat 3 '-' (scot %uv (sham now.bowl name))))
   =/  =beak  [who %groups da+now.bowl]
   =/  file=term  (cat 3 'pioneer-' name)
-  =/  start=start-args:spider  [~ `tid beak file !>(arg)]
+  =/  start=start-args:spider  [~ `tid beak file !>(`(unit json)`[~ arg])]
   =/  watch-path=path  /(scot %p who)/spider/thread-result/[tid]
   ;<  ~  bind:m  (watch-app watch-path [who %spider] /thread-result/[tid])
   ;<  ~  bind:m  (poke-app [who %spider] spider-start+start)
```

Fang-:
Sweet, thank you. I made some small adjustments and applied your changes, much appreciated.

> some scries needed mark suffixes

Huh. I ran all these threads and should've hit their scries. I also don't see anything of this sort in your patch. A quick search doesn't reveal any obvious misses here. Are we in the clear?

yapishu:
> Huh. I ran all these threads and should've hit their scries. I also don't see anything of this sort in your patch. A quick search doesn't reveal any obvious misses here. Are we in the clear?

my mistake, i misremembered something i had been testing with; the threads in the diff above work with the pioneer  calls

backend/run-tests.sh
```diff
@@ -113,50 +113,41 @@ run_click="$click -t $TIMEOUT -b $vere -i - -kp"
...
-
 echo "Updating base desk..."
 $run_click $pier <<EOF
-=/  m  (strand ,vase)  
-;<  our=ship  bind:m  get-our  
-;<  ~  bind:m  (poke [our %hood] kiln-commit+!>([%base |]))  
-(pure:m !>(%ok))  
+=/  m  (strand ,vase)
+;<  our=ship  bind:m  get-our
+;<  ~  bind:m  (poke [our %hood] kiln-commit+!>([%base |]))
```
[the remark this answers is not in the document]
mikolajpp: We don't download any ships here, our pill has these patched.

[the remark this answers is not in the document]
mikolajpp: Someone please make an LLM that constantly questions itself using scholastic method to discover what it does not know.

PR #5873, opened 2026-05-22
deps: patch RNGH to handle pointerEvents

jamesacklin:
## Summary

Fixes TLON-5659 where our InteractableChatListItem would fail to handle taps on left-hand action Pressables which were revealed via swipe.

ReanimatedSwipeable renders both the left-actions container and the right-actions container as absolute-fill overlays on top of the row. They animate from `opacity: 0` to `opacity: 1` based on swipe progress. 

On Android, an element at `opacity: 0` is still a live touch target, so the right-actions container was sitting in front of (or in the same z-stack as) the visible left actions and eating the taps. iOS handles this differently, which is why it only repro'd on Android.

Outstanding upstream issue: https://github.com/software-mansion/react-native-gesture-handler/issues/3223

## Changes

Patches react-native-gesture-handler to link `pointerEvents` to the same progress value as `opacity`. This stops the invisible (right-side) container from intercepting taps meant for the visible (left-side) container.

## How did I test?

Confirmed working in Android simulator (see video), no regression in iOS simulator.

## Risks and impact

  - [x] Other: gesture deps

## Rollback plan

git revert

## Screenshots / videos

[untitled.webm](https://github.com/user-attachments/assets/95c8d95c-c378-44e5-8f6d-0f0cd6ece543)


jamesacklin:
@claude review once

janicduplessis (approved):
Looks good! Surprised this hasn't been fixed upstream considering the big issue. Could be good to open upstream PR too to avoid having to keep this patch for too long.

PR #5877, opened 2026-05-26
groups: preserve local state when sending and ingesting updates

mikolajpp:
## Summary

The invited record in group admissions is local to a ship, but it was being overwritten when processing the group creation update. This in particular caused problems when a ship acquired admin rights, as it would receive new group state containing all available information, including that restricted to admins, at the same time clearing the invited record.

`active-channels` was similarly affected. however, the state is recreated on each update, which might explain why we have rarely, if ever, observed any problems.

Fixes TLON-5897

## Changes

Preserve `invited` and `active-channels` record when processing group creation updates.

## How did I test?

I had a regular member invite a third-party to a public group, and than had him acquire admin rights.
Observed the invitation record being preserved after this change, while it was lost previously.

Amending `active-channels` management follows the same code paths as for the invited record.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.


desk/app/groups.hoon
```diff
@@ -2710,12 +2710,20 @@
...
+      ::  the group, including information restricted to admins.
+      ::
       =+  paths=(se-ships-subscription-paths new-admins)
       ?:  =(~ paths)  se-core
       =/  time
         ?~  update=(ram:log-on:g log)  now.bowl
         -.u.update
-      =/  =update:g  [time %create group]
+      =/  gr=group:g  group
+      ::  clear local state
```
Fang-:
```suggestion
      ::  clear local state for the update
```

Fang- (approved):
Looks good! Small nit about the comments for clarity.

PR #5880, opened 2026-05-26
fix unread query invalidation

dnbrwstr:
## Summary

Fixes several unread state mismatches that could leave group, channel, activity, or badge state stale or inconsistent. Group and channel unread queries now invalidate from the tables they actually read, notification-only activity such as reactions is treated as unread activity, and badge/presented-notification gates now follow the backend `notify` signal instead of only message counts.

Fixes TLON-5871

## Changes

- Added missing unread query invalidation for group detail state and fixed channel unread count invalidation.
- Preserved base unread notification timestamps when deriving local unread rows for badge updates.
- Switched badge and presented notification gates to count local notifying unread sources, including notification-only rows.
- Mark channel/thread notification-only activity as read when the user opens the relevant source.
- Added shared unread activity helpers and a child thread unread activity count so parent DMs do not clear while a child thread is still notifying.
- Fixed nested thread unread upserts from channel unread inserts.
- Added DB/API coverage for base unread conversion, notification-only clearing, query invalidation metadata, and nested/thread unread behavior.

## How did I test?

Tested on device

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [x] Notifications

## Rollback plan

Revert.

### DIFF
--- a/packages/shared/src/db/queries.ts
+++ b/packages/shared/src/db/queries.ts
@@ -1931,23 +1931,6 @@ function threadUnreadActivityPredicate() {
   return or(ne($threadUnreads.count, 0), eq($threadUnreads.notify, true));
 }
 
-export const getThreadUnreadActivityCountByChannel = createReadQuery(
-  'getThreadUnreadActivityCountByChannel',
-  async ({ channelId }: { channelId: string }, ctx: QueryCtx) => {
-    const result = await ctx.db
-      .select({ count: count() })
-      .from($threadUnreads)
-      .where(
-        and(
-          eq($threadUnreads.channelId, channelId),
-          threadUnreadActivityPredicate()
-        )
-      );
-    return result[0]?.count ?? 0;
-  },
-  ['threadUnreads']
-);
-
 export interface GetUnreadsOptions {
   orderBy?: 'updatedAt';
   includeFullyRead?: boolean;
@@ -4977,15 +4960,14 @@ export const insertChannelUnreads = createWriteQuery(
       }
     });
   },
-  (unreads) =>
-    unreads.length
-      ? [
-          'channelUnreads',
-          ...(unreads.some((u) => (u.threadUnreads?.length ?? 0) > 0)
-            ? (['threadUnreads'] as const)
-            : []),
-        ]
-      : []
+  (unreads) => {
+    if (unreads.length === 0) {
+      return [];
+    }
+    return unreads.some((u) => (u.threadUnreads?.length ?? 0) > 0)
+      ? ['channelUnreads', 'threadUnreads']
+      : ['channelUnreads'];
+  }
 );
 
 export const clearChannelUnread = createWriteQuery(
--- a/packages/shared/src/db/queries.ts
+++ b/packages/shared/src/db/queries.ts
@@ -4949,59 +4949,6 @@ export const clearGroupUnread = createWriteQuery(
   ['groupUnreads']
 );
 
-export const clearStaleNotificationOnlyGroupUnread = createWriteQuery(
-  'clearStaleNotificationOnlyGroupUnread',
-  async ({ groupId }: { groupId: string }, ctx: QueryCtx) => {
-    const groupUnread = await ctx.db.query.groupUnreads.findFirst({
-      where: eq($groupUnreads.groupId, groupId),
-    });
-    if (!groupUnread?.notify || (groupUnread.count ?? 0) > 0) {
-      return false;
-    }
-
-    const [notifyingChannels] = await ctx.db
-      .select({ count: count() })
-      .from($channelUnreads)
-      .innerJoin($channels, eq($channelUnreads.channelId, $channels.id))
-      .where(
-        and(eq($channels.groupId, groupId), eq($channelUnreads.notify, true))
-      );
-    if ((notifyingChannels?.count ?? 0) > 0) {
-      return false;
-    }
-
-    const [notifyingThreads] = await ctx.db
-      .select({ count: count() })
-      .from($threadUnreads)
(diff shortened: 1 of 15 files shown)


dnbrwstr:
@claude review once

packages/shared/src/store/dbHooks.ts
```diff
@@ -290,6 +290,27 @@ export const useLiveThreadUnreadByParentId = (parentPostId: string | null) => {
   });
 };
 
+export const useLiveThreadUnreadsByChannel = (channelId: string | null) => {
+  const depsKey = useMemo(
+    () => (channelId ? keyFromQueryDeps(db.getThreadUnreadsByChannel) : null),
```
patosullivan: Why not use `useKeyFromQueryDeps` here?

packages/shared/src/store/channelActions.ts
```diff
@@ -631,6 +634,22 @@ export async function markChannelRead({
...
   }
+  const channelUnreadWasNotificationOnly =
+    existingUnread?.notify === true &&
+    (existingUnread.count ?? 0) === 0 &&
+    (existingUnread.countWithoutThreads ?? 0) === 0;
+  const groupUnreadLooksLikeSameNotification =
+    existingGroupUnread?.notify === true &&
+    (existingGroupUnread.count ?? 0) === 0 &&
+    (existingGroupUnread.notifyCount ?? 0) <= 1 &&
+    existingGroupUnread.updatedAt === existingUnread?.updatedAt;
```
patosullivan, line 641: I think we might be missing a case here. What if there are two channels with one unread each in the group (total notifyCount is 2)? User views one channel, it's marked read, but `notifyCount` is still <= 1, so we skip the cleanup and the group count stays at 2.

dnbrwstr: Addressed this

packages/shared/src/db/queries.ts
```diff
@@ -4936,27 +4969,39 @@ export const insertChannelUnreads = createWriteQuery(
...
 export const clearChannelUnread = createWriteQuery(
   'clearChannelUnread',
   async (channelId: string, ctx: QueryCtx) => {
     return ctx.db
       .update($channelUnreads)
-      .set({ count: 0, countWithoutThreads: 0, firstUnreadPostId: null })
+      .set({
+        count: 0,
+        countWithoutThreads: 0,
+        notify: false,
```
arthyn: same here

packages/shared/src/db/queries.ts
```diff
@@ -4967,15 +5012,19 @@ export const updateChannelUnreadCount = createWriteQuery(
...
       const existingCount = existingUnread.count ?? 0;
-      if (existingCount && existingCount - decrement >= 0) {
+      const nextCount = existingCount - decrement;
+      if (existingCount && nextCount >= 0) {
         return ctx.db
           .update($channelUnreads)
-          .set({ count: existingCount - decrement })
+          .set({
+            count: nextCount,
+            ...(nextCount === 0 ? { notify: false } : {}),
```
arthyn: same here

packages/shared/src/db/queries.ts
```diff
@@ -4891,10 +4920,14 @@ export const updateGroupUnreadCount = createWriteQuery(
...
       const existingCount = existingUnread.count ?? 0;
-      if (existingCount && existingCount - decrement >= 0) {
+      const nextCount = existingCount - decrement;
+      if (existingCount && nextCount >= 0) {
         return ctx.db
           .update($groupUnreads)
-          .set({ count: existingCount - decrement })
+          .set({
+            count: nextCount,
+            ...(nextCount === 0 ? { notify: false, notifyCount: 0 } : {}),
```
arthyn: would this temporarily mask if we have thread notifications when viewing chat/channel list?

packages/shared/src/db/queries.ts
```diff
@@ -1901,24 +1901,36 @@ export const removeChatMembers = createWriteQuery(
   ['chatMembers', 'groups']
 );
 
-export const getUnreadsCountWithoutMuted = createReadQuery(
-  'getUnreadsCountWithoutMuted',
-  async ({ type }: { type?: ChannelUnread['type'] }, ctx: QueryCtx) => {
-    const result = await ctx.db
+export const getNotifyingUnreadSourceCount = createReadQuery(
+  'getNotifyingUnreadSourceCount',
```
arthyn, line 1905: this query reads weirdly to me, we're just grabbing the first result from each of these queries and using the count? wouldn't you want to sum them at least?

arthyn (approved):
I'm approving this as a stop gap for those with backends synced to develop, but I have some ideas on what we can ship from the backend to prevent you having to do most of this and make the rest easier.

packages/shared/src/db/queries.ts
dnbrwstr: Fixed in 79a07c8

packages/shared/src/db/queries.ts
dnbrwstr: Fixed in 79a07c8

packages/shared/src/db/queries.ts
dnbrwstr: Fixed in 79a07c8

packages/shared/src/db/queries.ts
dnbrwstr: We don't need the sum of actual notifications here, just the number of sources that are still unread, so we know whether we should do our conservative badge clear. Cleaned up the naming a little to make it more clear.

PR #5883, opened 2026-05-27
Fix orphan role picker sheet on assign/remove

patosullivan:
## Summary

Fixes TLON-5891. Tapping a role in the role-assignment sheet left a stale, orphaned picker visible, made the parent profile sheet flicker, and let a second tap toggle the role back. Root cause was a nested-modal interaction with `@gorhom/bottom-sheet`'s default `stackBehavior='switch'`. The fix dismisses both sheets through Gorhom's normal animation flow instead of yanking the React subtree out from under the queue.

## Changes

Root cause: when the inner role picker presented, Gorhom called `.minimize()` on the parent (default `stackBehavior='switch'`). Our `BottomSheetWrapper` interpreted the resulting `onChange(-1)` as a user dismissal, propagated `onOpenChange(false)` up to the caller, and unmounted the entire `GroupMemberProfileSheet` subtree. The inner Gorhom modal was never told to dismiss, so it remained in `BottomSheetModalProvider._sheetsQueue` as a visible orphan rendering its last-rendered (stale, no-checkmark) state.

- `BottomSheetWrapper.types.ts` / `BottomSheetWrapper.native.tsx` / `BottomSheetWrapper.web.tsx` — added optional `stackBehavior?: 'replace' | 'push' | 'switch'` prop and forwarded it to Gorhom on native; web wrapper accepts and ignores it defensively.
- `ActionSheet.tsx` — exposes `stackBehavior`, destructured separately so it is routed only to `BottomSheetWrapper` on native and is not leaked into the Tamagui `<Sheet>` on web.
- `ProfileSheet.tsx` — sets `stackBehavior="push"` on the nested role-picker `ActionSheet` so the parent is no longer minimized. After a successful role tap, the inner picker closes via `setOpen(false)` and `ProfileSheet` schedules a 300ms `requestParentClose()` so the parent animates closed through Gorhom's normal dismiss flow. The deferred-close timer is owned by `ProfileSheet` (not by `RoleAssignmentSheet`) so that a self-demoting non-host admin — whose role mutation unmounts the admin action group and the picker as a side effect — does not cancel its own parent-close timer. Cleanup is tied to `contactId` change / `ProfileSheet` unmount. Host/admin guard preserved (the host's own `admin` row closes the picker only, no mutation, parent stays open) and `closingRef` suppresses double-taps.
- `GroupMemberProfileSheet.tsx` — owns local `parentOpen` state driving `ProfileSheet`'s `open` prop, so a caller-initiated close runs through Gorhom's animate-to-`-1` and `enableDismissOnClose` cleanup before the caller unmounts the subtree. Hidden from callers  (`GroupMembersScreenView.tsx` and `chatDetails.tsx` are unchanged).

## How did I test?

Manual QA on iOS and Android

## Risks and impact

  - [x] Other: group member role management UI (admin-facing)

Notes:

- The 300ms `PARENT_CLOSE_DELAY_MS` / `PARENT_CLOSE_ANIMATION_MS` are heuristics matched to `BottomSheetWrapper`'s `quick` transition duration. If that duration changes these need to move with it. Cleaner long-term fix would be to plumb Gorhom's `onDismiss` through `BottomSheetWrapper` and anchor parent-close to actual animation completion, probably out of scope here.
- `stackBehavior` is now plumbed through the wrapper, but `RoleAssignmentSheet` is the only consumer at the moment. We actually don't have nested-modal sheets anywhere else in the  app. If anyone *adds* a new nested-modal sheet in the future, they should know to set `stackBehavior="push"`.

## Rollback plan

Revert.

## Screenshots / videos

https://github.com/user-attachments/assets/e27d5e22-d849-4b41-9f0b-fbda95f86b26

### DIFF
--- a/packages/app/ui/components/ActionSheet.tsx
+++ b/packages/app/ui/components/ActionSheet.tsx
@@ -149,6 +149,7 @@ const ActionSheetComponent = ({
   closeButton,
   footerComponent,
   unmountOnClose,
+  stackBehavior,
   ...props
 }: PropsWithChildren<
   ActionSheetProps &
@@ -159,6 +160,7 @@ const ActionSheetComponent = ({
       | 'hasScrollableContent'
       | 'keyboardBehavior'
       | 'unmountOnClose'
+      | 'stackBehavior'
     >
 >) => {
   const mode = useAdaptiveMode(forcedMode);
@@ -357,6 +359,7 @@ const ActionSheetComponent = ({
       footerComponent={footerComponent}
       hasScrollableContent={hasScrollableContent}
       unmountOnClose={unmountOnClose}
+      stackBehavior={stackBehavior}
       frameStyle={{}}
     >
       <ActionSheetContext.Provider value={actionSheetContextValue}>
--- a/packages/app/ui/components/BottomSheetWrapper.native.tsx
+++ b/packages/app/ui/components/BottomSheetWrapper.native.tsx
@@ -146,6 +146,7 @@ export const BottomSheetWrapper = forwardRef<
       enableContentPanningGesture,
       enableDynamicSizing,
       unmountOnClose = false,
+      stackBehavior,
     },
     ref
   ) => {
@@ -465,6 +466,7 @@ export const BottomSheetWrapper = forwardRef<
           key={mountKey}
           ref={bottomSheetModalRef}
           accessibilityViewIsModal={true}
+          stackBehavior={stackBehavior}
           {...commonProps}
           {...commonOverrides}
         >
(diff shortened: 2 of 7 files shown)


patosullivan:
@claude review

patosullivan:
@claude review

dnbrwstr (changes requested):
There's a lot of bespoke logic in these two components to manage this specific sheet situation which could pop up elsewhere. Could we either use the existing pattern for navigating within a sheet here or abstract this logic out so it could be reused elsewhere?

patosullivan:
@claude review once

dnbrwstr (approved):
Lgtm!

PR #5885, opened 2026-05-29
update signup flow to allow for plain signups + invites

dosullivan:
## Summary

Opens up registration through the mobile app by adding an un-gated signup path alongside the existing invite flow. When a user has no invite, the app now sends no lure (instead of falling back to DEFAULT_LURE), so the backend routes them to the default no-group policy; users with an invite still paste/deeplink a code and join the inviter's group as before. Also fixes a URLSearchParams bug where an absent lure serialized to the literal string "undefined", and updates the welcome/signup screens to surface both paths. Update the copy as you see fit, but this should work fine against the /v1/sign-up endpoint for regular sign-ups and invites. 
## Changes

- `packages/app/contexts/branch.tsx` — useSignupParams no longer falls back to DEFAULT_LURE/DEFAULT_PRIORITY_TOKEN; lureId/priorityToken stay undefined when there's no invite, so un-invited signups send no lure. Removed the now-unused imports.
- `packages/api/src/client/hostingApi.ts` — getHostingAvailability builds its query string conditionally (only sets email/lure/priorityToken when present), fixing the bug where new URLSearchParams({lure: undefined}) serialized to the literal string "undefined".
- `packages/shared/src/store/hostingActions.ts` — signUpHostedUser's inviteId and priorityToken params are now optional (omitted from the request body for open signups).
- `apps/tlon-mobile/src/screens/Onboarding/WelcomeScreen.tsx` — with no invite, "Sign up" now navigates straight to Signup (open path); added a secondary "I have an invite code" button → PasteInviteLink.
- `apps/tlon-mobile/src/screens/Onboarding/SignupScreen.tsx` — screen title is "Accept invite" only when there's a lure, otherwise "Sign up".
- `apps/tlon-mobile/src/screens/Onboarding/PasteInviteLinkScreen.tsx` — the fallback button changed from "No invite? Join waitlist" (→ JoinWaitList) to "Sign up without an invite" (→ Signup).
## How did I test?

have not tested (don't have the tooling for that locally)

## Risks and impact

  - [X] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

### DIFF
--- a/apps/tlon-mobile/src/hooks/useReviveSavedOnboarding.ts
+++ b/apps/tlon-mobile/src/hooks/useReviveSavedOnboarding.ts
@@ -1,6 +1,5 @@
 import { NavigationProp, useNavigation } from '@react-navigation/native';
 import * as api from '@tloncorp/api';
-import { useLureMetadata } from '@tloncorp/app/contexts/branch';
 import { useShip } from '@tloncorp/app/contexts/ship';
 import { AnalyticsEvent, createDevLogger } from '@tloncorp/shared';
 import { signupData } from '@tloncorp/shared/db';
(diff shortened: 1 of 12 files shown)


latter-bolden (approved):
This should be ready to go, but we need to hold on merging until all hosting support is deployed and confirmed functional

PR #5886, opened 2026-05-29
Upgrade mobile to Expo SDK 56

janicduplessis:
## Summary

Upgrades the mobile app from Expo SDK 54 to SDK 56 (React Native 0.81 → 0.85, React 19.1 → 19.2). Both iOS and Android build and run on the new architecture with prebuilt React Native — no `buildReactNativeFromSource`.

## Changes

- **Dependencies:** bump `expo` to `~56.0.x` and the SDK-bundled `expo-*` packages, `react-native` 0.85.3, `react` 19.2, `@react-native-firebase/*` 24, Skia, reanimated/worklets, etc. Root `pnpm.overrides` pin the SDK 56 versions to keep the workspace's `*` peer deps from dragging in off-SDK transitives.
    - `@shopify/flash-list` is kept at **2.3.1** (the version `develop` uses) rather than the `2.0.2` the SDK 56 upgrade tooling reset it to.
    - `react-native-safe-area-context` (5.7.0) and `react-native-screens` (4.25.2) are pinned so web/mobile/app/ui all resolve a **single** copy of each (they had drifted to two versions).
    - `react-test-renderer` is pinned via overrides to `19.2.3` so it matches `react` exactly (the renderer enforces an exact-version check at init).
- **OTA runtime version:** `app.config.ts` switches from a hardcoded `runtimeVersion: '4.0.2'` (shared with the SDK 54 builds) to `{ policy: 'fingerprint' }`. The runtime version now tracks the native layer automatically, so a native change like this SDK bump can't serve incompatible JS to older native clients over EAS Update.
- **Android build system:** SDK 56 removed `useExpoModules()` / `scripts/autolinking.gradle`, so `settings.gradle` moves to the `expo-autolinking-settings` plugin + `expoAutolinking.useExpoModules()/useExpoVersionCatalog()`. `app/build.gradle` renames the `reactAndroidLibs` catalog to `expoLibs` and enables core library desugaring (required by recaptcha-enterprise). `MainApplication.kt` migrates to `ExpoReactHostFactory.getDefaultReactHost`.
- **iOS build system:** `AppDelegate.swift` migrates to the SDK 56 API (drops the `React_RCTAppDelegate` import — those symbols come through `React` now — uses `internal import Expo`, removes the deleted `bindReactNativeFactory`). RNFB is kept out of the prebuilt-React framework-module conflict via `ios.forceStaticLinking`, and `Error+logWithDomain.swift` uses the direct `FirebaseCrashlytics` API since static-linked RNFB exposes no Swift module.
- **Native lockfiles:** the Android `gradle.lockfile` and iOS `Podfile.lock` are regenerated for the SDK 56 dependency set (Android uses `dependencyLocking { lockAllConfigurations() }`, so a stale lock fails resolution outright).
- **expo-av → expo-audio:** voice-memo recording/playback migrated to the new hooks API; `expo-av` removed.
- **Production web boot fix:** Expo's winter runtime makes `expo/fetch` the global `fetch`, and `SafeAreaProvider` has no synchronous insets on the web's first paint. `ToastView` (rendered eagerly, before its `visible` guard) calls `useSafeAreaInsets()`, which now throws on a null value — crashing the production build straight to the error boundary. `BaseProviderStack` passes `initialWindowMetrics` to `SafeAreaProvider` so insets exist on first render.

### Patches

Every patch was reconciled against the SDK 56 dependency tree.

- **Recreated / re-ported for the bumped versions:** `@gorhom/bottom-sheet` → 5.2.14 (the `unstable_getBoundingClientRect` crash fix is upstream now, so the patch is slimmed to just the flex-overflow fix); `react-native-gesture-handler` → 2.31.2; `react-native-reanimated` → 4.3.1; `expo-image-manipulator` → 56.0.19; `expo-background-task` → 56.0.19.
- **New:** tentap patch for RN 0.85's removal of `UIManagerModule.onBatchComplete()`.
- **Moved to userland:** the RN uncontrolled-`TextInput` measurement fix — `BaseTextInputShadowNode` was refactored in 0.85 so the old patch no longer applies; the chat input height reset is now handled in app code instead of patching `react-native`.
- **Dropped — the fix is upstream in the bumped version:** `@tiptap/react` (the `!editor.isDestroyed` guard, in 2.6.6); `expo-image-picker` (`shouldDownloadFromNetwork`, in 56.0.14); `expo-task-manager` (the headless-JS timer keep-alive, in 56.0.15).
- **Dropped — no longer applicable:** `usehooks-ts` (no longer in the dependency tree); `@10play/tentap-editor@0.4.55` (stale duplicate of the live `0.5.21` patch); `expo-share-intent` (the share-extension customization now lives in committed native source at `apps/tlon-mobile/ios/ShareExtension/ShareViewController.swift`).
- **Restored:** `@urbit/http-api`. The 3.1.0-dev-3 → 3.2.0-dev bump dropped the `urlTransformer` patch, but 3.2.0-dev doesn't carry it upstream (its constructor is still `(url, code, desk, fetchFn)`), while tlon-web still passes `hostingUrl` as the 5th constructor arg ([api.ts](apps/tlon-web/src/api.ts)). Without it, hosted-ship channel pokes silently lose their `?mark=&actions=&kind=` annotations, so the patch was re-ported onto 3.2.0-dev.

`pnpm.allowNonAppliedPatches` is `false`, so `pnpm install` fails if any patch stops applying — all listed patches apply against the SDK 56 tree.

## How did I test?

- **iOS:** `expo run:ios` — `Build Succeeded`, installs and boots on the simulator, connects to Metro, and loads a real ship (authenticated channel traffic over the event stream). Built with prebuilt React Native (SDK 56 default).
- **Android:** `expo run:android` / gradle `assembleProductionDebug` — `BUILD SUCCESSFUL`, debug APK installs and launches on an emulator.
- **CI (green):** `test-build` (typecheck + prettier), the **Production Build Smoke Test** (the web prod build now boots past the safe-area crash), and the parallel E2E suite all pass.
- `expo-doctor` clean; `pnpm -r tsc` passes; `pnpm install` is green with all patches applying and a single resolved copy of the pinned RN libraries.

## Risks and impact

- **OTA behavior change (fingerprint):** because `runtimeVersion` is now the fingerprint policy, the EAS build + `eas update` pipeline must compute the fingerprint consistently, and native-affecting changes will require new binaries more often (that's the compatibility guarantee, not a regression). Worth confirming the update workflow before relying on OTA.
- Safe to rollback without consulting PR author? **Yes** — revert the PR to return to SDK 54, but it requires `pnpm install` + a clean native rebuild (pods / gradle) afterward since native folders and the lockfile change.
  - [ ] Onboarding
  - [x] State / providers (provider stack: `SafeAreaProvider` initial metrics)
  - [ ] Message sync
  - [ ] Channel display
  - [x] Notifications
  - [x] Other: native build system (iOS Podfile/AppDelegate, Android gradle/MainApplication), app startup, dependency tree, OTA runtime version

  *(Notifications: `AppDelegate.swift` remote-notification forwarding and `MainApplication.kt` notification-channel setup are in the touched startup files; the handler logic itself is unchanged.)*

## Rollback plan

Revert the PR's commits (or revert the merge), then run `pnpm install`, `pod install`, and a clean gradle build. No data migrations or server-side changes are involved, so rollback is purely client-side.

## Screenshots / videos

No user-facing UI change. Onboarding/welcome, sign-up, and authenticated app boot render correctly on both iOS and Android simulators. Can attach captures on request.


dnbrwstr (approved):
Changes lgtm

PR #5888, opened 2026-05-29
activity: fix v9 json conversions

mikolajpp:
## Summary

We missed a few JSON conversions in activity. This was found on 408k, which builds all marks, but we back-port the fix. 

## Changes

1. Fill in missing activity JSON conversions.

## How did I test?

Compiled.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan
Revert.

### DIFF
--- a/desk/lib/activity-json.hoon
+++ b/desk/lib/activity-json.hoon
@@ -261,6 +261,31 @@
       :~  time+s+(scot %ud time.te)
           event+(event event.te)
       ==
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Good call on backporting, thank you!

PR #5874, opened 2026-05-22
Aqua tests: groups agent coverage – part 1

mikolajpp:
## Summary

This PR contains another batch of model-generated aqua tests, as well as various improvements to the tooling and documentation.

## Changes

1. Generated aqua tests for various group join, invite revocation and group preview scenarios.
2. Various other aqua test adjustements
3. Reorganize docs
4. Temporarily adjust test runner for 408k

## How did I test?

Run aqua tests locally with the runner.

Note: only aqua tests will show up as passing in CI. There is no e2e zod available for 408k yet, so aqua tests use their own set of fakeships. 

## Risks and impact

- Safe to rollback without consulting PR author? Yes
- Affects important code area:
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.

PR #5889, opened 2026-06-01
groups: fix pending join not preserving roles

mikolajpp:
## Summary

Found while expanding test coverage. The group host would actually emit a fact containing assigned roles, but would not update them in the state. Fortunately, aqua sees the perspective of both ships, and so when we examine the final group creation fact we could see the roles were missing.

## Changes

1. Fix pending join not preserving roles.

## How did I test?

Tested in aqua on #5874. Here we only backport the fix.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.

### DIFF
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -2646,6 +2646,8 @@
         =+  seat=(~(got by seats.group.se-core) ship)
         =?  roles.seat  ?=(^ roles)
           (~(uni in roles.seat) u.roles)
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Obvious in hindsight. Test expansion working as intended! Thanks.

PR #5890, opened 2026-06-02
fix chat previews after database purge

dnbrwstr:
## Summary

Fixes missing chat previews after a local SQLite purge or fresh database rebuild, and addresses a foreground refetch race that could leave chat-list previews stale after sync. Fixes three issues:
- Sync and init timestamps were not completely reset when we dropped the database
- `setLastPosts` was a no-op when called from `insertChannels`, `insertGroups`, etc.
- Table-based query invalidation skipped queries that were already fetching, so `getChats` could miss writes that landed during an active refetch

Fixes TLON-5915
Fixes TLON-5852

## Changes

- Reset all DB sync markers when native, web, or electron SQLite storage is purged or recreated.
- Make `setLastPosts` repair previews from local rows when channel or group metadata arrives after posts.
- Backfill post `groupId` during last-post repair so posts inserted before channel rows still attach to their group.
- Preserve newer server sequence watermarks while repairing local preview metadata.
- Only invalidate `posts` after group inserts when the incoming payload includes embedded channels.
- Invalidate table-dependent queries even if they are already fetching, so active chat-list refetches get a follow-up read after later writes.
- Add tests for post-before-channel ordering, first-sync init data repair, metadata repair, dynamic table effects, and active-refetch invalidation.

## How did I test?

Reproduced on physical Android device from a clean/purged DB state, then verified previews were repaired after first sync. Also ran:

- `pnpm --filter @tloncorp/shared tsc`
- `pnpm --filter @tloncorp/shared exec vitest run src/db/query.test.ts src/db/queries.test.ts src/store/sync/sync.test.ts`

## Risks and impact

- [ ] Onboarding
- [x] State / providers
- [x] Message sync
- [x] Channel display
- [ ] Notifications
- [ ] Other:

### DIFF
--- a/packages/app/lib/baseDb.ts
+++ b/packages/app/lib/baseDb.ts
@@ -1,5 +1,5 @@
 import { createDevLogger } from '@tloncorp/shared';
-import { handleChange } from '@tloncorp/shared/db';
+import * as sharedDb from '@tloncorp/shared/db';
 import { perfMark } from '@tloncorp/shared/perfLog';
 import { sql } from 'drizzle-orm';
 import { integer, sqliteTable, text } from 'drizzle-orm/sqlite-core';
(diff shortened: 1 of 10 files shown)


dnbrwstr:
@claude review once

patosullivan (approved):
Lgtm

PR #5893, opened 2026-06-03
signup: avoid showing web prompt immediately on home screen

latter-bolden:
## Summary

We recently added a "Try Tlon on Web" sheet to make mobile signups aware of our web app. Right now, it displays immediately when you land in the app. This creates issues for our E2E test, but also feels jarring — you have to both register the information about web while also seeing the home screen in the background for the first time.

I'd like to defer this slightly so it's shown the second time the app is opened rather than immediately.

## Changes

- add key value variable to track opens for the web prompt
- only display it after the app has been opened at least twice

## How did I test?

iOS sim

## Risks and impact

  - [x] Onboarding


patosullivan (approved):
Lgtm

PR #5895, opened 2026-06-04
fix reaction taps

dnbrwstr:
## Summary

Fixes iOS reaction chips so tapping an existing emoji adds your reaction. The native tooltip trigger does not own the press target, so the chip now keeps the tooltip wrapper while the rendered button handles taps. 

Fixes TLON-5943

## Changes

- Move reaction tap and long-press handlers from `Tooltip.Trigger` to the rendered reaction chip.

## How did I test?

Tested on device


dnbrwstr:
@claude review once

patosullivan (approved):

PR #5897, opened 2026-06-04
notes group integration: generic channel-host convention + membership

arthyn:
## Summary

This improves %groups ability to have non-%channels agents integrate with it as well as makes the integration with our client more "real" and less of a hack. This is in service of us trying to build an actual UI integration with %notes.

Companion to https://github.com/arthyn/notes/pull/7

## Changes

Adds support for `%notes` notebooks as a group channel type, via a generic "channel-host" convention so `%groups` can drive a backing agent it doesn't know about.

**`%groups` agent**
- Generic channel-host convention: a nest whose kind isn't `%chat`/`%diary`/`%heap` is backed by the agent named by the kind. `join-channels`/`leave-channels` poke `%<kind>-join`/`%<kind>-leave`; `is-joined` reads a conventional `/joined/<host>/<name>` scry for `active-channels`. chat/diary/heap keep their existing `%channels` fast-path untouched. No `%groups` state change.
- New `group-channel-active` poke (+ `mar/group/channel-active`) lets a backing agent report channel join/leave so `active-channels` stays current without `%groups` subscribing to it.

**Client**
- Notes channels are created via the `%notes` HTTP API (`POST /notes/~/v1/notebooks`), which returns the server-assigned flag deterministically (added `Urbit.requestJson` + an `api.requestJson` wrapper).
- Membership for notes channels is reconciled from each group's `active-channels` set (`setJoinedNotesChannels`, fed from `initApi`) rather than `%channels`' joined list, which doesn't track notes. `setLeftGroupChannels` is scoped to `%channels`-backed kinds.
- Join/leave for notes channels are routed to `%notes` instead of `%channels`; `%channels`-only affordances that don't apply to notes are suppressed (hooks preview, mark-as-read).

## How did I test?

- `%groups` compiles against the `develop` groups desk; the `%notes` desk's unit suite passes (84/84), including the state migration.
- Manually on a dev ship: create-in-group, join/leave a notes channel, and `active-channels` population (host- and subscriber-side).

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [x] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [x] Other: `%groups` agent (channel-host convention)

Additive: the generic channel-host path only fires for non-`%channels` kinds (unknown/uninstalled kinds no-op), and notes-specific client branches are gated on the `notes/` id prefix, so chat/diary/heap behavior is unchanged. No `%groups` state migration.

## Rollback plan

Revert the commits on this branch. There is no `%groups` state migration to unwind, and the client changes are gated to notes channels, so reverting restores prior behavior cleanly. The `%notes` companion desk is deployed separately and is inert for clients without these changes.


desk/app/groups.hoon
```diff
@@ -497,6 +497,22 @@
...
+      ?>  =(our src):bowl
+      =+  !<([=flag:g =nest:g joined=?] vase)
+      ?~  net-group=(~(get by groups) flag)  cor
+      =*  group  u.net-group
+      %_  cor  groups
+        %+  ~(put by groups)  flag
+        %_  group  active-channels
+          ?:  joined
+            (~(put in active-channels.group) nest)
+          (~(del in active-channels.group) nest)
```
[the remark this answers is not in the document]
arthyn: Acknowledged — deferring this for now (tracked alongside the PR, not fixed in this change).

The fix needs a genuine new `%groups` response variant: we explored re-emitting the existing `%join` fact, but that carries the deprecated per-channel auto-join *setting* semantics, not "membership of this nest changed," so the client would misinterpret it. A correct fix is a dedicated `active-channels`-changed `r-groups`/`go-response` that the client maps to the same join/leave reconcile path the full init sync uses.

Practical impact today is limited to multi-device: a second active client for the same ship lags on notes join/leave until its next full init sync (single-device and fresh loads are unaffected, since membership is derived from `active-channels` in the init payload). We will follow up with the new response variant rather than overload `%join`.

packages/shared/src/store/channelActions.ts
```diff
@@ -43,7 +43,6 @@ export async function createChannel({
       groupId,
       title,
       description: rawDescription,
-      readers,
     });
```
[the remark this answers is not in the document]
arthyn: Fixed in 56d3f23e8 (client) + notes `2f32c52` (agent). `createNotesChannel` now forwards the `readers` roles it receives into the `POST /v1/notebooks` body, and the %notes host registers the group channel with those reader roles via a new `%create-group-notebook` action (group-action-4 channel-add carries `readers` instead of `~`). Private notes channels now keep their role restrictions; `go-can-read` gates them correctly. Covered by a new agent test asserting the emitted poke carries the readers verbatim.

desk/app/groups.hoon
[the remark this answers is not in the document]
arthyn: Update: implemented in 660918de8 rather than deferring. Added a local-only `%active-channels` `r-group` response variant carrying the per-nest membership delta; the `%group-channel-active` arm emits it on `/v1/groups` (a `from-self`-only path, so it never crosses the wire) after updating `active-channels`. The client reduces it to the existing `joinChannel`/`leaveChannel` `GroupUpdate`, so a second same-ship client now reconciles membership without waiting for a full init sync. Discipline on `group-response-1` was relaxed (`&`→`|`) since its carried type changed.

desk/app/groups.hoon
```diff
@@ -1713,6 +1738,25 @@
 ++  watch-channels
   (safe-watch /channels [our.bowl %channels] /v1)
 ::
+::  +set-active-channel: record a channel-host membership change in the
```
Fang-: The wording here (and in the description of this in general) is _very_ overloaded lol. Taking out the "membership" alone would make it less confusing.

desk/app/groups.hoon
```diff
@@ -1713,6 +1738,25 @@
...
+::  group's active-channels set and push a local-only %active-channels delta
+::  to same-ship clients. /v1/groups is a from-self-only sub, so the fact
+::  never crosses the wire — it just lets a second local client reconcile
+::  join/leave without a full init sync. no-op if already in that state.
```
Fang-: I'm having trouble understanding what this is trying to say. Isn't this just doing a normal local state update and corresponding response fact?

desk/app/groups.hoon
```diff
@@ -3452,13 +3470,17 @@
...
+      ?:  ?=(?(%chat %diary %heap) p.nes)
+        =/  =dock  [our.bowl %channels]
+        =/  action=a-channels:v9:dv  [%channel nes %leave ~]
+        `[%pass wire %agent dock %poke channel-action-1+!>(action)]
+      ::  generic channel-host: the nest kind names the backing agent on our
+      ::  ship; poke %<kind>-leave carrying the nest. An uninstalled kind just
+      ::  nacks (logged in go-agent, otherwise harmless).
+      =/  =dock     [our.bowl p.nes]
+      =/  =mark     `@tas`(crip (weld (trip p.nes) "-leave"))
+      `[%pass wire %agent dock %poke mark !>(nes)]
```
Fang-: This should probably just be a static `%group-channel-leave` mark. (And likewise for join.) The data is of a fixed shape, the meaning is fixed. Not obvious to me why we need per-kind marks.

Fang- (commented):
As discussed out of band, I don't hate this. Some comments/questions below. In particular, not clear to me why we would want to do dynamic marks for the status update pokes that come out of groups agent.

desk/sur/groups.hoon
```diff
@@ -278,6 +274,16 @@
...
+::  $channel-active: channel join/leave report. Third-party backing agents
+::  poke this to keep $active-channels current.
+::
++$  channel-active  [=flag =nest joined=?]
+::  $channel-join / $channel-leave: %groups tells a channel-host agent (named
+::  by the nest kind) to join or leave `nest`. join carries the group flag so
+::  the host can record affiliation. See +join-channels / +leave-channels.
+::
++$  channel-join   [=nest group=flag]
++$  channel-leave  [=nest]
```
Fang-:
```suggestion
+$  channel-join   [=nest group=flag]
+$  channel-leave  nest
```

desk/app/groups.hoon
```diff
@@ -3476,11 +3475,10 @@
...
       ::  generic channel-host: the nest kind names the backing agent on our
-      ::  ship; poke %<kind>-leave carrying the nest. An uninstalled kind just
-      ::  nacks (logged in go-agent, otherwise harmless).
-      =/  =dock     [our.bowl p.nes]
-      =/  =mark     `@tas`(crip (weld (trip p.nes) "-leave"))
-      `[%pass wire %agent dock %poke mark !>(nes)]
+      ::  ship; poke it %group-channel-leave with the nest. An uninstalled
+      ::  kind just nacks (logged in go-agent, otherwise harmless).
+      =/  =dock  [our.bowl p.nes]
+      `[%pass wire %agent dock %poke group-channel-leave+!>(`channel-leave:g`[nes])]
```
Fang-: Eh, depending on the situation it might queue up indefinitely in local gall state. But we aren't proactively pushing content into it or anything, so I'm not concerned about those queues getting big.

desk/app/groups.hoon
```diff
@@ -3495,10 +3493,10 @@
...
         =/  action=a-channels:v9:dv  [%channel nes %join flag]
         `[%pass wire %agent dock %poke channel-action-1+!>(action)]
-      ::  generic channel-host: poke %<kind>-join carrying [nest group-flag].
-      =/  =dock     [our.bowl p.nes]
-      =/  =mark     `@tas`(crip (weld (trip p.nes) "-join"))
-      `[%pass wire %agent dock %poke mark !>([nes flag])]
+      ::  generic channel-host: poke the kind-named agent %group-channel-join
+      ::  carrying [nest group-flag].
+      =/  =dock  [our.bowl p.nes]
+      `[%pass wire %agent dock %poke group-channel-join+!>(`channel-join:g`[nes flag])]
```
Fang-: This (and above) feels distinctly non-human (the comment needlessly verbose/redundant with the code, defining `=dock` ahead of time but doing the cage inline instead of the other way around). Kind of odd too since I don't think there's too many examples of code like that, but might be a side-effect of where it's putting attention/focus.

Fine™ as-is either way.

Fang- (commented):
Changes look good, thank you. Curious to hear @mikolajpp's thoughts too, but good to go afaic unless he has big objections.

dnbrwstr:
Took a look at the client code here -- the main thing I struggle with a bit is having to differentiate between groups-backed and notes-backed channels. Ideally the client shouldn't have to know about that. Is there a way to collapse or abstract over the difference?

desk/sur/groups.hoon
```diff
@@ -274,6 +274,16 @@
...
 ::
 +$  flagged-content  (map nest (jug plan ship))
+::  $channel-active: channel join/leave report. Third-party backing agents
+::  poke this to keep $active-channels current.
+::
++$  channel-active  [=flag =nest joined=?]
+::  $channel-join / $channel-leave: %groups tells a channel-host agent (named
+::  by the nest kind) to join or leave `nest`. join carries the group flag so
+::  the host can record affiliation. See +join-channels / +leave-channels.
+::
```
mikolajpp, line 284:
```suggestion
::  $channel-join / $channel-leave: %groups tells a channel-host agent (named
::  by the nest kind) to join or leave .nest. join carries the group flag so
::  the host can record affiliation. see +join-channels / +leave-channels in groups.
::
```
We generally don't use markdown in comments, but have our own Hoon-esque convention.

desk/sur/groups.hoon
```diff
@@ -274,6 +274,16 @@
 ::  $flagged-content: flagged posts and replies that need admin review
 ::
 +$  flagged-content  (map nest (jug plan ship))
+::  $channel-active: channel join/leave report. Third-party backing agents
+::  poke this to keep $active-channels current.
+::
```
mikolajpp:
```suggestion
::  $channel-active: channel join/leave report. third-party backing agents
::  poke this to keep $active-channels current.
::
```
Somebody please remind the robot to keep comments in style.

desk/sur/groups.hoon
```diff
@@ -564,6 +574,11 @@
       [%section =section-id =r-section]
       [%section-order order=(list section-id)]
       [%flag-content =nest =plan src=ship]
+      ::  per-nest active-channels (membership) delta. emitted local-only on
+      ::  /v1/groups so a second same-ship client learns join/leave without a
+      ::  full init sync; never crosses the wire. See the %group-channel-active
+      ::  poke arm in app/groups.hoon.
```
mikolajpp: I assume these are model artifacts. It should not be necessary to put so much detail for what is obviously just a new case in a groups response. If the mechanism requires explanation, it should be located close to the actual implementation. Even a model will be able to locate it without any problem by grepping for call sites as they usually do.

```suggestion
```

desk/sur/groups-ver.hoon
```diff
@@ -126,6 +126,8 @@
           [%section =section-id =r-section]
           [%section-order order=(list section-id)]
           [%flag-content =nest =plan src=ship]
+          ::  per-nest active-channels (membership) delta; local-only response.
+          [%active-channels =nest joined=?]
```
mikolajpp:
```suggestion
          [%active-channel =nest joined=?]
```
The documentation style for addition of type variants in `*-ver` files is to put the new/modified variant with a comment at top level, as we do for `%section-order`. 

We should use a singular here, unless we wanted to transfer the whole set of active channels.
(Looks like `%seat` and `%role` violate this naming scheme in the opposite direction, but out of scope here.)

That said, I am not sure if we can get away with retroactively changing v10 types. There are two questions to consider:
1. Does discipline marked this type strict? If so, a new mark and a new version is required.
2. Will the client silently drop unknown cases of this type?

desk/mar/group/channel-join.hoon
```diff
@@ -0,0 +1,12 @@
+/-  g=groups
```
mikolajpp, line 1: Looks like the intention for this and other marks is to be transferred only between urbit agents. If that is the case, they can remain at current version.

desk/sur/groups.hoon
```diff
@@ -564,6 +574,11 @@
...
+      [%active-channels =nest joined=?]
```
mikolajpp:
```suggestion
      [%active-channel =nest joined=?]
```
We would say `%active-channels` in the plural if the delta carried the whole set of active channels, which is not the case here.

desk/lib/groups-json.hoon
```diff
@@ -209,8 +209,14 @@
             'r-section'^(r-section r-section.r-group)
         ==
       ::
-          %section-order  
+          %section-order
         (frond 'section-order' a+(turn order.r-group (lead %s)))
+      ::
+          %active-channels
```
mikolajpp:
```suggestion
          %active-channel
```

desk/lib/groups-conv.hoon
```diff
@@ -143,6 +143,9 @@
         |=  [=r-group:v9:gv seats=(map ship seat:v9:gv) =admissions:v9:gv]
         ^-  (list diff:v2:gv)
         =/  rs-group-7=(list r-group:v7:gv)
+          ::  %active-channels is a local-only response with no v7/v2
+          ::  backcompat representation; drop it from the legacy diff.
+          ?:  ?=([%active-channels *] r-group)  ~
```
mikolajpp:
```suggestion
          ?:  ?=([%active-channel *] r-group)  ~
```

desk/app/groups.hoon
```diff
@@ -84,8 +87,13 @@
           :+  %channel-preview    &  -:!>(*vale:m-channel-preview)
           :+  %channel-preview-1  &  -:!>(*vale:m-channel-preview-1)
         ::
-          :+  %group-response-1   &  -:!>(*vale:m-group-response-1)
+          ::  relaxed (|): r-group gained the local-only %active-channels
+          ::  variant, so its type hash must not be pinned into negotiation.
+          :+  %group-response-1   |  -:!>(*vale:m-group-response-1)
```
mikolajpp: After we switch off the flag, we lose the advantage of discipline on this type. Furthermore there is no hope of ever turning it strict again because that would prevent legacy ships from upgrading due to nest fails on this type. (Because it was strict once in the past.)

We should have a new mark here. We can probably reuse the same subscription path (if the client can handle that gracefully), provided we update the set of facts in discipline. 

I also don't think it has anything to do with negotiation? In fact, the opposite is the case: we use relaxed flags for marks that *are* shielded with lib-negotiate. (Dear model: there is also no type hash pinning going on, I think discipline simply runs nest checks).

We can only do relaxed -> strict transition once. After that types can only remain set in stone.

desk/app/groups.hoon
```diff
@@ -84,8 +87,13 @@
...
           :+  %group-action-3     &  -:!>(*vale:m-group-action-3)
+          :+  %group-channel-active  &  -:!>(*vale:m-group-channel-active)
+          :+  %group-channel-join    &  -:!>(*vale:m-group-channel-join)
+          :+  %group-channel-leave   &  -:!>(*vale:m-group-channel-leave)
```
mikolajpp: What's the indented interface model here? Should not third party agents use protocol negotiation to interface with groups, and shouldn't groups bump its protocol with these changes? These marks use current types, so it looks like we want to make them relaxed.

desk/app/groups.hoon
```diff
@@ -649,6 +664,21 @@
   ^-  path
   /(scot %p our.bowl)/channels/(scot %da now.bowl)/[p.nest]/(scot %p p.q.nest)/[q.q.nest]
 ::
+::  +is-joined: are we subscribed to (or host of) this channel? Built-in
+::  channels kinds scry %channels; any other kind uses the generic
+::  channel-host convention — the nest kind names the backing agent and we
+::  %gu its /joined/<host>/<name> path. An uninstalled kind reads as |.
```
mikolajpp:
```suggestion
::  %gu its /joined/<host>/<name> path. An unsupported kind reads as |.
```

desk/app/groups.hoon
```diff
@@ -649,6 +660,21 @@
...
+++  is-joined
+  |=  =nest:g
+  ^-  ?
+  ?:  ?=(kind:d p.nest)
+    .^(? %gu (channels-scry nest))
+  ::  generic channel-host: the kind agent answers a /joined/<host>/<name>
+  ::  scry with a loob.
+  =/  =path
+    /(scot %p our.bowl)/[p.nest]/(scot %da now.bowl)/joined/(scot %p p.q.nest)/[q.q.nest]
+  .^(? %gu path)
```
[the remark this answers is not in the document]
mikolajpp: Waiting to be addressed @arthyn ?

desk/app/groups.hoon
```diff
@@ -1713,6 +1742,20 @@
...
+  ^+  cor
+  ?~  net-group=(~(get by groups) flag)  cor
+  =/  [=net:g =group:g]  u.net-group
+  ?:  =(joined (~(has in active-channels.group) nest))  cor
+  =.  active-channels.group
+    ?:  joined
+      (~(put in active-channels.group) nest)
+    (~(del in active-channels.group) nest)
+  =.  groups  (~(put by groups) flag net group)
+  =/  =r-groups:v9:gv  [flag [%active-channels nest joined]]
```
mikolajpp:
```suggestion
  =/  =r-groups:v9:gv  [flag [%active-channel nest joined]]
```

desk/app/groups.hoon
```diff
@@ -1727,39 +1770,15 @@
...
-      =*  flag  group.perm.rc
-      ?.  (~(has by groups) flag)  cor
-      =+  group=(~(got by groups) flag)
-      %_  cor  groups
-        %+  ~(put by groups)  flag
-        %_  group  active-channels
-          (~(put in active-channels.group) nest.r-channels)
-        ==
-      ==
+      (update-active-channel group.perm.rc nest.r-channels &)
```
mikolajpp, line 1744: How this turns out is pretty neat!

desk/app/groups.hoon
```diff
@@ -3452,13 +3471,16 @@
...
-      =/  =cage  channel-action-1+!>(action)
       =/  =wire  (snoc go-area %leave-channels)
-      `[%pass wire %agent dock %poke cage]
+      ?:  ?=(?(%chat %diary %heap) p.nes)
+        =/  =dock  [our.bowl %channels]
+        =/  action=a-channels:v9:dv  [%channel nes %leave ~]
+        `[%pass wire %agent dock %poke channel-action-1+!>(action)]
+      ::  generic channel-host: the nest kind names the backing agent on our
+      ::  ship; poke it %group-channel-leave with the nest. An uninstalled
+      ::  kind just nacks (logged in go-agent, otherwise harmless).
```
mikolajpp:
```suggestion
      ::  ship; poke it %group-channel-leave with the nest. an unsupported
      ::  kind just nacks (logged in +go-agent, otherwise harmless).
```

desk/app/groups.hoon
mikolajpp: I would say it is *very* clear from context what is happening, but given that this is new and unusual logic the comments can stay. I would indeed inline dock and define the cage if anything.

mikolajpp (changes requested):
Overall this looks good. I like the minimally invasive integration. 

There is a whole bunch of nits and styling comments.

The two issues to be addressed surround discipline and negotiation. Firstly, we absolutely should not relax marks that were once strict, and are used by any external integrations, like the client. Doing so defeats the purpose of the discipline.

Secondly, it looks like the third-party channel host can be a remote ship too. If that's the case, it will eventually be necessary to bump groups protocol. Perhaps it is fine to have an internal version running first without that. CC @Fang-

dnbrwstr (approved):
Nice, lgtm!

desk/sur/groups.hoon
```diff
@@ -274,6 +274,16 @@
...
 +$  flagged-content  (map nest (jug plan ship))
+::  $channel-active: channel join/leave report. third-party backing agents
+::  poke this to keep $active-channels current.
+::
++$  channel-active  [=flag =nest joined=?]
+::  $channel-join / $channel-leave: %groups tells a channel-host agent (named
+::  by the nest kind) to join or leave .nest. join carries the group flag so
+::  the host can record affiliation. see +join-channels / +leave-channels in groups.
+::
++$  channel-join   [=nest group=flag]
```
mikolajpp: We use `flag` everywhere, while `group` usually implies actual group structure.
```suggestion
+$  channel-join   [=nest =flag]
```

desk/app/groups.hoon
```diff
@@ -497,6 +506,13 @@
         ?.  valid.i.invites  $(invites t.invites)
         token.i.invites.u.far
       fi-abet:(fi-join:(fi-abed:fi-core flag.join) tok)
+    ::
+    ::  allows third-party agents (e.g. %notes) to report channel join/leave
+    ::  for active-channels bookkeeping.
+        %group-channel-active
```
mikolajpp, line 476:
```suggestion
        ::  allows third-party agents (e.g. %notes) to report channel join/leave
        ::  for active-channels bookkeeping.
        ::
        %group-channel-active
```

desk/app/groups.hoon
```diff
@@ -645,6 +661,24 @@
   ^-  path
   /(scot %p our.bowl)/channels/(scot %da now.bowl)/[p.nest]/(scot %p p.q.nest)/[q.q.nest]
 ::
+::  +is-joined: are we subscribed to (or host of) this channel? Built-in
+::  channels kinds scry %channels; any other kind uses the generic
+::  channel-host convention — the nest kind names the backing agent and we
+::  %gu its /joined/<host>/<name> path. An unsupported kind reads as |.
+++  is-joined
```
mikolajpp:
```suggestion
::  +is-joined: are we subscribed to (or host of) this channel? 
::
::  built-in channels kinds scry %channels; any other kind uses the generic
::  channel-host convention — the nest kind names the backing agent and we
::  %gu its /joined/<host>/<name> path. An unsupported kind reads as |.
::
++  is-joined
```

We could probably use an explicit styling guide for comments if this was model generated. The convention is to `+arm: brief description`, followed by a gap and one or more paragraphs in funny lowercase.

desk/app/groups.hoon
```diff
@@ -1104,6 +1138,10 @@
     ::
   ::
     [%v1 %groups ~]  ?>(from-self cor)
+  ::
+    ::  v2 firehose: full v10 r-group, incl. the %active-channel membership
+    ::  delta. v1 stays group-response-1 for agents that blind-cast to v9.
+    [%v2 %groups ~]  ?>(from-self cor)
```
mikolajpp:
```suggestion
    [%v2 %groups ~]  ?>(from-self cor)
```
I am not sure we need a comment here at all. These info can easily be found elsewhere.

The second part of this comment is not clear to me: `v1` just continues to disseminate `v9` updates, is there anything special there?

desk/app/groups.hoon
```diff
@@ -3483,13 +3510,16 @@
...
       =/  =wire  (snoc go-area %leave-channels)
-      `[%pass wire %agent dock %poke cage]
+      ?:  ?=(?(%chat %diary %heap) p.nes)
+        =/  =dock  [our.bowl %channels]
+        =/  action=a-channels:v9:dv  [%channel nes %leave ~]
+        `[%pass wire %agent dock %poke channel-action-1+!>(action)]
+      ::  generic channel-host: the nest kind names the backing agent on our
+      ::  ship; poke it %group-channel-leave with the nest. an unsupported
+      ::  kind just nacks (logged in +go-agent, otherwise harmless).
+      =/  =dock  [our.bowl p.nes]
```
mikolajpp:
```suggestion
      ::  generic channel-host: the nest kind names the backing agent on our ship.
      ::
      =/  =dock  [our.bowl p.nes]
```

mikolajpp: We should also add the same comment for joins below.

desk/app/groups.hoon
```diff
@@ -4636,6 +4660,12 @@
     =*  log  ~(. l `'group-join')
     ::=.  cor  (tell:log %dbug leaf+"+go-response /v1/groups fact {<r-groups-9>}" ~)
     =.  cor  (give %fact v1-paths group-response-1+!>(r-groups-9))
+    ::  v2 response: same r-group as v1, carried by the v10 mark so v2
+    ::  subscribers also receive the %active-channel deltas from elsewhere.
+    ::
```
mikolajpp:
```suggestion
    ::  v2 response: same $r-group payload as v1, carried carried by the v10 mark so v2
    ::  subscribers can also receive the %active-channel deltas sent out by +update-active-channel.
    ::
```

mikolajpp (approved):
Good job, the code looks really good now. Left some styling/commenting nits, otherwise it is good to merge.

desk/app/groups.hoon
arthyn: yeah agreed

PR #5898, opened 2026-06-05
Route group-invite push notifs to the invite preview sheet

patosullivan:
## Summary

Fixes TLON-5887: group-invite push notifications not routing anywhere when tapped. The parser returned `null` for `group-invite` events, so the notification listener had nothing to navigate to. This parses the event and routes the tap to the chat list, opening the group preview sheet in a loading state until the invite row lands, then showing Accept/Reject.

## Changes

- Parsed `group-invite` push events into a routeable `{ type: 'groupInvite', groupId }` payload and added `GroupInviteNotificationData` to the notification union, in `apps/tlon-mobile/src/lib/notificationPayload.ts`. Removed `group-invite` from the `return null` fallthrough (other unsupported events still return `null`).
- Added a `groupInvite` case to the notification listener in `apps/tlon-mobile/src/hooks/useNotificationListener.ts`. It resets to `ChatList` via the exported `groupInvitePreviewRouteStack(groupId)` helper, using the existing `createTypedReset` pattern so the destination is deterministic from any starting stack (cold start, a channel screen, or already on ChatList).
- Added a `previewGroupFromInviteNotification?: boolean` marker to the `ChatList` route params in `packages/app/navigation/types.ts`. It is set only by the notification route; deep links and other callers omit it.
- Added notification-scoped invite gating to `packages/app/features/top/ChatListScreen.tsx`. It tracks the invite-notification group distinctly from normal selections, reacts to a later `previewGroupId` param (mirroring desktop `HomeSidebar`) so a tap while already mounted reopens the sheet, and runs a group-keyed, sync-readiness-gated bounded timer that only counts down once `connStatus === 'Connected'` and `session.phase` is `low`/`ready`. If the invite never resolves within the window it closes the sheet and logs `ErrorPushNotifNavigate` rather than degrading to generic group-preview actions.
- Added `packages/app/features/top/groupInvitePreview.ts`, a pure module holding the sheet open/group/terminal-close decision logic (the component still owns timer setup and sync-readiness gating). `isGroupInviteReady` is true only on a landed invite or existing membership (deliberately excludes `haveRequestedInvite`), and `getGroupInviteSheetState` reduces non-notification selections to today's exact behavior while giving notification selections the loading-spinner-until-ready, never-hand-a-non-ready-row, bounded-close treatment.
- Added tests: a `group-invite` parser test plus route-stack and classification asserts in `apps/tlon-mobile/src/__uitests__/useNotificationListener.test.ts`, and unit coverage for `isGroupInviteReady` and `getGroupInviteSheetState` (including the stale-A-vs-B and terminal-close cases) in `packages/app/features/top/groupInvitePreview.test.ts`.

## How did I test?

- Manually verified on physical Android and iOS devices.

## Risks and impact

  - [x] Notifications

## Rollback plan

Revert.


dnbrwstr (approved):
Lgtm

PR #5904, opened 2026-06-09
Add drag and drop for pinned chat list items

patosullivan:
## Summary

Resolves [TLON-5948](https://linear.app/tlon/issue/TLON-5948/pinned-dmsgroups-are-glued)

Adds drag-and-drop reordering for pinned items in the chat list. Pinned DMs, groups, and channels were previously locked into pin-time order; users can now enter a sort mode and manually reorder them on web, iOS, and Android.

## Changes

**App UI (**`packages/app`**)**

* Moved the pinned section into the ChatList `FlashList` header with a Sort/Done toggle (`PinnedSortToggle` in `ChatList.tsx`). Only the non-pinned sections feed the virtualized list; the pinned block renders as the list header.
* Added the sortable pinned list: web via `@dnd-kit` (`SortablePinnedChats.web.tsx`), native via `react-native-sortables` (`SortablePinnedChats.tsx`). Sortable identity is `pin.itemId` (a pinned group channel's `chat.id` differs from its persisted pin id). Failed persists roll back the optimistic order on both platforms.
* Added the `usePinnedChatOrdering` hook (plus `usePinnedChatOrdering.helpers.ts`), which maps the visible (tab-filtered) pinned chats to sortable items and reconstructs the full global order from a visible-subset reorder before persisting, so reordering in a filtered tab never drops or relocates a pin the user cannot see.
* `MessagesSidebar`, `HomeSidebar`, and `ChatListScreen` now pass the full pinned set (`allPinnedChats`) so reorders in filtered tabs preserve hidden pins. Restored the render gate in `MessagesSidebar` to key off section data presence.

**Shared (**`packages/shared`**)**

* Added `setPinnedItemsOrder` (rewrites each pin's index to its position) and `clearPinnedItems` (explicit clear for the failure-reconcile empty-snapshot case) DB queries. Added `pins` to `getChats` query dependencies so reorder-only writes invalidate the chat-list query.
* Added the `reorderPinnedItems` store action in `channelActions.ts`. It writes the full merged order locally (optimistic plus a post-success reassert), but the backend poke and reassert send only the user-reordered visible ids, so hidden pins in filtered views keep their server slots even under cross-device drift. The action never throws (boolean contract), reconciles from the backend on poke failure, and rolls back locally (including clearing pins on an authoritative empty snapshot) when reconcile also fails.

**API (**`packages/api`**)**

* Added the `setPinnedItemOrder` client poke to `groups-ui`.

**Backend (Hoon,** `desk/`**)**

* Added a `%set-order` variant to `a-pins` in `sur/ui.hoon` and handled it in the `groups-ui` agent (`app/groups-ui.hoon`). The handler is slot-preserving: it reorders only the named (visible) pins into the slots they currently occupy, leaves omitted pins fixed in place, dedupes, and ignores unknown ids.
* Added the `set-order` JSON dejs path in `lib/groups-json.hoon`.

**Tests**

* Shared unit tests (`reorderPinnedItems.test.ts`, `queries.test.ts`), app helper tests (`usePinnedChatOrdering.helpers.test.ts`, `ChatList.helpers.test.ts`), backend desk tests (`desk/tests/app/groups-ui.hoon`, `desk/tests/lib/groups-json.hoon`), and a web e2e (`apps/tlon-web/e2e/pinned-reorder.spec.ts`) covering a Home reorder and persistence across reload.

## How did I test?

* Ran the shared unit tests (store reorder logic, including the in-flight sync race and empty-snapshot reconcile cases) and the app helper tests.
* Ran the backend desk tests on local fakezod (`groups-ui` and `groups-json` suites), plus a live agent round-trip confirming slot-preserving reorder, unknown-id rejection, and dedup.
* Ran the web e2e (`pinned-reorder.spec.ts`): pins three groups, drags the top pin below the bottom, asserts the optimistic order, then reloads and asserts persistence.
* Manually verified web drag-reorder in local dev, and confirmed drag-reorder and persistence on iOS (simulator) and Android, including a pinned group channel and the native sortable grid nested inside the `FlashList` header.

## Risks and impact

* Safe to rollback without consulting PR author? No
* Affects important code area:
  - [ ] Onboarding
  - [X] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [X] Other: pin ordering (groups-ui agent + pins DB queries)

This touches a backend agent action, pin DB queries, and chat-list rendering, so it is not a trivial UI-only change. The `reorderPinnedItems` store action concentrates the distributed-systems edge cases (cross-device drift, in-flight sync, rollback) and is the main area to watch.

## Rollback plan

Revert. Backend changes require no migrations.

## Screenshots / videos

https://github.com/user-attachments/assets/46a21981-4751-412c-8e91-166326ab5bf0


Fang- (commented):
Backend stuff here is essentially just list ordering algo stuff, if it works it works. Looks fine to me.

arthyn (approved):
lgtm, would be nice to get an emitted fact when this happens so you see it update live cross-device

PR #5905, opened 2026-06-09
tlon-skill: stop reporting false success on rejected group admin actions

patosullivan:
## Summary

Fixes TLON-5954: broken admin promotion (and the other state-changing admin commands) in the `tlon-skill` `groups` CLI, where actions against a group hosted on another ship printed `✅` on a local poke ack even when the remote host rejected the change. The acting ship now confirms its permission before mutating and verifies the resulting state by scry before reporting success.

## Changes

- In `packages/tlon-skill/scripts/groups.ts`, added a pre-flight permission check (`assertActingShipCanAdminister`) that reads the group's actual state and refuses when the acting ship is neither the host nor an admin, with an error naming the acting ship and host plus a non-zero exit. It retries with the same `VERIFY_ATTEMPTS`/`VERIFY_DELAY_MS` as the verification paths so a stale local snapshot on a foreign-hosted group doesn't reject a just-granted admin; the happy path still returns on the first attempt.
- In `packages/tlon-skill/scripts/groups.ts`, added batched post-poke scry-poll verification (`verifyShips`, `verifyShipsHaveRole`, `verifyShipsAreAdmin`, `verifyShipsRemoved`, `verifyShipsBanned`) that confirms the change landed before printing `✅`, using one scry per attempt against all target ships and throwing naming only the still-unverified ships on failure.
- Wired the pre-flight check and verification into the 7 affected commands (promote, demote, assign-role, remove-role, kick, ban, unban) and replaced the unconditional `✅` lines with post-verification output. `assign-role`/`remove-role` verify the **specific role** was gained/lost; `promote`/`demote` verify **true admin status** — that the seat holds (or no longer holds) any role the group marks as admin (`admins`), not the literal `admin` role id; `kick` verifies the seat was removed; and `ban`/`unban` verify the ban-list state.
- Hardened the admin mutations against partial or false success (from review feedback):
  - `demote` removes **every** admin-marked role (sourced from `admins`, not just the literal `admin`) in a **single atomic `del-roles` seat poke**, so a self-inclusive demotion can't strip the caller's own admin role part-way through and get the remaining removals rejected — the backend checks `se-src-is-admin` once per action.
  - `demote` refuses to demote the **group host**, which is an admin unconditionally (`se-is-admin`), instead of stripping roles and reporting a no-op as success.
  - `promote` refuses an unmarked `admin`-role **collision**: if a group already has a non-admin role whose id is `admin`, `ensureAdminRole` throws rather than marking it admin, since marking it would grant admin to every existing holder of that role, not just the promote target.
- Added `packages/tlon-skill/scripts/commands/groups-verification.ts` with pure, side-effect-free predicates (`actingShipCanAdminister`, `seatHasRole`, `seatHasAdminRole`, `shipIsSeated`, `shipIsBanned`, `getShipRecordValue`) and the `RawGroupForAdminVerification` type, extended with `admissions.banned.ships`. The module avoids importing `./api-client` or `@tloncorp/api` so it stays unit-testable and passes the existing `command-contract.test.ts` forbidden-globals scan.
- Added `packages/tlon-skill/scripts/commands/groups-verification.test.ts` with bun:test coverage for host/admin/custom-admin-role/non-admin/absent ships, normalized-key lookups, role/seat/ban present-absent checks, partial multi-target failures that name only the unverified ship, and the admin-role-collision and partial-demote (still-admin) predicates.
- Moved `getShipRecordValue` and `RawGroupForAdminVerification` into the new module (`getShipRecordValue` now takes a `normalizeShip` argument) and updated the 5 `create-owned` callers accordingly.

## How did I test?

- `bun test ./scripts` (103 pass), including the new `groups-verification.test.ts` unit coverage.
- `bun test ./tests/hermetic` (135 pass).
- `pnpm typecheck` across src and test (clean).
- Reused the verification pattern already proven by `create-owned`'s `verifyOwnerAdmin`, which reads `/v2/ui/groups/{id}` rather than trusting the poke ack.

## Risks and impact

  - [x] Other: `tlon-skill`

Notes for reviewers:
- The pre-flight check and verification add scry round-trips to these commands. The happy path stays fast (the pre-flight returns on its first attempt when the ship is already a synced admin/host); only a rejection or a lagging foreign-host snapshot waits out the retries (worst case about 2.5s, reusing `VERIFY_ATTEMPTS=5` / `VERIFY_DELAY_MS=500`).
- Admin permission and `promote`/`demote` verification are checked against the group's raw `admins` role-id list rather than the literal `admin` role, so custom admin roles work.
- ban/unban verification depends on `admissions.banned.ships` being present in the `/v2/ui/groups/{id}` response (confirmed against the `Admissions` interface in `packages/api/src/urbit/groups.ts`).
- Intentionally does not introduce `trackedPoke`: scry-polling the resulting state is the source of truth, which is correct for the foreign-host case where the host emits no rejection fact.
- No CLI syntax changes, no `@tloncorp/api` changes. Failure is signaled purely via a non-zero exit. The change stays out of the in-progress TLON-5771 `run(args, deps)` runner migration while extracting the new module in a way compatible with it.

## Rollback plan

Revert

## Out of scope / follow-ups
- `tlonbot` prompt-layer mitigations

### DIFF
--- a/packages/tlon-skill/scripts/commands/groups-verification.test.ts
+++ b/packages/tlon-skill/scripts/commands/groups-verification.test.ts
@@ -1,0 +1,176 @@
+import { describe, expect, it } from 'bun:test';
+
+import {
+  type RawGroupForAdminVerification,
+  actingShipCanAdminister,
+  getShipRecordValue,
+  seatHasRole,
+  shipIsBanned,
+  shipIsSeated,
+} from './groups-verification';
+
+// Minimal stand-in for the real `preSig`-based normalizer: ensure a leading sig.
+const normalizeShip = (ship: string): string =>
+  ship.startsWith('~') ? ship : `~${ship}`;
+
+const HOST = '~ravmel-ropdyl';
+const ADMIN = '~admin-ship';
+const MEMBER = '~lorhep-binfen';
+
+describe('getShipRecordValue', () => {
+  it('returns undefined for a missing record', () => {
+    expect(getShipRecordValue(undefined, MEMBER, normalizeShip)).toBeUndefined();
+  });
+
+  it('matches on the raw key directly', () => {
+    const record = { [MEMBER]: { roles: ['admin'] } };
+    expect(getShipRecordValue(record, MEMBER, normalizeShip)).toEqual({
+      roles: ['admin'],
+    });
+  });
+
+  it('falls back to a normalized comparison for unnormalized keys', () => {
+    // record key has the sig, lookup ship does not
+    const record = { '~zod': 7 };
+    expect(getShipRecordValue(record, 'zod', normalizeShip)).toBe(7);
+  });
+
+  it('returns undefined when the ship is absent', () => {
+    expect(
+      getShipRecordValue({ [HOST]: 1 }, MEMBER, normalizeShip)
+    ).toBeUndefined();
+  });
+});
+
+describe('seatHasRole', () => {
+  const group: RawGroupForAdminVerification = {
+    seats: { [MEMBER]: { roles: ['admin', 'editor'] } },
+  };
+
+  it('is true when the seat holds the role', () => {
+    expect(seatHasRole(group, MEMBER, 'admin', normalizeShip)).toBe(true);
+  });
+
+  it('is false when the seat lacks the role', () => {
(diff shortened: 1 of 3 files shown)


arthyn (approved):
lgtm

PR #5907, opened 2026-06-09
Remove expo-av and the unused features that depended on it

janicduplessis:
## Summary

`expo-av` is deprecated and is removed in upcoming Expo SDKs. Auditing its last consumers showed they were all part of unused or unreachable features — the experimental custom-channel mic/audio post type and an orphaned embed-rendering system — so instead of migrating them to `expo-audio`, this PR removes them and drops the dependency. Voice memos, the one production audio feature, were already on `expo-audio` and are untouched.

Extracted from #5886 so the dep removal can land independently of the SDK 56 upgrade.

## Changes

- **Remove the `mic` draft input / `audio` post type** (confirmed unused by the team): `MicInput`, `AudioPost`, their registrations in `ComponentsKitProvider`, and the `tlon.r0.input.mic` / `tlon.r0.content.audio` registry entries. Channels that still carry that config degrade gracefully — unknown renderer ids fall back to the channel-type default (`PostView`), unknown input ids render no composer (`DraftInputView`). These are channel-level config ids, not message types: posts created through `MicInput` are plain chat posts with an audio-file attachment, so no message on the network references anything removed here.
- **Remove the `customChannelCreation` feature flag and the channel-configuration UI** it gated: the "Configure view" sheet action, `ChannelConfigurationBar`/`ConfigInput` in `CreateChannelSheet`, and the `Channel` ref plumbing (`forwardRef`/`useImperativeHandle`) that existed only to open the bar. The action was already non-functional — `ChatOptionsSheet` never forwarded `onPressConfigureChannel` past its top-level component, so the button did nothing from anywhere. Flags persisted in local storage that are no longer in `featureMeta` are tolerated on load.
- **Remove the orphaned embed-rendering system**: chat links stopped rendering through `EmbedContent` when they switched to preview cards (af5e9b96f, ~5 months ago), leaving `EmbedContent`, `AudioEmbed`, `AudioPlayer` (+`.web`), `Embed`, `EmbedWebView` (+`.web`), the YouTube/Spotify/Twitter/TikTok `providers` HTML generators, and the `useEmbed` hook reachable only from Cosmos fixtures. `fetchEmbed`/`isTrustedEmbed` in `@tloncorp/shared` stay — the link-preview metadata path (`metagrabActions`) still uses them for oEmbed enrichment.
- **Remove the write-orphaned remainder of the custom-channel registry.** Deleting the configuration UI left no code path that can assign these ids to a channel, so the components behind them are unreachable for any new data: collection renderers `carousel`/`cards`/`sign`/`boardroom`/`strobe`/`summaries` (all of `postCollectionViews/` except the standard `ListPostCollection`), content renderers `color`/`raw`/`yell`/`scratchpad`, and draft inputs `yo`/`color`. The `summaries` view was also the only navigator to `PostUsingContentConfigurationScreen`, so that screen and `DetailPostUsingContentConfiguration` go with it. The `notes` ids stay — notes channels are a live feature with their own creation path (`createNotesChannel`). Legacy channels carrying removed ids degrade the same way as the mic/audio ones.
- **Rename `VideoEmbed` → `VideoPreview`** and flatten the now single-file `Embed/` directory. The component renders an uploaded video attachment (poster, play badge, duration) and opens the media viewer — it isn't an embed; `BlockRenderer`'s `VideoBlock` keeps wrapping it.
- **Drop `expo-av`** from `apps/tlon-mobile` and `packages/ui` package.json plus the pnpm, CocoaPods, and gradle lockfiles (the Android lockfile loses ExoPlayer 2.18 and the guava transitives that only `expo-av` was pulling).

## How did I test?

- Typechecked `api`, `app`, `shared`, `tlon-mobile`, and `tlon-web`; `packages/api` tests pass (192/192) after the registry-entry removal and `packages/shared` logic tests pass (108/108) after the `useEmbed` removal; eslint on touched files shows only pre-existing warnings.
- App builds and runs with `expo-av` removed on **iOS** (iPhone 17 Pro simulator / iOS 26.2 and physical iPhone 17 Pro / iOS 26.5) and **Android** (Medium_Phone_API_35 emulator, productionDebug). Voice memo record/playback verified unaffected.
- The removed surfaces had no reachable entry points to regression-test: link embeds stopped rendering through `EmbedContent` five months ago, and "Configure view" was dead due to the unforwarded prop.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: removal of unused/unreachable custom-channel and embed code; no behavior change to reachable surfaces

Worst case: a channel somewhere still has a removed content config (`mic`/`audio` or one of the experimental renderers above) (required the Tlon-employee flag plus the broken Configure view, so realistically only internal test channels) — its old posts render via the default chat renderer as attachment blocks instead of an audio player, and it shows no composer until reconfigured. Voice memos, video blocks, and link previews are untouched.

## Rollback plan

Revert the PR

## Screenshots / videos

N/A — no UI changes to reachable surfaces.


dnbrwstr:
I may have missed some discussion on this -- my initial thought is that I'd prefer to remove just the pathways that require `expo-av` while leaving in place the custom channel plumbing under the hood. How does that strike you?

janicduplessis:
@dnbrwstr It was discussed here /1/chan/chat/~bitpyx-dildus/interface/msg/170141184508006854228615737489947099136

the idea is if we ever want to bring back the feature we can always dig it up in git history.

dnbrwstr (approved):
Discussed offline -- lgtm!

PR #5920, opened 2026-06-10
tlon-skill: Port posts react command runner to tlon-skill and restore build-smoke CI gate

patosullivan:
## Summary

Re-integrates work into the in-monorepo `packages/tlon-skill` CLI as PR 1 of 3 stacked PRs for the current phase of TLON-5771. The monorepo import predated standalone PR #93 (the `posts react` migration) and dropped the built-binary CI gate, so this ports the migration and restores the gate. This is a port of already-reviewed code (standalone PR: [tloncorp/tlon-skill#93](https://github.com/tloncorp/tlon-skill/pull/93); diff-of-record is commit [`253a73d`](https://github.com/tloncorp/tlon-skill/commit/253a73d35feb80b7d02cc137bc347693e18cb9df)), reformatted with monorepo prettier; logic is unchanged.

## Changes

- Added `packages/tlon-skill/scripts/commands/posts.ts`, a pure `posts react` runner (`run(args, deps): Promise<number>`) with self-contained `extractNumericId`, `formatUd`, and `formatPostId` helpers. These are deliberately not imported from the legacy `scripts/posts.ts` (which auto-runs on import). Help short-circuits before auth, and missing args throw usage errors before auth.
- Added `packages/tlon-skill/scripts/posts-runtime.ts`, the process-backed deps. `authenticate` calls `ensureClient(['channels'])`, the current user is read via `@tloncorp/api`, and `addReaction` failures are rethrown as `commandError`.
- Added `packages/tlon-skill/scripts/commands/posts.test.ts`, in-process tests with fake deps covering the help short-circuit, the missing-arg matrix before auth, post-id formatting (dotted, undotted, `~ship/`-prefixed), the exact `✓ Reaction added` output, CommandError formatting, and unexpected-error propagation. Includes one addition over the standalone file: a `run(['unreact'], deps)` test pinning that non-`react` subcommands return a usage error with no auth or API calls (the runner becomes the posts-family dispatcher in a later PR).
- Updated `scripts/main.ts` so that in `case 'posts'`, `posts react ...` routes through the migrated runner (without mutating `process.argv`). All other `posts` paths still fall through to the legacy import-side-effect module.
- Updated `scripts/cli-test-matrix.ts` with matrix cases (`posts react --help`, the same with global `--url`/`--cookie` flags to prove they are stripped before dispatch, `posts unreact --help` to prove non-react paths stay on the legacy dispatch, and an auth-required `posts react chat/~host/channel 170.141 👍` case). Also centralized the hostile-help command list into a new exported `HOSTILE_HELP_COMMANDS` and added a nested `posts react --help` entry.
- Updated `tests/hermetic/cli.test.ts` and `scripts/build-smoke.ts` to consume the shared `HOSTILE_HELP_COMMANDS` instead of each rebuilding the list, removing duplication and adding nested-subcommand help coverage to both the source hermetic suite and the built-binary suite.
- Updated `.github/workflows/ci.yml` with a `tlon-skill build smoke` step (`pnpm --filter '@tloncorp/tlon-skill' build:smoke`) after "Run Tests" and before "Build packages". This restores the built-binary golden matrix gate dropped in the monorepo move; it compiles the binary via bun `--compile` and runs the full golden matrix under a hermetic env. Bun 1.3.4 is already set up earlier in the job.

Decisions worth noting:

- The coverage artifact from the standalone `ci.yml` was not ported. The suite is subprocess-heavy, so Bun coverage mostly measures the test process; the plan treats coverage as reporting-only. `test:coverage` remains in `package.json` for local use.
- The root `test:ci` trailing-arg quirk (`pnpm run -r test run` appends `run` to tlon-skill's chained `test` script, producing `bun test ./tests/hermetic run` where `run` acts as a bun path filter) was verified harmless (all tests still execute). Documenting it here rather than changing the shared root script, since other packages rely on the `run` arg for vitest.

## How did I test?

- `pnpm --filter '@tloncorp/tlon-skill' check` (typecheck plus 93 unit tests, 141 hermetic tests, build smoke), exit 0.
- `pnpm --filter '@tloncorp/tlon-skill' test run` passes 234 tests, reproducing the root `test:ci` trailing-arg shape.
- `pnpm prettier --check` on all touched files, clean.
- Built golden matrix confirms the new cases pass, including `posts react --help` under hostile `TLON_CONFIG_FILE=/nonexistent` and `--config /nonexistent` (exit 0, prints usage, empty stderr, no credential lookup).

## Risks and impact

- Affects important code area: skill

## Rollback plan

Revert

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,9 @@ jobs:
       - name: Run Tests
         run: pnpm test:ci
 
+      - name: tlon-skill build smoke
+        run: pnpm --filter '@tloncorp/tlon-skill' build:smoke
+
       - name: Build packages
         run: pnpm build:all
         env:
--- a/packages/tlon-skill/scripts/build-smoke.ts
+++ b/packages/tlon-skill/scripts/build-smoke.ts
@@ -5,8 +5,8 @@ import { join, resolve } from 'node:path';
 
 import {
   CLI_MATRIX_CASES,
-  COMMAND_FAMILIES,
   type CliCase,
+  HOSTILE_HELP_COMMANDS,
   normalizeCliOutput,
 } from './cli-test-matrix';
 
@@ -155,15 +155,7 @@ for (const testCase of CLI_MATRIX_CASES) {
   assertCase(testCase);
 }
 
-const hostileHelpCommands = [
-  { name: 'top-level', args: ['--help'] },
-  ...COMMAND_FAMILIES.map((family) => ({
-    name: family,
-    args: [family, '--help'],
-  })),
-];
-
-for (const command of hostileHelpCommands) {
+for (const command of HOSTILE_HELP_COMMANDS) {
   assertCase(
     {
       name: `${command.name} help with nonexistent TLON_CONFIG_FILE`,
--- a/packages/tlon-skill/scripts/cli-test-matrix.ts
+++ b/packages/tlon-skill/scripts/cli-test-matrix.ts
@@ -319,9 +319,39 @@ export const NESTED_HELP_CASES: CliCase[] = [
     ['groups', 'info', '--help'],
     'Usage: tlon groups info'
   ),
+  helpCase(
+    'posts react --help',
+    ['posts', 'react', '--help'],
+    'Usage: tlon posts react'
+  ),
+  helpCase(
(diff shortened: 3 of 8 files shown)


arthyn (approved):
lgtm

PR #5926, opened 2026-06-11
Add %context-lens agent for durable owner-side bot run history

wca4a:
## Summary

- Adds the `%context-lens` gall agent (split out of #5925 so the ship-side store can land and be vetted independently of the client UI). One agent, two roles:
  - **Bot ship**: accepts `%context-lens-action-1` pokes from the local OpenClaw gateway and fans run records out to configured owners as `%context-lens-signal-1` pokes (ames retries until ack).
  - **Owner ship**: stores runs keyed `[bot id-run]` with `src.bowl` as the trust anchor, gives `%context-lens-update-1` facts on `/v1`, and answers `/x/recent` and `/x/run` scries with per-bot retention (1000 runs / 90 days), enforced on store, on a daily behn timer, and filtered at read time.
- Run payloads relay as cords (serialized JSON) rather than `$json` to keep the agent schema-agnostic.
- Named `%context-lens` rather than `%lens` because `%base` already bills `%lens` and gall agent names are global across desks — the collision makes the groups desk commit silently revert.
- Also fixes arm64/aarch64 vere arch detection in `backend/run-tests.sh` so the desk test suite runs on Linux arm64.

## Dependencies

This is the ship-side hub of a three-part stack — it has no code dependencies on the other PRs, but they depend on it:

- **Upstream producer**: tloncorp/openclaw-tlon#161 — the gateway pokes this agent on the bot ship with finalized runs. Without this agent installed, those pokes nack harmlessly (the gateway logs and continues).
- **Downstream consumer**: #5925 — the client UI subscribes to `/v1`, scries `/x/recent` / `/x/run`, and pokes `%configure` via `lensApi`. Without this agent, the client's lens subscription no-ops gracefully and the UI shows no run history.
- **Suggested landing order**: this PR → #5925 → openclaw-tlon#161 (any order is safe, but this order means no consumer ever points at a missing agent).

## Test plan

- [x] `desk/tests/app/context-lens.hoon` covers configure, run-final fan-out, owner-side store/retention, and scries (`backend/run-tests.sh`)
- [x] Exercised end-to-end against dev ship `~malmur-halmex` with the gateway and client PRs (run records fan out, facts drive live client updates)

### DIFF
--- a/desk/app/context-lens.hoon
+++ b/desk/app/context-lens.hoon
@@ -145,7 +145,7 @@
   $(targets t.targets)
 ::
 ++  store
-  |=  [bot=ship =id-run:l payload=json complete=?]
+  |=  [bot=ship =id-run:l payload=@t complete=?]
   ^+  cor
   ::  never demote a finalized run back to partial
   ::
--- a/desk/app/context-lens.hoon
+++ b/desk/app/context-lens.hoon
@@ -16,7 +16,7 @@
 /+  default-agent, verb, dbug
 |%
 +$  card  card:agent:gall
-::  $state-0: agent state
+::  $state-1: agent state (%0 -> %1 arms the prune timer on upgrade)
 ::
 ::    .owners: ships that receive run fan-out (bot role; ~ = inert)
 ::    .runs: stored run records keyed by bot ship and run id (owner role)
@@ -27,14 +27,21 @@
       runs=(map [bot=ship =id-run:l] run:l)
   ==
 ::
-+$  versioned-state  $%(state-0)
-+$  current-state  state-0
++$  state-1
+  $:  %1
+      owners=(set ship)
+      runs=(map [bot=ship =id-run:l] run:l)
+  ==
+::
++$  versioned-state  $%(state-0 state-1)
++$  current-state  state-1
 ::
 ++  migrate-state
   |=  old=versioned-state
   ^-  current-state
   ?-  -.old
-    %0  old
+    %0  [%1 +.old]
+    %1  old
   ==
 --
 =|  current-state
@@ -47,12 +54,20 @@
   +*  this  .
       def   ~(. (default-agent this %.n) bowl)
       cor   ~(. +> [bowl ~])
-  ++  on-init  `this
+  ++  on-init
+    ^-  (quip card _this)
+    [~[wait-prune:cor] this]
   ++  on-save  !>(state)
+  ::  arm the prune timer exactly once when upgrading from a version
+  ::  that didn't have it; arming on every load would stack timers
+  ::
   ++  on-load
     |=  ole=vase
     ^-  (quip card _this)
-    [~ this(state (migrate-state !<(versioned-state ole)))]
+    =/  old  !<(versioned-state ole)
+    =/  cards=(list card)
+      ?:(?=(%0 -.old) ~[wait-prune:cor] ~)
+    [cards this(state (migrate-state old))]
   ++  on-poke
     |=  [=mark =vase]
     ^-  (quip card _this)
@@ -77,6 +92,7 @@
       =/  bot  (slav %p i.t.t.path)
       =/  =id-run:l  i.t.t.t.path
       ?~  r=(~(get by runs) [bot id-run])  [~ ~]
+      ?:  (expired:cor u.r)  [~ ~]
       ``context-lens-update-1+!>(`update:v1:l`[%run bot id-run u.r])
     ==
   ++  on-agent
@@ -84,7 +100,11 @@
     ^-  (quip card _this)
     =^  cards  state  abet:(agent:cor wire sign)
     [cards this]
-  ++  on-arvo  on-arvo:def
+  ++  on-arvo
+    |=  [=wire =sign-arvo]
+    ^-  (quip card _this)
+    =^  cards  state  abet:(arvo:cor wire sign-arvo)
+    [cards this]
   ++  on-leave  |=(path `this)
   ++  on-fail
     |=  [=term =tang]
@@ -97,10 +117,22 @@
 ++  abet  [(flop cards) state]
 ++  emit  |=(=card cor(cards [card cards]))
 ++  give  |=(=gift:agent:gall (emit %give gift))
-::  retention bounds for stored runs (owner role)
+::  retention bounds for stored runs (owner role): enforced on store
+::  and on a daily timer; reads also filter expired runs so the age
+::  bound holds even if a bot goes quiet between timer fires
 ::
-++  max-runs-per-bot  500
-++  max-run-age  ~d30
+++  max-runs-per-bot  1.000
+++  max-run-age  ~d90
+++  prune-interval  ~d1
+::
+++  wait-prune
+  ^-  card
+  [%pass /prune %arvo %b %wait (add now.bowl prune-interval)]
+::
+++  expired
+  |=  =run:l
+  ^-  ?
+  (lth received.run (sub now.bowl max-run-age))
 ::
 ++  poke
   |=  [=mark =vase]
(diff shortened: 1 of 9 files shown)


desk/sur/context-lens.hoon
```diff
@@ -0,0 +1,44 @@
...
+::  $id-run: gateway-assigned run identifier (the lensId stamped on posts)
+::
++$  id-run  @t
+::  $run: a stored run record
+::
+::    payload is opaque serialized JSON with an inner schemaVersion;
+::    ships relay and store it without interpreting its contents.
+::    it is a cord rather than a $json because embedding the recursive
+::    $json type in mark sample types makes ford's tube nest checks
+::    diverge (stack overflow on every json conversion build).
```
Fang-: Sounds like it wrote and subsequently misunderstood an infinite loop? `@t` instead of `$json` for opaque client-only payloads is standard for us now though, so all good.

desk/app/context-lens.hoon
```diff
@@ -0,0 +1,272 @@
...
+  ==
+::
+++  arvo
+  |=  [=wire sign=sign-arvo]
+  ^+  cor
+  ?+  wire  cor
+      [%prune ~]
+    ?.  ?=([%behn %wake *] sign)  cor
+    =?  cor  ?=(^ error.sign)
+      ((slog 'context-lens: prune wake failed' u.error.sign) cor)
```
Fang-: In practice agents don't get `%wake` error reporting when they crash during their processing, they'll have their `+on-fail` called by gall instead. (The event as a whole never crashes.)

Until `+on-fail` gets a better api (gall 2026 fixes this!) there's nothing we can reliably do, timer error handling wise.

desk/app/context-lens.hoon
```diff
@@ -0,0 +1,272 @@
...
++$  card  card:agent:gall
+::  $state-1: agent state (%0 -> %1 arms the prune timer on upgrade)
+::
+::    .owners: ships that receive run fan-out (bot role; ~ = inert)
+::    .runs: stored run records keyed by bot ship and run id (owner role)
+::
++$  state-0
+  $:  %0
+      owners=(set ship)
+      runs=(map [bot=ship =id-run:l] run:l)
```
Fang-: Crazy structure. Per the processing in `+prune` and co (which currently go through all kinds of hoops and inefficient logic to re-group the runs), we probably want something more in the direction of `(mip ship id-run:l run:l))`, or perhaps just `(jar ship [id-run:l run:l])` if the lists we keep won't get too big.

desk/app/context-lens.hoon
```diff
@@ -0,0 +1,272 @@
...
+        our.bowl
+        id-run.signal
+        payload.signal
+        ?=(%run-final -.signal)
+      ==
+    %-  emit
+    :^    %pass
+        /signal/(scot %p i.targets)/(scot %t id-run.signal)
+      %agent
+    [[i.targets %context-lens] %poke %context-lens-signal-1 !>(signal)]
```
Fang-: Given that we don't do anything on-nack, we might want to remove the details from the wire and just do `/signal`. The `i.targets` comes back in the `src.bowl` anyway, and all adding the `id-run.signal` into the wire does is cause ames flow proliferation, losing us any delivery ordering guarantees while increasing network state bloat.

desk/sur/context-lens.hoon
```diff
@@ -0,0 +1,44 @@
...
+  $%  [%configure owners=(set ship)]
+      [%run-event =id-run payload=@t]
+      [%run-final =id-run payload=@t]
+  ==
+::  $signal: bot-ship to owner-ship fan-out pokes
+::
++$  signal
+  $%  [%run-event =id-run payload=@t]
+      [%run-final =id-run payload=@t]
+  ==
```
Fang-: Given that these get processed almost identically, this could just be `[=id-run payload=@t final=?]`, no `$%` needed.

(And likewise for the corresponding action, unify into a single `%run-event` case.)

desk/app/context-lens.hoon
```diff
@@ -0,0 +1,272 @@
...
+        [%x %recent ~]
+      ``context-lens-update-1+!>(`update:v1:l`[%runs recent:cor])
+    ::
+        [%x %run @ @ ~]
+      =/  bot  (slav %p i.t.t.path)
+      =/  =id-run:l  i.t.t.t.path
+      ?~  r=(~(get by runs) [bot id-run])  [~ ~]
+      ?:  (expired:cor u.r)  [~ ~]
+      ``context-lens-update-1+!>(`update:v1:l`[%run bot id-run u.r])
+    ==
```
Fang-: I feel like it's slightly unusual to produce `$update`s in scry results, but perhaps not unheard of.

desk/app/context-lens.hoon
```diff
@@ -0,0 +1,272 @@
...
+    ^-  (quip card _this)
+    =^  cards  state  abet:(poke:cor mark vase)
+    [cards this]
+  ++  on-watch
+    |=  =path
+    ^-  (quip card _this)
+    ?>  =(src our):bowl
+    ?+  path  (on-watch:def path)
+      [%v1 ~]  `this
+    ==
```
Fang-: We don't give a `%runs` update here because we expect the client to have scried that out already?

Fang- (commented):
Some initial thoughts. Primarily, the state data structure can be better.

wca4a:
Thanks @Fang- — agree with almost all of this. Here's the plan, grouped by what I'll change vs. confirm.

### Structural changes (agent-local, this PR)

- **State → `(mip ship id-run run)`.** You're right that `+prune` (runs on every store), `+prune-all`, and `+recent` all currently re-derive the per-bot grouping by tapping the whole map and skimming on `=(bot for)`. A `mip` makes a bot's runs `(~(get by runs) bot)` and keeps `+store`'s keyed upsert/dedup O(1) via `bi`. I'll go with `mip` rather than `jar` specifically because the `jar`'s list would turn our dedup-by-`id-run` (and the late-partial-drop guard in `+store`) into a list scan + rebuild — the keyed inner map preserves the semantics we rely on.
- **Flatten the poke wires** (`/signal/<ship>/<id-run>` → `/signal`, and `/command/<ship>/<id-run>` → `/command`). We only `slog` on nack, so the wire carries no actionable info; encoding `id-run` just spawns a flow per run, bloating network state and losing ordering. Flattening collapses to one ordered flow per target (recoverable from `src.bowl`), and `+store`'s defensive reorder handling means we lose no safety. I'll update the wire assertions in the agent tests to match.

### Comment / cleanup

- **`@t` over `$json`:** keeping it — opaque passthrough is the right call. I'll just trim the comment; the "ford tube nest checks diverge / stack overflow" rationale is a misdiagnosis and shouldn't be the documented reason.
- **`%wake` error slog:** removing the `=? cor ?=(^ error.sign)` branch. Agreed it implies timer-crash reporting gall doesn't actually deliver (that's `+on-fail`'s job, weak as it is today) — keeping it is misleading.

### Confirming your reads (no change)

- **Updates in scry results:** intentional — reusing the `update-1` mark for both scries and `/v1` facts lets the client share one decoder. Happy to split to a dedicated scry mark if you'd prefer.
- **No `%runs` on `[%v1 ~]` watch:** correct — the client scries `/x/recent` for backfill on mount, then subscribes for live `%run` facts; emitting `%runs` on watch would double-deliver.

### Coordinated change (separate, not agent-local)

- **Unify `%run-event`/`%run-final` → `[=id-run payload=@t final=?]`.** Agreed they're processed identically. Holding this one because it's cross-repo: the openclaw gateway emits two distinct json fronds when poking, so the agent mark and the gateway poke builder have to change in lockstep. I'll land it alongside the gateway change rather than half-migrating the protocol here.

arthyn:
closing in favor of #5971

PR #5928, opened 2026-06-12
Multi-bot thinking indicator for chat channels

wca4a:
## Summary
- Reworks `useConversationComputingState` to return per-ship computing states with a combined label ("X and Y are thinking...", "N bots are thinking..."), sorted by ship so heartbeat re-publishes don't reorder.
- Rewrites `ThinkingState` as an animated avatar stack (max 3 avatars + "+N" overflow) and enables it in group chat channels (`dm` + `chat`), gated on read permission.
- Shows the thinking indicator in thread views (`DetailView`), above the reply composer.
- Fixes two `%presence` agent bugs:
  - **Host blind spot**: a `%set` with empty `disclose` (= public) was fanned out to subscribers but never stored on the context host itself, so the host's own client never saw presence for channels it hosts.
  - **Bounded nack retry**: a nacked context watch previously dropped the desire permanently (one transient nack = no presence until next full setup). Now retries with linear backoff (try × 5min) via a new `tries` map in `%1` state, giving up after 5 consecutive nacks so a permanently-failing host doesn't become an eternal timer.

## Known limitation: thread indicator is channel-scoped
Computing presence contexts are keyed by channel/DM only (`/channel/chat/~host/name`), so the indicator in a thread reflects bots thinking **anywhere in that channel**, not specifically in that thread. The ideal end state is thread-scoped presence contexts (e.g. `/channel/chat/~host/name/thread/<parent-post-id>`) published by the gateway, with the channel view aggregating thread contexts by prefix. That requires changes across the presence agent, the gateway publisher, and the client presence API, so it's deferred to a follow-up.

## Test plan
- [x] `tsc --noEmit` clean on packages/app
- [x] Hoon build check on a live ship (incl. `%0` → `%1` state migration)
- [x] iOS sim, live ships: DM single-bot indicator; group channel two-ship avatar stack + combined label; remote-channel presence mirror populates during a run and clears at run end
- [x] 3-ship and 4-ship (+1 overflow) states verified via mocked presence
- [x] Thread view indicator verified via mocked presence (iOS sim)
- [x] Gallery/notebook channels show no indicator
- [ ] Desktop web spot-check
- [ ] Android pass

### DIFF
--- a/desk/app/presence.hoon
+++ b/desk/app/presence.hoon
@@ -42,14 +42,24 @@
 +$  state-0
   $:  %0
       =places
+      want=(set [ship context])
+      subs=(jug context ship)
+  ==
++$  state-1
+  $:  %1
+      =places
       :: ships=(jug ship context)
     ::
-      want=(set [ship context])  ::  desired outgoing subs
-      subs=(jug context ship)    ::  real incoming subs
-      ::TODO  wack=(set ship)    ::  waiting for acks (load prevention)
+      want=(set [ship context])   ::  desired outgoing subs
+      subs=(jug context ship)     ::  real incoming subs
+      tries=(map [ship context] @ud)  ::  consecutive sub nacks
+      ::TODO  wack=(set ship)     ::  waiting for acks (load prevention)
   ==
++$  versioned-state  $%(state-0 state-1)
 ::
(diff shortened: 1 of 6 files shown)


arthyn (approved):
lgtm

PR #5936, opened 2026-06-15
openclaw: context lens — durable run history, ship sync, run-capture hardening 2/5

patosullivan:
## Summary
Streams bot-run introspection events (trigger, context sources, tool calls, writes, timeline) from the gateway, models runs beyond replies (cron/background/internal), and hardens tool lifecycle tracking, event delivery, and run-queue handling.

- **Durable run history**: finalized runs persist to a disk store and mirror to the bot ship's `%context-lens` agent (payloads as serialized-JSON cords), fanning out to owner ships — no gateway reachability required from devices.
- **Config & auth**: zod config schema, bearer-token routes, effective-enablement gating, `botShip` stamping in the reference blob.
- **Full tool arguments** as expandable detail (`argumentDetail`) for the client inspector.

## Stacking / migration
- Migrated from **tloncorp/openclaw-tlon#161** (authored by ~wsa). Rebased onto current openclaw-tlon `master`, relocated to `packages/openclaw`, squashed.
- **Based on #5935** (the #160 presence-fixes sync) — retarget to `develop` once that merges.
- This is the gateway/**producer** end of the context-lens stack; the client (#5925) and `%context-lens` agent (#5926) consume what it produces.

## Security note
`argumentDetail` ships **raw tool parameters** in run payloads synced to the ship. Run visibility is **owner-only** today; these payloads must gain redaction/filtering **before** visibility is ever widened (secrets/tokens/file contents can appear in tool args).

## Test plan
- [x] `pnpm tsc --noEmit` clean
- [x] Unit tests: 640/640 passing (`pnpm test` in `packages/openclaw`)
- [x] Applies cleanly on the #160 baseline (only the pre-existing a2ui fragment in `media.ts` needed reconciliation — #161's positive allowlist supersedes it and still skips a2ui)


packages/openclaw/src/types.ts
```diff
@@ -203,11 +316,33 @@ export function resolveTlonAccount(
...
   };
 }
 
+/**
+ * Context lens is effectively on only when enabled AND at least one reader
+ * path exists: an auth token (gateway HTTP/SSE routes) or owner ships
+ * (%context-lens ship sync). Without either, recording and blob stamping would
+ * produce data nothing can read.
+ */
+export function isContextLensEnabled(
```
arthyn, line 332: `isContextLensEnabled()` is dead code — grep finds zero callers across the monorepo. It also duplicates `isContextLensEffectivelyEnabled()` in `context-lens-ship-sync.ts:131`, which is the variant actually used (by `monitor/index.ts`). Both encode the same predicate (`enabled && (authToken || owner || ownerShip)`) via slightly different paths, so they can silently drift. Per CLAUDE.md ("Reuse before adding — don't ship parallel implementations"), suggest deleting this one (or having one delegate to the other) so there's a single source of truth.

packages/openclaw/src/context-lens-store.ts
```diff
@@ -0,0 +1,228 @@
...
+import type { ContextLens, ContextLensStatus } from './context-lens.js';
+import { sharedSlot } from './shared-state.js';
+import { resolveTlonAccount } from './types.js';
+
+export const DEFAULT_STORE_RETAIN_DAYS = 30;
+export const DEFAULT_STORE_MAX_STORED = 500;
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+
+const TERMINAL_STATUSES: ReadonlySet<ContextLensStatus> = new Set([
```
arthyn, line 16: `TERMINAL_STATUSES` is defined byte-identically here and in `context-lens-ship-sync.ts:21`, both derived from `ContextLensStatus` (which lives in `context-lens.ts`). Adding or renaming a terminal status means editing both Sets — miss one and the disk store and ship-sync disagree on what counts as "final" (a run persisted by one reader but dropped as non-terminal by the other). Suggest exporting `TERMINAL_STATUSES` once from `context-lens.ts` and importing it in both places.

packages/openclaw/src/monitor/index.ts
```diff
@@ -471,6 +496,31 @@ export async function monitorTlonProvider(
...
+      opts.accountId ?? undefined
+    );
+    const contextLenses = createContextLensRegistry({
+      ttlMs: contextLensConfig.ttlMs ?? undefined,
+      maxEntries: contextLensConfig.maxEntries ?? undefined,
+      visibilityDefault: contextLensConfig.visibilityDefault,
+      disabled: !contextLensEnabled,
+    });
+    setContextLensEventCapacity(contextLensConfig.maxEntries);
+    const logContextLens = (
```
arthyn, line 585: `logContextLens` eagerly builds `JSON.stringify({ phase, detail, ...snapshot })` — spreading the entire lens snapshot (which can carry hundreds of tool runs + source previews) — on every lifecycle event (created/dispatching/model_selected/final, 5+ per run). The `?.` short-circuits when `runtime.log` is undefined, but if `runtime.log` is a level-filtering logger the full serialization still runs and the result is then discarded. Suggest gating the stringify behind a debug check or passing a lazy formatter so suppressed logs cost nothing. (The `contextLenses.get()` clone is fine — it feeds `publishContextLensEvent`.)

arthyn (commented):
A few cleanup items from a review pass — all in the context-lens producer code. Non-blocking, but worth tidying before this lands since the duplicated `TERMINAL_STATUSES` can cause the two reader paths to silently disagree.

arthyn (approved):
lgtm, I'm not sure we really need the local store and gateway routes, but lets leave it for now

PR #5944, opened 2026-06-16
ChatDetails, ChannelDetails: add channel host info block

jamesacklin:
## Summary

Adds a channel host block to the `ChannelDetails` version of `ChatDetails`.

## Changes

- Adds a ChannelHost padded block, which shows the current channel's host in its "Channel Info" (profile) screen
- Navigates the user to the channel host's profile on tap
- Imports the `getChannelHost` util, returns null if we're looking at a group or a DM, which forces an early return and no render

## How did I test?

Ran on web

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: chat info

## Rollback plan

git revert

## Screenshots / videos

image


dnbrwstr (approved):
Lgtm

PR #5945, opened 2026-06-16
ops: auto deploy hermes

latter-bolden:
## Summary

We need to have Hermes automatically redeploy to internal when changes land on `develop`. Helpfully, the existing Openclaw action already does what we need. It already triggers the relevant set of user nodes (internal or all tlonbot) to reload with updated changes from `tlon-apps` when run (irrespective of harness). 

I think the only change we need to make here is to watch the Hermes adaptor source files in addition to the Openclaw ones.

As an aside, I'd like to reorganize the files so we don't have to list them all out explicitly like this. I'd prefer to get auto-deploys working first, then handle that as a follow up.

## Changes

- update the *Restart Bot Ships* action to apply when Hermes changes land
- rename the file to be harness agnostic

## How did I test?

I'll trigger a contrived change on `develop` after this is merged and confirm it updates my Hermes agent.

## Risks and impact

Fixes TLON-6008


latter-bolden:
Fixes TLON-6008

patosullivan (approved):

PR #5948, opened 2026-06-17
fix (openclaw): default dev tlon-skill override to the prebuilt binary

patosullivan:
## Summary

`pnpm dev` from `packages/openclaw` failed for Walt on docker desktop's VirtioFS file-sharing backend while building the local `@tloncorp/tlon-skill` override:

```
failed to rename /run/host_virtiofs/Users/.../packages/tlon-skill/.xxx.bun-build
  to /workspace/tlon-apps/packages/tlon-skill/dist/tlon: ENOENT
```

`bun build --compile` writes its output to a temp file and renames it onto `--outfile`. On VirtioFS the temp lands at the host _realpath_ (`/run/host_virtiofs/…`) while `--outfile` is the _mount_ path (`/workspace/…`); those are two namespaces for the same directory and the cross-namespace rename fails with `ENOENT`. Other backends (e.g. gRPC FUSE) don't have that realpath divergence, so the build only fails on some setups.

The entrypoint already installs `@tloncorp/tlon-skill` from npm — including a working prebuilt platform binary — before the override runs. Building from source only matters when you're actively editing tlon-skill's source, a power-user case. So building-from-source by default was breaking the common "just run the bot" path for no benefit.

## Change

`dev/build-local-skill-override.sh`: gate the source build behind an opt-in `TLON_SKILL_FROM_SOURCE=1`. The default now hydrates the prebuilt binary from the plugin's npm install (a plain `cp`, no bun `--compile`, so the VirtioFS rename issue can't occur). Anyone editing tlon-skill source sets the flag to restore the rebuild-on-start behavior.

## Testing

Brought the dev stack up locally with the new default (flag unset): the override skips the source build, hydrates the `linux-arm64` prebuilt binary, links it into `/workspace/tlon/node_modules/@tloncorp/tlon-skill`, and the gateway reaches `ready` with the tlon tool registered and the firehose connected.

Caveat: I can't reproduce VirtioFS locally (my Docker backend builds fine either way), so I couldn't reproduce the original failure myself. Going to test with Walt to see if this fixes the issue.

### DIFF
--- a/packages/openclaw/dev/build-local-skill-override.sh
+++ b/packages/openclaw/dev/build-local-skill-override.sh
@@ -44,19 +44,27 @@ ln -s "$TLON_SKILL_DIR" "$TARGET"
 # After symlinking, the fallback resolves from the realpath ($TLON_SKILL_DIR/bin/)
 # walking up via the local checkout's node_modules — which on a darwin host won't
 # have the linux binary package installed. So we always produce $TLON_SKILL_DIR/bin/tlon
-# inside the container, either by rebuilding from source (preferred — picks up local
-# edits) or by hydrating the matching platform-arch binary from openclaw-tlon's npm
-# install (fallback when no source is present).
+# inside the container, either by hydrating the matching platform-arch binary from
+# the plugin's npm install (the default — works on any Docker backend) or, when
+# opted in, by rebuilding from source to pick up local edits.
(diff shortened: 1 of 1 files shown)


patosullivan:
Confirmed that this fixed walt's issue.

jamesacklin (approved):

PR #5949, opened 2026-06-17
ops: fix rube piers and update deploy flows

arthyn:
## Summary

This PR fixes our rube ships and refactors our deploys and desk ops to use peru for gathering dependencies from other repos.

## Changes

**Vendor Hoon desk dependencies with peru (replaces the full base-dev rsync)**

- Add `peru.yaml` pinning the urbit (`408k-rc2`) and landscape (`5193635`) revs and picking only the transitively-required base-dev + landscape files, instead of rsyncing all ~119 base-dev files (many unused, some that don't compile on the new kelvin).
- Vendored files land in a separate, **gitignored `desk-deps/`** tree; `desk/` stays purely our committed source. `desk/lib/verb.hoon` (patched 3-arg variant) stays committed and is not picked.
- Add `scripts/sync-deps.sh` (`peru sync`) and `scripts/assemble-desk.sh <target>` (rsync `desk-deps/` with `--delete`, then `desk/` on top, then stamp `commit.txt`).

**Rewire desk-assembly consumers to peru**

- `.github/helpers/deploy.sh`: assemble + `peru sync` in the runner, `scp` the finished desk to the ship; the remote no longer clones urbit/landscape or rsyncs base-dev (no peru needed on ship hosts).
- `apps/tlon-web/rube/{construct-desk.sh,index.ts}`: `peru sync` once, then assemble each ship's desk via `assemble-desk.sh`. Dockerfile installs `peru`/`rsync`/`python3` and copies `peru.yaml` + `scripts/`.
- `ci.yml`: install peru before the bare-runner production smoke test (the parallel e2e job gets it from the Docker image).
- Pin moved in-repo: dropped the now-dead `URBIT_REPO_TAG` / `URBIT_REPO_TAG_KERNEL` Actions vars from the deploy workflows (kernel-integration deploys take their kelvin from the deployed branch's `peru.yaml`).

**`archive-piers.sh`**

- `--skip-prepare` (`SKIP_PREPARE=true`) to roll/chop/archive piers already in `dist/` without re-extracting from the manifest.
- Support both `dist/<ship>/<ship>` and single-nested `dist/<ship>` layouts.

**Cleanup**

- Remove the unused `tm-alpha` desk + its deploy jobs.
- Remove the unused `ops/` + `bin/` bouncer dev tooling.
- Document the workflow in `CLAUDE.md` and `DEVELOPMENT.md`.

## How did I test?

- The assembled `%groups` desk commits clean on a 408 ship.
- Static check: every `/+`/`/-` import reachable by live agents resolves from the peru subset; the only unresolved names are out-of-scope (in-flight contacts→landscape migration, the `de-html-entities` dash/slash quirk, and the aqua-pill `ph-*` helpers).
- Locally verified `assemble-desk.sh` produces the correct tree (`desk-deps/` ∪ `desk/`, disjoint, patched `verb.hoon` preserved, `commit.txt` stamped) and that `archive-piers.sh --skip-prepare` resolves single-nested piers and tars a correctly-rooted `<ship>/` archive.
- CI confirms the pipeline works end to end: `peru` installs on both the bare runner and in the Docker image, and the e2e run logs show `peru sync` → `Assembled desk into …/groups` → `commit %groups` → agents reloading → `~zod/~ten is ready`.

**Still outstanding / not yet verified here:**
- Full e2e green is gated on rebuilding the rube pier archives — the previous archives had a stale `%landscape` desk (old kelvin → `%docket` not running → `/apps/landscape/` 503), being addressed by re-bumping the piers with an updated landscape desk.
- The new `deploy.sh` path (`gcloud compute scp --tunnel-through-iap`) has not been exercised against a live ship yet — recommend a canary/ephemeral deploy before relying on it for mainnet.

## Risks and impact

- Safe to rollback without consulting PR author? **No** — reverting the code restores the old base-dev-rsync flow, but the rube pier archives were re-versioned and `shipManifest.json`/`Dockerfile` point at the new ones; coordinate so the manifest references a known-good archive.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: build/deploy & e2e tooling, Hoon desk assembly (no app runtime code)

## Rollback plan

- Revert the PR commits. The old `deploy.sh`/rube re-clone base-dev and rsync it, so prior behavior is restored; the gitignored `desk-deps/`, `peru.yaml`, and `scripts/*` become inert.
- Confirm `shipManifest.json` and `apps/tlon-web/rube/Dockerfile` point at a pier-archive version that boots cleanly (the previous, pre-bump versions if the new ones are suspect).
- Note: deploy hosts gained no new dependency (peru runs only in the runner), so no host cleanup is required.


Fang- (commented):
Per our discussion out of band, overall approach here is way better than status quo. It'll result in smaller/cruftless desks and gives us explicit control rather than "just copy and hope for the best". Still not all-green in CI obviously and the comments/inline docs could use some trimming & polish but this'll be a good improvement.

PR #5951, opened 2026-06-17
normalize image orientation with renderer

dnbrwstr:
## Summary

Fixes the mobile upload failure for HDR/10-bit HEIC images by patching Expo ImageManipulator's iOS orientation transformer. The previous transformer always built a raw `CGContext` with the source image's bit depth/color space plus a hard-coded bitmap layout, which can fail with `Image context has been lost`. This pathway was easy to hit if the user had hdr enabled for screenshots on ios and attempted to upload an hdr screenshot.

The patch skips already-upright images and uses Expo's `UIGraphicsImageRenderer` helper for non-upright orientation normalization, so UIKit can choose a supported backing context while still applying orientation/mirroring before resize.

Fixes TLON-5979

## Changes

- Patch `expo-image-manipulator@14.0.8` to return early for `.up` images.
- Use `UIGraphicsImageRenderer` + `UIImage.draw(in:)` for non-upright orientation normalization instead of constructing a raw bitmap context.
- Register the updated patch hash in pnpm.
- Document the local patch and removal conditions in `patches/README.md`.

## How did I test?

- Reproduced the original failure path on device with an HDR HEIC screenshot.
- Tested on device against hdr screenshots as well as a number of my other photos and images


janicduplessis:
Can you test with a large video / image, I think changing `preferredAssetRepresentationMode` could regress on performance that we fixed a while back.

dnbrwstr:
@janicduplessis yeah, I do think that would be a concern with the previous solution. I did a little more digging and was able to fully reproduce the issue -- it was occurring when hdr was enabled for screenshots at the iOS level and the user attempted to upload an hdr heic screenshot.

The actual error was occurring inside `expo-image-manipulator`'s orientation normalization, which was failing to create a raw `CGContext` for the image. I fixed it by patching `expo-image-manipulator` to use expo's existing `UIGraphicsImageRenderer` helper + `UIImage.draw(in:)`, which lets UIKit choose a supported backing context while still applying orientation/mirroring before resize.

Tested on hdr screenshots and a number of other photos and worked fine.

janicduplessis:
Nice! Do you think this is upstreamable as is? Looks like a really clean fix if we can confirm the expo helper does pretty much the same as the code that was there.

janicduplessis (approved):

dnbrwstr:
I think it should be upstreamable? The main difference is color profile handling. The helper creates a new renderer context without explicitly passing the source `cgImage.colorSpace`, so it may not preserve that info in exactly the same way.

Previously, when orientation needed normalization and we were doing a simple operation like crop, convert, or compress, expo redrew through a raw `CGContext` using the source color space where possible. With this change, orientation normalization goes through the helper instead. Resize, rotate, and flip already used this helper later in the pipeline, so those flows may already have had the same color profile behavior.

That said, I'm not sure there's any strong intent behind expo's current different treatment of orientation normalization vs resize etc. Looking at the history, it seems like it may just be an old style of manipulation that didn't get cleaned up in recent refactoring.

Expo history: The orientation path was split into its own transformer in [`4e8fafc779c0`](https://github.com/expo/expo/commit/4e8fafc779c0), resize/rotate/flip were moved to `UIGraphicsImageRenderer` in [`f4bdf73de692`](https://github.com/expo/expo/commit/f4bdf73de692). Orientation just stayed on the old raw `CGContext` path.

PR #5952, opened 2026-06-17
android: enable R8 + resource shrinking on release builds

janicduplessis:
## Summary

Enable R8 code shrinking and resource shrinking for Android **release** builds — both are off by default in the RN/Expo template (`android.enableProguardInReleaseBuilds` / `android.enableShrinkResourcesInReleaseBuilds` default `false`). On a release build this cut the APK ~31% and shaved ~216ms off cold-start time-to-login on a low-end device.

## Changes

- `android/gradle.properties`: set `android.enableProguardInReleaseBuilds=true` and `android.enableShrinkResourcesInReleaseBuilds=true`.
- `android/app/proguard-rules.pro`: keep `expo.modules.adapters.react.**`. R8 strips the reflection-loaded `RNHeadlessAppLoader` (used for Expo headless JS — background tasks, notifications, share-intent), so the runtime class lookup throws `ClassNotFoundException` without an explicit keep. This is an upstream bug — `expo-modules-core` annotates `RNHeadlessAppLoader`'s constructor with `@DoNotStrip` instead of the class, so its own bundled keep rule misses it. Fixed in [expo/expo#46920](https://github.com/expo/expo/pull/46920) (class-level `@DoNotStrip`), landing in SDK 57; the keep can be removed once we're on an `expo-modules-core` release that includes it.

## How did I test?

Measured on a physical Galaxy A03s (Android 13, arm64), preview release, **unauthenticated** login-screen cold start, 10 cold starts each (median + IQR):

| | APK (arm64) | tap → login |
|---|---|---|
| baseline | 124.4 MB | 4845 ms (IQR 4784–4907) |
| R8 + shrink | 85.6 MB | 4629 ms (IQR 4588–4651) |
| Δ | **−40.7 MB (−31%)** | **−216 ms** (IQRs don't overlap) |

The app boots to the login screen and navigates (the login-options bottom sheet) correctly, and with the keep rule the `RNHeadlessAppLoader` `ClassNotFound` no longer appears.

## Risks and impact

- Safe to rollback without consulting PR author? **Yes** — revert this PR, or set both flags back to `false`.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [x] Notifications
  - [x] Other: background sync + share-intent (anything relying on Expo headless JS / reflection-loaded classes)

R8 shrinks and obfuscates the entire release class set. Most libraries ship consumer ProGuard rules that are applied automatically, but reflection-loaded classes without a keep rule fail at runtime — I hit one (`RNHeadlessAppLoader`) and added a keep; others may only surface on paths I couldn't exercise from the login screen. The validation checklist above is the gate before this leaves draft.

## Rollback plan

Revert this PR (or set `android.enableProguardInReleaseBuilds` / `android.enableShrinkResourcesInReleaseBuilds` back to `false`). No data or schema impact; the next release build reverts to the un-shrunk APK.

## Screenshots / videos

No UI change — the login screen renders identically with R8 on. Happy to attach a before/after APK-size or cold-start capture if useful.


dnbrwstr (approved):
Lgtm!

PR #5956, opened 2026-06-18
Dedupe phone-number libraries (drop google-libphonenumber)

janicduplessis:
## Summary

The app bundles **two** phone-number libraries: `libphonenumber-js` (used directly in ~7 places for parse/validate) and `google-libphonenumber` (~688KB), the latter pulled in *only* by `react-native-phone-input`. This patches `react-native-phone-input` to use `libphonenumber-js` instead, dropping `google-libphonenumber` from the bundle entirely (~570KB off the release JS bundle in measurement).

## Changes

- `pnpm patch react-native-phone-input@1.3.7` → rewrites `dist/PhoneNumber.js` to use `libphonenumber-js` (`parsePhoneNumberFromString`, `AsYouType`, `.isValid()`, `.getType()`) in place of google-libphonenumber's `PhoneNumberUtil` + `AsYouTypeFormatter`.
- Normalizes the component's lowercase iso2 (e.g. `us`) to the uppercase `CountryCode` libphonenumber-js requires (`US`) — without this every parse/validate silently returns nothing.

## How did I test?

- **Functional test** of the patched module: parse / format / validate / number-type / country-detection across US + international numbers, including the lowercase-iso2 path.
- **On-device** (Galaxy A03s), phone-login screen: typed a number and verified as-you-type formatting, validation (Send button enabling), and the country flag. Before/after below.
- **Bundle**: confirmed `google-libphonenumber` (`i18n.phonenumbers` namespace) is gone from the release bundle; the JS bundle shrank ~570KB.

⚠️ **One visible change:** libphonenumber-js formats the national part with **spaces** (`+1 202 555 0123`) where google used **dashes** (`+1 202-555-0123`). Validation, country detection, and the value submitted are unchanged — only the separator style differs (see the video). Flagging in case design wants the dashes kept.

## Risks and impact

- Safe to rollback without consulting PR author? **Yes** — revert this PR (removes the patch, restores google-libphonenumber).
  - [x] Onboarding (phone-number input in signup + phone login)
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: any flow using the `PhoneInput` component
- The patch is pinned to `react-native-phone-input@1.3.7` (must be re-applied if bumped — and 1.4.x still depends on google-libphonenumber, so upstream won't fix this soon; worth upstreaming the libphonenumber-js switch). Before merge, QA the full phone flow: signup phone entry, **non-US countries**, paste, and edit/delete-while-typing.

## Rollback plan

Revert this PR — drops the patch and restores `google-libphonenumber`. No data/schema impact.

## Screenshots / videos

Phone login on a Galaxy A03s — **left: google-libphonenumber (before), right: libphonenumber-js (after)**. Same number, same as-you-type behavior + validation; note the dash→space separator difference.

https://github.com/user-attachments/assets/5bd2aaf7-783c-4228-9970-4b3df4ef13ad

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,8 @@
       "@tamagui/sheet@2.0.0-rc.41": "patches/@tamagui__sheet@2.0.0-rc.41.patch",
       "@tamagui/static@2.0.0-rc.41": "patches/@tamagui__static@2.0.0-rc.41.patch",
       "react-native-reanimated@4.1.6": "patches/react-native-reanimated@4.1.6.patch",
-      "react-native-gesture-handler@2.28.0": "patches/react-native-gesture-handler@2.28.0.patch"
+      "react-native-gesture-handler@2.28.0": "patches/react-native-gesture-handler@2.28.0.patch",
+      "react-native-phone-input@1.3.7": "patches/react-native-phone-input@1.3.7.patch"
     },
     "allowNonAppliedPatches": true,
     "overrides": {
--- a/patches/README.md
+++ b/patches/README.md
@@ -221,3 +221,49 @@ Removal:
 Remove this patch once `react-native-gesture-handler` ships a version of
 `ReanimatedSwipeable` that disables pointer events on the hidden action
 container, and we confirm the Android repro no longer needs the local fix.
+
+## react-native-phone-input@1.3.7
+
+Why:
+The package bundles `google-libphonenumber` (~688KB) solely for phone-number
+parsing, validation, and as-you-type formatting. We already ship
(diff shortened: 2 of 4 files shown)


janicduplessis:
Closing — going to explore a different approach for the phone-number input rather than patching `react-native-phone-input`. Branch `janic/dedup-phone-lib` is kept for reference (the patch + before/after video show the dedup works; the open question was the dash→space formatting change and the dist-patch fragility).

PR #5957, opened 2026-06-18
fix: uploads broken by Expo 56's global expo/fetch; drop legacy fetch polyfills

janicduplessis:
Fixes TLON-6070

## Summary

Since #5886, image/file uploads are broken on develop — every `fetch().blob()` on native throws `Creating blobs from 'ArrayBuffer' and 'ArrayBufferView' are not supported` (e.g. changing a group icon). The cause: Expo SDK 56's winter runtime [installs `expo/fetch` as the global `fetch` on native by default](https://github.com/expo/expo/blob/f5d56c05565e8dbb0cdc4bb5390f712f0ca620d4/packages/expo/src/winter/runtime.native.ts#L41-L52), and its [`Response.blob()` does `new Blob([ArrayBuffer])`](https://github.com/expo/expo/blob/f5d56c05565e8dbb0cdc4bb5390f712f0ca620d4/packages/expo/src/winter/fetch/FetchResponse.ts#L286-L290), which React Native's built-in `Blob` rejects. There's no upstream fix — Expo installs `fetch` but no matching `Blob`, on `main` too; reported upstream as expo/expo#47468 with a minimal repro.

This PR makes the app work with (and lean into) the new default:

- installs `expo-blob`'s spec-compliant `Blob` as the global, so blob construction from `ArrayBuffer`/`Uint8Array` works app-wide instead of patching each `blob()` call site;
- stops round-tripping uploads and binary scries through blobs where the bytes were never needed;
- removes the community fetch/streaming polyfills that Expo 56 makes redundant (`react-native-fetch-api`, `react-native-polyfill-globals`, and their transitive `text-encoding` — ~706KB of legacy encoding tables):

| global / capability | now provided by |
|---|---|
| streaming `fetch` | `expo/fetch` (the global fetch) |
| `TextDecoder` | Expo winter runtime |
| `TextEncoder` | Hermes (built-in) |
| global `ReadableStream` | Expo's Metro-injected `expo/virtual/streams` |

## Changes

- `packages/app/platform/polyfills.native.ts`: `initializePolyfills` installs `expo-blob`'s `Blob` as the global; the encoding + readable-stream polyfills are removed.
- The `platformFetch` indirection is gone. It existed to give the Urbit SSE channel a streaming fetch; `expo/fetch` resolves to the same module the winter runtime installs as the global, so `apiFetch` just calls `fetch` on both platforms — module-identical, no behavior change for the channel.
- `packages/api/src/http-api/Urbit.ts`: the binary scry reads `response.arrayBuffer()` directly (native under `expo/fetch`), replacing the `blob()` → `readArrayBufferFromBlob` round-trip.
- `packages/shared/src/store/storage/storageActions.ts`: `performUpload` no longer buffers the local file into a Blob via `fetch(file://).blob()` just to read its size. It uses the intent's known `size`/`mimeType`/`name` when present (`fileUri` intents carry them; their `-1` unknown-size sentinel is treated as missing), and otherwise sizes the file through a new platform-split helper: `getLocalFileSize.native.ts` stats via `FileSystem.getInfoAsync`, while the web variant fetches the blob (`blob:`/`data:` URIs can't be statted). The bytes were never the upload body; `uploadFile` streams from the uri.
- `prepareAsset` now reports the image manipulator's *output* format, so a HEIC photo re-encoded to JPEG uploads as `image/jpeg` — previously the stale `image/heic` could be sent.
- Native lockfiles pick up `expo-blob`: `Podfile.lock` gains the `ExpoBlob` pod, and the Android `gradle.lockfile` is regenerated (`lockAllConfigurations()` fails resolution on the shifted transitives otherwise).

## How did I test?

The develop failure reproduces on an iOS dev build: group info → change icon → upload throws the Blob error above.

On this branch, with temporary debug logging on the upload path (intent metadata, resolved size/content type, upload route, transport status), manually exercised the upload matrix against real hosting (memex) storage, logged into a real ship:

| Scenario | Platform | Observed |
|---|---|---|
| HEIC photo | iOS sim | re-encoded to `image/jpeg`, size statted from the manipulator's output file, `contentLength` matched, upload 200 |
| PNG image | Android emulator | stays `image/png`, statted size, 200 |
| 11.2MB video (`fileUri` intent) | Android emulator | intent-carried size trusted — no stat, 200 |
| video poster (webp) | Android emulator | no intent size → statted, 200 |
| JPEG image | web | `blob:` URI sized via the web `getLocalFileSize` (`fetch().blob()`), body uploaded as a Blob, PUT 200 |

Both native apps also log a boot-time check confirming the globals are wired as described: global `fetch` carries Expo's `expo.builtin` symbol and global `Blob` is `expo-blob`'s.

Earlier validation on a Galaxy A03s (preview release):
- Channel connects and **real-time message delivery works** (messages sent from another device arrive live, no refresh) — the SSE stream over the global `expo/fetch`.
- No `ReadableStream` / `TextDecoder` / channel errors at runtime.
- Confirmed via the production bundle's sourcemap that `expo/virtual/streams` (global `ReadableStream`) and Expo winter's `TextDecoder` are present and `text-encoding` is gone — so the removed polyfills are genuinely platform-provided, not missing.

## Risks and impact

- Safe to rollback without consulting PR author? **Yes**, with a caveat — see rollback plan.
  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync (the realtime event channel)
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: image upload / any `blob()` path (global `Blob` now comes from `expo-blob`), and anything relying on global `TextDecoder` / `TextEncoder` / `ReadableStream` (all now platform-provided)
- The realtime channel is validated live on Android, and the upload/`Blob` paths are exercised on all three platforms (see test matrix above).

## Rollback plan

A plain revert returns develop to its current state, where uploads are still broken by the SDK 56 fetch default — the escape hatch there is building with `EXPO_PUBLIC_USE_RN_FETCH=1`, which restores RN's fetch. The revert itself is clean: restores `react-native-fetch-api` / `react-native-polyfill-globals`, drops `expo-blob`, no data/schema impact.

## Screenshots / videos

No UI change. Validated live on a ship (real-time message arrival on the polyfill-free build).


dnbrwstr (approved):
Lgtm

dnbrwstr (approved):

PR #5958, opened 2026-06-18
Clear hosting cookie on Android (CookieManager.clearByName is iOS-only)

janicduplessis:
## Summary

`clearHostingNativeCookie` used `CookieManager.clearByName`, which has **no Android implementation** — the native module (`CookieManagerModule.java`) throws `"Cannot remove a single cookie by name on Android"`. The call is `try/catch`-wrapped, so on Android it just logged `error clearing hosting native cookie: …` and the hosting `SolarisSession` cookie was **never actually cleared**.

It runs during **login/OTP** ([CheckOTPScreen](apps/tlon-mobile/src/screens/Onboarding/CheckOTPScreen.tsx), [useOnboardingHelpers](apps/tlon-mobile/src/hooks/useOnboardingHelpers.ts)) to drop a stale hosting session before establishing a new one.

## Changes

- `apps/tlon-mobile/src/lib/hostingAuth.ts`: on Android, delete `SolarisSession` by passing a raw expired `Set-Cookie` (`SolarisSession=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; Max-Age=0`) through `CookieManager.setFromResponse`, which forwards it straight to WebView's `CookieManager`, where a past `Expires` / `Max-Age=0` removes the entry. iOS keeps using `clearByName`, which works there.

## How did I test?

Validated on a **native `productionDebug` build of this branch** on a Pixel 6 emulator. A self-test seeds a fake `SolarisSession`, then fires a real `fetch` to `tlon.network` **before and after** `clearHostingNativeCookie()`, with the device routed through a local proxy that records the exact `Cookie` header the OS attaches to each request:

```
seed  SolarisSession=FAKE-TOKEN-12345
BEFORE  request to tlon.network  →  Cookie: SolarisSession=FAKE-TOKEN-12345   (sent)
clearHostingNativeCookie()       →  no error
AFTER   request to tlon.network  →  Cookie: (none)                           (not sent)
```

Captured identically on both sides — the app's view of the request and the proxy's independent view of what actually went over the wire. `CookieManager.get('…tlon.network')` corroborates (seeded value → `{}`). So the cookie is genuinely attached to outbound hosting requests before the clear and not sent at all after.

## Risks and impact

- Safe to rollback without consulting PR author? **Yes** — revert (returns to the iOS-only `clearByName` that no-ops on Android).
  - [x] Onboarding (login / OTP)
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: hosting session / cookie handling
- Low blast radius: an Android-only branch inside an already-guarded helper. Worst case it still doesn't clear (same as today's behavior) — no regression.

## Rollback plan

Revert this PR.

## Screenshots / videos

N/A — no UI change.


dnbrwstr (approved):
Lgtm

PR #5961, opened 2026-06-18
fix: country picker search input text is vertically clipped

janicduplessis:
## Summary

The country-codes picker's search field renders its text vertically clipped (only a thin horizontal slice of the placeholder/typed text is visible). The flag/calling-code rows below are fine — just the search input at the top.

## Changes

The picker's search field has a fixed height of `40` (`react-native-country-codes-picker`'s `searchBar` default), but our `textInput` style overrode it with `padding: 16` — 32px of vertical padding on a 40px box leaves only ~8px for the text, so it clips. Switched to `paddingHorizontal: 16`, which keeps the horizontal inset without collapsing the line height, in the three places that style the picker:

- `packages/app/ui/components/Form/PhoneNumberInput.tsx`
- `apps/tlon-mobile/src/components/OnboardingInputs.tsx`
- `apps/tlon-mobile/src/screens/Onboarding/RequestPhoneVerifyScreen.tsx`

## How did I test?

On an Android emulator: opened the country picker from the phone-number login screen and confirmed the search text now renders fully (both the placeholder and typed text), with the rest of the picker unchanged.

## Risks and impact

  - [x] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert the commit — it only changes one style value (`padding` → `paddingHorizontal`) in three files.

## Screenshots / videos

Country picker search field (typed "France"):

**Before** — text clipped:

before

**After** — renders fully:

after


dnbrwstr (approved):
Good catch, lgtm.

PR #5965, opened 2026-06-18
ops: sync staging 6 18

arthyn:
## Summary

Merged m/next-kelvin-408 (staging) into develop, fixes conflicts so we can actually sync back.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:


Fang- (dismissed):
Looks like it could be staging, yes. Except for red X's!

PR #5966, opened 2026-06-18
Migrate onboarding phone input to react-native-transformer-text-input

janicduplessis:
## Summary

Replaces the onboarding phone-number field (`react-native-phone-input`, which bundles google-libphonenumber) with one built on [`react-native-transformer-text-input`](https://github.com/AppAndFlow/react-native-transformer-text-input)'s international `PhoneNumberTransformer`. Formatting runs on the UI thread via a worklet, which removes the per-keystroke flicker the old input has (most visible when spam-deleting — the `+` and flag jump around) and lets us drop the google-libphonenumber dependency entirely.

## Changes

- **`PhoneNumberInput`** — rebuilt on rntti's international `PhoneNumberTransformer`. The calling code is editable inline and the country is detected as you type (`+1 514` → 🇨🇦, `+1 212` → 🇺🇸, `+44` → 🇬🇧); the country picker is still available.
- **`phoneNumberValue.ts`** helpers: `toE164` (normalize to the stored E.164 value), `getCallingCode`/`getFlag`, and `detectCountryFromInput`. The last resolves the NANP territory from the **area code alone, as you type** — libphonenumber only assigns a country once the national number is complete, so it pads the area code to a full valid number (`+1 514` → `+1 514 200-0000` → CA) to disambiguate US/CA/Caribbean immediately, matching the old lib's behavior.
- All onboarding call sites (`OnboardingInputs`, `RequestPhoneVerifyScreen`) now consume the single shared `PhoneNumberInput`.
- Adds `react-native-transformer-text-input ^0.3.0` and **removes `react-native-phone-input`** along with its transitive **google-libphonenumber**.

## How did I test?

- `vitest` unit tests for `phoneNumberValue` (`toE164`, `detectCountryFromInput`) — 9/9 pass; `tsc` and `eslint` clean.
- On device (Android + iOS): typed numbers from several countries, confirmed formatting, as-you-type flag detection, E.164 output, and the country picker.
- Recorded an old-vs-new comparison (below). All numbers are fictional (NANP `555-01xx`, UK Ofcom drama range `7700 900xxx`).

## Risks and impact

  - [x] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

The field only produces an E.164 string; nothing downstream changes. Isolated to the onboarding phone-entry screens.

## Rollback plan

Revert the PR — onboarding returns to the `react-native-phone-input` field and the revert restores the dependency. No data or persisted state involved.

## Screenshots / videos

Captured with a temporary dev toggle (used only to record this, not shipped) comparing the old field (`react-native-phone-input`, gray toggle) with the new one (`rntti`, green toggle). Watch the spam-delete: the old input's flag flickers 🇨🇦→🇺🇸→empty and the `+` blinks; the new one stays steady.

https://github.com/user-attachments/assets/b890dd33-bd00-4b39-a618-d7b10cec56f4

<details><summary>Full-screen recording (with keyboard)</summary>

https://github.com/user-attachments/assets/a594f64b-aa71-41cd-951f-142ba7436188

</details>

### Web

`PhoneNumberInput` lives in shared `packages/app` (only the mobile onboarding screens consume it today, but it ships a Cosmos fixture). It's web-safe: `react-native-transformer-text-input` provides a `.web` build — a plain react-native-web `TextInput`, with no native component or worklet — and that's what the web bundle resolves to via the `.web.*` extension preference. Rendered it under react-native-web and confirmed the transformer formats and emits E.164 the same as on device:

Phone input rendered on web (react-native-web)


dnbrwstr (approved):
Overall lgtm, would just address the codex comments if valid.

Also, you mention removing libphonenumber in the description, but it seems to still be included -- is that intentional?

janicduplessis:
@dnbrwstr The lib we are dropping is `google-libphonenumber` (transitive dep of `react-native-phone-input`) and keeping `libphonenumber-js` which we already use in some app code.

image

PR #5975, opened 2026-06-19
feat(steward/lens): %retry action + configurable count-only retention

wca4a:
Stacked on top of `hm/steward` (PR #5971). Two additions to the lens module — Hunter can fold these in or land separately.

## 1. `%retry` action on the lens module

Adds a `%retry` action variant so an owner can ask the bot ship to re-dispatch a finalized run (failed, aborted, or `no_reply`). The downstream `%context-lens` agent we'd been running had a retry surface; with `%context-lens` folded into `%steward`, there was none — this restores it inside the harness-agnostic umbrella so both openclaw and hermes pick it up.

- `sur/steward.hoon`: lens action becomes a `\$%` union (`%entry` existing, `%retry` new). Lens update becomes a union too (`%entry` for stored records, `%retry-requested` for the retry fact).
- `app/steward.hoon`: poke gate moves from blanket-at-top to per-variant. `%entry` keeps the existing self-or-sponsored-moon gate. `%retry` uses `src=our OR src=configured-owner` — covers a cross-ship poke from the owner planet to the bot moon (the entry gate rejects this direction since the owner isn't a moon the bot sponsors).
- `%retry` doesn't mutate state — it emits a `%retry-requested` fact on `/v1/lens`; the bot's local gateway is subscribed and dispatches a new run, which lands later as an `%entry` from the gateway with a fresh `id` and `retryOf` linking back.
- `mar/steward/action-1.hoon` and `update-1.hoon`: JSON grow/grab updated for the union shapes.

**Wire-shape break:** TS clients poking `{lens:{id,payload,final}}` must move to `{lens:{entry:{...}}}`; updates similarly nested under `%entry`/`%retry-requested`. Gateway + client + agent ship together for our rollout.

## 2. Configurable count-only retention (no time cap)

Lens runs are durable memory of agent activity. The current 90d cap silently ages them out; users want that to stop.

- Drops the 90d `max-run-age`, `le-expired` filter, the `/lens/prune` periodic wake, and the on-init `wait-prune` card.
- Promotes `max-runs-per-bot` from a core arm into lens state, configurable per ship.
- Bumps the default from 1000 → 10000 (~200MB per bot at 20KB/run; well within modern looms).
- Adds `[%lens %configure max-runs-per-bot=@ud]`; the new cap applies to every bot immediately.
- Pruning happens at insert-time (per-bot) and on `%configure` (all bots). No periodic timer.

**State bump to state-1.** `on-load` tries `state-1` first, then `state-0` with migration (legacy bare-map lens wraps into the new record, default cap applied), then bunt on parse failure.

## Scry surface

- `/x/v1/lens/recent` (unchanged): newest 50 as `%entry` updates.
- `/x/v1/lens/recent/<count>`: newest `count`.
- `/x/v1/lens/since/<da>`: every entry with `received >= cutoff`, newest first. Paginate back by passing the oldest `received` from the previous page.
- All parsed-arg paths use `slaw` so malformed atoms return `[~ ~]` instead of crashing the peek.

## Tests

- Existing entry tests reshaped for the union (`[%lens %entry ...]`).
- 6 retry tests: owner-accepted, self-accepted, foreign-rejected, no-owner-foreign-rejected, store-not-mutated, existing-entry-preserved.
- 5 retention tests: init seeds default cap, state-0 migration, count cap prunes oldest on insert, %configure applies across all bots, cap=0 drops everything.
- 2 scry tests: windowed `/recent/<count>`, `/since/<da>`.
- 1 robustness test: malformed scry args return `[~ ~]`.

Removed: `test-old-runs-pruned-by-age`, `test-init-arms-prune-timer-and-activity`, `test-prune-wake-reclaims-and-rearms`, `test-expired-runs-hidden-from-reads` — all obsolete with the time cap gone.

## Rollback plan

`git revert` on the two commits restores state-0 and the existing action surfaces. Existing state-1 ships fall through to bunt on next load (acceptable since lens history is durable on the gateway side).

### DIFF
--- a/desk/app/steward.hoon
+++ b/desk/app/steward.hoon
@@ -213,6 +213,13 @@
   ::  state-0 migration). No time-based expiry — lens runs are durable
   ::  memory of agent activity, not transient logs.
   ::
+  ::  payloads are opaque to %steward, but a sponsored moon could send
+  ::  arbitrarily large cords. cap incoming bytes so a misbehaving (or
+  ::  compromised) gateway can't blow up loom usage with a single poke.
+  ::  the gateway-side truncates to ~50KB; this is a hard ceiling.
+  ::
+  ++  max-payload-bytes  ^-  @ud  524.288
+  ::
   ++  le-handle-configure
     |=  cap=@ud
     ^+  cor
@@ -267,6 +274,12 @@
   ++  le-handle-entry
     |=  [=id:lens:s =payload:lens:s final=?]
     ^+  cor
+    ::  drop oversized payloads to keep loom usage bounded; a sponsored
+    ::  moon misbehaving (or compromised) can't force unbounded growth.
+    ::
+    ?:  (gth (met 3 payload) max-payload-bytes)
+      %-  (slog leaf+"steward: lens payload oversized, dropping" ~)
+      cor
     ?:  =(src.bowl our.bowl)
       (le-fan-out id payload final)
     (le-store src.bowl id payload final)
--- a/desk/app/steward.hoon
+++ b/desk/app/steward.hoon
@@ -161,6 +161,13 @@
       ((slog 'steward: lens run fan-out nacked' u.p.sign) cor)
     ==
   ::
+      [%lens %retry *]
+    ?+  -.sign  cor
+        %poke-ack
+      ?~  p.sign  cor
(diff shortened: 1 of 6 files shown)


arthyn:
closing in favor of #5988

PR #5978, opened 2026-06-22
Tap active bottom tab to scroll to top

janicduplessis:
## Summary

Re-tapping the tab you're already on now scrolls that tab's list back to the top — the standard mobile gesture — on all three main tabs: Home (ChatList), Notifications (Activity), and Contacts.

## Changes

The app uses a custom `NavBarView` over a stack navigator, not a tab navigator, so React Navigation's `useScrollToTop` doesn't apply (it relies on the `tabPress` event only tab navigators emit). So this wires it directly:

- `NavBarView` routes a press on the **already-active** tab to a new optional `onPressActiveTab` callback; inactive taps still navigate exactly as before, so there's no change when switching tabs.
- A small `useScrollTabToTop` hook hands each tab screen a `scrollRef` + an `onPressActiveTab` handler. Each screen threads the ref down to its list and passes the handler to `NavBarView`.
- `scrollContainerToTop` covers the three list kinds in play: `FlashList`/`FlatList` via `scrollToOffset`, `SectionList` via its scroll responder, `ScrollView` via `scrollTo`.

Settings isn't a bottom tab (it's reached from a gear in the Contacts header), so it's out of scope.

## How did I test?

On an Android emulator running a `develop` build: for each tab, scrolled the list down, tapped the active tab icon, and confirmed it animates back to the top. Inactive-tab taps still switch tabs as before. Video below.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: bottom nav + main tab screens (additive; inactive-tab navigation unchanged)

## Rollback plan

Revert this PR. It's a single additive commit — no migrations, schema, or state changes.

## Screenshots / videos

https://github.com/user-attachments/assets/f886866c-bf28-45e9-bbd9-5595fdd85dc0

### DIFF
--- a/packages/app/features/chat-list/ChatList.tsx
+++ b/packages/app/features/chat-list/ChatList.tsx
@@ -1,4 +1,4 @@
-import { FlashList, ListRenderItem } from '@shopify/flash-list';
+import { FlashList, FlashListRef, ListRenderItem } from '@shopify/flash-list';
 import * as db from '@tloncorp/shared/db';
 import { isEqual } from 'lodash';
 import React, { useCallback, useMemo } from 'react';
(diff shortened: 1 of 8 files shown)


dnbrwstr (approved):
Lgtm!

PR #5983, opened 2026-06-22
Link hosted web users to hosting's bot settings page

patosullivan:
## Summary

Fixes TLON-6016

Adds web access to bot settings by linking hosted web users out to the central bot settings page at `https://tlon.network/tlonbot`. This is intentionally narrower than an in-app web implementation, since hosted web has ship auth but not central hosting auth, so in-app calls to the hosting and Tlonbot APIs cannot be relied on without a larger auth and CORS project. Native mobile behavior is unchanged and still navigates to the internal `BotSettings` route.

## Changes

- Added `packages/app/utils/botSettings.ts` with the shared link-out behavior:
  - `BOT_SETTINGS_URL` constant pointing at `https://tlon.network/tlonbot`.
  - `openExternalBotSettings()`, which opens a new tab via `window.open(url, '_blank', 'noopener,noreferrer')` on web and falls back to `Linking.openURL` on native.
  - `useHasExpectedBotDm(currentUserId, enabled)`, a hook that returns true when the user has an established (non-pending, non-invite) DM with their expected bot id (`~pinser-botter-<self>`). This is used as a local proxy for "the bot is provisioned", because `hostingBotEnabled` is not available on hosted web.
- Updated `packages/app/features/settings/SettingsScreen.tsx` (mobile) and `packages/app/navigation/desktop/SettingsNavigator.tsx` (desktop/web) so the "Bot Settings" row shows on web for hosted users gated by `useHasExpectedBotDm`, while native keeps the existing `isHostedUser && hostingBotEnabled` gate. Pressing the row opens the external URL on web and navigates to the internal `BotSettings` route on native.
- Updated `packages/app/features/top/UserProfileScreen.tsx` so the own-bot profile "Bot settings" entry also shows on web, gated on hosted status, `useHasExpectedBotDm`, and `isOwnBotProfile`. The hosted check is wrapped in `getCurrentUserIsHostedSafely()` because `getCurrentUserIsHosted()` can throw before the Urbit client is initialized. Pressing the entry opens the external URL on web and navigates internally on native.

Out of scope for this patch: in-app web MCP OAuth, iframing `tlon.network/tlonbot` (it sends `x-frame-options: SAMEORIGIN`), adding hosting auth to the web app, and any hosting-service CORS or auth changes. `BotSettings`, `BotMcpSettings`, and `BotOtherSettings` remain unreachable from web.

## How did I test?

- Manual web verification
- Native smoke check: app settings and own-bot profile still navigate to the internal `BotSettings` route, and `Connect MCP` / `External Services` behavior is unchanged.

## Risks and impact

  - [x] Other: Settings and user-profile bot settings entry points

## Rollback plan

Revert.

## Screenshots / videos

image

### DIFF
--- a/packages/app/features/settings/SettingsScreen.tsx
+++ b/packages/app/features/settings/SettingsScreen.tsx
@@ -12,6 +12,10 @@ import { useHandleLogout } from '../../hooks/useHandleLogout';
 import { useResetDb } from '../../hooks/useResetDb';
 import { RootStackParamList } from '../../navigation/types';
 import { SettingsScreenView, View, openTlonWebApp } from '../../ui';
+import {
+  openExternalBotSettings,
+  useHasExpectedBotDm,
+} from '../../utils/botSettings';
 
 type Props = NativeStackScreenProps<RootStackParamList, 'Settings'>;
 
@@ -23,7 +27,14 @@ export default function SettingsScreen(props: Props) {
   const hasHostedAuth = useHasHostedAuth();
   const hostingBotEnabled = db.hostingBotEnabled.useValue();
   const isHostedUser = getCurrentUserIsHosted();
-  const botEnabled = Platform.OS !== 'web' && isHostedUser && hostingBotEnabled;
+  const hasExpectedBotDm = useHasExpectedBotDm(
+    currentUserId,
+    Platform.OS === 'web' && isHostedUser
+  );
+  const botEnabled =
+    Platform.OS === 'web'
+      ? isHostedUser && hasExpectedBotDm
+      : isHostedUser && hostingBotEnabled;
   const navigationRef = useMutableRef(props.navigation);
 
   const onAppInfoPressed = useCallback(() => {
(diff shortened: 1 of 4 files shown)


packages/app/features/settings/SettingsScreen.tsx
```diff
@@ -23,7 +27,14 @@ export default function SettingsScreen(props: Props) {
   const hasHostedAuth = useHasHostedAuth();
   const hostingBotEnabled = db.hostingBotEnabled.useValue();
   const isHostedUser = getCurrentUserIsHosted();
-  const botEnabled = Platform.OS !== 'web' && isHostedUser && hostingBotEnabled;
+  const hasExpectedBotDm = useHasExpectedBotDm(
```
dnbrwstr, line 30: This sequence is repeated in several places, I'd centralize.

packages/app/features/top/UserProfileScreen.tsx
```diff
@@ -21,12 +21,24 @@ import {
...
+  openExternalBotSettings,
+  useHasExpectedBotDm,
+} from '../../utils/botSettings';
 import { useShipConnectionStatus } from './useShipConnectionStatus';
 
 type Props = NativeStackScreenProps<RootStackParamList, 'UserProfile'>;
 
 const logger = createDevLogger('UserProfileScreen', false);
 
+function getCurrentUserIsHostedSafely() {
```
dnbrwstr, line 34: Why does only this screen need this check? Can we make more consistent?

dnbrwstr (approved):
Lgtm, couple nits

jamesacklin (approved):

PR #5987, opened 2026-06-23
context lens: bot run inspector + %steward client 4/5

arthyn:
Bot-run introspection for Tlon: a per-message inspector showing what a bot did for any stamped message, backed by durable owner-side run history on the `%steward` agent. Absorbs and supersedes #5925 (now retargeted from the removed `%context-lens` agent to `%steward`, and rebased onto `develop`).

## Feature

- **Bot run inspector UI**, gated behind the `contextLens` feature flag (off by default): status, trigger, context sources, tool calls (expandable args), writes, model, runtime, and a run timeline.
- **Ship-path data flow**: the `%steward` agent (lens module) stores durable run history owner-side; runs sync to the client db via `/v1/lens` facts with scry backfill, driving React Query hooks (`useContextLensRun`, `useRecentContextLensRuns`). No gateway reachability needed from the device.
- **Wide (desktop web)**: channel-header toggle opens a 360px sidebar panel with live stream, message-select inspection, and recent-run history.
- **Narrow / native**: inline "⟐ Bot run" badge and long-press message action open a per-message run sheet (avatar header + run summary + Expand); channel-header button pushes a recent-runs screen; rows and Expand push a full run-inspector screen (`lens/runs`, `lens/run/:botShip/:lensId`).
- **Retry**: a failed run can be re-dispatched from the inspector — `retryLensRun` pokes `%steward`, which routes to the owning bot's gateway.

## `%steward` client surface (`packages/api`)

Targets the per-module `%steward` contract (typed `$json` payloads, no cord parsing):

- App `steward`; poke mark `steward-lens-action-1` (`{ entry }`, `{ retry }`); core `steward-action-1` `{ configure }`.
- Subscribe `/v1/lens` (handles `%entry`; ignores `%retry-requested`, which targets the bot's gateway).
- Scry `/v1/lens/recent[/<count>]`, `/v1/lens/since/<da>` (paginate history), `/v1/lens/run/<bot>/<id>` — all tagged `LensUpdate` / `LensRecentScry`.

## Dependencies

The ship + gateway halves of the same feature:

- **#5971** — the `%steward` umbrella agent (lens module: per-module marks, typed JSON payload). The store this UI reads.
- **#5988** — `%retry` action + count-only retention + the `%recent` / `/since` scries this client consumes. (Stacked on #5971.)
- **#5972** — the openclaw producer: records runs, stamps outbound bot messages with the lens reference blob this UI keys off, and pokes `%steward` with finalized runs.

[...]

PR #5988, opened 2026-06-23
feat(steward/lens): %retry action + count-only retention 5/5

arthyn:
Reconciles #5975's lens-module features (retry + retention) onto the per-module + typed-`$json` structure from #5971. #5975 was written against the old umbrella marks and cord payload, so it can't merge directly onto the split — this re-implements its semantics on the current base. Once this lands, #5975 can close.

Stacked on #5971 (base `hm/steward`).

## What's preserved from #5975

- **`%retry` action + cross-ship routing.** Owner-initiated re-dispatch of a failed run. `bot == our` → emit a `%retry-requested` fact on `/v1/lens` for the local gateway; `bot != our` → relay a cross-ship `%retry` poke to the bot's steward. The symmetric case to `%entry` (bot → owner): retry is owner → bot.
- **Count-only retention.** `max-runs-per-bot` (default 10,000), set via `%configure`. No time-based expiry, no prune timer — lens runs are durable memory, not transient logs. Pruned on insert and on `%configure`.
- **Payload size cap** — incoming `%entry` payloads over 512KB (jammed) are dropped to bound loom usage.
- **`/since/[da]` scry** for incremental backfill, alongside `/recent` and `/recent/[count]`.

## How it differs from #5975 (adapted to #5971)

- Lens action/update are tagged unions on the **per-module** `%steward-lens-action-1` / `%steward-lens-update-1` marks (not the umbrella `%steward-action-1` with a `{lens:…}` envelope).
- `%entry` payload stays a typed **`$json`** value (not a `@t` cord); the size cap measures the jammed noun.
- The `%recent` scry result folds into the `%steward-lens-update-1` mark (`%recent` variant), replacing the separate `update-list-1` mark added in #5971.
- Auth is per-variant (entry: self / sponsored-moon via `+sein:title`; retry: self / configured-owner; configure: self).
- Agent state bumps `%0 → %1`; on-load migrates the legacy bare-map lens state by seeding the default cap.

## Downstream follow-ups (separate branches, mine)

- **Producer (#5972)**: must wrap lens pokes in `%entry` (`{ entry: { id, payload, final } }`) now that the action is tagged.
- **Client (#5987)**: `/v1/lens` facts and `/run` scry now arrive tagged (`{ entry: … }`); `/recent` returns `{ recent: [ … ] }`. The client must unwrap. `retryLensRun` already pokes `steward-lens-action-1 { retry }`, which this PR's `%retry` variant accepts.

## Verification

- Full desk builds + **38 steward** and **6 gateway-status** tests pass on **zuse 408** (simtyc, matching CI).


desk/app/steward.hoon
```diff
@@ -36,16 +49,29 @@
...
+      [~ this(state p.one)]
+    =/  zero  (mule |.(!<(state-0 ole)))
+    ?:  ?=(%& -.zero)
+      =/  upgraded=state-1
+        :*  %1
+            owner.p.zero
+            [lens.p.zero default-max-runs-per-bot]
+            gateway.p.zero
+        ==
+      [~ this(state upgraded)]
```
mikolajpp: What is the reason we resort to mule here?

desk/app/steward.hoon
```diff
@@ -16,15 +16,28 @@
 /+  default-agent, verb, dbug
 |%
 +$  card  card:agent:gall
+::  legacy state: lens was a bare map with no retention config. on-load
+::  migrates it into state-1 with a default cap.
+::
```
mikolajpp:
```suggestion
```

desk/app/steward.hoon
```diff
@@ -16,15 +16,28 @@
...
+  $:  %1
       owner=(unit ship)
       lens=state:v1:l
       gateway=state:v1:g
   ==
+::  default cap on first install or state-0 → state-1 migration. generous:
+::  at ~20KB/run that's ~200MB per bot, within loom budgets. ships wanting
+::  more or less can poke %steward-lens-action-1 %configure.
 ::
+++  default-max-runs-per-bot  ^-  @ud  10.000
```
mikolajpp:
```suggestion
++  default-max-runs-per-bot  10.000
```
I keep seeing these casts. Why are they necessary?

desk/app/steward.hoon
mikolajpp: We talked off band, it would be good to clean it up and use the standard migrate pattern. Curious to know what made the model resort to this.

desk/app/steward.hoon
```diff
@@ -192,35 +212,100 @@
...
+      %+  give  %fact
+      :*  ~[/v1/lens]
+          %steward-lens-update-1
+          !>(`update:v1:sl`[%retry-requested id requester])
+      ==
+    %-  emit
+    :^    %pass
+        /lens/retry/(scot %p bot)/(scot %t id)
+      %agent
+    [[bot %steward] %poke %steward-lens-action-1 !>(`action:v1:sl`[%retry bot id])]
```
mikolajpp, line 307: As mentioned by the machine reviewer, this is not the logic we want. We should rather distinguish on `src.bowl == our.bowl` to trigger sending of retry. If the poke is not local, we should then assert `bot == our.bowl`, to prevent accidental proxying.

desk/app/steward.hoon
```diff
@@ -235,14 +320,24 @@
...
+    ::
+        [%v1 %since @ ~]
+      =/  parsed  (slaw %da i.t.t.path)
+      ?~  parsed  [~ ~]
+      ``steward-lens-update-1+!>(`update:v1:sl`[%recent (le-since u.parsed)])
     ::
         [%v1 %run @ @ ~]
-      =/  bot  (slav %p i.t.t.path)
+      =/  parsed-bot  (slaw %p i.t.t.path)
+      ?~  parsed-bot  [~ ~]
```
mikolajpp: Isn't the first version better? `[~ ~]` crashes without line number I think.

desk/app/steward.hoon
```diff
@@ -252,81 +347,102 @@
...
-    ::  we mark any live runs above the fixed limit as dead
-    ::
-    =?  dead  (gth (lent live) max-runs-per-bot)
-      =/  sorted
-        %+  sort  live
-        |=  [a=[* =run:v1:sl] b=[* =run:v1:sl]]
-        (lth received.run.a received.run.b)
-      (weld dead (scag (sub (lent live) max-runs-per-bot) sorted))
-    =/  keys  (turn dead |=([k=[bot=ship =id:v1:sl] *] k))
+    ?:  (lte (lent mine) max-runs-per-bot.lens.state)
```
mikolajpp, line 382: Resume review here.

desk/app/steward.hoon
```diff
@@ -252,81 +347,102 @@
...
+    :*  ~[/v1/lens]
+        %steward-lens-update-1
+        !>(`update:v1:sl`[%entry [bot id] run])
+    ==
+  ::
+  ::  trim a single bot's records to .max-runs-per-bot, dropping the oldest
+  ::  by .received first. invoked on every insert and on %configure.
   ::
   ++  le-prune
     |=  for=ship
```
mikolajpp:
```suggestion
    |=  who=ship
```
or just `=ship` would be more standard.

desk/app/steward.hoon
```diff
@@ -252,81 +347,102 @@
...
+      |=  [a=[* =run:v1:sl] b=[* =run:v1:sl]]
+      (gth received.run.a received.run.b)
+    %+  turn  (scag count sorted)
+    |=  [[bot=ship =id:v1:sl] =run:v1:sl]
+    `entry:v1:sl`[[bot id] run]
+  ::
+  ::  every entry with .received >= cutoff, newest first; paginate history by
+  ::  passing the oldest .received from the last page
+  ::
+  ++  le-since
```
mikolajpp: I can't see any pagination going on here...

mikolajpp (changes requested):
Overall this looks good. There is one issue regarding the logic for retries, already caught by machine reviewer.

After that is addressed it should be good to go.

arthyn:
Added a commit collapsing the agent to a single `state-0` with a bunt-only on-load — since %steward is greenfield, the `state-0→state-1` migration only bridged branch-only state (per "never ship cleanup for branch-only state"). Versioned state + real migrations get added back when steward actually ships. 37 tests green on 408.

PR #6011, opened 2026-06-26
Return to 409k

mikolajpp:
## Summary

Given the uncertainty surrounding 408k release, which may yet take a long time, we have decided to go back to releasing off 409k-compatible branches. At the same time, branching off of old 409k release would lose us important backend bug fixes, guardrails work and otherwise diverge from the mainline body of work.

This PR is a structural reversal of select commits strictly related to 408k, while keeping all other backend and client commits and fixes made since.

This was only possible thanks to @Fang- terrific commit discipline during construction of 408k PR.

This is a dual purpose PR:
- It can be used as a starting point for cutting 409k releases from now on.
- It can also be rebased on top of develop to recover a clean 409k view if we decide we want to do a hard stop on 408k. 

There is also a clear path to re-releasing 408k when the time comes: many of the changes necessary for 408k were not at all specific to the kelvin change, but were related to broken marks or bug fixes. The few reverted commits can be re-applied in the future when 408k proves stable.

## Changes

The specific commits reverted are laid out in the history. 

## How did I test?

Applied this branch to an up-to-date livenet moon running on 409k, with an already pending 408k update.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:


mikolajpp:
We have done a release using this branch as a basis, but the actual release was based off a separate branch [tloncorp/release-july](https://github.com/tloncorp/tlon-apps/tree/release-july)

arthyn:
won't need since 408 is going out

PR #6014, opened 2026-06-26
feat(steward/lens): explicit trusted-bots %entry gate + retry/retention hardening

wca4a:
Stacked on `hm/steward` (#5971). Reopens the scope of the now-closed #5975 and adds the trusted-bots auth fix on top. Steward-only diff (app + marks + sur + tests + docs); 7 commits.

## Headline: explicit trusted-bots `%entry` gate (replaces moon-sponsor-only)

#5971 gated lens `%entry` fan-out to **moons sponsored by the owner**. That rejects a planet→planet setup: bot `~sitrul-nacwyl` (planet) fanning runs to owner `~malmur-halmex` (planet) is refused.

This replaces the moon-sponsor check with an explicit, **ship-class-agnostic trusted-bots set**:
- `state-2` adds `bots=(set ship)`. Owner accepts `%entry` from `our` **or** any ship in `bots`.
- Self-only `%trust-bot` / `%untrust-bot` pokes manage the set.
- `on-load` migration is load-bearing: `state-2 → state-1 (+bots=~) → state-0 (+cap +bots=~) → bunt`. In-place mutation of `state-1` would silently bunt-reset live ships (wiping owner/runs/gateway). Trust is **not** auto-granted by moon sponsorship — operators must re-grant with `%trust-bot` post-upgrade.

## Also in this branch (the former #5975 scope)

- **`%retry` action** — owner-initiated re-dispatch of a finalized run; routes via the owner ship when `bot != our`. Emits a `%retry-requested` fact on `/v1/lens` rather than mutating state.
- **Count-only retention** — drops the 90d time cap, the prune timer, and `le-expired`; promotes `max-runs-per-bot` into state, configurable via `[%lens %configure]`, default 10000.
- **Payload cap** — hard 512KB ceiling on incoming `%entry` payloads so a compromised gateway can't blow up loom usage.
- **Scry JSON-friendliness** — `/x/v1/lens/recent`, `/recent/<count>`, `/since/<da>`; malformed args `slaw` to `[~ ~]` instead of crashing the peek.

## Mark build fixes (this session)

- `action-1` grab: manual key dispatch couldn't narrow a bare `@` into the `@p` aura the action union expects — routed through the `of` reparser.
- `update-1` grow: a multiline tall-form `?~` inside a wide `[ ]` cell fails to parse — inlined as `?~(...)`.

Both marks now build (verified against a live 408 kernel).

## Test plan

- [ ] `desk/tests/app/steward.hoon`: entry-union, 6 retry, 5 retention, 2 scry-window, 1 malformed-args robustness.
- [ ] Trusted-bots: `%trust-bot`/`%untrust-bot` self-only; `%entry` accepted from a trusted planet, rejected from an untrusted ship.
- [ ] `state-1 → state-2` migration preserves owner/runs/gateway (no bunt-reset).
- [ ] Marks build on a 408 desk.

### DIFF
--- a/desk/app/steward.hoon
+++ b/desk/app/steward.hoon
@@ -213,6 +213,13 @@
   ::  state-0 migration). No time-based expiry — lens runs are durable
   ::  memory of agent activity, not transient logs.
   ::
+  ::  payloads are opaque to %steward, but a sponsored moon could send
+  ::  arbitrarily large cords. cap incoming bytes so a misbehaving (or
+  ::  compromised) gateway can't blow up loom usage with a single poke.
+  ::  the gateway-side truncates to ~50KB; this is a hard ceiling.
+  ::
+  ++  max-payload-bytes  ^-  @ud  524.288
+  ::
   ++  le-handle-configure
     |=  cap=@ud
     ^+  cor
@@ -267,6 +274,12 @@
   ++  le-handle-entry
     |=  [=id:lens:s =payload:lens:s final=?]
     ^+  cor
+    ::  drop oversized payloads to keep loom usage bounded; a sponsored
+    ::  moon misbehaving (or compromised) can't force unbounded growth.
+    ::
+    ?:  (gth (met 3 payload) max-payload-bytes)
+      %-  (slog leaf+"steward: lens payload oversized, dropping" ~)
+      cor
     ?:  =(src.bowl our.bowl)
       (le-fan-out id payload final)
     (le-store src.bowl id payload final)
--- a/desk/app/steward.hoon
+++ b/desk/app/steward.hoon
@@ -161,6 +161,13 @@
       ((slog 'steward: lens run fan-out nacked' u.p.sign) cor)
     ==
   ::
+      [%lens %retry *]
+    ?+  -.sign  cor
+        %poke-ack
+      ?~  p.sign  cor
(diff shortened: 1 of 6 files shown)


arthyn:
Closing in favor of the per-module stack. We're settling on the **per-module** steward contract (per-module marks, unwrapped JSON, typed-`$json` payload — Mikolaj's review direction), which is what's implemented end-to-end now: the agent (#5971 → #5988), the openclaw producer (#5972), and the client (#6007–6010). This PR is on the umbrella contract (rewrites `action-1` to `{lens:{…}}` dispatch, re-adds the umbrella `update-1` mark, cord payload), so it can't merge alongside that stack.

The genuinely new piece here — the **trusted-bots `%entry` gate** (the planet→planet fix) — is worth keeping, so I ported it onto the per-module structure in **#6018** (stacked on #5988): `state-2` + `bots=(set ship)`, core `%trust-bot`/`%untrust-bot` pokes, the `%2→%1→%0` migration chain, and removal of moon-sponsor auto-trust. 43 steward tests green on 408.

The retry / count-only retention / payload-cap / scry-window work here already lives in #5988 (per-module). So everything in #6014 is now captured across #5988 + #6018 on the per-module contract. Thanks for the trusted-bots design — carried it over as-is.

PR #6016, opened 2026-06-26
fix(openclaw/lens): per-account global services + route telemetry

wca4a:
Stacks on Pat's context-lens gateway port (#5936, `po/migrate-context-lens`) as a hardening delta. Addresses two correctness gaps in how the process-global lens services resolve config under a multi-account `channels.tlon`.

## Summary
- **Global lens services resolve the right account.** The context-lens HTTP routes, disk store, and %steward ship sync are process-global but previously resolved config against the `default` account only. A lens configured under a named account (with no top-level block) silently stayed off. New `resolveLensAccountId(cfg)` returns the first lens-enabled account, preferring `default`; all three init paths now use it.
- **Outbound-route telemetry attributes the right account.** `setOutboundRouteReporter` lives in a single process-global slot (last monitor to register wins) but hard-bound owner/bot ship to its own closure, mis-attributing sibling accounts' sends. `TlonOutboundRouteEvent` now carries `accountId` (populated from the `message_sending` hook context), and the reporter resolves owner/bot from it — matching the debug/error reporters.

## Test plan
- [x] `pnpm tsc --noEmit` clean
- [x] `pnpm exec vitest run src/` — 743/743 pass (added 3 `resolveLensAccountId` cases)
- [x] `prettier --check` clean
- [ ] Manual: verify a named-account-only lens config registers routes + ship sync

### DIFF
--- a/packages/openclaw/README.md
+++ b/packages/openclaw/README.md
@@ -82,9 +82,11 @@ channels:
 
 ## Telemetry
 
-Telemetry is disabled by default. The plugin only sends tlemetry events when `channels.tlon.telemetry.enabled` is set to `true` and an API key is configured.
+Telemetry is disabled by default. The plugin only sends telemetry events when `channels.tlon.telemetry.enabled` is set to `true` and an API key is configured.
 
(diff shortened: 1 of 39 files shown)


arthyn:
closing in favor of #6021

PR #6017, opened 2026-06-26
fix(lens/api): repoint client at %steward umbrella module

wca4a:
Stacks on Hunter's #6007 (`hm/lens-1-api`) as a single delta. Aligns the lens **client** API with the module-nested `%steward` agent (#5971), which #6007's client still addresses with the old flat lens-only protocol.

## The gap this closes
The agent (#5971) moved lens under a `{ lens: ... }` umbrella module. #6007's client still speaks the pre-umbrella protocol, so its lens traffic wouldn't match the current agent:
- retry poke mark `steward-lens-action-1` → **`steward-action-1`** with `{ lens: { retry: { bot, id } } }`
- scry/response shape `response.recent` → **`response.lens.recent`** (`LensRecentScry = { lens: { recent } }`)
- payload is now a serialized-JSON CORD parsed at the API edge (`parsePayload`)
- adds `LensModuleUpdate` / `StewardUpdate` / `LensModuleAction` types + `isLensEntryUpdate` guard

Only `packages/api/src/client/lensApi.ts` and `urbit/lens.ts` change (~120 lines). This is the missing glue between #5971 (agent) and the #6007→#6010 client stack.

## Test plan
- [x] `pnpm --dir packages/api tsc` clean against current `hm/lens-1-api`
- [ ] Manual: client lens recent/retry round-trips against a ship running the umbrella `%steward`

### DIFF
--- a/packages/api/src/client/lensApi.ts
+++ b/packages/api/src/client/lensApi.ts
@@ -22,29 +22,44 @@ export const toLensRun = (entry: ub.LensRunEntry): LensRun => {
     lensId: entry.id,
     complete: entry.complete,
     receivedAt: parseReceived(entry.received),
-    // %steward relays the run record as structured JSON, not a cord — pass
-    // it through unchanged.
-    payload: entry.payload,
+    payload: parsePayload(entry.payload),
   };
 };
 
-export const getRecentLensRuns = async (count?: number): Promise<LensRun[]> => {
+// The agent relays payloads as opaque serialized-JSON cords; parse once at
+// the API edge so the rest of the client sees structured data.
+function parsePayload(payload: string): unknown {
+  try {
+    return JSON.parse(payload);
+  } catch {
+    logger.log('failed to parse lens payload', payload.slice(0, 100));
+    return null;
+  }
+}
+
+// %steward emits three lens-update variants on the wire. %entry / %recent
+// carry data; %retry-requested is a signal for the gateway, not the client.
+function isLensEntryUpdate(
+  update: ub.LensModuleUpdate
+): update is { entry: ub.LensRunEntry } {
+  return 'entry' in update;
+}
+
+export const getRecentLensRuns = async (
+  count?: number
+): Promise<LensRun[]> => {
   const path =
     count && count > 0 ? `/v1/lens/recent/${count}` : '/v1/lens/recent';
   const response = await scry<ub.LensRecentScry>({ app: 'steward', path });
-
-  return response.recent.map(toLensRun);
+  return response.lens.recent.map(toLensRun);
 };
 
-// Paginate run history backwards: pass the oldest receivedAt (@da string)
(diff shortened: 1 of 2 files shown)


wca4a:
## Why this repoint is correct: the canonical wire contract is the one the openclaw gateway speaks

An automated review flagged three findings here (recent shape, payload parsing, retry mark) arguing this PR diverges from "the lens module that landed." That's judging against the *base client's* per-module assumption (`hm/lens-1-api`), which **no shipping steward agent actually implements**. The ground truth is the gateway plugin (`@tloncorp/openclaw`), since it's the thing that pokes `%steward` with run records.

### What the gateway plugin speaks (write side)
`context-lens-ship-sync.ts`:
- mark **`steward-action-1`** (the umbrella mark — *not* `steward-lens-action-1`)
- body **`{ lens: { entry: { id, payload, final } } }`**
- payload is **serialized to a cord / JSON string** (explicit comment: "the agent stores payloads as cords")

So the gateway pins an **umbrella + cord-payload** contract.

### The three contracts in play
| | accepts gateway write (`steward-action-1` / `{lens:{entry}}` / cord) | serves client read `{lens:{recent}}` | internally consistent |
|---|---|---|---|
| `hm/steward` (#5971) | yes | **no — `/recent` returns a bare unmarked `noun`** | **no — base client uses `steward-lens-action-1`, a mark this agent never defines** |
| our steward fork (`wsa/steward-retry-retention`) | yes | yes (`%recent` variant) | yes (write + read + retry all umbrella) |

`hm/steward` only defines two marks — `steward-action-1` and `steward-update-1`. There is no `steward-lens-action-1` / `steward-lens-update-1`. Its `/run` and facts already emit `{lens:{...}}` and it accepts retry as `steward-action-1` + `{lens:{retry}}`. The one gap is `/x/v1/lens/recent`, which returns `noun+!>(le-recent)` — an unmarked noun that won't cleanly convert to either JSON shape.

### Therefore the three findings are stale against the real contract
- **Retry mark:** gateway uses `steward-action-1`; `steward-lens-action-1` exists on no agent. This PR is right.
- **Payload:** the wire value is a cord (`+$ payload @t`, mar emits `s+payload`), i.e. a JSON *string*. `parsePayload` is correct; passing it through unchanged hands the UI a raw string. This PR fixes a latent bug.
- **Recent shape:** the canonical agent serves `{lens:{recent}}`, so `response.lens.recent` is correct.

### What this implies for the merge
This PR is correct as-is. The convergence work belongs on the agent side: `hm/steward`'s `/recent` peek should emit the umbrella `{lens:{recent}}` shape (as our fork does) rather than a bare noun, so the client read path matches the umbrella write path the gateway already uses.

arthyn:
this was written against out of date %steward agent, we intentionally diverged from an all encompassing action set in https://github.com/tloncorp/tlon-apps/pull/5971

PR #6028, opened 2026-06-29
lens: open panel from bot run button + sync toggle to settings

arthyn:
## Summary

Improves the desktop experience so that when the bot run button is clicked, it opens the inspector panel and highlights the selected message. Also switches lens to a settings backed feature flag instead of client only. 

## Changes

**Bot run button → context lens panel (desktop)**
- On desktop, the "⟐ Bot run" badge and the "View bot run" action now open the context lens panel with that message selected, instead of the `ContextLensRunSheet` bottom sheet. Mobile/narrow keeps the sheet as a fallback.
- The inspected message is highlighted in the scroller, reusing the existing `isSelected` selection visual (gated to when the panel is open on desktop).

**Back the lens toggle with the synced settings store**
- Moved the context lens enable toggle off the client-local feature flag store and onto the Urbit-backed `%settings` store, so it persists and syncs across a user's devices.
- Added `contextLensEnabled` to the settings schema, API models/urbit types, bucket routing (`groups`), and `toClientSettings`; added an `updateContextLensEnabled` action and a `useContextLensEnabled` hook.
- Dropped `contextLens` from `featureMeta`; the two readers (`useContextLensAvailable` / `useContextLensGatewayConfig`) now read the setting, and the Feature Flags screen renders a settings-backed row for it.
- No Hoon change (the `%settings` agent stores arbitrary keys in the groups bucket). Regenerated the consolidated bootstrap migration.

## How did I test?

- `tsc --noEmit` passes clean for `packages/api`, `packages/shared`, and `packages/app`.
- Regenerated the migration via `pnpm generate:migration` and verified the consolidated bootstrap migration contains both the existing `context_lens_runs` table and the new `context_lens_enabled` column (nothing dropped).
- Not yet exercised in a running app / e2e — manual verification of the desktop panel-open + highlight behavior still pending.

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers (settings store)
  - [ ] Message sync
  - [x] Channel display (context lens panel / bot run button)
  - [ ] Notifications
  - [x] Other: settings schema + migration regeneration

Note: existing users who had the old client-local `contextLens` flag enabled will see it default to `false` until they re-toggle it (no migration of the old local value). Acceptable for a desktop/experimental flag; can add a one-time migration if desired.

## Rollback plan

Revert the two commits on this branch. The new `contextLensEnabled` settings key is additive and harmless if left behind; the schema/migration change is regenerated from `schema.ts`, so reverting the schema and re-running `pnpm generate:migration` restores the prior migration.


jamesacklin (approved):
Well done on the negative LOC

patosullivan (approved):

PR #6030, opened 2026-06-30
logs: grafana backend

mikolajpp:
## Summary

Grafana/loki stack is much more suitable for gathering and processing backend telemetry than Posthog. As a first step towards moving our telemetry to grafana, we implement a grafana logs backend that will run alongside the posthog for the time being, allowing us to explore the setup before migrating our dashboards.

Closes TLON-6076

## Changes

1. Extend the crash type with richer semantics and destructure the crash stacktrace, which by convention contains a crash tag, such as %poke-ack or %watch-ack.
2. Implement OTEL compatible log thread.
3. Augment logs agent api with an otel service configuration. This is going to be configured by hosting.

## How did I test?

Tested the setup on a hosted ship.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.


desk/lib/logs.hoon
```diff
@@ -1,46 +1,67 @@
 /-  *logs
 =<
-  |_  [our=ship =wire]
+  |_  [our=ship dap=term =wire]
```
Fang-: Arguably this should just take the `bowl:gall` so that we have everything we could want and won't have to change the signature. (Or at least, not as frequently.)

desk/lib/logs.hoon
```diff
@@ -1,46 +1,67 @@
...
   ::
   ++  fail
-    |=  [desc=term trace=tang =log-data]
+    |=  [vol=volume =echo =tang =log-data]
```
Fang-, line 6: Shouldn't `+fail` always be considered `%crit` or `%error` or `%fatal` or w/e is the heaviest volume?

desk/sur/logs.hoon
```diff
@@ -3,36 +3,42 @@
...
+::  %set-volume: global logging threshold
 ::
 +$  a-log
   $%  [%log event=log-event data=log-data]
-      [%set-posthog vol=(unit volume)]
+      [%set-volume vol=(unit volume)]
+      [%set-otel url=(unit @t)]
   ==
+++  v1  .
+++  v0
```
Fang-: Do we actually use the old version anywhere? Logs seemed like the kind of internal infra we could get away with not versioning.

desk/lib/logs.hoon
mikolajpp: As discussed off-band, we also report remote crashes through fail.

desk/sur/logs.hoon
mikolajpp: Cleaned up a little: moved the only required conversions to main agent file near the state migration.

Fang- (approved):
Changes look good, thank you!

PR #6035, opened 2026-07-01
Share bot e2e harness between openclaw and hermes

patosullivan:
## Summary

Fixes TLON-5930

Adds a shared `@tloncorp/tlon-bot-e2e` harness for running common bot E2E scenarios against Hermes and OpenClaw. Wires the shared runtime into CI and migrates baseline coverage so both bot runtimes exercise the same core Tlon behaviors.

## Changes

- Added `packages/tlon-bot-e2e` with a shared CLI runner, Docker Compose runtime, Hermes and OpenClaw drivers, fake model server, Tlon actor helpers, diagnostics, scenario DSL, and unit tests
- Added common E2E scenarios for connectivity, owner and allowlisted DM replies, unauthorized DM filtering, Tlon tool-call loops, owner-listen channel behavior, settings persistence, and loop protection
- Added hardened runtime handling for scrubbed compose environments, package-local `.env` allowlisting, dynamic and fixed port allocation, preflight port checks, deterministic fake ship credentials, and failure diagnostics
- Added Hermes shared E2E support with dedicated compose files, E2E Dockerfile and entrypoint, generated Hermes config, fake-model wiring, package smoke coverage, and fake ship cache support
- Migrated OpenClaw shared coverage and package-local integration cases onto the new harness, with `test:integration:shared` running the shared common suite plus OpenClaw `test/cases/*.test.ts`
- Updated Hermes adapter parsing to handle BotProfile-shaped authors in channel and DM events
- Wired CI through a new Hermes workflow, updated OpenClaw CI to use the shared `test:integration:shared` replacement path, and excluded `packages/tlon-bot-e2e` from the main app CI path filter
- Updated package docs, env examples, lockfile entries, and `.gitignore` for the shared E2E workflow

## How did I test?

- Ran `pnpm --filter @tloncorp/tlon-bot-e2e test:unit`
- Ran `pnpm --filter @tloncorp/tlon-bot-e2e tsc`
- Ran `pnpm --filter @tloncorp/tlon-bot-e2e test:e2e`, covering the full shared common suite against both OpenClaw and Hermes
- Ran `pnpm --filter @tloncorp/tlon-bot-e2e test:e2e:openclaw:package`
- Ran `pnpm --filter @tloncorp/tlon-bot-e2e test:e2e:hermes:package`
- Ran focused OpenClaw and Hermes adapter checks during the cleanup and review passes
- After wiring `test:integration:shared` to the full OpenClaw shared replacement path, reran shared harness unit tests, typecheck, and diff validation
- Ran `git diff --check`

## Risks and impact

## Rollback plan

Revert


packages/tlon-bot-e2e/vitest.e2e.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  cacheDir: '../../node_modules/.vite/tlon-bot-e2e-e2e',
```
arthyn: extra `-e2e`?

packages/tlon-bot-e2e/src/tlon/actor.ts
```diff
@@ -0,0 +1,647 @@
...
+          },
+        },
+      },
+    };
+
+    await this.urbit.poke({
+      app: 'chat',
+      mark: 'chat-dm-action',
+      json: action,
+    });
```
arthyn: any reason we can't use the api call for this like we do in `sendChannelPost`?

packages/tlon-bot-e2e/src/tlon/actor.ts
```diff
@@ -0,0 +1,647 @@
...
+  return Math.random().toString(36).slice(2, 8);
+}
+
+function requireEnv(name: string): string {
+  const value = process.env[name];
+  if (!value) {
+    throw new Error(`Missing required environment variable: ${name}`);
+  }
+  return value;
+}
```
arthyn: probably worth making as utils

arthyn (approved):
I didn't look through everything but I think generally this is g2g, left a few comments but no blockers.

here's what claude had to say about the architecture

```
Architecture notes (bigger picture)

  The overall shape here is right: a driver interface + shared scenarios run against both runtimes, backed by a scriptable fake model and a real Tlon
  ship, is the correct way to get behavioral parity between Hermes and OpenClaw. These are the notes I'd want us to address so the harness is
  trustworthy long-term — none of them are blockers on the concept, just on its durability.

  1. Keep "shared" genuinely shared. Runtime-specific details are already leaking into the shared layer (e.g. model.ts hard-codes OpenClaw's [OpenClaw
  heartbeat poll] string; drivers carry runtime-specific config/log-marker assertions). Fine at two runtimes, but this is how the suite rots into
  "mostly shared + a pile of if-hermes/if-openclaw." Push these quirks behind the driver interface (e.g. a driver-provided isBenignCall predicate)
  rather than into common code.

  2. Stop reimplementing infrastructure by hand. We've hand-rolled port allocation (racy), .env parsing (regressed), HTTP-poll waiters (can hang with
  no timeout), and docker compose JSON parsing (version-fragile). Most of the review's confirmed bugs are this class. A conformance suite people are
  supposed to trust can't be this bespoke underneath — every hand-built piece is a flake source, and flaky parity tests get muted. Lean on proven
  tooling (testcontainers, dotenv, wait-on/undici-with-signal).

  3. Read authoritative state, don't reconstruct it. The log-tail-window races and the fabricated-post-id logic reconstruct facts (readiness, post IDs)
  from log scraping and local timestamps instead of reading ship state. This is the same cross-ship-sync timing fragility our web E2E suite already
  fights. Prefer subscribing to/querying real state over polling logs and deriving IDs locally.

  4. De-duplicate config generation. dev/write_config.py and dev/e2e/write_config.py are two renderers for the same config and have already drifted.
  The E2E variant should be derived from the dev one, not forked.

  5. Snapshot/restore real isolation state. isolation.ts restores settings to a hand-maintained defaultSettingValue map (already missing a key). It
  silently goes stale whenever the bot adds a setting. Snapshot-and-restore actual state instead of restoring to a hardcoded mirror.

  6. Confirm CI cost is justified. 8vcpu / 35-min, builds a bun binary, installs workspace deps, boots multiple ships + fake model, runs the common
  suite against both drivers per PR. The path-filter scoping is good; just want to confirm the parity coverage warrants running the full suite ×2 on
  every relevant change.

  Bottom line: solid foundation, correct seams, worth building on — but the harness needs the same reliability bar as production code. Priorities: (a)
  replace hand-rolled runtime plumbing with proven libraries, (b) keep runtime-specific logic behind the driver interface, (c) de-dupe the config
  renderers.
```

PR #6038, opened 2026-07-01
fix(ios): native crash when attaching a video (SDK 56 image-manipulator)

janicduplessis:
## Summary

Attaching a video hard-crashes the app (native Swift fatalError, not catchable in JS) on SDK 56. The crash happens in the poster-generation flow when the expo-video thumbnail is handed to `ImageManipulator.manipulate()`. One-line ordering fix in our existing `expo-image-manipulator` patch.

## Changes

`videoPreviewData.native.ts` passes a `VideoThumbnail` (a `SharedRef<UIImage>`, per expo-video's API) to `manipulate(source: Either<URL, SharedRef<UIImage>>)`. This is type-correct, but crashes in expo's argument converter:

1. `DynamicEitherType.cast` probes branches in declared order under `try?`, expecting failures to throw
2. The `URL` branch converts the JS value via `getAny()` — a deprecated, *asserting* accessor that recursively walks the object's properties
3. On a shared-object instance that walk hits a function-valued property → `FatalError.unimplemented()` — a Swift **trap**, not a thrown error — so `try?` can't rescue it and the process dies before the `SharedRef` branch is ever tried

The patch swaps the probe order to `Either<SharedRef<UIImage>, URL>`. Shared refs now match directly without touching `getAny()`; string/URL sources fail the SharedRef branch with a *catchable* `NativeSharedObjectNotFoundException` and fall through to `URL` exactly as before, so all existing callers (including legacy `manipulateAsync(uri)`) are unchanged. The function body needs no changes — its `source.get()` lookups are by type, not by position.

Upstream `expo/expo` main still has the same signature and the same trapping converter (checked today); this patch hunk carries us until an upstream fix lands. iOS builds `expo-image-manipulator` from source (`expo.autolinking.ios.buildFromSource`), so the Swift patch takes effect. Android has the same ordering (`EitherOfThree<Uri, ...>`) but a separate converter implementation, no observed crash, and isn't built from source — deliberately left alone.

## How did I test?

- Reproduced on the iOS simulator (real ship): attaching a video crashed with `EXC_BREAKPOINT` in `DynamicEitherType.cast` → `JavaScriptValue.getAny()` (crash report symbolicated, poster flow confirmed as the trigger)
- With the patch: same flow — video attaches, poster/thumbnail generates, upload completes, no crash
- `pnpm install --frozen-lockfile` green (`allowNonAppliedPatches: false`, so the patch is verified to apply strictly); the pre-existing orientation-transformer hunk is intact

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: native image manipulation (resize on upload, video posters) on iOS

Worst case is a regression in the other `manipulate()` callers (string-URI paths); the typed fall-through behavior above is why that's not expected, and image upload (which resizes via the same module) was exercised in testing.

## Rollback plan

Revert the commit and `pnpm install` (restores the previous patch file and lockfile hash). No data or schema impact.

## Screenshots / videos

No UI change — fixes a hard crash on video attach.

### DIFF
--- a/patches/README.md
+++ b/patches/README.md
@@ -195,6 +195,13 @@ container, and we confirm the Android repro no longer needs the local fix.
 
 ## expo-image-manipulator@56.0.19
 
+This patch carries two independent iOS hunks.
+
+Local patch:
+`patches/expo-image-manipulator@56.0.19.patch`
+
+### Orientation normalization (HDR HEIC uploads)
+
 Why:
 Expo's iOS orientation transformer normalizes images by manually creating a
 `CGContext` with the source image's bit depth and color space, but with a
@@ -209,9 +216,6 @@ Skips orientation normalization when `UIImage.imageOrientation` is already
 UIKit choose a supported backing context while still applying orientation and
 mirroring before the requested resize runs.
 
-Local patch:
-`patches/expo-image-manipulator@56.0.19.patch`
-
 Upstream:
 - no matching Expo upstream fix found as of June 2026
 - related public reports: `Expensify/App#81702`,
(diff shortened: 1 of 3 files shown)


dnbrwstr (approved):
Lgtm

PR #6042, opened 2026-07-02
pioneer: test runner fixes

mikolajpp:
## Summary

Of the two problems in the pioneer PR: (1) poking with cages rather than pages and (2) having to clam thread results, only (1) turned out to be solvable. (2) is blocked by some annoying hoon/clay bug which makes it impossible to use marks with a vase sample (as would by required to capture thread results).

_Important: when merging the destination branch into develop, care must be taken to adjust the pill in the unified test runner; there likely won't be a merge conflict on that line._

## Changes

1. Add missing %spider-start, %spider-stop marks. They should be upstreamed, but for now we content ourselves with vendoring them, given that we have full control over the test pill and can patch base easily.
2. Update guard to account for new marks
3. Remove stray aqua docs

## How did I test?

Tests pass.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.
## Screenshots / videos

*picture of developer banging hoon compiler*


arthyn (approved):
lgtm

PR #6047, opened 2026-07-02
notes: migrate from original repo into %groups

arthyn:
Fixes TLON-6113

## Changes

- Import the arthyn/notes shared-notebook Gall backend into the `%groups` desk: `app/notes` + `notes-ui`, `sur/notes`, `sur/mcp-proxy`, `lib/notes-*`, `mar/notes/*`, `mar/mcp-proxy-action`, `tests/app/notes`.
- Register `%notes` in `desk.bill`.
- Source matches notes-prototype's `kelvin-408k` branch; the only source adaptation is this desk's patched 3-arg `verb` (`%^ verb | %warn`).
- Reuse existing shared marks (`json`, `html`, `group/channel-{join,leave}`) and vendored deps instead of the notes repo's copies.
- Add opt-in self-poke draining to `lib/test-agent` (`do-poke-drain` / `do-agent-drain` + `drain-self-pokes`/`drain-loop`). The notes agent routes local actions through an async self-poke, so its tests route through the draining variants; the shared `do-poke`/`do-agent` stay non-draining so other agents' tests are unaffected.

## How did I test?

Built and committed the assembled `%groups` desk on a kelvin-408 dev ship (~sidwyn): clean build, no compile errors. Ran the notes agent tests (`/tests/app/notes`) — 83/83 pass — and the full `/tests/app` suite with no regressions.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: new backend agent (`%notes`); additive, no changes to existing agents. Shared `lib/test-agent` change is additive (opt-in variants).

## Rollback plan

Revert the commit / drop `%notes` from `desk.bill`. The change is purely additive to the desk (a new agent plus opt-in test-harness arms) and does not touch existing agents or their state.

## Screenshots / videos

n/a


scripts/assemble-desk.sh
```diff
@@ -35,8 +35,12 @@ fi
...
 # 1. vendored deps first, clearing anything stale in the target
 rsync -aL --delete desk-deps/ "$target/"
-# 2. our own source on top (wins on overlap)
-rsync -aL desk/ "$target/"
+# 2. our own source on top (wins on overlap). Excludes:
+#   - app/notes-ui/  the raw single-file HTML app kept in-repo only as an
+#     editable source; it's hand-copied into lib/notes-ui.hoon (+index), which
+#     is what %notes actually serves. It's not desk source, so keep it off the
+#     assembled desk.
+rsync -aL --exclude 'app/notes-ui/' desk/ "$target/"
```
mikolajpp: I don't understand the need for this. Why not store the html file directly in desk in a dedicated html file and import it where it needs to be used?

scripts/build-notes-openapi.sh
```diff
@@ -0,0 +1,62 @@
...
+TMP_JSON=$(mktemp -t notes-openapi-XXXX).json
+trap 'rm -f "$TMP_JSON"' EXIT
+node -e 'const y=require("js-yaml"),fs=require("fs");process.stdout.write(JSON.stringify(y.load(fs.readFileSync(process.argv[1],"utf8")),null,2)+"\n")' "$YAML" > "$TMP_JSON"
+
+if grep -q "'''" "$TMP_JSON"; then
+  echo "error: generated JSON contains a triple-quote ('''), which would break the cord wrapper" >&2
+  exit 2
+fi
+
+awk -v json_file="$TMP_JSON" '
```
mikolajpp: plus for using awk ;-)

scripts/build-notes-ui.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh
+# Regenerate the +index arm of desk/lib/notes-ui.hoon from the current
+# desk/app/notes-ui/index.html. The +manifest, +service-worker, +favicon-svg,
+# and +icon-svg arms are hand-edited and left untouched.
+set -e
```
mikolajpp: I can't see any actual meaningful processing going on here. The file should just be in the desk and imported in an arm.

desk/sur/notes.hoon
```diff
@@ -0,0 +1,634 @@
+::  notes: shared notebook surface types
+::
+|%
++$  role
+  ?(%owner %editor %viewer)
+::
+::  notebook visibility: private (default) rejects joins from non-members;
+::  public allows anyone to join.
++$  visibility
```
mikolajpp: Something is going on with inverted flag comments. I used to do them in some cases, but @Fang- we generally use the flag convention unless in specific cases.

desk/sur/mcp-proxy.hoon
```diff
@@ -0,0 +1,44 @@
+::  mcp-proxy: vendored type subset for the helper-poke that
+::  registers %notes with the local mcp-proxy.
+::
+::  Only the action variants we send (%add-server, %refresh-spec)
+::  and the mcp-server record they reference are kept. Full type
+::  source: github.com/tloncorp/mcp desk/sur/mcp-proxy.hoon
+::
```
mikolajpp: We have peru now. Why not just vendor the whole file that way?

desk/mar/notes/action-1.hoon
```diff
@@ -0,0 +1,17 @@
+::  notes-action-1: HTTP / request-id wrapped action
+::    POST body JSON: {"requestId": "0v...", "action": <a-notes JSON>}
+::
```
mikolajpp:
```suggestion
::
```
I think agents are more than capable of getting the json shape from json conversion library, we shouldn't need such ad-hoc comments.

desk/mar/notes/response-1.hoon
```diff
@@ -0,0 +1,18 @@
+::  notes-response-1: subscriber → client response.
+::    Faceted on /v1/notes/~ship/name/request/<uv> and returned as the
+::    body of POST /notes/~/v1 (or its GET /request/<uv> follow-up).
```
mikolajpp: This information is properly described in the agent.

desk/lib/notes-openapi.hoon
```diff
@@ -0,0 +1,2785 @@
...
+::  ++json body is regenerated from docs/openapi.yaml by
+::  scripts/build-notes-openapi.sh — edit the YAML, not this file.
+::  (Stored as JSON because %mcp-proxy parses the cached spec with
+::   de:json:html and doesn't accept YAML.)
+::
+|%
+++  json
+^-  @t
+'''
+{
```
mikolajpp: As we talked, this can probably be a real json file.

desk/lib/notes-share.hoon
```diff
@@ -0,0 +1,124 @@
+^-  @t
+'''
```
mikolajpp: And this can be a html file.

desk/lib/notes-ui.hoon
```diff
@@ -0,0 +1,6325 @@
+::  notes-ui: HTML/CSS/JS + PWA static assets bundled for the agent
+::
+|%
+::
+::  +index: full single-file HTML app served at /notes/ui and as the
+::  fallback for any non-asset /notes/* URL.
+++  index
+^-  @t
+'''
+<!DOCTYPE html>
```
mikolajpp: As mentioned elsewhere, we should simply import the html file from the desk.

desk/lib/test-agent.hoon
```diff
@@ -234,12 +234,84 @@
   ;<  ~              bind:m  (set-scry-gate old-scry)
   (pure:m c)
 ::
+::  +drain-self-pokes: re-deliver self-directed %poke cards through on-poke,
+::  recursively until fixpoint or depth=16.  Returns (each [...] tang).
+::  A "self-poke" matches [%pass * %agent [our.bowl dap.bowl] %poke *].
+::  pending: cards to scan for self-pokes in this round
+::  acc:     all cards collected so far (grows monotonically)
+::  st:      current harness state
+++  drain-self-pokes
```
mikolajpp: It is hard to understand what is going on here. Why do we ever need more than 2 hops if we proxy pokes? (The case being a local poke getting proxied and reinjected, if that's what is happening here.)

desk/lib/notes/json.hoon
```diff
@@ -0,0 +1,704 @@
+::  lib/notes-json: JSON encoding/decoding for notes types
+::
+/-  n=notes
+|%
+::  +da-to-unix: convert @da to unix seconds
+++  da-to-unix
+  |=  da=@da
+  ^-  @ud
+  (div (sub da ~1970.1.1) ~s1)
```
mikolajpp: I am astonished this is not in zuse, but I searched and could not find anything... @Fang-

desk/lib/notes/json.hoon
```diff
@@ -0,0 +1,704 @@
...
+    ^-  json
+    %-  pairs
+    :~  ['id' (numb id.nb)]
+        ['title' s+title.nb]
+        ::  rootFolderId is deterministically id+1 (set by se-create-notebook)
+        ['rootFolderId' (numb +(id.nb))]
+        ['createdBy' s+(scot %p created-by.nb)]
+        ['createdAt' (numb (da-to-unix created-at.nb))]
+        ['updatedAt' (numb (da-to-unix updated-at.nb))]
+        ['updatedBy' s+(scot %p updated-by.nb)]
```
mikolajpp: Here and elsewhere we stray from established convention of doing 
```
'id'^(numb id.nb)
```

desk/lib/notes/json.hoon
```diff
@@ -0,0 +1,704 @@
...
+          %created
+        :~  ['type' s+'notebook-created']
+            ['notebook' (notebook notebook.upd)]
+            ['visibility' s+(scot %tas visibility.upd)]
+        ==
+          %updated
+        :~  ['type' s+'notebook-updated']
+            ['notebook' (notebook notebook.upd)]
+        ==
+          %deleted
```
mikolajpp, line 471: This was machine generated, but then again it was machine generated so it should be _easy_ to follow whatever conventions we have regarding comment line breaks etc. :-)

desk/lib/notes/json.hoon
```diff
@@ -0,0 +1,704 @@
...
+        ==
+          %note
+        :~  ['type' s+'note-update']
+            ['noteUpdate' (u-note id.upd u-note.upd)]
+        ==
+      ==
+    --
+  --
+::
+::  +dejs: decode JSON to notes types
```
mikolajpp: I see inverted flags comments everywhere...

mikolajpp (changes requested):
Left some minor comments. The only file left to review is the main agent file. Meanwhile I have also the comments from the first round of a review (still back at the notes repo): https://gist.github.com/mikolajpp/5bc4a8009a29b659e4aa40c590d2b008

desk/sur/notes.hoon
arthyn: Flipped the inverted comment blocks to the flag form (description, then a bare `::` above the code) across sur/notes, lib/notes/json, app/notes, and tests/app/notes. (7614e1797)

desk/mar/notes/action-1.hoon
arthyn: Dropped the ad-hoc JSON-shape comment. (7614e1797)

desk/mar/notes/response-1.hoon
arthyn: Removed — it's described in the agent. (7614e1797)

desk/lib/notes/json.hoon
arthyn: Switched the enjs pairs to the `'key'^value` form throughout json.hoon. (7614e1797)

desk/lib/notes/json.hoon
arthyn: Tightened the comment. (7614e1797)

desk/lib/notes/json.hoon
arthyn: Flipped throughout the notes files — see the sur/notes.hoon thread. (7614e1797)

desk/lib/notes/json.hoon
arthyn: Couldn't find a zuse equivalent either — leaving `da-to-unix` as-is for now; happy to swap if @Fang- knows one.

desk/lib/test-agent.hoon
arthyn: I think its just trying to emulate what gall does when poking yourself

desk/app/notes.hoon
```diff
@@ -0,0 +1,2818 @@
...
+    ^-  ?
+    ?~  grp=group.notebook-state
+      !=(~ (~(get by members.notebook-state) who))
+    (group-can-read u.grp flag who)
+  ::
+  ++  se-can-edit
+    |=  who=ship
+    ^-  ?
+    =/  r=(unit role:n)
+      (~(get by members.notebook-state) who)
```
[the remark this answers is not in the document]
arthyn: Good catch. Fixed in 0c38744dd: se-can-edit now re-checks the group's live can-read for group-mode notebooks (in addition to the members-map role), mirroring se-can-view — a revoked member can no longer write even though the members map isn't pruned. Added a regression test (test-group-edit-revoked-rejects).

desk/lib/notes/ui.html
```diff
@@ -0,0 +1,6249 @@
...
+  if (!n) return;
+  const title = document.getElementById("note-title-input").value.trim() || "Untitled";
+  const body = document.getElementById("editor").value;
+  const nb = notebooks[n.notebookId];
+  const isLocal = nb && SHIP && nb.host.replace(/^~/, "") === SHIP.replace(/^~/, "");
+  const sentRev = isLocal ? savedRevision : 0;
+  const items = [];
+  const updateAction = { "update-note": { notebookId: n.notebookId, noteId: activeNoteId, bodyMd: body, expectedRevision: sentRev } };
+  if (activeNotebookFlag) updateAction._flag = activeNotebookFlag;
+  items.push({ id: msgId++, action: "poke", ship: SHIP.replace("~",""), app: "notes", mark: "notes-action", json: updateAction });
```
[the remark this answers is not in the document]
arthyn: Fixed in 0c38744dd: flushSaveBeacon now builds the current type-tagged a-notes actions (via the pokeNoteAction shape) and POSTs them to /notes/~/v1 with keepalive, instead of the stale {"update-note": ...} channel poke. The old path also had wrong field names (bodyMd vs body), so unload saves were being dropped.

desk/lib/notes/json.hoon
Fang-: This is `+unt:chrono:userlib`. (;

desk/app/notes/favicon.svg
Fang-, line 1: I think static files, ui logic etc, you generally want to put under `/app/notes/` rather than `/lib/notes/`. At least, it'd feel more at home there. Under `/lib` it suggests it might be useful to outside callers or otherwise be part of the public api, too.

desk/sur/notes.hoon
```diff
@@ -0,0 +1,619 @@
...
++$  notebook-state-13
+  $:  =notebook
+      =members
+      =visibility
+      folders=(map @ud folder)
+      notes=(map @ud note)
+      history=(map note-id=@ud (list note-revision))
+  ==
+::  Actions (client → agent)
+::  ============================================================
```
Fang-:
```suggestion
::  actions (client -> agent)
::  
```

I know the tools like these things (big separators and such) but they're not considered good hoon style. If we don't want this stuff to proliferate we should probably avoid having any of it end up in our mainline branches, lest it start taking it as examples.

(There's a similar risk there with the LLM-y hoon style in general, where the more of it we accept into our codebase, the more entrenched it becomes. These things often try (but not always succeed, obviously) to mimic local style, and as the local codebase gains more generated code, the generated style will overtake everything else...)

desk/lib/test-agent.hoon
```diff
@@ -234,12 +234,84 @@
...
   %-  do
   |=  s=state
   (~(on-poke agent.s bowl.s) mark vase)
 ::
+::  +do-poke-drain: like +do-poke, but simulates Gall's move loopback by
+::  re-delivering any self-directed %poke cards (see +drain-self-pokes) until
+::  fixpoint. Use for agents that route local actions through a self-poke and
+::  only mutate state on the redelivered event (e.g. %notes). Plain +do-poke
+::  leaves those self-pokes inert, so callers that inspect state must use this.
+++  do-poke-drain
```
Fang-: As mentioned out of band, this isn't safe/sane to do in the general case: move/effect ordering may be different than it is in practice.

Perhaps here (and most of the time) that's fine and not important, but I can't condone it as an additional to `/lib/test-agent` proper. If it starts seeing usage elsewhere it may mask subtle ordering bugs, giving false confidence in hard-to-find areas.

You're welcome to keep this approach for the notes test, but I'd move these helpers to be inside of `/tests/notes` instead.

desk/app/notes.hoon
```diff
@@ -0,0 +1,2825 @@
...
+  =?  old  ?=(%5 -.old)  (state-5-to-6 old)
+  =?  old  ?=(%6 -.old)  (state-6-to-7 old)
+  =?  old  ?=(%7 -.old)  (state-7-to-8 old)
+  =?  old  ?=(%8 -.old)  (state-8-to-9 old)
+  =?  old  ?=(%9 -.old)  (state-9-to-10 old)
+  =?  old  ?=(%10 -.old)  (state-10-to-11 old)
+  =?  old  ?=(%11 -.old)  (state-11-to-12 old)
+  =?  old  ?=(%12 -.old)  (state-12-to-13 old)
+  =?  old  ?=(%13 -.old)  (state-13-to-14 old)
+  ?>  ?=(%14 -.old)
```
Fang-: Since this is coming in fresh and the current install base is pretty small, you could consider trimming this down and just accepting state-14 (or states 12-14 or w/e) only, to help clean this up a little bit.

desk/app/notes.hoon
```diff
@@ -0,0 +1,2825 @@
...
+      %handle-http-request
+    ::  Normalize src.bowl to our.bowl for the rest of the http path.
+    ::  Eyre sets src to a synthetic '<random>--<host>' guest @p for
+    ::  unauthenticated browser sessions, which then fails downstream
+    ::  permission checks like +se-is-owner even when the caller has a
+    ::  valid X-Api-Key (or eyre session). The api-key IS the host's
+    ::  capability and a valid session IS into our ship, so once the
+    ::  request authorizes inside +handle-v1-post the effective actor
+    ::  is our.bowl regardless of which @p eyre put in src.
+    =.  bowl  bowl(src our.bowl)
```
Fang-: Crazy that it doesn't just write `=.  src.bowl  etc`.

If that comment's the case, shouldn't we do this at the point where we verify the authentication? A line like this, this early, unconditionally, feels kinda risky.

desk/app/notes.hoon
```diff
@@ -0,0 +1,2825 @@
...
+    ?:  =(our.bowl ship.flag)  cor
+    ?.  (~(has by books) flag)  cor
+    =/  rid=request-id:v1:n  `@uv`(mix eny.bowl rid-counter)
+    =.  rid-counter  +(rid-counter)
+    (leave-remote-v1 rid flag)
+  ==
+  ::  no |^ arms remain — the invite/join helpers used to live here but
+  ::  moved to the top level so +dispatch-v1-action (called by both this
+  ::  arm and +handle-v1-post) can reach them.
+  ::
```
Fang-: LLMs are great at leaving these kind of planning/history details in that don't really add anything to the code... I don't think this is important context for reading the code, is it?

desk/app/notes.hoon
```diff
@@ -0,0 +1,2825 @@
...
+    ?:  =(%'POST' method)  (handle-v1-post eyre-id inbound-request)
+    (http-error eyre-id 405 'method not allowed')
+  ?:  =("/notes/~/v1/request/" (scag 20 url-path))
+    ?:  =(%'GET' method)  (handle-v1-get-request eyre-id (slag 20 url-path) inbound-request)
+    (http-error eyre-id 405 'method not allowed')
+  ?:  =("/notes/~/v1/" (scag 12 url-path))
+    ::  other /notes/~/v1/* — GET reads, POST/PATCH/DELETE first-class writes
+    ?:  =(%'GET' method)     (handle-v1-read eyre-id url-path inbound-request)
+    ::  vere's runtime HTTP server rejects PATCH (400 before reaching the
+    ::  agent), so note-body updates use PUT.
```
Fang-, line 272: Crazy, but correct. There's a trivial `// TODO ??` in http.c, can't see any reason not to support `PATCH` there...

desk/app/notes.hoon
```diff
@@ -0,0 +1,2825 @@
...
+  ::  step 6: prefix "n-" if first char is a digit
+  =/  prefixed=tape
+    ?.  &(?=(^ base) (gte i.base '0') (lte i.base '9'))
+      base
+    (weld "n-" base)
+  ::  step 7: build suffix string (strip dots from scot %ud)
+  =/  raw-suf=tape  (trip (scot %ud suffix))
+  =/  suf-tape=tape
+    %+  skim  raw-suf
+    |=(c=@t !=(c '.'))
```
Fang-: Very funny.

```suggestion
  =/  suf-tape=tape  (a-co:co suffix)
```

Fang- (commented):
Some comments. Not exhaustive, but this is yuge, proof is in the used functionality of it I suppose.

desk/lib/notes/json.hoon
arthyn: Thanks! Swapped da-to-unix for `unt:chrono:userlib` — confirmed it's the seconds variant (byte-identical), `unm` being the ms one. Done in the latest commit.

desk/app/notes/favicon.svg
arthyn: Moved the served static files (ui.html, share.html, openapi/manifest/sw/favicon/icon) to app/notes/; kept the json codec lib in lib/notes/. (0d4264940)

desk/sur/notes.hoon
arthyn: Removed the ==== banner separators across sur/notes.hoon and app/notes.hoon. (0d4264940)

desk/lib/test-agent.hoon
arthyn: Agreed — reverted lib/test-agent.hoon to pristine and moved the self-poke draining (drain-self-pokes/loop, do-poke-drain/agent-drain) into tests/app/notes.hoon, since it's only sound for %notes. (0d4264940)

desk/app/notes.hoon
arthyn: Switched to `=. src.bowl our.bowl` and moved it to after the auth check in handle-v1-post/handle-v1-write; reads already use our.bowl explicitly. (0d4264940)

desk/app/notes.hoon
arthyn: Removed. (0d4264940)

desk/app/notes.hoon
arthyn: Ha — left the note as-is; agreed it'd be nice if vere's http.c supported PATCH. Using PUT for now.

desk/app/notes.hoon
arthyn: Done — `(a-co:co suffix)`. (0d4264940)

desk/app/notes.hoon
```diff
@@ -0,0 +1,2825 @@
...
+    ^+  se-core
+    ?>  ?=(%note -.c-notebook.cmd)
+    ?>  ?=(%delete -.a-note.c-notebook.cmd)
+    ?>  (se-can-edit src.bowl)
+    =*  nid  id.c-notebook.cmd
+    =/  =note:n
+      (~(got by notes.notebook-state) nid)
+    =.  notes.notebook-state
+      (~(del by notes.notebook-state) nid)
+    (se-update [%note nid [%deleted ~]])
```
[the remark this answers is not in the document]
arthyn: Fixed in 997e3361b: se-delete-note now removes the note's `published` entry, and the recursive folder-delete path does the same for each removed note.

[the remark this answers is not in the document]
arthyn: Fixed in 997e3361b: se-delete-note clears the note's archived `history` (recursive folder-delete clears it per removed note too). Added test-delete-note-clears-history.

desk/app/notes.hoon
```diff
@@ -0,0 +1,2825 @@
...
+    ::  clean up published entries for this notebook
+    =.  published
+      %-  malt
+      %+  skip  ~(tap by published)
+      |=  [k=[=flag:n note-id=@ud] v=@t]
+      =(flag.k flag)
+    ::  group affiliation, history and visibility live in notebook-state,
+    ::  deleted with the book via the gone flag
+    =.  se-core  (se-update [%deleted ~])
+    se-core(gone &)
```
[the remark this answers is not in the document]
arthyn: Fixed in 997e3361b: group-mode notebook delete now pokes %groups with a `%channel … %del` (mirror of the %add on create). Added test-delete-group-notebook-removes-channel.

desk/app/notes.hoon
```diff
@@ -0,0 +1,2815 @@
...
+::    - %ok / %no-change: drop after 5m
+::    - %error: drop only after the client has fetched it
+::
+++  cleanup-requests
+  |=  now=@da
+  ^-  requests:v1:n
+  %-  ~(rep by requests)
+  |=  [[id=request-id:v1:n req=incoming-request:v1:n] out=requests:v1:n]
+  ?:  |(?=(~ result.req) ?=([~ %pending *] result.req))
+    (~(put by out) id req)
```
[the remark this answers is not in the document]
arthyn: Fixed in 997e3361b: finalize-pending stamps final-at at the timeout, and cleanup-requests no longer keeps %pending unconditionally — dead pending requests age out at 24h.

mikolajpp (commented):
I had previously reviewed everything except the agent file. This round, I have taken a look solely at the agent code. The agent is rather large: it maintains several api surfaces: a REST api, and standard gall apis for client/server components.

Overall the code seems to possess a structure, no doubt resulted from judicious steering. There are many places with rather dubious Hoon patterns, unconventional code etc. Some of them were already addressed. At the end of this comment I enclose some things that caught my attention while skimming through code.

I have not checked: internal logic of the application, the coherence of external api surfaces and other such things we typically do during a full PR review, but which require a solid mental model of the code. To do so would require significantly more time and effort than we seem to have presently.

## Review notes:
   - Some issues from my previous notes remain: for instance `$c-cmd` carries the flag and is accepted as argument to arms, which already have flag in core context. 
   - There is a really bad pattern of dispatch arms asserting on command variant. They should only receive the arguments of a particular variant they execute. See `+se-create-folder` for instance.
   - `+serve-http` should not route by raw string prefixes / hardcoded `scag`/`slag` lengths.
 Use zuse URL parsers (`de-purl:html`, e.g. `apat`/`yque`) once, then pattern match on parsed
 path segments and query args.
   - For REST JSON field extraction, after `get`ing a field from the JSON map,
 we should use standard JSON value decoders (`so`, `ni`, etc.) rather than manual tag checks and parsers.
   - We should not introduce production workarounds for `test-agent` behavior. If `test-agent` does not advance
 `eny`, that should be flagged/fixed in tests, not accommodated in app logic.
   - `+se-create-folder` should not design core behavior around caller mistakes. (see comment therein.) If omitted
 `parent` is invalid, require it in the command/decoder or default it at the boundary.

desk/app/notes.hoon
arthyn: Done in 6bbf3a52f — collapsed to state-14 as the sole base state. Groups-hosted %notes always inits fresh (channels suspends any standalone %notes desk and force-starts ours via kiln rein), so the whole migration chain was dead code here. Removed state-0..12, all state-N-to-M arms, and the migration-only frozen types.

desk/app/notes.hoon
```diff
@@ -0,0 +1,2802 @@
...
+  =?  old  ?=(%10 -.old)  (state-10-to-11 old)
+  =?  old  ?=(%11 -.old)  (state-11-to-12 old)
+  =?  old  ?=(%12 -.old)  (state-12-to-13 old)
+  =?  old  ?=(%13 -.old)  (state-13-to-14 old)
+  ?>  ?=(%14 -.old)
+  =.  state  old
+  ::  ships migrating from state-11 land here with api-key=~; mint one
+  ::  so the bypass is usable out of the box. Operators can rotate or
+  ::  clear afterward.
+  =?  api-key.state  ?=(~ api-key.state)  `(scot %uv eny.bowl)
```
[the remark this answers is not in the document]
arthyn: Fixed in 6bbf3a52f: removed the load-time api-key mint guard. `+on-init` still mints on fresh install, but at load an empty api-key now means an operator ran `%clear-api-key`, so it stays disabled rather than being silently re-minted. (The guard only existed to backfill the field for state-11 migrants, which no longer exist after the state-14 collapse.)

desk/app/notes.hoon
```diff
@@ -0,0 +1,2798 @@
...
+    ~>  %spin.['state-1-to-2']
+    |=  s=state-1:n
+    ^-  state-2:n
+    [%2 books.s next-id.s ~]
+  ::
+  ++  state-2-to-3
+    ~>  %spin.['state-2-to-3']
+    |=  s=state-2:n
+    ^-  state-3:n
+    [%3 books.s next-id.s ~]
```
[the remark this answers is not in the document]
arthyn: Moot as of 6bbf3a52f: the migration chain was removed entirely (groups-hosted %notes always inits fresh), so state-2→3 no longer exists.

desk/app/notes.hoon
```diff
@@ -0,0 +1,2792 @@
...
+      %+  turn  ~(tap by published.s)
+      |=  [[f=flag-v9:n nid=@ud] html=@t]
+      =/  =flag:n  (fall (~(get by xlat) f) [ship.f `@tas`name.f])
+      [[flag nid] html]
+    =/  new-invites=(map flag:n invite-info:n)
+      %-  malt
+      %+  turn  ~(tap by invites.s)
+      |=  [f=flag-v9:n info=invite-info:n]
+      =/  =flag:n  (fall (~(get by xlat) f) [ship.f `@tas`name.f])
+      [flag info]
```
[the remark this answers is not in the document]
arthyn: Moot as of 6bbf3a52f: the state-9 flag re-key migration was removed with the rest of the chain.

desk/app/notes.hoon
```diff
@@ -0,0 +1,2783 @@
...
+      no-core
+    ==
+  ::  +no-peek: handle per-notebook scry requests
+  ::  kind: the path segment after /v0/ (e.g. %notebook, %notes, %note, etc.)
+  ::  rest: the remainder of the pole after kind/ship/name (typed as *)
+  ::
+  ++  no-peek
+    |=  [kind=@ rest=*]
+    ^-  (unit (unit cage))
+    ?>  ?=(^ (~(get by members.notebook-state) src.bowl))
```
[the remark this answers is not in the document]
arthyn: Fixed in 1d827267a: `+no-peek` now gates reads on `+can-view-flag` (group-aware can-read recheck, guarded by `+group-synced`) instead of the raw members map, mirroring the edit-side `+se-can-edit` fix. Also tightened `+group-can-read`'s self-allow from `our.bowl` to `ship.flag` so a revoked local subscriber is still checked.

desk/app/notes/openapi.json
```diff
@@ -0,0 +1,2773 @@
...
+                  "folderName"
+                ],
+                "properties": {
+                  "folderName": {
+                    "type": "string",
+                    "description": "Name of the new folder. (Distinct from the path's `{name}` which is the notebook slug — when mcp-proxy flattens path + body fields into a single tool input, a colliding name would conflate the two.)"
+                  },
+                  "parent": {
+                    "type": "integer",
+                    "description": "Parent folder id. If omitted (or null), the folder lands under the notebook's root (`notebook.rootFolderId`). An explicit id that doesn't exist in the notebook is rejected."
```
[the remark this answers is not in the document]
arthyn: Fixed in fa584b8b7: `parent` is now marked required and the description says to pass the notebook's root folder id for top-level folders (the server requires it as of the create-folder boundary change).

desk/app/notes/openapi.json
```diff
@@ -0,0 +1,2773 @@
...
+        "properties": {
+          "type": {
+            "type": "string",
+            "enum": [
+              "notebook"
+            ]
+          },
+          "flag": {
+            "$ref": "#/components/schemas/Flag"
+          },
```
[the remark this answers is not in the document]
arthyn: Fixed in fa584b8b7: the generic notebook-action envelope's `flag` is now a `~ship/name` string. (The `{host, flagName}` object form is only used by create-group-notebook.)

desk/app/notes/ui.html
```diff
@@ -0,0 +1,6255 @@
...
+  s = s.replace(/\*\*\*(.+?)\*\*\*/g, "<strong><em>$1</em></strong>");
+  s = s.replace(/___(.+?)___/g,       "<strong><em>$1</em></strong>");
+  s = s.replace(/\*\*(.+?)\*\*/g,     "<strong>$1</strong>");
+  s = s.replace(/__(.+?)__/g,         "<strong>$1</strong>");
+  s = s.replace(/\*(.+?)\*/g,         "<em>$1</em>");
+  s = s.replace(/_(.+?)_/g,           "<em>$1</em>");
+  s = s.replace(/~~(.+?)~~/g,         "<del>$1</del>");
+
+  // Swap tokens back in.
+  return s.replace(/TOK(\d+)/g, (_m, i) => tokens[+i]);
```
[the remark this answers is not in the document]
arthyn: Deferring: the notes web client (`ui.html`) is being extracted from this desk and will be maintained separately. Tracking the remaining web-client correctness/XSS findings in its new home rather than here.

desk/app/notes/ui.html
```diff
@@ -0,0 +1,6255 @@
...
+
+async function loadFolders() {
+  if (!activeNotebookFlag) return;
+  const data = await scry(`/v0/folders/${activeNotebookFlag}`);
+  // Only replace the cache when we got fresh data — otherwise preserve what
+  // we already have so offline/reconnecting users can still walk around.
+  if (data) {
+    folders = {};
+    (data || []).forEach(f => folders[f.id] = f);
+    foldersByNb[activeNotebookFlag] = { ...folders };
```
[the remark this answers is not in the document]
arthyn: Deferring: the notes web client (`ui.html`) is being extracted from this desk and will be maintained separately. Tracking the remaining web-client correctness/XSS findings in its new home rather than here.

desk/app/notes/ui.html
```diff
@@ -0,0 +1,6255 @@
...
+
+  // Strict revision check on the host — always send the real revision; the
+  // stream keeps subscribers' savedRevision current so this is correct on
+  // both host-owned and subscribed notebooks.
+  const sentRev = savedRevision;
+  // Pre-bump savedRevision BEFORE sending so the echoed note-updated event
+  // (which may arrive before pokeAction's ack resolves) is recognized as our
+  // own save via isOwnSaveEcho's revision check.
+  const rollbackRev = savedRevision;
+  savedRevision = savedRevision + 1;
```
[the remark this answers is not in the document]
arthyn: Deferring: the notes web client (`ui.html`) is being extracted from this desk and will be maintained separately. Tracking the remaining web-client correctness/XSS findings in its new home rather than here.

desk/app/notes/ui.html
```diff
@@ -0,0 +1,6255 @@
...
+// Pure echo-detection. Kept pure (no global access) so it's unit-testable;
+// the canonical tests live at tests/frontend/run.html — keep in sync.
+// A note-updated event is our own save echo if:
+//   (a) its revision matches or trails what we already expect after our
+//       optimistic pre-bump in autoSave (covers local-host saves), OR
+//   (b) we issued a save very recently from this ship (covers remote saves
+//       where we can't predict the host's new revision ahead of time).
+function isOwnSaveEchoPure(evt, savedRev, shipId, lastSaveAt, now) {
+  const r = evt && evt.revision;
+  if (typeof r === "number" && r <= savedRev) return true;
```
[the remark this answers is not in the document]
arthyn: Deferring: the notes web client (`ui.html`) is being extracted from this desk and will be maintained separately. Tracking the remaining web-client correctness/XSS findings in its new home rather than here.

desk/app/notes/ui.html
```diff
@@ -0,0 +1,6262 @@
...
+  savedRevision = n.revision;
+  document.getElementById("note-rev").textContent = `rev ${n.revision}`;
+  document.title = `${n.title || "Untitled"} — Notes`;
+  clearDirty();
+  // If a local draft exists from a prior session that was never saved,
+  // restore it. Only restore when it was based on a revision <= the
+  // current server revision (otherwise it's stale or already superseded).
+  const draft = readDraft(activeNotebookFlag, id);
+  if (draft && draft.baseRevision <= n.revision &&
+      (draft.title !== n.title || draft.body !== n.bodyMd)) {
```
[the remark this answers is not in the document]
arthyn: Deferring: the notes web client (`ui.html`) is being extracted from this desk and will be maintained separately. Tracking the remaining web-client correctness/XSS findings in its new home rather than here.

desk/app/notes/ui.html
```diff
@@ -0,0 +1,6262 @@
...
+function clearDirty() {
+  dirty = false;
+  if (saveTimer) { clearTimeout(saveTimer); saveTimer = null; }
+}
+
+async function confirmDirty() {
+  if (!dirty) return true;
+  // Try to save first
+  await autoSave();
+  return true;
```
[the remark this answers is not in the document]
arthyn: Deferring: the notes web client (`ui.html`) is being extracted from this desk and will be maintained separately. Tracking the remaining web-client correctness/XSS findings in its new home rather than here.

desk/app/notes/ui.html
```diff
@@ -0,0 +1,6262 @@
...
+//   note-update   → { type: "note-update",   noteUpdate:   { type: "note-{created|updated|deleted|published|unpublished|history-archived}", ... } }
+
+function applyNotebookUpdate(u) {
+  const type = u.type;
+  if (!type) return;
+  if (type === "notebook-created" || type === "notebook-updated" || type === "notebook-visibility-changed") {
+    loadNotebooks();
+  } else if (type === "notebook-deleted") {
+    // notebookId is now not directly in the update; reload sidebar
+    loadNotebooks();
```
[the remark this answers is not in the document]
arthyn: Deferring: the notes web client (`ui.html`) is being extracted from this desk and will be maintained separately. Tracking the remaining web-client correctness/XSS findings in its new home rather than here.

desk/app/notes/ui.html
```diff
@@ -0,0 +1,6262 @@
...
+  const type = nu.type;
+  if (type === "note-history-archived") {
+    // nu.revision is the archived note-revision object (from u-note %history-archived)
+    if (nu.revision) applyArchivedRevision({ revision: nu.revision, noteId: nu.id });
+    return;
+  }
+  if (type === "note-created" && nu.note) {
+    notifyNoteCreated(nu.note);
+  }
+  loadNotes();
```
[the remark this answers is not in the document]
arthyn: Deferring: the notes web client (`ui.html`) is being extracted from this desk and will be maintained separately. Tracking the remaining web-client correctness/XSS findings in its new home rather than here.

desk/app/notes/ui.html
```diff
@@ -0,0 +1,6262 @@
...
+  }
+}
+
+async function publishNote() {
+  closeOverflow();
+  if (!activeNoteId || !activeNotebookId) return;
+
+  // Save first if dirty
+  if (dirty) await autoSave();
+  if (!(await publishCurrentNoteHtml())) return;
```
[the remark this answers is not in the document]
arthyn: Deferring: the notes web client (`ui.html`) is being extracted from this desk and will be maintained separately. Tracking the remaining web-client correctness/XSS findings in its new home rather than here.

desk/app/notes.hoon
```diff
@@ -0,0 +1,2588 @@
...
+  ==
+::  +load: groups-hosted %notes always initializes fresh (channels suspends
+::  any standalone %notes desk and force-starts ours via kiln rein), so no
+::  pre-existing on-disk save is ever migrated in-place here. state-14 is the
+::  sole base state — decode directly.
+::
+++  load
+  |=  =vase
+  ^+  cor
+  =.  state  !<(state-14:n vase)
```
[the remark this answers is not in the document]
arthyn: Intentional. %notes was alpha software with a very small install base, and most of those installs have already taken the state-14 transition, so we're accepting state-14 as the sole base state rather than carrying the migration chain. Any rare pre-14 standalone save can reset. Won't fix.

PR #6073, opened 2026-07-06
Fix presence expire timers eating refreshed runs

arthyn:
## Changes

- `desk/app/presence.hoon`: embed the timer's `end` date in the `%expire` behn wire (`[%expire (scot %da end) (scot %p ship) topic context]`) at both the host `on-poke %presence-command-1 %set` and subscriber `on-agent %fact %set` scheduling sites.
- `on-arvo` now matches `[%expire @ @ topic *]` and makes the wake **self-validating**: it clears the presence only when the currently-stored presence's end is not later than the fired `end`; superseded wakes (a later `%set` slid the window forward) and wakes for already-gone presences no-op.
- Removed `+cancel-expire` and its two `%clear` call sites — proactive `%rest` is no longer needed now that stale timers validate themselves on fire.

### Why

The old wire was keyed only by `[ship topic context]`, and `%set` never rested the prior timer, so refreshes stacked multiple behn timers on the same wire. `on-arvo` deleted unconditionally, so the earliest timer fired and cleared a still-active presence — the computing/typing indicator flickered off roughly every timeout window during long runs, on both the host-local view and every subscriber. The plugin side (`packages/openclaw`) is designed for a sliding window (90s timeout, re-publish within 30s, 20s keepalive, `maxDurationMs: 0`) and correctly relies on each `%set` extending the ship-side expiry; the bug was entirely backend.

## How did I test?

- Diff reviewed for compile-correctness against `desk/sur/presence.hoon` (wire narrowing, `since.u.pre`/`timeout.u.pre` wing resolution, quip return types, `;: weld` arity). Traced the timer lifecycle by hand across set / refresh / clear / clear-then-reset races.
- ⚠️ Not yet exercised on a running `%groups` ship — needs a manual pass confirming the indicator stays lit through a long multi-tool run and clears promptly on completion.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: `%presence` agent (computing/typing indicators)

## Rollback plan

Revert this commit. The change is confined to `desk/app/presence.hoon`; the `%expire` wire format is internal to the agent (not persisted state or a cross-ship protocol mark), so no migration or coordinated deploy is required.

### DIFF
--- a/desk/app/presence.hoon
+++ b/desk/app/presence.hoon
@@ -139,20 +139,6 @@
   ?:  =(~ tos)  (~(del by places) context)
   (~(put by places) context tos)
 ::
-++  cancel-expire
-  |=  [=places key]
-  ^-  (list card)
-  =/  tos  (~(gut by places) context *topics)
-  =/  pes  (~(gut by tos) topic *people)
-  ?~  pre=(~(get by pes) ship)  ~
-  =/  end=@da
-    %+  add  since.u.pre
-    (fall timeout.u.pre (default-timeout topic))
-  :_  ~
-  :+  %pass
-    [%expire (scot %p ship) topic context]
-  [%arvo %b %rest end]
-::
 ++  give-update
   |=  [subs=(jug context ship) disclose=(set ship) upd=update-1]
   ^-  (list card)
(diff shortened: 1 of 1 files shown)


arthyn:
Closing in favor of #6074 (Bill), which fixes the same root cause — stale `%expire` timers deleting presence that later `%set`s kept alive — with a cleaner approach: a time-based check (`now >= since + timeout`) in the single `on-arvo %expire` handler, so it covers both the host-local and subscriber paths without changing the wire format or touching the `%set` sites. It also ships agent tests and live fakezod verification. Only thing not carried over is removing the now-redundant `+cancel-expire`, which is a trivial follow-up if wanted.

PR #6074, opened 2026-07-06
presence: don't let stale expiry timers delete fresh presence

wca4a:
Fixes TLON-6112

## Summary

Fixes the bot "thinking" indicator bouncing on/off during long runs (and any other presence flicker under periodic keepalives).

**Root cause:** on the subscriber's ship, every incoming `%set` fact arms its own behn expiry timer without cancelling prior ones, and the `%expire` wake handler deleted the entry unconditionally. So the timer armed by an earlier `%set` fires exactly `timeout` after that set and deletes presence that later `%set`s had kept alive — the indicator drops until the next keepalive arrives. With the gateway's 30s keepalive + `~s90` timeout, every dropout on a live ship landed exactly 90s after an earlier set (4/4 timestamped predictions).

**Fix:** the `%expire` wake handler now looks up the entry and only deletes (and gives `%gone`) once `since + timeout` (falling back to the topic default) has actually passed. Stale wakes are no-ops. No cancellation bookkeeping, no state change, no protocol change.

## Testing

- New agent tests in `desk/tests/app/presence.hoon` covering the subscriber-side timer behavior:
  1. incoming `%set` fact stores the (context-translated) entry, gives `%here`, arms a timer
  2. a stale timer's wake preserves presence refreshed by a later `%set`; true expiry still deletes and gives one `%gone`
  3. a wake with no matching entry is a no-op
- Negative check: tests 2 and 3 fail on the unfixed code ("got more cards than expected").
- Full desk test suite green with the fix.
- Live verification on e2e fakezods: with a gateway-identical `presence-action-1` keepalive loop (~ten → ~zod DM), unfixed code dropped the indicator on schedule per the stale-timer arithmetic; after hot-swapping the fixed agent, zero dropouts under the identical loop.

### DIFF
--- a/desk/app/presence.hoon
+++ b/desk/app/presence.hoon
@@ -636,7 +636,19 @@
     =*  topic    i.t.t.wire
     =*  context  t.t.t.wire
     =*  key      [context ship topic]
-    ::TODO  no-op if we didn't have it anyway
+    ::  every %set arms its own expiry timer without cancelling prior ones,
+    ::  so a stale timer may fire while fresher %sets keep the entry alive.
+    ::  only delete once the entry has actually expired.
+    ::
+    =/  tos  (~(gut by places) context *topics)
+    =/  pes  (~(gut by tos) topic *people)
+    ?~  pre=(~(get by pes) ship)
+      [~ this]
+    =/  end=@da
+      %+  add  since.u.pre
+      (fall timeout.u.pre (default-timeout topic))
+    ?:  (gth end now.bowl)
+      [~ this]
     =.  places   (del-presence places key)
     [[(give-response %gone key)]~ this]
   ==
--- a/desk/tests/app/presence.hoon
+++ b/desk/tests/app/presence.hoon
@@ -1,0 +1,99 @@
+::  tests for %presence
+::
+::    focused on expiry-timer behavior on the subscriber's ship: every
+::    incoming %set fact arms its own expiry timer without cancelling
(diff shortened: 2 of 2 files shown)


arthyn (approved):
lgtm!

desk/app/presence.hoon
```diff
@@ -636,7 +636,19 @@
     =*  topic    i.t.t.wire
     =*  context  t.t.t.wire
     =*  key      [context ship topic]
-    ::TODO  no-op if we didn't have it anyway
+    ::  every %set arms its own expiry timer without cancelling prior ones,
+    ::  so a stale timer may fire while fresher %sets keep the entry alive.
+    ::  only delete once the entry has actually expired.
+    ::
```
Fang-, line 642:
```suggestion
    ::  only delete if the current entry has actually expired
    ::
```

Was wondering whether this logic would be more cleanly added inside of `+del-presence`, but there's other callsites that just want to delete unconditionally, it'd clean this up but make a mess elsewhere.

Fang- (commented):
Right, this will do you in. We do have `+cancel-inspire` though (and I _thought_ it was used to prevent this but apparently overlooked this), so:

> Consider using `+cancel-expire` to drop the expiration timer eagerly (during presence overwrite, like we do with deletion) rather than no-opping when it fires. Is the resulting logic simpler/easier to reason about? If so, fix it that way instead.

PR #6086, opened 2026-07-08
Paste images into chat inputs

janicduplessis:
Fixes TLON-6143

## Summary

Pasting an image into the native chat/DM composer now attaches it inline, matching what web already does. This removes the "Paste from Clipboard" item in the `+` attachment menu, which was a workaround for the native `TextInput` not being able to receive a pasted image.

## Changes

The active chat composer (`BareChatInput`) is a plain React Native `TextInput`, which can't surface pasted image data.

- **`PasteableTextInput` wrapper** — a `.native.tsx` / `.tsx` platform split. Native renders `PasteInput` from `@mattermost/react-native-paste-input` (a drop-in `TextInput` replacement with an `onPaste` callback); web/default renders the plain `TextInput`. The split keeps the native module out of the web (vite) bundle — it's imported only from the `.native.tsx` file.
- **Paste handling** — `onPaste` filters to image files, resolves dimensions via the existing `imageSize` util, and adds each as an attachment using the same shape the web `usePasteHandler` already produces (`{ type: 'image', file: { uri, width, height, mimeType, fileSize } }`). Web is unchanged; native paste is image-only (web additionally accepts pasted video, as before).
- **Workaround removed** — the "Paste from Clipboard" action and its `clipboardUtils` image helpers are deleted. `expo-clipboard` stays (still used by copy actions, invite screens, etc.); only the image-specific helpers are gone.
- **Scope note (behavior change):** the removal applies everywhere the shared attachment sheet is used. The long-form/notebook/gallery inputs are a WebView editor that `react-native-paste-input` can't cover, so they **lose clipboard-paste** — they keep the picker and file upload.
- **iOS native** — registers the paste-input module in `AppDelegate` (after `startReactNative`), plus a local `pnpm` patch to the library. Without the patch, paste silently never works on iOS: `PasteInputModule` captures `reactHost` once in `+setup:`, but under Expo (bridgeless) the `RCTHost` is created lazily *after* that call, so the Fabric surface presenter can't be resolved and the paste swizzle is never installed. The patch re-reads `reactHost` from the `rootViewFactory` lazily at lookup time. Documented in `patches/README.md`; upstream fix filed as [mattermost/react-native-paste-input#56](https://github.com/mattermost/react-native-paste-input/pull/56).

## How did I test?

Device-verified image paste end to end on **RN 0.85.3 / Expo SDK 56 (New Arch)**:

- **iOS** (simulator): copy an image → paste in a chat → attaches as a preview chip and sends.
- **Android** (emulator): copy an image (Gboard clipboard → IME `commitContent`) → attaches. Screenshot below.
- `+` menu no longer shows "Paste from Clipboard"; the other items (Photo Library, Take a Photo, Upload a File, Voice Memo) remain.
- Unit tests for the paste→attachment conversion (Vitest, 3 cases: image conversion, non-image ignored, dimension fallback). `pnpm -r tsc` and `packages/app` lint are clean; the `packages/app` Vitest suite passes.

Android — pasted image attached in the composer

## Risks and impact

- Safe to rollback without consulting PR author? **Yes** — self-contained, no migrations or state changes.
  - [x] Other: chat message composer (input + attachments)

- **iOS depends on a local patch** to `react-native-paste-input` (the bridgeless `reactHost` fix above). It's a 3-line, no-op-on-non-Expo change; remove once [upstream #56](https://github.com/mattermost/react-native-paste-input/pull/56) ships. `findNodeHandle` (upstream #55) was evaluated and is **not** needed here — on RN 0.85 the input's `__nativeTag` is populated.
- **Scope:** non-chat WebView inputs (notebook/gallery/long-form) lose clipboard-paste; they keep the image picker and file upload.

## Rollback plan

Revert the PR. No migrations or persisted-state changes. The revert restores the "Paste from Clipboard" menu item and removes the local patch. Other devs need `pnpm install` (applies/removes the patch) + a native rebuild.

## Screenshots / videos

Android paste (image attached) shown above. iOS verified on the simulator with the same flow.


wca4a:
this would be amazing

dnbrwstr (approved):

PR #6088, opened 2026-07-08
chore: migrate to TypeScript 7 and oxc toolchain (oxlint, oxfmt, tsdown)

janicduplessis:
## Summary

Upgrades the workspace to **TypeScript 7** (the native `tsgo` compiler, published as `typescript@7.0.2`) and replaces the build/lint/format tooling that TS 7 breaks. TS 7 removed the legacy JavaScript compiler API (`require('typescript')` now exposes only `version`), so everything that consumed it — typescript-eslint, tsup's `rollup-plugin-dts`, prettier's TS integration — no longer runs. Rather than pin those tools to an old TS, this moves the toolchain to the oxc stack (**oxlint + oxfmt + tsdown**), none of which touch the TS API.

## Changes

- **TypeScript 7.** Bumps the `typescript` override to `7.0.2`. Migrates tsconfigs off options TS 7 removed (`baseUrl`, `downlevelIteration`, `moduleResolution: node`), makes `paths` relative, and adds `skipLibCheck` to the desktop solution config.
- **Genuine type fixes.** TS 7's checker is stricter; fixes span FlashList/`ListRenderItem` generics, query-fn nullability, `ComponentType` prop constraints, null-vs-undefined variance, and IPC/`better-sqlite3` typing in desktop main. No `as any` / `@ts-ignore`.
- **react-navigation.** Its types aren't TS-7-compatible yet (upstream fix in-flight but unreleased — react-navigation/react-navigation#13163). The ~56 resulting errors are quarantined behind `@ts-expect-error` comments referencing that PR. `@ts-expect-error` self-fails once the directive is unnecessary, so they flag themselves for removal the moment a compatible react-navigation ships.
- **Lint → oxlint.** Replaces ESLint + typescript-eslint. Our custom `no-restricted-syntax` guards (the TLON-5598 `navigate()`/`pop:true` rule, `getToken`→`getTokenValue`, `onPress`-on-`Stack` ban, `CommonActions.reset` ban) and the api/shared import boundaries have no native oxlint equivalent, so they're reimplemented as a local oxlint JS plugin (`oxlint/tlon-plugin.cjs`) loaded via `jsPlugins`. `pnpm -r lint` is unchanged.
- **Format → oxfmt.** Prettier config migrated to `.oxfmtrc.json`; oxfmt also handles import-order, replacing the `@trivago` plugin. Scoped to code files; the one-time reformat touches ~5% of `.ts/.tsx`. CI's format step now runs `pnpm format:check`.
- **`packages/api` build → tsdown.** Replaces tsup. Uses `dts: { tsgo: true }` so declaration bundling runs through the native compiler instead of `rollup-plugin-dts`; output extensions kept as `.js`/`.d.ts` so the published package's `exports` are unchanged.
- **Patch `@tamagui/build`.** Guards a module-level `ts.sys` access so `tamagui-build --skip-types` (the `@tloncorp/ui` build) doesn't crash on load under TS 7.

## How did I test?

All CI-equivalent gates run locally and green:

- `pnpm install`, `pnpm -r tsc` (tsgo, **0 errors across all 8 packages** incl. `tsc -b` desktop), `pnpm -r lint` (oxlint, 0 errors), `pnpm format:check` (oxfmt, clean).
- Builds: `build:editor`, `build:packages` (incl. tamagui `ui`), `build:web`, desktop `build:ts`, tlon-skill `build:smoke`. Tests: `shared` + `app` pass.
- The custom-guard oxlint plugin is verified firing on fixtures, including the `pop:true` escape-hatch and `as never` cast handling.

**Not run locally:** the full `build:mobile` metro bundle (heavy; metro/babel don't touch the TS API and mobile typechecks clean, so low risk — relying on CI).

## Risks and impact

Foundational change — it touches the whole repo's toolchain, so the blast radius is every build/lint/format path. Worst cases are mostly "a tool misbehaves", not runtime app breakage:

- **Web tamagui falls back to runtime styling under TS 7 (perf deopt, not breakage).** `build:web` succeeds and produces a working bundle, but tamagui's vite-plugin config bundling hits the same missing-TS-API wall (`ts.sys.fileExists`), so compile-time style extraction is off until tamagui ships TS-7 support. This is the most consequential real-world side effect and worth QA on web perf.
- **oxlint custom rules depend on `jsPlugins`, which is alpha** (not semver-stable). If it regresses, only our custom guards are affected; the rest of linting is native. The guards are behavior-verified in this PR.
- **oxfmt is young (v0.58).** It's scoped to `.ts/.tsx/.js` only — YAML/markdown/JSON/openapi specs are left to their existing formatting.
- **react-navigation typings** are suppressed, not fixed — see above; they self-flag for cleanup on the upstream release.

- Safe to rollback without consulting PR author? **Yes** — it's a revert of the toolchain swap (mechanical), though the revert diff is large because of the one-time reformat.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: **build/lint/format toolchain + TypeScript version** (no product logic changes beyond type annotations)

## Rollback plan

Revert the PR. That restores the `typescript@5.9` override, the ESLint/Prettier configs and deps, the tsup api build, and removes the oxc configs and the tamagui patch. No data/migration state involved.

## Screenshots / videos

N/A — tooling and type-level change, no user-facing UI difference.


Valiums:
Following up on [the Blacksmith failure report](https://github.com/tloncorp/tlon-apps/pull/6088#issuecomment-5608080411): the quick group is visible in the failure screenshot, but the test expects `ChatListItem-Untitled group-unpinned` while the trace shows `ChatListItem--unpinned`.

Quick groups have an empty title. The UI displays the fallback “Untitled group”, but the test ID uses the raw title (`??` preserves an empty string). The rendered ID and the test selector need to be aligned to resolve this timeout.

PR #6089, opened 2026-07-09
fix ios mention insertion spacing

wca4a:
Fixes TLON-6141

## What changed

- Return the selected mention replacement text together with the intended cursor position from the bare chat input mention hook.
- On native platforms, push both the updated text and selection into the `TextInput` via `setNativeProps` after a mention is selected.
- Replace any overlapping mention metadata when selecting a mention, so repeat selection at the same range cannot render duplicated mention chips.

## Why

On iOS, selecting a mention visually inserted the mention but left the native input buffer/caret at the end of the mention text with no effective trailing separator. The next typed characters attached to the mention and could break the mention span. If the same selection fired again over the same range, the hook also appended duplicate mention metadata, which could make the mention appear to replicate while typing.

## Impact

After selecting a mention suggestion, the caret lands after the inserted trailing space so continued typing starts as normal text while the mention metadata remains intact. Re-selecting a mention over the same text range replaces the existing mention record instead of duplicating it.

## Validation

- Confirmed the supplied screen recordings reproduced the missing separator/caret behavior and duplicate rendered mention behavior.
- `git diff --check`
- `pnpm --filter '@tloncorp/app' exec eslint ui/components/BareChatInput/index.tsx ui/components/BareChatInput/useMentions.tsx` passed with existing warnings only.
- `pnpm run build:editor`
- `pnpm --filter '@tloncorp/app' tsc`


wca4a:
@codex review

dnbrwstr (approved):
I'd fix the codex comment, otherwise looks fine to me.

PR #6099, opened 2026-07-10
metagrab: protect from infinite redirect loop

mikolajpp:
## Summary

Metagrab suffers from infinite redirect loop bug. This manifest as infinite loop on the ship when a user pastes a link. It can only be stopped from terminal, to which ordinary hosted users have no access to.

Metagrab caches results of url fetches. However, as we don't implement cookie handling, we are unable to properly resolve stateful redirects. This poisons our cache, since following redirects without supplying the received session cookie can lead to an infinite redirect loop when we are finally redirected back to the original url but don't have valid cookie for the request.

The proper fix would be to handle the typical flow of cross-domain cookie synchronization. However, given that this bug causes ship to hang, we first implement a fix to guard against infinite redirection loops: we detect any infinite loops and cap the maximum number of requests to 10.

The required changes are rather delicate. The code includes a distributed proof described in the comment, which proves that an infinite loop can't happen. 

Closes TLON-6077

## Changes

Carefully protect against redirection loops, cache poisoning and infinite request chains.
Log when an error is encountered.

## How did I test?

Triggered the bug using a link. Deployed the fix, observed that the loop stopped and an error was reported.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: It renders the ship unresponsive until manual intervention. 

## Rollback plan

Revert.

### DIFF
--- a/desk/app/metagrab.hoon
+++ b/desk/app/metagrab.hoon
@@ -452,7 +452,13 @@
       :-  (give-response (~(get ju await) url) now.bowl %300 nex)
       this(await (~(del by await) url))
     ::TODO  deduplicate with %handle-http-request somehow?
+    =+  redir-max=10
+    =+  orig-nex=u.nex
     |-  ^-  (quip card _this)
+    ?.  (gth redir-max 0)
+      %-  (tell:l %warn 'maximum redirects exceeded' orig-nex ~)
+      :-  (give-response (~(get ju await) url) now.bowl %300 nex)
+      this(await (~(del by await) url))
     ::  move awaiters over to the next target
     ::
     =.  await
@@ -476,7 +482,7 @@
     ?:  ?&  ?=([%300 ~ @] wat.u.entry)
             ?=(^ (de-purl:html u.nex.wat.u.entry))
         ==
-      $(u.nex u.nex.wat.u.entry)
+      $(u.nex u.nex.wat.u.entry, redir-max (dec redir-max))
     ::  otherwise, serve the response from cache
     ::
     [(give-response (~(get ju await) u.nex) u.entry) this]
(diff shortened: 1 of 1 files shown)


desk/app/metagrab.hoon
```diff
@@ -476,7 +478,7 @@
     ?:  ?&  ?=([%300 ~ @] wat.u.entry)
             ?=(^ (de-purl:html u.nex.wat.u.entry))
         ==
-      $(u.nex u.nex.wat.u.entry)
+      $(u.nex u.nex.wat.u.entry, redir-max (dec redir-max))
```
Fang-: You say we "cap the maximum number of requests to 10 and log the original request url", but codex is right, we don't actually catch the `0` case anywhere, so we just hit this decrement again and crash.

So at the very least, you'll want to handle that, unless you're explicitly fine with serving `500` responses in these scenarios.

A technically more accurate redirect _loop_ detection is to, instead of keeping a counter, keep a `=|  redirs=(set @t)`, then here `(~(put in redirs) u.nex)`, and somewhere near the top, if we're `(~(has in redirs) u.nex)`, early-out with an appropriate error response.

(Either way this resolves the `::TODO  detect redirect loops` a couple lines up.)

Fang- (changes requested):

desk/app/metagrab.hoon
mikolajpp: Oh, looks like a hunk wasn't commited. I have a guard locally.

Fang- (dismissed):
Changes look good, good catches all around. Thank you!

mikolajpp:
Do not merge yet. I am working on another approach to this fix that will avoid the problem of poisoned cache triggering another infinite loop via `+on-poke`

mikolajpp:
@arthyn ready for review.

arthyn (approved):
lgtm

PR #6118, opened 2026-07-13
chat: retain link attachment when removing link text

jamesacklin:
## Summary

Fixes [TLON-6072](https://linear.app/tlon/issue/TLON-6072): pasting a link in Chat shows a preview, but replacing the link text with a comment removed the preview. Once a link preview has appeared it now persists when its URL is edited out of the text, so you can send the link with a comment instead of the raw URL.

## Changes

- `BareChatInput`: the debounced URL-watching effect no longer removes link attachments when their URL leaves the input text (a regression from 9013560, which added URL-sync removal). The preview's remove button is now the only way to dismiss it. This is safe because the attachment carries the URL and is sent as its own `link` block, so the post still contains the clickable link.
- Added a guard so that a URL deleted *before* its metadata arrives doesn't pop in a surprise preview after the fact — checked against the live input text at fetch-resolution time.
- Factored the URL extraction (code-fence stripping, hash normalization, dedupe) into a shared `extractPreviewUrls` helper used by both the effect and the guard.
- The duplicate-preview-on-edit protection from d50ad72 is preserved via the existing-attachment check.

## How did I test?

Drove the real web app (local ~zod ship + vite dev server, `BareChatInput` is shared across web/iOS/Android with no platform branches in the touched code):

- Typed a URL → preview appeared → replaced the whole text with a comment → preview persisted (pre-fix it disappears after the 500ms debounce) → sent → post renders the link preview block plus the comment.
- Clicked the preview's X while the URL was still in the text → removed, and it does not re-add while typing continues.
- Deleted a URL while its metadata fetch was still in flight (local HTTP server with a 6s-delayed response as the link target) → no ghost preview appeared after the fetch resolved.
- `tsc`, eslint, and prettier clean for `packages/app`.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: chat message input / link preview attachments

## Rollback plan

Revert the single commit; no schema, API, or backend changes.

## Screenshots / videos

Verified flow: preview card stays above the input after the URL text is replaced with "check out this cool site", and the sent message shows the link preview block with the comment beneath it.

### DIFF
--- a/packages/app/ui/components/BareChatInput/index.tsx
+++ b/packages/app/ui/components/BareChatInput/index.tsx
@@ -603,11 +603,17 @@ function BareChatInput(
 
             // If the URL was deleted from the text before its metadata
             // arrived, the user never saw a preview - don't surprise them
-            // with one now.
+            // with one now. Un-mark the URL as seen so that if it is typed
+            // again, the next debounce pass re-fetches it (the deletion may
+            // never reach debouncedText, so this pass's marking would
+            // otherwise suppress the re-added URL forever).
             if (!extractPreviewUrls(latestTextRef.current).includes(url)) {
               bareChatInputLogger.log('skipping link preview for removed url', {
                 url,
               });
+              prevUrlsRef.current = prevUrlsRef.current.filter(
+                (u) => u !== url
+              );
               return;
             }
 
--- a/packages/app/ui/components/BareChatInput/index.tsx
+++ b/packages/app/ui/components/BareChatInput/index.tsx
@@ -76,6 +76,21 @@ function normalizePreviewUrl(url: string) {
   }
 }
 
+// Extract normalized, deduplicated URLs eligible for link previews. Strips
+// backtick code spans/blocks (including unclosed ones) first so that URLs
+// inside code fences are not turned into link attachments. Triple-backtick
+// alternation is listed first so that ``` is not accidentally consumed by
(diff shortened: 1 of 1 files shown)


dnbrwstr (approved):
I think this behavior is confusing + removing the preview when the url isn't present makes more sense and is more consistent. But code looks fine.

jamesacklin:
> I think this behavior is confusing + removing the preview when the url isn't present makes more sense and is more consistent. But code looks fine.

@e-shee care to lobby?

e-shee:
> > I think this behavior is confusing + removing the preview when the url isn't present makes more sense and is more consistent. But code looks fine.
> 
> @e-shee care to lobby?

The behavior I'm trying to capture is when someone wants to send a link and a comment, but not the URL text.

dnbrwstr:
Are there other apps that do it that way? Most apps (eg signal) treat the preview a preview of the link rather than an attachment, so when you remove the link there's nothing to preview and the block gets removed.

e-shee:
Not sure—but I'm fairly confident we used to support the behavior, hence the surprise when testing. Don't really have a strong opinion either way!

dnbrwstr:
Makes sense! @jamesacklin my vote is no, but do as thou wilt.

PR #6127, opened 2026-07-13
bot e2e: shared parity scenarios for reactions, threading, restart, cron delivery

patosullivan:
## Summary

Fixes TLON-6100

Adds four parity scenario groups to the shared bot E2E harness (`packages/tlon-bot-e2e`), run against both OpenClaw and Hermes: inbound reactions on the bot's own posts, channel and DM threading, restart/reconnect no-double-reply, and Hermes cron delivery. Includes two OpenClaw production fixes surfaced by the new scenarios.

**Stacked on #6108**: the reaction and DM-thread scenarios depend on its Hermes adapter/CLI behavior.

## Changes

New shared scenarios (both drivers unless noted):

- Inbound reaction on the bot's own post dispatches a reply
- Channel threading (root anchoring plus participated-thread follow, with a non-participated control) and DM threading
- Restart/reconnect no-double-reply, **merged skipped on both drivers**: it exposed a real product bug where messages received while the bot is down are dropped after reconnect on both runtimes. Evidence filed on TLON-6098; the scenario is the regression gate once replay lands
- Hermes-only cron delivery targeting the home conversation, via a `no_agent` job plus direct `run_one_job` (no model turn); model-driven cron round-trip is follow-up TLON-6150

Harness plumbing:

- Auxiliary-call-aware fake model with server-side classification of Hermes title-generation calls
- `addReact` and `postWithReplies` actor methods in `src/tlon/actor.ts`
- Docker-by-label lifecycle helper (stop/start/exec) in `src/runtime/docker-direct.ts` for the restart scenario
- `orderDependent` scenarios now run last; first-class `skipReason` surfaces as a visible `test.skip`

OpenClaw fixes:

- Reaction-target resolution now fetches on cache miss via host-assigned IDs, covering both top-level posts and the exact-reply route; unresolved reactions are logged instead of misclassified
- Extracted `recordSentTlonReply` output recording, restoring group-send context-lens recording

## How did I test?

- `tlon-bot-e2e` unit tests: 80/80 passing
- `tsc` clean for `tlon-bot-e2e` and `openclaw`
- OpenClaw unit suite: 829 passed
- Full E2E baselines green on both drivers: Hermes 18 passed + 1 skipped (run twice), OpenClaw 17 passed + 1 skipped (re-run after the final reply-reaction fix)

## Risks and impact

## Rollback plan

Revert


packages/openclaw/src/monitor/output.ts
```diff
@@ -0,0 +1,51 @@
...
+  replyPreview: string;
+  replyText: string;
+  senderShip: string;
+}): void {
+  if (!outputMessageId) {
+    return;
+  }
+
+  if (!isGroup) {
+    cacheMessage(`dm/${senderShip}`, {
```
arthyn, line 34: Coverage ask: this DM send-time cache entry is one of the PR's two production fixes — it exists so a reaction on the bot's DM reply can be classified before the echo arrives (DMs have no scry fallback) — but no E2E scenario ever reacts to a bot DM message. Coverage is `output.test.ts` unit assertions only, which pin that the entry is written, not that a DM reaction dispatches end-to-end.

The pieces are already on the table: `addReact` takes `parentId`/`parentAuthorId`, and the `dm-thread-anchoring` scenario produces a bot-authored DM reply to react to. A `dm-reaction-on-bot-reply` scenario would close the loop on the exact race this fix targets.

packages/openclaw/src/monitor/history.ts
```diff
@@ -286,6 +353,38 @@ async function fetchParentPost(
   }
 }
 
+async function fetchReplyHistoryEntry(
```
arthyn, line 416: Coverage ask: the exact-reply route never fires in a scenario — the E2E reaction scenario reacts on a root post only, so this half of `lookupOrFetchCachedChannelMessage` is unit-tested but has no regression gate. Notably the PR description says the OpenClaw baseline was re-run "after the final reply-reaction fix", which implies reply reactions are exactly where a bug surfaced.

The `channel-thread-anchoring-and-follow` scenario already leaves a bot thread reply sitting in a channel; reacting to it (with `parentId` set to the root) would exercise this route end-to-end in a few lines.

arthyn (approved):
Solid work — the threading and cron scenarios pin real routing properties, the auxiliary-call redesign removes a whole class of counting flakiness, and the skipped restart scenario is handled the way deferred coverage should be. Approving with two coverage asks, both places where this PR's own production fixes outrun its E2E scenarios (inline comments): a DM reaction scenario to cover the send-time cache fix, and a reply-reaction to cover the exact-reply fetch route. Both look cheap with the harness pieces already in this PR.

PR #6140, opened 2026-07-15
Show dates on Context Lens run rows

wca4a:
## What changed

- show each Context Lens run's calendar date alongside its existing time
- apply the compact locale-aware timestamp to both recent-run list surfaces
- keep detailed inspector timestamps time-only
- cover the new formatter and missing timestamp fallback with unit tests

## Why

Time-only labels make runs from different days difficult to distinguish in collapsed run rows. Adding the date gives enough context without changing the row layout or Ochre styling.

## Validation

- `pnpm --filter @tloncorp/app test --run ui/components/Channel/ContextLens/format.test.ts`
- `pnpm exec prettier --check` on changed files
- `git diff --check`
- visually verified in the Tlon Cosmos component preview


wca4a:
@codex review

PR #6147, opened 2026-07-16
desk: add channel-action spider threads

arthyn:
## Summary

Adds spider threads that expose the `%channels` action poke over an HTTP JSON endpoint, so external services can post to a channel with a plain authenticated `POST` instead of holding an eyre channel open. Living in the groups desk means they stay in sync with `sur`/`mar` changes rather than rotting in a frozen snapshot desk.

## Changes

Following the existing versioned-thread pattern (`ted/group/create.hoon` / `create-1.hoon`), there is one thread per channel-action mark version, each typed to exactly one input:

- `%channel-action` — v7 payloads: `/spider/groups/channel-action/channel-action/json`
- `%channel-action-1` — v9 payloads: `/spider/groups/channel-action-1/channel-action-1/json`
- `%channel-action-2` — v10 payloads: `/spider/groups/channel-action-2/channel-action-2/json`

Each thread `!<`s its input at exactly its own version, then pokes `%channels` with the mark it was given. No version sniffing happens inside a thread, and no conversion happens either — `app/channels` already accepts every action mark version it has shipped and runs the `channel-conv` chain itself.

The one thing a thread does touch is `.sent`, and only for post adds: it's the client id `%channels` dedupes posts on, and an HTTP caller has no way to pick a good one (callers commonly send `0`, which would make every post after the first a silent duplicate). Every other action passes through untouched.

## How did I test?

Assembled the desk onto a 408-kelvin dev ship and committed it.

- All three threads build (`-build-file`).
- Posted through all three endpoints over HTTP with `sent: 0`; each message landed in the channel with a distinct, non-zero `.sent`, confirming the stamp works and dedupe doesn't swallow repeats.
- Deleted those test posts through the same endpoint, confirming a non-post-add action passes through unmodified.
- A malformed arg is rejected by the typed `!<` (nest-fail → `%false-start`) instead of being mis-parsed.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: new spider threads only; no existing code paths touched

## Rollback plan

Revert the PR. The threads are additive — nothing else in the desk references them, and no agent, mark, or type changed.

## Screenshots / videos

N/A

Fixes TLON-6172

### DIFF
--- a/desk/ted/channel/action.hoon
+++ b/desk/ted/channel/action.hoon
@@ -5,30 +5,27 @@
 ::    version of the channel action mark — the input-mark segment of the
 ::    spider url picks the json parser, e.g.
 ::    /spider/groups/channel-action-2/channel-action/json — and
-::    dispatches on the shape received, newest first.
+::    upconverts older shapes before poking %channels.
 ::
 /-  spider, cv=channels-ver
-/+  io=strandio
+/+  io=strandio, ccv=channel-conv
 =,  strand=strand:spider
 ::
 ^-  thread:spider
 |=  arg=vase
 =/  m  (strand ,vase)
 ^-  form:m
-;<  =bowl:strand  bind:m  get-bowl:io
 =/  new  ((soft (unit a-channels:v10:cv)) q.arg)
-?^  new
-  ?~  u.new  (pure:m !>(~))
-  =/  a-channels  u.u.new
-  ?>  ?=([%channel ^ %post %add *] a-channels)
-  =.  sent.essay.c-post.a-channel.a-channels  now.bowl
-  ;<  ~  bind:m  (poke:io [our.bowl %channels] channel-action-2+!>(a-channels))
-  (pure:m !>(~))
-=/  old  ((soft (unit a-channels:v7:cv)) q.arg)
-?~  old  ~|(%unknown-channel-action-version !!)
-?~  u.old  (pure:m !>(~))
-=/  a-channels  u.u.old
-?>  ?=([%channel ^ %post %add *] a-channels)
-=.  sent.essay.c-post.a-channel.a-channels  now.bowl
-;<  ~  bind:m  (poke:io [our.bowl %channels] channel-action+!>(a-channels))
+=/  a-channels=(unit a-channels:v10:cv)
+  ?^  new  u.new
+  =/  old  ((soft (unit a-channels:v7:cv)) q.arg)
+  ?~  old  ~|(%unknown-channel-action-version !!)
+  ?~  u.old  ~
+  `(v10:a-channels:v9:ccv (v9:a-channels:v7:ccv u.u.old))
+?~  a-channels  (pure:m !>(~))
+=/  action  u.a-channels
+?>  ?=([%channel ^ %post %add *] action)
+;<  =bowl:strand  bind:m  get-bowl:io
+=.  sent.essay.c-post.a-channel.action  now.bowl
+;<  ~  bind:m  (poke:io [our.bowl %channels] channel-action-2+!>(action))
 (pure:m !>(~))
--- a/desk/ted/channel/action.hoon
+++ b/desk/ted/channel/action.hoon
@@ -1,0 +1,22 @@
+::  channel-action: send actions to %channels
+::
+::    HTTP endpoint for posting to a channel as this ship, used by the
+::    alert bot (serverless-infra /api/sendAlertBotMessage). Takes a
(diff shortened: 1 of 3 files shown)


desk/ted/channel/action.hoon
```diff
@@ -0,0 +1,34 @@
...
+|=  arg=vase
+=/  m  (strand ,vase)
+^-  form:m
+;<  =bowl:strand  bind:m  get-bowl:io
+=/  new  ((soft (unit a-channels:v10:cv)) q.arg)
+?^  new
+  ?~  u.new  (pure:m !>(~))
+  =/  a-channels  u.u.new
+  ?>  ?=([%channel ^ %post %add *] a-channels)
+  =.  sent.essay.c-post.a-channel.a-channels  now.bowl
```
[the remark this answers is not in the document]
arthyn: sure but you have to be authenticated as this ship for that to happen, not a real issue.

desk/ted/channel/action.hoon
```diff
@@ -0,0 +1,35 @@
+::  channel-action: send actions to %channels
+::
+::    HTTP endpoint for posting to a channel as this ship, used by the
+::    alert bot (serverless-infra /api/sendAlertBotMessage). Takes any
+::    version of the channel action mark (v7, v9, or v10) — the
+::    input-mark segment of the spider url picks the json parser, e.g.
+::    /spider/groups/channel-action-2/channel-action/json — and
+::    upconverts older shapes before poking %channels.
```
mikolajpp:
```suggestion
::    thread for posting to a channel as this ship, used by the
::    alert bot and typically invoked through eyre spider api.
```
let's not start documenting eyre api in our threads :-)

desk/ted/channel/action.hoon
```diff
@@ -0,0 +1,35 @@
...
+  =/  new  ((soft (unit a-channels:v10:cv)) q.arg)
+  ?^  new  u.new
+  =/  mid  ((soft (unit a-channels:v9:cv)) q.arg)
+  ?^  mid
+    ?~  u.mid  ~
+    `(v10:a-channels:v9:ccv u.u.mid)
+  =/  old  ((soft (unit a-channels:v7:cv)) q.arg)
+  ?~  old  ~|(%unknown-channel-action-version !!)
+  ?~  u.old  ~
+  `(v10:a-channels:v9:ccv (v9:a-channels:v7:ccv u.u.old))
```
mikolajpp: I see the intention here, which is to accommodate callers running on different api versions. This is something spider has no support for. It should ideally be fixed to pass the mark together with the vase, so we could handle this case properly. That is obviously out of scope here.

As this change stands, I am not sure if we should do this kind of versioning in the thread: we have an existing pattern that is used in groups creation thread, which is to simply copy over the thread with a version tag. This way we steer clear from this kind of type handling. Any client that wants to use a newer interface simply targets the new version of the thread.

If the alert bot continues to function correctly, I would say let's wait until next week for @Fang- opinion on this.

mikolajpp (commented):
Reviewed. The pattern of matching directly on types is highly unusual in userspace, or even in the kernel. This is a deficiency of spider, but we have an existing alternative way for handling it (see group creation thread).

Deferring for @Fang- opinion.

desk/ted/channel/action.hoon
Fang-: Right, this is a pattern I don't think we have/use anywhere yet. I think we should be careful around adding new ones. @mikolajpp describes an existing pattern (though I haven't fact-checked/looked closely at threads recently), so I'd be curious to hear if there's any reason not to follow that. The PR description describes what we do/introduce here, but not _why_.

Fang-: (And yes spider by virtue of doing userspace-within-userspace drops some otherwise gall-level affordances on the floor without reintroducing them. "Accept multiple marks & do conversions" is one of those.)

arthyn: Reworked in d24916a0a to follow the group-create pattern: one thread per mark version — %channel-action (v7), %channel-action-1 (v9), %channel-action-2 (v10) — each typed to a single input with `!<`, upconverting to v10 before the poke. No version sniffing inside a thread anymore.

mikolajpp (approved):
This looks good. The threads becoming generic poke threads brings a question why do we even need those in the first place. Surely poking over generic HTTP interface would just work.

(The bit in daa8313  about client not being able to set `.sent` surely is wrong. It's just a timestamp)

Issue #29, opened 2022-04-19
message input: inserting + previewing clearnet URLs


jamesacklin:
Two stories at work:

> As a member, I want to insert a http/https resource into the chat so I can share off-Urbit content with the chat.

> As a member inserting a link, I want to see a preview of the link immediately near my message text so that I can be sure I’m referencing the right content.

**Acceptance criteria**

Standalone links:
- Given write privileges to a chat channel and no other text in the message input...
- When I insert an http / https URL...
- I should see an abbreviated preview of the web resource I'm linking to with some metadata (oEmbed?) such as title, URL, favicon, etc. near the message input

Inline links:
- Given write privileges to a chat channel and other text present in the message input...
- When I insert an http / https URL...
- I should see the link I inserted represented as an inline URL with a "tag" representation

Removing:
- Given a URL in my message...
- When I click a "X" mechanism on either representation...
- My link attachment should be removed

PR #6156, opened 2026-07-20
notes: cite notes in chat — /v0/said preview surface + client note references

arthyn:
## Summary

Lets users reference a %notes note in chat the way channel posts are cited. A note gets a copyable in-app reference path (`/1/chan/notes/~host/notebook/note/<id>`); pasting it into any chat input turns it into a reference attachment, and readers' clients hydrate a preview card (title / author / snippet) through a new permission-gated `/v0/said` surface on the %notes agent, mirroring %channels' `/said`. Wikilink-based backlinks are a separate follow-up.

Fixes TLON-6180

## Changes

**Backend (`desk/`)**
- `sur/notes.hoon`: new `note-preview` + `said` types (snippet is the leading ≤400 chars of body-md, cut on codepoint boundaries).
- New marks `%notes-said` (JSON via new `said:enjs` in `lib/notes/json.hoon`) and `%notes-denied` (null sentinel, mirrors `%channel-denied`).
- `app/notes.hoon`: `/v0/said/~host/name/note/<id>` watch path — single-shot fact + kick. Hosts and local replicas answer from state; otherwise the agent proxies one watch to the notebook host and relays the answer (watch-nack coerced to `%notes-denied`, upstream sub torn down after the fact). Read gate: public notebooks preview for anyone; private/group notebooks reuse `can-view-flag`.
- Documented the new subscription in `docs/notes/asyncapi.yaml`.

**Client (`packages/`)**
- `NoteReference` variant in the `ContentReference` union; `toContentReference` (both copies) recognizes `notes/…` chan cites with `/note/<id>` paths; `contentReferenceToCite` round-trips them.
- `getNoteReference` (subscribeOnce to the local agent's `/v0/said`, dot-grouped id for `+slav %ud`) + `useNoteReference` react-query hook, injected through the `useRequests` context with a backward-compatible default.
- `NotesNoteReferenceLoader` renderer: notebook-styled preview card; taps navigate to `NotesDetail`; denied/missing shows the standard not-found skeleton.
- Insertion affordances: "Copy link to note" in the native note-row action sheet and "Copy Tlon reference" in the standalone notes web app (`ui.html`). The existing `REF_REGEX` paste flow converts the path into a reference attachment unchanged.

**Infra**
- `backend/run-tests.sh`: pin the test fakezod's ames port (`-p 31999`) so booting it no longer collides with other local fakeships (`mesa: bind: address already in use`).

## How did I test?

- 6 new agent tests in `desk/tests/app/notes.hoon` (host answers member, private denies stranger, public answers stranger, missing note denied, proxy + relay flow, nack→denied). Full Hoon suite green on a fresh fakezod: 290 unit tests + aqua/ph integration tests.
- New vitest cases for note-cite parse/serialize round-trips (`packages/api/src/__tests__/references.test.ts`); full `packages/api` suite green (295 tests). `tsc` clean across api/shared/app.
- Live end-to-end on a dev moon: deployed the desk, created a notebook + note, subscribed to `/v0/said/...` over an eyre channel — got exactly one `notes-said` fact with the expected JSON payload followed by the kick.

## Risks and impact

- Backend surface is additive: a new watch path + two new marks; no state schema change (state-15 unchanged), no changes to existing paths or the HTTP API.
- Client parsing change: `toContentReference` now branches on `notes/…` nests before the channel-post regex. Nest kinds other than `notes` are unaffected; older clients render an unknown-reference fallback for note cites rather than crashing (cite blocks were already opaque to them).
- The proxy path only fetches over the network on behalf of the local ship (`src.bowl` guard), and hosts strictly permission-check every request.

## Rollback plan

Revert the commit. No migrations to unwind; previously sent note cites would simply render as unsupported references again.

## Screenshots

N/A (preview card uses the existing notebook reference styling; can add screenshots once design takes a pass).


desk/app/notes.hoon
```diff
@@ -625,6 +625,22 @@
...
+    =/  nid=@ud  (slav %ud id.pole)
+    ?:  ?|  =(our.bowl ship.flag)
+            (~(has by books) flag)
+        ==
+      (give-said ~ flag nid src.bowl)
+    ::  refuse to fetch over the network on another ship's behalf
+    ?>  =(src.bowl our.bowl)
+    =/  =wire  /said/(scot %p ship.flag)/[name.flag]/note/(scot %ud nid)
+    ?:  (~(has by wex.bowl) [wire ship.flag %notes])
+      cor
```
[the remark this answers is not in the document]
arthyn: Fixed in 9528ad27c by noncing the proxy wire (entropy segment) and dropping the wex de-dup guard, rather than adding a result cache — a repeat request during the un-acked %leave window now just opens its own upstream watch.

desk/app/notes.hoon
```diff
@@ -959,6 +1006,55 @@
...
+  =/  =cage
+    ?~  entry=(get-book flag)  notes-denied+!>(~)
+    =*  bs  notebook-state.u.entry
+    =/  can-view=?
+      ?~  grp=group.bs
+        !=(~ (~(get by members.bs) who))
+      ?&  (group-synced u.grp)
+          (group-can-read u.grp flag who)
+      ==
+    ?.  |(=(%public visibility.bs) can-view)
```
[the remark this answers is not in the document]
arthyn: Fixed in 4bb37377b — the %public bypass now only applies to non-group notebooks, mirroring +se-member-join, with a regression test. Worth noting the practical severity is lower than P1: the leaking state is unreachable from the UI and requires the notebook owner to explicitly poke %public on a group notebook, a toggle that is otherwise inert for group mode.

patosullivan (approved):
Frontend code looks good.

desk/app/notes.hoon
```diff
@@ -625,6 +625,27 @@
...
+      ?:  ?=(%pub -.net.u.entry)  &
+      init.net.u.entry
+    ?:  |(=(our.bowl ship.flag) ready)
+      (give-said ~ flag nid src.bowl)
+    ::  refuse to fetch over the network on another ship's behalf
+    ?>  =(src.bowl our.bowl)
+    ::  nonce'd wire: a finished request's sub can linger in wex until the
+    ::  host acks the %leave, which would block an identical re-watch.
+    =/  =wire
+      /said/(scot %p ship.flag)/[name.flag]/note/(scot %ud nid)/(scot %uv eny.bowl)
```
Fang-: Pretty sure that's bogus, gall expels the subscription from the outgoing set as soon as it process the `%leave` locally.

Did you observe this being a problem in practice? If it's a real problem we'll want channels to do this too, but I assume you're fighting ghosts here.

desk/app/notes.hoon
```diff
@@ -959,6 +1006,59 @@
   ::  revocation rather than a replication gap.
   ?.  (group-synced u.grp)  &
   (group-can-read u.grp flag who)
+::  +said-path: subscription path for a single note-preview request
+::
+++  said-path
+  |=  [=flag:n nid=@ud]
+  ^-  path
+  /v0/said/(scot %p ship.flag)/[name.flag]/note/(scot %ud nid)
```
Fang-, line 1026: Channels puts the kind at the start. We don't do that here because the target agent (notes) only has one "kind" of channel, right?

We put `/note` in place of `/post`. I think for consistency we might want to do `/post` here too for "individual piece of content addressed by id". Probably doesn't matter too much, but if the segment is meaningless and too costly to strip out, then at least we can keep it consistent.  
But I could be convinced either way.

desk/app/notes.hoon
```diff
@@ -959,6 +1006,59 @@
...
+::  +said-snippet: leading slice of body-md, cut on a codepoint (not
+::  byte) boundary so multi-byte UTF-8 never gets split.
+::
+++  said-snippet
+  |=  body=@t
+  ^-  @t
+  =/  limit=@ud  400
+  =/  chars=(list @c)  (tuba (trip body))
+  ?:  (lte (lent chars) limit)  body
+  (crip (tufa (scag limit chars)))
```
Fang-: Doesn't save you from cutting in the middle of multi-codepoint emoji though. For example, you might separate these parents from their kids! 👨‍👩‍👧‍👦

Fang- (changes requested):
You might be shadow-boxing here.

Also I continue to detest the way your tool writes comments. But it's just notes codebase, so /shrug

desk/app/notes.hoon
arthyn: You were right — fought a ghost. Verified in gall: +ap-handle-peers deletes the boat.yoke entry (= wex) in the same event that processes the %leave, and gall nonces the underlying wires itself. Never observed in practice, only inferred. Reverted in 8b6a419d6 to the plain wire with the wex de-dup guard, same shape as channels's safe-watch.

desk/app/notes.hoon
arthyn: Fixed in 8b6a419d6: after the codepoint cut, the snippet backs off while the boundary touches cluster glue (zwj, variation selectors, skin tones, combining marks), so a trailing family keeps its kids (by losing the whole family). Regression test straddles a 4-person zwj family across the 400-codepoint limit.

desk/app/notes.hoon
arthyn: Went back and forth and landed on keeping /note: the rest of the notes surface consistently says "note" for this resource (REST /notes/{id}, cite wers /note/<id> matching diary's convention), so within-agent consistency seemed to beat cross-agent consistency with channels. Happy to flip it if you feel strongly — it's a one-string change while nothing depends on the path yet.

arthyn: sorry didn't mean for it to comment here, but I think we should keep note, its not a post really.

backend/run-tests.sh
```diff
@@ -79,10 +79,13 @@ then
 fi
 
 http_port=9090
+# Fixed ames port: fake ~zod's default port collides with any other fakezod
+# on the machine (e.g. the rube/playwright dev ships).
```
mikolajpp:
```suggestion
```

mikolajpp: That said, I think vere support `-p 0` to auto-select a free port?

backend/run-tests.sh
```diff
@@ -92,8 +95,11 @@ then
 fi
 
 echo "Booting ship"
-($vere --loom 33 --http-port $http_port -t $pier) &
+($vere --loom 33 --http-port $http_port -p $ames_port -t $pier) &
 vere_pid=$!
+# An orphaned vere keeps the pier and poisons later runs (commits silently
+# no-op against its stale mount mirror) — always reap it, even on ^C/kill.
```
mikolajpp:
```suggestion
```
Please don't include comments for self explanatory improvements.

desk/sur/notes.hoon
```diff
@@ -309,6 +309,20 @@
 ::  HTML body is served via /notes/pub/~host/name/note-id, not this list.
 ::
 +$  published-record  [=flag note-id=@ud]
+::  $note-preview: trimmed note view served on /v0/said — snippet is the
+::  leading slice of body-md, not the full document.
+::
```
mikolajpp:
```suggestion
::  $note-preview: trimmed note view served
::
::  note previews are served on /v0/said
::  .snippet is the leading slice of $note's body-md
::
```

I would not expect a _snippet_ to be a full document in any case!

desk/sur/notes.hoon
```diff
@@ -309,6 +309,20 @@
...
+  $:  id=@ud
+      title=@t
+      snippet=@t
+      author=ship
+      updated-at=@da
+      notebook-title=@t
+  ==
+::  $said: single-shot /v0/said response (fact then immediate kick)
+::
++$  said  [=flag =note-preview]
```
mikolajpp:
```suggestion
::  $said: single-shot /v0/said response
::
+$  said  [=flag =note-preview]
```

Can the model be instructed to stop doing tutorial-style, self-help kind of comments? It does not add anything.

desk/app/notes.hoon
```diff
@@ -625,6 +625,26 @@
       [%v0 %inbox %stream ~]
     ?>  =(src.bowl our.bowl)
     cor
+  ::
+      [%v0 %said ship=@ name=@ %note id=@ ~]
+    ::  single-shot note reference preview: answer from state if we can,
+    ::  else proxy one watch to the host and relay its answer in +agent.
+    =/  =flag:n  [(slav %p ship.pole) `@tas`name.pole]
```
mikolajpp, line 636:
```suggestion
    ::  single-shot note reference preview: answer from state if we can,
    ::  else proxy one watch to the host and relay its answer in +agent.
    ::
    =/  =flag:n  [(slav %p ship.pole) `@tas`name.pole]
```

Here and elsewhere, we tend to prefer flag comments when commenting about a code segment.

desk/tests/app/notes.hoon
```diff
@@ -2164,5 +2164,253 @@
   ?~  vis
     |+['snapshot missing visibility field']~
   &+[~ s]
+::  ====  /v0/said: single-shot note reference previews  ====
```
mikolajpp: Do we do this asciidoc-style headers on purpose? It's not even markdown, and for sure a new style in the codebase. I guess the model just followed what was already there, but we should considered adhering to our existing conventions.

desk/tests/app/notes.hoon
```diff
@@ -2164,5 +2164,253 @@
...
+  /v0/said/(scot %p ship.f)/[name.f]/note/(scot %ud nid)
+::  +setup-said-notebook: notebook 'NB' + one note ('T' / body 'B', id 3)
+::
+++  setup-said-notebook
+  =/  m  (mare ,[f=flag:n =bowl:gall])
+  =*  b  bind:m
+  ^-  form:m
+  ;<  ~  b  init-zod
+  ;<  =bowl:gall  b  get-bowl
+  ;<  *  b  (poke-a [%create-notebook 'NB'])
```
mikolajpp:
```suggestion
  ;<  *  b  (poke-a %create-notebook 'NB')
```
We badly need an official styling guide. It gets tiring to correct all these tiny styling mistakes, but if we don't do it eventually our codebase will be full of those because if models can do anything, it's imitation. Perhaps we should get together with @Fang- and figure out a way forward.

desk/tests/app/notes.hoon
```diff
@@ -2164,5 +2164,253 @@
...
+  =/  f=flag:n  (nb-flag our.bowl 'NB' 1)
+  ;<  *  b  (poke-a [%notebook f [%create-note 2 'T' 'B']])
```
mikolajpp:
```suggestion
  ;<  *  b  (poke-a %notebook f [%create-note 2 'T' 'B'])
```
We could even do away with the inner `[]`, idk. Certainly the outer bracket does not add to readability.

desk/tests/app/notes.hoon
```diff
@@ -2164,5 +2164,253 @@
...
+::  ====  test-said-host-answers-member  ====
+::  Host answers its own /v0/said watch with the preview fact + kick.
+::
+++  test-said-host-answers-member
+  %-  eval-mare
+  =/  m  (mare ,~)
+  =*  b  bind:m
+  ^-  form:m
+  ;<  [f=flag:n =bowl:gall]  b  setup-said-notebook
+  ;<  caz=(list card)  b  (do-watch (said-watch-path f 3))
```
mikolajpp: Pardon me asking, but was this hand-coded? Surprised to see us having to alias `b`, whereas `bind:m` is pretty much the pattern everywhere. Spelling `bind:m` adds volume to monadic lines and I personally find it more pleasant to read. I can yet be convinced  departing from established convention is a good change. CC @Fang- 

PS
If this was generated let's browbeat the model into using `bind:m` everywhere :-)

desk/tests/app/notes.hoon
```diff
@@ -2164,5 +2164,253 @@
...
+::  A non-member watching a private notebook's /v0/said gets %notes-denied.
+::
+++  test-said-private-denies-stranger
+  %-  eval-mare
+  =/  m  (mare ,~)
+  =*  b  bind:m
+  ^-  form:m
+  ;<  [f=flag:n =bowl:gall]  b  setup-said-notebook
+  ;<  *  b  (set-src ~bus)
+  ;<  caz=(list card)  b  (do-watch (said-watch-path f 3))
```
mikolajpp: We can also use the `+do-as` operator here.

desk/tests/app/notes.hoon
```diff
@@ -2164,5 +2164,253 @@
...
+::
+++  test-said-public-answers-stranger
+  %-  eval-mare
+  =/  m  (mare ,~)
+  =*  b  bind:m
+  ^-  form:m
+  ;<  [f=flag:n =bowl:gall]  b  setup-said-notebook
+  ;<  *  b  (poke-a [%notebook f [%visibility %public]])
+  ;<  *  b  (set-src ~bus)
+  ;<  caz=(list card)  b  (do-watch (said-watch-path f 3))
```
mikolajpp: See comment about `+do-as` above.

desk/sur/notes.hoon
```diff
@@ -309,6 +309,20 @@
...
 ::
 +$  published-record  [=flag note-id=@ud]
+::  $note-preview: trimmed note view served on /v0/said — snippet is the
+::  leading slice of body-md, not the full document.
+::
++$  note-preview
+  $:  id=@ud
+      title=@t
+      snippet=@t
+      author=ship
```
mikolajpp, line 320: Do we not want to support bot authorship through the `$author` type? Or did that fall out of favor already.

desk/sur/notes.hoon
```diff
@@ -309,6 +309,20 @@
 ::  HTML body is served via /notes/pub/~host/name/note-id, not this list.
 ::
 +$  published-record  [=flag note-id=@ud]
+::  $note-preview: trimmed note view served on /v0/said — snippet is the
+::  leading slice of body-md, not the full document.
+::
++$  note-preview
+  $:  id=@ud
```
mikolajpp, line 317: I mentioned it with last review, I will mention it again: we should consider whether we want to continue using `@ud` as the opaque note id, whereas `@uv` is generally used throughout the system. (I guess there are exceptions @Fang- ?)

 There are some practical implications, because I remember us generating those sequentially. It at least psychologically prioritizes earlier notes, which users might not like.
It also leaks information about the size of the notes collection anytime note id is shared.

desk/mar/notes/said.hoon
```diff
@@ -0,0 +1,14 @@
+/-  n=notes
+/=  notes-json  /lib/notes/json
```
mikolajpp: What is happening here? This should be a library import!

desk/app/notes.hoon
```diff
@@ -774,6 +794,31 @@
...
+      [%said ship=@ name=@ %note id=@ ~]
+    ::  proxied note-preview answer from a notebook host: relay to our
+    ::  /v0/said subscribers (nack/foreign mark coerced to %notes-denied).
+    =/  =flag:n  [(slav %p ship.pole) `@tas`name.pole]
+    =/  nid=@ud  (slav %ud id.pole)
+    =/  paths=(list path)  ~[(said-path flag nid)]
+    ?+  -.sign  cor
+        %watch-ack
+      ?~  p.sign  cor
+      =.  cor  (give %fact paths notes-denied+!>(~))
```
mikolajpp: As mentioned elsewhere, I believe this is a bad pattern because it confuses an unrelated bug and masks it as a fixed error message. It also undermines our effort to bring in better error semantics to the system at large.

`notes-denied` should be reserved to a specific condition, likely a permission error, and should not occur otherwise. If we want to signal errors back to subscribers, we should have another generic `notes-error` fact that can either carry nothing or transmit a genuine error message generated server-side.

Alternatively, I can also accept notes-denied itself being changed to a variant type such as 
```
?(%access-denied %crash)
```
or some such.

desk/app/notes.hoon
```diff
@@ -959,6 +1004,75 @@
...
+  =/  limit=@ud  400
+  =/  chars=(list @c)  (tuba (trip body))
+  ?:  (lte (lent chars) limit)  body
+  =/  glue
+    |=  c=@c
+    ?|  =(`@`c 0x200d)                        ::  zero-width joiner
+        =(`@`c 0xfe0e)  =(`@`c 0xfe0f)        ::  variation selectors
+        &((gte `@`c 0x1.f3fb) (lte `@`c 0x1.f3ff))  ::  skin tones
+        &((gte `@`c 0x300) (lte `@`c 0x36f))  ::  combining marks
+    ==
```
mikolajpp: This looks too simple to be working robustly. A model claims there is a bunch of unhandled cases here, but I have not consulted unicode spec for this. Is this snippet lifted off somewhere legitimate?

A while ago we have ported over unicode confusable spec to handle id spoofing. We should at know when this will break and document it, because this will surely manifest rendering bug for users who do anything outside of the covered set of cases.

desk/app/notes.hoon
```diff
@@ -959,6 +1004,75 @@
...
+  |-  ^-  @t
+  ?~  peek  ''
+  ?.  |((glue i.peek) (glue nxt))
+    (crip (tufa (flop peek)))
+  $(peek t.peek, nxt i.peek)
+::  +give-said: one %fact (preview or %notes-denied) then an immediate
+::  %kick, mirroring %channels' single-shot said flow. Public notebooks
+::  preview for anyone. Unlike +can-view-flag (which treats an unsynced
+::  group as viewable so subscriptions only drop on real revocations),
+::  previews fail closed: no synced group, no snippet.
```
mikolajpp:
```suggestion
::  +give-said: one %fact (preview or %notes-denied) then an immediate
::  %kick, mirroring %channels' single-shot said flow. public notebooks
::  preview for anyone. unlike +can-view-flag (which treats an unsynced
::  group as viewable so subscriptions only drop on real revocations),
::  previews fail closed: no synced group, no snippet.
```
AI should make up its mind. It can't go on casually inserting uppercase everywhere like a sane human would!

desk/app/notes.hoon
```diff
@@ -959,6 +1004,75 @@
...
+    ?~  entry=(get-book flag)  notes-denied+!>(~)
+    =*  bs  notebook-state.u.entry
+    ::  mirrors +se-member-join: visibility only means something for
+    ::  non-group notebooks; group mode's sole authority is can-read
+    =/  can-view=?
+      ?~  grp=group.bs
+        ?|  =(%public visibility.bs)
+            !=(~ (~(get by members.bs) who))
+        ==
+      ?&  (group-synced u.grp)
```
mikolajpp, line 1072: `+group-synced` is slightly confusing. Clearly we check for local existence, and this used to be nomenclature.

mikolajpp (changes requested):
Overall this looks good, logic wise. A whole bunch of styling nits as usual.

I have two points that need to be addressed before we merge this:
1. We use %notes-denied to conceal crashes occurring notes host side. We need to fix this by not crashing on host if there is a semantically clear error, and also capturing legitimate crashes in a separate category/mark.
2. Unicode logic must be ironed out. At the very least we must *know* when it is going to be broken.

desk/app/notes.hoon
arthyn: Split in 5840c7a7f: %notes-denied now strictly means no permission. A new null-payload %notes-error covers everything else — missing note for an authorized viewer, proxy watch-nacks, foreign marks. Missing notebooks still answer denied so existence can't be probed by strangers. Went with the separate-mark option so denied stays a clean single condition; can move to the variant-payload shape instead if you prefer it on the wire.

desk/app/notes.hoon
arthyn: Honest answer: hand-written approximation, not lifted. In 5840c7a7f the glue set is extended (zwnj, all variation selectors, every combining-mark block, regional-indicator flags — conservatively drops a trailing flag run rather than splitting one — and tag sequences) and the arm comment now documents exactly what's covered and the known gaps: hangul jamo composition and virama-joined indic conjuncts can still split, yielding valid but visually altered text. Full UAX #29 segmentation needs the property tables like the confusables port did — happy to do that as a follow-up if we decide it matters for note bodies.

backend/run-tests.sh
arthyn: Comment dropped. On -p 0: plausible, but I couldn't verify vere 4.5's semantics for it locally, and a fixed port keeps failures reproducible — took the deletion, kept the pin. Happy to switch if you know 0 works.

desk/tests/app/notes.hoon
arthyn: Generated, imitating this file — every pre-existing test in it uses `=*  b  bind:m`. The new said section now spells `bind:m` per your preference (5840c7a7f); left the ~60 older tests for a file-wide pass once you and @Fang- land the style guide, so the file isn't half-and-half by accident.

desk/tests/app/notes.hoon
arthyn: These match the file's pre-existing section headers (e.g. test-join-public-accepts up top) — imitation, as predicted. No attachment to them; whatever the style guide lands on, this file can be normalized in one pass.

desk/sur/notes.hoon
arthyn: Notes models authors as plain ships everywhere today (note.created-by, note-revision.author), so the preview just mirrors that. Adopting $author is a cross-cutting notes change — seems like it belongs with the notes-bot-awareness work (TLON-6182) rather than smuggled in here.

desk/sur/notes.hoon
arthyn: Agreed the tradeoffs are real (ordering leak, collection-size leak). It's the pre-existing id scheme across all of notes though — every REST path, cite, and state map keys on the sequential @ud — so switching to @uv is its own migration-bearing PR, not something this one should smuggle in.

desk/app/notes.hoon
arthyn: Fair — it predates this PR (added with the group-notebook work), this diff just calls it. +group-known or +group-replicated would say it better; follow-up rename so this diff doesn't grow.

desk/tests/app/notes.hoon
arthyn: Correction to my earlier reply: those headers were themselves model-written in a previous session — not a real convention, no other test file uses them. Normalized the whole file in 8432084df to the flag style groups' tests use (`::  +test-name: description`) with plain lowercase section banners; all 88 tests still green.

desk/mar/notes/said.hoon
arthyn: My earlier reply was wrong — /+ segments hyphenated names, so `/+  notes-json` resolves lib/notes/json.hoon just fine (as the channel marks demonstrate). Converted the new mark plus all pre-existing notes marks, the agent, and the tests in 17900112f; verified building and green on a live ship.

desk/app/notes.hoon
```diff
@@ -625,6 +624,27 @@
...
+      ?~  entry=(get-book flag)  |
+      ?:  ?=(%pub -.net.u.entry)  &
+      init.net.u.entry
+    ?:  |(=(our.bowl ship.flag) ready)
+      (give-said ~ flag nid src.bowl)
+    ::  refuse to fetch over the network on another ship's behalf
+    ?>  =(src.bowl our.bowl)
+    =/  =wire  /said/(scot %p ship.flag)/[name.flag]/note/(scot %ud nid)
+    ?:  (~(has by wex.bowl) [wire ship.flag %notes])
+      cor
```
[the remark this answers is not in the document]
arthyn: Not a real window: gall expels the outgoing subscription from wex in the same event that processes the %leave (+ap-handle-peers deletes the boat.yoke entry immediately, and gall nonces the underlying wires itself). A later /v0/said subscription can never observe the stale entry. This was litigated with @Fang- above — the nonce fix was reverted at his request after verifying gall source, not by accident.

desk/app/notes.hoon
```diff
@@ -774,6 +794,33 @@
...
+    =/  nid=@ud  (slav %ud id.pole)
+    =/  paths=(list path)  ~[(said-path flag nid)]
+    ?+  -.sign  cor
+        %watch-ack
+      ?~  p.sign  cor
+      =.  cor  (give %fact paths notes-error+!>(~))
+      (give %kick paths ~)
+    ::
+        %kick
+      (give %kick paths ~)
```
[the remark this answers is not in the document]
arthyn: The premise doesn't hold: the host answers inside on-watch itself — accept, fact, and kick are one atomic event, so "accepted but died before answering" isn't a reachable state. Failure of that event is a watch-nack, which we already map to %notes-error. Agent reloads don't kick subscribers and ship restarts don't drop subscriptions. The bare re-kick on kick-without-fact matches channels' take-said behavior for the same (operator-exotica-only) situation.

desk/app/notes.hoon
```diff
@@ -959,6 +1006,93 @@
...
+::  composition and virama-joined indic conjuncts can still split —
+::  valid codepoints, visually altered. full coverage needs the uax
+::  #29 property tables; port them if this ever matters in practice.
+::
+++  said-snippet
+  |=  body=@t
+  ^-  @t
+  =/  limit=@ud  400
+  =/  chars=(list @c)  (tuba (trip body))
+  ?:  (lte (lent chars) limit)  body
```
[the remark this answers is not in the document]
arthyn: Fixed in a0e3c0696: bodies over 1.600 bytes (4x the 400-codepoint limit, so any capped body necessarily exceeds the limit) are sliced as atoms before any decode, with a +utf8-head helper trimming a split trailing sequence; when the trim leaves fewer codepoints than the limit, the unknown next codepoint is treated as cluster glue. Two regression tests cover the split-sequence boundary and the exactly-at-limit slice.

desk/app/notes.hoon
```diff
@@ -1023,10 +1034,16 @@
...
+        &((gte `@`c 0xfe00) (lte `@`c 0xfe0f))    ::  variation selectors
         &((gte `@`c 0x1.f3fb) (lte `@`c 0x1.f3ff))  ::  skin tones
-        &((gte `@`c 0x300) (lte `@`c 0x36f))  ::  combining marks
+        &((gte `@`c 0x1.f1e6) (lte `@`c 0x1.f1ff))  ::  regional indicators
+        &((gte `@`c 0xe.0001) (lte `@`c 0xe.007f))  ::  tag sequences
+        &((gte `@`c 0x300) (lte `@`c 0x36f))      ::  combining diacritics
+        &((gte `@`c 0x1ab0) (lte `@`c 0x1aff))    ::  combining extended
+        &((gte `@`c 0x1dc0) (lte `@`c 0x1dff))    ::  combining supplement
+        &((gte `@`c 0x20d0) (lte `@`c 0x20ff))    ::  combining for symbols
+        &((gte `@`c 0xfe20) (lte `@`c 0xfe2f))    ::  combining half marks
```
Fang-: Why don't we slop together a real `/lib/unicode.hoon` at this point? It's the kind of spec-driven development the new tools are good at. Though I guess build times for a 20 mb .hoon file will become a concern...

Fang- (commented):
Changes look fine I guess. The whole unicode thing is overkill, I shouldn't have brought it up...

desk/app/notes.hoon
arthyn: Update: b033129ee replaces the grapheme backoff entirely with something much simpler at Hunter's request — truncate to 400 bytes, back off only to a utf-8 boundary, no decoding. The family CAN now lose its kids, but that's a documented limitation with a test asserting the exact split behavior, rather than a half-approximation of uax #29. If we ever want real cluster handling the honest path is porting the property tables.

desk/app/notes.hoon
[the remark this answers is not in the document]
arthyn: Superseded by b033129ee: the bounded-decode machinery is gone entirely — the snippet is now a 400-byte atom slice backed off to a utf-8 boundary, zero decoding, so cost is O(400) regardless of body size. Same DoS conclusion, simpler mechanism.

desk/app/notes.hoon
arthyn: we probably should have one in base tbh lol. I just reverted to the initial version so we may see some emoji families separated RIP.

desk/app/notes.hoon
```diff
@@ -797,24 +798,26 @@
   ::
       [%said ship=@ name=@ %note id=@ ~]
     ::  proxied note-preview answer from a notebook host: relay to our
-    ::  /v0/said subscribers (nack/foreign mark coerced to %notes-denied).
+    ::  /v0/said subscribers. a nack or foreign mark is a host-side
+    ::  failure, not a permission answer, so it becomes %notes-error.
+    ::
```
mikolajpp: `a nack or foreign mark...` part is not necessary, the code is obvious.

I just realized we don't log these anywhere, but we should! The pattern in our agents is that remote crashes get reported locally, since they don't trigger `+on-fail` at the host.

If notes are not released to users yet it is fine to do logging instrumentation at another time.

desk/app/notes.hoon
```diff
@@ -797,24 +798,26 @@
...
         %kick
       (give %kick paths ~)
     ::
         %fact
       =.  cor
-        ?:  ?=(?(%notes-said %notes-denied) p.cage.sign)
+        ?:  ?=(?(%notes-said %notes-denied %notes-error) p.cage.sign)
           (give %fact paths cage.sign)
-        (give %fact paths notes-denied+!>(~))
+        (give %fact paths notes-error+!>(~))
```
mikolajpp, line 831: Likewise marking this as a place where we should log eventually.

mikolajpp (approved):
Changes look good, I consider it ready to merge. Some feedback in the comments, notably about lack of logging.

desk/app/notes.hoon
arthyn: Done now rather than later in 115c906db — nack tang slogged locally per the remote-crashes-don't-reach-host-+on-fail convention, unexpected fact marks noted likewise, and the obvious half of the comment trimmed.

arthyn: Upgraded in be9cf73d9: bare slogs replaced with the logging library — fail:log carries the remote nack tang, tell:log covers unexpected marks, and +on-fail now routes through %logs like chat/channels/groups (plus the /logs poke-ack wire swallowed in +agent).

PR #6158, opened 2026-07-20
tlonbot: virtual bot identity

arthyn:
## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:


Fang- (changes requested):
tldr from what we discussed out of band:

The problem to solve is "bot doesn't see everything I see". The approach this PR takes is "facilitate running the bot through your own ship" by letting it pretend to be a real, distinct ship still. This introduces (& tries to solve) a new problem, which status quo already has "for free": the ability to DM the bot directly.

In other words, this approach creates an additional problem, then solves both problems. But I'd rather not create an additional problem by doing a simpler solution instead. (The whole "pretend there's a real ship there" thing goes against the grain on our architecture across the board, too.)

Bunch of ways to solve for getting the bot into all the channels its owner is in. The owner could expose a subscription endpoint that lets the bot listen to joined/left channels and "follow along". The owner could auto-invite the bot into any channels/groups it joins (iiuc the bot already auto-responds to invites by joining).

There's broader concerns/tradeoffs/directions in the idea maze, but I'm just answering for the problem at hand here.

PR #6159, opened 2026-07-20
add model-driven cron capability partition to shared bot e2e harness

patosullivan:
## Summary

Fixes TLON-6150

Adds the full model-driven cron round trip to the shared bot e2e harness (`packages/tlon-bot-e2e`), covering both the Hermes and OpenClaw drivers. The model schedules a cron job via a tool call, the runtime's own scheduler fires it, the fired job runs an agent turn against the scripted fake model, and the output delivers to the intended conversation.

Prior work (TLON-6100) only covered cron _delivery targeting_ on Hermes, using a `no_agent` job fired directly via `run_one_job`, with no model turn involved. This fills in the missing piece.

## Changes

-   Added a `'cron'` `RuntimeCapability` and a `cron` scenario partition (`TLON_BOT_E2E_SCENARIO_PARTITIONS=cron`), so cron tooling is enabled only in that partition and the baseline partition is unchanged.
-   OpenClaw: `cron` capability maps to the core `cron` tool in `tools.allow`. The job is scheduled as `{kind:'at'}` at +180s with `deleteAfterRun`, so it fires once and self-removes.
-   Hermes: `HERMES_E2E_ENABLE_CRONJOB` conditionally adds `cronjob` to `platform_toolsets.tlon` and drops it from `agent.disabled_toolsets`; `assertHermesConfig` is now per-partition. The job uses the execution-relative `'1m'` schedule with explicit `deliver: 'origin'`.
-   Fake-model strategy: the fired-turn `[tlon-test:KEY]` tag is embedded in the scheduled prompt so the cron-fired turn matches a registered script, asserted via `latest-user` provenance.
-   New `src/scenarios/shared/cron.ts` with driver-aware cron lifecycle helpers: all-jobs listing, race-safe creation and ID capture, and bounded ID-addressed teardown with asserted postconditions.
-   `execInComposeService`/`resolveComposeContainer` now share a single operation deadline. Previously container resolution could consume its own 60s on top of the exec timeout. Backward compatible: omitted timeouts keep the existing 60s default.

## How did I test?

-   Real Docker e2e runs of the new partition, both drivers green:
    -   `TLON_BOT_E2E_SCENARIO_PARTITIONS=cron pnpm --filter @tloncorp/tlon-bot-e2e test:e2e:hermes` — passed (scenario 180s).
    -   `TLON_BOT_E2E_SCENARIO_PARTITIONS=cron pnpm --filter @tloncorp/tlon-bot-e2e test:e2e:openclaw` — passed (scenario 267s). A first run failed in teardown only (gateway device-pairing scope upgrade on `cron disable`); the teardown was fixed to be presence-check-first (see Notes) and the re-run passed end to end.
-   `pnpm --filter @tloncorp/tlon-bot-e2e tsc` passes.
-   `pnpm --filter @tloncorp/tlon-bot-e2e test -- run`: 108/108 tests across 16/16 files.
-   `python3 -m unittest test_e2e_config` in `packages/hermes-tlon-adapter`: 3/3.
-   Prettier clean on all changed files.

## Risks and impact

Notes:

-   OpenClaw teardown is deliberately presence-check-first: the gateway CLI resolves least-privilege operator scopes per method (`cron.list` READ, `cron.update`/`cron.remove` WRITE), and the container's CLI device identity pairs with the scopes of its first call — a later WRITE is a scope upgrade requiring pairing approval. On passing runs the one-shot has already self-removed (`deleteAfterRun`), so cleanup lists first and skips gateway WRITE calls entirely; mutations only run when a job is actually left behind (a path where the scenario has already failed). This was found and fixed via the real e2e runs above.
-   CI runs the cron partition on **both** drivers, so this is a real regression gate rather than a manual-only test: `hermes-ci.yml` gains a `TLON_BOT_E2E_SCENARIO_PARTITIONS=cron` step, and `openclaw-ci.yml`'s shared-E2E step moves to `baseline,cron`. Each adds roughly one extra compose-stack boot (~5-8 min) to jobs that run ~13-15 min against a 35-minute timeout.
-   The remaining `RuntimeCapability` values (`image_search`, `upload_storage`, `media_blob`, `external_credentials`) have no shared scenarios declaring them, so there is no partition to run for those yet — that coverage currently lives in the OpenClaw package suite (`test/cases/06-media.test.ts`, `07-blobs.test.ts`), which already runs in CI with the configured secrets. Writing shared scenarios for those capabilities is a separate follow-up.
-   `docker-compose.hermes.yml` shows four incidentally reformatted `HERMES_MODEL_*` lines (Prettier reflowed pre-existing lines when the file was touched). YAML-parsing both revisions confirms the only semantic change is the added `HERMES_E2E_ENABLE_CRONJOB` key.
-   Two pre-existing harness weaknesses were left out of scope and deserve separate follow-ups: (1) `compose.ts`'s `down()` runs with `allowFailure: true` and no timeout, so a failed `docker compose down -v` is silently treated as success; (2) `sanitizeRunId` truncates to 32 chars both before and after appending the partition suffix, so a 32-character explicit `TLON_BOT_E2E_RUN_ID` would drop `-baseline`/`-cron` and collide two partitions on one compose project name.

## Rollback plan

Revert


arthyn (approved):
lgtm

PR #6168, opened 2026-07-21
openclaw: reduce diagnostic output from plugin

latter-bolden:
## Summary

We need to cut down on volume of OTLP diagnostics we're capturing from OpenClaw gateway before deploying the latest plugin across the fleet. Estimate is it'll emit 70% fewer logs, 60% fewer bytes stored.

## Changes
- Removed successful poke and SSE event acknowledgement logs; failures remain observable
- Stopped logging routine channel-discovery polls, unchanged results, and repeated pending-invite counts
- Suppressed unchanged five-minute settings snapshots while retaining the startup snapshot and real updates
- Promoted discovery and settings failures to error-level logs
- Added regression tests confirming routine success paths stay silent and failures are still logged

## How did I test?

Confirmed runs locally using `pnpm dev` Docker setup.

## Risks and impact

Fixes TLON-6208


packages/openclaw/src/urbit/sse-client.ts
```diff
@@ -371,9 +371,6 @@ export class UrbitSSEClient {
       if (eventId > this.lastHeardEventId) {
         this.lastHeardEventId = eventId;
         if (eventId - this.lastAcknowledgedEventId > this.ackThreshold) {
-          this.logger.log?.(
```
patosullivan, line 374: Thank you

patosullivan (approved):

PR #6175, opened 2026-07-22
hermes: resume the eyre channel on SSE reconnect instead of abandoning it

patosullivan:
## Summary

Fixes TLON-6171

TLON-6171 was filed as an E2E flake: `known bot loop protection resets on human dispatch` failing with `Timeout waiting for model call(s) … after 90000ms`, eight times across six branches, blamed on the `ships` container dying mid-suite. The container never died. The adapter was losing the message. On any SSE stream fault, including the aiohttp `sock_read` idle timeout, it abandoned its Eyre channel and `open()` minted a fresh id, while the ship went on holding the old channel and the events buffered on it, with no `Last-Event-ID` ever sent to retrieve them. Nothing in the system errors on this path, which is why the only available symptom was a timeout four hops downstream. Transport faults on an established stream now resume the same channel; rebuild is reserved for subscription nack/quit, channel reap, and setup failures. The same window silently drops real mentions and DMs to hosted bots.

## Changes

-   On a transport fault with an established stream, re-GET the same Eyre channel with `Last-Event-ID` instead of opening a new one. Rebuild is now reserved for subscription nack/quit, channel reap (404), and any failure during setup.
-   Added `TlonChannelError` carrying the HTTP status so the adapter branches on data rather than message text. A fixed-cookie 401/403 is fatal (re-auth cannot mint a new cookie); 404 still rebuilds.
-   Drop events with `id <= last_heard` client-side, since Eyre redelivers unacked events on resume. `_mark_seen` covers the rebuild case where ids restart at 0.
-   Reset reconnect backoff when the stream is established rather than on the first event, so an idle-but-healthy stream no longer ratchets its delay.
-   Release the SSE resources on any setup failure, and close the event generator when the stream loop exits.
-   `sse_reconnect` telemetry gains a `mode` field (`resume` / `rebuild`).
-   Added `disconnectComposeNetwork`, `connectComposeNetwork` and `readComposeServiceLogs` to the tlon-bot-e2e Docker runtime, plus a hermes-only `sse-resume-replays-events-missed-while-disconnected` scenario.

Scope: this is lossless for ordinary transport reconnects only. Messages missed while the bot process is down stay out of scope (TLON-6098). OpenClaw's own client has the same fresh-channel-on-reconnect behavior and needs its own ticket; it isn't exposed today because its stale-stream watchdog is 180s rather than a 60s read timeout.

## How did I test?

-   735 adapter unit tests pass, 21 of them new, covering resume/rebuild classification, cursor handling, replay dropping, resource release, auth modes and dedup.
-   Mutation-tested the unit suite: removing the `Last-Event-ID` header, swapping ack-before-drop ordering, reverting resume to rebuild, and dropping the fatal-path cleanup were each caught.
-   tlon-bot-e2e: tsc clean, 91 vitest pass, prettier clean.
-   Ran the new e2e scenario against local Urbit ships in Docker both ways. With the fix stashed it fails after 176.6s with the original flake's exact signature (20 passed, 1 failed); with the fix it passes in 89.5s (21 passed). The scenario detaches the bot container from the compose network, posts a mention while it is isolated, waits to observe the stream fault in the container logs, reconnects, and asserts delivery.
-   Confirmed the ticket's original theory was wrong: all three ships answer `/~/login` with 200 after the failure, and the `ConnectionRefusedError` lines are gateway boot noise that also appears in passing runs.

## Risks and impact

Main risks are double-dispatch on replayed events (guarded twice, by the client-side id filter and by `_mark_seen`) and misclassifying a subscription fault as a transport fault (subscription nack/quit raise `TlonChannelError` and stay on the rebuild path).

## Rollback plan

Revert


packages/hermes-tlon-adapter/adapter.py
latter-bolden, line 1: **Reading the HTTP 500 body can defeat dead-channel recovery.** 

The code awaits resp.text() before raising TlonChannelError. A stalled/truncated body can raise a timeout or payload error first, sending the adapter down the generic “resume this channel” path indefinitely. Since status 500 is sufficient—and @tloncorp/api resets immediately—raise before consuming the body or use a strictly bounded, nonessential

latter-bolden, line 1: *Invalid fixed cookies are only fatal after startup.** 

The new fatal check lives in _run_stream, but initial startup calls _connect_sse() from connect() first. A rejected TLON_COOKIE therefore falls through the generic connect handler without setting a fatal error, causing retry/restart churn. I reproduced connect_result=False with fatal_error=None. Handle the same TlonTerminalActionError 401/403 case in connect(), ideally through a shared helper.

latter-bolden (approved):
Makes sense, this'll be good to get in!

Flagging 2 comments my local review turned up, otherwise lgtm

packages/hermes-tlon-adapter/adapter.py
```diff
@@ -2466,20 +2493,51 @@ async def _run_stream(self) -> None:
...
+                        self._set_fatal_error("auth", str(exc), retryable=False)
+                        return
+                    logger.warning("[tlon] SSE setup failed, retrying: %s", exc)
+                    await _backoff_and_report(exc, mode="rebuild")
+                else:
+                    logger.warning(
+                        "[tlon] SSE stream error (resuming from event %s): %s",
+                        self._sse.last_heard_event_id, exc,
+                    )
+                    await _backoff_and_report(exc, mode="resume")
```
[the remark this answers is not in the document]
patosullivan: This is an edge case. Deferred, filed an issue for it: TLON-6240.

PR #6188, opened 2026-07-23
activity: notes support

arthyn:
## Summary

Adds notes to the `%activity` system: `%notes` now reports note creates, edits, and deletes to each member ship's `%activity`, which tracks per-note read state and rolls unreads up note → notebook → (group | base). This is the backend half of unread dots and notifications for notebook channels; the client work is stacked on top in a follow-up PR.

Part of TLON-6197.

## Changes

**`%activity` v10** (`sur/activity-ver` freezes v9; current sur updated):

- New source kinds `[%notebook flag group=(unit flag)]` and `[%note id=@ud notebook=flag group=(unit flag)]` with static parent chains `note → notebook → (group | base)`. Group affiliation is embedded in the source (same trick as `%channel`), so `lib/activity`'s hierarchy stays static — `%activity` never learns about notes folders; folder-path rollup is a client concern.
- New event kinds `%note-create` / `%note-edit` carrying `[id folder notebook group title author]`. They default to `[unreads=& notify=&]`, matching `%post`. Note events carry no message-key, so they count like `%group-ask` in unread summaries (no unread-point anchor).
- v10 → v9 conversions in `lib/activity-conv` drop notes sources/events, so `/v4`/`/v5` subscribers (existing clients, `%steward`, `%presence`) are unaffected.
- New `/v6` subscription + scry surface alongside `/v4`/`/v5` (feed at `/v7`), new marks (`activity-action-2`, `update-6`, `full-6`, `summary-6`, `settings-2`, `stream-2`, `event-2`, `feed-7`, `feed-init-7`), state migration to `%11` with `%base` volume-map backfill.
- New scry `/v6/activity/notes/{host}/{name}` returning per-note summaries for a notebook.
- `app/notify` moves to `/v6/notifications` + `activity-event-2` (state-9 stores v10 events) so note events reach push.

**`%notes` emission:**

- Every ship that applies a note mutation submits to its own `%activity` (host via `+se-update`, subscribers via `+no-response`); self-authored changes submit nothing, which also covers the echo of your own edit returning from the host.
- Creates and edits are **trail-debounced**: each change (re)arms a behn timer for `+edit-activity-window` (~m2) past it, with the note's `updated-at` encoded in the wire. At fire, `+note-activity-wake` submits only if the wire's `updated-at` still matches the note — newer changes re-arm with their own timers, so stale timers no-op. No timer bookkeeping, no new state (`%notes` stays on state-15). The event kind is picked from the record at fire time: untouched-since-creation → `%note-create`, else `%note-edit`, so the client's create-then-immediately-title flow resolves to one event instead of two notifications.
- Deletes clean up: note delete → `%del` of the note source; notebook delete/leave → `%del` of the notebook source (children included); recursive folder deletes clean up their notes' sources host-side, where the deleted ids are still known.
- Submissions are guarded by a `%gu` liveness check on `%activity` and ride a fire-and-forget `/activity/submit` wire.

**Also:** `app/channels` pins its `%bump` source construction to v9 types (it pokes with the v9 mark).

## How did I test?

- Agent tests on a dev ship: notes suite 83/83 (new coverage: debounced create, self-silence, create+edit collapse with stale-timer no-ops, note/notebook delete hygiene) and activity suite 9/9 (note-event rollup through notebook → group → base with default notify, standalone notebooks parenting to base, del-event coalescing, `read %all` clearing the chain, notebook `%del` dropping note children).
- Live dojo verification: note events roll up with correct counts at every level, reads zero the chain, sources delete cleanly.
- End-to-end with the stacked client PR: dots, notifications, sidebar reordering, and read-clearing against a dev ship.

## Risks and impact

- Safe to rollback without consulting PR author? (Yes — see rollback plan)
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [x] Notifications
  - [x] Other: `%activity` type versioning (v9 → v10), `%notify` subscription surface

Main risk is the v10 versioning pass itself: every existing consumer stays on `/v4`/`/v5` and receives down-converted updates with notes data dropped, so old clients see no behavior change. `%activity` state migrates `%10 → %11` (widening only); `%notify` migrates state-8 → state-9 (event wrapper widening only). `%notes` gains no state.

## Rollback plan

Revert the merge commit. Pre-deploy, reverting is consequence-free since nothing else emits v10 events; post-deploy, ships that have stored note sources/events would need those dropped (`%del` per notebook source) before an old `%activity` could load, so prefer roll-forward once released.

## Screenshots / videos

Backend-only; see the stacked client PR for visuals.


desk/app/notify.hoon
```diff
@@ -424,11 +447,20 @@
         (activity-event-to-json time-id u.event-8)
       ::
           %activity-event-1
+        ?~  event-9=(v9:event:v10:ac event.u.event)
+          [[404 ~] ~]
```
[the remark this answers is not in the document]
arthyn: addressed in https://github.com/tloncorp/tlon-apps/pull/6200

desk/lib/activity-conv.hoon
```diff
@@ -2,6 +2,188 @@
...
+      `[u.src (v9:^activity-summary activity-summary)]
+    --
+  ++  activity-summary
+    |%
+    ++  v9
+      |=  =activity-summary:v10:av
+      ^-  activity-summary:v9:av
+      :*  newest.activity-summary
+          count.activity-summary
+          notify-count.activity-summary
```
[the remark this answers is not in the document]
arthyn: Declining: this matches the accepted pattern for dropped event kinds — the v9→v8 conversion has kept react/contact-derived counts in legacy summaries since reactions shipped. If we ever fix it, it should be generic across dropped kinds, not notes-specific.

desk/app/activity.hoon
```diff
@@ -494,18 +527,21 @@
     =/  =feed:a  (feed type.pole start count)
     ?-    ver.pole
         %v5
-      ``activity-feed-5+!>((v8:feed:v9:ac feed))
+      ``activity-feed-5+!>((v8:feed:v9:ac (v9:feed:v10:ac feed)))
     ::
         %v6
-      ``activity-feed-6+!>(`feed:v9:av`feed)
+      ``activity-feed-6+!>((v9:feed:v10:ac feed))
```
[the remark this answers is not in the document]
arthyn: Declining: same pre-existing pattern — v5 feed pages built from v9 data already thin out when the window is full of react/contact events. Low exposure for notes; a generic over-fetch would be the right fix if it ever bites.

desk/app/notes.hoon
```diff
@@ -2554,6 +2698,13 @@
       (emil ?~(rep ~[stream] ~[stream u.rep]))
     ::
         %update
+      =?  cor  ?=(%note -.u-notebook.update.response)
+        %:  note-activity
+          flag
+          group.notebook-state
+          id.u-notebook.update.response
+          u-note.u-notebook.update.response
+        ==
```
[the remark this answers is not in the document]
arthyn: Declining for now: acceptable gap — notebook content still catches up via the snapshot; only unread/notification state for changes missed while unsubscribed is skipped. Snapshot diffing can be added later if this bites in practice.

desk/sur/activity-ver.hoon
```diff
@@ -11,6 +11,201 @@
...
+  ::  $activity-summary: depends on $source
+  +$  activity-summary
+    $~  [*@da 0 0 | ~ ~ ~]
+    $:  newest=time
+        count=@ud
+        notify-count=@ud
+        notify=_|
+        unread=(unit unread-point)
+        children=(set source)
+        reads=*  ::  DO NOT USE, 🚨 ⚠️ REMOVE
```
mikolajpp: Can we finally remove it in the current body of types during this migration? I don't think there is a reason to keep it anymore?

desk/tests/app/activity.hoon
```diff
@@ -27,17 +27,152 @@
   ;<  *  bind:m  (do-init dap activity-agent)
```
mikolajpp, line 27: I think at this point it is fair to settle that we are not required to review tests during PR reviews: we tend to generate them and it should be the responsibility of the author to make sure they do what they claim, given that it is pretty much the only bridge between the model generated code and reality.

Personally I plan to do light skimming to check we are not contaminating codebase with slop-style, and *maybe* unleash my own model to review, but nothing more beyond that. Time is better spent on the core of changes.

CC @Fang-

desk/tests/app/notes.hoon
```diff
@@ -2287,6 +2319,290 @@
   ?~  vis
     |+['snapshot missing visibility field']~
   &+[~ s]
+::  ====  %activity submission tests  ========================================
+::
```
mikolajpp: Needs a pass to remove headers.

desk/sur/activity.hoon
```diff
@@ -162,6 +164,19 @@
...
   ==
+::  $note-event: a note created or edited in a notebook
+::
+::    .notebook mirrors the shape of a group flag: host ship + name.
+::    .group is set when the notebook is a group channel.
+::
++$  note-event
+  $:  id=@ud
+      folder=@ud
+      notebook=flag:gv
```
mikolajpp: I did not notice we reuse the same type for notebook ids as we do for groups. We should at least define the flag type separately in `/sur/notes.hoon` and use that.

desk/sur/activity.hoon
```diff
@@ -172,6 +187,8 @@
       [%dm =whom]
       [%dm-thread key=message-key =whom]
       [%contact who=ship]
+      [%notebook =flag:gv group=(unit flag:gv)]
```
mikolajpp:
```suggestion
      [%notebook =flag:nv group=(unit flag:gv)]
```
When we follow advice above.

desk/sur/activity.hoon
```diff
@@ -172,6 +187,8 @@
...
+      [%note id=@ud notebook=flag:gv group=(unit flag:gv)]
```
mikolajpp:
```suggestion
      [%note id=@ud notebook=flag:nv group=(unit flag:gv)]
```
Same as above.

desk/sur/activity-ver.hoon
```diff
@@ -11,6 +11,201 @@
...
+        [%group-invite & &]
+        [%group-ask & &]
+        [%flag-post & &]
+        [%flag-reply & &]
+        [%group-kick & |]
+        [%group-join & |]
+        [%group-role & |]
+        [%contact | |]
+        [%note-create & &]
+        [%note-edit & &]
```
mikolajpp, line 205: By default we both notify and show unreads? Would not this trigger every ~10s or so when someone keeps editing their notebook?

desk/sur/activity.hoon
```diff
@@ -108,6 +108,8 @@
       [%flag-post key=message-key channel=nest:cv group=flag:gv]
       [%flag-reply key=message-key parent=message-key channel=nest:cv group=flag:gv]
       [%contact contact-event]
+      [%note-create note-event]
+      [%note-edit note-event]
```
mikolajpp, line 113: Do we plan to add notifications about notebook and folder creation later? Only notifying about notes feels pretty basic.

desk/lib/activity.hoon
```diff
@@ -49,14 +60,20 @@
...
+      %group     |=  [[src=s *] out=(list s)]  =;  f=?  ?:(f [src out] out)
+                 ?|  &(?=(%channel -.src) =(flag.source group.src))
+                     &(?=(%notebook -.src) =(`flag.source group.src))
+                 ==
+      %channel   |=  [[src=s *] out=(list s)]  =;  f=?  ?:(f [src out] out)
+                 &(?=(%thread -.src) =(nest.source channel.src))
+      %dm        |=  [[src=s *] out=(list s)]  =;  f=?  ?:(f [src out] out)
+                 &(?=(%dm-thread -.src) =(whom.source whom.src))
+      %notebook  |=  [[src=s *] out=(list s)]  =;  f=?  ?:(f [src out] out)
+                 &(?=(%note -.src) =(flag.source notebook.src))
```
mikolajpp, line 76: I forgot we have this particular skeleton in our closet, but apparently this is performance related. 

I hope you double checked it's correct because equality is wonderfully untyped in Hoon and we could be comparing two things which could never be equal, as happened many times in the past.

desk/lib/activity.hoon
```diff
@@ -310,19 +337,22 @@
...
+            ?=(?(%dm-post %dm-reply %post %reply %group-ask %note-create %note-edit) -<.event)
         ==
       $(stream rest)
     =.  total  +(total)
     =.  main   +(main)
     =?  main-notified  &(notify:volume notified.event)  &
-    ?:  ?=(%group-ask -<.event)
+    ::  note events carry no message-key, so they can't anchor
+    ::  an unread-point; count them like %group-ask
+    ?:  ?=(?(%group-ask %note-create %note-edit) -<.event)
```
mikolajpp, line 360: Isn't this something we want though? Seeing badges/unreads in some form on notebooks and folders, or even notes would definitely be useful.

desk/tests/app/activity.hoon
```diff
@@ -27,17 +27,152 @@
...
+  =/  note-src=source:v10:av  [%note 42 book ~]
+  =/  create=incoming-event:v10:av  [%note-create 42 7 book ~ 'T' ~wet]
+  =/  edit1=incoming-event:v10:av   [%note-edit 42 7 book ~ 'T' ~wet]
+  =/  edit2=incoming-event:v10:av   [%note-edit 42 3 book ~ 'T2' ~wet]
+  ;<  *  bind:m
+    (do-poke activity-action-2+!>(`action:v10:av`[%add create]))
+  ;<  *  bind:m
+    (do-poke activity-action-2+!>(`action:v10:av`[%add edit1]))
+  ;<  *  bind:m
+    (do-poke activity-action-2+!>(`action:v10:av`[%del-event note-src edit1]))
```
mikolajpp: `%del-event` is not actually emitted by the %notes agent, how could this feature pass a real test then?

desk/tests/app/notes.hoon
```diff
@@ -2287,6 +2319,290 @@
...
+  =/  act  (snag 0 acts)
+  ?.  ?=(%del -.act)
+    |+['expected a %del submission']~
+  ?.  ?=(%notebook -.source.act)
+    |+['expected %del of the notebook source']~
+  &+[~ s]
+::  a subscriber receiving %folder %deleted mirrors the host's recursive
+::  subtree removal and %dels the removed notes' activity sources — the
+::  wire update only names the folder
+::
```
mikolajpp, line 2584: This test does not seem to do what it claims: there is no recursive substree removal, we only test whether deleting a folder deletes its notes as well.

desk/lib/activity-conv.hoon
```diff
@@ -2,6 +2,188 @@
...
 |%
+++  v10
+  |%
+  ++  source
+    |%
+    ++  v9
+      |=  =source:v10:av
+      ^-  (unit source:v9:av)
+      ?:  ?=(?(%notebook %note) -.source)  ~
+      `source
```
mikolajpp, line 13: This will cause client using v9 api to display incorrect counts, as happened with previous changes to activity.

desk/app/notes.hoon
```diff
@@ -1234,6 +1372,9 @@
   ^+  cor
   =/  existing=(unit incoming-request:v1:n)  (~(get by requests) rid)
   ?^  existing  cor
+  ::  evict terminated records while we're here — the map only grows on
+  ::  registration, so sweeping at add time bounds it without a timer
+  =.  requests  (cleanup-requests now.bowl)
```
mikolajpp, line 1405: We never reach this part of code...
+handle-v1-post registers a request on its own, then calls +dispatch-v1-action which +register-request again, but the existing is always full.

desk/app/activity.hoon
```diff
@@ -182,12 +182,22 @@
...
-  ?>  ?=(%10 -.old)
+  =?  old  ?=(%10 -.old)
+    ::  insert missing volume defaults to %base
+    ::
+    =/  base-volume=volume-map:v10:av
+      (~(gut by volume-settings.old) [%base ~] default-volumes:v9:av)
+    =/  =volume-settings:v10:av
+      %+  ~(put by `volume-settings:v10:av`volume-settings.old)  [%base ~]
+      (~(uni by default-volumes:v10:av) base-volume)
+    old(- %11, volume-settings volume-settings)
```
mikolajpp: For a notebook anchored under a group, this migration will not properly set default volumes. The note events keys remain missing after the migration and will default to another set of defaults than what we set in default volumes.

mikolajpp (changes requested):
The PR looks complete, the code style is more in line with our standards which is good.

There are two bugs I found, rather serious in consequnces.

The cleanup logic was modified to remove the timers, but the new clean up logic does not actually fire, leaking space.

Secondly, the coalescing of activity event is not wired up in the agent: we never generate a `%del-event` in %notes. Surely this is an omission and explicitly contradicts the activity unit tests which pretends a behavior that is not real.

My takeaway from this review is that test-agent framework is generally ill-suited for agentic engineering era. We should really start leaning heavily towards shipping our code with aqua tests that actually exercise most, if not all of the implemented features on a real ship.

I am working on even better tooling in that regard, but there is enough existing tests and documentation to get started :-)

desk/sur/activity-ver.hoon
mikolajpp: Following up, I see there is a debounce window of 2m. I am still not sure if this is what we want (what if someones is editing for 10 minutes - doesn't this mean about 5 notifications?)

arthyn: potentially yeah, I think we have to feel this out. I think upping the timer to 5mins could be reasonable. would be curious what the team thinks, feels like something we just need to play with to really see what makes sense.

desk/sur/activity.hoon
arthyn: well we don't currently notify on creation of new channels which is the equivalent of notebooks so I wouldn't add that. folders seem like a weird thing to notify about, since they're generally just organization not actually meaningful content.

desk/lib/activity.hoon
arthyn: this is more about where to put the unread point in the backscroll of a channel which really doesn't map cleanly. the system as it stands will give badges/unreads on notebooks and folders recursively until you get to the note itself.

desk/tests/app/activity.hoon
arthyn: sorry this test didn't get updated after some changes I made

desk/lib/activity.hoon
arthyn: yep these are correct

desk/sur/activity.hoon
mikolajpp: I guess it is fair not to notify about new notebooks, but notifying about folders can be meaningful in itself. Imagine you are in a school group and teacher adds a "Exam Results" folder, or some such scenario. In many cases creating a folder in an established notebook can be more like posting.

desk/app/activity.hoon
```diff
@@ -190,7 +190,15 @@
     =/  =volume-settings:v10:av
       %+  ~(put by `volume-settings:v10:av`volume-settings.old)  [%base ~]
       (~(uni by default-volumes:v10:av) base-volume)
-    old(- %11, volume-settings volume-settings)
+    ::  v10 drops the unused reads field from activity-summary, so the
+    ::  stored summaries need reshaping, not just a head swap
+    ::
```
mikolajpp, line 192:
```suggestion
    ::  v10 drops the unused reads field from activity-summary, so the
    ::  stored summaries need reshaping.
    ::
```
A fine point in hoon commenting style: once you have used punctuation marks in a comment it should end with a period.

mikolajpp (approved):
LGTM!

desk/app/notes.hoon
```diff
@@ -1124,6 +1141,133 @@
...
+    ::  the wire carries the updated-at this timer was armed against, so
+    ::  +note-activity-wake can tell whether it is the note's latest
+    ::  change — no timer bookkeeping needed, stale timers just no-op
+    %-  emit
+    :*  %pass
+        %+  weld
+          /activity/edit/(scot %p ship.flag)/[name.flag]
+        /(scot %ud id)/(scot %da updated-at.note.upd)
+        %arvo  %b  %wait
+        (add updated-at.note.upd edit-activity-window)
```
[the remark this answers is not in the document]
arthyn: Known edge, accepting for now: the stale badge is bounded by the debounce window and self-heals the next time the note is opened. A read during the window has nothing to land on (the source does not exist yet) and %activity cannot backdate events, so a real fix means either recording read floors for not-yet-created sources or a client-side re-clear while viewing — deferring until we see whether it bites in practice.

PR #6193, opened 2026-07-24
openclaw: harden outbound image upload

patosullivan:
## Summary

Fixes TLON-6199: the OpenClaw Tlon plugin's outbound media handling silently fabricated success. From the field report: a bot generated an image, passed its server-local `/pier/...` path to the message tool, and the adapter logged `Invalid URL` — but still posted the message and reported `deliveryStatus: sent`. The bot then told the user the image was delivered while the user saw a broken block, since the posted image `src` was a path no client can reach.

Root cause: `uploadImageFromUrl` only understood http(s) URLs, and its catch-all returned the _original input string_ on any failure — local path, fetch error, or `No storage credentials configured` alike. `buildMediaStory` then matched the `.png` extension and emitted an image block anyway, and the post itself succeeded, so nothing upstream ever learned the media was broken. The plugin also ignored OpenClaw core's agent media-access machinery, so workspace-local file paths (the natural output of a bot generating an image) couldn't work in the first place, even though they work on other channel plugins.

Now media loads through core's guarded loader and uploads real bytes to Tlon storage; unrecoverable failures surface as a failed send with an actionable error, so the model can recover instead of claiming success (a remote-https upload failure falls back to hotlinking the original URL). Scoped to `packages/openclaw`.

## Why is this diff so large?

A fair question for "fix image upload" — the core fix (load through the guarded loader, fail loudly, classify by bytes) is only a couple hundred lines. The rest breaks down as:

-   **~69% is tests and fixtures.** The parsing and sanitization code here is adversarial-input handling, so it's tested against both real encoder output and deliberately malformed bytes. `test-fixtures.ts` is roughly split between hand-written malformed-container builders (the adversarial cases), a smaller block of base64 from genuine ffmpeg/cwebp-produced files, and comments — data and fixtures to skim, not logic to review.
-   **Error/filename sanitization (~a third of `upload.ts`).** The naive version of this fix leaks secrets: propagating loader errors into thrown messages echoes signed-URL query strings into channel messages, and deriving upload filenames from caller input echoes local paths into storage metadata. Review iterations kept finding new leak paths through raw external text, so the design removes it from all sinks — fixed category messages, fixed labels, synthetic filenames. That costs more lines than pattern-based redaction but is actually closeable.
-   **`image-dimensions.ts` (~540 lines) is a hand-rolled structural validator** for PNG/JPEG/GIF/WebP headers — no decoding, bounded walks only. We deliberately avoided adding an image-parsing dependency in a path that feeds attacker-influenced bytes to storage, and existing header-only libraries (e.g. `image-size`) read the dimension field without requiring the surrounding data/terminators, so they'd re-accept exactly the truncated/malformed bytes this fix rejects. It's a self-contained pure function (`bytes in → dimensions or null`), reviewable in isolation.

The comment layer and some dead generality were trimmed after review (redundant JSDoc, an unused filename sanitizer, a test-only export); the structural validation and the credential-leak boundaries were kept intact.

Suggested review order: `channel.runtime.ts` + `send.ts` (behavior wiring), `upload.ts` (policy), `image-dimensions.ts` (pure parsing), then spot-check tests.

## Changes

-   Replaced `uploadImageFromUrl` with `prepareOutboundMedia` in `src/urbit/upload.ts`: loads media via core's root-allowlisted, SSRF-guarded, byte-verifying `loadWebMedia` (preserving `hostReadCapability`), uploads the bytes, and throws on failure instead of posting a broken image.
-   `sendMedia` (`src/channel.runtime.ts`) now threads core's `mediaAccess` / `mediaLocalRoots` / `mediaReadFile`, so agent-workspace-local file paths work.
-   Added `src/urbit/image-dimensions.ts` (`parseRasterHeader`): structural header validation and real dimensions for PNG/JPEG/GIF/WebP; malformed bytes are rejected rather than posted at 0x0.
-   `classifyLoadedMedia` decides image-vs-link from sniffed bytes (not URL extension): only PNG/JPEG/GIF/WebP post inline; AVIF/HEIC/BMP/ICO and local SVG are rejected with a convert-to-PNG/JPEG error; remote SVG posts as a link; SVG is never inlined.
-   Security: error messages and logs no longer echo URL credentials or signed-query secrets; local source references collapse to a fixed placeholder; uploaded filenames for local sources are synthesized (never derived from caller path text); uploaded URLs are validated as credential-free https with no fragment.
-   `buildMediaStory` (`src/urbit/send.ts`) now takes prepared media and emits real dimensions; removed the extension-based `isImageUrl` from `src/urbit/story.ts`.
-   Updated tool-hint copy in `src/channel.ts` so the model knows the media contract and that failed media now fails the send.

## How did I test?

From `packages/openclaw`: `pnpm test` (1090 passing, 61 files), `pnpm tsc --noEmit` clean, `pnpm lint` 0 errors, `prettier --check` clean on changed files. New unit tests cover `parseRasterHeader`, `classifyLoadedMedia`, `prepareOutboundMedia` (mocked-remote and real-loader local-path via temp-dir fixtures), error-category and credential-safety paths, and `sendMedia` media-access threading. The storage-gated end-to-end image DM case (`test/cases/06-media.test.ts`) exists; the workspace-local variant was left out because there is no harness plumbing to place a file in the gateway agent workspace.

## Risks and impact

Behavior change: sends that previously "succeeded" now fail loudly (unreadable local paths, malformed images, AVIF/HEIC/BMP/ICO, local SVG, non-https/invalid uploaded URLs), and remote SVG now posts as a link instead of an inline image. This is intentional (the point is to stop fabricating success), but operators may see new tool-call failures where bots used to post broken or extension-matched blocks. Loss of AVIF/BMP/ICO inline support is a minor accepted regression.

## Rollback plan

Revert.

### DIFF
--- a/packages/openclaw/package.json
+++ b/packages/openclaw/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tloncorp/openclaw",
-  "version": "0.14.0",
+  "version": "0.15.0",
   "description": "OpenClaw Tlon/Urbit channel plugin",
   "license": "MIT",
   "repository": {
--- a/packages/openclaw/src/channel.runtime.test.ts
+++ b/packages/openclaw/src/channel.runtime.test.ts
@@ -1,0 +1,123 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Mock the media pipeline, the authenticated-api wrapper, and the send helpers
+// so sendMedia can be exercised without a ship. The account/target resolution
+// (types.js, targets.js) runs for real against a minimal valid config.
+vi.mock('@tloncorp/api', () => ({
+  scry: vi.fn(),
+}));
+vi.mock('./urbit/upload.js', () => ({
+  prepareOutboundMedia: vi.fn(),
+}));
+vi.mock('./urbit/api-client.js', () => ({
+  withAuthenticatedTlonApi: vi.fn(),
+}));
+vi.mock('./urbit/send.js', () => ({
+  buildMediaStory: vi.fn(() => [{ inline: [''] }]),
+  sendDm: vi.fn(),
+  sendDmWithStory: vi.fn(),
+  sendChannelPost: vi.fn(),
+}));
+vi.mock('./monitor/index.js', () => ({
+  monitorTlonProvider: vi.fn(),
+}));
+vi.mock('./setup-surface.js', () => ({
+  tlonSetupWizard: {},
+}));
+
+const cfg = {
+  channels: {
+    tlon: {
+      ship: '~zod',
+      url: 'http://localhost:8080',
+      code: 'lidlut-tabwed-pillex-ridruc',
+    },
+  },
+} as never;
+
+describe('sendMedia', () => {
+  let logSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(async () => {
+    vi.clearAllMocks();
+    logSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
+    const { withAuthenticatedTlonApi } = await import('./urbit/api-client.js');
+    vi.mocked(withAuthenticatedTlonApi).mockImplementation(
+      async (_params, fn) => fn()
+    );
+    const { sendDmWithStory, sendChannelPost } = await import(
+      './urbit/send.js'
+    );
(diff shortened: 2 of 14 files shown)


patosullivan:
This was just too overwrought and didn't really do enough to address the real problem. Will put up a new PR soon.

PR #6194, opened 2026-07-27
logs: modify send-otel thread to use base alloy url

mikolajpp:
## Summary

In the initial deployment we have used the full alloy url including the endpoint. This has caused some confusion on the hosting side, so we modify `-send-otel` thread to accept base alloy url and append the fixed endpoint by itself.

Note that deploying this will require coordinating with hosting, since all ships will need to be reconfigured. This breaks grafana logging backend until reconfiguration takes place, but it is acceptable given that Posthog remains our default telemetry dashboard for now.

## Changes

1. Assume `otel` url is the bare alloy address.

## How did I test?

Deployed and tested on a hosted moon.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

## Rollback plan

Revert.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ dump.sqlite3
 clurd
 
 .obsidian
+.pi
--- a/.prettierignore
+++ b/.prettierignore
@@ -39,6 +39,7 @@ dump.sqlite3
 .gitattributes
 clurd
 .obsidian
+.pi
 # From apps/tlon-desktop/.gitignore
 /apps/tlon-desktop/build/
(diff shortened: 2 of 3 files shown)


mikolajpp:
After talking with @yapishu we decided it would be more robust to handle both kind of URLs. Implementation pending.

arthyn (approved):

yapishu (approved):
lgtm!

arthyn (approved):

PR #6197, opened 2026-07-27
TLON-6253 Register OAuth credentials and models

yapishu:
- Adds provider connection, status, model discovery, and disconnect routes
- Stores OAuth credentials in sqlite auth-profile store
- Discovers OpenAI models directly from the connected ChatGPT account so the UI only shows models available to that  subscription
- Exposes Anthropic subscription models through OpenClaw’s provider catalog
- Detects expired or invalid credentials and reports when reconnection is required
- Handles the root-owned managed config without attempting to write openclaw.json
- Registers the bundled OpenAI and Anthropic provider runtimes through an in-memory allowlist overlay
- Ensures OAuth-backed OpenAI models use the ChatGPT subscription transport rather than the API transport
- Adds diagnostics for unavailable subscription provider runtimes

Managed OpenClaw 7.1 installations use a restrictive plugin allowlist. Provider models could appear in `/models`, but their runtime hooks were not registered, causing inference to fail with errors like `FailoverError: Unknown model: openai/gpt-5.6-luna`

The OAuth profile was present in openclaw-agent.sqlite, but model resolution failed before authentication was consulted. The plugin registers the required runtime hooks without modifying the OC configuration or tlawn.py


yapishu:
closing in favor of standalone plugin

PR #6209, opened 2026-07-30
groups: add group-level custom blob field

arthyn:
## Summary

Adds an opaque, admin-writable `blob=(unit @t)` field to `$group` in the %groups agent — a group-level analogue of the channel-level custom `meta` field — for client-owned group config and bot/automation metadata, plus full frontend plumbing.

Fixes [TLON-6274](https://linear.app/tlon/issue/TLON-6274/percentgroups-add-group-level-opaque-blob-field-custom-metadata-like)

CI will fail until we cut a new pier archives, waiting to do that until its confirmed (or we could PR having the harnesses pull in latest desk changes)

## Changes

**Backend**

* `blob=(unit @t)` on `$group`; `%blob` variants on `a-group`/`c-group`/`u-group`/`r-group`; new `++ v11` dependency closure in `sur/groups-ver`
* Host setter (admin-gated, 256KB size cap, idempotent) and client apply; responses ride a new `/v3/groups` lane (`group-response-3`) emitted first, with `groups-conv` downgraders keeping v2/v1/v0 blob-free (same playbook as `%active-channel` / 39458f2ba)
* Actions ride a new strict `group-action-5` mark (`a-groups:v11`); `group-action-4` stays frozen at v8 so older clients are unaffected
* State migration `%11 → %12` (blob defaults to `~`, pub logs re-walked for embedded `%create` groups); groups negotiate version `%2 → %3` (update/log/command wire shapes change), bumped in `%groups`, `%channels`, and `%channels-server`
* New read surfaces: `/x/v3/{groups,changes,groups/…,ui/groups/…}`, `/x/v4/init`, and `groups-ui` `/v9/init` + `/v10/changes` (activity intentionally stays v8 — the unconsumed init-8/activity-v9 bump is not bundled in); new marks `group-3`, `groups-3`, `group-ui-3`, `group/changed-groups-3`, `group/response-3`, `ui/init-9` registered in rail/discipline/mark-warmer

**TypeScript**

* Wire types for the `%blob` action/response; `updateGroupBlob` poke via `groupAction5()` tracked on `/v3/groups`; a third `subscribeGroups` subscription filtered to blob events (`.catch` for older backends)
* `getGroup`/`getGroups` moved to the v3 scries; init → `/v9/init`, changes → `/v10/changes`
* `groups.blob` db column (migration regenerated), `insertGroups` conflict set, `editGroupBlob` sync case, optimistic `updateGroupBlob` store action

## How did I test?

* `./backend/run-tests.sh` fully green on a fresh pier: 317 OK, **Unit tests passed ✅** and **Aqua tests passed ✅**, including a new `+test-c-group-blob` (admin set/clear, idempotence, non-admin/non-member nacks)
* `pnpm -r tsc` clean workspace-wide; api + shared vitest green, including new tests for the `toV1GroupsUpdate` blob arm and the `editGroupBlob` sync handler
* Live smoke test on the \~rapsed dev moon: `%11 → %12` migration ran cleanly against real group state; blob set + clear round-trips via `group-action-5`; `/v3` scries, `/v9/init`, and `/v10/changes` carry the blob while `/v2` surfaces stay blob-free

## Risks and impact

* Safe to rollback without consulting PR author? (Yes)
* Affects important code area:
  - [ ] Onboarding
  - [X] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [X] Other: %groups agent state migration + negotiate version bump

Notes: the negotiate bump pauses groups sync between ships until both sides OTA (same as the request-timestamps bump, 027013bed). New frontend requires the new desk (`/v3` scries and `/v9/init` are a hard cutover, matching prior init bumps). `%create` responses on v1/v2 are blob-stripped; a just-joined group briefly lacks its blob until init/changes/getGroup or the next blob update.

## Rollback plan

Revert the two commits. The desk state migration is forward-only (`%12`), so rolled-back code must not ship to ships that already ran it — roll forward with a fix instead for any ship that has OTA'd.

## Screenshots / videos

N/A — backend field + API plumbing; no UI changes.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


desk/lib/mark-warmer.hoon
```diff
@@ -6,6 +6,7 @@
 /$  init-5      %ui-init-5            %json
```
mikolajpp: Why do we touch this file? On 408 all marks are build by default, this should render it unnecessary. CC @pkova for confirmation.

desk/app/groups.hoon
```diff
@@ -45,7 +52,9 @@
           %channel-preview-1
           %group-response-1
           %group-response-2
+          %group-response-3
           %group-action-3
+          %group-action-5
```
mikolajpp, line 58: I think we forgot to add %group-action-4 at some point. We can do it by the way in this PR.

desk/app/groups.hoon
```diff
@@ -318,6 +333,23 @@
...
+        go-abet:(go-a-group:group-core a-group.a-groups)
+      ::
+          %invite
+        =/  group-core  (go-abed:go-core flag.a-groups)
+        go-abet:(go-a-invite:group-core [ships a-invite]:a-groups)
+      ::
+          %leave
+        =/  group-core  (go-abed:go-core flag.a-groups)
+        go-abet:(go-leave:group-core &)
+      ==
```
mikolajpp, line 349: This way of structuring code will bloat. Our established pattern is, yes, to handle things at the latest version, but at the same time we should short-circuit old pokes to the new one when possible. See the relationship between %group-action-3 and %group-action-4.

What we should do here:
1. Leave %group-action-5 as is
2. Deflate %group-action-4 by upgrading to %group-action-5. This should be possible as is, the old type nests under the new one.

desk/app/groups.hoon
```diff
@@ -348,7 +380,7 @@
           [%group flag [%delete ~]]
         $(+< group-command+!>(c-groups))
       ?:  ?=(%secret -.diff)
-        =/  =group:v9:gv  +:(~(got by groups) flag)
+        =/  =group:g  +:(~(got by groups) flag)
```
mikolajpp, line 367: yes, ok, i guess this survived thus far.

desk/app/groups.hoon
```diff
@@ -1231,48 +1285,62 @@
...
+    ``unsafe+noun+!>([groups-light-ui-11 foreigns-8])
+  ::
+       [%x ver=?(%v0 %v1 %v2 %v3) %groups ~]
+    =/  groups-11=groups:v11:gv  (~(run by groups) tail)
+    =*  groups-9  (~(run by groups-11) v9:group:v11:gc)
     ?-    ver.pole
         %v0  ``groups+(~(run by groups-9) v2:group:v9:gc)
         %v1  ``groups-1+(~(run by groups-9) v5:group:v9:gc)
-        %v2  ``groups-2+groups-9
+        %v2  ``groups-2+`groups:v9:gv`groups-9
```
mikolajpp: Why is this cast necessary? Rail gives us a typed interface, does it not?
```suggestion
        %v2  ``groups-2+groups-9
```

desk/app/groups.hoon
```diff
@@ -1231,48 +1285,62 @@
...
     =/  groups-9=groups:v9:gv
       %-  ~(run by groups)
-      |=  [=net:v9:gv =group:v9:gv]
-      (drop-seats:group:v9:gc group our.bowl)
+      |=  [=net:g =group:g]
+      (v9:group:v11:gc (drop-seats:group:v11:gc group our.bowl))
     ?-    ver.pole
         %v0  ``groups+(~(run by groups-9) v2:group:v9:gc)
         %v1  ``groups-1+(~(run by groups-9) v5:group:v9:gc)
         %v2  ``groups-2+groups-9
```
mikolajpp, line 1313: Is this endpoint in use by anything? If not, we should mark it for removal later. Otherwise it needs updating too.

desk/app/groups.hoon
```diff
@@ -1231,48 +1285,62 @@
...
+      |=  [=flag:g =net:g =group:g]
       =/  =status:neg
         (read-status:neg bowl [p.flag %groups])
-      (group-ui:v2:group:v9:gc status net group)
+      (group-ui:v2:group:v11:gc status net group)
     ::
-      %v1  ``groups-ui-1+(~(run by net-groups-9) group-ui:v5:group:v9:gc)
-      %v2  ``groups-ui-2+(~(run by net-groups-9) group-ui:group:v9:gc)
+      %v1  ``groups-ui-1+(~(run by groups) group-ui:v5:group:v11:gc)
+      %v2  ``groups-ui-2+(~(run by groups) group-ui:v9:group:v11:gc)
```
mikolajpp: Same comment as above, we don't expose `v11` here.

desk/app/groups.hoon
```diff
@@ -1301,16 +1375,19 @@
...
         %v1
-      ``group-1+(v5:group:v9:gc +.u.net-group)
+      ``group-1+(v5:group:v11:gc +.u.net-group)
     ::
         %v2
-      ``group-2+`group:v9:gv`+.u.net-group
+      ``group-2+(v9:group:v11:gc +.u.net-group)
+    ::
+        %v3
+      ``group-3+`group:v11:gv`+.u.net-group
```
mikolajpp: Rails are typed, we should not need casts anymore.

desk/app/groups.hoon
```diff
@@ -4586,18 +4687,29 @@
...
+    ::  v1 response: %blob and %active-channel are absent
+    ::
+    =/  r-group-9=(unit r-group:v9:gv)  (v9:r-group:v11:gc r-group)
+    ?~  r-group-9  go-core
+    =/  r-groups-9=r-groups:v9:gv  [flag u.r-group-9]
     =/  v1-paths  ~[/v1/groups [%v1 go-area]]
     =.  cor  (give %fact v1-paths group-response-1+r-groups-9)
-    =/  r-groups-10=r-groups:v10:gv  [flag r-group]
-    =/  v2-paths  ~[/v2/groups [%v2 go-area]]
-    =.  cor  (give %fact v2-paths group-response-2+r-groups-10)
```
mikolajpp: I don't like the structure here. The code paths for v9 and v10 do the same thing, why should they look different? I prefer the pattern with `?~` rather than `=?`, because we mostly want to send out the update, it's only the one case of %blob that's getting nullified.

mikolajpp (approved):
This looks good.

Some nits in the comments. Appreciate the less-verbose coding style :-)

jamesacklin:
Seems good to me, my stuff could write to this easily

desk/lib/mark-warmer.hoon
arthyn: Removed the whole file in f3d6e73f rather than just my additions — it is only `/$` runes, imported by `app/contacts` and `app/groups-ui` but never referenced, so on 408 none of it earns its keep. Shout if @pkova disagrees and I will restore it.

desk/app/groups.hoon
arthyn: Cast dropped, but the accessor needed a change too: `net-groups` is `(map flag [net group])`, a faceless tuple, so `q.u.net-group` is `-find.q`. It is now a plain `=/` binding of `+.u.net-group`.

desk/app/groups.hoon
arthyn: Checked — nothing reads it: no reference to `/x/vN/light/groups` or the `/x/groups/light` alias anywhere in `packages/`, `apps/`, or `desk/`. Deleted outright rather than marked, in the stacked #6224.

desk/app/groups.hoon
arthyn: Same story: the only reference to the plural `/x/vN/ui/groups` map was a request-timeout test using the path as filler. The in-desk create threads read the per-flag `/gx/groups/vN/ui/groups/<ship>/<name>`, which is a different arm and stays. So rather than expose v11 there, #6224 deletes the plural map. Left unresolved for your call.

packages/shared/src/store/groupActions.ts
```diff
@@ -490,6 +490,32 @@ export async function updateGroupMeta(
...
+) {
+  logger.log('updating group blob', group.id);
+
+  const existingGroup = await db.getGroup({ id: group.id });
+
+  // optimistic update
+  await db.updateGroup({ id: group.id, blob });
+
+  try {
+    await api.updateGroupBlob({ groupId: group.id, blob });
```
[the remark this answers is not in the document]
arthyn: Fixed in a0850077. `updateGroupBlob` now returns early when `(existingGroup.blob ?? null) === blob`, so no poke is issued and there is nothing to time out or roll back. Added four tests in `packages/shared/src/store/groupActions.test.ts` (unchanged, both-empty, changed, cleared); the two no-op cases fail without the guard. Worth noting `updateGroupMeta` has the same latent issue on develop, since `se-c-group`'s `%meta` arm short-circuits identically — left alone here as out of scope.

apps/tlon-web/rube/archive-piers.sh
```diff
@@ -123,10 +125,22 @@ get_current_version() {
...
     local ship=$1
+    if [ -n "$ARCHIVE_TAG" ]; then
+        echo "$ARCHIVE_TAG"
+        return 0
+    fi
     local current_version=$(get_current_version "$ship")
+    if [ -z "$current_version" ]; then
+        print_error "Cannot derive a version for ~$ship from $MANIFEST_FILE"
+        print_info "The manifest holds a non-numeric archive name; set ARCHIVE_TAG explicitly."
+        return 1
```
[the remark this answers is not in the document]
arthyn: Confirmed and fixed in a0850077 — thanks, this was a guard that did nothing. Reproduced both halves: `print_error`/`print_info` echo to stdout with no `>&2`, and `local x=$(f)` returns the status of `local`, so the caller got `status=0` with the diagnostics as the version string. Now the two prints are redirected to stderr (matching `archive_pier`'s existing per-call convention) and the caller does `local next_version` then `if ! next_version=$(...)` so the failure aborts. Verified: aborts with an empty variable.

apps/tlon-web/rube/archive-piers.sh
```diff
@@ -123,10 +125,22 @@ get_current_version() {
...
-# Function to increment version
+# Function to increment version, or use a literal tag when pinned.
+# ARCHIVE_TAG names archives after what they contain rather than a position in
+# the numbered lineage (e.g. ARCHIVE_TAG=-group-blob -> rube-zod-group-blob.tgz),
+# for piers tied to a branch.
 get_next_version() {
     local ship=$1
+    if [ -n "$ARCHIVE_TAG" ]; then
+        echo "$ARCHIVE_TAG"
+        return 0
```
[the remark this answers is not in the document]
arthyn: Fixed in a0850077: the pattern is now `rube-${ship}[^ ]*\.tgz`, verified to repoint an already-tagged name. One correction on the premise — the Dockerfile was not left stale on this run. `[0-9]*` matched the *old* numeric name still in the file, so the diff shows `rube-zod27.tgz` → `rube-zod-group-blob.tgz` correctly. The bug would have bitten the next tagged run, once the file held a non-numeric name.

desk/app/groups-ui.hoon
```diff
@@ -311,6 +327,21 @@
         'groups'^groups
         'contacts'^contacts
     ==
+  ::
+      [%x %v10 %changes since=@ ~]
+    =+  .^(activity=json (scry %gx %activity /v4/activity/changes/[since.pole]/json))
```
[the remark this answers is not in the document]
arthyn: Declining this one — `/v4` and `/v5/activity/changes` return the same bytes, so nothing is down-converted away.

Both `init` and `changes` carry exactly one activity-derived field: `$activity` = `(map source activity-summary)`. Activity v9 never defines `activity`, `activity-summary`, or `source` — it inherits all three from v8 through `=,  v8` (`sur/activity-ver.hoon:15-181`), so `activity:v9:av` and `activity:v8:av` resolve to the same definition.

Three confirmations:

- `lib/activity-conv.hoon:44-71` — `v8:activity:v9:ac` restates all seven summary fields verbatim and casts `source:v9` to `source:v8` with a bare cast. It's an identity function.
- `lib/activity-json.hoon:70-72` — v9's enjs core is `=,  v8` with no `activity-summary` override, so v9 serializes summaries via v8's encoder. That only typechecks because the types are identical.
- `mar/activity/summary-4.hoon` and `summary-5.hoon` differ only in the version face.

Reactions live in `$incoming-event`/`$event`, which v9 *does* redefine — but those ride the feed (`/v5|/v6/feed/…`) and subscription (`/v4|/v5`) endpoints, which the client already selects dynamically via `getActivitySupportsReactions()`. They never appear in the summaries map. Reaction-driven unread state still reaches the client on `/v4`, because reacts bump `newest`/`count`/`notify-count` in the agent's own v9-native state before either scry converts.

Leaving unresolved in case whoever owns %activity sees something I don't.

desk/mar/ui/init-10.hoon
```diff
@@ -0,0 +1,38 @@
...
+++  grow
+  |%
+  ++  noun  init
+  ++  json
+    =,  enjs:format
+    ^-  ^json
+    %-  pairs
+    :~  groups/(groups-ui:v11:enjs:gj groups.init)
+        foreigns/(foreigns:v8:enjs:gj foreigns.init)
+        activity/(activity:v8:enjs:aj activity.init |)
```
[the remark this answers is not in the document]
arthyn: Fixed in 2bed30e43 — `mar/ui/init-10.hoon` now calls `activity:v10:enjs:aj`.

One correction to the diagnosis: this was a build break, not a lossy encode. `activity:v10` does not nest in `activity:v8` (v10 adds `%notebook`/`%note` to `$source`), so the mark could not compile at all. The flip was made while resolving the develop merge but never staged — the file was not among the conflicted paths, so it was left out of the merge commit.

latter-bolden (approved):

PR #6234, opened 2026-08-07
tlon-skill: add a Hosted tlonbots section for the no-storage upload rule

arthyn:
Context: [TLON-6298](https://linear.app/tlon/issue/TLON-6298) (no auto-close — that issue tracks tlonbot e2e work; this is a spin-off finding).

A hosted tlonbot runs on a **moon**, and a moon has no storage of its own, so uploads have to authenticate as the owner ship. SKILL.md documented this only as a trailing sentence in the **OpenClaw** section — *after* two bare `tlon upload` examples — and the **Upload** reference section ~500 lines later didn't mention it at all.

But the rule isn't a property of OpenClaw. It's a property of *hosted tlonbots*, and applies equally under Hermes. So it gets its own section, and Hermes / OpenClaw / the Upload reference all link to it.

```
## Hermes            → link
## OpenClaw          → link
## Hosted tlonbots   ← the rule, and what it doesn't cover
## Installation
```

## Scope

Deliberately tight, because this is the **generic** skill doc — most people run the CLI against their own ship, which has storage:

- Upload reference examples stay **bare**; that's the correct default.
- Its new note states the constraint generically ("authenticate as a ship that has storage") without naming `$TLON_OWNER_CONFIG_PATH`.
- That variable is injected by tlonbot's `entrypoint/tlawn.py` and **is not defined anywhere in this repo** — the one pre-existing mention was the OpenClaw line. It now appears only in the Hosted tlonbots section, which also states what it does *not* cover: a self-hosted OpenClaw or Hermes deployment against a ship you control.

Earlier revisions of this PR pushed the owner-config form into the generic reference and the Hermes section; that was wrong for this reason and has been reverted.

## How it turned up

tlonbot's e2e nightly. On two of three `post-image` variants the bot called bare `tlon upload` first, got an error, then retried with `--config …/ships/<owner>.json` and succeeded:

| # | call | result | duration |
| --- | --- | --- | --- |
| 1 | `tlon upload <url>` | error | 433ms |
| 2 | `tlon --config …/ships/risref-hoplux.json upload <url>` | ok | 2925ms |

## Related — the misleading error, fixed but not shipped

That failure surfaced as `Error: Note: Credentials cached for <moon>` — the step that *succeeded*. The note is written to stderr during auth setup, so on failure stderr holds the note followed by the real error (`main.ts:222`), and a consumer surfacing the first stderr line reports the hint instead of the failure.

eeae99cb9 (2026-08-04) already gates that note on `process.stderr.isTTY`. It's on **`develop` but not `master`**, and unpublished: npm's latest `@tloncorp/tlon-skill` is 0.4.5 from 2026-07-29, and both branches still declare 0.4.5. Since `packages/openclaw` depends on `@tloncorp/tlon-skill: workspace:^` and plugin checkouts resolve that to a registry version, bots get the stale published build. Shipping it needs develop → master → version bump + publish; no code change.

Matching prompt fixes on the tlonbot side: tloncorp/tlonbot#168.

### DIFF
--- a/packages/tlon-skill/SKILL.md
+++ b/packages/tlon-skill/SKILL.md
@@ -37,7 +37,9 @@ and `posts send` to a current chat/DM conversation (reply normally instead).
 Current-gallery `posts send` creates a new item and is allowed. Image sends
 (`--image`) are allowed anywhere,
 including the current conversation: `tlon upload <direct-image-url>`, then
-`posts send <target> [caption] --image <uploaded-url>`.
+`posts send <target> [caption] --image <uploaded-url>`. If you're a hosted
+tlonbot, that upload needs the owner config — see
+[Hosted tlonbots](#hosted-tlonbots).
 
 ## OpenClaw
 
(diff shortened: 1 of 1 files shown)


latter-bolden:
Closing since we're giving bots their own memex soon, which should alleviate the need for owner upload.

PR #6235, opened 2026-08-07
notes: apply write responses and stream updates as deltas; batch + merge imports

arthyn:
## Summary

Notes mutations refetched a full notebook snapshot on every write — often several times per mutation — even though the backend already tells us exactly what changed. The `%ok` write envelope carries the applied `u-notebook` update with the complete entity, and the per-notebook stream broadcasts the same typed updates for every change. This applies both as deltas, so a mutation costs one request instead of 13+.

The same reasoning applies to import, which drove its own serial walk of `createFolder`/`createNote` calls even though `%notes` has had `batch-import-tree` all along. Import is now one poke, and the folder-merging that made it correct moved to the agent, where authoritative state actually lives.

Fixes TLON-6242

Fixes TLON-6311. Also fixes TLON-6307.

### 1. Mutations apply their own response (`66fad7f11`)

**Before**, `createNotebookNote` did: a baseline snapshot fetch (4 GETs + per-note hydration) → POST → `syncNotesNotebookUntil` poll (4+ GETs, up to 8 retries) → a second poll with hydration (5+ GETs). Then the stream fact for our own write triggered *another* full snapshot fetch. Move/rename/delete each cost a write plus a 4-GET poll loop.

**After**, each of those is one request whose response advances local state directly.

- **`packages/api`** — parse `u-notebook` payloads into a typed `NotesUpdate` union; write helpers return the update they applied instead of `void`; `NotesStreamEvent` delivers its payload parsed rather than as an opaque "something changed" signal.
- **`packages/shared/db`** — per-entity writes: `upsertNotesNote`, `upsertNotesFolder`, `replaceNotesMemberRoles`, `updateNotesNotebook`. `upsertNotesNote` carries the same revision-monotonic guard as `updateNotesNote` and the snapshot merge, expressed as `ON CONFLICT DO UPDATE ... WHERE` so the compare and the write are one atomic statement.
- **`packages/shared/store`** — `applyNotesUpdate`, one reducer shared by the write path and the stream handler, serialized per notebook so overlapping updates commit in arrival order (folder rows have no revision to break ties with).

Full snapshot sync stays for initial load, reconnect (`snapshot` events), and as the fallback whenever a response or fact carries no update we can model.

### 2. Import uses the batch API (`7b95fd633`)

`useNotesImportController` walked its items one at a time — a `createNotebookFolder` per missing path segment, a `createNotebookNote` per file, all sequential. A 200-note folder import was 200+ serial round trips.

`se-batch-import-tree` takes a parent folder and a nested folder/note tree, creates the lot host-side, and emits a `%created` update per entity — which the new stream handling already applies. `batchImportNotesTreeV1` submits it; `generateNotesRequestId` mints the `@uv` the envelope requires; `importNotebookTree` pokes once and syncs once (the `%ok` envelope carries only the *last* of the many updates the poke emits, so converging on the stream's copies beats reconstructing the batch).

### 3. Folder merging moved to the agent (`8e817f27f`)

The backend created every folder in a submitted tree unconditionally, so importing `docs/a.md` into a notebook that already had `docs/` produced a second `docs/`. Commit 2 worked around this client-side, walking existing path prefixes and anchoring each batch at the deepest folder it believed existed.

**That workaround could never be correct.** It planned against the client's own replica, which can be one unsynced fact behind the folder list the host will actually merge against — so a folder another client created moments ago still got duplicated. Only the agent sees authoritative state at apply time.

`+se-batch-import-tree` now merges a `%folder` node into an existing same-named child instead of creating one:
- Case-insensitive, preserving what the import feature already did (`normalizeTitleKey` lowercased).
- Ties resolve to the lowest folder id — duplicate sibling names are legal (`se-create-folder` permits them) and map traversal order isn't stable.
- A merge emits no `%created` update, because nothing was created.

**TLON-6307, same arms:** `se-batch-import` assigned its `folder` id into every imported note without checking it existed, and `se-batch-import-tree` had the identical gap on `parent`. Either persists a batch of notes that are real but invisible to folder traversal. Both now assert, matching `+se-create-note`'s contract. That issue's suggested client-side mitigation was explicitly "a pre-flight courtesy check"; this is the actual fix.

Frontend follows: `planNotesImport`'s ~90 lines of anchoring collapse to `buildNotesImportTree`, the now-unused `folders` prop leaves the import hook, and an import is always **one** poke rather than one per existing folder it touched.

### Deliberate behavior changes

- Analytics for move/delete/rename now fire on the write's `%ok` response rather than on read-back convergence. The response *is* the host's confirmation — the old poll only watched our own replica catch up, so a lagging replica silently suppressed the event.
- Per-note hydration (`getNote` per created note) is gone: update payloads carry `bodyMd` directly.
- Import title deduping still applies to the target folder and within the import, but no longer to nested folders — the host may merge into one whose contents the client can't know.
- Import error messages name a batch rather than a single file, since a batch is what succeeds or fails now.

### Kept on purpose

The client's pre-flight folder GET before a batch import is redundant against the new assertion, but the version-skew failure modes are asymmetric: an old host without the merge gives duplicate folders (visible, fixable), while an old host without the assertion silently buries a whole batch under a dangling id. One GET *per import* is cheap insurance, and it yields a typed `NotesUnknownFolderError` rather than a bare crash from the poke.

Publish/unpublish still confirm via the `/v0/published` scry — those go through a `%notes` poke, not a v1 route, so there's no envelope to read, and the confirmation gates a "link copied" signal that must fail loudly. Wiring them to the stream's `note-published` fact needs an event bus the plain async actions can't reach today. Follow-up.

## How did I test?

**Desk (`backend/run-tests.sh`): 310 tests OK, 0 failures**, 109 in `/tests/app/notes`, including five new ones for folder merging (existing folder, case-insensitive, siblings-only) and the TLON-6307 id assertions on both batch arms. The merge test is load-bearing by construction: without merging, `Sub` takes id 4 and the note id 5, so both the folder-count assertion and the note-id lookup fail.

**JS unit tests: 1394 pass** (`api` 650, `shared` 395, `app` 349). Covers update parsing across every variant, unmodeled/malformed payloads degrading to `null`, stream events arriving parsed, `applyNotesUpdate` per variant (incl. folder-subtree deletion cascading to notes, current-user role, concurrent updates converging on the highest revision), and `buildNotesImportTree`. The DB revision guard was checked by disabling its `WHERE` clause and confirming the stale-update test fails.

**E2E (`notes-channel.spec.ts`, new — no Playwright spec covered `%notes` before this):**

- ✅ **mutations apply their write response** — create folder, create note, rename, delete; each asserts ≥1 write and **zero** notebook-scoped GETs in its window. Green against live ships across five separate runs.
- ✅ **import merges into an existing folder** — creates `docs/`, drops `docs/first.md`, `docs/second.md`, `top.md`; asserts exactly one `docs` row and exactly **one** write for the whole import. The `se-batch-import-tree` merge, end to end. Green across the same runs.
- ⚠️ **cross-ship stream convergence** — written but **never verified locally**. It fails inside the shared `acceptGroupInvite` helper, whose fixed 15s wait for `ChannelListItem-General` is short for cross-ship sync on my machine; its own assertions were never reached. Other specs use that helper successfully in CI, so **CI is the first real signal for this test** — if it proves flaky there, it should be dropped rather than papered over.

Three harness facts the spec had to accommodate, each commented in place: destructive notes actions confirm via `window.confirm` (Playwright dismisses dialogs by default, silently cancelling every delete); opening a channel kicks off an initial snapshot sync whose GETs would otherwise land in the next mutation's window; and a just-joined group's channel list churns, so the open is retried as a whole.

Note the cross-ship test runs **first** in the file on purpose — `createGroup`'s fallback identifies its group by the shared "Untitled group" label, which stops being unique once a second exists, and the fixtures' cleanup cannot reliably remove them from the desktop group list. That is a pre-existing sharp edge any spec creating two groups would hit.

`pnpm -r tsc` clean. (The `@tloncorp/editor/dist/editorHtml` error some may see is a missing build artifact, not a code error — `pnpm build:packages` fixes it, and it also blocks the Vite dev server used by e2e.)

## Risks and impact

  - [x] State / providers
  - [x] Message sync
  - [x] Channel display
  - [ ] Onboarding
  - [ ] Notifications
  - [x] Other: %notes agent + write/import paths

Main risk is a backend update variant the parser doesn't model — handled: parsing returns `null` and the caller falls back to a full sync, today's behavior. The revision guard makes re-applying an update (the stream echoing our own write) a no-op.

The desk change is the part to scrutinize: it alters `se-batch-import-tree`'s semantics, and adds assertions that turn a previously-silent bad write into a crashed poke. `batch-import-tree` had no frontend caller before this PR, and `tlon-skill` uses only the flat `batchImport`, so the blast radius is small.

## Rollback plan

Revert any of the three commits independently — 3 depends on 2 only in that it simplifies what 2 added. No schema migration and no wire-format change; the backend already emitted the updates and already accepted `batch-import-tree`.


arthyn:
closing until more ready

PR #6252, opened 2026-08-11
select bot slash commands by the harness advertised in %contacts

patosullivan:
## Summary

A Hermes-backed bot currently opens the slash-command popup with OpenClaw suggestions, so people are offered commands that bot cannot run. The client hardcodes OpenClaw because bots provide no runtime identity; this adds a versioned `bot-info` claim in `%contacts` so the claimed harness (the bot runtime: the TypeScript OpenClaw plugin or the Python Hermes agent) selects the matching app-owned command list. It separately fixes OpenClaw registry drift—`/tlon` and `/migrate` are real plugin commands missing from its list—and gives Hermes a reachable list audited against the adapter and pinned host core.

Fixes TLON-6300

## Changes

**Identity publishing and client selection**

- Each runtime publishes `v`, `harness`, its plugin or adapter `version`, and an optional host `harnessVersion` as JSON in a `%text` `bot-info` field on its self-contact. OpenClaw and Hermes compare against the current field at boot and reconnect, poke only on a difference, retry transient write failures, and treat publishing failure as non-fatal. A failed self-contact read never causes a blind write. The poke's shape comes from a small `contactSelfFieldPoke` builder exported by `@tloncorp/api`.
- The client carries `bot-info` through the contact API and a new nullable `contacts.bot_info` column, reading whatever claim ordinary contact sync has delivered — the contact book, `/v1/directory`, the `%meet` a profile view fires, or subscription facts. It fetches nothing of its own: a never-met bot shows the default list until one of those paths delivers its profile, the same degradation as a missing nickname for an unmet peer (deliberate scope, recorded in `docs/bot-info.md`). The web localStorage snapshot fills missing rows only and never overwrites a stored claim.
- `parseBotInfo` accepts only version 1 with non-empty string fields, caps the raw JSON at 512 UTF-8 bytes and each field at 64 Unicode code points, and ignores unknown fields for forward compatibility. Missing, malformed, mis-cased, or unknown harnesses select the OpenClaw fallback.
- Bot DMs resolve the counterpart ship and select that contact's list. The home-group bot channel has no counterpart ship signal, so it continues to use the fallback list; identifying its bot member is tracked separately in TLON-6301.

**Command lists and drift protection**

- OpenClaw's app list adds the missing `/tlon` and `/migrate` commands, and `/model` joins the host-core suggestions. Its ten plugin commands now come from one registration table that also produces the committed token fixture; the four host-core suggestions remain audit-pinned app constants.
- The dormant Hermes constant does not include any adapter control commands. The selectable Hermes list contains the adapter's ten advertised commands plus six host-core commands audited against the pinned Hermes agent. Its new eleven-row adapter registry also keeps the `/tlon-version` legacy alias handled but deliberately excludes that alias from suggestions.
- A shared contract compares each runtime fixture with the corresponding app-owned runtime list after sorting. Exact sequence equality catches additions, removals, and duplicates, while popup order remains an app concern expressed through priorities. The contract also runs in `bot-checks`, including on bot-only pull requests.
- Popup icon lookup now accepts only real own exports from the icon namespace, excludes module metadata, and falls back to the generic command glyph. Tests assert that every static entry resolves and that both harnesses retain their curated order.
- The Hermes deployment filter now covers every top-level runtime Python module while excluding tests, so the identity publisher and command registry participate in hosted bot restarts.
- The shared bot E2E suite gains a scenario asserting the claim reaches both the bot's own profile and the owner's peer contact against real ships.

The wire schema, validation limits, version-source rules, two-phase command-removal rule, and explicit clear-to-null procedure are documented in `docs/bot-info.md`. No Hoon or desk changes are required because v1 contacts already carry open key-value fields through their normal sync paths.

## How did I test?

The shared, API, app, OpenClaw, and Hermes adapter test suites pass, along with repo-wide TypeScript checks and Prettier on the changed files. Focused coverage exercises claim parsing and publishing, retry and compare-before-poke behavior, directory carry-through, runtime selection and fallback, command registry/fixture parity, popup ordering, and icon resolution. The drift contract was also mutation-checked with added, removed, and duplicated tokens in both runtime fixtures.

A shared E2E scenario, `bot-info-publishes-and-replicates`, exercises the publish path against real ships in Docker for both runtimes: it asserts the claim on the bot's own `/v1/self` and on the owner's per-ship contact, byte-identical, carrying the harness and that runtime's real package version. This is the part no unit test can reach — that a live `%contacts` agent accepts the poke, that the namespaced key survives `sane-contact`, and that it reaches the profile the client actually reads. Both drivers pass, publishing `{"v":1,"harness":"hermes","version":"0.15.0","harnessVersion":"0.17.0 (2026.6.19)"}` and `{"v":1,"harness":"openclaw","version":"0.19.0","harnessVersion":"2026.5.28"}` respectively. Disabling a publisher fails that scenario and only that scenario.

Its scope is publication and replication. Compare-then-poke idempotence remains unit-tested in both runtimes rather than asserted there, because asserting it end to end required restarting the bot container mid-suite, which the OpenClaw gateway does not survive.

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [x] Other: contacts sync/schema and bot runtime publishing

The identity claim is self-published, but it can only select between static lists shipped by the app; no wire-provided command, label, or icon is rendered. Invalid claims and failed reads fall back to OpenClaw, and publish failures do not stop either bot runtime.

The database change is an additive nullable column. The visible OpenClaw list gains `/tlon`, `/migrate`, and `/model`; sending `/migrate` invokes the existing diary-to-notes migration flow. Hermes selection is DM-only in this scope, and home-group conversations keep the fallback list.

Command lists ship with the app, so a bot-side command change reaches users on an app release rather than a bot deploy, and a third-party bot cannot advertise commands of its own — an unrecognized harness gets the fallback list.

The hosted-bot deploy trigger is intentionally wider for Hermes top-level runtime modules. Command removals require a two-phase rollout: keep the runtime handler until an app release no longer suggests it. Hosted deploys build `@tloncorp/api` from the same checkout as the plugin, so the new builder export is available there; the next npm publish of the plugin, however, needs an `@tloncorp/api` release that carries it.

## Rollback plan

Revert this pull request. The base app does not read `bot-info`, so claims already published in contact profiles are inert after an app rollback. If the identity key should also be removed from bot profiles, send the documented `%self` poke with `bot-info: null`; stopping a publisher alone does not delete a merged contact key.

## Screenshots / videos

Not included; this changes which commands populate the existing popup, not its layout.


packages/shared/src/db/migrations/0000_hesitant_ultimo.sql
```diff
@@ -172,6 +172,7 @@ CREATE TABLE `contacts` (
 	`status` text,
 	`color` text,
 	`coverImage` text,
+	`bot_commands` text,
```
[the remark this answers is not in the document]
patosullivan: declined. the desktop electron app is currently dead code.

packages/openclaw/src/bot-info.ts
```diff
@@ -0,0 +1,113 @@
...
+  selfContact?: SelfContactRead
+): Promise<BotCommandManifestPublishResult | 'skipped'> {
+  try {
+    return await maybePublishBotCommandManifest(
+      api,
+      selfContact ?? (await readSelfContact(api)),
+      desiredValue
+    );
+  } catch {
+    return 'skipped';
```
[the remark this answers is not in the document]
patosullivan: Fixed: the publish poke now retries in place — 3 attempts with 2s/8s backoff between them, never after the last — then falls through to the existing non-fatal skip. Failed self-contact reads are still never retried, and an unchanged compare still performs no pokes. The null-clear path gets the same retry since it's the documented rollback procedure.

packages/hermes-tlon-adapter/adapter.py
```diff
@@ -2463,17 +2475,60 @@ async def _send_control_reply(
...
+        try:
+            desired = build_command_manifest_json()
+            if extract_bot_commands_value(self_contact) == desired:
+                return
+            await self._sse.poke(
+                "contacts", BOT_COMMANDS_CONTACT_MARK, build_bot_commands_poke(desired)
+            )
+            logger.info("[tlon] published bot command manifest")
+        except Exception as exc:
+            logger.warning("[tlon] could not publish bot command manifest: %s", exc)
```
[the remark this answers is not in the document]
patosullivan: Fixed the same way as the OpenClaw side: bounded in-place retry around the poke (3 attempts, 2s/8s between them), then the existing non-fatal warning. Read failures and unchanged compares are unaffected.

packages/shared/src/store/useBotSlashCommandManifest.ts
```diff
@@ -39,9 +85,52 @@ export const useBotSlashCommandManifest = (
...
+      !shouldBackfillBotCommands({
+        enabled,
+        botShipId,
+        contactQuerySettled: isFetched,
+        hasAdvertisedManifest: !!advertisedManifest,
+      })
+    ) {
+      return;
+    }
+    ensureBotCommandsSynced(botShipId!);
```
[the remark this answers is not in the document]
patosullivan: This one is deliberate and disclosed in the PR description: the backfill is best-effort and a transient failure leaves that one conversation on the static fallback until remount. The blast radius is a cosmetic downgrade in a single conversation for a session, which we didn't think justified retry scheduling with unmount/account-change cancellation in the hook. Bounded retry here is noted as follow-up work.

packages/shared/src/store/contactActions.ts
```diff
@@ -10,6 +11,83 @@ import { syncGroup } from './sync/syncGroup';
...
+    botCommandsBackfillInFlight.add(key);
+    reservedKey = key;
+
+    const contact = await db.getContact({ id: ship });
+    // Only a *usable* manifest means there is nothing to fetch: a stale,
+    // malformed or wrong-version value reads as no manifest everywhere else
+    // (the hook falls back to the static list), so it must not pin the
+    // backfill off either.
+    if (domain.parseBotCommandManifest(contact?.botCommands)) {
+      return;
```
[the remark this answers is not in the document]
patosullivan: Known cost of the provenance design, now disclosed in the description: preserve-on-v0 is what stops the lossy `/all` scry from clobbering good data, and the price is that a manifest changed while the app was offline persists as its cached value until the bot next publishes — which every bot boot does via compare-then-poke when content changed. The real fix is moving initial peer sync off the lossy v0 scry entirely, which is contacts-sync work we're tracking separately rather than bolting onto this PR.

packages/openclaw/src/bot-command-manifest.ts
```diff
@@ -0,0 +1,148 @@
...
+// skipping entirely (see SelfContactRead).
+export const BOT_COMMANDS_PUBLISH_ATTEMPTS = 3;
+export const BOT_COMMANDS_PUBLISH_BACKOFF_MS: readonly number[] = [
+  2_000, 8_000,
+];
+
+export type Sleeper = (ms: number) => Promise<void>;
+
+const defaultSleep: Sleeper = (ms) =>
+  new Promise((resolve) => setTimeout(resolve, ms));
```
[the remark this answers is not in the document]
patosullivan: Fixed following the auth-backoff pattern in this file: the sleeper takes the monitor's `opts.abortSignal`, clears the pending timer and rejects on abort, and the rejection lands in the existing non-fatal skip — so a retired monitor stops retrying immediately instead of lingering through the backoff. Hermes needs no change: its backoff is `asyncio.sleep`, which task cancellation already interrupts (`CancelledError` escapes the `except Exception` handlers there).

packages/api/src/urbit/contact.ts
```diff
@@ -95,6 +95,10 @@ export interface ContactBookProfile {
   ['lanyard-phone-0-sign']?: AttestationSignature;
   ['lanyard-twitter-0-url']?: AttestationProviderUrl;
   ['lanyard-phone-0-url']?: AttestationProviderUrl;
+  // Self-published bot identity claim (JSON-in-text). Declared as %text but
+  // treated as untrusted at runtime — any ship can publish any value type
+  // under an unknown key.
```
arthyn: unnecessary comment

packages/app/ui/components/slashCommandIcon.ts
```diff
@@ -0,0 +1,40 @@
...
+ * has a null prototype and no `__esModule`, which is why this is only reachable
+ * in some builds — and why the tests inject a CJS-shaped namespace rather than
+ * relying on whatever the test runner happens to produce.)
+ *
+ * Membership is an own-enumerable-key test (`Object.keys` skips the prototype
+ * chain) minus that metadata. It deliberately does not inspect the *value*: an
+ * icon is a component in the app but a string under the test runner's SVG
+ * transform, so a shape check would reject every real icon in one of those
+ * worlds.
+ */
```
arthyn, line 22: shorten

packages/openclaw/src/bot-info.ts
```diff
@@ -0,0 +1,203 @@
...
+    const timer = setTimeout(() => {
+      signal?.removeEventListener('abort', onAbort);
+      resolve();
+    }, ms);
+    const onAbort = () => {
+      clearTimeout(timer);
+      reject(new Error('Aborted'));
+    };
+    signal?.addEventListener('abort', onAbort, { once: true });
+  });
```
arthyn: belongs in utils

packages/openclaw/src/bot-info.ts
```diff
@@ -0,0 +1,203 @@
...
+
+// A self-contact read that failed is not the same as one that succeeded
+// without the key: only the latter proves the key is absent. Publishing on a
+// failed read defeats compare-then-poke exactly when the ship is unhealthy.
+export type SelfContactRead =
+  | { ok: true; contact: unknown }
+  | { ok: false; error: unknown };
+
+export async function readSelfContact(
+  api: BotInfoScryApi
```
arthyn, line 69: are these not exposed by api lib?

packages/openclaw/src/bot-info.ts
```diff
@@ -0,0 +1,203 @@
...
+// log-and-continue.
+export async function publishBotInfo(
+  api: BotInfoPokeApi,
+  desiredValue: string | null,
+  sleep: Sleeper = defaultSleep,
+  abortSignal?: AbortSignal
+): Promise<BotInfoPublishResult> {
+  for (let attempt = 1; ; attempt++) {
+    try {
+      await api.poke({
```
arthyn: use api lib

packages/shared/src/domain/slashCommands.ts
```diff
@@ -5,22 +5,129 @@ export interface SlashCommandOption {
...
+}
+
+/**
+ * Parse a bot's identity claim off a synced contact record. Takes `unknown` —
+ * the value comes from the network and the TS declaration of the contact field
+ * proves nothing at runtime. Returns null when the claim is absent, malformed,
+ * over-long, or the wrong version; callers then treat the bot as unidentified
+ * and fall back to the default command list.
+ */
+export function parseBotInfo(raw: unknown): BotInfo | null {
```
arthyn, line 62: I'm thinking we probably need to do something to stop the bots from building all these parsers for stuff that is generally pretty consistent. we have types and the backend is controlled, so we shouldn't really be seeing anything here unexpected.

packages/shared/src/domain/slashCommands.ts
```diff
@@ -109,99 +192,233 @@ const OPENCLAW_COMMANDS: SlashCommandOption[] = [
...
+    title: 'Migrate diary to notes',
+    subtitle: 'Run or clean up a diary-to-notes migration',
+    icon: 'Copy',
+    keywords: ['migrate', 'diary', 'notes', 'migration'],
+    priority: 13,
+  },
 ];
 
-const HERMES_COMMANDS: SlashCommandOption[] = [
+// OpenClaw core. Audit-verified against core at the plugin's dev pin
```
arthyn, line 213: why don't we expose /model?

packages/shared/src/store/contactActions.ts
```diff
@@ -10,6 +11,83 @@ import { syncGroup } from './sync/syncGroup';
...
+// initial v0 `/all` peers scry strips the namespaced `bot-info` key, and
+// waiting for the bot's next republish could take weeks. Fetch the ship's full
+// v1 profile on demand instead.
+const BOT_INFO_BACKFILL_MAX_ATTEMPTS = 3;
+// Session-scoped bookkeeping, keyed by `${currentUserId}:${ship}` so a
+// switched account starts fresh.
+const botInfoBackfillInFlight = new Set<string>();
+const botInfoBackfillAttempts = new Map<string, number>();
+
+export async function ensureBotInfoSynced(ship: string): Promise<void> {
```
arthyn: hopefully unnecessary after #6301

arthyn (commented):
we should get #6301 in first and rebase off of it, should cut some fluff from this PR. not finished reviewing, but I'd rather see that go through first.

packages/shared/src/domain/slashCommands.ts
patosullivan: Because we had some discussion a couple weeks ago about not making that one as easy to find. I left it in for hermes, though, on accident. I don't mind keeping it or restricting it either way. I can add it back for now.

packages/shared/src/domain/slashCommands.ts
patosullivan: Probably true in general, but this one isn't really backend-controlled data tho. It's just an arbitrary key we stuff in.

packages/shared/src/db/migrations/meta/_journal.json
```diff
@@ -5,9 +5,9 @@
     {
       "idx": 0,
       "version": "6",
-      "when": 1782835292031,
-      "tag": "0000_unique_texas_twister",
+      "when": 1786472821543,
+      "tag": "0000_pink_forgotten_one",
```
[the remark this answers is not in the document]
patosullivan: Wrong, this is not how migrations work in this codebase. There's always a single migration file.

packages/api/src/client/contactsApi.ts
```diff
@@ -437,6 +451,47 @@ function parseContactAttestations(
...
+    return null;
+  }
+  const candidate = field as { type?: unknown; value?: unknown };
+  if (candidate.type !== 'text' || typeof candidate.value !== 'string') {
+    return null;
+  }
+  return candidate.value;
+};
+
+// Fetch a single peer's full v1 contact profile. Used to backfill the
```
arthyn: this doesn't really make sense. the directory already contains this if its present. if it isn't present this scry will return null, so it can't really be used for this.

packages/openclaw/src/bot-info.ts
```diff
@@ -0,0 +1,178 @@
+import { contactSelfFieldPoke } from '@tloncorp/api';
```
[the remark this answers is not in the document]
patosullivan: This is the publish choreography already recorded in the PR description's Risks section, not a defect in this branch: hosted deploys build `@tloncorp/api` from this same checkout (tlawn.py wires it in as a `file:` tarball; the registry is only a fallback), and the next npm publish of the plugin is sequenced behind an api release carrying this export.

packages/openclaw/src/bot-info.ts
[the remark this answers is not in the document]
patosullivan: The bounded exponential-backoff retry requested here already exists: `publishBotInfo` loops `BOT_INFO_PUBLISH_ATTEMPTS` with `BOT_INFO_PUBLISH_BACKOFF_MS` sleeps between attempts (`src/bot-info.ts`). What remains after exhaustion is deliberate: publishing is non-fatal and re-runs on reconnect catch-up.

arthyn (approved):

PR #6259, opened 2026-08-11
fix: show thread unread dots for replies received while the channel was closed

patosullivan:
## Summary

Fixes TLON-6304: intermittent missing thread-unread dots on the "N replies" row. The channel post-list query never populates `threadUnread` on posts, and per-post cache entries expire after 30s with no refetch, so a reply that arrives while the channel is closed (app open, but on another screen, like home) can leave the dot unpainted after the entry ages out and the channel reopens.

## Changes

-   Added `packages/app/ui/contexts/threadUnreads.tsx`: a provider backed by `useLiveThreadUnreadsByChannel` (already reactive, already used by the notes tree) exposing a `Map<threadId, ThreadUnreadState>` for the current channel.
-   `Channel/index.tsx`: mounts the provider inside `ChannelProvider`, scoped to `chat`, `dm`, `groupDm`, `notebook` (gallery/notes untouched, so no new broadly-invalidated query is added there). The existing `useLiveThreadUnreadsByChannel` call feeding `hasMainChannelUnreadActivity` is left as-is; React Query dedupes the identical key.
-   `ChatMessageReplySummary.tsx`: when the context map is present it's authoritative for dot state (a missing entry means no dot); falls back to `post.threadUnread` outside a provider. Authoritative, not merely a fallback, because the live query uses `excludeRead: true` and a stale `post.threadUnread` could otherwise resurrect a dot for an already-read thread.

Note: a separate backend defect (a self-read floor collision in `%activity`/`%channels`) can produce the same symptom and will be filed separately; this addresses the client-side cause only.

## How did I test?

-   `pnpm -r tsc` passes, prettier clean on changed files.
-   Unit tests for `resolveThreadUnread` (`packages/shared/src/logic/threadUnreads.test.ts`), which encodes the subtle part: a present channel map is authoritative, so a missing entry means no dot even when the post still carries a stale `threadUnread`. Mutation-checked — reintroducing the fallback fails the authoritative test, restoring it passes.
-   Integration tests for `getThreadUnreadsByChannel` (`packages/shared/src/db/queries.test.ts`) covering the contract the overlay depends on: rows keyed by parent post id, `excludeRead` dropping read threads, and channel scoping.
-   Added one dot-visible assertion each to `thread-unread-clearing.spec.ts` and `dm-thread-unread-clearing.spec.ts` covering live-arrival in group and DM threads; these pass pre-fix by design and guard against the new overlay suppressing a live dot.
-   An e2e spec for the closed-channel scenario was attempted and dropped: reproducing it needs a prior channel visit, a reply while closed, and a wait past the 30s per-post cache TTL, which pushed a CI shard past its 30-minute limit. The unit and integration tests above cover the same invariants deterministically.

## Risks and impact

    -   [ ] Onboarding
    -   [x] State / providers
    -   [ ] Message sync
    -   [x] Channel display
    -   [ ] Notifications

## Rollback plan

Revert.


arthyn (approved):
lgtm

PR #6275, opened 2026-08-13
staging: fix omnibus 2

latter-bolden:
- Fix Android bottom nav Contacts tab (keep new Android component, replace default b&w sigil with person icon)
- Fix "login after logging out" hard crash (due to unavailable DB relations)
- Fix missing note blue dots (would show on top level Channel List but not Note Channel List)
- Fix more sheets rendering below the new bottom nav


patosullivan (approved):
lgtm, but why create the new TopLevelTabNavigator.react.tsx?

PR #6295, opened 2026-08-17
prevent openclaw api client races

dnbrwstr:
## Summary

Prevents OpenClaw monitors, gateway heartbeats, and outbound operations from replacing one process-global Tlon API client while another operation is using it.

This fixes the production failure observed on `~dordev-ragsyn`: a cron-generated Notes entry completed successfully, but publishing failed with `JSON requests not supported on this client shim`. A gateway heartbeat installed the monitor's poke-only client between the authenticated setup and the later `notes.getNotebook()` / `notes.createNote()` JSON requests.

The generated content was not the cause.

## Changes

- Adds a small client-resolver hook to `@tloncorp/api`. App clients still use the existing configured singleton by default; server runtimes can resolve a client from their current async context.
- Gives each OpenClaw monitor its own async client scope before it creates SSE callbacks or background work.
- Runs each outbound send or action in a nested, full client scope containing `poke`, `scry`, and `requestJson`. Concurrent accounts cannot replace each other's client, and the previous monitor scope is restored automatically afterward.
- Keeps scoped authentication with the scoped client. Scry retries once after a 403, JSON requests retry once after a 401 or 403, and concurrent reauthentication is deduplicated within that operation.
- Runs gateway configure, start, heartbeat, and stop calls through the normal API `poke()` inside a short scope bound to the monitor's SSE transport. This keeps the existing API signatures and the generic poke logging, error reporting, and duration instrumentation.
- Shares the async scope through OpenClaw's existing cross-module registry so it survives OpenClaw's separate extension and runtime module-loader contexts.
- Removes every OpenClaw production call to `configureClient`; monitor restarts, heartbeats, and concurrent outbound operations no longer mutate the API package's singleton.

## Why this is safe

The underlying transports are unchanged. Monitor and gateway pokes still use the existing authenticated `UrbitSSEClient`; one-shot outbound work still uses the existing authenticated HTTP poke client and SSRF-guarded fetch path.

The client is selected once at the start of each `poke`, `scry`, or `requestJson` call, so another async scope cannot change the client midway through that request. Scoped clients own their own authentication and never enter the API singleton's reauthentication flow.

Gateway traffic uses the ordinary `@tloncorp/api` gateway helpers and `poke()` wrapper. There is no custom instrumentation path and no added transport parameter on the gateway API surface.

This is intentionally smaller than converting the entire API package to client factories. The API package keeps its existing singleton as the default for mobile, web, and other app clients. The resolver covers the client proxy plus `poke`, `scry`, and `requestJson`, which are the shared-client paths OpenClaw uses. Singleton-oriented subscription APIs are unchanged and are not routed through OpenClaw's scoped shim.

The existing single-account constraint for gateway-status and reengagement remains. It is now a product/shared-gateway-transport constraint, not a client-race constraint.

## Risks and impact

- Adds Node `AsyncLocalStorage` only inside the OpenClaw package; the shared API package has no Node dependency.
- A scoped client does not use the app singleton's auth callbacks; OpenClaw owns scoped auth and retry.
- There is no protocol, payload, database, or configuration change.
- Safe to roll back by reverting this PR.

## How did I test?

From a clean worktree based on current `origin/develop`:

- `packages/api`: TypeScript and package build passed
- `packages/api`: full unit suite — 35 files and 745 tests passed
- `packages/openclaw`: TypeScript and package build passed
- `packages/openclaw`: full unit suite — 68 files and 1,363 tests passed
- Temporary merge with current `db/agent-onboarding-v2`, including Notes publishing, after resolving the expected overlaps in `monitor/index.ts` and `urbit/api-client.ts`: OpenClaw build passed and 69 files / 1,403 tests passed
- Regression tests cover concurrent account scopes, normal gateway poke routing, descriptive failures for unsupported gateway operations, the empty activation slot, scoped `poke` / `scry` / `requestJson`, JSON reauthentication, and restoration after nested outbound work
- Targeted ESLint on changed files — no errors; existing warnings are unchanged
- `git diff --check`

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tloncorp/api",
-  "version": "0.0.10",
+  "version": "0.0.12",
   "type": "module",
   "files": [
     "dist",
--- a/packages/api/src/__tests__/urbitClientResolver.test.ts
+++ b/packages/api/src/__tests__/urbitClientResolver.test.ts
@@ -1,0 +1,69 @@
+import { afterEach, describe, expect, test, vi } from 'vitest';
+
+import {
+  BadResponseError,
+  poke,
+  requestJson,
+  scry,
+  setClientResolver,
+} from '../client/urbit';
+import { AuthError, type Urbit } from '../http-api';
+
+afterEach(() => {
+  setClientResolver(null);
+});
+
+describe('client resolver', () => {
+  test('routes an operation through the resolved client', async () => {
+    const scopedClient = {
+      poke: vi.fn().mockResolvedValue(7),
+      requestJson: vi.fn().mockResolvedValue({ ok: true }),
+      scryWithInfo: vi.fn().mockResolvedValue({
+        result: { ship: '~zod' },
+        responseSizeInBytes: 0,
+        responseStatus: 200,
+      }),
+    };
+    setClientResolver(() => scopedClient as unknown as Urbit);
+
+    await expect(poke({ app: 'chat', mark: 'test', json: {} })).resolves.toBe(
+      7
+    );
+    await expect(scry({ app: 'contacts', path: '/v1/self' })).resolves.toEqual({
+      ship: '~zod',
+    });
+    await expect(requestJson('/notes', 'GET')).resolves.toEqual({ ok: true });
+  });
+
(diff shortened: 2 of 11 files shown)


latter-bolden (approved):
Looks good to me

Let's make sure to bump the API module and publish

PR #6311, opened 2026-08-19
web: fix notes overwriting each other as you swap around

latter-bolden:
## Summary

Unsaved notebook edits could overwrite another note when rapidly switching between notes. The shared editor retained draft state while delayed save callbacks continued updating it, allowing one note’s content to become associated with another note’s save target.

The fix prevents asynchronous saves from one note from mutating another note’s editor state. Saves remain ordered in the background, while note switching stays immediate and drafts are safely restored and rebased.

## Changes

- Guarded save completions using note identity.
- Preserved one FIFO save chain across note switches.
- Restored pending drafts synchronously from memory.
- Added observable per-note pending-save tracking across component remounts.

## Risks and impact

  - [x] State / providers
  
Fixes TLON-6356

### DIFF
--- a/packages/app/ui/components/NotesChannel/NotesNativeChannel.tsx
+++ b/packages/app/ui/components/NotesChannel/NotesNativeChannel.tsx
@@ -1093,7 +1093,6 @@ export function NotesNativeChannel({
         />
       ) : (
         <NotesNoteDetail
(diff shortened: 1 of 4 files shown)


jamesacklin (approved):
Looks good to me. Let's also get this out to internal on \~marnec-dozzod-marnus so others don't get bitten in the web editor.

PR #6361, opened 2026-08-26
Sync staging into develop (8-26)

arthyn:
## Summary

Syncs `staging` into `develop`, bringing across 42 commits — most notably the group blob feature (#6209), which shipped from staging.

Routine sync in the `sync-staging-dev-*` series. Thirteen paths conflicted; four were mechanical, five needed real integration because develop and staging had each done genuine, different work to the same components.

## Changes

**Mechanical conflicts**

- `db/migrations` — both sides had regenerated the single baseline (`0000_bitter_hammerhead` on develop, `0000_special_colossus` on staging). Rather than hand-resolve the snapshot JSON, reran `node reset-migrations` against the merged schema. The result, `0000_aspiring_klaw`, differs from develop's baseline by exactly one line: `` `blob` text `` on the groups table.
- `queries.test.ts` — each side renamed a different import. Kept develop's `directoryToClientProfiles` + `ContactsDirectoryScryResult1` and staging's `toClientGroups`.
- `desk.docket-0` — took develop's glob; the glob bot rewrites it regardless.

**Integration conflicts**

| File | Resolution |
| --- | --- |
| `BotApiKeySettingsScreen`, `BotOpenAISubscriptionScreen` | staging genericized OpenAI-specific naming (`disconnectLLMSubscription`, `subscriptionName`, `providerName`) while develop reformatted the same lines with oxfmt. Took staging's identifiers, reformatted for develop. |
| `BotSettingsUI` | develop aliased `BotSettingsSection` onto the shared `SettingsSection` (native settings chrome migration); staging had added `subtitle` and `valueColor`. Kept develop's alias and carried both props forward — `subtitle` onto the shared `SettingsSection`, `valueColor` back onto `BotSettingsRow`. |
| `BotModelSettingsScreen` | staging's `defaultStep` provider/model flow is the live logic; develop's side contributed `placement="navigation"`. Kept both. |
| `SplashSequence` | develop's Aug-25 work (onboarding handoff, uncontrolled-input fixes, Android IME) is six days newer and supersedes staging's Aug-19 invite-button tweak, which targeted markup develop has since replaced. Took develop's, then repointed it at staging's renamed `LLMSubscriptionAuthView` and wired the now-required `onSubmitToken` to the hook's `completeToken`, matching staging's own caller. |

## How did I test?

- `./backend/run-tests.sh`: **344 OK**, **Unit tests passed ✅**, **Aqua tests passed ✅** — the desk compiles with staging's blob work on top of develop's tree.
- `tsc --noEmit` clean in `packages/api`, `packages/shared`, and `packages/app`.
- vitest: 102 tests across `queries.test.ts`, `groupActions.test.ts`, `sync.test.ts`.
- `oxfmt` on every file touched during resolution.

Two resolution bugs were caught by the `app` typecheck rather than by reading, and both are fixed: `git checkout --ours` on `BotSettingsUI.tsx` initially discarded *all* of staging's changes to that file, dropping `valueColor` (which `BotSettingsScreen` passes in two places); and taking develop's `SplashSequence` wholesale broke an import, since staging had renamed the component and added a required prop.

The one remaining local typecheck error is `@tloncorp/editor/dist/editorHtml`, an unbuilt artifact in a fresh worktree, not a merge issue — CI builds packages.

## Risks and impact

  - [x] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: bot settings UI, group blob feature, db migration baseline

The bot settings and onboarding resolutions are the parts most worth a second pair of eyes — particularly `SplashSequence`, where staging's invite-button change is intentionally dropped as superseded. If that call is wrong, it should be re-applied on top of develop's newer markup rather than by reverting this resolution.

## Rollback plan

Revert the merge commit.

## Screenshots

n/a


latter-bolden (approved):
We should wait to get passing CI before merging

PR #6374, opened 2026-08-27
auto-release Android preview builds

dnbrwstr:
## Summary

Make Android preview builds available to internal testers without a manual Play Console release step.

## Changes

- complete preview Android submissions on the internal track
- keep production Android submissions as drafts

## How did I test?

- validated the EAS submit profiles with `jq`
- confirmed the preview Gradle flavor uses `io.tlon.groups.preview`
- confirmed preview builds still submit to the `internal` track
- confirmed the production submit profile is unchanged
- `git diff --check`


patosullivan (approved):
🙏

PR #6376, opened 2026-08-28
fix profile header layout

dnbrwstr:
## Summary

Restore profile names and normal spacing after the Bot badge layout caused the header to collapse.

Fixes TLON-6408

## Changes

- give the profile identity wrapper a definite full width
- preserve long-name truncation and the trailing Bot badge

## How did I test?

- `packages/app` TypeScript check passed
- targeted formatting passed
- targeted lint completed with two existing warnings
- `git diff --check` passed
- verified normal and Bot profiles on a physical Galaxy S24
- verified long Bot names truncate while the badge remains visible
- verified profile scrolling down and back up

## Screenshots

| Before | After |
| --- | --- |
| Profile before the fix | Profile after the fix |

### DIFF
--- a/packages/app/ui/components/UserProfileScreenView.tsx
+++ b/packages/app/ui/components/UserProfileScreenView.tsx
@@ -115,7 +115,7 @@ export function UserProfileScreenView(props: Props) {
           flexDirection: 'row',
         }}
       >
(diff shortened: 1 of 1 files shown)


latter-bolden (approved):
Had the same fix locally, confirmed fixes on emulator

PR #6377, opened 2026-08-28
ops: bump mobile version to 9.5.0

dnbrwstr:
## Summary

Bump the mobile app version from 9.4.3 to 9.5.0 so the production iOS build can be submitted.

## Changes

- update the mobile package version
- update Android versionName
- update iOS MARKETING_VERSION

## How did I test?

- verified all three mobile version sources are 9.5.0
- verified no 9.4.3 or 9.4.4 references remain in those sources
- `git diff --check` passed
- confirmed the previous production build succeeded and only the iOS submission failed because 9.4.3 had already been submitted

### DIFF
--- a/apps/tlon-mobile/android/app/build.gradle
+++ b/apps/tlon-mobile/android/app/build.gradle
@@ -132,7 +132,7 @@ android {
         targetSdkVersion rootProject.ext.targetSdkVersion
         compileSdk rootProject.ext.compileSdkVersion
         versionCode 108
-        versionName "9.4.4"
+        versionName "9.5.0"
     }
     compileOptions {
         coreLibraryDesugaringEnabled true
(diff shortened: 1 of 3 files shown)


apps/tlon-mobile/android/app/build.gradle
```diff
@@ -132,7 +132,7 @@ android {
         targetSdkVersion rootProject.ext.targetSdkVersion
         compileSdk rootProject.ext.compileSdkVersion
         versionCode 108
-        versionName "9.4.3"
+        versionName "9.4.4"
```
jamesacklin: 9.5.0

jamesacklin (changes requested):

jamesacklin (approved):

PR #6393, opened 2026-08-31
fix iOS chat input resize flicker

dnbrwstr:
## Summary

Prevents the iOS chat scroller from flashing or jumping when the message input changes height, while preserving the existing Android behavior.

Fixes TLON-6420

## Changes

- feed the floating composer height directly into the iOS list content inset
- keep older messages anchored while browsing history and continue following the latest message at the end
- commit iOS inset and offset updates together in the keyboard controller
- avoid double-counting composer padding in compact conversation lists

## How did I test?

- tested one-line and multiline composer growth at the bottom and mid-history on an iPhone 17 Pro simulator running iOS 26.4
- tested emoji keyboard resizing, momentum scrolling, chat remounting, context menus, and standard Dynamic Type sizes through 3XL
- verified the frozen pnpm install and dependency patch application
- `corepack pnpm --filter @tloncorp/app tsc`
- targeted `oxfmt` and `oxlint` passed with existing warnings only
- `git diff --check`

### DIFF
--- a/packages/app/ui/components/Channel/DraftInputView.tsx
+++ b/packages/app/ui/components/Channel/DraftInputView.tsx
@@ -1,5 +1,5 @@
 import { DraftInputId } from '@tloncorp/api';
-import { ComponentProps, PropsWithChildren } from 'react';
+import { ComponentProps, PropsWithChildren, useEffect } from 'react';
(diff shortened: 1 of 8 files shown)


janicduplessis (approved):
Sounds good, I 'm fine with that!

PR #6400, opened 2026-09-01
fix repeated android notification routing

dnbrwstr:
## Summary

Stops Android from reusing the same notification tap when the native notification module is recreated. This applies Expo’s accepted native fix and drains queued responses after they are delivered.

Fixes TLON-6424

## Changes

- patch `expo-notifications` 57.0.6 with the upstream queue-draining fix
- build `expo-notifications` from source on Android so the patch is included instead of the prebuilt AAR
- refresh the Gradle dependency lock and document when to remove the patch

## How did I test?

- built and launched `productionDebug` with Stim on a physical Samsung SM-S921U
- cold-started the app with a synthetic notification response, then reloaded React Native without killing Android; the response was emitted once before the reload and was not emitted again
- inspected the built APK to confirm the queue-clearing native code is present
- `corepack pnpm install --frozen-lockfile --ignore-scripts --offline`
- `stim logs --errors`
- `git diff --check`


janicduplessis (approved):
LGTM!

PR #6410, opened 2026-09-01
restore the iOS channel header fade

dnbrwstr:
## Summary

Restores the fade behind channel headers on iOS 26 while preserving the existing spacing for notices and side panels.

## Changes

- keep the channel scroll view underneath the transparent header so iOS can draw its top edge effect
- let visible join-request notices, fullscreen drafts, and the context panel own their header spacing without duplicating it
- add focused coverage for the list inset behavior

## How did I test?

- used Stim for the isolated mobile worktree and its worktree-owned Metro server
- built, installed, and launched the exact PR head on an iPhone 17 Pro simulator running iOS 26.4 with the repo build wrapper
- signed in to the preview app and visually verified that channel messages fade beneath the transparent top header while scrolling, with the existing bottom fade still visible
- `corepack pnpm --filter @tloncorp/app exec vitest run ui/components/conversationInsets.test.ts` (7 tests passed)
- `corepack pnpm --filter @tloncorp/app tsc`
- targeted `oxfmt --check` passed
- targeted `oxlint` passed with existing warnings only
- `git diff --check`


patosullivan (approved):
Lgtm

PR #6426, opened 2026-09-02
Keep the thinking indicator above the composer in empty conversations

jamesacklin:
## Summary

Fixes [TLON-6447](https://linear.app/tlon/issue/TLON-6447) and [TLON-6440](https://linear.app/tlon/issue/TLON-6440): in a brand-new agent group the thinking indicator floated well above the message input instead of sitting directly on top of it. The empty conversation list left its scroll offset unreconciled, so the footer landed one header height (plus the composer inset) too high on iOS.

## Changes

- **`Scroller.tsx`**: for an empty iOS conversation, size the list content to the list frame minus the composer inset instead of a viewport-filling `flexGrow`. This is the same geometry LegendList produces via `alignItemsAtEnd` once it has rows, so the footer rests directly above the composer at offset 0 with no scroll range to drift into. Other layouts and platforms keep their existing empty-state padding.
- **`PostList.tsx`**: while the conversation has no rows, scroll the native scroll view to its end whenever its frame or content size settles. LegendList skips end alignment and `maintainScrollAtEnd` without rows, and its `initialScrollAtEnd` for empty data resolves to an offset equal to the top content inset; the keyboard-controller inset reaction can also run before the scroll view has reported its size. The callback exits immediately once rows exist, so populated lists take exactly the same path as before.
- **`Channel.fixture.tsx`**: two cosmos fixtures used to reproduce and verify: `agentGroupSetupSequence` (timed replay: loading → empty channel → indicator → first entry → indicator clears, with a transparent-header top inset) and `chatWithThinking` (populated chat with the indicator).

### Root cause detail

LegendList only computes `alignItemsAtEndPadding` when `data.length > 0`. With zero rows it falls back to `flexGrow: 1; justifyContent: flex-end`, so the footer's on-screen position is dictated purely by scroll offset. `calculateOffsetWithOffsetPosition` returns `stylePaddingTop` for an empty end target (our top content inset, ~116pt for the transparent iOS 26 header), and `useExtraContentPadding` in react-native-keyboard-controller then adds the 98pt composer inset when it fires before layout is known. 116 + 98 matches the gap in the issue screenshot. Once the first post arrives LegendList's normal end-alignment math takes over, which is why the first message rendered in the right place.

## How did I test?

iOS Simulator (iPhone 17 Pro, iOS 26.5) running the existing debug build against Metro in cosmos mode, measuring the indicator's window position with a temporary `measureInWindow` beacon (composer top edge is at 776pt on this device):

| Scenario | Before | After |
| --- | --- | --- |
| Empty channel, indicator appears after load | 874 (hidden behind composer) | 776 |
| Same, composer inset arriving after list mount (production ordering) | drifted by top inset + composer inset | 776 |
| First post arrives, then indicator clears | — | post and indicator both end at 776 |
| Populated chat (100 posts) with indicator | 776 | 776 (unchanged) |
| Plain empty channel welcome notice | offset-dependent | rests above composer |

Also: `tsc` for `packages/app`, `oxfmt --check`, and vitest for `ThinkingState`, `postListInitialization`, `conversationInsets`, and `postVisibility` (32 tests) pass.

Not verified: keyboard open on an empty channel (simulator tap injection doesn't work on this machine), and the real agent-provisioning flow (needs the private tlonbot sandbox). Both are reasoned through in code comments; the settle callback uses the native `scrollToEnd`, which includes the live keyboard inset.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

Scope is limited to conversations with zero rows on iOS (content sizing) plus a no-op-once-populated settle callback. LegendList never consumes `onContentSizeChange` and chains `onLayout`, so the new props do not shadow its handlers. Thread views always contain the parent post and never hit the empty branch.

## Rollback plan

Revert this commit. No data or schema changes.

## Screenshots / videos

| Before (issue, build 460) | After (simulator, empty channel, indicator visible) |
| --- | --- |
| before: indicator floating above the composer | after: indicator directly above the composer |

| After: first post arrived, indicator cleared | After: populated chat with indicator (unchanged path) |
| --- | --- |
| after: first post directly above the composer | after: populated chat with indicator at the end |

### DIFF
--- a/apps/tlon-mobile/README.md
+++ b/apps/tlon-mobile/README.md
@@ -117,6 +117,13 @@ DEFAULT_SHIP_LOGIN_ACCESS_CODE=
 
 See `.env.sample` for other configurable env variables.
 
+In a dev build, setting both `DEFAULT_SHIP_LOGIN_URL` and
+`DEFAULT_SHIP_LOGIN_ACCESS_CODE` leaves the ship login screen ready to submit —
+press `Connect` and you are logged in, with no 2FA. Hosted (`*.tlon.network`)
+URLs are accepted there in dev builds too. Note this yields a self-hosted style
+session, so hosting-account flows (node status, revival, bot config) are not
+exercised.
+
 ### Notifications
 
 **Note:** dev mode push notifications are only sent to connected devices, not simulators.
--- a/apps/tlon-mobile/src/screens/Onboarding/ShipLoginScreen.tsx
+++ b/apps/tlon-mobile/src/screens/Onboarding/ShipLoginScreen.tsx
@@ -68,7 +68,8 @@ export const ShipLoginScreen = ({ navigation }: Props) => {
     if (!urlPattern.test(url)) {
       return false;
     }
-    if (hostedPattern.test(url)) {
+    // dev helper: allow logging in to hosted ships with their access code
+    if (hostedPattern.test(url) && !__DEV__) {
       return 'hosted';
     }
     return true;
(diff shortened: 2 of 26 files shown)


dnbrwstr (approved):
Nice, confirmed that it also fixes TLON-6440

PR #6445, opened 2026-09-04
mobile: telemetry for bot settings changes

latter-bolden:
Adds telemetry for capturing bot settings changes. Notably so we can tell how much usage zero data retention is getting


jamesacklin (approved):
Cool, looks comprehensive

PR #6448, opened 2026-09-04
ios: fix onboarding category picker after custom submission

latter-bolden:
## Summary

The visual effect in the video is strange, seemingly half the component looked "disabled". The issue was likely a UIKit race: the bottom-edge fade interaction recalculated while the keyboard was dismissing, then remained attached using the keyboard-adjusted geometry. That left an oversized translucent region over the category picker, making it look disabled even though it was still interactive.

The fix listens for keyboardDidHide and reattaches the scroll-edge interaction after UIKit finishes updating the layout.

## Changes

- Added an observer for `UIResponder.keyboardDidHideNotification` in `ScrollEdgeElementContainer`
- When the keyboard finishes hiding, the code removes and reattaches `UIScrollEdgeElementContainerInteraction`, forcing UIKit to recalculate the bottom fade with the correct layout

## How did I test?

Couldn’t reproduce the original artifact in the simulator, but the fixed flow behaved correctly in fixtures.

## Risks and impact

Fixes TLON-6464

### DIFF
--- a/apps/tlon-mobile/modules/tlon-scroll-edge-effect/ios/ScrollEdgeElementContainer.swift
+++ b/apps/tlon-mobile/modules/tlon-scroll-edge-effect/ios/ScrollEdgeElementContainer.swift
@@ -45,11 +45,19 @@ public final class ScrollEdgeElementContainer: ExpoView {
             interaction.edge = edge
             edgeInteraction = interaction
             addInteraction(interaction)
+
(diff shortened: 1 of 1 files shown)


jamesacklin (commented):
Seems plausible, will wait for @dnbrwstr 's review to approve

dnbrwstr (approved):

PR #6468, opened 2026-09-06
api/app: catch floating promises that spray unhandled rejections

jamesacklin:
Fixes TLON-6487

Follow-up to #6467. That PR covered the two `useAppUpdates` pollers; this one covers the remaining `TypeError: Failed to fetch` groupings in the web Sentry project. They're all the same class of bug — **a promise nobody catches, rejecting when the network blips** — spread across three subsystems.

## Summary

| Sentry | Cause | Events | Users |
| --- | --- | --- | --- |
| [XJ](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-XJ) | vitals connection check | 1,103 | 9 |
| [1DJ](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-1DJ) [12W](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-12W) [12D](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-12D) [12C](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-12C) [13V](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-13V) | hosting heartbeat on auth failure | ~265 | 1–2 each |
| [10T](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-10T) | `markInvitesRead` | 90 | 2 |

## Changes

**Vitals connection check** — `useShipConnectionStatus`'s queryFn calls `api.checkConnectionStatus(...)` without awaiting. That's deliberate (the queryFn returns cached data immediately; the subscription callback fills it in later), but the promise floats. It fires in bursts because React Query's `onOnline` refetches every connection-status query the moment the browser reconnects — exactly when the subscribe is most likely to fail. `checkConnectionStatus` also fired an unawaited `poke`. Both now catch.

**Hosting heartbeat** — `onAuthFailure` awaited `api.getHostingHeartBeat()` uncaught. The comment above it already claimed *"this will never trigger if you're offline"* — but offline makes it **throw**, and the rejection escaped. Now caught and treated as "not known to be expired", which is what the comment intended. This only ever suppressed a logout we didn't want anyway.

**markInvitesRead** — called unawaited inside a `syncQueue` action in `ChatListScreen` and `HomeSidebar`, so its `backOff` promise floated past the queue's own try/catch. Worth noting why the fix isn't `return markInvitesRead()`: `syncQueue.add()` *rejects* the promise it hands back and neither call site consumes it, so awaiting inside the action would only **relocate** the unhandled rejection. Awaiting would also hold one of the queue's three threads for the full ~14s of backoff retries (2s + 4s + 8s), on exactly the network blip that triggered it. So it stays fire-and-forget and the catch goes directly on `markInvitesRead()`.

## Deliberately not in this PR

- **[1BQ](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-1BQ) + [1NX](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-1NX) (~207 events)** — `packages/api/src/http-api/Urbit.ts` floats `this.ack(eventId)` and `this.subscribe(pa)` inside its SSE `onmessage` handler. That file is being rewritten by #6454 (TLON-6470); folding it in would conflict. Should be picked up there or after it lands.
- **[1KY](https://tlon-corporation.sentry.io/issues/JAVASCRIPT-REACT-1KY) (62 events)** — a `scry` rejection reaching a `Promise.all` under `SyncQueue.syncNext`. This is the general case of the `markInvitesRead` note above: *any* `syncQueue.add` whose returned promise is discarded turns an action failure into an unhandled rejection. That wants a design decision (should the queue reject a promise nobody awaits?) rather than a spot fix.

## How did I test?

- `tsc --noEmit` clean on both `packages/api` and `packages/app`.
- `pnpm format:check` passes.
- No tests added, consistent with the surrounding code: these are `.catch()` clauses on paths whose only observable behavior was the Sentry report. The one behavioral assertion worth making — that a failed heartbeat no longer logs the user out — is the explicitly documented intent of the code it sits in.

## Risks and impact

Low, but not zero — unlike #6467 this touches shared packages, so it affects mobile as well as web.

Every success path is unchanged. The behavior changes are all in the failure direction, and all of them replace "throw an unhandled rejection" with "do the thing the surrounding comment said we do":

- A failed connection check leaves the status query showing its previous/placeholder value instead of rejecting into the void. Same as today from the user's perspective.
- A failed hosting heartbeat no longer logs you out. It never did — it threw first — so this only makes the existing behavior intentional.
- A failed `markInvitesRead` is logged instead of unhandled. Invites stay unread until the next focus, same as today.

Failures are logged through `createDevLogger(..., false)`, which is off by default and is not forwarded to Sentry.

## Rollback plan

Revert the commit. No migrations, no persisted state, no backend changes.

## Screenshots

n/a — no user-visible UI change.

### DIFF
--- a/packages/api/src/client/vitalsApi.ts
+++ b/packages/api/src/client/vitalsApi.ts
@@ -1,8 +1,19 @@
 import { createDevLogger } from '../lib/logger';
+import { AnalyticsEvent } from '../types/analytics';
 import * as ub from '../urbit';
 import { poke, scry, subscribe, unsubscribe } from './urbit';
 
 const logger = createDevLogger('vitalsApi', false);
+
+// Counted, not silently dropped. trackEvent routes to PostHog only; trackError
+// would also reach Sentry, which is the spray these catches exist to prevent.
+// No contactId in the payload — ship names should not go into analytics.
+function reportBackgroundFailure(context: string, e: unknown) {
+  logger.trackEvent(AnalyticsEvent.BackgroundRequestFailed, {
+    context,
+    errorMessage: e instanceof Error ? e.message : String(e),
+  });
+}
 
 export const getLastConnectionStatus = async (contactId: string) => {
   const result = await scry<ub.ConnectionUpdate>({
@@ -34,10 +45,7 @@ export const checkConnectionStatus = async (
         // Fire-and-forget from a void callback. `unsubscribe` rejects on a
         // failed channel PUT, so catch it here or it escapes unhandled.
         unsubscribe(id).catch((e) => {
-          logger.log(
-            `Failed to unsubscribe connection check for ${contactId}:`,
-            e
-          );
+          reportBackgroundFailure('vitals unsubscribe', e);
         });
       }
     }
@@ -51,7 +59,7 @@ export const checkConnectionStatus = async (
     mark: 'run-check',
     json: contactId,
   }).catch((e) => {
-    logger.log(`Failed to poke connection check for ${contactId}:`, e);
+    reportBackgroundFailure('vitals poke', e);
   });
 
   return subscription;
--- a/packages/api/src/client/vitalsApi.ts
+++ b/packages/api/src/client/vitalsApi.ts
@@ -5,8 +5,9 @@ import { poke, scry, subscribe, unsubscribe } from './urbit';
 
 const logger = createDevLogger('vitalsApi', false);
 
-// Counted, not silently dropped. trackEvent routes to PostHog only; trackError
-// would also reach Sentry, which is the spray these catches exist to prevent.
+// Counted, not silently dropped. trackEvent keeps this in PostHog; trackError
+// would report as `app_error`, which the composite logger forwards to Sentry —
+// the spray these catches exist to prevent.
 // No contactId in the payload — ship names should not go into analytics.
 function reportBackgroundFailure(context: string, e: unknown) {
   logger.trackEvent(AnalyticsEvent.BackgroundRequestFailed, {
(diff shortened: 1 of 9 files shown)


packages/api/src/client/vitalsApi.ts
```diff
@@ -1,6 +1,21 @@
...
 import * as ub from '../urbit';
 import { poke, scry, subscribe, unsubscribe } from './urbit';
 
+const logger = createDevLogger('vitalsApi', false);
+
+// Counted, not silently dropped. trackEvent keeps this in PostHog; trackError
+// would report as `app_error`, which the composite logger forwards to Sentry —
+// the spray these catches exist to prevent.
+// No contactId in the payload — ship names should not go into analytics.
+function reportBackgroundFailure(context: string, e: unknown) {
```
arthyn: if this took a `logger` as a param you could put it in a more generic place like utils or analytics so everything could use it. also if you made this curried you could just do stuff like `.catch(reportBackgroundFailure(logger, 'some reason'))`

arthyn (changes requested):
I would consider making the util, and using it in all the catches here. I would also remove all the comments added here since they're just noise.

the bigger question is are these errors useful in general, but at least this puts them somewhere and we can clear the console. this makes me wonder if there's a more generic place we could catch these though.

packages/api/src/client/vitalsApi.ts
jamesacklin: Done in 896834f — curried, takes the logger, and moved to `packages/api/src/lib/logger.ts` next to `createDevLogger` (it needs `AnalyticsEvent`, which is a dependency-free enum, so no cycle). Exported from the api root so `packages/app` shares the same one. Call sites now read `.catch(reportBackgroundFailure(logger, 'vitals poke'))`. Comments dropped too. Net 28 lines lighter.

The hosting-heartbeat site is the one exception — its catch has to return `null` to feed the expired check, so it keeps a block body.

arthyn (approved):

PR #6473, opened 2026-09-08
Add support@tlon.io links to offline and stopped-node screens

jamesacklin:
## Summary

The app's offline screen and the stopped-node (hosting) screen are both dead ends — the offline screen offers no action at all, and the stopped-node screen leaves the user waiting with nothing to do if their node never wakes up. This adds a pressable `support@tlon.io` address to each.

Fixes TLON-6499

## Changes

- New `EmailSupportLink` component (`packages/app/ui/components/EmailSupportLink.tsx`): renders a prompt (default `Need help? Email`) followed by the underlined support address, and hands off to the mail app via `openComposer`. If the device has no mail app, it shows an alert — the address stays on screen either way. Follows the existing `TlonText.Text` + `TlonText.RawText` link pattern used by the onboarding screens.
- `App.main.tsx`: added the link to the offline screen, subject `Help! I can't connect to Tlon.`. Sized `$label/l` so it isn't fine print under the large offline message; the surrounding `View` became a `YStack` to get gap spacing.
- `GettingNodeReadyScreen.tsx`: added the link below the "Feel free to close the app…" footnote, subject `Help! My node won't wake up.`
- Added `SUPPORT_EMAIL` to `packages/app/constants.ts` and routed the two pre-existing hardcoded copies of the address (`AppInfoScreen`, `UnderMaintenance`) through it, so it lives in one place.

Note: the separate `UnderMaintenance` ("Needs Repair") screen already had an Email Support button, so it was left as-is apart from the constant swap. It only appears after a re-check still reports maintenance — happy to make that unconditional in a follow-up if that's wanted.

## How did I test?

- `pnpm tsc` clean in both `packages/app` and `apps/tlon-mobile`.
- `pnpm format:check` clean; `pnpm lint` produces no new warnings in the touched files.
- Rendered both placements in React Cosmos (temporary fixture, removed before commit) at mobile (375×812) and desktop widths to check copy, underline, and spacing.
- Not exercised on a device: the offline branch requires an actual network drop and the stopped-node screen requires a paused hosted node. The `openComposer` call is the same pattern already shipping in `UnderMaintenance` and `AppInfoScreen`.

## Risks and impact

  - [x] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [ ] Other:

Additive, presentation-only. Worst case is a mail composer that doesn't open, which falls back to an alert; the address is visible as text regardless.

## Rollback plan

Revert the commit. No migrations, no persisted state, no backend changes.

## Screenshots / videos

Offline screen (Cosmos, mobile width): the existing "You are offline. Please connect to the internet and try again." message, with `Need help? Email support@tlon.io` centered below it, the address underlined.

### DIFF
--- a/apps/tlon-mobile/src/App.main.tsx
+++ b/apps/tlon-mobile/src/App.main.tsx
@@ -209,6 +209,7 @@ const MainApp = () => {
           </Text>
           <EmailSupportLink
             size="$label/l"
+            prompt="Back online and still stuck? Email"
             subject="Help! I can't connect to Tlon."
           />
         </YStack>
--- a/apps/tlon-mobile/src/App.main.tsx
+++ b/apps/tlon-mobile/src/App.main.tsx
@@ -23,10 +23,12 @@ import { AppDataProvider } from '@tloncorp/app/provider/AppDataProvider';
 import { BaseProviderStack } from '@tloncorp/app/provider/BaseProviderStack';
 import {
   AgentOnboardingSequence,
+  EmailSupportLink,
   LoadingSpinner,
   SplashSequence,
   Text,
   View,
+  YStack,
   ZStack,
   usePreloadedEmojis,
 } from '@tloncorp/app/ui';
@@ -195,16 +197,21 @@ const MainApp = () => {
           <OnboardingStack />
         )
       ) : (
-        <View
+        <YStack
           height="100%"
           padding="$l"
+          gap="$3xl"
           justifyContent="center"
           alignItems="center"
         >
           <Text textAlign="center" fontSize="$xl" color="$primaryText">
             You are offline. Please connect to the internet and try again.
           </Text>
-        </View>
+          <EmailSupportLink
+            size="$label/l"
+            subject="Help! I can't connect to Tlon."
+          />
+        </YStack>
       )}
       <StatusBar
         backgroundColor={isDarkMode ? 'black' : 'white'}
--- a/apps/tlon-mobile/src/screens/Onboarding/GettingNodeReadyScreen.tsx
+++ b/apps/tlon-mobile/src/screens/Onboarding/GettingNodeReadyScreen.tsx
@@ -16,6 +16,7 @@ import {
 import {
   AppDataContextProvider,
   ArvosDiscussing,
+  EmailSupportLink,
   IconType,
   ListItem,
   LoadingSpinner,
@@ -245,6 +246,7 @@ export function GettingNodeReadyScreen({
               Feel free to close the app if this takes too long. We’ll send you
               a notification when your node is ready.
             </TlonText.Text>
+            <EmailSupportLink subject="Help! My node won’t wake up." />
           </YStack>
           <StoppedNodePushSheet
             notifPerms={notifPerms}
--- a/apps/tlon-mobile/src/screens/Onboarding/GettingNodeReadyScreen.tsx
+++ b/apps/tlon-mobile/src/screens/Onboarding/GettingNodeReadyScreen.tsx
@@ -238,15 +238,17 @@ export function GettingNodeReadyScreen({
                 )}
               </ListItem.EndContent>
             </ListItem>
-            <TlonText.Text
-              size="$label/s"
-              color="$secondaryText"
-              textAlign="center"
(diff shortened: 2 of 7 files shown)


packages/app/ui/components/EmailSupportLink.tsx
```diff
@@ -0,0 +1,52 @@
...
+      size="$label/s"
+      color="$secondaryText"
+      textAlign="center"
+      {...textProps}
+    >
+      {prompt}{' '}
+      <TlonText.RawText
+        accessible
+        accessibilityRole="link"
+        accessibilityLabel={`Email ${SUPPORT_EMAIL}`}
```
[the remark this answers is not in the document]
jamesacklin: Good catch — moved press + accessibility onto a `Pressable` wrapper (matching the existing TlonLogin pattern) so the line is one activatable element; the inner address text is now purely decorative and the whole line is the tap target.

apps/tlon-mobile/src/screens/Onboarding/GettingNodeReadyScreen.tsx
```diff
@@ -245,6 +246,7 @@ export function GettingNodeReadyScreen({
               Feel free to close the app if this takes too long. We’ll send you
               a notification when your node is ready.
             </TlonText.Text>
+            <EmailSupportLink subject="Help! My node won’t wake up." />
```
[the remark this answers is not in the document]
jamesacklin: Grouped the link with the footnote it continues in a `$m` stack instead of letting it take a full `$3xl` slot, cutting the height this PR adds from 48pt to 24pt; the screen's fixed-height layout is tight on 320x568 independent of this change, so I left restructuring it to a follow-up rather than reworking an onboarding screen here.

apps/tlon-mobile/src/App.main.tsx
```diff
@@ -195,16 +197,21 @@ const MainApp = () => {
...
           padding="$l"
+          gap="$3xl"
           justifyContent="center"
           alignItems="center"
         >
           <Text textAlign="center" fontSize="$xl" color="$primaryText">
             You are offline. Please connect to the internet and try again.
           </Text>
-        </View>
+          <EmailSupportLink
```
dnbrwstr, line 210: Is it confusing to show this in a screen that exists explicitly to warn the user they currently should not be able to connect to Tlon? The offline screen only shows up when there's actually no network connection, not just when it can't connect to the ship etc.

apps/tlon-mobile/src/screens/Onboarding/GettingNodeReadyScreen.tsx
```diff
@@ -237,14 +238,17 @@ export function GettingNodeReadyScreen({
...
+            <YStack gap="$m">
+              <TlonText.Text
+                size="$label/s"
+                color="$secondaryText"
+                textAlign="center"
+              >
+                Feel free to close the app if this takes too long. We’ll send
+                you a notification when your node is ready.
+              </TlonText.Text>
+              <EmailSupportLink subject="Help! My node won’t wake up." />
```
dnbrwstr: I would delay showing this by, idk, 20 seconds? So it's only visible in cases where there actually might be something wrong.

dnbrwstr (approved):

apps/tlon-mobile/src/screens/Onboarding/GettingNodeReadyScreen.tsx
jamesacklin: Done — gated behind a 20s timer (`SUPPORT_LINK_DELAY`), so it only appears once the wait is long enough that something might actually be wrong.

apps/tlon-mobile/src/App.main.tsx
jamesacklin: Fair point — the offline link was an explicit ask for this PR so I've kept it rather than dropping it unilaterally, but reworded to "Back online and still stuck? Email support@tlon.io" so it reads as the next step after reconnecting instead of an action they can't take; happy to remove it entirely if you'd still rather it not be there.

PR #6484, opened 2026-09-09
Disallow creating new %diary channels

jamesacklin:
## Summary

%notes replaces %diary, so this closes every path in the app that still mints a diary channel. Existing diary channels are untouched — they stay readable, writable and moderatable; only creation is refused.

Fixes TLON-6480

## Changes

**One chokepoint.** `createChannel` (`packages/shared/src/store/channelActions.ts`) refuses `channelType: 'notebook'` before its optimistic insert, so nothing has to be rolled back and nothing reaches the backend. It fires the new `Blocked Diary Channel Creation` analytics event and throws `DiaryCreationBlockedError`. That single guard covers the create-channel sheet, its private-channel follow-on, and the channel-as-template flow.

**Create-channel sheet.** The diary option is gone; the sheet always offers `[Chat, Notebook (%notes), Gallery]`. The `useNotesDeskAvailable` gate went with it — %notes runs on every ship via the `%kiln-rein` override, and the probe returned `false` on *any* error, so one transient failure was enough to hide the longform option entirely. That hook had no other callers and is deleted.

**Group templates.** All 7 carried a diary channel ("Reviews", "Recipes", "Playlists", …); they now carry `type: 'notes'`. A %notes channel can't ride the %groups create poke the way %channels kinds can — it's created against the %notes API and then bound to the group — so `createGroupFromTemplate` sends chat/gallery in the poke and adds each notebook afterwards, via the same path `agentGroupOnboarding` already uses. A notebook that fails is logged and skipped rather than rejecting, so a group that was in fact created is never reported to the caller as a failure.

**Channel-as-template.** "Use channel as template" is hidden for bulletins in `ChatOptionsSheet`, since templating copies the source channel's type. `ChannelFromTemplateView` refuses directly as well, because that screen is also reachable by direct navigation.

**Naming.** The manage-channels list rendered `capitalize(channel.type)`, so a channel the sheet had just called "Notebook" listed as "Notes". A new `getChannelTypeLabel` sits beside the existing `getChannelTypeIcon` — %notes is "Notebook", %diary is "Bulletin" — and is used by both the list and `ChannelHeader`, which drops its own `notesAvailable` conditional for the same reason the sheet did.

## How did I test?

`pnpm -r tsc`, `pnpm run lint:all` (warnings only, all pre-existing) and `pnpm format:check` are clean. Unit suites pass: shared 762, api 882, app 623.

Three new unit tests:
- `createChannel` refuses a diary channel — asserts no poke is sent and no channel row is left behind, since the guard runs ahead of the optimistic insert.
- `createGroupFromTemplate` keeps notes out of the group poke and creates it after.
- `createGroupFromTemplate` still returns the group when its notebook fails.

E2E: `create-group.spec.ts` expectations moved to the %notes type and the "Notebook" label. The four `notebook-*.spec.ts` specs cover legacy diary read/write — precisely the behavior that must keep working — and used to build their fixture through the create sheet. They now call a new `createDiaryChannel` helper that pokes `%channels` with `channel-action-2` directly, reading the group flag out of the page URL.

Both were then run against local rube ships:

- `notebook-creation.spec.ts` **passed**, confirming the new `createDiaryChannel` poke helper works against a real ship.
- `create-group.spec.ts` initially failed on `ChannelItem-Notebook-2` / `ChannelItem-Reviews-2`. The notebook was in fact created correctly — the failure was ordering. `verifyGroupChannels` pinned the last expected channel to a fixed index, and a template's notebook is now created after the group rather than in the poke, so it lands at index 0 rather than last. Fixed in `8b0c8bb22d` by asserting the channel count instead; the per-channel checks below it already matched any index.

`should create a quick group with default single channel` was failing across branches. **That fix now lives on `develop` (`51b3d3b048`) and is merged in here; this PR no longer carries its own.**

For the record, since I initially misdiagnosed it: the primary fault is a strict-mode violation, not a timeout. The `groupDestination` locator added in #6450 is `getByTestId('ChannelHeaderTitle').or(getByTestId('GroupChannelsHeaderTrigger'))`, and on desktop both panes render side by side, so it resolves to two elements. Whether the second pane has rendered when the locator first resolves is a race, which is what made it intermittent from #6450's merge onward rather than starting at a particular time. The empty-title testID problem I found (`??` not falling back on `''`) is real but secondary: it made the `catch` fallback unreachable for a quick group, turning a recoverable race into a hard failure. `develop`'s fix addresses both and covers the channel branch too, so my partial version was dropped in the merge.

## Risks and impact

  - [x] Onboarding — group templates now create their notebook in a second step
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display — channel-type labels changed in the manage-channels list and channel header
  - [ ] Notifications
  - [x] Other: channel creation

Two behavior changes worth a reviewer's attention:

1. **No longform option if %notes is genuinely missing.** Previously a ship without the notes desk fell back to offering diary. It now offers only Chat and Gallery, and a notes creation would error. This is deliberate — %notes has been on every ship since TLON-6113 — but it is the one case that gets worse rather than better.
2. **Template groups take a second round trip.** Their notebook arrives just after the group rather than with it, and can be missing if the %notes call fails. The group is still created and usable, and the failure is reported through `trackError`.

## Rollback plan

`git revert` the two commits on this branch. Nothing is persisted or migrated, and no backend change is involved — %diary is untouched on the ship side, so reverting restores the old creation options immediately. Diary channels created before or after this change behave identically either way.

## Screenshots / videos

n/a — the visible change is the removal of one option from the create-channel sheet and two channel-type labels.

## Not in this PR

`scaffoldPersonalGroup` (`packages/shared/src/store/groupActions.ts`) still builds a diary channel, but it is dead code — exported with no callers anywhere in the monorepo. It is now the only place in app code that constructs one. Deleting it is tracked separately; note that `getPersonalGroupKeys().notebookChannelId` must stay, since `wayfinding.ts` uses it for existing personal groups.


packages/app/ui/components/ChannelFromTemplateView.tsx
```diff
@@ -66,6 +68,14 @@ export function ChannelFromTemplateView({
         return;
       }
 
+      // A template copies the source channel's type, and bulletins (%diary) can
+      // no longer be created. ChatOptionsSheet hides the entry point for them,
+      // but this screen is also reachable by direct navigation.
+      if (isDiaryChannelType(channel.type)) {
```
dnbrwstr: Afaict this is unnecessary -- not possible to access this function with directly modifying code in which case `createChannel` would fail anyway.

dnbrwstr (approved):
Looks fine but I would remove some of the error plumbing around blocking diary creation attempts -- afaict there are no code paths that reach creation with a diary value, even the identified "direct navigation" thing isn't actually reachable. If we want to validate i would just do a check at channel creation time that the type is one of our blessed types, throw if it's not, and leave it at that.

packages/app/ui/components/ChannelFromTemplateView.tsx
jamesacklin: You're right — removed in b6613e243a. I'd justified it as "reachable by direct navigation", but ChannelTemplate has no entry in linking.ts, so there's no URL path to it; the only caller is onPressChannelTemplate from ChatOptionsSheet, where the action is already hidden for bulletins. `DIARY_CREATION_BLOCKED_MESSAGE` lost its last outside consumer with it, so it's no longer exported.

PR #6493, opened 2026-09-10
mobile: mandate valid hosting authentication

latter-bolden:
## Summary

We want to move to enforcing Hosting auth remain valid — sessions are sufficiently long lived now and without it we have no way of querying Hosting about the status of the ship.

Going forward, we'll attempt to refresh the Hosting auth during background sync.

The major downside here is users who are logged in with already expired Hosting auth. To combat the poor UX of having to fully log them out, we add a new screen that reduces the friction of re-authenticating:

Security Check-selection

## Changes

- Report actual background-sync outcomes (Fixes TLON-6404): propagate sync and contact-discovery failures to telemetry and the scheduler, correlate sync errors by task ID, and distinguish skipped auth/session checks from completed syncs.
- Detects confirmed expired Hosting auth without affecting self-hosted or unknown states.
- Locks authenticated app surfaces behind a new “Security check” OTP screen.
- Automatically sends an OTP to the account’s email or phone, with resend and logout options.
- Restores and persists the renewed Hosting session after verification.
- Checks auth during startup, foregrounding, and background sync.
- Preserves normal login and access-code recovery flows.
- Adds light/dark Cosmos fixtures and API/mobile test coverage.

## How did I test?

Backend endpoints are on Hosting test env, testing in progress...

Background-sync outcome follow-up: 95 targeted API/shared/app/mobile tests pass, including auth refresh, session changes, success, failure, cancellation, stale-data discard, and contact-discovery failures. Shared/app/mobile typechecks pass. This validates outcome reporting; the production client-startup cause remains inconclusive, and no additional device verification was performed.

## Risks and impact

  - [x] Other: Auth

### DIFF
--- a/apps/tlon-mobile/cosmos.imports.ts
+++ b/apps/tlon-mobile/cosmos.imports.ts
@@ -80,6 +80,7 @@ import * as fixture75 from './src/App.fixture';
 import * as fixture76 from './src/fixtures/InputToolbar.fixture';
 import * as fixture77 from './src/fixtures/Onboarding.fixture';
 import * as fixture78 from './src/fixtures/SetNicknameScreen.fixture';
+import * as fixture79 from './src/fixtures/HostingAuthReconnectScreen.fixture';
 import * as decorator1 from './src/fixtures/cosmos.decorator';
 
 export const rendererConfig: RendererConfig = {
@@ -252,6 +253,7 @@ const fixtures = {
     module: fixture74,
   },
   'src/App.fixture.tsx': { module: fixture75 },
+  'src/fixtures/HostingAuthReconnectScreen.fixture.tsx': { module: fixture79 },
   'src/fixtures/InputToolbar.fixture.tsx': { module: fixture76 },
   'src/fixtures/Onboarding.fixture.tsx': { module: fixture77 },
   'src/fixtures/SetNicknameScreen.fixture.tsx': { module: fixture78 },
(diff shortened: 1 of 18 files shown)


arthyn (commented):
a few issues that seem like we should address before merging from the clanker verified by other clanker: 

1. The ship-auth failure path still force-logs-out the exact users this PR is for. In useConfigureUrbitClient.ts, the client's onAuthFailure handler calls the Hosting heartbeat and
  runs logout() when it returns expired. A long-idle hosted user often has both an expired Hosting session and a lapsed ship cookie. Their first request 401s, getCode tries Hosting,
  fails, and the handler wipes the session and drops them to onboarding before the gate ever renders. This path should set hostingAuthExpired and hand off to the gate instead of
  logging out.

  2. Pat's remount comment from Sep 10 is still unaddressed. When the gate flips to expired, AuthenticatedAppContent unmounts. After OTP success it remounts and calls syncStart()
  cold. I verified the chain: isSyncing resets in a finally, configureClient reuses the existing client object, and subscribe in the api package always opens a fresh subscription id
  with no path-level dedupe. So the same eyre channel ends up with two copies of every subscription and every event is handled twice until logout. The cleanest fix is to keep the
  content mounted and render the reconnect screen as an overlay, the way authenticatedOverlay already works. That also preserves navigation state.

  4. The tlonbot revival prompt bypasses the gate. maybeShowPrompt still calls refreshHostingAuth({ force: true }) and discards the result. If that call is the one that detects
  expiry, the flag is written but hostingAuthState stays valid, the revival sheet opens, and its Hosting calls 401. The gate only catches up on the next foreground. It should call
  requireHostingAuth or drop the refresh, since the status handler already checked a moment earlier.

PR #6503, opened 2026-09-10
logs: fingerprint every %fail and send crashes to PostHog error tracking

arthyn:
## Why

Both PostHog and Grafana classify backend crashes today with hand-maintained substring lists over stack traces (140 entries in the PostHog "Unknown Crashes" series, a regex in the Grafana alert). Those lists key on `file:line:col` spans, so every commit rots them, and nobody can tell a new crash from a known one without editing a dashboard.

## What

**`%logs` names every crash before sending it.** `+fingerprint:logs` builds a signature from parts that survive a recompile:

```
/gall/groups | groups failed | nest-fail | on-agent/contact/fact | -have.@p -need.%full | app/groups>sys/vane/gall
```

origin; the developer's message (first line of the echo); the first error tag in the tang (why); the last tag, normally a `~|` dispatch hint (where); type/status detail from `-have`/`-need` lines and `code=` tokens; and the chain of source files in the stack with line numbers stripped. The `fingerprint` is eight hex digits of its `+mug`. `fingerprint_exact` additionally hashes the innermost span with its line numbers, so it separates crash sites within a release while `fingerprint` stays constant across releases. All three ride along as `log-data`, so they reach PostHog as event properties and Loki as OTLP attributes with no collector changes. Gall's own wire labels (`watch-ack`, `poke-fail`), commit markers, `key=value` data dumps, and bare ship names are never used, so user data can't leak into a signature and cardinality stays bounded.

**Dispatch hints so unexpected crashes have a "where".** Most `%fail` volume outside the known signatures has no message ("groups failed", or nothing) and a generic tag like `nest-fail`, which isn't enough to tell bugs apart. A `~|` hint at the top of `on-poke`, `on-watch`, `on-agent` and `on-arvo` in **groups, channels, notify and presence** puts the mark (or wire head and sign tag) into the tang of every crash beneath it, for free at runtime, so the fingerprint gets `on-poke/group-action-4` or `on-agent/contact/fact` without anyone tagging individual sites. Other agents can adopt the same four lines as needed.

**Crashes also go to PostHog Error Tracking.** `ted/posthog.hoon` now posts to `/batch/` and, for `%fail` events with a fingerprint, adds a `$exception` event carrying `$exception_fingerprint` and an `$exception_list` entry with the message and the stack as raw frames. That gets grouping, first-seen, and regression alerts from the product; the issue is named after the signature. The existing `Backend Log` event is unchanged so current dashboards keep working.

**Every event names the ship that caused it.** The logging door attaches `src` (`src.bowl`) to every event: the peer that nacked, kicked, or sent the fact in `on-agent`, the poker or subscriber in `on-poke`/`on-watch`, ourselves elsewhere. The `on-agent` hints carry `src.bowl` too, so unexpected crashes there show the peer in the trace. That is what separates "one host is nacking everyone" from "everyone is nacking". Ship names stay out of the fingerprint, so grouping is unaffected.

**Convention, no code:** call sites that log an *expected* failure should pass `'code'^s+'<stable-name>'` in `log-data`. Anything without a code is unexpected by definition. Dashboards and alerts can key on `code` presence instead of denylists as call sites adopt it.

## End-to-end check (2026-09-11, ~simtyc on this branch)

Synthetic `%fail` events poked into `%logs` on the dev moon, plus a real `%groups` crash from a bad poke:

- The real crash's nack tang carried the new `[%on-poke %noun]` dispatch hint.
- PostHog received the `Backend Log` event with `fingerprint`, `fingerprint_exact` and `signature`, and the companion `$exception` event. With the documented raw-frame shape (`platform`, `lang`, `lineno`, `colno`, `resolved`, plus `$lib`) PostHog Error Tracking created an issue for it; without those fields the event was stored but never became an issue, which is why the frame shape matters.
- With `%logs` pointed at the primary OTLP ingress, the record landed in primary Loki with the same attributes and with `ship` as a label, confirming #6481 too.

## Tests

- `tests/lib/logs.hoon`: signature shape, fingerprint invariance under line-number changes (and `fingerprint_exact` sensitivity), bracketed `%tags` with knots, dispatch hint + have/need detail, `code=` detail with a ship name dropped, tag-less tangs, `%tell` has none, and the door stamps `src` on every event.
- `tests/app/logs.hoon`: updated for the two new properties on `%fail`.
- Compiled and run on the ~simtyc dev moon: all eight `tests/lib/logs` tests pass; `tests/app/logs/test-log-fail` matches on the two new properties and differs only on the pre-existing `commit` value, which is `development` in the repo but a real hash on any ship with a committed desk.

## Follow-ups

- Dedupe repeated fingerprints per ship inside `%logs` (a bot moon currently logs the same join failure every two minutes).
- Grafana alert on a fingerprint first seen in the last hour on 3+ ships exists but is paused until the hint sweep is live and the top signatures look like distinct bugs.

Part of TLON-6530

### DIFF
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -214,6 +214,7 @@
   ++  on-poke
     %-  on-poke:guard
     |=  =rail
+    ~|  [%on-poke -.rail]
     %-  step:un:guard
     ^-  (quip card _this)
     =^  cards  state
@@ -222,6 +223,7 @@
   ::
   ++  on-watch
     |=  =path
+    ~|  [%on-watch ?~(path %$ i.path)]
     %-  step:un:guard
     ^-  (quip card _this)
     =^  cards  state
@@ -240,6 +242,7 @@
   ++  on-agent
     %-  on-agent:guard
     |=  [=wire =sign:guard]
+    ~|  [%on-agent ?~(wire %$ i.wire) -.sign]
     %-  step:un:guard
     ^-  (quip card _this)
     =^  cards  state
@@ -248,6 +251,7 @@
   ::
   ++  on-arvo
     |=  [=wire sign=sign-arvo]
+    ~|  [%on-arvo ?~(wire %$ i.wire)]
     %-  step:un:guard
     ^-  (quip card _this)
     =^  cards  state
--- a/desk/app/channels.hoon
+++ b/desk/app/channels.hoon
@@ -242,7 +242,7 @@
   ++  on-agent
     %-  on-agent:guard
     |=  [=wire =sign:guard]
-    ~|  [%on-agent ?~(wire %$ i.wire) -.sign]
+    ~|  [%on-agent ?~(wire %$ i.wire) -.sign src.bowl]
     %-  step:un:guard
     ^-  (quip card _this)
     =^  cards  state
--- a/desk/app/groups.hoon
+++ b/desk/app/groups.hoon
@@ -216,7 +216,7 @@
   ::
   ++  on-agent
     |=  [=wire =sign:agent:gall]
-    ~|  [%on-agent ?~(wire %$ i.wire) -.sign]
+    ~|  [%on-agent ?~(wire %$ i.wire) -.sign src.bowl]
     %-  step:un:guard
     ^-  (quip card _this)
     =^  cards  state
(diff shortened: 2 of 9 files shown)


arthyn:
@mikolajpp going to merge but still interested to see what you think

PR #6514, opened 2026-09-11
openclaw: state the owner-only policy in the blocked-tool reason

patosullivan:
## Summary

Non-owner requests blocked from the owner-only `tlon`, `cron`, or `read` tools now receive an explicit policy reason, with instructions to explain the denial without inventing outages or reload problems. OpenClaw core hands the hook's `blockReason` to the model verbatim as a plain, non-error tool result (verified on 2026.5.28 through 2026.8.2), and the old `The <tool> tool is not available.` led bots to confabulate: one told a group the upload tool "isn't exposed after the reload", another told a DM sender the Tlon CLI was unreachable. The restricted tools and the authorization rule are unchanged. Linear: https://linear.app/tlon/issue/TLON-6363

## Changes

- `packages/openclaw/src/owner-only-tools.ts` (new): the owner-only tool set, the role rule (only an explicit `user` role blocks; owner and role-less internal sessions pass), and the reason text, one line of 197 characters so it survives the first-line caps observers apply (core's `after_tool_call` error at 400, the owner "silent failure" notice at 200).
- `packages/openclaw/index.ts`: the `before_tool_call` gate takes its decision and reason from that module.
- `packages/openclaw/src/monitor/index.ts`: the ContextLens `ownerOnlyAvailable` advertisement reads the shared set.
- `packages/openclaw/test/cases/04-security.test.ts`: the non-owner case also asserts, scoped by `tool_call_id`, that the follow-up model call carried exactly the new reason.
- `packages/openclaw/SECURITY.md` §13: documents the reason and says "DM or group" where it said "DM", since group senders are gated identically.

Not included: no change to the gating rule or tool set, no prompt or tool-description changes, no Hermes change (its adapter already states its policy).

## How did I test?

- In `packages/openclaw`: `pnpm tsc`, `pnpm test` (1733 tests, including a new unit file pinning the tool set, role rule, and reason text) and `oxfmt --check` clean; no new `oxlint` warnings. Six mutations (old wording, flipped role rule, inserted newline, dropped `read`, reordered set, dropped instruction sentences) each fail the new unit file.
- Shared Docker e2e `04-security` (fake model, ephemeral fakezods): 11 passed, 1 skipped on OpenClaw 2026.5.28 (the image default) and again on 2026.7.1 (production). The shared compose passes no build args, so the 2026.7.1 run was made by temporarily setting the `OPENCLAW_CORE_VERSION` default in `packages/openclaw/dev/Dockerfile.test`; the image build logged `openclaw-core-version=2026.7.1`. That file is not part of this PR.
- The e2e checks the reason delivered to the model; the model's reply is scripted (`Tool rejected by policy.`), so real-model wording is not exercised. The case scripts a reply and therefore does not exercise the owner "silent failure" notice.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: OpenClaw bot plugin tool gate (model-facing text only)

Owner-visible surfaces (the Lens tool-run error and, on 2026.7.1 which counts a block as a tool error, the owner "silent failure" notice) now carry the longer reason.

## Rollback plan

Revert the merge commit. No config, data, or ship-side migration; the plugin reaches bots through the usual develop → master → fleet plugin refresh.

## Screenshots / videos

None.

### DIFF
--- a/packages/openclaw/SECURITY.md
+++ b/packages/openclaw/SECURITY.md
@@ -378,21 +378,21 @@ try {
 | Scenario | Behavior |
 | -------- | -------- |
 | Owner uses restricted tool | ✅ Allowed |
-| Non-owner uses restricted tool | ❌ Blocked with error message |
+| Non-owner uses restricted tool (DM or group) | ❌ Blocked; the tool result tells the model the tool is owner-only and what to say |
 | Non-owner tricks LLM into using tool | ❌ Still blocked (hook-level enforcement) |
-| Internal session (heartbeat, cron) | ✅ Allowed (no role = not a user DM) |
+| Internal session (heartbeat, cron) | ✅ Allowed (no stored role = not a user-initiated turn) |
 
 **Implementation:**
-- `before_tool_call` hook intercepts calls to restricted tools
-- Checks SenderRole from session tracker
-- Only blocks when role is explicitly `"user"` (non-owner DM)
+- `before_tool_call` hook intercepts calls to restricted tools (policy in `src/owner-only-tools.ts`: `OWNER_ONLY_TOOLS`, `resolveOwnerOnlyToolBlock`)
+- Checks SenderRole from session tracker (stored the same way for DM and group senders)
+- Only blocks when role is explicitly `"user"` (a non-owner sender, DM or group)
 - Owner sessions (`"owner"`) and internal sessions (`undefined` role) are allowed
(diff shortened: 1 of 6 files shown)


arthyn (approved):
lgtm

PR #6523, opened 2026-09-12
fix(chat): show one tombstone when a message is deleted

janicduplessis:
## Summary

Deleting a chat message in a group channel can render two "Message deleted" rows for the one post and raise a duplicate-key error from the message list (`Encountered two children with the same key` on FlatList; `[legend-list] Error: Detected overlapping key` plus a blank gap on LegendList). It only shows when the deleted post is already in the channel's loaded pages, which is why the ticket's steps reproduce intermittently.

The `/v4` `%channels` subscription encodes a deleted post as a `%set` whose value is a tombstone object (`type: 'tombstone'`), not `null` ([`channel-json.hoon#L216`](https://github.com/tloncorp/tlon-apps/blob/ce462c16fd11b384327f5eae835ba7cd42f30365/desk/lib/channel-json.hoon#L216)). `toChannelsUpdate` only checked `set !== null`, so every delete decoded as `addPost`. The tombstone then lands in the channel's in-memory `newPosts` with a `sentAt` derived from the post id rather than `essay.sent` ([`postsApi.ts#L1288`](https://github.com/tloncorp/tlon-apps/blob/ce462c16fd11b384327f5eae835ba7cd42f30365/packages/api/src/client/postsApi.ts#L1288)), and `mergePendingPosts` dedupes `newPosts` against the loaded pages by `sentAt`, so both copies survive.

## Changes

- `channelsApi.ts`: a `%set` whose value is a tombstone decodes to the existing `deletePost` update, for posts and replies, so it takes the `markPostAsDeleted` path instead of `handleAddPost`. This is the path the `null` encoding took before tombstones existed.
- `sync.ts`: the `deletePost` case also calls `deleteFromChannelPosts`, the overlay a local delete uses, so a post that is only in `newPosts` (sent this visit, deleted elsewhere) is tombstoned or hidden at once instead of staying live until remount.
- `sync.ts`: tombstones in the changes feed (used on resume) go to `deleteFromChannelPosts` instead of `addToChannelPosts`; `insertChanges` still upserts the row.
- Tests: `channelsApi.test.ts` and `deletePostListeners.test.ts` cover the decoding and the listener routing; each fails without its change.

Not changed: a reply tombstone carries recomputed reply `meta` that the `deletePost` update drops, so the parent's reply count refreshes on the next thread fetch. The old path did not apply it either.

## How did I test?

iOS simulator (iPhone 17, iOS 26.5) and Android emulator, debug builds against a dev ship, "Show deleted messages" on in Settings > Appearance, in a throwaway solo group's chat channel.

Before, both platforms: send a message, leave the channel, re-enter it, long-press the message, Delete message, confirm. Observed: a blank gap where the tombstone should be, the red error toast, and `[legend-list] Error: Detected overlapping key (<post id>)` in the logs. The Android client sitting in the same channel also hit the error when the post was deleted from iOS.

After, same steps on both platforms: one "Message deleted" row, no gap, no toast, `stim logs --errors` clean.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

**Worst case if wrong: a remote delete is not reflected until the channel is reopened.** Scope is decoding of group-channel post and reply delete events on the `/v4` subscription and the changes feed, all platforms; DMs use `%chat` and are not touched. `markPostAsDeleted` is an update, so a tombstone for a post the client never received writes nothing; the changes feed and page backfill still insert it. No query keys, table dependencies, or schema changed.

## Rollback plan

Revert the PR.

## Screenshots / videos

iOS before:

https://github.com/user-attachments/assets/ae0b1ae5-da36-4c1a-94aa-73955e2a191d

iOS after:

https://github.com/user-attachments/assets/7d7ac80a-66f9-41bb-8cb0-398dd375a451

Android before:

https://github.com/user-attachments/assets/dece12bb-8de8-4f64-9e71-6c19e96d7950

Android after:

https://github.com/user-attachments/assets/fa460611-3ef2-45c3-9917-677bdb66e6b9

### DIFF
--- a/packages/api/src/client/channelsApi.test.ts
+++ b/packages/api/src/client/channelsApi.test.ts
@@ -1,0 +1,72 @@
+import { expect, test } from 'vitest';
+
+import type * as ub from '../urbit';
+import { toChannelsUpdate } from './channelsApi';
+
+const nest = 'chat/~zod/test';
+const postId = '170.141.184.508.156.853.761.867.717.508.987.879.424';
+const replyId = '170.141.184.508.156.856.230.907.384.944.729.784.320';
+
+const tombstone: ub.PostTombstone = {
+  id: postId,
+  author: '~zod',
+  seq: 5,
+  'deleted-at': 1789187650000,
+  type: 'tombstone',
+};
+
+const post: ub.Post = {
+  seal: {
+    id: postId,
+    reacts: {},
+    replies: null,
+    meta: { replyCount: 0, lastRepliers: [], lastReply: null },
+    seq: 5,
+  },
+  essay: {
+    content: [{ inline: ['hi'] }],
+    author: '~zod',
+    sent: 1789187602325,
+    kind: '/chat',
+    blob: null,
+    meta: null,
+  },
+  type: 'post',
+};
+
+function postEvent(rPost: ub.PostResponse): ub.ChannelsSubscribeResponse {
+  return {
+    nest,
+    response: { post: { id: postId, 'r-post': rPost } },
+  } as ub.ChannelsSubscribeResponse;
+}
+
+test('a live post set decodes to addPost', () => {
+  const update = toChannelsUpdate(postEvent({ set: post }));
+  expect(update.type).toBe('addPost');
+});
+
+test('a tombstone post set decodes to deletePost', () => {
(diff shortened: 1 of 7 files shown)


janicduplessis:

PR #6524, opened 2026-09-13
fix(chat): keep the delivery indicator off grouped media

janicduplessis:
## Summary

Grouped chat posts (no author row) draw the `>>` send-status chevrons over their own content, which for image, video, link and reference posts is the media. This keeps the indicator at the top-right and reserves a 20pt strip above full-width media while the status shows, so the chevrons sit in clear space. Text posts and posts with an author row are unchanged. Applies to the main scroll and threads.

Linear: [TLON-5430](https://linear.app/tlon/issue/TLON-5430/arrows-show-up-on-top-of-video-when-above-a-text-chat-message-from)

```
[avatar] Author           22:04      <- author row: >> at right edge, as today
         [ image post           ]
                              >>     <- grouped media: 20pt strip, >> at right
         [ image post           ]
```

## Changes

- `StaticChatMessage` adds `paddingTop: 20` to the content column when the post has a visible delivery status, no author row, and a first block of `image`, `video`, `link` or `reference`. The indicator is 24pt at `top: 8`; the block wrapper's own 12pt top padding covers the rest.
- **Trade-off:** the strip is transient, so the post shrinks by 20pt and everything above it shifts down when the status clears; for `enqueued` (offline) and `needs_verification` that happens whenever the row reconciles. The alternative is the empty avatar gutter on the left: no shift, but the indicator moves.
- A `MessageStates` specimen in the Cosmos `ChatMessage` fixture for a sent, grouped image post.

The ticket's guess that TLON-5421 (missing blob fields in replies) is involved does not hold: it reproduces with a plain image post in the main scroll.

## How did I test?

Steps: with my own image post as the last message, attach a photo and send it, so two image posts from the same author sit in a row and the second is grouped.

- iOS Simulator (iPhone 17, iOS 26.5), debug build. Before: while the second image is in flight the `>>` renders over its top-right corner. After: it sits in the strip between the two images, overlapping neither; when it clears, the content above shifts down 20pt.
- Android emulator (API 36), `productionDebug`: same before and after.

On a local ship the in-flight window is under a second, so the indicator is visible for a few frames of each clip. Web was not exercised.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

Scope is grouped chat and thread posts with full-width media while their send status is showing, all platforms. Worst case is the layout shift described under Changes.

## Rollback plan

Revert the PR. No data, schema, or settings changes.

### iOS

Second image sent right after my own image.

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/bd6a28a6-4233-4a90-a28c-36d11bbc4746 | https://github.com/user-attachments/assets/01f9be01-f550-4dec-ae12-7f2bc46ec7e1 |

### Android

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/dc819270-f81d-48d2-b3e3-8d90ce45cda7 | https://github.com/user-attachments/assets/7c4c095e-553a-4fe7-9612-d2e05e22413d |


janicduplessis:

PR #6525, opened 2026-09-13
fix(channels): show a single tombstone when a chat message is deleted

janicduplessis:
## Summary

Deleting a message in a group chat channel could leave two "Message deleted" tombstones for one post and raise a duplicate-key error from the message list (`Encountered two children with the same key` on FlatList; `[legend-list] Error: Detected overlapping key` plus a blank gap on LegendList). It happens when the deleted post is already in the channel's loaded pages: a message you sent before you last opened the channel, or any message someone else deletes while you are in it. A message deleted in the same visit it was sent does not show it, which is why it looked intermittent.

The `/v4` `%channels` subscription encodes a deleted post as a **tombstone object**, not `null` ([`channel-json.hoon#L216`](https://github.com/tloncorp/tlon-apps/blob/ce462c16fd11b384327f5eae835ba7cd42f30365/desk/lib/channel-json.hoon#L216)), and `toChannelsUpdate` only checked `set !== null` ([`channelsApi.ts#L275`](https://github.com/tloncorp/tlon-apps/blob/ce462c16fd11b384327f5eae835ba7cd42f30365/packages/api/src/client/channelsApi.ts#L275)), so every delete was classified as `addPost`. The tombstone's `sentAt` is derived from its id, so it differs from the loaded copy's `essay.sent`, and `mergePendingPosts` dedupes by `sentAt`, not id ([`useMergePendingPosts.ts#L107`](https://github.com/tloncorp/tlon-apps/blob/ce462c16fd11b384327f5eae835ba7cd42f30365/packages/shared/src/store/useMergePendingPosts.ts#L107)): both copies stay in the list.

## Changes

- A `%set` whose value is a tombstone decodes to the existing `deletePost` update, for posts and for replies, so it takes the `markPostAsDeleted` / `recomputeChannelLastPost` path instead of `handleAddPost`. This is the path the `null` encoding took before tombstones existed. For replies it also stops `handleAddPost` from bumping the parent's reply count for a tombstone it cannot match by cache id.
- **Remote deletes now show live on a mounted channel.** The `deletePost` case in `handleChannelsUpdate` also calls `deleteFromChannelPosts`, the overlay a local delete already uses; the channel's loaded pages never refetch, so `markPostAsDeleted` alone only showed on remount.
- Tombstones on the changes feed (`/changes`, used on resume and after a discontinuity) go to `deleteFromChannelPosts` instead of `addToChannelPosts`: same path, same `sentAt` mismatch. `insertChanges` still upserts the row.
- `PostResponse.set` and `ReplyResponse.set` include `PostTombstone`, matching what the backend sends.
- `channelsApi.test.ts`: a tombstone `set` for a post and for a reply decode to `deletePost` (both fail on `develop`); a `null` set still does; a live post `set` still decodes to `addPost`.

## How did I test?

iOS simulator (iPhone 17, 26.5) and Android emulator, debug builds against a dev ship, "Show deleted messages" on, in a throwaway solo group's chat channel.

Before, both platforms: send a message, leave the channel, re-enter, long-press it, Delete message, confirm. The red error toast appears (`[legend-list] Error: Detected overlapping key (<post id>)` in `stim logs --errors`) and the tombstone slot renders with a gap. The Android client, sitting in the same channel, also hit the error when the iOS side deleted a post.

After, same steps on both platforms: one tombstone, no gap, no toast, nothing in `stim logs --errors`. An iOS client mounted on the channel while Android deleted a post rendered a single tombstone live.

**Not exercised:** reply deletion from inside a thread; the resume (changes-feed) path; "Show deleted messages" off.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

Scope: decoding of group-channel post and reply delete events on the `/v4` channels subscription, and how those and changes-feed tombstones reach the mounted channel's post listeners, all platforms. DMs and group DMs use `%chat` and are not touched.

**Worst case if wrong:** a delete arriving over the subscription is not reflected until the next sync of that channel's posts. The old path upserted the tombstone as a row (rewriting `sentAt`, nulling `deliveryStatus`, storing `deletedAt`); `markPostAsDeleted` only flips `isDeleted` and clears content. Nothing reads `deletedAt`. A delete for a post the client never synced is a no-op until the next post sync returns the tombstone, as before tombstones existed.

Two adjacent gaps stay as they are: a tombstone returned by a post sync (`/changes`, sequenced pages) is still upserted with `sentAt` set to the id time, so a copy already in `newPosts` can collide with it on a later page fetch; and a remote reply delete does not apply the reply meta the host sends, so the parent's reply count waits for a resync, as it does for a local reply delete.

## Rollback plan

Revert the PR. No schema, migration, or settings change.

## Screenshots / videos

Before (iOS, re-entering the channel and deleting a message that was already loaded): 

https://github.com/user-attachments/assets/32e77b44-d6e5-42b0-a12e-5de7f27f397c

Before (Android, same flow): 

https://github.com/user-attachments/assets/b6f9dfc1-5d46-4c20-999d-6a144535fc29

After (iOS): 

https://github.com/user-attachments/assets/1979633e-effa-4cf5-8c5c-b100c45a4fc3

After (Android): 

https://github.com/user-attachments/assets/f1613822-ce7b-4f31-8ef5-acc7e2a4a7e8


janicduplessis:

PR #6526, opened 2026-09-13
fix(mobile): keep editor content after its web process is terminated

janicduplessis:
## Summary

On iOS, when the OS terminates the rich text editor's web process (typically while the app is backgrounded), `MessageInput` calls `WKWebView.reload()`, which has no page data to reload because the editor is loaded from an HTML string. The editor stays blank until the user leaves and reopens it, at which point the draft comes back. This change remounts the WebView instead of reloading it, so the editor loads again and the existing content restore runs.

**The Android bug as filed in [TLON-5012](https://linear.app/tlon/issue/TLON-5012/android-tapping-settings-from-text-block-editor-clears-post) predates #5688 and no longer reproduces on `develop`:** Android's `reload()` does bring the crashed WebView back, and #5688 re-applies the content. iOS was left broken because the reload itself does nothing there.

## Changes

`reloadWebview` bumps a `key` on `RichText` instead of calling `webviewRef.current.reload()`. Remounting reissues the HTML-string load on both platforms, which is also the recovery Android's `onRenderProcessGone` contract asks for. Everything downstream (`editor-ready` -> `reloadContentRef` -> `setContent`) is unchanged.

## How did I test?

Gallery channel, "New" -> "Text", typed a few words, then killed the editor's web process with the app in the foreground and again with it backgrounded via Home:

- iOS simulator (iPhone 17, iOS 26.5): `kill -9` on `com.apple.WebKit.WebContent`. Before: editor blank, stays blank. After: editor reloads with the typed text.
- Android emulator (API 36): CDP `Page.crash` over the WebView's devtools socket. Before and after: editor reloads with the typed text. After the remount, tapping the editor focuses it and typing works; no keyboard pops up on its own on return.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: message and post editor recovery after a WebView crash (iOS and Android)

Only runs when the WebView reports a crash, terminated content process, load error, or a script error inside the editor. Worst case is what happens today: an empty editor whose draft comes back on reopen. **One tradeoff:** a page that throws on every load already reloads in a loop; it now allocates a new WebView per iteration instead of reusing one.

## Rollback plan

Revert the PR.

## Screenshots / videos

**iOS before** (develop): WebContent killed at ~2s, editor goes blank and stays blank.

https://github.com/user-attachments/assets/5b449257-d881-42fa-a09f-9d6af5dd6e32

**iOS after**: WebContent killed at ~3s, editor reloads with the text.

https://github.com/user-attachments/assets/4051cfde-9685-495b-b7b6-9031c9ac4f5b

**Android before** (develop): renderer crashed at ~10s (keyboard drops), text restored by the existing reload path.

https://github.com/user-attachments/assets/b8c6b220-a811-4648-9f40-77f218345d11

**Android after**: Home, renderer crashed while backgrounded, return: text restored.

https://github.com/user-attachments/assets/f5e50a74-6741-43fc-b10a-5c834d309f2f


janicduplessis:

PR #6528, opened 2026-09-13
fix(chat): keep the delivery indicator off grouped block content

janicduplessis:
## Summary

Fixes [TLON-5430](https://linear.app/tlon/issue/TLON-5430/arrows-show-up-on-top-of-video-when-above-a-text-chat-message-from). When a chat message is grouped under the sender's previous message (no author row), the `>>` delivery indicator was drawn over the top-right corner of its first content block, so for images, videos and other card-shaped blocks it sat on top of the media, in the main scroll and in threads.

## Changes

`StaticChatMessage` positions the delivery indicator absolutely at the top right. With an author row that is the header line; without one it is the content. A grouped post whose first block is not inline text now gets a 20px strip above its content, so the indicator sits in the strip instead of on the block.

**The strip does not depend on the delivery status.** The indicator is transient — it clears on the server echo — so reserving its space only while it shows would move the post the moment it disappeared. The reservation depends on the post's shape alone: a grouped post whose first block is not inline text keeps the strip whether or not an indicator is ever drawn in it. The cost is 20px above grouped media posts that never show an indicator; the benefit is that nothing moves under the reader.

Inline text is an allowlist (`paragraph`, `header`, `list`, `blockquote`, `bigEmoji`), so a new block type gets the strip by default rather than silently regressing onto the media.

The indicator's own rendering is unchanged apart from hoisting the `deliveryStatus && !== 'failed'` test into `visibleDeliveryStatus`, so the same condition drives the element and nothing else.

## How did I test?

React Cosmos, `ChatMessage / MessageStates`, which now carries a **paired specimen**: `Sent (showAuthor=false, image)` and `No status (showAuthor=false, image)` — the same post, differing only in `deliveryStatus`.

Measured in the rendered fixture, card top to image top is **22px in both**. That is the regression this PR is really about: before, the two differed by the strip height, which is what produced the jump on the server echo.

The pair is deliberately left in the fixture rather than captured as a screenshot. The state being proved lasts only from send to echo, so a recording shows it for a few frames and a still shows only one side of the comparison; two specimens side by side stay checkable by anyone who opens Cosmos, and fail visibly if the reservation is ever made conditional again.

Also checked in the same fixture that the author-row cases (`Sent`, `Pending`, `Failed`) are untouched, and that text-first grouped posts still get no strip.

Gates: `pnpm format:check` clean across 1823 files, `tsc --noEmit` clean for `@tloncorp/app`.

## Risks and impact

- Safe to rollback without consulting PR author? Yes — revert restores the overlap.
  - [x] Other: chat message layout. No query, schema or migration changes, so the repo's `staleTime: Infinity` invalidation rules and migration generation are not involved.

The visible change for existing users is 20px of extra space above a grouped post whose first block is media, link, reference, code or file. Posts with an author row, and grouped posts that start with text, render exactly as before.

## Rollback plan

Revert the PR. The strip is additive and self-contained in `StaticChatMessage`.

## Screenshots / videos

Not attached deliberately — see **How did I test?**. The comparison lives in Cosmos as two adjacent specimens, which is more durable than a still of a state that exists for under a second.


janicduplessis:
Thanks. One change of direction on the placement: keep the indicator on the right, but put it above the media rather than moving it into the left gutter. Reserve just enough room above the block so the chevrons never overlap the image, and please check the case of two image posts in a row from the same sender, so the second post's indicator sits between the images without touching either. Any full-width block (image, video, link, reference, and anything else that renders as a card) should get the same treatment. Re-record the two-images case and update the tables.

janicduplessis:
That is the placement this branch already has: the indicator stays at the top right and the grouped post reserves a 20px strip above its content, so the chevrons sit above the block rather than on it. It never moved into the left gutter. Every block type outside the inline-text allowlist (`paragraph`, `header`, `list`, `blockquote`, `bigEmoji`) gets the strip, so image, video, link, reference, code, file, voice memo, table and A2UI cards are all covered, as is anything added later.

Two image posts in a row from the same sender: the previous image ends 12px above the post boundary (the block wrapper's bottom padding), the chevron glyphs occupy 12..28px into the new post, and the next image starts at 32px. So the indicator sits 24px below the first image and 4px above the second, touching neither. Re-recorded on both platforms; the clips are in a third column of each table in the description, and step 4 of "How did I test?" covers it. No code change was needed for this.

janicduplessis:
Closing to redo TLON-5430 from scratch through the mobile workflow, so the fix and its PR are produced by the loop rather than by hand. The branch is kept in case this needs reverting; the redo opens its own PR.

PR #6529, opened 2026-09-13
fix(mobile): keep editor content after its web process is terminated

janicduplessis:
## Summary

On iOS, when the OS terminates the rich text editor's web process (usually while the app is backgrounded), the editor comes back blank until the user leaves and reopens it. This remounts the WebView after such a termination instead of calling `reload()` on it, so the editor loads again and the existing content restore runs. Fixes [TLON-5012](https://linear.app/tlon/issue/TLON-5012/android-tapping-settings-from-text-block-editor-clears-post).

The ticket names Android, which already recovers through `reload()` since #5688. iOS does not: the page is loaded from an HTML string, and for that source `reload` is a plain [`[_webView reload]`](https://github.com/react-native-webview/react-native-webview/blob/f087ea96e71f55c189b666713d9895936afe0948/apple/RNCWebViewImpl.m#L1609-L1623), which leaves WKWebView empty once its content process is gone.

## Changes

`reloadWebview` bumps a `key` on `RichText` instead of calling `webviewRef.current.reload()`. Remounting reissues the HTML-string load on both platforms, which is also what Android's `onRenderProcessGone` contract asks for; everything downstream (`editor-ready` -> `reloadContentRef` -> `setContent`) is unchanged.

## How did I test?

Gallery channel, "New" -> "Text", typed a line, Home, killed the editor's web process while backgrounded, returned. Both platforms, since the crash path is native to each.

- iOS simulator (iPhone 17, iOS 26.5): `kill -9` on the simulator's `com.apple.WebKit.WebContent`. Before: blank, stays blank. After: text is back, typing works.
- Android emulator (API 36): CDP `Page.crash` over the WebView devtools socket. Before and after: text is back, typing works.

Backgrounding alone (Home, or the ticket's keyboard settings activity) does not reclaim the process on a simulator or emulator; the kill stands in for the OS doing it under memory pressure.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: message and post editor recovery after a WebView crash, iOS and Android

Only runs when the WebView reports a crash, terminated content process, load error, or script error. Worst case is today's behavior: an empty editor whose draft comes back on reopen. A page that throws on every load already reloaded in a loop; it now allocates a new WebView per iteration.

## Rollback plan

Revert the PR.

### iOS

Home, WebContent killed while backgrounded, back to the app.

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/44b27470-f2e7-4411-98a8-12bf47d80611 | https://github.com/user-attachments/assets/e25b28f2-7b0d-4033-9cba-e7f10069449b |

### Android

Home, renderer crashed while backgrounded, back to the app.

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/bec683b2-7740-45ab-b80b-a7617e88e66e | https://github.com/user-attachments/assets/140f4f9c-6ae1-4eb8-885e-92648e99a8bd |


janicduplessis:
Closing to redo TLON-5012 from scratch through the mobile workflow, so the fix and its PR are produced by the loop rather than by hand. The branch is kept in case this needs reverting; the redo opens its own PR.

PR #6530, opened 2026-09-13
fix(channels): show one tombstone when a chat message is deleted

janicduplessis:
## Summary

Deleting a message in a group chat channel could leave two "Message deleted" rows for one post and raise a duplicate-key error from the message list. On FlatList that was `Encountered two children with the same key`; on the current LegendList it is `[legend-list] Error: Detected overlapping key (<post id>)` plus a blank gap. It only shows when the deleted post is already in the channel's loaded pages (re-enter the channel before deleting), so the ticket's steps reproduce intermittently.

Root cause: the `/v4` `%channels` subscription sends a deleted post as a `%set` whose value is a tombstone object, not `null` ([`channel-json.hoon#L216`](https://github.com/tloncorp/tlon-apps/blob/ce462c16fd11b384327f5eae835ba7cd42f30365/desk/lib/channel-json.hoon#L216)), and `toChannelsUpdate` only checked `set !== null`, so every delete was decoded as `addPost`. The tombstone then entered the mounted channel's `newPosts` with a `sentAt` derived from its id, and `mergePendingPosts` dedupes by `sentAt`, not id ([`useMergePendingPosts.ts#L107`](https://github.com/tloncorp/tlon-apps/blob/ce462c16fd11b384327f5eae835ba7cd42f30365/packages/shared/src/store/useMergePendingPosts.ts#L107)), so both copies survived.

## Changes

A `%set` whose value is a tombstone now decodes to the existing `deletePost` update, for posts and for replies, so it never enters the pending-post merge. This is the path the `null` encoding took before tombstones existed.

The `deletePost` handler also notifies the deleted-post listeners, the overlay a local delete already applies. The channel's loaded pages are not invalidated by DB writes, so without this a delete from another client would only show after remount; the old path showed it live, with the duplicate.

Not changed: tombstones on the changes feed (`syncLatestChanges`, used on resume) still go to `addToChannelPosts` with the same `sentAt` mismatch. Not on this ticket's repro path; follow-up.

## How did I test?

iOS simulator (iPhone 17, iOS 26.5) and Android emulator (Android 16), debug builds against a dev ship, in a throwaway solo group's chat channel. Both platforms because the ticket reports Android only and the change is shared code.

Before, both platforms: send a message, leave the channel, re-enter it, long-press the message, Delete message, confirm. Observed: a blank gap where the tombstone should be, the red error toast, and `[legend-list] Error: Detected overlapping key (<post id>)` in `stim logs --errors`.

After, same steps on both platforms: one "Message deleted" row, no gap, no toast, `stim logs --errors` clean. `channelsApi.test.ts` covers the decoding: tombstone `set` events for a post and for a reply decode to `deletePost` (both fail on `develop`); `null` and live `set` events decode as before.

Not exercised: deleting a reply from inside a thread, a delete arriving while the app is backgrounded (the changes-feed path above), and "Show deleted messages" off.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

Scope: group-channel post and reply delete events on the `/v4` channels subscription, all platforms. DMs use `%chat` and are not touched. No query keys, table dependencies, or schema changed.

Behavior the old `addPost` path had that `deletePost` does not:

- It upserted the tombstone row, rewriting `sentAt` to the id time and storing `deletedAt`; `markPostAsDeleted` only flips `isDeleted` and clears content, and nothing reads `deletedAt`.
- For a reply it bumped the parent's reply count; that now waits for the next thread fetch, as for a local reply delete.
- A delete for a post the client never synced is a no-op until the next post sync returns the tombstone.

Worst case if wrong: a delete arriving over the subscription is not reflected until the channel is reopened.

## Rollback plan

Revert the PR.

### iOS

Delete an already-loaded message; before shows the gap and the error toast, after shows one tombstone.

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/9ab76af8-f3db-40ea-a633-dd4ff12ae4f4 | https://github.com/user-attachments/assets/ccbf2e4b-c180-4c77-bf78-2c80bdfe074b |

### Android

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/c4b3adbf-ec6a-47d9-864a-b54c44ce9495 | https://github.com/user-attachments/assets/392f0418-ac4e-4714-a66e-e4e90f109f7d |


janicduplessis:
Closing to redo TLON-6178 from scratch through the mobile workflow, so the fix and its PR are produced by the loop rather than by hand. The branch is kept in case this needs reverting; the redo opens its own PR.

PR #6531, opened 2026-09-14
tlon-skill, openclaw: descriptive User-Agent and one bounded 429/503 retry on media fetches (TLON-6362)

patosullivan:
## Summary

`tlon upload <url>` and other guarded media fetches intermittently failed with HTTP 429 because Wikimedia rate-limits by User-Agent, and Bun's default UA put every bot in the same shared bucket. This adds a compliant descriptive User-Agent to guarded/outbound media fetches, a single bounded retry on 429/503 that honors `Retry-After`, and distinct rate-limit/unavailable errors so the model can tell a throttle from a dead URL.

## Changes

- **tlon-skill CLI** (`packages/tlon-skill/scripts/media-guard.ts`, used by `tlon upload <url>`, `posts send --image`, `dms send --image`): sends `TlonBot/<version> (https://tlon.io; support@tlon.io) tlon-cli/<version>` in place of the bare `tlon-cli` UA; retries once on HTTP 429/503, honoring `Retry-After` only when it can be parsed (delta-seconds or an exact IMF-fixdate), capped at 10s and the remaining deadline, with a 5s default when the header is absent; adds fixed-literal errors `FETCH_RATE_LIMITED_ERROR` / `FETCH_UNAVAILABLE_ERROR` for these cases instead of the generic fetch-failed error. Version resolution moved into a new shared `packages/tlon-skill/scripts/version.ts`, also used by `main.ts`.
- **OpenClaw plugin**: adds `tlonMediaUserAgent()` in `src/version.ts`, sent on outbound media fetch (`src/urbit/upload.ts`) and inbound media download (`src/monitor/media.ts`). No retry logic added on this path.
- **Docs**: `SKILL.md` documents the upload UA/retry contract and rate-limit guidance for the model; Hermes `image_search.py` and `tlon_tool.py` tool descriptions updated with matching rate-limit guidance.

## How did I test?

- tlon-skill: full unit suite and typecheck on Bun 1.3.14 (the CI-pinned version; local Bun 1.3.4 fails four pre-existing TLS tests unrelated to this change), oxfmt clean, and a build smoke test confirming the injected version still prints via `--version`.
- OpenClaw plugin: tsc and vitest for the two touched files.
- Hermes: unittest for the two edited modules under Python 3.10.
- Ran a mutation check against `media-guard.ts` covering the throttle branch, the sleep call, retry-reset scoping, the UA string, the wait caps, `Retry-After` parsing, and the no-echo/no-guess invariants; each mutant was caught by at least one test.
- Not tested: an actual Wikimedia 429 in production (per-User-Agent bucketing isn't reproducible from a residential IP).

## Risks and impact

- Behavior change on 429/503 responses: a failing fetch can now take up to ~10s longer before surfacing an error, still within the existing 120s/30s deadlines.
- Two new error strings are now visible to the model in place of the generic fetch-failed message.
- User-Agent changes on all guarded/outbound media fetches, CLI and plugin.
- Deployment: the CLI half only reaches production through an npm publish of `@tloncorp/tlon-skill` (bots resolve `latest` at rebuild). That publish also ships every CLI commit merged since 0.5.0, including `group-action-5`/`/v3/groups`, which requires target ships on desk >= 12.2.0. Publish before promoting to master so the plugin rebuild picks up the new version, and rebuild the Hermes image separately.
  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications

## Rollback plan

Revert the PR.

## Screenshots / videos

N/A


jamesacklin (approved):
Looks good to me

PR #6536, opened 2026-09-14
web: sign the dev server in to the ship with the mobile login variables

janicduplessis:
## Summary

The mobile app signs itself in from `DEFAULT_SHIP_LOGIN_URL` and `DEFAULT_SHIP_LOGIN_ACCESS_CODE`. Web had no equivalent: the dev server proxies to the ship, and the browser meets the ship's login page. Now the same two variables sign both apps in — the dev server logs in to the ship at startup and attaches the session cookie to the browser's requests, so no login page appears and nothing needs typing.

## Changes

A dev-only Vite plugin, `tlon:ship-login`, in `apps/tlon-web/vite.config.mts`. `apply: 'serve'` keeps it out of every build; `vite preview` constructs it but never runs `configureServer`, so it is inert there too. The `mock`, `staging` and `electron` plugin lists don't include it.

**Two prerequisites that turned out not to be obvious.** `loadEnv` reads only `VITE_`-prefixed keys by default — which is why the `SHIP_ACCESS_CODE` many of us have in `.env.local` has never done anything. The login pair is now loaded explicitly, and kept *un*prefixed so the code can never reach the client bundle (`envPrefix` stays `VITE_`/`TAMAGUI_`). And the ship answers a rejected code with **HTTP 400 while still setting an `urbauth` cookie**, so the status is what decides success, not the cookie's presence.

**The cookie makes the server act as the user, so who gets it is the whole design:**

- Only requests from the person's own browser on this machine: the **socket address** must be loopback — a `Host` header is forgeable by any script on the LAN under `--host`, the socket is not — and `Host`, `Origin` and `Sec-Fetch-Site`, when present, must all agree it is same-origin. Vite answers cross-origin requests with `Access-Control-Allow-Origin: *` and checks no `Host` — harmless today because the browser holds no localhost cookie, but with the server supplying one it would let any open website, or anyone on the LAN under `--host`, read the ship and poke as the user.
- Only when the proxy targets the ship the code belongs to. `SHIP_URL` (the e2e ships), `dev2`, or a `VITE_SHIP_URL` that differs all make it no-op with a log line naming both URLs, so a personal `+code` is never POSTed to `~zod`.
- Injection checks the exact cookie name, because another local ship's `urbauth-~bus` on the same host must not suppress it — cookies scope by host, not port.
- A session the ship no longer knows gets **401 on every path, `/~/login` included**, which would lock the browser out as well. On a 401 the cookie is dropped and sign-in repeated once; while that's pending nothing is injected and the ship's own login page works. Only the cookie that earned the 401 is dropped, so a slow response sent with the previous session cannot undo a sign-in that already replaced it.

`DEFAULT_SHIP_LOGIN_URL` stands in for `VITE_SHIP_URL` in the bundle as well as for the proxy target — the app reads it to decide whether the ship is hosted — so a `.env.local` holding just the mobile pair behaves the same as one with both.

`CLAUDE.md` gains the line.

## How did I test?

Against the real dev ship, through the proxy, with a client sending no cookie — a request the ship refuses directly with 403:

| case | plugin | proxied `/~/scry/…` |
| --- | --- | --- |
| correct code | `signed in to …` | **200** |
| wrong code | `the ship rejected DEFAULT_SHIP_LOGIN_ACCESS_CODE; the ship's login page will appear` | 403 |
| unset | *(silent)* | 403 — today's behaviour, unchanged |

Then each guard, same live server:

| probe | result |
| --- | --- |
| own browser (no `Origin`, loopback `Host`) | 200 |
| `Origin: http://evil.example` | 403 |
| `Sec-Fetch-Site: cross-site` | 403 |
| `Host: evil.example` (DNS-rebinding shape) | 403 |
| same-origin fetch (`Origin` matches `Host`) | 200 |
| carrying `urbauth-~bus=…` for another ship | 200 |
| `SHIP_URL=http://localhost:8080` with the code for `pilwes-hosren` | `skipped, the proxy targets … and DEFAULT_SHIP_LOGIN_URL is …` |
| `POST /~/logout` through the proxy, then request again | 401 → re-login → **200** |
| under `--host`: LAN address with a **forged** `Host: localhost:<port>` | 403 |
| `VITE_SHIP_URL` empty, only the mobile pair set | bundle's `import.meta.env.VITE_SHIP_URL` is the ship |
| slow request in flight on session 1, sessions cleared, fast requests re-login, slow one lands with a late 401 | fresh session survives; next request 200 |

The last two rows were reproduced against a stub ship speaking Eyre's login/401 protocol, by a second independent review, and re-run after the fix.

Two independent reviews produced the guards above. The first found the cross-origin exposure, a real regression the draft would have shipped; the second found that a `Host` check alone does not stop a scripted LAN client under `--host`, which my own comment had claimed it did.

`pnpm format:check` and `tsc --noEmit` clean.

## Risks and impact

  - [x] Other: dev-server auth. No app code, no query, no schema — the `staleTime: Infinity` invalidation rules and migration generation are untouched.

The residual risk is the one the design bounds rather than removes: a dev server with the code set is, to the person's own browser on that machine, already signed in to their ship. That is the feature, and it is the same trust the mobile debug build extends. It is not extended to other origins, other hosts, other ships, or builds.

Not addressed: `SHIP_ACCESS_CODE` stays inert. It is in people's `.env.local`, not the repo, so there is nothing here to remove; the new name is the one that works.

## Rollback plan

Revert the commit. Anyone without the variable set sees no difference before or after.

## Screenshots / videos

None — there is no UI in this change; the evidence is the table above.


janicduplessis:
Addressed the Codex review on `2b64472f6e`:

- **P1 — verify client address before injecting.** Already handled in `e0454ae29c` before this comment landed: the gate now requires `req.socket.remoteAddress` to be loopback, since a `Host` header is forgeable under `--host`. Verified: LAN address with `Host: localhost:<port>` → 403, loopback → 200.
- **P2 — normalize target before the login URL.** Fixed. The login now POSTs to the trailing-slash-stripped base, so a `DEFAULT_SHIP_LOGIN_URL` ending in `/` hits `/~/login`, not `//~/login`. Verified against a stub with both URLs trailing-slashed.
- **P2 — replace a stale browser cookie.** Fixed. Injection now drops any same-name `urbauth-*` cookie the browser holds and substitutes the dev-server session, rather than suppressing injection when one is present. A request carrying a dead `urbauth-~zod` now authenticates (was 401).
- **P2 — honor the proxy TLS setting.** Fixed. The login request runs through `node:https` with `rejectUnauthorized: false`, matching the proxy's `secure: false`, so a self-signed dev ship the proxy can reach can also be signed in to. (First attempt used an undici `Agent` dispatcher; that broke the working real-ship path because the hoisted undici's interceptor API doesn't match the one built-in `fetch` uses, so the login moved to node's own http/https modules.)

All four verified live — the real hosted ship still signs in and proxies at 200, and the stub-ship cases each pass.

PR #6542, opened 2026-09-15
fix(chat): keep the delivery status indicator off a grouped message's media

janicduplessis:
## Summary

[TLON-5430](https://linear.app/tlon/issue/TLON-5430). The delivery-status chevrons are absolutely positioned at the top right of a chat post. A post grouped under the sender's previous message draws no author row, so the chevrons land on the first content block instead of beside a name — on the photo for an image, on the **Copy** button of a code block, on the open-in-new arrow of a file card. Same in the main scroll and in threads.

This reserves room above the content when a grouped post's first rendered block is full width.

```
before                    after
┌──────────────────┐      ┌──────────────────┐
│ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ›› │      │               ›› │
│ ▓▓▓ image ▓▓▓▓▓▓ │      │ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ │
│ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ │      │ ▓▓▓ image ▓▓▓▓▓▓ │
└──────────────────┘      └──────────────────┘
```

**One thing to decide:** `MessageContainer` — the Android/web action-sheet preview — renders its own author row above a `StaticChatMessage` that gets no `showAuthor`, so previews of full-width posts pick up the gap as well. The indicator genuinely would overlap there, so it isn't wrong, but that preview is height-capped with `overflow: hidden` and loses ~22px of visible content. Happy to special-case it if you'd rather.

## Changes

- **Shape, not status.** The gap is keyed to the post being grouped with a full-width first block, never to `post.deliveryStatus` — that clears on the server echo a moment after a send, so a gap depending on it would make the post jump as it lands. The trade is that a grouped media post carries the gap even with no indicator showing.
- **Text blocks keep today's layout; everything else reserves.** `paragraph`, `header`, `list`, `blockquote` and `bigEmoji` are excluded by name; image, video, link preview, reference, code, file, voice memo and table all reserve. A block type added later reserves by default, which is the safer way to be wrong.
- **Two Cosmos fixtures.** `GroupedDeliveryStatus` renders seven post shapes twice side by side, with a `pending` status and with none, so the no-jump property is checkable by eye. `GroupedImageRun` is the two-images-in-a-row case. Both are eyeball-only — nothing automated pins this, and nothing ties the `$2xl` padding to the indicator's hardcoded 24px height and 8px inset.

## How did I test?

iOS simulator (iPhone 17, 26.5) and web desktop at 1280px, against a self-hosted dev ship. Sent two images back to back in a throwaway group and watched the second one — the grouped post — through the send window, on both. Not tested on Android: the change is a tamagui `paddingTop` in shared code with no platform branch, and both token scales are covered by iOS and web.

Verified in the DOM rather than from the screenshots, in the real web app and in Cosmos: the first block's top moves from 10px to 32px inside the post, exactly where the indicator's box ends, and every pair in `GroupedDeliveryStatus` has identical post heights with and without a status.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

Layout only, one `paddingTop` on one container, no data or query changes.

The web renderers cap media at 600px (`CHAT_REF_LIKE_MAX_WIDTH`, `WebChatImageRenderer`), so in a wide desktop column a block often never reaches the chevrons and there was nothing to overlap — the gap is added there anyway. Same for a small image on any platform. Conditioning on measured width would reintroduce the jump, so I didn't.

## Rollback plan

Revert the PR.

### iOS

Sending a second image right under the first. The indicator is on screen for about half a second; the stills are frames from these clips, at the moment it shows.

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/18f155c2-e12b-42a8-9bc0-0bd9ccde17c3 | https://github.com/user-attachments/assets/cfe4a051-a94c-4611-9011-8dd49cca33a8 |

The "before" frame, zoomed on the second image's top-right corner — a chevron is drawn over the photo and clipped by its rounded corner:

### Cosmos — `ChatMessage / GroupedDeliveryStatus`

Seven post shapes, each rendered twice: left half has a `pending` status, right half has none. The two halves line up in both columns, which is the no-jump property.

| Before | After |
|---|---|
|  |  |

### Cosmos — `ChatMessage / GroupedImageRun`

Two images in a row from the same sender; the second post's indicator has to sit between them, touching neither.

| Before | After |
|---|---|
|  |  |

### Web desktop

The same three posts in the running app at 1280px — the grouped ones gain the reserved row.

| Before | After |
|---|---|
|  |  |


dnbrwstr:
I agree this is annoying, but don't love adding extra space between messages. Could we do something like Signal here? White sending indicator, slight gradient underneath.

2026 9 15 12 00 55 59db 22d0 e560 4189-269E03A6-2A48-4739-93F5-08140FFDB370

janicduplessis:
I agree, was mostly testing the new pr workflow. In our case since the loading indicator only appear briefly should we have that gradient only visible which it is there or all the time? Or should we generally rethink the design of it?

PR #6543, opened 2026-09-15
fix(chat): duplicate "Message deleted" tombstone after deleting a loaded message

janicduplessis:
## Summary

[TLON-6178](https://linear.app/tlon/issue/TLON-6178) — deleting a chat message that was already
loaded into the list renders a second "Message deleted" tombstone and throws React's
`Encountered two children with the same key`. The key is the deleted post's own id.

Not a sentinel id, as the ticket guessed: a tombstone's `sentAt` differs from the live post's, and
`mergePendingPosts` dedupes on `sentAt`, so the same post gets woven into the list twice.

**Known gap:** this does not cover deleting your own message inside the `markPostSent` catch-up
window, where the still-pending row and the tombstone carry different ids. See Risks.

<details>
<summary>Where the two <code>sentAt</code> values come from</summary>

`toPostData` builds a tombstone's `sentAt` from the host-assigned id
([postsApi.ts#L1288](https://github.com/tloncorp/tlon-apps/blob/c7f3a97a9bb34fe024d9b20f56bfb604aeaf7347/packages/api/src/client/postsApi.ts#L1288));
a live post uses `essay.sent`
([L1365](https://github.com/tloncorp/tlon-apps/blob/c7f3a97a9bb34fe024d9b20f56bfb604aeaf7347/packages/api/src/client/postsApi.ts#L1365)).
When the tombstone arrives through the live listener while the original is still in the paginated
query page, the two rows disagree. Logging the merge output on device: same id, `sentAt`
1789431942733 vs 1789431916682.

</details>

## Changes

Two changes in `mergePendingPosts`, both needed:

- **Dedupe new and pending posts against the ids already in the page**, so a server-backed post
  can't be woven in a second time when its `sentAt` moves. `sentAt` stays the dedupe key for
  everything else — it's the only thing that can match an optimistic row (local id) to its
  confirmed echo (canonical id).
- **Propagate the deleted overlay by id**, not just by the `channelId:sentAt` merge key. Without
  this, dropping the duplicate would leave the surviving row rendering as a normal message when
  someone *else* deletes it: the tombstone was the only carrier of the deletion in that render, and
  the merge key can't match it for the same reason `sentAt` couldn't.

## How did I test?

Android (Pixel emulator, `productionDebug`) and web (desktop navigation, 1440x900), in throwaway
groups. Both platforms, because this is shared code under `packages/shared`. No Cosmos specimen:
the change is in a pure merge function, and the fixtures render single messages, not a merged list.

Repro on both: post a message, leave and re-enter the channel so the post is served from the
paginated query page rather than the live listener, then delete it. Deleting a message that has
only ever been in the live page does **not** reproduce — `mergePendingPosts` early-returns before
the pagination merge when the page is empty, which is probably why this looked intermittent.

- **Android** — before: `stim logs --errors` shows `[legend-list] Error: Detected overlapping key
  (170.141.184.508.161.387.297.604.387.955.198.132.224)`, a red error toast, and the "Today"
  divider disappears as legend-list drops the collided row. After: one tombstone, divider intact,
  no error records.
- **Web** — counted tombstones and console errors around a single delete. Before: 2 → 4 tombstones
  for one deletion, with three `Encountered two children with the same key` errors. After: 3 → 4,
  zero. (The delete was driven from the phone so the web client exercised the sync-delivered
  tombstone path; its own action menu doesn't respond to synthetic clicks.)
- Two regression tests in `mergePendingPosts.test.ts` cover the local and remote delete shapes;
  both fail on `develop`.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [x] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [ ] Other:

This sits in the merge every channel view runs on every render, so the blast radius is wide even
though the change is small. The risk is dropping a row that should render: the new filter only
bites when an incoming post shares an id with one already in the page, and an edit preserves
`essay.sent`, so in practice the tombstone is the only shape that reaches it. Optimistic rows are
unaffected — their ids are locally generated and never match a canonical one. No React Query key,
dependency declaration, or schema touched.

**Not fixed here:** deleting your own message inside the `markPostSent` catch-up window, where the
still-pending row (local id) and the server tombstone (canonical id) can both render as tombstones.
Different ids, so no key error, and I haven't reproduced it — noting it rather than widening this
change.

## Rollback plan

Revert the PR.

### Android

Delete an already-loaded message. Before: the tombstone appears, then the "Today" divider vanishes
and the duplicate-key error toast comes up. After: one tombstone, divider stays, no toast.

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/49df1bde-0955-433f-862f-3bc1518f24b5 | https://github.com/user-attachments/assets/78e45b17-3bc2-408f-8c3e-361996eb9619 |

### Web

Same repro, counted on the page. Before has four tombstones for three deleted messages; after has
four for four.

| Before | After |
|---|---|
|  |  |


patosullivan (approved):
Lgtm!

PR #6547, opened 2026-09-15
Sentry triage batch 3: baseline migration replay, notes watermark invariant, stale sync discard

patosullivan:
## Summary

Third Sentry triage batch under TLON-6478. Three designed or benign paths were reaching Sentry as errors: the first migration attempt failing on native before the designed purge and retry (REACT-NATIVE-F), the sequence watermark invariant firing on every notes channel (REACT-NATIVE-15R), and `syncSince` discarding stale fetched changes (JAVASCRIPT-REACT-1S9, REACT-NATIVE-XM). Each now reports to PostHog only, with the surrounding behaviour unchanged.

## Changes

**`packages/app/lib/nativeDb.ts`** (REACT-NATIVE-F, 3,270 events / 21 users since November)

- The client ships one regenerated baseline migration, so after any schema change drizzle replays it against a populated database, op-sqlite throws `table … already exists`, and `runMigrations` purges and retries by design (PostHog, 90 days: 876 purges, 864 successful retries, 0 retry failures). That catch logged at Critical severity, which `compositeLogger` forwards to Sentry.
- The outcome of the initial attempt never affects control flow: every failure purges and retries, and the purge and retry-migrate catches already report Critical and rethrow. The initial catch therefore now logs `NativeDbDebug` at Low severity with the same diagnostic payload, unconditionally. A recovered first attempt is counted in PostHog; a genuine failure still reaches Sentry through the retry. This matches what `webDb.runMigrations` already does.
- Purge and retry behaviour is untouched.

**`packages/shared/src/store/useChannelPosts/useChannelPosts.ts`** (REACT-NATIVE-15R)

- `hasNewerPosts` asserted a sequence-number watermark for every channel, but third-party channels (notes) have no sequence numbers and `getChannelPosts` returns a null watermark for them, so every notes channel fired the invariant as `app_error`.
- It now returns `false` for third-party channels before reading the watermark. Its only caller feeds `canFetchNewerPosts`, and `true` claimed there might be sequenced newer posts to page toward, which can never hold for these channels. DMs, clubs and chat/diary/heap channels are unaffected. The function is exported for its test.

**`packages/shared/src/store/sync/sync.ts`** (JAVASCRIPT-REACT-1S9, REACT-NATIVE-XM)

- `syncLatestChanges` discards fetched changes when the fetch outlived its freshness threshold (a suspension, or a slow or wedged request) by throwing; `syncSince`'s catch reported that as `sync since failed` / `app_error`, which is forwarded to Sentry.
- The throw is now a typed `StaleSyncDataError` carrying `runningForMs`, and the catch reports it as a plain `sync since discarded stale data` event (PostHog only). Message text, threshold, the discard, `result = 'error'` and watermark handling are unchanged.

**Tests**

- `packages/app/lib/nativeDb.test.ts` confirms an initial failure logs Low and still purges and retries, and that a retry-attempt failure keeps the Critical report; new `staleSyncData.test.ts` and `useChannelPosts.test.ts` cover the stale-sync classification and the notes early return.

## How did I test?

- `pnpm format:check`, `pnpm -r tsc` and package lint all pass.
- vitest for `packages/app` and `packages/shared`.
- Mutation checks on every guard (initial-catch severity, third-party early return, `StaleSyncDataError` branch) confirm the tests fail when the guard is removed.

## Risks and impact

  - [ ] Onboarding
  - [x] State / providers
  - [x] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [x] Other: native DB migration reporting (`packages/app/lib/nativeDb.ts`)

The only behaviour change beyond reporting is `hasNewerPosts` returning `false` instead of `true` for third-party channels, which stops a newer-page pagination flag that could never have been satisfied for them.

## Rollback plan

Revert the commit. No schema, data, or API change.

## Screenshots / videos

N/A


dnbrwstr:
**nativeDb: suggest replacing the collision classifier with a severity rule**

Observation: the outcome of the initial migrate attempt does not affect control flow. Any error leads to purge and retry, and the retry path already reports Critical on purge failure and on retry migrate failure. A genuine failure therefore still reaches Sentry via the retry.

The classifier only changes reporting for one case: an initial attempt that fails for a non-collision reason and then succeeds on retry. That is a recovered state and can be counted in PostHog rather than paged.

Proposed change:

- In the initial catch block, always emit `NativeDbDebug` at `AnalyticsSeverity.Low` with `errorMessage` attached.
- Remove `readRecordedMigrationWhen`, `BUNDLED_MIGRATION_WHEN`, `EXPECTED_BASELINE_REPLAY_COLLISION`, and `isExpectedBaselineReplay`.
- Reduce `nativeDb.test.ts` to: initial failure logs Low, retry failure logs Critical.
- For visibility on non-collision initial failures (5s timeout, partial DB), add a PostHog insight filtered on `errorMessage` not containing `already exists`.

Additional notes:

- The regex depends on an exact op-sqlite error prefix from `cpp/bridge.cpp` and on drizzle and the transaction wrapper not rewrapping the error. This is a fragile dependency on third-party internals.
- `webDb.runMigrations` already implements the proposed behaviour: any initial failure is logged, the DB is purged, and migrate runs again. It does not report the initial failure, which is why this Sentry issue only has a native fingerprint.
- The journal predicate is not portable to web. `webMigrator` tracks applied migrations by tag, not `created_at`, so the platforms would classify the same event using different evidence. The severity rule keeps them aligned, and web can adopt the same two events later.

dnbrwstr (approved):
I think we can simplify the nativeDb error handling (see above bot comment) but otherwise lgtm.

patosullivan:
Agreed, taking this. The initial catch will log `NativeDbDebug` at Low unconditionally; the collision regex, the journal read and `isExpectedBaselineReplay` go away, and the tests reduce to initial-failure-logs-Low / retry-failure-logs-Critical. Push coming on this PR.

patosullivan:
Merged with the severity rule. The insight you suggested is up: [Native DB: initial migrate failures that are not baseline collisions](https://eu.posthog.com/project/7233/insights/KPx9fJBg) (Native DB Debug, initial-attempt context, errorMessage not containing "already exists", broken down by message). It stays empty until the first build with this change ships.

PR #6583, opened 2026-09-21
fix(chat): keep a huge image from taking over the conversation on native

jamesacklin:
## Summary

[TLON-6526](https://linear.app/tlon/issue/TLON-6526/chat-cannot-scroll-to-bottom-when-a-huge-image-is-inserted). On native, a chat image is laid out at the message column width times whatever aspect ratio the post declares, with no cap — so an image block carrying large dimensions becomes a conversation row several viewports tall. A 1200×9000 block lands at roughly 2,600pt on a phone. This caps it: **7 viewports of scrollable content → 4** on the test conversation.

Two things to know before reviewing:

- **I could not reproduce the trapped scroll in the ticket's recording on develop.** The reporter was on `ios-production-720` (cut 2026-08-20) and two production builds have shipped since. I tried four shapes, including the exact one in the clip; the bottom was reachable every time. What I could verify is the unbounded layout the ticket's *Expected behavior* asks to fix, so that is what this addresses. If it recurs on a current build it needs the ship and channel, or a fresh recording, plus a separate look at the list.
- **Only reachable from outside the app.** Our uploader resizes every attachment to a [1200px long edge](https://github.com/tloncorp/tlon-apps/blob/0baa9b2ef793b19d3a581b70add887850e651373/packages/shared/src/store/storage/storageActions.ts#L232-L238), so app-sent images top out around 1200pt of row height — a 1290×18000 PNG arrived as 86×1200. Rows big enough to matter need a block from another client, a bot, or a remote URL.

## Changes

`ImageBlock` takes an optional `maxWindowHeightFraction`. When set it measures the column it is in and renders the largest box with the image's own proportions that fits both that column and the cap, never wider in points than the image is in pixels. Chat passes 0.85 of the window; notebooks, galleries, activity and link previews pass nothing and keep today's behavior exactly.

This reuses the sizing `constrainedSize` already did rather than adding a second path — it computed `min(maxWidth, maxHeight × ratio)` and only ever needed real limits instead of fixed ones, so web chat now passes its pixel caps and native chat passes its measured column and window-derived height into the same calculation.

**Measuring is gated on whether it can matter.** A column is never wider than the window and an image is never rendered wider than its own pixels, so `min(window, pixels) / ratio` is the tallest a block could ever be. When even that fits the cap, no column can make it overflow and the block renders on the plain path. Ordinary photos therefore keep exactly the pre-change render path — no probe, no extra layout pass, no state update — and only the unusually tall images the cap exists for pay for any of it.

Two things it deliberately does not do, both of which I tried first and backed out of:

- **It does not cap by aspect ratio.** The height a block takes is its rendered width over its ratio, and rendered width is already bounded below the column by the pre-existing `maxWidth: dimensions.width`. A ratio floor crops a 120×400 image that was never too tall.
- **It does not crop to the cap.** Nothing caps image width on native, so a portrait photo across a 1000pt tablet column stands 1333pt tall — cropping it to fit would take a third of the picture, where narrowing it to 652pt keeps all of it.

## How did I test?

iOS 26.5 simulator against a self-hosted dev ship. iOS only: the new prop is passed solely by the native chat renderer, so web and desktop take the identical path they do today — I ran the `ChatMessage` Cosmos fixture to confirm the shared component still renders there. To author the fixture I temporarily bypassed the upload resize above, then reverted it; both recordings run unmodified code.

- scrollable content **7 viewports → 4**, off the scroll indicator's page count
- an 86×1200 image renders **53pt wide** — exactly `min(330 column, 86 pixels, 743 cap × 0.0717)`, complete rather than cropped
- a phone screenshot still renders at **full column width**, uncropped
- tapping a fitted image still opens it full-size in the viewer
- `stim logs --errors` clean

`imageSizing.test.ts` is new: representative image and device shapes rather than a cross product, plus the contract the measurement gate rests on — anything it skips must be something the fit would have left alone.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [ ] Notifications
  - [x] Other: shared `ImageBlock`, used by every post-content surface

**The one thing worth checking in review:** `ImageBlock` is shared, so the blast radius is every post-content surface, and what keeps the others byte-identical is that the prop defaults to undefined and only the native chat renderer sets it. That is the whole containment argument.

The visible trade-off is that a very tall image now renders narrow rather than tall — a 1200×9000 block comes out about 99pt wide on a phone, legible only by tapping through, with no affordance saying so. That is the same bargain web has always made, and it beats both alternatives (develop leaves it unbounded; cropping would take 60% of a screenshot on a tablet column), but it is a real change for anyone who posts a long screenshot. Notebooks have the same unbounded behavior and are deliberately untouched.

Not covered by tests: the React Native layout and `GestureTrigger` wiring — that the probe is inserted only when the gate says so, that its measured width reaches the calculation, and that the viewer's trigger node still wraps the image. Typecheck and the unit tests say nothing about any of it, so it is worth an eye.

No React Query or schema changes.

## Rollback plan

Revert the PR.

### iOS

Scrolling the same conversation from the newest message to the top. Before, the tall image takes several full screens; after, every image fits inside one.

| Before | After |
|---|---|
| https://github.com/user-attachments/assets/35330768-ae5b-483d-8851-c1b3f2d33815 | https://github.com/user-attachments/assets/738cb602-9a6e-4a04-a840-ce0f3dc7eff7 |


packages/app/ui/components/PostContent/BlockRenderer.tsx
```diff
@@ -723,13 +777,37 @@ export function ImageBlock({
...
+      {triggered}
+    </RNView>
+  );
 }
 
+const styles = StyleSheet.create({
+  // Full width so the layout reports the column, and flex-start so the trigger
+  // node inside shrink-wraps the image -- the viewer zooms from that node, and
+  // a stretched one hands it a rect wider than the picture being tapped.
+  measureAvailableWidth: { width: '100%', alignItems: 'flex-start' },
```
jamesacklin: `alignItems: flex-start` is doing real work here, not styling. `GestureTrigger` registers its own wrapping `View` as the node the media viewer zooms from, and that View is a stretch child of this probe — so without this it would be the full column width while the picture the user tapped is only as wide as the fitted box, and the open transition would start from a rect up to 3.3x too wide. The probe keeps `width: 100%`, so the column measurement is unaffected.

Related: while writing this I had the probe *inside* `GestureTrigger` for a while. That is silently broken — `GestureTrigger` does `cloneElement(child, { onPress })` and only registers the node from that injected handler, so handing it a plain `RNView` (no `onPress`) drops the prop, the node never registers, and both the open zoom and swipe-to-dismiss die with no warning, because the dev-mode "No onPress found" log lives inside the handler that never runs.

jamesacklin:
Requesting @janicduplessis and @dnbrwstr — Janic for `BlockRenderer.tsx`, where the change lives, and Dan for `StaticChatMessage.tsx` and the conversation geometry around it.

Two things I would most like a second opinion on, both called out in the description:

1. **The containment argument.** `ImageBlock` is shared by every post-content surface, and what keeps notebooks, galleries, activity and link previews byte-identical is that `maxWindowHeightFraction` defaults to undefined and only the native chat renderer sets it. That is the whole blast-radius story, and it is worth an independent check rather than my word.
2. **The trade-off.** A very tall image now renders narrow instead of tall — a 1200x9000 block is about 99pt wide on a phone, readable only by tapping through to the viewer, with nothing on screen saying so. It beats the alternatives I tried (develop leaves it unbounded; cropping takes 60% of a screenshot on a tablet column), but whether that is the right product call is yours, not mine.

Also worth knowing: I could not reproduce the trapped scroll from the ticket on develop, so this fixes the unbounded layout the ticket asks for rather than a reproduced scroll trap. Details in the description.

dnbrwstr (approved):
Approved. The product direction is sound, and measuring the real column is justified if the requirements are no cropping, no letterboxing, and correct media-viewer transition geometry.

Non-blocking: this still feels overengineered in proportion to the issue. The core rule is `width = min(column width, natural width, height cap × aspect ratio)`, but the PR adds 365 lines, including 189 lines of tests that mostly reassert properties of that calculation while the riskier React Native layout and `GestureTrigger` wiring remains untested. Every native chat image also takes the extra wrapper, layout measurement, state update, and rerender, even though only unusually tall externally-authored images need it.

I would prefer a follow-up simplification: keep the chat-only cap and real-column measurement, fold the fit calculation into the existing constrained-size path, only measure when the image could exceed the cap, and replace the cross-product suite with a few representative cases plus coverage of the component wiring.

Also worth keeping explicit: the reported scroll trap was not reproduced on develop, so this fixes the unbounded image layout rather than proving the original failure is fixed.

jamesacklin:
Thanks @dnbrwstr — taking the follow-up framing, and agreeing with most of it.

**The per-image cost is a fair hit and the cheapest to retire.** Every native chat image currently takes the probe, a layout pass and a state update even though almost none of them need it. It is fixable without measuring first: the widest an image can ever render is `min(window width, natural width)`, and the column is never wider than the window — so when `min(windowWidth, naturalWidth) / aspect <= cap` the image provably cannot exceed the cap at any column, and the probe can be skipped outright. That drops ordinary photos back to today's render path entirely, and takes the opacity-0 first frame with them.

**The test balance is wrong, and I would weight it the other way round.** The cross-product does mostly reassert arithmetic; the `GestureTrigger` wiring is exactly where the one real bug in this change lived (probe inside the trigger, `cloneElement` dropping `onPress` onto a plain `RNView`, viewer transitions silently dead), and 125 green assertions said nothing about it. Representative cases plus a render test asserting the wiring is the better trade.

One factual note on the size, not a defence of it: of the 365 added lines, 189 are tests and 43 are comments/docs, so the component delta is 83 lines in `BlockRenderer.tsx` and 71 in the helper. Still more than the one-line rule deserves, and folding the fit into the existing `constrainedSize` path is the right way to collapse it — that path already computes `min(maxW, maxH × aspect)` and only ever needed a real column width instead of a fixed one.

Happy either way: I can do the simplification in this PR now, or land this and follow up. Say which you prefer and I will not merge until then.

And yes — keeping the scope caveat explicit. It is in the description and on the Linear issue: the reported trap was never reproduced on develop, so this fixes the unbounded layout rather than proving the original failure is gone.

jamesacklin:
Done in `cd48f8c` — all three, @dnbrwstr.

**Folded into `constrainedSize`.** The separate fitted-size branch is gone; both callers share the one calculation and differ only in where the limits come from. Only the measured path holds an image to its own pixels, which is what the fixed-cap callers have never done, so web chat behaviour is byte-identical.

**Measuring is gated.** `min(window, pixels) / ratio` is the tallest a block could ever be, so when that already fits the cap no column can make it overflow and the block takes the plain path. Ordinary photos get no probe, no extra layout pass, no state update and no invisible first frame — exactly the pre-change path. Only the tall images the cap exists for pay anything.

**Tests: 125 cases → 18.** Representative shapes instead of the cross product, plus the contract the gate rests on (anything it skips must be something the fit would have left alone) — that one is worth having, since the gate is now load-bearing for correctness, not just cost.

Net effect on size: 365 → 350 added lines, and one sizing concept instead of two.

Where I did **not** follow you: component wiring is still untested. The `packages/app` vitest setup mocks `react-native` down to five exports, so a render test of `ImageBlock` is its own piece of infrastructure rather than a few cases, and I did not want to smuggle that in here. It is called out in the description as the known gap — and it is the honest one, since the only real bug in this change lived exactly there.

Behaviour on device is unchanged: still 4 viewports, the 86×1200 image still renders 53pt wide, so the attached recordings still stand.

PR #6585, opened 2026-09-21
feat(mobile): bottom tab bar, revised Settings, and onboarding in the bot DM

jamesacklin:
## Summary

Merges the first four `next-app` PRs onto `develop`: the bottom tab bar (#6559), nav position surviving a cold start (#6566), the activity tab (#6567), and the revised Settings tab (#6570). It is a merge of `next-app` at `a5c955e9`, not a re-implementation — the 78 commits and their review history come across intact, and both merges (next-app in, then current `develop` in) were textually clean.

**Two things to decide before merging.**

1. `next-app` has already moved past this state — #6576–#6579 replace this bottom tab bar with a slide-out side drawer. Those four are excluded here, so merging this commits `develop` to a navigation shape the drawer work intends to replace. That is deliberate: it ships the tab bar as reviewed and lets the drawer land on its own.
2. **This breaks the Maestro suite** (details under Risks). It is not covered by this PR's CI, so nothing here will go red — `mobile-testing.yml` triggers on `push` to `develop`, so it fails *after* merge.

## Changes

- **Navigation.** A bottom tab bar carries Workspaces, Activity and Settings, plus a Bot tab that only mounts when `botDm.enabled`. The activity bell moves out of the Workspaces header into the bar, and unread badging moves into `activityBadges.ts` so one message doesn't light several indicators at once.
- **Settings.** Rows show their current value, so Appearance reads "Auto" without a tap; both labels come from the screen the row opens, so they can't disagree. The bot's own controls (identity, permissions, providers) are surfaced on the tab, and the Activity row is gone since the tab bar carries it.
- **Onboarding.** First run happens in the bot DM rather than in a generated group, and the plugin tracks where onboarding stands per-DM so finished conversations stop paying for history reads.
- **Agent tooling.** `mobile-login.mjs` gated on the literal `Home`, so after the rename its already-signed-in short-circuit never fired and a *successful* login blocked 60s before reporting failure. It now matches the chat list by label — `text="Workspaces"` hits both a navigation-bar and a text node, and an exact match on two elements is an `AMBIGUOUS_MATCH` rather than a hit.

## How did I test?

Built and ran the merged branch on an iPhone 17 simulator (iOS 26) against a self-hosted dev ship. Signed in through the self-hosted path, switched between all three tabs, and opened Settings to confirm the revised rows and their value labels. The Bot tab correctly does not appear on a ship with no agent. The fixed `mobile-login.mjs` was verified end-to-end against the signed-in app.

The only error logged while exercising the tabs was `Query data cannot be undefined … ["baseVolumeLevel",{}]`, from `useBaseVolumeLevel` — untouched by this merge, already called from five places, and it falls back to `'medium'` by design.

Web renders clean signed-in at 1440×900; its three console warnings are identical on `develop`, checked on a second server.

**Android has now been exercised** on a physical Pixel 7a (Android 17), on `productionDebug` against a hosted ship with a bot. The tab bar, all four tabs, drill-down and back, `backBehavior="history"`, the cold-start position restore (#6566) and the revised Settings value labels all pass; `stim logs --errors` is clean. The `baseVolumeLevel` warning above reproduces identically there, which supports reading it as pre-existing. Evidence and the full checklist are in a comment below.

That pass turned up one defect, fixed here in `02b8b98` and `ddb729e`: the Bot tab's header blanked its title, so the bot DM showed an avatar and caret with nothing to say whose conversation it is. `ChannelHeader.tsx` is now back to its state on develop — the `isBotDm` flag and the `isTopLevelTab` prop it was the only reader of are gone.

The flag suppressed the title *and* the subtitle, but only the title suppression ever did anything. It required `isWindowNarrow`, while `ScreenHeader` renders a subtitle only under `useHorizontalTitleLayout`, which `ChannelHeader` sets to `!isWindowNarrow`. Those are mutually exclusive, so the subtitle half was unreachable on a narrow layout and the flag was already false on a wide one. Mobile channel headers render no subtitle either way; changing that would touch every channel header, not just the bot's, and is not attempted here.

Before/after evidence for each feature lives on the source PRs (#6559, #6566, #6567, #6570). The captures below only show that the features still render after merging onto current `develop`; they are cropped because this repo is public and the dev ship holds real conversations.

## Risks and impact

  - [x] Onboarding
  - [x] State / providers
  - [ ] Message sync
  - [x] Channel display
  - [x] Notifications
  - [ ] Other: mobile E2E (Maestro)

**The Maestro suite breaks on merge, and it was the mobile-navigation regression net.** Renaming the `Home` tab to `Workspaces` and moving `Contacts` off the tab bar invalidates **30 `.maestro` files**. `mobile-testing.yml` runs three flows that `assertVisible: 'Home'` immediately after login; `mobile-reliability-nightly.yml` routes all 12 reliability flows through `subflows/begin.yaml`, which waits on `^(Home|Contacts)$` and then taps a Contacts tab that no longer exists — on iOS by a positional tap calibrated for a 3-tab bar. `03-signup.yaml` breaks twice over: a hosted signup now lands on the bot DM tab, whose header is deliberately blank.

I did **not** fix these. The literal rename is mechanical, but the Contacts-tab removal, the positional taps and the signup landing are structural, and I can't run Maestro Cloud locally to verify a blind fix — 30 files of unverifiable churn that still wouldn't go green seemed worse than saying so plainly. Post-merge there is effectively no automated coverage of mobile navigation until they're updated.

**Blast radius is large.** This replaces the app's top-level navigation and touches shared state hooks, so almost every mobile screen is downstream. The compensating factor: all of it already went through review on `next-app`; what is new is its interaction with the 122 commits `develop` gained since.

**Two new React Query readers** land in `dbHooks.ts` — `['channelUnread', depsKey, channelId]` and `['channelThreadUnreads', threadDepsKey, channelId]`. Both put the `useKeyFromQueryDeps` set at key index 1, which is where the invalidation predicate looks; a key that puts it elsewhere silently never refreshes again. No schema change, so no migration.

**`packages/openclaw` needs a version bump before release.** develop bumped it to `0.30.0`; this branch then adds ~1,100 lines of monitor behavior under that same version, and `openclaw-publish.yml` verifies tag-vs-version. Hosted ships are unaffected (`bot-harness-deploy.yml` pulls the branch ref, not the npm version).

## Rollback plan

Revert the PR. No migration runs and no setting is written, so nothing persists past a revert.

### iOS

The merged branch on iOS: the revised Settings tab, and the tab bar cropped to its chrome.

| Settings tab | Tab bar chrome |
|---|---|
|  |  |


packages/app/navigation/topLevelTabs.ts
```diff
@@ -5,17 +5,21 @@ import type { RootStackParamList, TopLevelTabParamList } from './types';
...
 
 export const TOP_LEVEL_TABS = {
+  BotChat: {
+    title: 'Bot',
+    analyticsLabel: 'Bot Chat',
+  },
   ChatList: {
-    title: 'Home',
-    analyticsLabel: 'Home',
+    title: 'Workspaces',
```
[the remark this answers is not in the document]
jamesacklin: Confirmed, and thanks — this matches what I found before opening, so I'll leave the thread open rather than resolve it: it's a genuine pre-merge gate and reviewers should see it.

It's written up under **Risks and impact** in the PR body. Scope is a bit wider than the flows you named — **30 files** under `.maestro` reference the literal `Home`. One addition worth flagging, because it fails *silently wrong* rather than timing out: `reliability/channels.yaml:131-136` guards a Back tap on `notVisible: '^Home$'`, which is now always true, so it pops one screen too far before failing its assert.

I deliberately did not fix it in this PR. The literal rename is mechanical, but the Contacts-tab removal, the iOS positional taps in `contacts-tab.yaml` calibrated for a 3-tab bar, and the signup flow landing on the bot DM (whose header is deliberately blank) are structural. I can't run Maestro Cloud locally to verify any of it, and 30 files of unverifiable churn that still wouldn't go green seemed worse than saying so plainly. It's tracked as separate work, already in progress.

Worth noting for whoever picks it up: neither `mobile-testing.yml` nor `mobile-reliability-nightly.yml` runs on pull requests, so nothing goes red here — the failure lands on `develop` after merge.

apps/tlon-mobile/src/hooks/useDeepLinkListener.ts
```diff
@@ -54,8 +54,12 @@ export const useDeepLinkListener = () => {
               const inviter = lure.inviterUserId;
               if (inviter) {
                 logger.log(`handling deep link to user`, inviter);
+                // Contacts is a stack screen now, not a tab, so seat it over
```
dnbrwstr: Extraneous comment

apps/tlon-mobile/src/hooks/useNotificationListener.ts
```diff
@@ -291,8 +291,8 @@ export default function useNotificationListener() {
     }
 
     async function goToContacts() {
-      const route = getTopLevelTabRoute('Contacts');
-      navigation.navigate(route.name, route.params, { pop: true });
+      // Contacts is a stack screen now, not a tab.
```
dnbrwstr: Rm comment

jamesacklin:
## Android verification (Pixel 7a, physical device)

The PR body says **"Android was not exercised."** It has been now, on a physical
Pixel 7a (Android 17) against a hosted ship with a bot, on `productionDebug`.

One fix came out of it (`02b8b98`); everything else below passed as written.

### What was tested

| Check | Result |
|---|---|
| Bottom tab bar renders | Pass — four tabs: Bot, Workspaces, Activity, Settings |
| Bot tab mounts only when `botDm.enabled` | Pass — a ship with no agent showed three tabs, no Bot |
| Switching between all four tabs | Pass |
| Drill-down and back inside a tab | Pass — Settings → Choose Model → back; tab bar correctly absent on the pushed screen and restored on return |
| `backBehavior="history"` | Pass — Settings → Workspaces → system back → Settings |
| Nav position survives a cold start (#6566) | Pass — left on Settings, `am force-stop`, relaunch, came back to Settings rather than the default tab |
| Revised Settings value labels | Pass — rows show their current value without a tap |
| `stim logs --errors` | Clean, exit 0 |

The one runtime warning is the `Query data cannot be undefined … ["baseVolumeLevel",{}]`
the body already reports from iOS. It reproduces identically on Android, which
supports the body's reading of it as pre-existing and unrelated to this merge.

### Fix: the Bot tab header had no name

`ChannelHeader` blanked the title for the bot DM when it backs the first tab,
leaving the avatar and caret with nothing to say whose conversation it is.
Restored the title; the subtitle stays suppressed.

Captures are cropped or blurred — this repository is public and the ship holds
real conversations. The blur leaves the screen header and the tab bar legible,
which is where the navigation evidence lives.

https://github.com/user-attachments/assets/a5e584cd-196c-4945-b2a4-0df6e70f9203

Bot tab header: before (no name) and after (openjames)

Cold start: on Settings before force-stop, and back on Settings after relaunch

Revised Settings rows showing their current values

packages/openclaw/src/monitor/agent-onboarding.ts
```diff
@@ -1878,6 +1978,39 @@ async function postFirstRunServices(
...
+  // be attributed to this run, and claiming one that is not ours would report
+  // the wrong note as the owner's first. Those stay a failure.
+  const botShip = normalizeShip(correlation.context.botShip);
+  return listed
+    .filter(
+      (note) =>
+        note.createdAt != null &&
+        note.createdAt >= earliest &&
+        note.createdBy != null &&
+        normalizeShip(note.createdBy) === botShip
```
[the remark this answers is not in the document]
jamesacklin: Fixed in 6add2bf — the window now closes before the listing (same clock slack the lower bound already carried) and the oldest match wins rather than the newest, which had leaned towards exactly the later entry you describe. There is no run-scoped identifier to key on, since the entry's title is written by the run itself, so this narrows the window rather than closing it; a baseline note id captured at enqueue would settle it properly, at the cost of a listing on the happy path.

packages/app/navigation/topLevelTabs.ts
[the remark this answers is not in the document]
jamesacklin: Correct, and deliberate — the Maestro suite is updated in #6587, which targets this branch rather than develop so the fix merges with the breakage. It is now well past the rename: `to-tab-root`/`workspaces-tab`/`settings-tab`/`contacts-screen` replace the removed subflows, `begin.yaml` confirms the ship through Settings, `03-signup` no longer backs out of the Bot tab, and the two Android regressions this shook out are fixed and verified on a physical device. The flows added to this branch since #6587 opened are carried over there too.

Worth stating the dependency plainly: #6587 has to land into this branch before this one merges to develop, or the coverage gap you describe is real.

jamesacklin:
## Android re-verification on the merged tree (`3b8733e`)

The earlier Android pass described `ddb729e`. Since then this branch took two
merges from `develop` — 159 commits, including the first-week campaign revert
(#6591) — which touched onboarding code paths. Re-ran the same checks on the
merged head, on the physical Pixel 7a (Android 17), `productionDebug`.

Everything still passes; nothing regressed.

| Check | Result on `3b8733e` |
|---|---|
| Bottom tab bar renders | Pass — Bot, Workspaces, Activity, Settings |
| Switching between all four tabs | Pass |
| Bot tab header shows the bot's name | Pass — reads `openjames`, so the header fix survived both merges |
| Drill-down and back inside a tab | Pass — Settings → Choose Model → back; tab bar absent on the pushed screen, restored on return |
| `backBehavior="history"` | Pass — Settings → Workspaces → system back → Settings |
| Nav position survives a cold start (#6566) | Pass — force-stopped on Settings, relaunched, returned to Settings |
| `stim logs --errors` | Clean, exit 0 |

One note on method, because it nearly invalidated the run: `stim stop` left the
pre-merge Metro alive (its file watcher had died with `ENOENT`) and started a
second one on a new port. The dev client was still pointed at the stale server.
Repointing it produced a fresh 9,963-module bundle; without that this would have
been a recording of pre-merge JavaScript.

Captures are blurred — this repository is public and the ship holds real
conversations. The blur leaves the screen header and the tab bar legible, which
is where the navigation evidence is.

https://github.com/user-attachments/assets/7ee6f231-9d20-4423-98e1-f4d71e0a8705

Cold start on the merged tree: Settings before force-stop, Settings after relaunch

packages/openclaw/src/monitor/agent-onboarding.ts
```diff
@@ -1869,6 +1969,49 @@ async function postFirstRunServices(
...
+  // Require the bot's own authorship as well; an entry missing either cannot
+  // be attributed to this run, and claiming one that is not ours would report
+  // the wrong note as the owner's first. Those stay a failure.
+  const botShip = normalizeShip(correlation.context.botShip);
+  return (
+    listed
+      .filter(
+        (note) =>
+          note.createdAt != null &&
+          note.createdAt >= earliest &&
```
dnbrwstr, line 2012: The added upper bound and oldest-first order make recovery safer, but time and bot authorship still do not correlate a note to this run. A delayed bot note from an earlier provision within the window could be selected first. Could we capture a baseline note ID at enqueue or use another run-specific marker? This is the remaining gap after the latest fix.

packages/shared/src/store/agentGroupOnboarding.ts
```diff
@@ -674,10 +774,17 @@ async function ensureIntroRequest(
     groupId,
     ...(isFirstGroup ? { isFirstGroup: true } : {}),
   });
+  // Sending needs the channel row locally, and on a fresh account the bot's
+  // DM may so far exist only on the ship — Hosting made it, sync has not
+  // caught up. Starting the DM from here is the same pending row the New
+  // Message flow uses; the first writ lands in the DM the ship already has.
+  if (channelType === 'dm') {
```
dnbrwstr: If the intro request already exists remotely but the Bot DM row has not synced locally, the `alreadyPosted` return above skips this upsert. The landing consumer does not wait for the DM row and can acknowledge the destination while `ChannelScreen` still renders `null`. Could we ensure the local DM exists before the early return, or wait for it before consuming the landing?

apps/tlon-mobile/src/App.main.tsx
```diff
@@ -253,10 +262,78 @@ function ConnectedNavigationContent({
...
+  // Backgrounded apps get evicted, and the relaunch that follows is a cold
+  // start: without this the navigator rebuilds from `initialRouteName` and the
+  // user loses their place. Resolved once, before the navigator mounts,
+  // because `initialState` is read only on the first render.
+  const [restoredState, setRestoredState] = useState<{
+    ready: boolean;
+    initialState?: NavigationState;
+  }>({ ready: false });
+
+  useEffect(() => {
```
dnbrwstr, line 279: Could the restore lifecycle move into a navigation-owned hook or component? `App.main` now reads saved state, decides whether onboarding overrides it, records a module-level restore flag, refreshes its age, and writes subsequent navigation changes; the tab navigator interprets the flag separately. Keeping that policy together would make future route changes easier to reason about and test.

apps/tlon-mobile/src/lib/navigationStatePersistence.ts
```diff
@@ -0,0 +1,225 @@
...
+export const NAVIGATION_STATE_MAX_AGE_MS = 24 * 60 * 60 * 1000;
+
+/**
+ * Params that carry a one-time instruction rather than a position. Almost
+ * nothing clears its own params after acting on them, so persisting these
+ * would replay the instruction on every launch inside the window: reopening an
+ * invite preview that was already dismissed, or reopening the composer and
+ * keyboard on a channel entered through a share link.
+ */
+const ONE_SHOT_PARAM_KEYS = [
```
dnbrwstr, line 20: This blacklist means every new one-shot route param is persisted for up to 24 hours unless its author remembers to add it here. Could restorable params/routes be an explicit contract, ideally close to route definitions, instead of relying on a central list of exceptions? This is a maintenance concern, not a claimed current failure.

packages/app/features/settings/bot/botSettingsDraftStore.ts
```diff
@@ -0,0 +1,146 @@
...
+    initialized: false,
+    baseline: EMPTY_VALUES,
+    draft: EMPTY_VALUES,
+    applying: false,
+    applyError: null,
+    setApplying: (applying) => set({ applying }),
+    setApplyError: (applyError) => set({ applyError }),
+    syncServerValues: (scopeKey, values) => {
+      const current = get();
+      const hasLocalChanges =
```
dnbrwstr, line 126: The store decides whether a draft has changes using whole-value `valuesEqual`, while `useBotSettingsDraft` independently computes pending fields one by one for the Apply UI. Can both paths use the same comparison/field definition? Otherwise adding a setting can make server-sync protection and the Apply bar disagree. The shared draft itself makes sense for the multi-screen flow; it is the duplicate definition of “dirty” I would simplify.

packages/app/hooks/useChannelNavigation.ts
```diff
@@ -85,25 +85,51 @@ export const useChannelNavigation = ({ channelId }: { channelId: string }) => {
     [navigation]
   );
 
+  const { navigation: rootNavigation } = useRootNavigation();
+  // Mounted as the BotChat tab, the nearest navigator is the tab bar, which
+  // has no push(); those screens live on the root stack, so go there instead.
+  const pushScreen = useCallback(
```
dnbrwstr, line 91: `pushScreen` discovers at runtime whether its typed navigation object has `push`, then casts both dispatch paths. Could the Bot-tab-versus-root-stack choice be made explicitly in a navigation helper or from the known route, with typed actions? This hook currently has to know the navigator topology and gives up type checking for every screen it opens.

packages/app/ui/components/Wayfinding/SplashSequence.tsx
```diff
@@ -2037,32 +2037,38 @@ export function GroupsPane(props: {
...
-  const groupInviteHasError = homeGroupInviteState === 'unavailable';
-  const { doCopy: copyHomeGroupInvite, didCopy: didCopyHomeGroupInvite } =
-    useCopy(homeGroupInviteUrl ?? '');
-  const shareHomeGroupInvite = useCallback(async () => {
-    if (!homeGroupInviteUrl) return;
+  // Invites connect people to the user rather than into one stable group:
+  // onboarding creates groups through the bot DM, so there is no fixed group
+  // to hand out. Verification runs in the background; an absent link is
+  // loading until that records a failure.
+  const inviteUrl = db.personalInviteLink.useValue();
```
dnbrwstr, line 2044: With `useHomeGroupInviteLink` removed, `initializeCachedHostedInviteLinks` still writes `homeGroupInviteLink`, but I cannot find a consumer of that stored value. Can we remove the home-group caching branch and its storage item if nothing still needs them? That would finish the cleanup from switching onboarding to personal invites.

dnbrwstr (approved):
Reviewed at 6add2bf. I left two onboarding edge-case comments and five code-quality/architecture suggestions. The Bot DM/group-context issue from my earlier draft has been fixed. I have not run an end-to-end device flow; this approval is based on source review.

packages/openclaw/src/monitor/agent-onboarding.ts
jamesacklin: Done in f1fd3e1 — the notebook's highest note id is now read after the claim and before the enqueue, and recovery requires an id above it. Not taken at the two recovery call sites, where a correlation is rebuilt from a persisted record after a restart: by then the run's own note may already have landed, and a baseline there would exclude the entry being looked for. Those keep the weaker time-and-author test; carrying the baseline across a restart needs it in the run record, which is a larger change than this.

packages/shared/src/store/agentGroupOnboarding.ts
jamesacklin: Done in f1fd3e1 — `upsertDmChannel` moved above the `alreadyPosted` return, so the local row exists whether or not the request still needs sending. It is idempotent, so an already-synced DM comes back untouched.

packages/app/features/settings/bot/botSettingsDraftStore.ts
jamesacklin: Done in f1fd3e1 — `getPendingFields` moves into the store as the single definition of changed, `hasPendingChanges` derives from it, and all callers use it. There was a third copy too: `useBotSettingsDraft` had its own `!valuesEqual` for the sync effect, which had to agree with the store's guard and would have drifted the same way. The one remaining `valuesEqual` compares the baseline to an incoming server snapshot — a different question, still whole-value.

packages/app/ui/components/Wayfinding/SplashSequence.tsx
jamesacklin: Done in f1fd3e1 — confirmed one writer and no readers, so the branch and `homeGroupInviteLink` are gone, along with the `homeGroupInviteToken` plumbing that fed it through `assignShipToUser`, `reserveNode` and the boot sequence. `homeGroupLureToken` stays in the hosting response type, since that is the server's shape rather than ours.

apps/tlon-mobile/src/App.main.tsx
jamesacklin: Agreed on the diagnosis — the policy is split across `App.main`, a module-level flag and the tab navigator, and that is hard to reason about. Not changed here: pulling the restore lifecycle into a navigation-owned hook is a behavioural change to app startup, and the failure mode if it is wrong is landing users on the wrong screen or losing the saved position entirely. That wants the author's intent and a device run, not a blind refactor from review. Leaving this open for @dnbrwstr and the PR author to settle.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

apps/tlon-mobile/src/lib/navigationStatePersistence.ts
jamesacklin: Fair, and it is the kind of list that rots quietly — a new one-shot param is persisted for 24h and nobody finds out until a stale instruction replays. Not changed here, because an explicit contract worth having belongs next to the route param types, and reshaping those on this PR is a wider change than the review is asking for. Worth noting a cheaper middle: the omission could at least be made loud rather than silent, by keying the contract off the param types so a new param has to declare itself. Leaving open for the author to choose the shape.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

packages/app/hooks/useChannelNavigation.ts
jamesacklin: Agreed the runtime `push` probe plus casts is the weak point — it makes the hook know the navigator topology and drops type checking for every screen it opens. Not changed here: doing it properly means the Bot-tab-versus-root-stack choice becomes explicit in the navigation layer, which is the same structural question as the restore-lifecycle thread above and is better decided once, together, by the author. Leaving open.

_🤖 Reviewed by [Claude Code](https://claude.com/claude-code)_

packages/openclaw/src/monitor/agent-onboarding.ts
```diff
@@ -1982,6 +2140,7 @@ function setFirstRunCorrelation(
     jobId: string;
     notebookName?: string;
     enqueuedAt: number;
+    baselineNoteId?: number;
```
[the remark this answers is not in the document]
jamesacklin: Correct, and worse than it looks — the value was dropped on the floor, so the baseline was inert from the moment it landed and recovery went on accepting notes that predate the run. Fixed in b9f230f. My test missed it because it only asserted the listing happened and happened before the enqueue, both of which were true; it now reads the baseline back off the correlation, and deleting the one storing line fails it with `expected undefined to be 9`.

packages/openclaw/src/monitor/agent-onboarding.ts
```diff
@@ -2074,6 +2237,26 @@ async function ensureFirstRunEnqueued(
...
+        .then((entries) =>
+          entries.reduce<number | undefined>(
+            (highest, entry) =>
+              highest === undefined || entry.noteId > highest
+                ? entry.noteId
+                : highest,
+            undefined
+          )
+        )
+        .catch(() => undefined)
```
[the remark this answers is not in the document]
jamesacklin: Fixed in b9f230f — cancellation propagates now, while an ordinary listing failure stays best effort. I also release the claim on the way out, for the same reason a rejected enqueue does: otherwise the next pass reads it as another live attempt until the grace window expires. Covered by a test that aborts mid listing and asserts no run was enqueued.

PR #6589, opened 2026-09-22
fix(chat): place native cursor after mention and command selections

wca4a:
## Summary

Selecting a nickname was reported to leave the caret attached to the mention, so pressing Space broke its highlight. This change explicitly places the native cursor after the replacement and its trailing space; slash-command selection uses the same path. The reporter says the patched build works, but the recorded failure was not reproduced on the current unpatched development build.

## Changes

Use TextInput’s `setSelection` command after rendering the replacement, while leaving text ownership with the decorated children. This preserves the existing protection against duplicating child text. Four component tests cover the cursor command for iOS and Android, continued typing after a mention, and slash-command selection.

## How did I test?

On an iPhone 17 simulator running iOS 26.5, connected to a disposable local ship: enter `~ja`, select `james`, press the onscreen Space key, and type `hello`. The mention stays highlighted and the caret advances beyond it. Runtime error logs were empty.

Web: verified in desktop Chrome at 1440×1000 against the same local ship. Enter `~ja`, click `james`, press Space, and type `hello`: the mention remains highlighted and the textarea selection is collapsed at offset 13, after `@james  hello`.

**Coverage limit:** Android device verification remains outstanding; the native command tests use mocked inputs. The clips are smoke tests, not proof of a before/after fix. Web does not execute the native-only cursor command.

The web dev session emitted startup/configuration warnings (including a missing analytics token and an undefined settings query); no additional console errors appeared during the mention interaction.

## Risks and impact

  - [ ] Onboarding
  - [ ] State / providers
  - [ ] Message sync
  - [ ] Channel display
  - [ ] Notifications
  - [x] Other: native chat composer cursor placement

A native timing regression could place subsequent input at the wrong position after selecting a mention or command. The web selection path is unchanged.

## Rollback plan

Revert this PR. No data migration or backend change.

### iOS

Both clips preserve the nickname mention. The patched clip also shows continued typing.

| Before (unpatched dev build) | After (patched dev build) |
|---|---|
| https://github.com/user-attachments/assets/ea4cea18-ea58-4a33-b311-4ff7dab004fa | https://github.com/user-attachments/assets/5beda95f-810e-44bd-a3e9-5e6a4f484634 |

### Web

Select the nickname, press Space, and continue typing; the mention highlight remains intact.

https://github.com/user-attachments/assets/a90ccd39-8f7b-4196-977e-32920ca0a183


janicduplessis:
I am having trouble seeing the bug in the before video, but code change looks good!

janicduplessis (approved):

